PR #209, opened 2015-02-10
Constraint classes

ohickey:
The main part of this commit is a conversion of constraints/lb_boundaries into a somewhat unified class structure. The electrostatic and magnetic constraints should probably be separated as they make the constraint class significantly more  convoluted and prone to future errors. I would note that there was a small bug in GHOST_TRACE where an MPI_BARRIER caused Espresso to freeze under certain myconfigs. Also using the config nocheck-maxset the test case nsquared.tcl fails to work, but this is also true for the old master and seems to only be a result of the debug outputs. I have also included an incredibly ugly hack to make the current master compile and work, with the exception of cython. If one is interested only in fixing the latter problem it is just the commit 061b41160b9322f091abcadefacdbfcfc69f8062


fweik:
Working on it.

PR #214, opened 2015-03-12
fix build with cython-0.22

junghans:
Reported to me by @tomspur, found on Fedora, details here: http://koji.fedoraproject.org/koji/buildinfo?buildID=619654.

It might be time for a new release, too!

### DIFF
--- a/src/python/espressomd/particle_data.pxd
+++ b/src/python/espressomd/particle_data.pxd
@@ -185,5 +185,5 @@ cdef class ParticleHandle(object):
   cdef public int id
   cdef bint valid
   cdef Particle particleData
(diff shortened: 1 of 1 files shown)


junghans:
@tomspur , you could back-port that patch or disable the python interface for the rpm in fedora.

tomspur:
@junghans thanks for the quick fix :)
It seems to build [just fine](http://koji.fedoraproject.org/koji/taskinfo?taskID=9212042), but the mpich update broke some [environment variables](https://lists.fedoraproject.org/pipermail/devel/2015-March/209000.html) so it will take a bit untill that is also sorted out.

PR #218, opened 2015-04-01
CPU Thermostat

mkuron:
This pull request implements a CPU thermostat. This gives Espresso the ability to thermalize the CPU, which is quite a unique feature in MD simulation software, which typically only allows for thermalization of the objects simulated (particles, fluids, ...), but not of the hardware running the simulation.

### DIFF
--- a/doc/ug/setup.tex
+++ b/doc/ug/setup.tex
@@ -442,6 +442,22 @@ Be aware that this feature is neither properly examined for all
 systems nor is it maintained regularly. If you use it and notice
 strange behaviour, please contribute to solving the problem.
 
+\subsection{CPU thermostat}
+
+\begin{essyntax}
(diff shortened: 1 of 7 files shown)


mkuron:
You're right about the error message, it doesn't compile on my AIX 7.1 laptop. I'll see what I can come up with.

rempferg:
I won't pull this without documentation.

mkuron:
Give this update a try. The CPU thermostat now integrates with the other thermostats much more nicely. Here's an example that heats your simulation machine to a comfy 60°C:
```
thermostat cpu [expr 60+273]
setmd skin 0
setmd time_step 1
part 0 pos 0 0 0 type 1
integrate 1
```

fweik:
This is a portability nightmare. Needs some serious improvement in the future.

PR #219, opened 2015-04-10
Harmonic Dumbbell Bond

hmenke:


fweik:
Hey, this breaks the usual harminic bonds, because the names match in the first word. Please check again if the tcl works as expected. Breaks the maxset testcase because of that, please fix.

hmenke:
The problem was in the order of REGISTER_BOND.  "Obviously", the one with the longer name has to come last.

fweik:
Obviously! Da!

PR #300, opened 2015-05-27
Rigid bond added to python interface

jonaslandsgesell:


jonaslandsgesell:
Tests on Travis failed due to an OS error on their servers.

PR #302, opened 2015-05-27
Porting particle properties nearly complete

hmenke:


hmenke:
I had a look into the "failing" build and it turned out that the `lb.tcl`testcase failed with a too large deviation from the reference data.  As this only happened for the `with-mpi=false myconfig=maxset` case, I suspect this is a statistical error.  Also the code in my pull request is not even covered by the`lb.tcl`testcase.  I think you can safely merge it.

hmenke:
[Build Status](https://travis-ci.org/hmenke/espresso)

PR #307, opened 2015-05-27
Harmonic dumbbell

KaiSzuttor:


jonaslandsgesell:
The bug is not reproducible, so it is likely a hardware or OS problem. Cython is not compiling.

KaiSzuttor:
Thats not a hardware or cython issue, its more related to the disk space limitation on travis.

Issue #337, opened 2015-06-11
blockfile reading


KaiSzuttor:
When i restart simulation from blockfile checkpoint the following error occurs:
blockfile_read_auto_variable: setmd sd_radius -1.0 reported: radius must be > 0.

without using any feature of stokesian dynamics...

HaoZeke:
Yeah, I've been seeing this error too. Even in the sample scripts.

fweik:
Fixed by #338.

RudolfWeeber:
Looks good, except that project version needs to be changed in CMakeList.txt and maybe also in src/config/version.hpp.in
Once that's done, it's ready

RudolfWeeber:
Sorry, wrong ticket

PR #338, opened 2015-06-15
fixed that build failed if h5md is not present

dschwoerer:
If h5md is not present and configure is run with:
./configure  --with-h5md=no
The build failed as the header file was still tried to be included.


dschwoerer:
The fix for issue #337 allows a sd_radius of -1
Other negative values are not allowed, in order to avoid to accidentally set a negative timestep.

dschwoerer:
I didn't change anything in the lb part, therefore i think it should be save to merge anyway.

fweik:
Have you tried --without-h5md ?

dschwoerer:
I tried 
./configure  --with-h5md=no
Without that, configure failed.
With that, configure succeeded, but the build failed, as the #include h5md.h was outside the #ifdef, and therefore the compiler looked for it, irrespective of the configuration.

PR #340, opened 2015-06-16
Python: all dipolar interaction methods (2x direct sum, p3m, mdlc)

RudolfWeeber:
Ext magn field will come after constraint refactor


rempferg:
Merged via rebase.

PR #361, opened 2015-07-10
Virtual sites, dipoles, rotation, energy minimization

RudolfWeeber:
* Dipolar direct summation on gpu
* Switching of rotation per particle per axis
* Virtual sites relative: Rotational degrees of vs are now integrated, unless specified otherwise using ROTATION_PER_PARTICLE
* Steepest-descent propagator now acts on rotational degrees
* VS_Relative: Placing vs at the same position as non-virtual particle
* Fixes to blockfile support for dipoles


fweik:
Does not build with the maxset (engine_lbgpu.tcl failed, which is not checked on travis). Also there seem to be some debug messages still in.

RudolfWeeber:
Did you build in a clean directory? For me the probem vanished after I cleared the build directory.
The debug message is removed.
I also merged Joost's thermostat fix and cleaned up and deduplicated the Langevin thermostat.

fweik:
The error changed.

```
$ git clone https://github.com/RudolfWeeber/espresso.git 
$ cd espresso && mkdir build && cd build && ln -s ../maintainer/jenkins/configs/maxset.hpp myconfig.hpp 
$ ../bootstrap.sh && ../configure && make -j32
[...]
Error compiling Cython file:
------------------------------------------------------------
...
#  
import sys

cdef extern from "communication.hpp":
  void mpi_init(int* argc = NULL, char ***argv = NULL)
              ^
------------------------------------------------------------

/scratch/fweik/espresso/src/python/espressomd/_init.pyx:22:15: Extern functions cannot have default arguments values.
```

PR #377, opened 2015-07-29
 Review and correction of pore distance calculations. 

lufische:
- Corrected copy-paste-typo, leading to incorrect values in the force calculation (in the order of double numeric limit).
- Suggest to delete irrelevant calculation cases. Since the pore is created with (numerical) infinite outer radius (c2_or), the only cases where the corresponding condition is true is r=std::numeric_limits<double>max(), which should not occur in simulation.s

### DIFF
--- a/src/core/constraint.cpp
+++ b/src/core/constraint.cpp
@@ -1088,6 +1088,9 @@ void calculate_pore_dist(Particle *p1, double ppos[3], Particle *c_p, Constraint
   /* upper left smoothing circle */
   if (p2_z <= c1_z && r >= c1_or ) {
     /* distance from the smoothing center */
+    // Since: c2_or_or seems to be set to numeric_limits<double>max(), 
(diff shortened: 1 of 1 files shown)


fweik:
Actually, the outer radius can be set by the user (cf. src/tcl/constraint_tcl.cpp:879). It seems to me that the condition in front of the code blocks you refer to covers not all cases. I'll have a look again tomorrow. But I agree that the code as it is now is not correct. I also know that this has caused problems in the past. As for the typo, I think you are correct, that is obviously wrong.

lufische:
I'm sorry: the documentation mislead me - it is indeed configurable. 
I suggest an addition to the ug, listing that option which I can prepare.
Given this I have to admit, that I can't think of any physical system, where a pore with smoothed inner edges AND smoothed outer edges is applicable. When that bit of code is reviewed, it might be worth to discuss this again.

PR #384, opened 2015-07-31
Make Espresso build w/o h5md. Closes #373

### LINKED ISSUE #373: Configure should autodetect the presence of libhdf5 and switch the feature on or off accordingly

fweik:


KaiSzuttor:
The build system never builds h5md because of #ifdef H5MD. How can this be repaired?

PR #403, opened 2015-08-27
h5md build system repair

KaiSzuttor:


KaiSzuttor:
Was not that obvious! First commit didnt fix it.

fweik:
Don't we need the AC_CONDITIONAL to get H5MD defined?

PR #413, opened 2015-09-21
Shan-Chen on CPU

Marcello-Sega:
This patch introduces the CPU version of Shan-Chen bi-component fluid.


Marcello-Sega:
travis stoped on virtual-sites-rotation.tcl because it took too long to execute it. I reopened the request hoping to find a less loaded machine...

Marcello-Sega:
Found a bug in the MPI implementation. Will pull again once solved.

PR #415, opened 2015-09-23
disable mpi in order to use clang for clang builds in travis

dschwoerer:
I am not quite sure, but it seems to me that travis uses mpi for most build, although I assumed that it should use clang.


Marcello-Sega:
Shouldn't mpicc / mpic++ rely also on clang when requesting that compiler? I think it should be a good idea to test the parallel version also for clang as an underlying compiler.

dschwoerer:
I don't really know how the mpicc/mpic++ compilers work and whether they change there behaviour depending on the CC/CXX flags.
But my point was that with-mpi=false flag implied that the mpi compiler should not be used.
As compilation fails now, I think we ether should add the clang without mpi builds, or replace them, in case the mpi compiler is independent of the CC/CXX flag (which I don't know)

Marcello-Sega:
Well, you can understand which compiler is used by mpicc by calling

      mpicc -v
To switch off mpi in espresso when configuring you should give the flag

      --without-mpi
and not

      --with-mpi=false
 The latter doesn't work.
This being said, I still think it is good to test the parallel version with clang.

dschwoerer:
According to SO, this should add mpi with clang to the builds.
http://stackoverflow.com/questions/14464554/is-there-an-easy-way-to-use-clang-with-open-mpi

fweik:
David, you solution seems correct to me. I will merge this, but let me fix the things clang complains about first. Otherwise all the tests will always fail.

fweik:
Marcello, you are confusing arguments to configure to environment variables evaluated by the build script (which in turn calls configure with the correct arguments.) Davids version disables mpi.

Marcello-Sega:
I see the point.  Still, I think one should check clang in parallel. 




M






—

dschwoerer:
I think this is what the new test does. At first I dropped the parallel tests, but thanks to your suggestion, I added both non parallel and parallel tests for clang.
One might reduce the number of cases, but I did not feel comfortable doing this.

mkuron:
This pull request should be obsolete with CMake. It doesn't use the MPI compiler wrappers anymore, uses whatever your main compiler is, and tells it to link against the MPI libraries.

fweik:
I agree with Michael.

PR #437, opened 2015-10-15
Added codecov support

Marcello-Sega:
This might be pretty much useful to pinpoint which parts of the code need a test to be added to the testsuite. 

After the discussion during the developer meeting, I looked around for some solution that could help tracing how good the testcases are. There are several options around to integrate the CI builds with output from code coverage tools, such as [gcov](https://en.wikipedia.org/wiki/Gcov). I found particularly simple to add in travis support for [codecov](https://codecov.io).

With this patch I added code coverage to one of the element of the travis matrix. The outcome can be seen, for example, in this snapshot:

screen shot 2015-10-15 at 13 51 26

You can have a look directly at [the annotated code files of this patch](https://codecov.io/github/Marcello-Sega/espresso/src/core/integrate.cpp?ref=1c74fd659c32bdde7b5c4900527e2507e98194fe) (this is an example of the coverage of `integrate.cpp` with the `maxset` option,  which for example does not cover NPT.

Codecov support is free (as in free beer) for open source projects.

There is also a [plugin for chrome](https://chrome.google.com/webstore/detail/codecov-extension/keefkhehidemnokodkdkejapdgfjmijf?hl=en-US), for example, to show directly the code coverage in the github page, as in these snapshots

screen shot 2015-10-15 at 13 47 42

screen shot 2015-10-15 at 13 47 52

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ before_install:
  - sudo pip install cython --upgrade 
  - if [ "$with_mpi" != false ]; then sudo apt-get install libopenmpi-dev openmpi-bin; fi
  - if [ "$with_fftw" != false ]; then sudo apt-get install libfftw3-dev; fi
+ - pip install --user codecov
 
 env:
   - myconfig=default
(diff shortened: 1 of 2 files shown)


fweik:
That looks great!

PR #451, opened 2015-11-02
2 additional collision_detection schemes, cleanup, memleak fix for GHOSTS_HAVE_BONDS

RudolfWeeber:


src/core/collision.cpp
```diff
@@ -21,12 +21,15 @@
...
 #include "communication.hpp"
 #include "errorhandling.hpp"
+#include "grid.hpp"
+#include "domain_decomposition.hpp"
+
 
 using namespace std;
 
-// #define DEBUG
+#define DEBUG
```
fweik: Are you sure this is correct?

RudolfWeeber: > >  
> >  using namespace std;
> >  
> > -// #define DEBUG
> > +#define DEBUG
> 
> Are you sure this is correct?
It does nothing, as it is never used in the rest of the file and the
.cpp-file is not includexd anywhere else.
I still remoted it to avoid confusion.

PR #453, opened 2015-11-06
Rotational motion core fixes: adding mass and dimensionless model related parameters

psci2195:
Impact has been tested. Without the provided changes, a single possible equilibrium outcome is a suspended set of nanoparticles ("gas"). New changes introduces following use cases:

== Given ===
Device: Tesla C2070.
Number of particles: 5000.
=== When ===
inter magnetic 1.0 dds-gpu 
=== Then ===
Suspended set of nanoparticles ("gas")
=== When ===
inter magnetic 10.0 dds-gpu 
=== Then ===
Dense ferrofluid aggregate with linear chains.


RudolfWeeber:
commit b22b58d1c2173dc2cd581f0e1f7ac178603cb8dc:
This one fixes a bug in the dipolar direct sum on the gpu. Namely, the Bjerrum length was ignored.

commit 0250eb27995d1a38842bd43c8f8eedec59cb119d
This one is probably not correct and breaks existing simulations (It fails the testsuite).
Please explain further, what this is supposed to do.
The division by time_step in the energy calculation and the thermostat, e.g., is needed, because at this point in the integration, the forces are scaled.

As a side node: If you push new commits to an existing pull request, which qualitatively alter, what the pull request covers, please change the request's title to reflect that.

I included your fix for the dipolar direct sum, together with other fixes and a test case in pull request #456.

fweik:
I agree with Rudolf. The last commit does not seem correct to me.

psci2195:
Dear Dr. Rudolf Weeber,

Thank you for the message and advises.

Sorry for a confusion. You are right. We've investigated and now fully
understood the Espresso dimensionless model. We
made a refactoring - most core code has been reverted. All tests were
successful for default configuration. For some configurations couple of
tests are failing. We are fixing this now.

We propose to keep some fixes in this pull request (#453):

a/ 0.4 for rigid sphere moment of inertia (=2/5 M * SIGMA^2). The test
langevin.tcl was ok with the ROTATION feature enabled.
b/ masses feature enable for rotational motion
c/ fix of the call of the gaussian_random (typo)

We can provide mathematical/physical explanation (in doc file) of reasons
why these dimensionless-model-related corrections are proposing. It is well
aligned with publication (only different of Espresso dimensionless model
from this work is delta_t re-normalization of a velocity):

@article{Wang2003,
author = {Wang, Zuowei and Holm, Christian},
doi = {10.1103/PhysRevE.68.041401},
issn = {1063-651X},
journal = {Physical Review E},
month = oct,
number = {4},
pages = {041401},
title = {{Structure and magnetic properties of polydisperse ferrofluids: A
molecular dynamics study}},
tmpurl = {http://link.aps.org/doi/10.1103/PhysRevE.68.041401},
volume = {68},
year = {2003}
}


*Sincerely yours,Dr. Bogdan Tanygin*

psci2195:
Please note, that tests langevin.tcl and mass-and-intertia.tcl are passing. They are very sensitive to any changes of rotational motion dynamics and thermostat including distribution of energy among degrees of freedom including moment of inertia anisotropy. Failing of other tests can be related to already calculated reference data used in them which were calculated based on the rotational dimensionless model which we are trying to fix. We will continue investigation of failing of those tests.

Latest comment concerning 0250eb2 (already reverted here). There was same situation. Analyze.tcl test simply used already calculated energy using re-normalized velocity. When we call [analyze kinetic energy] within integration we received proper kT/2 per each degree of freedom in the console during whole simulation cycle. Hence, our commit was not incorrect physically by rather broke existing conventions in Espresso (which we should avoid, of course). But generally, when velocity is already dimensionless, additional re-normalization of it using delta_t, lead to need to keep this in mind in case of adding of each and every new physical equation to Espresso. Same as in case of viscous friction force (in thermostat.cpp/hpp) adding of any new physical equation which contain velocity should be supplemented by division "/delta_t" with same power as velocity. Otherwise, relation with real dimensions of real physical values can be disconnected.

Generally, maybe it make sense to update user guide by some table with normalized coefficients which maps real physical values to dimensionless ones, for different possibilities if there is more than one (we can help with this if needed). For instance if all masses are normalized to mass of some type of particle "M_dimless = M_real / M0" and distance is normalized to diameter of some type of particles "SIGMA_dimless = SIGMA_real / SIGMA0" then it automatically lead to momentum of inertia changes "I_dimless = 2/5 * M_real * SIGMA_real^2 / (M0 * SIGMA0^2)", hence "2/5" should be kept in simulation dynamics equations. Otherwise, for instance, rotational motion and translation motion will be physically not properly synchronized in time; also without this pull request increase of M_dimless does not impact momentum of inertia magnitude. Present pull request fixes this.

psci2195:
All tests have been passed.

psci2195:
Just in case: a clarification concerning the proposed dimensionless model (with a physically correct temporal synchronization of rotational & translational motion). Regardless of specific dimensionless model selection, the numerical value "I_dimless = 0.4" should be present in some other numeric parameter (hardcoded in the source code).

From http://arxiv.org/abs/cond-mat/0203445

motion_equations

uschille:
Is the 'hardcoded' value of 0.4 still valid if one specifies anisotropic moment of inertia tensor? Also, I can see the motivation for the scaling with the mass, but it changes the units of `p.rinertia` implicitly.

Marcello-Sega:
Hi,

sorry to jump in the middle of this. Just about the moment of inertia:
for an homogeneous sphere of radius R the inertia moment  is 2/5 mass
R^2, and sigma=2R is the diameter.

In dimensionless units the inertia tensor is 2/5 R^2/sigma^2 =   2/5 *
1/4 = 0.1, not 0.4.

Am I wrong?

Cheers,

M.




On Tue, Nov 17, 2015 at 10:46 PM, Kyiv Ferrofluid Research Group
<notifications@github.com> wrote:
>
> Agree, indeed, both proposed & existing way of the rotational properties definition requires (in order to have a self-consistent dimensionless model) some implicit relation between parameters (even per separate particle) setting by user. OK, we can make a design decision before we'll update this pull request. We can propose the following user experience for your review.
>
> Features configuration & corresponding proposed changes:
>
> - #undef MASS #define MASS
> #undef ROTATIONAL_INERTIA to hardcode p.rinertia[0] = p.rinertia[1] = p.rinertia[2] = 0.4. This approach treats default particle as a solid sphere. to hardcode p.rinertia[0] = p.rinertia[1] = p.rinertia[2] = 0.4 * PMASS(part[i]). This approach treats default particle as a solid sphere with a given mass.
> #define ROTATIONAL_INERTIA no changes in a code are required; one may add a sentence to User Guide with advises concerning values of a moment of inertia (to be defined by user in rinertia parameters) & how it should be defined in the Espresso dimensionless model, e.g. for a hollow or solid sphere, etc. In this case, a particle shape can be arbitrary. Here, a per-particle contradiction of particle properties is possible. The moment of inertia defined by user can be not proportional to the given particle mass, also, defined by the user. Solution: per-particle properties check with some accuracy & error message can be implemented. In this case, a particle shape can be arbitrary.
>
> Make sense ?
>
> —
> Reply to this email directly or view it on GitHub.




-- 
University of Vienna, Institute of Computational Physics

psci2195:
Dear Marcello,

Thanks, good point! Yes, depending on a way how we define a dimensionless distance, it can be 0.4 (a distance measured in a particle radius units) or 0.1 (a distance measured in a particle diameter units) or other value; and one makes a physical sense to add such coefficient to a rotational motion equation. So, the important outcome of your comment is that the rotational motion equation coefficient value depends on an overall units selection.

As we understood, the conventional Espresso selection is a distance normalized to a particle diameter. _However, according to User Guide, units selection should be quite flexible_.

Thanks again for your comment. We will think how to handle this coefficient in an elegant way, otherwise impact on overall dimensionless model and on other equations can be significant and many equations should be changed.

Maybe, it makes sense (simplest) to stick to the diameter in terms of distance units (to sacrifice a bit a flexibility of an units selection).

However, such kind of convention should be made taking into account full picture of Espresso.

Regards,
B.

RudolfWeeber:
I don't understand, why the code has to be changed at all to get the
correct translational and relaxational relaxation and diffusion behaviour for spherical particles.

Why can't one set 
mass =m
Inertial moment =2/5 m r^2


And for the thermostat
thermostat langevin $kT $gamma_t $gamma_r
where
the translaitonal friciont is gama_t =6 pi eta r
and the rotational one is gamma_r =8 pi eta r^2
@jdegraaf: What do you think?

BTW: I'm going to make a separate pull request for the fix of the
thermostat noises, but it still breaks the testsuite in the maxset
config.

jdegraaf:
Hi Rudolf,

Yes, you can also plug in mass and inertia parameters into the relaxation
times. But

(i) that is not what was advertised in the UG, in which it expressly stated
'these are friction coefficients' and
(ii) in the original code the rotational and translational relaxation could
not be independently set, hence you could not even do what you suggested.

I vastly prefer that they are friction coefficients, which leads to far
fewer IO (Idiot Operator) errors being made than when you have to specify
the relaxation times.

Also, what are you talking about "I'm going make a separate ... in the
maxset config"? I thought all of the fixes had been pulled a long time ago
and did not break anything. Why is this only now coming up?

Best, Joost

Marcello-Sega:
I think there should be no particular hardcoded inertia moment in the code.
So, I would suggest to keep by default (I thought it was like that) mass=1
and inertia tensor=diag(1,1,1), since particles are pointlike objects and
the concept of their size is accessory.

This way, by default accelerations = forces and angular accelerations =
torques.

M


On Wed, Nov 18, 2015 at 9:38 AM, jdegraaf <notifications@github.com> wrote:

> Hi Rudolf,
>
> Yes, you can also plug in mass and inertia parameters into the relaxation
> times. But
>
> (i) that is not what was advertised in the UG, in which it expressly stated
> 'these are friction coefficients' and
> (ii) in the original code the rotational and translational relaxation could
> not be independently set, hence you could not even do what you suggested.
>
> I vastly prefer that they are friction coefficients, which leads to far
> fewer IO (Idiot Operator) errors being made than when you have to specify
> the relaxation times.
>
> Also, what are you talking about "I'm going make a separate ... in the
> maxset config"? I thought all of the fixes had been pulled a long time ago
> and did not break anything. Why is this only now coming up?
>
> Best, Joost
>
> On 18 November 2015 at 07:27, RudolfWeeber <notifications@github.com>
> wrote:
>
> > I don't understand, why the code has to be changed at all to get the
> > correct translational and relaxational relaxation and diffusion behaviour
> > for spherical particles.
> >
> > Why can't one set
> > mass =m
> > Inertial moment =2/5 m r^2
> >
> >
> > And for the thermostat
> > thermostat langevin $kT $gamma_t $gamma_r
> > where
> > the translaitonal friciont is gama_t =6 pi eta r
> > and the rotational one is gamma_r =8 pi eta r^2
> > @jdegraaf: What do you think?
> >
> > BTW: I'm going to make a separate pull request for the fix of the
> > thermostat noises, but it still breaks the testsuite in the maxset
> > config.
> >
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/espressomd/espresso/pull/453#issuecomment-157631017
> >.
>
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/espressomd/espresso/pull/453#issuecomment-157645037>.
>



-- 
University of Vienna

uschille:
I agree very much with @RudolfWeeber. @jdegraaf: Isn't Rudolf's suggestion exactly what you say? His gammas are friction coefficients. Or are we talking about a different convention, i.e., is the force F=gamma*v or F=m*gamma*v?

In any case there are two things here. Apart from the friction, it seems the code automatically scales the moment of inertia tensor when the mass is changed. As Rudolf says, this can be done on the script level (keeping in mind that "ESPResSo does not do the physics for you").

Also agree with @Marcello-Sega.

uschille:
I think the best way forward is to separate the different aspects into individual commits and pull requests.

psci2195:
ok, agree, then, let's close this pull request meanwhile

RudolfWeeber:
On Wed, Nov 18, 2015 at 12:57:18AM -0800, Ulf Schiller wrote:
> I think the best way forward is to separate the different aspects into individual commits and pull requests.

Agreed.

PR #455, opened 2015-11-10
Implementation of Ladd Boundaries to lbgpu

rbardak:


src/core/lb.cpp
```diff
@@ -1254,10 +1263,13 @@ int lb_lbnode_get_pi_neq(int* ind, double* p_pi) {
 int lb_lbnode_get_boundary(int* ind, int* p_boundary) {
     if (lattice_switch & LATTICE_LB_GPU) {
 #ifdef LB_GPU
-        unsigned int host_flag;
+        int host_flag;
```
mkuron: This looks dubious. You declare the variable, but don't ever assign a value to it. So it has an undefined value, but you do an `if` below based on that value?

src/core/lbgpu.cpp
```diff
@@ -266,6 +270,33 @@ void lb_reinit_parameters_gpu() {
       /* Eq. (81) Duenweg, Schiller, Ladd, PRE 76(3):036704 (2007). */
       lbpar_gpu.gamma_bulk[ii] = 1. - 2./(9.*lbpar_gpu.bulk_viscosity[ii]*lbpar_gpu.tau/(lbpar_gpu.agrid*lbpar_gpu.agrid) + 1.);
     }
+    
+    
+    if (lbpar_gpu.is_TRT) {
```
mkuron: Looks like you already have some of Georg's TRT changes in your code. These are not on the master yet.

src/core/lb.cpp
```diff
@@ -1254,10 +1263,13 @@ int lb_lbnode_get_pi_neq(int* ind, double* p_pi) {
...
         int single_nodeindex = ind[0] + ind[1]*lbpar_gpu.dim_x + ind[2]*lbpar_gpu.dim_x*lbpar_gpu.dim_y;
         lb_get_boundary_flag_GPU(single_nodeindex, &host_flag);
```
rbardak: host_flag is set by this function. A wrapper function is used since the data has to be read via a cuda Kernelcall.

fweik:
Since this is based on the modified LB which is not merged yet this has to wait until this is fixed and merged.

rbardak:
outdated.

PR #457, opened 2015-11-10
Deprecation warnings for undocumented analyze cmds

rbardak:
Espresso will write out a warning to stderr whenever an analyze command is called, for which there is no documentation, nor a test case available. See open issues.

I also noticed that there might be problems with naming future commands if ARG1_IS_S is used, since it only requires the given command string to only match the first letters of the required string instead of matching exactly with the required string. This might cause the wrong function to be called. It can be dealt with by using ARG1_IS_S_EXACT, but since I don't know whether or not this is intentional, I didn't changed it.

### DIFF
--- a/src/tcl/statistics_tcl.cpp
+++ b/src/tcl/statistics_tcl.cpp
@@ -2737,46 +2737,48 @@ int tclcommand_analyze(ClientData data, Tcl_Interp *interp, int argc, char **arg
   else if (ARG1_IS_S(name)) err = parser(interp, argc - 2, argv + 2)
 #define REGISTER_ANALYSIS_W_ARG(name, parser, arg)			\
   else if (ARG1_IS_S(name)) err = parser(interp, arg, argc - 2, argv + 2)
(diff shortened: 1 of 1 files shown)


mkuron:
Some of these already don't exist anymore:
```
int tclcommand_analyze_parse_and_print_p_IK1(Tcl_Interp *interp, int argc, char **argv)
{
  Tcl_AppendResult(interp, "analyze p_IK no longer exists", (char *)NULL);
  return (TCL_ERROR);
}
 
int tclcommand_analyze_parse_bins(Tcl_Interp *interp, int argc, char **argv)
{
  Tcl_AppendResult(interp, "analyze bins no longer exists", (char *)NULL);
  return (TCL_ERROR);
}
```
[src/tcl/pressure_tcl.cpp](https://github.com/espressomd/espresso/blob/dca47de96d175bc382060a8f11cddbe7278ea310/src/tcl/pressure_tcl.cpp#L365)

PR #459, opened 2015-11-16
Ghost thermalization relaxation fix

rempferg:


rempferg:
engine_lbgpu test case fails. That's probably one with reference results and too tight tolerance.

rempferg:
Don't pull. engine_lbgpu test case fails.

uschille:
I think we should change the default behaviour to M10, i.e., MRT with ghost relaxation set to zero. The difference for BGK and TRT may in fact be a valid effect. On the long run, I would favour TRT as the default, but then the test cases need to be adjusted.

PR #465, opened 2015-11-19
Compilation fix for the GAUSSRANDOM* features

psci2195:

PR #468, opened 2015-11-23
Test coverage for different thermostat noise shapes and fixed mass-and-rinertia and virtual-sites-rotation testcases

### LINKED ISSUE #365: Test coverage for different noise shapes in thermostat missing
Via compile time switches, it is possible to switch the noise function between 
FLATNOISE, GAUSSRANDOM, GAUSSRANDOMCUT.
Only the flat noise (default) variant seems to be covered in the Travis testing, however.
Pleasse add default and maxset configurations with the other noise shapes to Travis.

RudolfWeeber:
This adds travis builds for maxset with guassian and with cutoff guassian thermostat noises (closes #365).
That the existing mass-and-rinertia and and virtual-sites-rotation tests passed with flat noise probably was a coincidence. The vs test did not pass for gaussian noise.

As the inertia tensor applies to the axes in the body frame, wehn judging the correctness of the rotational kinetic energy, the angular velocity has to be examined also in the body-fixed frame.
 
I randomized the values in the test case, to prvent further coincidental passes.


RudolfWeeber:
And of corse, it also fixes the build errors for the gaussian noises, i.e., it should replace #465

PR #469, opened 2015-11-25
p3m stress tensor

richter-t:
the p3m algorithm now also computes the stress tensor components


fweik:
Could you please provide a test case for the stress tensor? The existing one (in analysis.tcl) does not contain electrostatics. It would be nice if we could find some test for the off-diagonal elements. Maybe this can be achieved by rotating a system or from some analytical calculation, I don't know.

richter-t:
Hi,

ich hab den Rotations ansatz geschrieben, weiss aber nicht genau wie ich
da einen vollwertigen automatischen Testcase draus machen kann und wie das
genau eingepflegt werden soll/muss.

fweik:
p3m_stress_testcase.tcl fails in some situations. Please invetigate.

PR #473, opened 2015-11-26
Gaussian noises fix in case of LANGEVIN_PER_PARTICLE

psci2195:


RudolfWeeber:
The changes in the core look ok.

The changes to the test case, however, seem to remove the testing of the
Langevin thermostat for the case when LANGEVIN_PER_PARTICLE is
complied in but no particle-speciffic values are set.

I suggest to revert the changes to the test case, because if
only one situation is checked, it should be the most common one

The Langevin_per_particle feature seems to miss test coverage
entirely. I'll open a separate bug report.

psci2195:
I agree. The revert has been done.

PR #500, opened 2015-12-16
Icc python interface + Icc core fixes

KonradBreitsprecher:


fweik:
@KonradBreitsprecher could you please merge with the master. I'm not sure if I understand that correctly. This collides with the change of the verlet criterion by @RudolfWeeber

fweik:
Bump

.travis.yml
```diff
@@ -65,6 +64,3 @@ compiler:
   - clang
 
 script: ./maintainer/travis/build_cmake.sh
-
-after_success:
- - codecov
```
KaiSzuttor: Why is this file changed?

samples/python/p3m.py
```diff
@@ -115,10 +115,22 @@
...
 system.actors.add(p3m)
 
-print("P3M parameter:\n")
+print("\nSCRIPT--->P3M parameter:\n")
+p3m_params = p3m.get_params()
+for key in p3m_params.keys():
+    print("{} = {}".format(key, p3m_params[key]))
+
+print("\nSCRIPT--->Explicit tune call\n")
+p3m._tune()
```
KaiSzuttor: Maybe there should be a method of the P3M class to call the tune routine instead of using the as "internal" marked `_tune` method.

fweik:
Looks ok to me. But agree with @KaiSzuttor that there should be a public tune method. Can you add that? Otherwise I would merge.

KonradBreitsprecher:
Yes @KaiSzuttor has a point there. Will add this next time.

PR #503, opened 2015-12-27
Cmake

rbardak:
Note: the cmake build is still experimental and it isn't tested thoroughly. I will start working on migrating the testcases around mid january.

Also when merging: there are changes in src/core/config.hpp and src/python/espressomd/Makefile.am. As far as I know these two changes work as intended, but someone should glance over it just in case.


CMakeLists.txt
```diff
@@ -0,0 +1,121 @@
...
+#######################################################################
+if(WITH_MPI)
+  find_package(MPI REQUIRED)
+  list(APPEND CMAKE_C_FLAGS ${MPI_COMPILE_FLAGS})
+  list(APPEND CMAKE_EXE_LINKER_FLAGS ${MPI_LINK_FLAGS})
+  include_directories(${MPI_INCLUDE_PATH})
+  list(APPEND LIBRARIES ${MPI_LIBRARIES})
+  set(MPI_SOURCE)
+else(WITH_MPI)
+  include_directories(${CMAKE_SOURCE_DIR}/src/mpifake)
```
mkuron: Don't need `${CMAKE_SOURCE_DIR}` here, `include_directories` takes paths relative to the source directory.

CMakeLists.txt
```diff
@@ -0,0 +1,121 @@
...
+  list(APPEND LIBRARIES ${TCL_LIBRARY})
+endif()
+
+find_package(HDF5)
+if(HDF5_FOUND)
+  set(H5MD 1)
+  list(APPEND LIBRARIES ${HDF5_LIBRARIES})
+  include_directories(${HDF5_INCLUDE_DIRES})
+endif()
+
```
mkuron: I think some newer features require the `MAGMA` and `ARPACK` libraries. These are detected by the autotools build system, but not by your CMake.

cmake/RequireCXX11.cmake
```diff
@@ -0,0 +1,42 @@
...
+"Clang" )
+if ( CMAKE_CXX_COMPILER_VERSION VERSION_LESS
+${required_clang_version} )
+message(FATAL_ERROR "Clang version must be at least "
+"${required_clang_version} for C++11
+support, detected: "
+"${CMAKE_CXX_COMPILER_VERSION}")
+endif ()
+endif ()
+
```
mkuron: You can detect the Intel compiler using `if( ${CMAKE_CXX_COMPILER} MATCHES "icpc" OR "${CMAKE_CXX_COMPILER_ARG1}" MATCHES "icpc" )`. It also needs the `-std=c++11`.

The Cray compiler can be detected using
```
if( ${CMAKE_CXX_COMPILER_ID} MATCHES Cray )
    if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 8.4)
        message( FATAL_ERROR "Insufficient Cray Compiler Environment version" )
    endif()
    ...
endif()
```
and it needs the `add_flag ( CMAKE_CXX_FLAGS "-hstd=c++11" )` flag.

doc/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+add_subdirectory(dg)
+
+find_program(PDFLATEX pdflatex)
+find_program(MAKEINDEX makeindex)
+
+set(abs_srcdir ${CMAKE_CURRENT_SOURCE_DIR})
+set(abs_top_srcdir ${CMAKE_SOURCE_DIR})
+
+configure_file(latexit.sh.in
+               ${CMAKE_BINARY_DIR}/doc/latexit.sh
```
mkuron: Don't need ` ${CMAKE_BINARY_DIR}` here, `configure_file` takes a target path relative to the binary directory.

doc/dg/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+add_custom_target(dg
+        COMMAND ${CMAKE_BINARY_DIR}/doc/latexit.sh
+        ${CMAKE_CURRENT_SOURCE_DIR} dg
+        WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
```
mkuron: Don't need `${CMAKE_CURRENT_BINARY_DIR}` here, `add_custom_target`'s `WORKING_DIRECTORY` is relative to the current binary directory, so either use `.` or leave out the parameter.

doc/ug/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+add_custom_command(OUTPUT quickref.inp
+                   COMMAND ${AWK} -f
+                   ${CMAKE_CURRENT_SOURCE_DIR}/assemble_quickref.awk
+                   ${ug_TEXFILES_abs} > ${CMAKE_CURRENT_BINARY_DIR}/quickref.inp
+)
+
+add_custom_target(ug
+        COMMAND ${CMAKE_BINARY_DIR}/doc/latexit.sh
+        ${CMAKE_CURRENT_SOURCE_DIR} ug
+        WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
```
mkuron: Don't need `${CMAKE_CURRENT_BINARY_DIR}` here, `add_custom_target`'s `WORKING_DIRECTORY` is relative to the current binary directory, so either use `.` or leave out the parameter.

doc/ug/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+        DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/quickref.inp
+                ${CMAKE_BINARY_DIR}/doc/latexit.sh
+                ${CMAKE_SOURCE_DIR}/doc/logo/logo.pdf
+)
+
+#add_latex_document(ug.tex
```
mkuron: Why is this commented out? Delete if not needed.

src/CMakeLists.txt
```diff
@@ -19,6 +17,22 @@
...
 #
-ESPRESSO_SCRIPTS="@ESPRESSO_SCRIPTS@"
-export ESPRESSO_SCRIPTS
-@ESPRESSO_BIN@ $*
+cmake_minimum_required(VERSION 2.6)
+
+include_directories(${CMAKE_CURRENT_SOURCE_DIR})
+include_directories(${CMAKE_CURRENT_BINARY_DIR})
+
+configure_file(${CMAKE_SOURCE_DIR}/cmake/cmake_config.cmakein
```
mkuron: Don't need ` ${CMAKE_SOURCE_DIR}` here, `configure_file` takes a target path relative to the current source directory. So use `..` instead.

src/core/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
+
+set(CUDA_NVCC_FLAGS ${CUDA_NVCC_FLAGS};-O3 -arch=sm_20)#arch=compute_20,code=sm_20)
```
mkuron: Use `--arch=sm_20 --arch=sm_30` and get rid of the comment. That way, code is optimized for both 2.0 and 3.0 GPUs.

Also, don't put `-O3` here. _nvcc_ should use the global optimization level. I think it automatically gets called with `${CMAKE_CXX_FLAGS}`, so you don't need to add it to `CUDA_NVCC_FLAGS`. However, when compiling in debug mode (i.e. `${CMAKE_BUILD_TYPE}` is set to _Debug_), the flag `-G` should be added to `CUDA_NVCC_FLAGS` so that debugging information is generated for GPU code as well.

src/python/espressomd/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+
+add_custom_command(OUTPUT gen_pxiconfig.cpp
+                   COMMAND ${PYTHON_EXECUTABLE}
+                   ${CMAKE_CURRENT_SOURCE_DIR}/gen_pxiconfig.py
+                   ${CMAKE_SOURCE_DIR}/src/features.def ${CMAKE_CURRENT_BINARY_DIR}/gen_pxiconfig.cpp
+                   DEPENDS ${CMAKE_SOURCE_DIR}/src/features.def
+                  )
+
+include_directories(${CMAKE_BINARY_DIR}/src)
+include_directories(${CMAKE_SOURCE_DIR}/src/core)
```
mkuron: Don't need `${CMAKE_SOURCE_DIR}` here, `include_directories` takes paths relative to the current source directory. So you could use `../../core`.

src/tcl/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+file(GLOB_RECURSE Espresso_SRC
+          "*.cpp"
+          )
+
+include_directories(${CMAKE_SOURCE_DIR}/src/core)
+include_directories(${CMAKE_BINARY_DIR}/src/core)
+include_directories(${CMAKE_SOURCE_DIR}/src/tcl)
```
mkuron: Don't need `${CMAKE_SOURCE_DIR}` here, `include_directories` takes paths relative to the current source directory. So you could use `../../core` and `../../tcl`.

src/tcl/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+
+file(GLOB espresso_scripts ${CMAKE_SOURCE_DIR}/scripts/*tcl)
+
+file(COPY ${espresso_scripts} DESTINATION ${CMAKE_BINARY_DIR}/scripts)
```
mkuron: Can use relative path here: `../../scripts`.

PR #504, opened 2016-01-07
Py particle iterator and explicit particle creation

RudolfWeeber:
No range operations yet


RudolfWeeber:
this already has #499 merged

PR #515, opened 2016-01-26
Fix compiling on Mac

mkuron:
- Close unterminated comment
- Workaround for missing symbol ParticleProperties::mass
- Workaround for "static data member not allowed in anonymous struct"
- Fix unresolved symbols in Python interface
- Use correct shared library suffix
- Force building shared libraries for Tcl interface

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -55,6 +55,8 @@ EOF
 dnl overwrite _AC_PROC_CC_G to avoid check for -g and -O2 flags
 m4_define([_AC_PROG_CXX_G],[])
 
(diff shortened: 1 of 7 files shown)


mkuron:
Additionally, one currently needs to run `make install` to get the shared libraries copied to the right place. This is because libtool sets the `install_name` in the shared libraries and I don't know how to change that.

mkuron:
The OpenMPI `RTLD_GLOBAL` workaround wasn't working on Mac OS X. However, I found a much nicer workaround that uses `dlsym` to find out the full path to libmpi. One advantage of this is that we no longer need to carry a list of library names for all different platforms.

PR #518, opened 2016-01-30
Reactions

hmenke:
This pull-request contains the updated `reaction` method.  The old behaviour is preserved and selected by default.  The new method allows direction-dependent swapping of particles.  A python interface has also been made available.

The declarations `using namespace std;` have been purged from the header files `grid.hpp` and `fene.hpp`, as it is considered bad style [1,2].  Also, these declarations were leaking out into all `.cpp` files, which included these, which might ultimately lead to silent overloading of local funtions with a counterpart in `std::`.  To further support my argument, I quote “Thinking in C++” by Bruce Eckel

> However you'll virtually never see a using directive in a header file (at least not outside of scope). The reason is that using directive eliminate the protection of that particular namespace, and the effect last until the end of current compilation unit. If you put a using directive (outside of a scope) in a header file, it means that this loss of "namespace protection" will occur within any file that include this header, which often mean other header files.

[1] https://stackoverflow.com/questions/5849457/using-namespace-in-c-headers
[2] https://stackoverflow.com/questions/4872373/why-is-including-using-namespace-into-a-header-file-a-bad-idea-in-c


fweik:
using/ostringstream stuff is allready dealt with. Merged up to b1fe7bd in #531.

PR #525, opened 2016-02-03
Slice input for particle properties

tobiasrau:
Not all properties are implemented yet


RudolfWeeber:
As far as I can see, your implementation basically needs separate
code for each particle property which is very similar to the
setter/getter in the ParticleHandle class.
I'd find it worthwile to try to figure out a way to re-use the
access methods  in ParticleHandle to avoid code duplicity.

RudolfWeeber:
A rough sketch (untested pseudo-code) might perhaps look like this


class ParticleSlice:

  def __init__(self,s):
    self.slice=s
    # Here, we also need logic for un-limited slices,i.e., [:], [10:], [:10]
    self.n=(s.stop-s.start)/s.step 


  def particle_id_from_index(i):
    return self.slice.start+i*self.slice.step
    
  def __getattr__(self,a):
    # Get type and number of values for the respective property
    first_row=getattr(ParticleHandle(s.start),a)
    columns=len(first_row)
    result=np.zeros((slef.n,columns))
    result[0,:]=first_row
    for i in range(1,n,1):
      result[i,:]=getattr(PArticleHandle(particle_id_from_index(i),a))

  return result

  def __setattr__(self,a,values):
    for i in range(self.n):
        setattr(ParticleHandle(particle_id_from_index(i),a,values[i,:]))

PR #527, opened 2016-02-06
Cmake include actor for python, remove python tests from make check

rbardak:
this fixes building with ewaldgpu.

also changed the check target to not include python tests (check_py does that anyway), to avoid confusion among tests which have the same name


fweik:
Got more elegant solution...

PR #534, opened 2016-02-09
Scafacos support

fweik:


src/core/actor/Mmm1dgpuForce_cuda.cu
```diff
@@ -356,13 +356,13 @@ __global__ void forcesKernel(const __restrict__ mmm1dgpu_real *r, const __restri
 			sum_r *= sqpow(uz);
 			sum_z *= sqpow(uz);
 
-			sum_r += rxy*cbpow(rsqrt(rxy2+pow(z,2)));
-			sum_r += rxy*cbpow(rsqrt(rxy2+pow(z+boxz,2)));
-			sum_r += rxy*cbpow(rsqrt(rxy2+pow(z-boxz,2)));
+			sum_r += rxy*cbpow(rsqrt(rxy2+sqpow(2)));
```
mkuron: This is wrong and should fail the test case. Please change this line to `sum_r += rxy*cbpow(rsqrt(rxy2+sqpow(z)));`

src/core/actor/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+file(GLOB Actor_SRC *.cpp)
+
+add_library(Actor SHARED ${Actor_SRC})
```
mkuron: Are you sure this works? libActor will refer to symbols in libEspresso (e.g. `SystemInterface::gpuBegin`), creating the same kind of dependency loop that I was seeing with libEspressoCuda when built on Mac OS X.

Issue #533, opened 2016-02-09
Change in testsuite is not detected by cmake


fweik:

rbardak:
what kind of change?

fweik:
A change in the test scripts (.tcl)

mkuron:
Just symlink the tcl files instead of copying them.

rbardak:
$ cmake -E 2> >(grep -B1 symlink)
Available on UNIX only:
  create_symlink old new    - create a symbolic link new -> old

I can provide a horribly inelegant solution which only works on UNIX

mkuron:
What's that `test -d` for?

Hmm, I would have thought CMake could do symlinks on Windows too (they're called _reparse points_ there, but it's the same thing). Too bad.

Why do we need to copy the tcl scripts anyway? Can't we just call them with `${CMAKE_CURRENT_SRC_DIR}`? If it's necessary for the script to load blockfiles or other input data, just set the `WORKING_DIRECTORY` property to the source path.

rbardak:
I got some errors without it, apparently globbing also includes directories.

For one thing, a copy would be good because ctest creates testlogs in the working directory. I also just noticed that this method may not even need symlinks, because the testfiles are now part of a target.

fweik:
Another issue is that the tests do not depend on Espresso/pypresso. But I think that can now be easily fixed if there is a custom target.

rbardak:
the add_test commands can also be more generalised by referencing
add_executable(Espresso/pypresso) instead of filepaths. I will look into
that tomorrow.

PR #535, opened 2016-02-09
#533 changes in the testsuite src are now detected

rbardak:


rbardak:
Couldn't figure out how to make the testsuite target depend on targets from other CMakeLists

PR #545, opened 2016-02-23
Random

jonaslandsgesell:
Use the mersenne random number generator (from the c++ standard library) with it's huge period instead of the RAN1 random number generator.


jonaslandsgesell:
nvcc needs to be called with a flag that indicates that we want to use c++11. Compiles if  configure is called like this: ' configure --without-python-interface NVCCFLAGS="-std=c++11" '

src/python/espressomd/Makefile.am
```diff
@@ -58,6 +58,7 @@ PYXFILES =  \
 	magnetostatic_extensions.pyx \
 	particle_data.pyx \
 	reaction.pyx \
+	reaction_ensemble.pyx \
```
fweik: Was this on purpose? If so, the file is missing.

src/python/espressomd/Makefile.am
```diff
@@ -82,6 +83,7 @@ PXDFILES = \
 	p3m_common.pxd \
 	particle_data.pxd \
 	reaction.pxd \
+	reaction_ensemble.pxd \
```
fweik: Same as pyx

src/python/espressomd/Makefile.am
jonaslandsgesell: Good catch. this should be removed here. I did not see that i commited it. After removal the python iterface will likely build.

src/tcl/random_tcl.cpp
```diff
@@ -77,7 +77,7 @@ int tclcommand_t_random (ClientData data, Tcl_Interp *interp, int argc, char **a
   }
   else if ( ARG_IS_S(0,"seed") )    /* 't_random seed [<seed(0)> ... <seed(n_nodes-1)>]' */
   {
-    long *seed = (long *) Utils::malloc(n_nodes*sizeof(long));
+    int *seed = (int *) Utils::malloc(n_nodes*sizeof(int));
```
fweik: Seed is leaking on error. Use a std::vector.

src/core/random.hpp
```diff
@@ -27,183 +27,46 @@
...
-/*@{*/
-#define IA 16807
-#define IM 2147483647
-#define AM (1.0/2147483647.)
-#define IQ 127773
-#define IR 2836
-#define NDIV ((double) (1+(2147483647-1)/NTAB_RANDOM))
-#define RNMX (1.0-1.2e-7)
-# define NTAB_RANDOM  32
+extern std::mt19937 generator;
```
fweik: This construction defies the intent to inline the generation functions. (Other units see just the extern declaration, not the actual code => no inline)

mkuron:
This pull request shows that there are three policy questions that need to be decided ASAP:
- Requiring C++11. Right now that is optional. If we are ok with that and this pull request is merged, it will no longer be. The build system will also need to be modified to throw an error on old compilers and the `HAVE_CXX11` define should be removed.
- Dropping autotools. The two build systems are already starting to diverge, e.g. as Jonas's change requires manually adding to `NVCCFLAGS`. Can we drop autotools rather sooner than later?
- Can we require CUDA 7.0? This pull request requires a C++11-compatible CUDA compiler. If we are ok with that, the CMake CUDA detection will need to check the CUDA version number and ensure it's greater or equal to 7.0.

Personally, I think all these things are reasonable. All the major compilers have supported C++11 for half a decade and even Cray does now. CMake should be on every compute cluster already as e.g. Gromacs needs it too. And while CUDA 7.0 is pretty new, there is no reason why people shouldn't upgrade their machines to it. The only exception would be machines with pre-Fermi GPUs that are only supported up to CUDA 6.5 -- and we dropped support for those old GPUs already when we started requiring compute capability 2.0.

fweik:
I've integrated the stuff with the autotools (see #545), so we can postpone that point.
As for C++11, I think we should do that. We did also already announce that, for example at the developer meeting at the last Espresso school. If somebody would really need a version that builds with C++03, that could still be easily fixed by including boost, but at least from my perspective this is not worth the overhead because it will only affect corner cases.

I will post to the mailing list tomorrow and ask for feedback, (also to announce that we have change the PRNG) 
but agree with Michael on all points.

fweik:
It wrote to the mailing list, and will commit next week when I fixed travis, if there is not a very compelling new argument against it.

mkuron:
Is there any reason why we can't move to the new container-based Travis CI?

fweik:
Yes, apparently the containers are only Ubuntu 12.04.

mkuron:
Can we switch to the [Trusty beta Build Environment](https://docs.travis-ci.com/user/trusty-ci-environment)? That one is not container-based and has GCC 4.8 and Clang 3.5.

fweik:
Yes, that's the plan. I adjust the .travis.yml in my pull request to that end. Now there seems to be some other problem.

Issue #541, opened 2016-02-16
Cmake: In a build wihtout python, py testcases are reported as failed instead of as not run


RudolfWeeber:
94 - rotation_per_particle (Not Run)
         95 - sc_fluid_gpu (Not Run)
         96 - sd_ewald (Not Run)
         97 - sd_two_spheres (Not Run)
         98 - sd_thermalization (Not Run)
        100 - tunable_slip (Not Run)
        102 - virtual-sites (Not Run)
        103 - virtual-sites-rotation (Not Run)
        104 - bondedInteractions (Failed)
        105 - electrostaticInteractions (Failed)
        106 - nonBondedInteractions (Failed)
        107 - cellsystem (Failed)
        108 - ewald_gpu (Failed)
        109 - p3m_gpu (Failed)
        110 - magnetostaticInteractions (Failed)
Errors while running CTest
Makefile:75: recipe for target 'test' failed
make: *** [test] Error 8

Starting from 104, they are py test cases

RudolfWeeber:
I also encountered this in a build with python on the icp machines

Issue #540, opened 2016-02-16
Cmake: Auto-detection of some libs doesn't work


RudolfWeeber:
CUDA:
On a system without cuda:
$ cmake ~/espresso
-- Config file: /home/rudolf/espresso/src/core/myconfig-default.hpp
CMake Error at /usr/share/cmake-3.2/Modules/FindCUDA.cmake:586 (message):
  Specify CUDA_TOOLKIT_ROOT_DIR
  Call Stack (most recent call first):
    CMakeLists.txt:60 (find_package)

    -- Configuring incomplete, errors occurred!
    See also "/home/rudolf/es-build/CMakeFiles/CMakeOutput.log".

TCLTK: Apparently, find_package(TCL) only succeeds if tk is also there
$ cmake -D WITH_CUDA=0 ~/espresso
-- Config file: /home/rudolf/espresso/src/core/myconfig-default.hpp
-- Found PythonInterp: /usr/bin/python (found version "2.7.10")
-- Found Cython: /usr/local/bin/cython
-- Found PythonLibs: /usr/lib/x86_64-linux-gnu/libpython2.7.so (found version "2.7.10")
-- Found NumPy: /usr/lib/python2.7/dist-packages/numpy/core/include (found version "1.8.2")
-- NumPy ver. 1.8.2 found (include: /usr/lib/python2.7/dist-packages/numpy/core/include)
-- Could NOT find FFTW3 (missing:  FFTW3_LIBRARIES FFTW3_INCLUDE_DIR)
-- FFTW3 not found, diabling it.
-- Found Tclsh: /usr/bin/tclsh (found version "8.6")
-- Found TCL: /usr/lib/x86_64-linux-gnu/libtcl8.6.so
CMake Error at /usr/share/cmake-3.2/Modules/FindPackageHandleStandardArgs.cmake:138 (message):
  Could NOT find TCLTK (missing: TK_INCLUDE_PATH)
  Call Stack (most recent call first):
  /usr/share/cmake-3.2/Modules/FindPackageHandleStandardArgs.cmake:374 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake-3.2/Modules/FindTCL.cmake:225 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  CMakeLists.txt:85 (find_package)

  -- Configuring incomplete, errors occurred!
  See also "/home/rudolf/es-build/CMakeFiles/CMakeOutput.log".

[...]

Issue #537, opened 2016-02-15
Cmake: Scafacos is enabledd even if fcs.h is not found


RudolfWeeber:

fweik:
Should be fixed by 5176aefa608054c91fd5e3d2dd83977c558500a6

PR #549, opened 2016-02-25
Cmake update

### LINKED ISSUE #528: Skipped tests in CTest register as Error
Seems to be an upstream issue:
https://cmake.org/Bug/bug_relationship_graph.php?bug_id=15431

fweik:
Fixes #528 #537 #541. And parts of #540.


mkuron:
I think removing TK breaks all of the demos we show for _Kids Week_, _Girls' Day_, _Tag der Wissenschaft_, etc.

rbardak:
is there a reason not to use an older release version for the demos?

fweik:
We need annother plan for that anyway because we want to drop tcl (end hence won't work anymore).

mkuron:
I suppose that plan consists of using the Mayavi visualization and putting a custom GUI around it. But we're not there yet (the Mayavi visualization is not in the master yet because the boundaries are not in the Python interface yet), so I think it's a bit early to drop Tk.

PR #564, opened 2016-03-16
Mayavi

mkuron:
Here's the first fully-working version of the Mayavi live visualization. Please test and let me know what you think.

Simply add
```
from espressomd import visualization
mayavi = visualization.mayavi_live(system)
```
to your setup code and add
```
mayavi.update()
```
to your integration loop. Note that the GUI runs in the main thread, so it will react to button presses only when you call the `update` method the next time. To work around this, you could run your integrate loop in a separate thread, which would probably look something like this:
```
mayavi = visualization.mayavi_live(system)

def integrate_loop():
  for i in range(0,int_n_times):
    integrate.integrate(int_steps)
    mayavi.update()

t = Thread(target=integrate_loop)
t.start()
mayavi.run_gui_event_loop()
```
------------------

You may need to update Mayavi yourself if the one included in your Linux distribution isn't working:
```
pip install --user mayavi
```
If your Linux distribution ships with a VTK version that is too new, you need to get the most recent development version of 4.5.0 (see https://github.com/enthought/mayavi/issues/271).
```
pip install --user https://github.com/enthought/mayavi/archive/master.zip
```


mkuron:
Turns out threading doesn't work yet. It requires perfect separation, no Espresso functions may be called from the GUI thread and no Mayavi functions may be called from the integration thread.
Side note: the GUI always needs to run in the main thread for portability reasons.

The visualization of constrainta is also missing at the moment.

mkuron:
Visualization in the main thread and Espresso in a background thread works nicely now.
Matplotlib running in parallel with Mayavi works too.

This pull request also includes two CMake fixes (use `configure_file` at `make`-time instead of `cmake-command` copy at `cmake`-time) and one `ifdef` fix I made while working on this stuff.

Future work: constraint visualization, visualization of LB, DPD and EK fields. But the current state is already usable for pure particle simulations and thus I think is ready to be checked out by a larger audience and hopefully be merged soon.

KonradBreitsprecher:
I also had troubles installing mayavi with _no module named vtk_ and the suggested solution didn't help. But it seems like my pip was using python3 and vtk is for 2.7..

**pip2.7 install --user mayavi**
is working for me.

RudolfWeeber:
On some systems, like ubuntu, there is a separate pip2 command for
python2.x

PR #576, opened 2016-03-24
Travis

junghans:
`sudo` isn't really needed anymore, so let's switch to the container-based infrastructure.

Additionally I found a bug in `build_cmake.sh`, the `WITH_FFTW` option is gone, so let's replace it with `CMAKE_DISABLE_FIND_PACKAGE_FFTW3`, which forces cmake to not find fftw3 at all.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,42 @@
 language: cpp
 dist: trusty
-sudo: required
+sudo: false
 
 cache: 
   - apt
   - ccache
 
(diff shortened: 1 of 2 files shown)


fweik:
According to the travis documentation the container infrastructure supports only ubuntu 12.04 which has really old compilers that don't support c++11.

fweik:
OK, looks as if their documentation is outdated.

Issue #569, opened 2016-03-16
Python system interface lets user set properties that don't exist


rempferg:
The following should notify the user that system has no property thispropertydoesnotexist. Instead it quietly keeps going:

import espressomd._system as es
import espressomd

system = espressomd.System()
system.thispropertydoesnotexist = 0.001

rbardak:
that's because `espressomd.System` is a python class which inherits from the cython class `System` defined in `_system.pyx`. 

This means espressomd.System has a dynamic `__dict__` for its attribute list. This implies that `system.thispropertydoesnotexist = 0.001` is the same as `system.__dict__["this..."] = 0.001`, which is valid python code.

I don't understand why we have to wrap the cython class System, defined in espressomd/_system.pyx with the python class System, defined in espressomd/system.py in the first place. We don't really get any benefit from that, since we can still import cdef classes as extension types without the need of wrapping them. 
If we were to use the cython class directly, then we would get a runtime error if we tried to set properties which do not exist, therefore solving this issue.

RudolfWeeber:
I think that this was done at some point due to the way the mpi .so
was loaded. It might not be needed any more.

rbardak:
Here are two solutions to the problem.

The first one has no loopholes and is not runtime dependent, but it sets a limitation to the system class. With this fix, we have to require that all attributes of the system class need to be defined at class level. That means if a method sets a new attribute, it needs to be declared first. The limitation only affects the development process for the system class and right now does not have any impact on its functionality.

The second solution relies on the inspect module. It determines in what context `__settattr__` was called. If `__setattr__` is called within a method of the class, then allow the definition of new attributes (eliminating the limitation in the first solution). If `__setattr__` is called in any other context, only allow it if the attribute is already known. This solution has a loophole: if the caller of `__setattr__` has the same name as a method of System, then it also allows full access to the instance. It is also very inelegant and prone to failure, because we look at the stack to determine the context.

RudolfWeeber:
> The first one has no loopholes and is not runtime dependent, but it sets a limitation to the system class. With this fix, we have to require that all attributes of the system class need to be defined at class level. That means if a method sets a new attribute, it needs to be declared first. The limitation only affects the development process for the system class and right now does not have any impact on its functionality.
I'd be fine with that solution

fweik:
I'm not totally sure that this is a problem in the first place. Isn't that
part of what we wanted with python?
I mean that is the python way, if we do not want that we should use a
[...]

PR #578, opened 2016-04-05
Systemset

rbardak:
raise an exception if one tries to set a nonexisting attribute to an instance of System. This implies that methods called from System are also unable to set new attributes if these are not defined at the class level. This is to make sure that the right parameters are set and that the user is notified if a non-existing option was set.

To intentionally set nonexisting attributes, System now has the `create_attr(*args, **kwargs)` method.

### DIFF
--- a/src/python/espressomd/_system.pyx
+++ b/src/python/espressomd/_system.pyx
@@ -39,12 +39,17 @@ setable_properties=["box_l","max_num_cells","min_num_cells",
                     "time_step","timings"]
 
 cdef class System:
+    # NOTE: every attribute has to be declared at the class level.
+    # This means that methods cannot define an attribute by using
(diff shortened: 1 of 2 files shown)


fweik:
Please fix the python indentation

mkuron:
So why exactly do we need this? It seems unpythonic to do. The following example code works in plain Python and that probably is behavior that one would expect in Python:

```
class blah:
    pass

blah.foo = 17 # class attribute
b = blah()
b.bar = 42 # instance attribute
print b.foo, b.bar
```

fweik:
Discussion was in #569.

fweik:
So do we merge that now?

mkuron:
I still don't agree that this is necessary. But assuming it is necessary, this patch is the way to go. As everyone else seems to think it is necessary, please do merge this and close #569.

Issue #277, opened 2015-05-18
unexpected lbfluid + inter behavior


Marcello-Sega:
If an interaction is set up _after_ the LB fluid, the latter is brought to the initial state (see simple script below)

Is this the expected behaviour or should it be fixed? 
```tcl
setmd box_l 2 2 2
setmd periodic 1 1 1
setmd time_step 0.1
setmd skin 0.1
lbfluid cpu agrid 1 dens 1 visc 1.0 tau 0.1
lbnode 0 0 0 set v 0.1 0.0 0.0
integrate 0
# uncomment the line below to see the "unexpected behavior" (momentum= 0 0 0)
# inter 0 0 lennard-jones 1.0 1.0 0.1

puts "[analyze fluid momentum ]"  ; # this should yield 0.1 0 0 
```

fweik:
I'd say this is a bug. The current behaviour is not very intuitive, and there is no reason why the two things should be connected in any way.

rempferg:
Marcello's pull request was merged. I assume that fixed it.

PR #585, opened 2016-04-13
fixed reallog bug in LB

Marcello-Sega:
This should fix the realloc problem with swapping populations in LB


Marcello-Sega:
It's not the first time I'm having issues with (a) switching on/off interactions (b) the order with which interactions / methods are called (see also #277). 

I think it would be good at some stage to make some script that depending on the available features tries them in different orders, at least as much as possible, I guess being comprehensive is impossible.

PR #594, opened 2016-04-26
Replace mdlc test case

RudolfWeeber:
… testing the method

(magnetic dipolar layer correction).
The previous test even passed, when the algorithm was turned off.


KonradBreitsprecher:
langevin_per_particle.tcl testcase fails but no log is shown at
/home/travis/build/espressomd/espresso/testsuite/Testing/Temporary/LastTest.log. 
Commit didn't touch the langevin_per_particle.tcl testcase or related code and gcc run on the test passed. According to langevin_per_particle.tcl, the statistical test might fail in rare cases. Maybe just retry the commit?

fweik:
The langevin per particle issue is fixed in the master.

fweik:
The LastLog issue is also fixed (it should collect [...]/testsuite/tcl/Testing/[...] and so on)

PR #596, opened 2016-04-28
interaction_data: Recalc coulomb prefactor if temperature changes.

fweik:


mkuron:
This may be insufficient. I believe the GPU-based electrostatics algorithms only copy the prefactor to the GPU when their setup routine is called from the interface.

fweik:
Yes, they should react on the on_temperature_change hook. An easy way to fix this would be to let on_temp_chage call on_coulomb_change, but that does a lot of extra work...

mkuron:
Looks like Ewald and MMM1D don't even react to `on_coulomb_change`...

fweik:
U'd postpone that until after we have introduced signals.

PR #597, opened 2016-04-28
Oif periodic

icimrak:
Hi,

Together with my colleague Iveta Jancigova we made some adjustments in object-in-fluid module of ESPResSo. It is mostly only bug fixing, or adding new small features.

We however added new functionality to general LB code, namely to the vtk outputting. Now, it is possible to output also a 2D cut of the lbfluid in the vtk format by the command
lbfluid print vtk velocitycut cutx cuty cutz "outputfile.vtk"

We added a description into the UG as well, see section 12.8. Visualisation, the end of the paragraph.

I also ran the full feature compiled testsuite check, as suggested by Florian at my previous pull request. Checks went ok, however I did not have cude to check GPU. 

I hope the merge will be succesfull,
Thanks,

Ivan Cimrak


rempferg:
GPU tests pass. There are problems with ewald_gpu. These might not be due to this pull request.

icimrak:
Hi,

I don't know much about ewald_gpu. From my understanding, OIF do not interfere with ewald... Is there something I can do regarding the pull request?
Ivan

rempferg:
No, that's fine. I am currently reviewing it. We will pull it once I am done.

rempferg:
I replaced the vtk slice output by allowing the user to define a bounding box in the existing vtk output. You  slice output is a special case of this. Unfortunately, I am having a hard time testing this new feature since the output already fails with the current master on my machine. Need to investigate.

icimrak:
Ok, the bounding box is clever, more general...

I hope the solution will pop up soon... If I can be of any help, just let me know....

rempferg:
It's done, tested, and documented. See https://github.com/espressomd/espresso/pull/632 .
It's our coding day today, so there are a lot of pull requests that need to be merged into the master. We will merge within the hour.

Thank you for your contribution :) Good work!

Issue #561, opened 2016-03-16
MMM1D seg faults when trying to tune


richter-t:
MMM1D seg faults when trying to tune from the tcl interface. This segfault occurs when the periodicity is not set to "0 0 1" and the cellsystem is not set to nsquare.

jonaslandsgesell:
This issue can be closed. It is fixed with the above commit.

PR #599, opened 2016-04-29
added sanity check before tuning in mmm1d starts

jonaslandsgesell:
Fixes Bug #561


jonaslandsgesell:
The failed check has nothing to do with the change in the second commit (57ce007). The commit before (5328545) had no problems. I suppose that the test is inherently broken somehow but independent of this commit.

PR #603, opened 2016-05-06
P3m gpu bugfix

fweik:
- Fixes a bug in p3m-gpu that could lead to crashes and wrong calculations with non-cubic boxes under some circumastances.
- Tuning improvements
- Performance imrpovements for GPU error calculation


src/core/p3m_gpu_cuda.cu
```diff
@@ -426,7 +214,7 @@ __global__ void calculate_influence_function_device (const P3MGpuData p) {
   for(int i = 0; i < 3; ++i)
     Leni[i] = 1.0/p.box[i];
 
-  if((NX >= p.mesh[0]) || (NY >= p.mesh[1]) || (NZ >= p.mesh[2]))
+  if((NX >= p.mesh[0]) || (NY >= p.mesh[1]) || (NZ >= (p.mesh[2] /2 + 1)))
```
richter-t: I don't quite get the condition for NZ, why is it different than the others?

richter-t:
Looks all fine to me. I see mostly renaming and exchanging functions for more efficient ones. 
I hereby declare officially my consent to have those changes merged ;)

src/core/p3m_gpu_cuda.cu
fweik: It's a fft of a real function, so the fourier coefficients are not independent. only the independent ones are calculated, these are Nx * Ny * (Nz / 2 + 1) many. That Nz is the short direction is a choice of cuFFT.

PR #608, opened 2016-05-09
Fix warnings

### LINKED ISSUE #606: Runtime warnings are treated as errors
Creating a runtime warning leads to abort with background_error.

fweik:
Closes #606 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ addons:
       - libboost1.55-dev 
       - libboost-test1.55-dev
       - libboost-mpi1.55-dev
+      - libboost-serialization1.55-dev
       - libopenmpi-dev
       - openmpi-bin
       - libfftw3-dev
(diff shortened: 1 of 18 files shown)


fweik:
Espresso now continous on Runtime Warnings. Warnings are in tcl only returned as the result of the command, so they are only displayed if explicitly printed. Not sure if this is what we want.
This is now build in a way that it can easily be extended to a complete runtime message/logging system,
which probably is cleaner as all the printfs.

fweik:
Ok we can not just return the warnings as result... That may overwrite the result of the command.

fweik:
Ok now the warnings are just writen to stderr or stdout, in tcl only if warnings = 1. I can't think of a better mechanism right now.
Possible improvements: Limit number of errors/warnings, sort by level (related, you don't want warnings to shadow errors), rename to runtime messages, use for logging/debuging?

RudolfWeeber:
I think, this can be merged.
See separate issue (#609) regarding error handling in python

PR #609, opened 2016-05-11
fixed memory leak in cuda and roatation features together with Michael Kuron

jonaslandsgesell:


mkuron:
Now there's a matching _free_ for every _malloc_ in `gpu_change_number_of_part_to_comm()` for all combinations of enabled features..

fweik:
Looks good. Except for the fact that all of those should be thrust::device_vectors which would have  made this bug impossible :-)

PR #629, opened 2016-05-24
Minimize energy steepest descent now available in python

richter-t:


RudolfWeeber:
+++ b/doc/ug/run.tex
@@ -99,6 +99,18 @@ returns the wall runtime in seconds.
   \variant{1} minimize_energy \var{f_\mathrm{max}} \var{steps} \var{gamma} \var{maxdisplacement}
 \end{essyntax}
 
+\begin{pysyntax}
+	\object{
+		espressomd.minimize_energy
+	}{ init 
+	}{ f\textunderscore max = \arg{double},
+	   gamma			    = \arg{double}, 
+	   max\textunderscore steps = \arg{double},
+	   max\textunderscore displacement = \arg{double} 
+   }\\
+   \object{espressomd.minimize_energy}{minimize}{}
+\end{pysyntax}
As the user will typically run it as a member of the system class, please documen that.

 
 
--- /dev/null
+++ b/testsuite/python/minimize_energy.py
@@ -0,0 +1,55 @@
+import sys
+import espressomd._system as es
+import numpy as np
+from espressomd.interactions import *
+from espressomd import analyze
+import espressomd
+
+#from espressomd import minimize_energy
+
+
+system = espressomd.System()

Please use the python unit test framework for testing rather than writing a standalone test.
See other test cases for examples.

PR #630, opened 2016-05-24
Make the check_unit_tests target depend on EspressoConfig in order to…

KaiSzuttor:
… have the config-features.hpp written before.


fweik:
That does not solve the issue, because the executables do not depend on check_unit_test. And the order in which the depends of a target are build is not fixed. See #640 for a better solution.

PR #639, opened 2016-05-24
added diamond command for python interface

jonaslandsgesell:


src/python/espressomd/_system.pyx
```diff
@@ -343,92 +335,69 @@ cdef class System:
...
-                    mpi_random_seed(self.n_nodes,seed_array)
-                else:
-                        raise ValueError("The seed has to be an integer or a list of integers with one integer per node")
-
-            def __get__(self):
-                global __seed
-                return __seed
-        
+        def __set__(self, _seed):
+            print(_seed)
```
KaiSzuttor: Please remove this print statement.

src/python/espressomd/_system.pyx
```diff
@@ -33,10 +33,11 @@ from cellsystem import CellSystem
 
 import sys
 
-setable_properties=["box_l","max_num_cells","min_num_cells",
-                    "node_grid","npt_piston","npt_p_diff",
-                    "periodicity","skin","time",
-                    "time_step","timings"]
+__seed=-10
```
KaiSzuttor: Is this variable assignment needed?

src/python/espressomd/_system.pyx
jonaslandsgesell: yes, will remove debug print

src/python/espressomd/diamond.pyx
```diff
@@ -0,0 +1,70 @@
...
+            raise ValueError(
+                "The distance between two charged monomers' indices must be integer ", self._params[valid_keys[7]])
+        if(self._params[valid_keys[7]] == "nonet"):
+            self._params[valid_keys[7]] = 1
+        if(n_bonded_ia == 0):
+            raise ValueError(
+                "Please define a bonded interaction [0] before setting up polymers!")
+
+    def __set_params_in_es_core(self):
+        print(self._params["a"], self._params["bond_length"], self._params["MPC"], self._params["N_CI"], self._params[
```
jonaslandsgesell: yes

src/python/espressomd/_system.pyx
jonaslandsgesell: yes it is needed, since __seed otherwise is created in __set___() but then the property seed is not recognized as such by python

fweik: There has to be a better solution than this.

fweik:
Now h5md seems broken.

fweik:
Please remove the code completely instead of commenting it out.

jonaslandsgesell:
the h5md Code in the testcase was there to write h5md files from the Tests that rely on blockfiles. This then avoids reading and writing blockfiles from Python and one can Port the Python Tests, wich then only have to read h5md files. I think this is a good strategy to drop blockfiles from the tests.

fweik:
I agree, but then we should actually use hdf5 files and remove the blockfile code. Commented out code is just confusing. Either drop the blockfile code (and blockfiles) or drop the h5md code.

jonaslandsgesell:
I see your point, however reading h5md files is not implemented in tcl and the question is wether it is worth supporting it in the deprecating interface. Therefore my favoured solution is to live with the h5mdcode to write out the testdata in the tcl tests.

fweik:
Let's leave it removed. It is unlikely that somebody would want to write that again.

PR #640, opened 2016-05-24
cmake: Make unit test executables depend on Espresso config, unit_test macro and exclude unit test from all.

fweik:


KaiSzuttor:
Checked compilation with and without tests. Looks good to me. That's a nice solution to solve the dependency problem.
Can be merged.

PR #642, opened 2016-05-25
Jenkins cmake scripts

fweik:


maintainer/travis/build_cmake.sh
```diff
@@ -136,21 +128,21 @@ end "CONFIGURE"
 # BUILD
 start "BUILD"
 
-cmd "make" || exit $?
+cmd "make -j $check_procs" || exit $?
```
fweik: This is not correct. If you do it like that make tries to run $check_procs tests in parallel (while there really is onlye one.) Check procs needs to go as variable into the test. (-DTEST_NP=$check_procs iirc, please check.)

fweik:
@tigoesnumb3rs

PR #643, opened 2016-05-25
Script interface

fweik:
- C++ Interface for a generic script interface
- Factory template
- Cython bindings for this


RudolfWeeber:
This looks good. Just a few minor things:

diff --git a/src/core/shapes/Wall.cpp b/src/core/shapes/Wall.cpp
>new file mode 100644
>index 0000000..331309b
>[...]
>+void Wall::set_parameter(const string &name, const Variant &value) {
>+  /* We need to tranform the vector<double> to a Vector3d for normal */
>+  if(name == "normal") {
>+    /* Get the variant as vector, and explicitly construct a Vector3d
>+       from that. */
>+    n = Vector3d(boost::get<vector<double> >(value).data());   
I think, a SET_VEC3D_PARAMETER_HELPER might be nice as this is a standard
case, but doesn't have to be in this pull req.

>diff --git a/src/core/shapes/Wall.hpp b/src/core/shapes/Wall.hpp
>new file mode 100644
>index 0000000..efb5af3
>+    int calculate_dist(const double *ppos, double *dist, double *vec) const override;  
>+    std::map<std::string, ScriptInterface::Parameter> all_parameters() const override;
>+    std::map<std::string, Variant> get_parameters() const override;
Should we have convenience type defs such as parmeter_map and variant_map?

>diff --git a/src/core/unit_tests/Factory_test.cpp b/src/core/unit_tests/Factory_test.cpp
>new file mode 100644
>index 0000000..ecb9496
>[...]
>+
>+namespace Testing {
>+
>+struct TestClass {
>+  virtual void method() {}
>+};
>+
>+struct DerivedTestClass : public TestClass {};
>+
>+struct OtherDerivedTestClass : public TestClass {};
>+
+} /* namespace Testing */
Please put in a comment at the beginning, to explain
what this will test

>diff --git a/src/core/utils/Factory.hpp b/src/core/utils/Factory.hpp
>new file mode 100644
>index 0000000..b7ece56
>--- /dev/null
>+++ b/src/core/utils/Factory.hpp
>@@ -0,0 +1,100 @@
At the beginning of the file, please put some comments explaining
the different types such as pointer_type, builder_type, ...
Maybe even put an example based on the shapes
[...]
>+  static pointer_type make(const std::string &name) {
Please put a comment explaining what this does
[...]

>+template<class T>
>+std::map<std::string, typename Factory<T>::builder_type> Factory<T>::m_map;
Please put a comment explaining what this is for

>+  template<class T>
>+  auto factory_make(const std::string &name) -> typename Factory<T>::pointer_type {
>+    return Factory<T>::make(name);
>+  }
Same here


Could you also provide a py unit test, e.g. for setting up and retrieving a wall

fweik:
1. I am not sure if I really want to keep the Utils::Vector class. (These are stack vectors as opposed to std::vector which live on the heap). The reason to introduce those was that back then we did not have std::array, which is c++11. Instead the algebra methods could be implemented as free functions, which
would have the added benefit that they would work on all container types when implemented as template overloads. So maybe a better way is to replace the Utils::Vector3d by a std::array, and add those to
the variant type.

2. I'd propose to add 
```c++
typedef std::map<std::string, Variant> VariantMap;
typedef std::map<std::string, Parametert> ParameterMap;
```
 into Parameters.hpp. Our semi-consistent naming scheme is that user-defined types start with a capital letter and are camel-case. I like that because you can immediately see if something is a type or a value.

I will add more comments. pointer_type and so on follow the naming scheme of the stl, you can find those e.g. in std::vector (so much for semi-consistent).

fweik:
For the description of the types in Factory.hpp: They are explained at their declaration,
so that the comment does show up in the doxygen generated code documentation at the
right place. Also this is where you would end up if you use an IDE to jump to the declaration of the type.
I have extended the comment above the class with an usage example.

Regarding the python unit test: At this level we should have a unit test for the python bindings to ScriptInterfaceBase, with a mock ScriptInterfaceBase implementation, otherwise we are testing
a specific implementation on the C++ side and not the bindings. For this to work the test case
has to have a cython part in place (e.g. to include a to be written mock implementation from the c++ unit tests). I'm not sure we have the infrastructure in place to do that at the moment. Any suggestions on how to do that?

The implementation of ScriptInterfaceBase in the various cases should be tested in the c++ unit tests,
and can be automated (get the parameters, set them, get them back...., we could even use the Factories to get a list of classes to be tested). This will be in a separated pull request with the shapes logic. The wall is just in here to illustrate the ScriptInterface.

PR #647, opened 2016-05-27
unit_tests: Replace header only boost test by linked version to impro…

fweik:
…ve build time.


rbardak:
looks good, although I would have used `${Boost_UNIT_TEST_FRAMEWORK_LIBRARY}` instead. And why not link it in the CMakelists.txt of unit_test directly? That's the only place where it seems necessary.

Another point would be that `list(APPEND LIBRARIES ${Boost_LIBRARIES})` now occurs twice in the root level CMakelists.txt which might cause confusion.

fweik:
Agreed.

rbardak:
looks good, go ahead.

PR #657, opened 2016-06-01
Fixed flush for h5md output.

KaiSzuttor:


fweik:
Should not ``h5mdfile H5Fclose`` run flush? That would be the same behavior as for regular files.

KaiSzuttor:
According to evangelos it doesnt. But I'll check it myself tomorrow.

fweik:
I would argue that it should.

KaiSzuttor:
According to the API of hdf5 it should, youre right.

fortysixandtwo:
Oops, looks like I didn't call h5md_close.
Sorry for spreading missinformation :[

KaiSzuttor:
Forgot to check, so thanks Evangelos!

PR #661, opened 2016-06-02
UG: mention a cavet for newcomers when using setmd periodicity

jonaslandsgesell:


arnolda:
Ehm, as far as I can see, the force computations do obey the (non-)periodicity, either because no shift is set (domain_decomposition) or because they use get_mi_vector. So I do agree with the first sentence, but the second one about minimum image I don't understand.

doc/ug/setup.tex
```diff
@@ -127,7 +127,7 @@ \section{\texttt{setmd}: Setting global variables.}
 \item[periodicity] (bool[3]) Specifies periodicity for the three
   directions. If the feature PARTIAL_PERIODIC is set, this variable
   can be set to (1,1,1) or (0,0,0) at the moment.  If not it is
-  readonly and gives the default setting (1,1,1).
+  readonly and gives the default setting (1,1,1). Cavet: Be aware of the fact that setting a dimension to be not periodic does not hinder particles from leaving the box in this direction. In this case for keeping particles in the simulation box a constraint has to be set. Further setting a certain dimension to be not periodic does not affect the minimum image convention for this dimension which means that the box length in this direction still has some effect.
```
fweik: Do you mean 'caveat'?

jonaslandsgesell:
Regarding the second sentence I understood Michael in the following way: Assume you have a nonperiodic Dimension x. Now you have two particles. One at x=0 (lower boundary)  and one at x=box_l -dx.
If the types are such, that there is e.g. a lennard-jones interaction defined, then in this case Espresso will calculate some interaction between the minimum images. This I would not necessarily expect. But maybe I miss understood @Michael?

fweik:
I agree with Axel, the second sentence is not true.

doc/ug/setup.tex
jonaslandsgesell: :) yes of course

doc/ug/setup.tex
```diff
@@ -127,7 +127,7 @@ \section{\texttt{setmd}: Setting global variables.}
 \item[periodicity] (bool[3]) Specifies periodicity for the three
   directions. If the feature PARTIAL_PERIODIC is set, this variable
   can be set to (1,1,1) or (0,0,0) at the moment.  If not it is
```
arnolda: Ehm, just looking at the diff, I noted that line. That is wrong as well. I am pretty sure that MMM2D e.g. can be used only with 1, 1, 0 and MMM1D with 1, 0, 0, which according to this would mean they can't be used at all :-)

fweik:
Have a look at get_mi_vector,(https://github.com/espressomd/espresso/blob/master/src/core/grid.hpp#L188), this looks exactly as expect to me. Minimal image convention is only used for the periodic directions.

doc/ug/setup.tex
jonaslandsgesell: You can use the algorithms without constraints. However particles can then escape your simulation box during the integration which can lead to wrong physics. Which might be unexpected... Therefore I think it is worth mentioning this in the users guide?!

PR #662, opened 2016-06-06
Require CMake 3.0 or higher

mkuron:
CMake 2.8 does not have the `CMAKE_CXX_COMPILER_VERSION` variable, therefore the C++11 compiler flags are not set versions before 3.0.

This problem was observed by @AlexanderWeyman.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,7 +18,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
-cmake_minimum_required(VERSION 2.8)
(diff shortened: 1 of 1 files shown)


fweik:
The variable was introduced in 2.8.8
https://cmake.org/cmake/help/v2.8.8/cmake.html#section_VariablesforLanguages

mkuron:
2.8.12.2 definitely does not work. That's what the bee cluster defaults to. It does not set the `-std=c++11` flag. I guess it's a different but related issue then.

fweik:
Looks that way.

PR #668, opened 2016-06-09
Icc first particle + deterministic testcases

fweik:
Added option to set an particle id offset for dielectric pore.
Replaced tcl rand() in testcases by t_random to make them deterministric.
Sorry for the combined pull request.


KonradBreitsprecher:
rand to t_random looks trivial.

In the dielectrics.tcl, you probably need to set **global start_pid** and communicate it to the iccp3m initialization call via **iccp3m [...] first_id $start_pid**

fweik:
Apparently I broke sth

PR #670, opened 2016-06-13
Bugfix Vkappa (Python Interface)

AlexanderWeyman:
Wrongly commented out code caused error when calling analyze.Vkappa(system, 'reset')


fweik:
Looks good, will merge if the tests pass.

PR #672, opened 2016-06-13
Python interface for printing LB node velocity

KaiSzuttor:


RudolfWeeber:
diff --git a/src/python/espressomd/lb.pyx b/src/python/espressomd/lb.pyx
index 53c3eea..42e2875 100644
--- a/src/python/espressomd/lb.pyx
+++ b/src/python/espressomd/lb.pyx
@@ -161,6 +161,13 @@ IF LB_GPU or LB:
             lb_lbfluid_save_checkpoint(path, binary)
         def load_checkpoint(self, path, binary):
             lb_lbfluid_load_checkpoint(path, binary)
+        def lbnode_print_velocity(self, coord):
+            cdef double[3] double_return
+            cdef int[3] c_coord
+            for i in range(len(coord)):
+                c_coord[i] = int(coord[i])
+            lb_lbnode_get_u(c_coord, double_return)
+            return double_return
The function doesn't print anything. Please call it lbnode_Get_velocity() or similar.
Might also be nice to cast the return value to numpy.ndarray

KaiSzuttor:
You’re right @RudolfWeeber . I should name the method accordingly. Actually there are some other functions in the LB python interface where the name doesn’t fit the action, e.g. [https://github.com/espressomd/espresso/blob/f0999647a68a7de6c5711cd9251709928fe8be31/src/python/espressomd/lb.pyx#L152-L159] .

fweik:
@rempferg ?

KaiSzuttor:
Think there's no reason not to merge.

PR #676, opened 2016-06-14
Add MpiCallbacks to Makefile.am

hirschsn:


fweik:
You should use cmake.

PR #677, opened 2016-06-14
Py cleanup and fixes

RudolfWeeber:
BondedInteractions, NonBondedInteractions, Actors, ParticleHandles, ParticleSlices, ParticleLists are printable via implementation of __str__()

Energy and pressure components are returned in a meaningful order for better printability

Remove picke'ing of the seed in the system class, because that would not actually result in the simulation using the same rng state as at the time of checkpointing. Additionally, the current implementation breaks checkpointing of the system class.

Autogenerate the list of valid particle properties (in place of the incomplete manually generated list that was there)

Remove duplicate code from test cases

Make ewald_gpu and p3m_gpu test cases actually run (rather than skip)


fweik:
Can you please add a comment that explains why some attributes are removed from the particle attributes before pickeling?

RudolfWeeber:
Done

fweik:
Thx

PR #678, opened 2016-06-14
moved logic of calculating mindist to all constraints to the C_core. …

jonaslandsgesell:
…Useful for example when getting this mindist in the new python interface or when needing to consider this mindist in a monte carlo move


fweik:
In principle you have of course a valid point. But I am currently working on completely refactoring the constraints, so I won't merge this at this time.

PR #679, opened 2016-06-14
Update mmm1dgpu.tcl

jonaslandsgesell:
this should work without cell system nsquare


mkuron:
That is correct. `cellsystem nsquare` is only required on the CPU.

PR #680, opened 2016-06-16
mpiio: Removed dependency on the tcl interface.

fweik:


fweik:
@hirschsn

fweik:
Now :-)

hirschsn:
Works fine, can be merged. Thanks for fixing this.

PR #693, opened 2016-06-20
Py: Move analysis routines into a class

RudolfWeeber:
To avoid passing system to every single call of an analysis method.
Also make analysis available via ana property of System class.
This includes the interface itself, testcases, sample scripts and user guide.

### DIFF
--- a/doc/ug/analysis.tex
+++ b/doc/ug/analysis.tex
@@ -97,7 +97,7 @@ radius \var{r\_catch} around the position of the particle with number
 
 \begin{pysyntax}
   \object{
-    espressomd.analyze
+    espressomd.analyze.Analysis(system)
   }{
     distribution
   }{
@@ -186,7 +186,7 @@ of the currently identified properties.
 
 \begin{pysyntax}
   \object{
-    espressomd.analyze
+    espressomd.analyze.Analysis(system)
   }{
     cylindrical_average
   }{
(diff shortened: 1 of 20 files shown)


KaiSzuttor:
What's the benefit from that? Writing system.ana.XXX() vs. analyze.XXX(system) is the same effort and "ana" may not be that obviously meaning "analyze". Can you explain what's the point here @RudolfWeeber ?

fweik:
I'm in favor of writing it out.

RudolfWeeber:
> What's the benefit from that? Writing system.ana.XXX() vs. analyze.XXX(system) is the same effort and "ana" may not be that obviously meaning "analyze". Can you explain what's the point here @RudolfWeeber ?
Most analysis routines don't need a handle to the system class, but
some do. 
Some analysis routines want system as a keyword arg, others as a
non-keyword-arg. 
Removing the system argument from those that don't need it but
keeping it for the rest would decreaswe usability, since it is hard to remember which ones do and
which don't need it. 
The class Analysis stores the reference at construction and it
can be used as needed.

Concerning typing:
system.ana is a convenience thing like system.part and
system.non_bonded_inter. The analysis class isn't a singleton.
You can do 
a=Analysis(system)
print A.energy()

As for the attribute name system.ana:
ana is a compromise, just as part and non_bonded_inter. Sicne the
use of the routines is frequent, I gave shortness priority.

RudolfWeeber:
> I'm in favor of writing it out.
The "ana" attribute of the system class? 
I find ana clear enough, but I also thing, changing it wouldn't hurt much, 
because people can store a reference under a shosrt name anyway.
If we change to the long name, we should probably mention that
possibility in the doc.

fweik:
People can also use auto-completion :-)

PR #716, opened 2016-06-22
Python Tutorial Charged System

AlexanderWeyman:
rewrite tcl tutorial for the python interface. please keep in mind that the tutorials is incomplete due to missing python implementations. Latex lstlisting should be adjusted in common.tex for Python language.


rbardak:
the cmake stuff looks good

fweik:
Can you please merge the master into this branch. This should fix the python errors, just that we can see whether there are any other issues.

PR #721, opened 2016-06-26
Euler equations nonlinear term for the anisotropic case

psci2195:
Dear Espresso team,

This is just a proposal of a correction of the case of the anisotropic particle rotational motion. If we misunderstood something - please correct us. However, one looks like a very simple bug. At the beginning of the file "rotation.cpp", one can see "static double I[3] = { 1, 1, 1};". This variable is not changing within the whole source code. As a result, second term of Euler equations (nonlinear one) will be always zero:

Wd[0] =  (p->f.torque[0] + p->m.omega[1]*p->m.omega[2]*(I[1]-I[2]))/I[0]/p->p.rinertia[0];

cause I[1]-I[2] = 0. Which should not be the case in the eq. (3) of the reference http://ciks.cbt.nist.gov/~garbocz/dpd1/node2.html

===

Concerning proposed changes in tests.. This is can be false-negative vs false-positive discussion. Indeed, in case of an anisotropic body and after a nonzero values definition (the changes in the "rotation.cpp") of the nonlinear parts of Euler's rotational equations, it seems that exponential expected behaviour is a not a fully precise assumption. At least, nonlinear differential equations do not give exponent-like behaviour (usually). Such anisotropic decelleration should be validated using other equations or expected numerical values, which also, probably, could be obtained only using a simulation. This is why we've made a [set J "10 10 10"].

If someone could suggest a closed-form expression for this system of 3 nonlinear differential equations (decelleration case) then we will correct tests accordingly.

===

Please note, that other tests related to anisotropy are passing for the introduced fixes in the rotation.cpp without tests change, e.g. fragments like

set j1 [expr [t_random] * 20]
set j2 [expr [t_random] * 20]
set j3 [expr [t_random] * 20]

in different tests: virtual-sites-rotation.tcl, mass-and-rinertia.tcl, mass-and-rinertia_per_particle.tcl, etc.

Regards, Bogdan T.


rbardak:
When I was working on code to integrate the rotational EOM of anisotropic rigid bodies I remember doing the same changes to the rotation code. Back then I also double checked with the mentioned references in the code and came to the conclusion that these changes are consistent with the theory. 

I agree with your proposal, however I would introduce a slight change: in the case of `#ifdef ROTATIONAL_INERTIA`, where you introduced your changes, I would completely remove the `I` variable.

psci2195:
Indeed, the rotation according to Euler eq. (3) & quaternion-based scheme of Omelyan is performing with the rotating body-fixed reference frame aligned with principal axes of inertia. Actually, the method rotation.cpp is following this algorithm precisely. Hence, a diagonalizing of the I[i,j] is not required: this are always principal moments of inertia .

rbardak:
I agree that rotations.cpp follows the method precisely and diagonalization is not required, but that is not the point that I was trying to make in the last bit of my comment.
When I said related code I meant code in different sections of espresso which may or may not rely on the rotations code.

psci2195:
Concerning other Espresso code - I saw usage of p->p.rinertia[j], and, probably, no usage of I[j].

RudolfWeeber:
I did not look at the theory (in detail), but even without that, I agree that the current implementation makes no sense. 
My impression is that this pull request can be merged.
@jdegraaf: Any comments?

Marcello-Sega:
I remember we discussed this long ago, and I have a foggy memory that the solution similar or equal to the one proposed here had several problems. I would wait a bit before merging: I can check older messages, think a bit about it and report to you asap if I recall something more specific.

RudolfWeeber:
On Wed, Jun 29, 2016 at 08:09:58AM -0700, Marcello Sega wrote:
> I remember we discussed this long ago, and I have a foggy memory that the solution similar or equal to the one proposed here had several problems. I would wait a bit before merging: I can check older messages, think a bit about it and report to you asap if I recall something more specific. 
The suggested code is:
Wd[0] =  (p->f.torque[0] + p->m.omega[1]*p->m.omega[2]*(p->p.rinertia[1]-p->p.rinertia[2]))/p->p.rinertia[0];
Wd[1] =  (p->f.torque[1] + p->m.omega[2]*p->m.omega[0]*(p->p.rinertia[2]-p->p.rinertia[0]))/p->p.rinertia[1];
Wd[2] =  (p->f.torque[2] + p->m.omega[0]*p->m.omega[1]*(p->p.rinertia[0]-p->p.rinertia[1]))/p->p.rinertia[2];
Wikipedia (Euler equation rigid body dynamics) says
I_1\dot{\omega}_{1}+(I_3-I_2)\omega_2\omega_3 &= M_{1}\\
I_2\dot{\omega}_{2}+(I_1-I_3)\omega_3\omega_1 &= M_{2}\\
I_3\dot{\omega}_{3}+(I_2-I_1)\omega_1\omega_2 &= M_{3}

Reshuffling yields
I_1\dot{\omega}_{1} +(I_3-I_2)\omega_2\omega_3 + M_{1}
I_1\dot{\omega}_{1} -(I_2-I_3)\omega_2\omega_3 + M_{1}
\dot{\omega}_{1} =[(I_2-I_3)\omega_2\omega_3 + M_{1}]/I_1

Don't you think that the code above is correct?

The existing code in Espresso is
Wd[0] =  (p->f.torque[0] +   p->m.omega[1]*p->m.omega[2]*(I[1]-I[2]))/I[0]/p->p.rinertia[0];
which ignores two of the inertia moments of the particle under
consideratoin, and uses a sphere instead. It also contains an
additional 1/I[0] for the non-linear term, which only doesn't hurt,
because the one of a sphere of unit inertia moment is used.

Marcello-Sega:
Hi,

the equation are right, sure, I remember there were some more general
problems related to this, but I can't recall them exactly anymore, maybe
Axel can tell us something more.

There is one point about the fluctuation dissipation theorem, though, which
certainly applies here. If the inertia tensor is not isotropic, the torque
fluctuations are not simply delta correlated, but must have a definite
covariance, see for example https://goo.gl/xNqJ1U

If I'm not mistaken, this is not the case for espresso so far, so the patch
would work correctly only for the NVE case. This is probably of minor
relevance for the Espresso user base.

Regards,

M.




On Thu, Jun 30, 2016 at 12:37 PM, RudolfWeeber <notifications@github.com>
wrote:

> On Wed, Jun 29, 2016 at 08:09:58AM -0700, Marcello Sega wrote:
> > I remember we discussed this long ago, and I have a foggy memory that
> the solution similar or equal to the one proposed here had several
> problems. I would wait a bit before merging: I can check older messages,
> think a bit about it and report to you asap if I recall something more
> specific.
> The suggested code is:
> Wd[0] = (p->f.torque[0] +
> p->m.omega[1]*p->m.omega[2]*(p->p.rinertia[1]-p->p.rinertia[2]))/p->p.rinertia[0];
> Wd[1] = (p->f.torque[1] +
> p->m.omega[2]*p->m.omega[0]*(p->p.rinertia[2]-p->p.rinertia[0]))/p->p.rinertia[1];
> Wd[2] = (p->f.torque[2] +
> p->m.omega[0]*p->m.omega[1]*(p->p.rinertia[0]-p->p.rinertia[1]))/p->p.rinertia[2];
> Wikipedia (Euler equation rigid body dynamics) says
> I_1\dot{\omega}_{1}+(I_3-I_2)\omega_2\omega_3 &= M_{1}\\
> I_2\dot{\omega}_{2}+(I_1-I_3)\omega_3\omega_1 &= M_{2}\\
> I_3\dot{\omega}_{3}+(I_2-I_1)\omega_1\omega_2 &= M_{3}
>
> Reshuffling yields
> I_1\dot{\omega}_{1} +(I_3-I_2)\omega_2\omega_3 + M_{1}
> I_1\dot{\omega}_{1} -(I_2-I_3)\omega_2\omega_3 + M_{1}
> \dot{\omega}_{1} =[(I_2-I_3)\omega_2\omega_3 + M_{1}]/I_1
>
> Don't you think that the code above is correct?
>
> The existing code in Espresso is
> Wd[0] = (p->f.torque[0] +
> p->m.omega[1]*p->m.omega[2]*(I[1]-I[2]))/I[0]/p->p.rinertia[0];
> which ignores two of the inertia moments of the particle under
> consideratoin, and uses a sphere instead. It also contains an
> additional 1/I[0] for the non-linear term, which only doesn't hurt,
> because the one of a sphere of unit inertia moment is used.
>
>
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub
> <https://github.com/espressomd/espresso/pull/721#issuecomment-229623091>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAccj-koqPVXFkvO0IYTyLf-pOxMcjkdks5qQ5xVgaJpZM4I-me9>
> .
>



-- 
University of Vienna

RudolfWeeber:
> There is one point about the fluctuation dissipation theorem, though, which
> certainly applies here. If the inertia tensor is not isotropic, the torque
> fluctuations are not simply delta correlated, but must have a definite
> covariance, see for example https://goo.gl/xNqJ1U
The rotational Langevin thermostat in its current form is anyway not suitable for ellipsoid, because it does not have axis-dependent relaxation constants.

> If I'm not mistaken, this is not the case for espresso so far, so the patch
> would work correctly only for the NVE case. This is probably of minor
> relevance for the Espresso user base.
For ensemble averages, the proposed corretion will indeed make no difference. But it might be quite relevant for driven ellipsoids in an lb fluid, as the lb fluid should thermalize an ellipsoid based on the raspberry model correctly.

fweik:
Ok, looks like we want to merge this?

Marcello-Sega:
My personal opinion is that it is dangerous to have code that violates the
fluctuation dissipation theorem, so I would rather either implement the
proper thermostat, or avoid using anisotropic inertia tensors, as it was
till now (afaiu).

Regarding the comment about the raspberry model, I'm not sure I understood
what Rudolf meant :(

Cheers,

M


On Thu, Jun 30, 2016 at 5:31 PM, Florian Weik <notifications@github.com>
wrote:

> Ok, looks like we want to merge this?
>
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub
> <https://github.com/espressomd/espresso/pull/721#issuecomment-229695996>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAccj-8Ap0odIANhLno7dPz0vwExIxDXks5qQ-FUgaJpZM4I-me9>
> .
>



-- 
University of Vienna

fweik:
@Marcello-Sega Are you sure about the noise covariance? I think this is only true if you have deviation moments, anisotropy alone is not sufficient (c.f. Eq. (3) + (4) of the paper you cited, iff I is diagonal, the covariance matrix is also diagonal if I am not mistaken. Don't trust me on that, it is to hot here for calculus.)

Marcello-Sega:
Yes, it's diagonal in that case, but for an ellipse the 3 diagonal elements
are not the same, so this should be taken care of in the Langevin
thermostat.

M

On Thu, Jun 30, 2016 at 7:53 PM, Florian Weik <notifications@github.com>
wrote:

> @Marcello-Sega <https://github.com/Marcello-Sega> Are you sure about the
> noise covariance? I think this is only true if you have deviation moments,
> anisotropy alone is not sufficient (c.f. Eq. (3) + (4) of the paper you
> cited, iff I is diagonal, the covariance matrix is also diagonal if I am
> not mistaken. Don't trust me on that, it is to hot here for calculus.)
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/espressomd/espresso/pull/721#issuecomment-229737365>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAccj8seWoiR5CvIuE55-OXhDleJAIEeks5qRAKkgaJpZM4I-me9>
> .
>



-- 
University of Vienna

RudolfWeeber:
On Thu, Jun 30, 2016 at 11:33:47AM -0700, Marcello Sega wrote:
> Yes, it's diagonal in that case, but for an ellipse the 3 diagonal elements
> are not the same, so this should be taken care of in the Langevin
> thermostat.
I didn't get that. Is there a need for the correlation of noises for
different axes even if the inertia tensor is diagonal?
Because if not, the inertia moment probably does not enter, just
like the mass does not enter in the Langevin thermostat.

Regarding raspberry ellipsoids and lb: In that case, there is no
Langevin thermostat. The thermalization is done via the lb coupling
of all the points making up the raspberry. The inertia tensor
doesn't enter except in the equation of motion of the raspberry's
center of mass. And that equation of motion is wrong for an
ellipsoid, if the pull request is not applied.

If there is a change needed in the Langevin thermostat, imo, we should
implement that rather than keeping the wrong equation of motion.

Marcello-Sega:
On Thu, Jun 30, 2016 at 9:19 PM, RudolfWeeber <notifications@github.com>
wrote:

>
>
> If there is a change needed in the Langevin thermostat, imo, we should
> implement that rather than keeping the wrong equation of motion.
>
>
>
I totally agree on that, let's try to fix the complete thing.


M



> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/espressomd/espresso/pull/721#issuecomment-229761120>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAccj6Ogf0sQg1QV68RwwoiVpoBBQuduks5qRBa1gaJpZM4I-me9>
> .
>



-- 
University of Vienna

fweik:
I think you get a noise amplitude proportional to the sqrt of the corresponding diagonal entry of the inertia tensor (analog to the sqrt of the mass in the linear case), and no cross-correlation. 

But sure, let's fix it for the general case.

psci2195:
Hi @Marcello-Sega, could you please advise which tests could be written to understand problems of the existing rotational part of Langevin thermostat with proposed changes to a rotational motion? For instance, we could catch (by new tests) the violation of fluctuation dissipation theorem mentioned above. Maybe I'm mistaken, I thought that tests ```mass-and-rinertia.tcl``` and ```mass-and-rinertia_per_particle.tcl``` already validate this. I will read in more details reference which you cited and dive deeply into this. But maybe you already could advise which problems could be detected by tests in a test-driven way.

Marcello-Sega:
Hi,

I've checked thermostat.hpp, and it looks like the correct thermostat for
diagonal anisotropic tensors is there already, so what is currently missing
is a friction coefficient that depends on the axis of rotation. The tests
you mentioned should be fine, just one should make sure that they are taken
in the large friction limit.

Note that by including the nonlinear term \omega x I \omega, if your're not
in the overdamped limit, the tests can (wrongly!) fail, because < I
\omega^2> != kT (see the discussion about eq. 6&9 in the paper cited
before)

Cheers,

M





On Fri, Jul 1, 2016 at 3:18 PM, Dr. Bogdan Tanygin <notifications@github.com
> wrote:

> Hi @Marcello-Sega <https://github.com/Marcello-Sega>, could you please
> advise which tests could be written to understand problems of the existing
> rotational part of Langevin thermostat with proposed changes to a
> rotational motion? For instance, we could catch (by new tests) the
> violation of fluctuation dissipation theorem mentioned above. Maybe I'm
> mistaken, I thought that tests mass-and-rinertia.tcl and
> mass-and-rinertia_per_particle.tcl already validate this. I will read in
> more details reference which you cited and dive deeply into this. But maybe
> you already could advise which problems could be detected by tests in a
> test-driven way.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/espressomd/espresso/pull/721#issuecomment-229944371>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAccj3QzncGx52b5h4wQb5PmUoD3LVlUks5qRRPBgaJpZM4I-me9>
> .
>



-- 
University of Vienna

Marcello-Sega:
Another point: the kinetic energy contribution must be changed, right now
it also assumes an isotropic tensor (energy_inline.hpp)

  /* the rotational part is added to the total kinetic energy;

     at the moment, we assume unit inertia tensor I=(1,1,1)  */

  energy.data.e[0] += (SQR(p1->m.omega[0]) + SQR(p1->m.omega[1]) +
SQR(p1->m.omega[2]))*time_step*time_step;

Regards,


M




On Fri, Jul 1, 2016 at 5:31 PM, Marcello Sega <marcello.sega@gmail.com>
wrote:

> Hi,
>
> I've checked thermostat.hpp, and it looks like the correct thermostat for
> diagonal anisotropic tensors is there already, so what is currently missing
> is a friction coefficient that depends on the axis of rotation. The tests
> you mentioned should be fine, just one should make sure that they are taken
> in the large friction limit.
>
> Note that by including the nonlinear term \omega x I \omega, if your're
> not in the overdamped limit, the tests can (wrongly!) fail, because < I
> \omega^2> != kT (see the discussion about eq. 6&9 in the paper cited
> before)
>
> Cheers,
>
> M
>
>
>
>
>
> On Fri, Jul 1, 2016 at 3:18 PM, Dr. Bogdan Tanygin <
> notifications@github.com> wrote:
>
>> Hi @Marcello-Sega <https://github.com/Marcello-Sega>, could you please
>> advise which tests could be written to understand problems of the existing
>> rotational part of Langevin thermostat with proposed changes to a
>> rotational motion? For instance, we could catch (by new tests) the
>> violation of fluctuation dissipation theorem mentioned above. Maybe I'm
>> mistaken, I thought that tests mass-and-rinertia.tcl and
>> mass-and-rinertia_per_particle.tcl already validate this. I will read in
>> more details reference which you cited and dive deeply into this. But maybe
>> you already could advise which problems could be detected by tests in a
>> test-driven way.
>>
>> —
>> You are receiving this because you were mentioned.
>> Reply to this email directly, view it on GitHub
>> <https://github.com/espressomd/espresso/pull/721#issuecomment-229944371>,
>> or mute the thread
>> <https://github.com/notifications/unsubscribe/AAccj3QzncGx52b5h4wQb5PmUoD3LVlUks5qRRPBgaJpZM4I-me9>
>> .
>>
>
>
>
> --
> University of Vienna
>



-- 
University of Vienna

jdegraaf:
Dear All,

I believe that I implemented the right anisotropic rotational tensor as
well (at least the diagonal part), off-diagonal terms would require quite a
bit more work to implement. Marchello is probably right about the kinetic
energy, which I did not modify.

Best Wishes,

Joost

fweik:
@Marcello-Sega if you look at the lines directly above the part you quoted from energy_inline.hpp, (starting from line 523) doesn't that look correct to you?

On a more general note, I think we should start by removing all the specialized code for ROTATIONAL_INERTIA and make the rotational inertia of the particles a static constexpr as with
the mass. The compiler can than hopefully figure out which parts vanish at compile time. If this
leads to a slight performance decrease (which I don't believe) this a fair price to pay for the simplification. 
I'm going to start that right now, just that you are informed that there will be changes to the code in this area.

Marcello-Sega:
Whoops, yep, that's correct, I was fooled by the comment :P

Thanks !

M

On Sun, Jul 3, 2016 at 6:55 PM, Florian Weik <notifications@github.com>
wrote:

> @Marcello-Sega <https://github.com/Marcello-Sega> if you look at the
> lines directly above the part you quoted from energy_inline.hpp, (starting
> from line 523) doesn't that look correct to you?
>
> On a more general note, I think we should start by removing all the
> specialized code for ROTATIONAL_INERTIA and make the rotational inertia of
> the particles a static constexpr as with
> the mass. The compiler can than hopefully figure out which parts vanish at
> compile time. If this
> leads to a slight performance decrease (which I don't believe) this a fair
> price to pay for the simplification.
> I'm going to start that right now, just that you are informed that there
> will be changes to the code in this area.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/espressomd/espresso/pull/721#issuecomment-230163111>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAccj5vvY74HuO1SKjYAMXO5lkPO1iztks5qR-logaJpZM4I-me9>
> .
>



-- 
University of Vienna

rbardak:
> (starting from line 523) doesn't that look correct to you?

`p->m.omega` is in the basis of real space not body space. No it is not correct, as the inertia tensor is only a diagonal matrix in body space. Omega needs to be transformed first. There is an appropriate function in rotations.cpp which does that.

Also we don't need to consider off-diagonal elements at all. The inertia tensor is symmetric. Every real, symmetric matrix is diagonalizable. 

Since the rotation group is SO(3), which contains all basis transformations except for space inversions, the diagonalization can be accomplished by rotating the particles rotational degrees of freedom into an appropriate reference frame (body space) and do the calculations in that coordinate system.

The current implementation tracks the orientation of the particles, giving us the coordinate transformation that we need. All that needs to be done is to make sure that at every step of the way the calculations are done in the proper coordinate system. 
This is already the case when the rotational degrees of freedom are integrated within the functions of rotations.cpp. However the coordinate transformation needs to be considered in other parts of espresso where `p->rinertia` is used.

In general, if you're working with anisotropic particles, you also need to pass the orientation (either as quaternions or euler angles) to espresso, for which the inertia tensor is diagonal, this is already implemented IIRC.

jdegraaf:
I really don't know why there is so much discussion about this. It was
programmed in body space --- what is in and not in body space in the code
actually requires some tracking in the code, but everything having to do
with the updates for the rotational diffusion in the core is in that frame
--- and tested to yield the correct diffusivities for something with all
different diagonal elements...

Please go through the code very, very carefully before you start modifying
stuff which may not be broken to begin with...

Best, Joost

Marcello-Sega:
Hi,

@Robin if omega is in the lab-frame, then also the omega x L term

 Wd[0] =  (p->f.torque[0] + p->m.omega[1]*p->m.omega[2]*(I[1]-I[2]))/I[0
]/p->p.rinertia[0];

should be wrong (assume I[] to be the proper inertia tensor element). Or,
omega is just in the body-frame (I assumed so) and everything is fine.


M



On Sun, Jul 3, 2016 at 7:28 PM, Robin Bardakcioglu <notifications@github.com
> wrote:

> (starting from line 523) doesn't that look correct to you?
>
> p->m.omega is in the basis of real space not body space. No it is not
> correct, as the inertia tensor is only a diagonal matrix in body space.
> Omega needs to be transformed first. There is an appropriate function in
> rotations.cpp which does that.
>
> Also we don't need to consider off-diagonal elements at all. The inertia
> tensor is symmetric. Every real, symmetric matrix is diagonalizable.
>
> Since the rotation group is SO(3), which contains all basis
> transformations except for space inversions, the diagonalization can be
> accomplished by rotating the particles rotational degrees of freedom into
> an appropriate reference frame (body space) and do the calculations in that
> coordinate system.
>
> The current implementation tracks the orientation of the particles, giving
> us the coordinate transformation that we need. All that needs to be done is
> to make sure that at every step of the way the calculations are done in the
> proper coordinate system.
> This is already the case when the rotational degrees of freedom are
> integrated within the functions of rotations.cpp. However the coordinate
> transformation needs to be considered in other parts of espresso where
> p->rinertia is used.
>
> In general, if you're working with anisotropic particles, you also need to
> pass the orientation (either as quaternions or euler angles) to espresso,
> for which the inertia tensor is diagonal, this is already implemented IIRC.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/espressomd/espresso/pull/721#issuecomment-230164823>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAccj8Luwp169qfnl9mQSUVLTqey_o7Eks5qR_E5gaJpZM4I-me9>
> .
>



-- 
University of Vienna

fweik:
For the energy it does not matter btw. because lab- and body-frame are related only by a rotation.

Marcello-Sega:
Mhh, not if an anisotropic inertia tensor enters the game.

M.


On Sun, Jul 3, 2016 at 7:42 PM, Florian Weik <notifications@github.com>
wrote:

> For the energy it does not matter btw. because lab- and body-frame are
> related only by a rotation.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/espressomd/espresso/pull/721#issuecomment-230165502>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAccj_t9TV9bvVSbUDigMqkmN09hG5Rzks5qR_R-gaJpZM4I-me9>
> .
>



-- 
University of Vienna

rbardak:
sorry about that, yes omega is in body space, this is also noted in particle_data.hpp 


> For the energy it does not matter btw. because lab- and body-frame are related only by a rotation.
It does because tensors by definition transform under coordinate transformation. If the tensor transformation is not taken into account you're gonna have a bad time

RudolfWeeber:
> `p->m.omega` is in the basis of real space not body space. No it is not correct, as the inertia tensor is only a diagonal matrix in body space. Omega needs to be transformed first. There is an appropriate function in rotations.cpp which does that.
p->m.omega is in body frame at all times. It only gets converted to
lab space at the time of printing.

The torque, on the other hand is in lab frame part of the time.

> 
> Also we don't need to consider off-diagonal elements at all. The inertia tensor is symmetric. Every real, symmetric matrix is diagonalizable. 
And the omega is written in the basis spanned by the main inertia
axes in body frame.

>

psci2195:
Hi @Marcello-Sega,

Concerning
> I've checked thermostat.hpp, and it looks like the correct thermostat for diagonal anisotropic tensors is there already, so what is currently missing is a friction coefficient that depends on the axis of rotation. The tests you mentioned should be fine, just one should make sure that they are taken in the large friction limit.

I've prepared and committed these changes and new tests. Also, the mentioned reference work was analysed: the tests (existing/new ones) are consistent with it.

Maybe, we can merge this or something additional should be done?

Bogdan

fweik:
Looks good to me.

RudolfWeeber:
> 
> Concerning
> > I've checked thermostat.hpp, and it looks like the correct thermostat for diagonal anisotropic tensors is there already, so what is currently missing is a friction coefficient that depends on the axis of rotation. The tests you mentioned should be fine, just one should make sure that they are taken in the large friction limit.
> 
> I've prepared and committed these changes and new tests. Also, the mentioned reference work was analysed: the tests (existing/new ones) are consistent with it.
> 
> Maybe, we can merge this or something additional should be done?
It looks good to me.

We still need the change in the python interface:
src/python/espressomd/particle_data.pyx and .pxd analogous to
property rinertia
in line 361. 
And a getter in src/core/particle_data.?pp analogous to
pointer_to_rotational_inertia()
If you aren't comfortable with python, I can do the
change. In that case, please open an issue, once this pr is merged.

psci2195:
Hi @RudolfWeeber ,
I'm OK with Python, will supplement the pull request.
Concerning
> And a getter in src/core/particle_data.cpp analogous to pointer_to_rotational_inertia()
it is already in place - line 2256.

RudolfWeeber:
> Hi @RudolfWeeber ,
> I'm OK with Python, will supplement the pull request.
I forgot the testcase in testsuite/python/particle.py

psci2195:
Thanks, OK, will do

psci2195:
Hi @RudolfWeeber 
I've done the implementation according to your advice. Python build & tests were passed successfully with different feature combinations.

fweik:
So, we're good?

psci2195:
Done from my side..

RudolfWeeber:
The py stuff looks good. I think, this can be merged

PR #727, opened 2016-07-06
Python tests now working with Travis

tigoesnumb3rs:
Removed `anaconda`. Now there are no more problems with mysterious python versions.
Changed the build environment to [junest](https://github.com/fsquillace/junest). 

Since `vtk` in the junest-repositories does not have python bindings (like the `python-vtk` package in ubuntu), the testfiles have been modified to skip the tests instead of raising an error.


fweik:
Great. Could you please comment in the other configurations and the unit and tcl tests?

fweik:
@tigoesnumb3rs  bump

fweik:
If I understand this correctly this runs arch in a qemu vm? While that is creative, it is also insanely slow. The best solution for now is probably to run docker containers on travis.

RudolfWeeber:
It's more like a chroot. The test times are ~16 min, so comparable to berfore.

PR #728, opened 2016-07-06
Fixed memory leak when writing h5md files with the tcl interface.

KaiSzuttor:


fweik:
This is why we need an C++/RAII wrapper arround that :-)

KaiSzuttor:
Im working on that ;)

KaiSzuttor:
There still seems to be a memleak, cannot see where...

KaiSzuttor:
It occurs only if h5md_obervable_1D_write is used... I'm looking into that and hopefully will fix that before this can be merged.

KaiSzuttor:
In my opinion the whole interface code for h5md is not worth spending too much time to debug. I found most of the mem leaks and wont spend any additional time with this. Maybe you want to merge this until the h5md is available within the core and we can drop the whole current h5md stuff.

fweik:
Ok. I'll have a look at your ne h5md code right now, maybe I can contribute sth to that.

PR #730, opened 2016-07-11
Tutorial

richter-t:
Lennard Jones Tutorial in Python


fweik:
Looks good. Did somebody else besides you have a look at that? Maybe we should assign that to a student or somebody who has not yet worked with Espresso to test it?

richter-t:
As far as I know nobody has has a closer look yet. So I'm very much in favor of having somebody testing this tutorial.

PR #732, opened 2016-07-13
Removed custom {i,d}{min,max}, isnan and replaced by

fweik:
std function.


fweik:
Sorry for all the formatting changes. (I have clang-format on [TAB] in my emacs :-) ). This should not introduce any behavior changes. (Except for the default myconfig). Maybe the only interesting thing is that I changed the construction by which the features are included, so that it does not make any assumptions on the include order. Also, I removed debug.hpp from utils.hpp and included it implicitly where needed and where edited the files anyway moved up the feature check where possible to speed up compiles.

For the formatting: I am using now the LLVM style of clang-format. This is not because I like that one particularly, but it is the default style of clang-format an is what happens without configuration. Maybe we should do one big reformatting PR, when nothing else is open...

fweik:
Sorry @RudolfWeeber, wrong PR.

PR #735, opened 2016-07-18
Restoration of the commits 9e82b7b and 14d3657

psci2195:
Sorry, I do not know how my changes have reverted this.. Maybe issue of the back-merging.

This pull request is a restoration of the original commits made by @KaiSzuttor 

I will make my branching strategy more conventional to avoid such cases in future.


fweik:
I don't know how I could have missed this :-)

PR #745, opened 2016-07-21
H5mdcore

KaiSzuttor:
Implemented writing H5MD files from within the core. This should be easy extendible to multi-core writing.


KaiSzuttor:
I don't understand why the macro of the hdf5 lib doesn't seem to work on travis, does anybody have an idea?

MrTheodor:
Hi, perhaps it is because of the version of libhdf5 distributed by Ubuntu (and therefore uses by travis), in the package version wasn't compiled with some flags that enables new API. I remember that we had some issue with it, that is why in VOTCA the libhdf5 is build directly from the source.

KaiSzuttor:
thanks @MrTheodor for the comment. Can you give a more concrete hint? Since developing with hdf5 lib, i never encountered  a problem until now. Also tried to build on a newer ubuntu on a machine and it worked.

fweik:
From a first glance this looks like major progress. I'll give more detailed feedback next week.

MrTheodor:
If you try to use new interface v1.8 with travis, it will complain. This is because the Ubuntu 12.04 is compiled with the *--with-default-api-version=v16* flag.
See here: https://bugs.launchpad.net/ubuntu/+source/hdf5/+bug/73488

You can switch travis to use Trusty:  https://docs.travis-ci.com/user/trusty-ci-environment

In VOTCA the libhdf5 is built directly from the source code downloaded from hdf5 web site.
https://github.com/votca/csg/blob/master/.travis.yml

KaiSzuttor:
Thanks a lot @MrTheodor . I will try to do it the same way.

fweik:
https://docs.travis-ci.com/user/common-build-problems/#Troubleshooting-Locally-in-a-Docker-Image might be useful to you...

KaiSzuttor:
thanks @fweik for the docker hint. Wanna give some code review?

fweik:
When I have time. This is number 4 on the queue for Stuttgart I think :-)

fweik:
A couple of CMake things first:
- include_directories(${CMAKE_CURRENT_SOURCE_DIR}/libs/h5xx) Works only for in source builds.
  This should be CMAKE_SOURCE_DIR or similar, but see next point.
- CMake supports dependend includes: See https://cmake.org/cmake/help/v3.0/command/target_include_directories.html This is prefered to top level include.
- Build fails on ubuntu 16.04 with
```
[  0%] Building CXX object src/core/io/writer/h5md/CMakeFiles/H5mdCore.dir/h5md_core.cpp.o
In file included from /home/florian/Source/espresso/libs/h5xx/h5xx/h5xx.hpp:29:0,
                 from /home/florian/Source/espresso/src/core/io/writer/h5md/h5md_core.hpp:33,
                 from /home/florian/Source/espresso/src/core/io/writer/h5md/h5md_core.cpp:23:
/home/florian/Source/espresso/libs/h5xx/h5xx/file.hpp: In constructor ‘h5xx::file::file(const string&, MPI_Comm, MPI_Info, unsigned int)’:
/home/florian/Source/espresso/libs/h5xx/h5xx/file.hpp:146:39: error: ‘H5Pset_fapl_mpio’ was not declared in this scope
     H5Pset_fapl_mpio(plid_, comm, info);
```
with
```
florian@jupiter ~/Source/espresso/build (git)-[h5mdcore] % dpkg -l | grep hdf5                                                                                          :(
ii  hdf5-helpers                                    1.8.16+docs-4ubuntu1                                        amd64        Hierarchical Data Format 5 (HDF5) - Helper tools
ii  libhdf5-10:amd64                                1.8.16+docs-4ubuntu1                                        amd64        Hierarchical Data Format 5 (HDF5) - runtime files - serial version
ii  libhdf5-cpp-11:amd64                            1.8.16+docs-4ubuntu1                                        amd64        Hierarchical Data Format 5 (HDF5) - C++ libraries
ii  libhdf5-dev                                     1.8.16+docs-4ubuntu1                                        amd64        Hierarchical Data Format 5 (HDF5) - development files - serial version
ii  libhdf5-doc                                     1.8.16+docs-4ubuntu1                                        all          Hierarchical Data Format 5 (HDF5) - Documentation
ii  libhdf5-openmpi-10:amd64                        1.8.16+docs-4ubuntu1                                        amd64        Hierarchical Data Format 5 (HDF5) - runtime files - OpenMPI version
ii  libhdf5-openmpi-dev                             1.8.16+docs-4ubuntu1                                        amd64        Hierarchical Data Format 5 (HDF5) - development files - OpenMPI version
```
```
florian@jupiter /usr/include/hdf5/openmpi % dpkg -L libhdf5-openmpi-dev | xargs grep -F "fapl_id, MPI_Comm comm, MPI_Info info);"  2> /dev/null                       :(
/usr/include/hdf5/openmpi/H5FDmpio.h:H5_DLL herr_t H5Pset_fapl_mpio(hid_t fapl_id, MPI_Comm comm, MPI_Info info);
```
Wrong include dir?
- Maybe CMake should run git submodule, not sure about that. But with specific version pinned thats should be safe.

KaiSzuttor:
I think the reason for your failing build is that ubuntu allows to install hdf5 serial and parallel at the same time and the FindHDF5 macro is currently not capable of distinguishing between the two libhdf5.so files although it checks if the lib is build for parallel usage. I'm not a cmake pro but maybe @rbardak could easily adobt the macro.

pdebuyl:
Debian has (obviously) made the same change for HDF5 earlier. A recent enough cmake will detect the serial library with priority and this is settable with `HDF5_PREFER_PARALLEL=true` if you want the parallel version to have priority.

rbardak:
I sent a reply earlier, but it doesn't seem like it arrived.

as @pdebuyl said, setting the `HDF5_PREFER_PARALLEL` is the right choice for cmake 3.4 and greater versions. However we require a lower version. For this, I think we just need to copy paste the find macro from 3.4 into our `./cmake/` folder. Maybe rename the macro, depending on what folder takes priority.

rbardak:
This issue seems to be resolved since 2015. For cmake 3.4 and greater, try
setting the variable `HDF5_PREFER_PARALLEL` to true before finding it.

If this works, I think just copy-pasting the find macro from cmake 3.4 into
our cmake folder should do the trick for versions < 3.4. Maybe we need to
rename the script, depending on what path takes priority.

KaiSzuttor:
thanks @pdebuyl  for this remark. I've already seen this feature in the recent cmake version but since most of the linux distros don't come with cmake 3.4 yet, we can not depend on it. @rbardak I already tried exactly what you said, but this module has some further dependencies within cmake, so it's not that easy.

rbardak:
I have an Ansatz, but I need to test it with every cmake version between 2.8 and 3.4. I will do that when I find the time, I'm on a tight schedule right now. Am I right in the assesment, that if HDF5 is found, the parallel version is always required?

KaiSzuttor:
yes that is right.

CMakeLists.txt
```diff
@@ -96,8 +96,12 @@ if (WITH_CUDA)
...
       set(CUDA_NVCC_FLAGS "${CUDA_NVCC_FLAGS} -O3")
     endif()
-    set(CUDA_NVCC_FLAGS "${CUDA_NVCC_FLAGS} -gencode=arch=compute_20,code=sm_20 -gencode=arch=compute_30,code=sm_30 -gencode=arch=compute_50,code=sm_50")
-
+    set(CUDA_NVCC_FLAGS "${CUDA_NVCC_FLAGS} -gencode=arch=compute_20,code=sm_20 \
+                                            -gencode=arch=compute_30,code=sm_30 \
+                                            -gencode=arch=compute_35,code=sm_35 \
+                                            -gencode=arch=compute_50,code=sm_50 \
+                                            -gencode=arch=compute_52,code=sm_52 \
+                                            -gencode=arch=compute_52,code=compute_52")
```
fweik: ?

src/core/io/writer/h5md/h5md_core.cpp
```diff
@@ -0,0 +1,703 @@
...
+                vel[0][particle_index][1] = current_particle.m.v[1] / time_step,
+                vel[0][particle_index][2] = current_particle.m.v[2] / time_step;
+            if (force)
+                f[0][particle_index][0] = current_particle.f.f[0],
+                f[0][particle_index][1] = current_particle.f.f[1],
+                f[0][particle_index][2] = current_particle.f.f[2];
+            image[0][particle_index][0] = current_particle.l.i[0];
+            image[0][particle_index][1] = current_particle.l.i[1];
+            image[0][particle_index][2] = current_particle.l.i[2];
+            particle_index++;
```
fweik: Please use braces to mark blocks. Is The image stuff really correct?

CMakeLists.txt
rbardak: My thoughts exactly. Doesn't this wreck compile time? Considering that cuda already takes forever for one arch.

KaiSzuttor: Sorry, i mixed up another branch, had problems with my GPU, will revert this commit. Maybe we can even drop all of the flags but this is another story.

rbardak:
@KaiSzuttor I'm done with the find hdf5 stuff, how should we merge it? I was thinking of opening a pull request with your repo, but your branch isn't up to date with the current master

KaiSzuttor:
sounds good @rbardak . I just merged the current master, so you can open a pull request to the h5mdcore branch.

KaiSzuttor:
Travis actually builds, it was a network connection error that broke the last.

fweik:
This code seems awfully repetitive to me. Wouldn't it be easier to just define an list of all the needed groups, instead of repeating all the command for every group?

src/core/io/writer/h5md/h5md_core.cpp
```diff
@@ -0,0 +1,707 @@
...
+         * This has the advantage, that the new file can just be deleted if the simulation crashes at some point and we
+         * still have a valid trajectory, we can start from.
+        */
+        std::string temp_name = filename + "_tmp";
+        this->lastfile = h5xx::file(filename, MPI_COMM_WORLD, MPI_INFO_NULL,
+                                     h5xx::file::out);
+        this->h5md_file = h5xx::file(temp_name, MPI_COMM_WORLD, MPI_INFO_NULL,
+                                     h5xx::file::out);
+        /* particles -- atoms -- box -- edges */
+        H5Ocopy(lastfile.hid(), "/particles", this->h5md_file.hid(), "/particles", H5P_DEFAULT,
```
mkuron: Instead of copying the data, the better solution would be to use `H5Lcreate_external` to link to the old data from the new file.

src/core/io/writer/h5md/h5md_core.cpp
```diff
@@ -0,0 +1,707 @@
...
+        this->dataspace_particles_atoms_species = h5xx::dataspace
+                (chunk_dims_1d);
+        this->dataset_particles_atoms_species =
+                h5xx::create_dataset(this->group_particles_atoms, "species",
+                                     this->type_int, this->dataspace_particles_atoms_species);
+        this->WriteSpecies();
+        /* particles -- atoms -- parameters */
+        this->group_parameters = h5xx::group(this->h5md_file, "parameters");
+        this->group_parameters_vmd_structure = h5xx::group(
+                this->group_parameters, "vmd_structure");
```
mkuron: What's in the `/parameters/vmd_structure` group? There shouldn't be anything VMD-specific in these files as H5MD should be readable by all kinds of visualization programs.

KaiSzuttor: Actually the H5MD proposal allows any additional groups, so if we want to have it compatible to VMD, we can just add the respective structure and it does not hurt any other tool or program.

src/core/io/writer/h5md/h5md_core.cpp
KaiSzuttor: Thats not possible since the external links are actually hard-links. So if you append to these, you have exactly the problem I want to avoid.

src/core/io/writer/h5md/h5md_core.cpp
mkuron: I was just wondering if there's anything in there that might be interesting to other software as well.

src/core/io/writer/h5md/h5md_core.cpp
mkuron: Don't you store time series? So wouldn't this allow you to refer to another files for earlier time steps? Or is the problem that there is not one object per time step and data field, but one object per data field and each time step is appended to that same object?

KaiSzuttor: The latter, each timestep is appended to the same hdf5 object, i.e. to the dataset for the respective trajectory, velocity, etc.

mkuron: Ok, then that seems like a slightly poor choice in the design of H5MD (though they may have had other goals). There's nothing you can do about it.

pdebuyl: Hey,

One other solution would be to copy the file with a system call:
    mv file.h5 .file.backup.h5
    cp .file.backup.h5 file.h5

so that you can recover the backup in case of corruption. @petercolberg 's nano-dimer https://colberg.org/nano-dimer/ has a snapshot mechanism that is also interesting.

The storage as single datasets has advantages, as for too numerous datasets hdf5 files grow poorly (think of 10^5 - 10^7 steps).

pdebuyl: @mkuron Regarding the choice of dataset storage, the use of a large number of datasets is known to result in poor performance (from following the mailing list of the hdf group). If you have specific questions, feel free to contact the H5MD mailing list https://lists.nongnu.org/mailman/listinfo/h5md-user

KaiSzuttor: @pdebuyl Thank you very much for your comments. I just looked into the code of Peter Colberg but could not find the snapshot mechanism you mentioned in the source code. Can you guide me to the source?

pdebuyl: `nano-dimer` is implemented in Lua and OpenCL and is not easy to navigate at first. I tried to share my first steps in it in a blog post if you are interested: http://pdebuyl.be/blog/2015/opencl-nanomotors.html

Specifically for snapshotting:
1. The file `nanomotor/snapshot.lua` contains the generic mechanism for snapshotting. It relies on objects having defined a `snapshot` and `restore` function.
2. For specific objects, such as trajectories and observables, the functions are defined and explicity save the object's state to disk (see e.g. `nanomotor/thermo.lua`) that defines the storage of the thermodynamic observables.
3. The global state of the program collects all this information. This is implemented in `nanomotor/snapshot.lua` that takes care of a "domain" and is provided with observables (last function in the file).

The snapshot filename and interval are defined in the configuration file of a simulation. The snapshot file is different from the trajectory file in the sense that, for instance, the solvent information is stored fully for every snapshot. If I understand well, the snapshots are also deviating from the HDF5 definition.

A snapshot is only deleted after the next snapshot has been fully written. This is achieved by a system call to rename `snapshot_filename.h5.tmp` to `snapshot_filename.h5` after closing the HDF5 file. (again, end of `nanomotor/snapshot.lua`).

The code can be found at https://git.colberg.org/nano-dimer/tree/

hirschsn:
I just noticed that even the current sequential version (so this is not only a problem when going parallel) will not give you consistent trajectories. We have to use e.g.
```
pos[0][current_particle.p.identity][0] = current_particle.r.p[0];
```
in the particle loop in File::WriteTimeDependant3D to achieve that, since the ordering of particles in cells is somewhat arbitrary.

However, this will not work in parallel where we decided to dump the ids. The problem with this approach that the particles/position/value dataset won't contain valid trajectories either.

Any comments on this problem? (@KaiSzuttor,  @jonaslandsgesell, others welcome)

mkuron:
You need to write the ID for every particle in every time step. It is the viewer application's responsibility to define the "trajectory _x_" as "the position of the particle with ID _x_", instead of "the position of the particle number _x_".

KaiSzuttor:
@mkuron is right. Actually we will do it as we already discussed on the coding day. As far as i remember the plan was to do write IDs as a time-dependent group into the hdf5 file. Correct me, if I'm wrong.

hirschsn:
Fine with me. I suggest, we make writing out masses and types a feature (like positions, velocities and forces) and not statically dump them once. Then, the user can select to write them out if the number of particles changes. (Which is one point we agreed on to support.)

fweik:
Ok, that looks very good. It would be nice to have unit tests for the core class.
Could you please resolve the merge conflict.

KaiSzuttor:
@fweik can you please specify what should be covered by core unit tests? I already included interface tests.

fweik:
Can you please remove boost::{timer,chrono} from the required boost modules. They are not actually needed and are not installed in the docker containers used for testing.

fweik:
In a pull request $TRAVIS_BRANCH contains the branch the pull request is against

KaiSzuttor:
Yes but that breaks the possibility to run travis checks for the users repo

KaiSzuttor:
All tests pass except the non-docker tests for mac. submodule is not found, anybody has an idea?

mkuron:
Something must be wrong with your preprocessor include path if it doesn't find h5xx on Mac.

KaiSzuttor:
d656925c1336b8be3b946f4b8c0dea9b080a14c1 broke the test but I could not figure out why yet.

KaiSzuttor:
@fweik ready for review, i somehow cant restart the failing mac job

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "libs/h5xx"]
```
fweik: Maybe we should fix a revision of h5xx that is known to work?

maintainer/travis/build_docker.sh
```diff
@@ -17,15 +17,16 @@ if [ -z "$image" ]; then
 fi
 
 if [ "$TRAVIS_OS_NAME" = "linux" ]; then
-	image=espressomd/buildenv-espresso-$image
-	docker run -u espresso --env-file $ENV_FILE -v ${PWD}:/travis -it $image /bin/bash -c "git clone /travis && cd travis && maintainer/travis/build_cmake.sh"
+	image=espressomd/buildenv-espresso-$image:python
```
fweik: This should read $TRAVIS_BRANCH

mkuron: Why do we even have both branches on Docker Hub? Just use the Dockerfile from the python branch to build both the master and python branches.

.gitmodules
mkuron: It is already fixed to a version, fedf38a0bc58ff5ff30c46819d64eb7b8c644c60.

maintainer/travis/build_docker.sh
KaiSzuttor: If the tag is not hardcoded here, users cant check on travis in advance. Is this what we want?

src/core/particle_data.cpp
```diff
@@ -2304,8 +2295,3 @@ bool particle_exists(int part) {
         return true;
    return false;
 }
-
-#ifndef MASS
-constexpr double ParticleProperties::mass;
```
KaiSzuttor: Okay, I didn't know that and the tests succeeded, can you explain why this is needed?

maintainer/travis/build_docker.sh
mkuron: There is no reason why the Docker image should depend on whether we're building master or python. Docker Hub should use the Dockerfile from the python branch to build the default tag (latest), just as before.

KaiSzuttor: I think you're point is cleat @mkuron

src/core/particle_data.cpp
fweik: Because the standard says so (9.4.2): "A static data member of literal type can be declared in the class definition with the constexpr specifier; if so, its declaration shall specify a brace-or-equal-initializer in which every initializer-clause that is an assignment-expression is a constant expression. [ Note: In both these cases, the member may appear in constant expressions. — end note ] The member shall still be defined in a namespace scope if it is odr-used (3.2) in the program and the namespace scope definition shall not contain an initializer."

KaiSzuttor: I included the declaration again at https://github.com/KaiSzuttor/espresso/blob/4c15c144bc763e603cf910b1a50a36fcef2ae3e4/src/core/particle_data.cpp#L705-L707

fweik:
See #941

PR #749, opened 2016-07-22
Cmake install

rbardak:


fweik:
Should we install the samples or parts of the documentation, tutorials?

rbardak:
Depends. What's the main purpose of an installation? I don't know, but I could imagine the main purpose for that is to run espresso on multiple hosts. For that purpose I think a test script which checks that mpi doesn't throw a tantrum for the installed version would be more important.

Also, since one has to switch a few features on and off for a particular purpose and recompile espresso, I would think it's highly unlikely that a user doesn't have access to the source/docu one way or another. I think a simple `Espresso --help`, which mentions where to find the source code and documentation would be enough.

fweik:
One use for the installation would be distribution packages, with some "reasonable" feature set.

rbardak:
Depends on whether or not there is such a thing as a reasonable feature set, which covers most "day to day" use cases and only provides methods which can be used as a black box. In this case, adding some tutorials and docu which cover at least the feature set would be a good idea.

rbardak:
On a personal note, since espresso is used for scientific purposes, I believe shipping pre compiled packages without full source would convey the wrong message.

junghans:
I disagree - pre-compiled packages can act as a good starting point to attract new user und ultimately developers.

Looking at:
<https://build.opensuse.org/package/view_file/Education/espresso/espresso.spec?expand=1>
it seems installing documentation is not necessary!

fweik:
I'll merge it for now as it is. Maybe I'll add documentation or a man page later. Just so that if you install the package you can get a pointer as to where you can find further information.
I agree with Christoph that having the entry barrier as low as possible is beneficial for any open source project, scientific or not. I think that there is no real difference to other FOSS projects, the question whether it is PhD or spare time that goes into the project doesn't really matter.

PR #751, opened 2016-07-23
python: Removed tcl python modul(Broken, not easily fixed.). Removed …

fweik:
…some other tcl stuff and fixed build w/o tcl-dev.

Not sure if this needs discussion.

@RudolfWeeber  @tobiasrau


RudolfWeeber:
Can be merged.

PR #752, opened 2016-07-25
cmake: add so version to libs

junghans:


fweik:
Wouldn't it be more ... idiomatic to set the project version and have cmake set the shared object versions accordingly?

junghans:
Sure! That makes sense!

junghans:
Do you want to keep whole version or just the first digit, e.g.
https://github.com/votca/tools/blob/master/CMakeLists.txt#L6

fweik:
I think major and minor is the usual way, but it does not really matter because we don't guarantee anything about API/ABI stability

junghans:
Using the full version only makes sense if we guarantee API/ABI stability, which we don't. So, I just derived the version from the `PROJECT_VERSION`.

PR #754, opened 2016-07-25
added in some documentation on installing on ubuntu 16.04 and ccmake

tigoesnumb3rs:


doc/ug/introduction.tex
```diff
@@ -321,6 +321,19 @@ \section{Requirements}
   version 1.2).
 \end{description}
 
+\subsection{Installing Requirements on Ubuntu 16.04 LTS}
+\label{sec:installing_requirements}
+\index{installing_requirements}
+
+To make ESPResSo run on Ubuntu 16.04 LTS its dependencies can be installed with:
```
mkuron: This is not specific to Ubuntu or that version. I guess that it should also work on all supported Ubuntu versions (12.04, 14.04, 15.10 and 16.06) and Debian versions (7 and 8). Please check the package names on these distributions and make sure they already have the `apt` command (older versions may require `apt-get`). Then make the text more generic as appropriate.

doc/ug/introduction.tex
```diff
@@ -321,6 +321,19 @@ \section{Requirements}
...
+\begin{code}
+  apt install build-essential cmake cython python-numpy tcl-dev
+  tk-dev libboost-all-dev openmpi-common
```
fweik: TK is no longer needed (nor used)

RudolfWeeber: So did someone change the tcl finding routine in cmake, because that
at one point required tk, eventhough espresso wasn't using it.

fweik: Yes I did, see https://github.com/espressomd/espresso/commit/657a853ee7856170c8c357a794cd2a4d3a09a30d

fweik:
I am currently experimenting with docker to test building on different distros. It would be nice if we had a dependency list (let's say a minimal and a full one) that could be automatically tested and included in the docs. The same goes for build instructions.

doc/ug/introduction.tex
fweik: There are differences, because on the older ubuntus and debians cython is not new enough and has to be install via pip. 16.04 ships cython 0.23.4

RudolfWeeber:
I suggest to merge this as it is probably quite helpful to new users. 
We can add instructions fore more distributions, as we test them via docker or the like.

fweik:
Yes, just wanted to say that we should think about away way to write down install instructions that facilitates automation better than a latex document.

PR #757, opened 2016-07-27
added python interface for tabulated nonbonded

richter-t:


KaiSzuttor:
@richter-t testcase? :)

KaiSzuttor:
Can be merged.

RudolfWeeber:
There are still 2 issues with the test case.
1. Please cast the testcase into the python unit-test framework. See coulomb_cloud_wall ore something similar.
2. Please add the test to the CMAKeList.txt, so it actually gets run.

richter-t:
To point 1: not applicable, this is a physical test, point 2 done

fweik:
We use the unittest framework for all tests in python.

PR #758, opened 2016-07-27
added python interface for core random number generator

jonaslandsgesell:
pickle should now in principle be able to get and set the random number generator state in the core.


fweik:
Could you please add a C++ level function that returns the state size for the RNG? The interface should work with any engine.

jonaslandsgesell:
Added state size interface and added the possibility to set the state of the Mersenne Twister RNG with real random numbers from os.urandom() via pythons SystemRandom.

fweik:
Yes, that is clear. The Engine concept from the standard does not include the state size, because there are implementations for which this makes no sense (e.g. std::random_device).

PR #761, opened 2016-07-27
Changed seeding of rng

jonaslandsgesell:
According to sackoverflow this seeding procedure is suggested:
https://stackoverflow.com/questions/15509270/does-stdmt19937-require-warmup 
```
std::seed_seq seeder{seed};
generator.seed(seeder);
```
seed_seq "tries to come up with a sane initialization" (see second source). 
However this is still not enough for a good seeding, see e.g. http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0205r0.html#ref-01 or http://www.pcg-random.org/posts/cpp-seeding-surprises.html

Of course "good seeding" with only one 32bit integer cannot be done truely for the mersenne twister with its huge internal state which cannot be filled with only one 32bit integer. Maybe the most clean way to seed the mersenne twister in the core would be to provide a function set_random_seed() in the python interface which sets the state of the mersenne_twister in the core with true random numbers.

### DIFF
--- a/samples/python/minimal_random_number_generator.py
+++ b/samples/python/minimal_random_number_generator.py
@@ -38,5 +38,5 @@ for i in range(len(rng_state_read1)):
 	rng_state.append(i)
 system.random_number_generator_state=rng_state
 rng_state_read2=system.random_number_generator_state
-print "random number generator state read 2", rng_state_read2, "should be ", rng_state
+print "random number generator state read 2", rng_state_read2,
 
--- a/samples/python/minimal_random_number_generator.py
+++ b/samples/python/minimal_random_number_generator.py
@@ -18,10 +18,6 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 import espressomd
-from espressomd import thermostat
-from espressomd import integrate
-from espressomd import interactions
-from espressomd import diamond
 import numpy
 import sys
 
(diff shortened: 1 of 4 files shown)


fweik:
I don't see your point. Have you noticed correlations in the random numbers? You realize that the discussions that you cite are mostly about predictability, which is not relevant for our application.
I don't see the point for sampling applications to use std::random_device for initialization, what is the benefit of that for sampling applications?

jonaslandsgesell:
As outlined in the second source (http://www.pcg-random.org/posts/cpp-seeding-surprises.html) bad seeding of the RNG can also introduce a bias. In Monte-Carlo Simulations this bias might kill your statistics. Suppose you are on a lattice and want to propose a new coniguration within the original Metropolis algorithm. This algorithm assumes that the proposal probabilities are symmetric:
g(x->x')=g(x'->x)
If x' is however a number that cannot be drawn by the RNG due to a bias (e.g. x'=7 or x'=13 in the example of O'Neill ) then  you violate this assumtion of the algorithm. This could lead to wrong simulation results.

> But a good rule of thumb is that if you're so cheap that you want to use a single 32-bit integer to initialize your random number generator, you should probably use an RNG with a 32-bit state. Otherwise, you should use the amount of seed randomness your generator actually needs—don't try to give it less. O'Neill 

Therefore I do not see a point why we should initialize the RNG with less random bits than needed. It avoids the risk of running into bias problems.

And yes even the code provided above does not fix this problem.  It only invokes std::seed_seq which avoids too many 0s in the seed;

> The seed_seq algorithm avoids this issue (when given a single integer), resulting in an initialization that's relatively sane.  O'Neill 

 However it does not fix the problem of bias. This would only be fixed by introducing a seed_function which uses 624 real random numbers. So let me ask the counterquestion: Why should we expose the user to the risk of bias when we can avoid it?

fweik:
If you really concerned about this you can change the complete RNG state from the script level, or drop the first ~ 1e6 random numbers.

jonaslandsgesell:
I added dropping the first 1e6 random numbers after seeding with one random number.

PR #763, opened 2016-07-27
OpenGL Visualizer

KonradBreitsprecher:
..plus minimal example.


mkuron:
Could you rename the class to `espressomd.visualization.opengl_live` to be consistent with my `espressomd.visualization.mayavi_live`? Also, please either rename the `start` method to `run_gui_event_loop` or rename that method in the Mayavi stuff (sample, tutorial, and code)?

I think we should have all visualizers conforming to a simple, common interface so that one can easily switch between them. You should also add a `register_callback` method so one can run Matplotlib in parallel wirh the visualization.

RudolfWeeber:
On Wed, Jul 27, 2016 at 09:20:28AM -0700, Michael Kuron wrote:
> Could you rename the class to `espressomd.visualization.opengl_live` to be consistent with my `espressomd.visualization.mayavi_live`? Also, please either rename the `start` method to `run_gui_event_loop` or rename that method in the Mayavi stuff (sample, tutorial, and code)?
Wouldn't following pep8 mean MayaviLive and OpenGlLive or the like?

mkuron:
You're probably right about the PEP8 stuff. So Konrad, would you please make things consistent between the two visualizers and also rename my class in a PEP8-compliant fashion?

mkuron:
Maybe we should rename `run_gui_event_loop` to `start` to make it more obvious -- technical details like the existence of event loops need not concern the user.
samples/python/visualization_bonded.py and samples/python/lj-demo.py need to also be adapted to the new names of classes and methods.

I will do a full code review next week.

fweik:
I am currently experimenting with docker to test building on different distros. It would be nice if we had a dependency list (let's say a minimal and a full one) that could be automatically tested and included in the docs. The same goes for build instructions.

samples/python/visualization_bonded.py
```diff
@@ -35,10 +37,18 @@
...
-  integrate.integrate(int_steps)
-  mayavi.update()
+def main():
+    for i in range(0,int_n_times):
+        print(i)
+        integrate.integrate(int_steps)
+        visualizer.update()
+
+#Start simulation in seperate thread
+t = Thread(target=main)
```
mkuron: Threads are not strictly needed here because we don't register any callbacks.

mkuron:
Two things:
- make the naming consistent: currently we have `espressomd.visualization.mayaviLive` and `espressomd.visualizationOpenGL.openGLLive`. The latter should be named `espressomd.visualization.openGLLive`. I think you should rename _visualization.pyx_ to _visualizationMayavi.pyx_ and create a file _visualization.pyx_ with the following contents:
```
from .visualizationMayavi import mayaviLive
from .visualizationOpenGL import openGLLive
__all__ = ['mayaviLive', 'openGLLive']
```
- _samples/python/lj-demo.py` also uses visualization. That needs to be adapted to all name changes.

I didn't read through the entirety of _visualizationOpenGL.py_ (and I assume you will still be changing that file quite a bit in the near future), but everything else looks good.

mkuron:
Also, please try replacing pyopengl with pyopengl-accelerate if available. According to the documentation, the two are interchangeable, but the latter uses Cython to do native OpenGL calls. This should be able to speed up the visualization significantly.

KonradBreitsprecher:
Done with the stated issues. As far as I understood, I don't use any features that benefit from pyopengl-accelerate. And there is nothing like OpenGL_accelerate.GL (or GLU, GLUT).

mkuron:
_samples/python/lj-demo.py_ hasn't been adapted to the new class names yet. Please fix that.

Regarding `OpenGL_accelerate`: as the documentation is quite sparse, I checked the code on how it is used. It seems like the `OpenGL` module automatically imports it if it's present and its main purpose is to reduce conversions between C arrays and Python lists. It doesn't appear to help in optimizing PyOpenGL calls from Cython programs (like putting native OpenGL calls into the Cython-generated C code). So you don't need to do anything, but I would be interested to see some profiler output to see how much of a bottleneck PyOpenGL is.

KonradBreitsprecher:
Simply adding  `OpenGL_accelerate` did not change the performance in a naive timing test. Anyways, I the bottleneck is in the integration, a larger system can never be integrated and visualized smoothly, so some small improvement in PyOpenGL will not help too much.

These commits are now part of #817 due to a unclean starting version on the last coding day, sorry for that.

RudolfWeeber:
Can this be closed? (@KaiSzutor)

RudolfWeeber:
Sorry, wrong ticket

PR #765, opened 2016-07-27
cmake: fix regex for soversion & stricter check

junghans:


junghans:
Currently libs are installed as `libXXX.so.4.0-dev`, which is wrong. It should be `libXXX.so.4` instead.

PR #769, opened 2016-08-02
doc: make it work with texlive-2016 (fixes #768)

### LINKED ISSUE #768: Documentation doesn't build with texlive-2016
See https://build.opensuse.org/public/build/home:cjunghans:branches:Education/openSUSE_Factory/i586/espresso/_log:
```
! Class scrreprt Error: undefined old font command `\bf'.
```

I think there are two solutions:

1. replace `{\bf ...}` by `\textbf{...}`
2. Use class option `enabledeprecatedfontcommands`

junghans:
Replace \bf -> \textbf, \tt -> \texttt, \it -> \textit
        \rm -> \textrm

### DIFF
--- a/doc/tutorials/tcl/01-lennard_jones/01-lennard_jones.tex
+++ b/doc/tutorials/tcl/01-lennard_jones/01-lennard_jones.tex
@@ -1108,7 +1108,7 @@ exit
 
 \end{minipage}}\vspace{1cm}
 
-\noindent {\bf Hint:} \texttt{analyze pressure} (without \texttt{total}) returns the pressure and
+\noindent \textbf{Hint:} \texttt{analyze pressure} (without \texttt{total}) returns the pressure and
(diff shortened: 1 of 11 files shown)


fweik:
@hmenke Do we like that?

hmenke:
@fweik No, we don't like that, because this is just mindless search-and-replace.  For example `{\bf v}` in math mode should be replaced by `\mathbf{v}` (even nicer would be `\bm{v}` with the `bm` package, but that might entail a complete overhaul of the manual).  It might also cause error in the future when people copy-and-paste this stuff.  Try for instance `$\textbf{\epsilon}$`.

junghans:
I did another round with `\mathbf{}`.

hmenke:
@junghans Looks good.  Thanks for the effort.  @fweik Can be merged.

Issue #776, opened 2016-08-05
Cannot build with autotools anymore (on OpenSuse leap 42.1 at icp)


RudolfWeeber:
This was from a clean github master

# configure
[...]
****************************************************************
*                  Checking for programs                       *
****************************************************************
/home/weeber/es-tmp/configure: line 4745: syntax error near unexpected token `disable-static'
/home/weeber/es-tmp/configure: line 4745: `LT_INIT(disable-static)'

KaiSzuttor:
this is because bootstrap.sh previously fails, we have to run libtoolize before running bootstrap.sh.

fweik:
This seems to be an ICP issue. I can't reproduce this in a fresh opensuse container. This is odd.
@KaiSzuttor libtoolize should be run by autoreconf (call from bootstrap.sh), which should recognize
that libtool is used. This does not work on the ICP system.

fweik:
Why was there an alocal in /usr/local/bin in the first place?!

KaiSzuttor:
I dont know, maybe autotools package has been too old for us at some time? But the question is rather why does it corrupt the bootstrap.sh script?

PR #778, opened 2016-08-05
Added libtoolize command to boostrap.sh.

KaiSzuttor:
#776

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -2,5 +2,6 @@
 
 cd `dirname $0`
 
+libtoolize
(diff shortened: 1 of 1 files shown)


fweik:
Do you have any understanding of this? This seems OpenSuSE specific, but I can't figure out what is even the problem...

KaiSzuttor:
Actually this was just a dirty quick fix. I found the reason: aclocal from /usr/local is used for the configuration but doesnt seem to work probably, I now moved the bin in /usr/local to .bak and it works, so issue and PR can be closed.

PR #788, opened 2016-08-15
C++11 cuda build

psci2195:
Hi all,

Probably (please correct me if I'm wrong), the proposed here changes are required for any C++11 (```cmake``` based) builds of latest CUDA.

My environment:

```
nvcc --version
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2015 NVIDIA Corporation
Built on Tue_Aug_11_14:27:32_CDT_2015
Cuda compilation tools, release 7.5, V7.5.17
---
gcc --version
gcc (Ubuntu 4.8.5-2ubuntu1~14.04.1) 4.8.5
```

One looks like aligned with
https://lists.nongnu.org/archive/html/espressomd-users/2016-02/msg00002.html

The issue (similar nature): http://stackoverflow.com/questions/29121211/cuda-compilation-issue-with-cmake

Regards, Bogdan T.


psci2195:
Runtime details.. No changes were detected (compare to ```make``` version): performance of test ```dawaanr-and-dds-gpu.tcl``` is the same, it is passed well. Additional note: my CUDA compute capability is 2.0 (Tesla C2070).

fweik:
The HAVE_CXX11 define is actually only there for historic reasons and can be removed. Espresso needs a CUDA compiler that has c++11 support (e.g. 7.0 or newer), and builds cuda code as c++11.

psci2195:
Thanks. I've just completed some cautious cleanup of the HAVE_CXX11 and pushed here..

fweik:
Thank you for the cleanup.

Issue #742, opened 2016-07-19
py: Complete system class


RudolfWeeber:
* Support non-periodic boundary conditions
* Move stuff out of the system class, which doesn't belong there (see below)
* (probably) change read-only properties into functions
* Document

The following attributes of the system class should go to other places
(The reason that they are in the system class now is just that they are accessed via setmd in tcl)

This should go to the thermostat class, probably as part of its get_state method:
gamma
temperature
nptiso_gamma0
nptiso_gammav
npt_p_ext
npt_p_inst
npt_p_inst_av
npt_piston
npt_p_diff
dpd_gamma
dpd_r_cut
dpd_tgamma
dpd_tr_cut
dpd_twf
dpd_wf
thermo_switch

This should go to the particleList class:
max_part (PArticleList.highest_particle_id())
n_part (Is already implemented as len(ParticleLIst))
n_part_types 
n_rigidbonds

The following should probably go to the CellSystem class, part of it into the get_state() method.

local_box_l
max_num_cells
min_num_cells
max_cut
max_range
max_skin
n_layers
node_grid
cell_grid
cell_size
n_nodes
max_cut_bonded
max_cut_nonbonded
skin
verlet_reuse

tobiasrau:
I'm working on that.

Issue #741, opened 2016-07-19
Py: Complete thermostat


RudolfWeeber:
* Implement set_npt(), coordiante with #739
* Support rotational gamma for set_langevin()
* Implement pickle get_state and set_state methods
* Py docs for the thermostat

tobiasrau:
I'm working on that.

Issue #739, opened 2016-07-19
py: Complete integrators


RudolfWeeber:
* Integrators (at this point velocity_verlet and minimize_energy) should have a common interface and common base class. 
* Instances should be available as attributes of the System class like
  system.velocity_verlet
  system.minimize_energy
  system.brownian_dynamics (no py interface yet)
* The integrate/run method should probably be called the same for all of them.
* Add, whatever is missing regarding npt
* Python docs for all available integrators

tobiasrau:
I'm working on that.

PR #792, opened 2016-08-18
System, Integrator, Thermostat

tobiasrau:
Adresses #741, #742 and #739


RudolfWeeber:
Thx. I'll have a look.

RudolfWeeber:
In general, it looks good.
Please address the following issues, though:
* When building with the maxset myconfig, not all py tests pass
* In cellsystem.pyx, please remove the setter for node_grid, as this is not user-settable
* I had the notion that diffrent integrators woudl go into different classes,
  but I'm also with the way it's done here, as it also unifies the interface
* The tutorials scripts and tex files need to be adapted

RudolfWeeber:
* There still is a prob with the cellsystem test case
* (fragments of) the tutorial scripts are also in the .tex-files and also need to be changed

RudolfWeeber:
Looks good. I think, it can be merged.

fweik:
Handled elsewhere

PR #794, opened 2016-08-19
closes #793 generate code_info & set dependency

### LINKED ISSUE #793: Py: autogeneration of code_info.pyx does not work correctly
Currently, the file code_info.pyx is written into the source directory, but not registered for compilation, probably because the file match *.pyx is executed, before the code_info.pyx is generated (i did a parallel build).

Either, the remaining *.pyx and *.pxd builds need to depend on the code_info generation, so that the code_info.pyx is there in time, or, code_info.pyx needs to be written in the build dir and compiled there.

Workaround: delete src/python in the build dir and make. The second time, the pyx-file is already there at the beginning, so it works.

rbardak:
code_info.pyx is now generated in the build dir (as it should be)

removed its header as it serves no purpose

Note: please remove the code_info.pyx file which was generated in your src dir from previous builds. In the future please avoid generating and editing files in the source directory during build, the source dir should be read only.


fweik:
Well
```
CMake Error at src/python/espressomd/CMakeLists.txt:75 (add_library):

  add_library cannot create target "code_info" because another target with

  the same name already exists.  The existing target is a shared library

  created in source directory

  "/home/travis/build/espressomd/espresso/src/python/espressomd".  See

  documentation for policy CMP0002 for more details.
```

rbardak:
Works for me, I suspect the error is caused by an existing, untracked `code_info.pyx` in the source directory, which was generated by a previous build with an outdated `CmakeLists.txt`

fweik:
Nope, that was on travis

rbardak:
yea, I'm trying to reproduce that atm

rbardak:
Alright found the problem: travis builds inside the source dir. All builds which failed reconfigured after already building espresso successfully, thus causing a duplicate mention of `code_info.pyx` in the variable `${cython_SRC}` during the second configuration. Building outside of source does not cause this issue.

The error can be avoided with `list(REMOVE_DUPLICATES)` to counter the issue from the ill-advised in-source build. However we should adress the fact that travis builds in-source, as this may break the build in a way that may not be circumvented, for instance when generating files in the build dir with the same name as the original file.

RudolfWeeber:
That was fast. Thanks!

Issue #790, opened 2016-08-18
Py Bonded, non-bonded inter, actors need parameter syncing


RudolfWeeber:
after calling the setter in the core, 
self._parmas=self._get_params_from_es_core()

needs to be called, because otherwise parameters calculated by the core are not up to date.

PR #809, opened 2016-08-29
added params update implicit

tobiasrau:
#790


RudolfWeeber:
Looks good. 
However, generally speaking, we should probably not overdo it with this kind of magic

PR #812, opened 2016-08-29
Python 3 support

### LINKED ISSUE #780: Check if python interface can be easily made python3 compatible
This is a TODO for the next coding day.

mkuron:
Here's Python 3.x support. Closes #780.
Adding it was pretty much trivial, but below are a few explanations to what I did:
- Default arguments inside `cdef cppclass` don't work. Instead I now declare the individual overloads.
- Changed every use of `print` to the new function syntax.
- Cython is now called with `-2` or `-3` so that the Cython code's syntax needs to match that of the Python version in use. This isn't strictly necessary as the generated C code contains `#ifdef`s so it can be used with both Python 2 and 3, but I did it anyway for the sake of consistency.
- I ran `2to3`, which fixed things like the `cPickle` module or the `xrange` builtin having been eliminated and some explicit casts to `list` being required.
- Massive changes to the imports inside the `espressomd` module were necessary due to Python 3 dropping support for implicit relative imports, see [PEP 328](https://www.python.org/dev/peps/pep-0328/). So the following replacements were necessary:

  Old|New
  ---|---
  `import x` | `from . import x`
  `from x import y` | `from .x import y`
  `cimport x` | `from . cimport x`
  `from x cimport y` | ~~`from .x cimport y`~~*  `from espressomd.x cimport y`

Once @fweik publishes the [Espresso Docker file](/fweik/espresso-docker), we should duplicate it to also build a container that contains Python 3.5 instead of Python 2.7.

By default, Python 2 will be used if available. If it is not available but Python 3 is available, that is used instead.

Could one of the Cython experts (@tobiasrau, @RudolfWeeber) have a look at this pull request to make sure that none of the changes have adverse effects? I don't see any problematic ones, but especially the relative import stuff could use a second pair of eyes.

---
\* Logically, this should work, but Cython doesn't appear to support it (yet).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -114,7 +114,7 @@ if (WITH_PYTHON)
   find_package(NumPy REQUIRED)
   list(APPEND LIBRARIES ${PYTHON_LIBRARY})
   execute_process(COMMAND ${PYTHON_EXECUTABLE} 
-                -c "import distutils.sysconfig as cg; print cg.get_python_lib(1,0,prefix='${CMAKE_INSTALL_EXEC_PREFIX}')"
+                -c "import distutils.sysconfig as cg; print(cg.get_python_lib(1,0,prefix='${CMAKE_INSTALL_EXEC_PREFIX}'))"
                 OUTPUT_VARIABLE PYTHON_INSTDIR OUTPUT_STRIP_TRAILING_WHITESPACE)
 endif(WITH_PYTHON)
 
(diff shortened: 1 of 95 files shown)


RudolfWeeber:
It looks ok to me.

Is it possible to switch the python version manually, if both are available. 
We need to keep in mind that, once the tutorials have converged, they need to be 2to3'd, too.

mkuron:
Yes, you can manually force Python 3 if your system has both 2 and 3 installed as follows:
```
rm -f CMackeCache.txt
cmake .. -DPYTHON_EXECUTABLE=$(which python3)
```

Note that you cannot switch the Python version after you have run `cmake`, you will get a messy combination of libraries and interpreters from both Python versions. You need to delete your _CMakeCache.txt_ to force re-detection of Python.

tobiasrau:
There are many unnecessary brackets now.

`- print("Features: ", espressomd.features())`
`+ print(("Features: ", espressomd.features()))`

Apart from that, it looks good.

mkuron:
Parentheses fixed and PR is thus ready to merge.

Issue #789, opened 2016-08-17
Py: In many places, runtime error collection is missing


RudolfWeeber:
for instance ParticleHandle.vs_auto_relate_to()

PR #815, opened 2016-08-29
Runtime error handling

hmenke:
Addresses #789 

- Runtime error handling moved to `utils.pyx`
- Runtime error handling in interface
- Removed some unimplemented function prototypes

### DIFF
--- a/src/core/pressure.hpp
+++ b/src/core/pressure.hpp
@@ -526,16 +526,7 @@ int local_stress_tensor_calc (DoubleList *TensorInBin, int bins[3], int periodic
 int observable_compute_stress_tensor(int v_comp, double *A, unsigned int n_A);
 
 void update_pressure(int v_comp);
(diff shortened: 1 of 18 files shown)


RudolfWeeber:
This pr will catch runtime errors which occur without the call to the core returning an error.
However, in this pr, the call to handle_error() is often placed after the exception is thrown. this means that runtime error messages from the core will often not be displayed.

E.g.
if espresso_core_call():
    raise Exception("It didn't work")
handle_errors("message")

The solution implemented for tcl is as follows:

if handle_errors(espresso_core_call()):
  raise Exception("It did not work")

Then, handle_errors():
* displays runtime error msgs
* returns 1 if the core call returned 1 or there were runteim errors
* returns 0 otherwise

There are probably also other solutions.

richter-t:
#847

PR #814, opened 2016-08-29
Various py fixes and extensions

RudolfWeeber:
See commit summary.

This should be merged after the complete_system_class branch from @fweik, which contains refactoring done by @tobiasrau.

PR #817, opened 2016-08-29
Python Tutorial 02-charged_systems + OpenGL Visualization

KonradBreitsprecher:
-Revised the tutorial (removed pickle stuff and offline analysis script). 
-Added analyze.append() method.
-Fixed Bug in actors (set_stuff_in_es_core was called on actor instantiation)
-Merged the OpenGL visualization
-TODO: Once we have constraints, I will add the partial periodic section in the tutorial


mkuron:
The OpenGL stuff from #763 is fine, all the stuff discussed over there has been addressed already.

RudolfWeeber:
> -TODO: Once we have constraints, I will add the partial periodic section in the tutorial
I removed the restriction to 3d periodicity in #814, but probably,
the wall constraint is also needed for this tutorial?

KonradBreitsprecher:
Found the issue related to the failing check_python of this PR:

Actor should not call `_set_params_in_es_core()` on instantiation:
When creating an actor (without adding it to the system), `__init__()` calls `_set_params_in_es_core()`.
One rat's tail was a crash on integration when just creating and not adding a P3M actor.
So I removed the `_set_params_in_es_core()` and added it in the `activate()` method of the electrostaticInteractions, if it wasn't already there. 

Now the unit-tests of P3M and DH fail because they don't add the actors and still expect the params from the core. So this fixes it for P3M and DH:
```
Inter = interClass(**params)    #<--this should not set the params in the core
Inter.validate_params()
Inter._set_params_in_es_core()  #<--Do it here instead
outParams = Inter.get_params()
```

Now local python checks are passing, also for other actors like lb and magnetostatic.
After pull, other actors should be tested as well: create actor without adding it to the system then integrate(0) should do it.
Maybe @KaiSzuttor for LB and @RudolfWeeber for magnetostatics?

testsuite/python/tests_common.py
```diff
@@ -81,6 +81,7 @@ def func(self):
         # set Parameter
         Inter = interClass(**params)
         Inter.validate_params()
+        Inter._set_params_in_es_core()
```
mkuron: Don't do that. Methods prefixed with `_` are private and should not be called from outside the class.

RudolfWeeber: On Wed, Sep 07, 2016 at 11:42:15AM -0700, Michael Kuron wrote:
> > @@ -81,6 +81,7 @@ def func(self):
> >          # set Parameter
> >          Inter = interClass(**params)
> >          Inter.validate_params()
> > +        Inter._set_params_in_es_core()
> 
> Don't do that. Methods prefixed with `_` are private and should not be called from outside the class.
Actually, by now, it should be possible to actiave and dactiave
actors by adding and remong them from system.actors.
See the coulomb_cloud_wall.py testcase.
At the point in time when I wrote the decorators for the actor
tests, that was not possible, so original test did not cover actual
activation, just interface consistency.

mkuron:
Is that last commit related to #831? You can try running the test 100 times. If that doesn't provoke the test to fail, then it is probably fixed.
`ctest -R mass-and-rinertia_per_particle -V --repeat-until-fail 100`.

KonradBreitsprecher:
That one commit simply added 200 integration steps in the testcase 72: mass-and-rinertia_per_particle, I quickly adopted it to get the green light.

mkuron:
Make sure it actually fixes it and doesn't just make it a bit more unlikely to fail. Run it 100 times with that `ctest` command.

fweik:
Can you please check if the issue remains with #835

KonradBreitsprecher:
The failing mass-and-inertia tests are not directly related to this pull request but a general problem.
And #835 doesn't solve it I think.

This one can be pulled.

PR #820, opened 2016-08-30
make hdf5 optional

lahnerml:
Enable switching off HDF5


rbardak:
Might conflict with #745. Otherwise it's alright, however why provide an option?

lahnerml:
Providing an option allows disabling HDF5 when it is neither installed nor needed. :-)

mkuron:
If HDF5 isn't installed, CMake should proceed without it. It's not a required library. So your change will only be relevant if HDF5 is installed but should not be used?

rbardak:
It's not a requires package, so providing an option if it's not installed
is not necessary. I didn't think it would matter in terms of compilation
time if it's not needed. It might just as well be a hassle to deal with the
option at this point. However I might be wrong, considering that #745 might
significantly increase compilation time.

Could you also update the user guide? The option needs to be mentioned
somewhere.

lahnerml:
Well, I want to have this option on the machines of our department, because not all hdf5 libraries that are used in Espresso are installed. In this case the build fails.
I will have to maintain it on my own, if you decide against including this into the official version of espresso. However, I do not see where this change might hurt you as long as HDF5 remains an optional dependency?

rbardak:
I didn't know this case exists. Considering that there is a possibility for
the build to fail, this option is necessary.

mkuron:
No, the correct way to handle this case is to detect on the CMake level whether all HDF5 libraries we require are present and only if that is the case,`HDF5_FOUND` should be set.

RudolfWeeber:
On Tue, Aug 30, 2016 at 07:52:09AM -0700, Michael Kuron wrote:
> No, the correct way to handle this case is to detect on the CMake level whether all HDF5 libraries we require are present and only if that is the case,`HDF5_FOUND` should be set.
That is certainly true. I still find it usefule to be able to switch
off optional libraries, for instance to get rid of unrelated
valgrind messages. Since the library is optional, the espresso code
can handle the case of it not being linked, anyway.

fweik:
I agree that the detection should work out of the box (this has been an issue before). Bit an option does not hury, esoecially because with the next h5md update that stuff will depend on a submodule...

junghans:
Adding this option was kind of redundant, `-DCMAKE_DISABLE_FIND_PACKAGE_HDF5=ON` would have done the job. That is cmake's (not very well-documented) internal way to disable the detection of optional package inside `find_package()`.

PR #823, opened 2016-08-31
Update Dockerfiles

mkuron:
After @fweik added the Dockerfiles in #822, I connected Docker Hub to my Github account and made sure that automatic building works. As each `RUN` line causes one image layer to be created, I merged all commands into one to reflect accepted Docker best practices (there's a maximum number of layers that you can have in an image, somewhere around 40).

I also duplicated the Ubuntu Dockerfile to create one that contains Python 3.

---
On a related note: Can you set up the auto-builds on [Docker Hub](https://hub.docker.com)? Go to Create, Create Automated Build, link your Github account, and select the espressomd/espresso repository. Name the repository _espresso-buildenv-debian_ and click Create. In Build Settings, set the Dockerfile Location to _/maintainer/docker/debian/Dockerfile_. Repeat the same for Ubuntu and for the Python3 dockerfiles.

Once you have done that, please update _.travis-ci.yml_ so that it builds on all three Docker images instead of the one we are currently using.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,14 +3,12 @@ sudo: required
 
 services: docker
 
-before_install:
-        - docker pull espressomd/travis-build:latest
-
-
 env:
   matrix:
     - myconfig=default
     - myconfig=maxset
+    - myconfig=maxset image=ubuntu-python3
+    - myconfig=maxset image=debian
     - myconfig=maxset-gaussrandom
     - myconfig=maxset-gaussrandomcut
     - myconfig=molcut
--- a/maintainer/docker/debian/Dockerfile
+++ b/maintainer/docker/debian/Dockerfile
@@ -1,21 +1,23 @@
 FROM debian:jessie
 MAINTAINER Florian Weik <fweik@icp.uni-stuttgart.de>
 ENV DEBIAN_FRONTEND noninteractive
-RUN apt-get update
-RUN apt-get install -y apt-utils
-RUN apt-get install -y cmake 
-RUN apt-get install -y build-essential
-RUN apt-get install -y clang
-RUN apt-get install -y openmpi-bin
-RUN apt-get install -y libfftw3-dev
-RUN apt-get install -y libboost-dev libboost-serialization-dev libboost-mpi-dev libboost-filesystem-dev libboost-test-dev
-RUN apt-get install -y python python-numpy python-pyvtk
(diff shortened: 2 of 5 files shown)


RudolfWeeber:
On Wed, Aug 31, 2016 at 01:38:41AM -0700, Michael Kuron wrote:
> I also duplicated the Ubuntu Dockerfile to create one that contains Python 3.
[...] 
> Once you have done that, please update _.travis-ci.yml_ so that it builds on all three Docker images instead of the one we are currently using.
> You can view, comment on, or merge this pull request online at:
CI testing for both python versions is a really good thing.

What does this do to the runtime of the ci testing? Isn't it
multiplied by a factor of 3?

Couldn't python2 and python3 be tested from a single image
contiangin both, python2 andf python3 packages and than switching
the python version via cmake?

mkuron:
Yes, that would multiply the runtime by a factor of three. But there is no way around it. After switching to a different Python version, CMake will rebuild everything, so it doesn't make a difference whether we switch to a different Docker image or we delete CMakeCache.txt, re-run cmake with `-DPYTHON_EXECUTABLE=$(which python3)`, and re-build.
I think it would be sufficient to build one configuration (e.g. maxset) with Python 3 and do everything else with Python 2 for now.

mkuron:
I've modified the Travis config appropriately. Of course, now the Travis build fails because the images are not yet on Docker Hub. But once @fweik creates them, the build should succeed again.

RudolfWeeber:
> Yes, that would multiply the runtime by a factor of three. But there is no way around it. After switching to a different Python version, CMake will rebuild everything, so it doesn't make a difference whether we switch to a different Docker image or we delete CMakeCache.txt, re-run cmake with `-DPYTHON_EXECUTABLE=$(which python3)`, and re-build.
True. Pulling the docker image is probably insignifficant compared to
the build time.

On a different note:
What's the reason for testing on Debian? Old enough boost libs?

mkuron:
Yes, we should diversify the testing platforms a bit to catch library and compiler incompatibilities before users report them. I would also like to add CentOS 6 and 7 and OpenSUSE 13.2 and 42.1 so we can cover all distributions that have major marketshare.

fweik:
I'm not sure if we need two compilers. Something like -pedandic should do the trick. If we want to test more I'd rather use clang and do sanitized builds and tests, I think that catches more stuff...

PR #827, opened 2016-08-31
Fix tests with Python 3

### LINKED ISSUE #826: Some python tests fail with python 3
https://travis-ci.org/espressomd/espresso/jobs/156591475

mkuron:
These changes were present in 5b0909bcbcbd8cf5c4f7ca8ea0b30238c167287e, but got lost in 6daf4e85da43c3891ec25f2e711799b975d23ae0

Closes #826

### DIFF
--- a/src/core/interaction_data.cpp
+++ b/src/core/interaction_data.cpp
@@ -870,6 +870,8 @@ void make_bond_type_exist(int type)
 	 bonded_ia_params[type].p.tab.npoints > 0 ) {
       free(bonded_ia_params[type].p.tab.f);
       free(bonded_ia_params[type].p.tab.e);
+      bonded_ia_params[type].p.tab.f = NULL;
+      bonded_ia_params[type].p.tab.e = NULL;
     }
 #endif 
 #ifdef OVERLAPPED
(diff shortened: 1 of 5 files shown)


mkuron:
Now it's a different error, but that one also comes from merging @RudolfWeeber's code:
```
  File "bondedInteractions.py", line 82, in func
    **params).type_name() + ": value set and value gotten back differ for bond id " + str(bondId) + ": " + params.__str__() + " vs. " + outParams.__str__())
AssertionError: False is not true : TABULATED: value set and value gotten back differ for bond id 0: {'type': 'distance', 'filename': 'lj1.tab'} vs. {'maxval': 1.12246, 'minval': 0.0, 'invstepsize': 890.9003438875327, 'type': 'distance', 'npoints': 1001, 'filename': b'lj1.tab'}
```
To me this looks ok. SetParams added a couple extra attributes to the dict, but the ones set are still there.

mkuron:
I think I've finally fixed it. There were some reads of uninitialized memory that tripped Python 3's conversion of C-strings to Unicode strings. While fixing those, I noticed that `make_bond_type_exist` couldn't safely be re-run, which I've also fixed now.

mkuron:
Yup, this is it. Please merge, @fweik.

fweik:
Thank you for fixing this, I messed this up. Good that we have python 3 checks now.

PR #828, opened 2016-08-31
Gpu test fix proposal

psci2195:
At least on Tesla C2070, there is not enough accuracy to achieve the requirement ```$dE <=1E-3```:

```Energy difference to ref. value 0.007545170147636782. Is 148.93475031985236, should be 148.94229549
Difference in energy too large.```

### CODE REVIEW

**Comment by fweik** [comment]
If it is still the Ewald test case that is failing I think something is actually broken. It should be able to get this precision.

**Comment by mkuron** [comment]
I remember the Ewald test used to fail on compute capability 1.x GPUs back when we supported them. The difference is in the fifth digit, so if it's summing up a few thousand numbers, it is already close to the accuracy of single precision. And there the implementation of the arithmetic units could make enough of a difference to pass or fail the test.

**Comment by psci2195** [comment]
Here, we use Tesla C2070 with capability 2.0, hence, the weakest one from that which are supported by Espresso. Maybe, ```Energy difference to ref. value 0.008``` corresponds to such compute capability? Does anyone use GPU with a compute capability 2.0 and passing test ```coulomb_cloud_wall.tcl```?

As other possible solution: we could introduce a compute capability as a parameter of this test? However, such approach will generally make all GPU tests here more complicated.

We could suggest to modify a threshold (as suggested in this changes) starting from the minimal compute capability supported by Espresso (2.0).

**Comment by psci2195** [comment]
Wow, I've merged latest 24 hours changes from the Espresso master and accuracy is increased.

Hence, these changes are no more needed for the 2.0 compute capability.

**Comment by mkuron** [comment]
Could you try running the test a few more times? Summing in parallel is not deterministic, so it's possible the threshold is violated occasionally.

**Comment by psci2195** [comment]
@mkuron , thanks, indeed, it is quite stably failing on the compute capability 2.0:

```
---------------------------------------------------------------
- Testcase coulomb_cloud_wall.tcl running on 01 nodes:
---------------------------------------------------------------
...
Testing EWALD-GPU...
{coulomb 1.0 ewaldgpu 2.70746 64 64 64 1.001 0.0 }
  rms_force_error=3.737679e-04
Energy difference to ref. value 0.007545170147636782. Is 148.93475031985236, should be 148.94229549
Difference in energy too large.
```

I understand that my device is a weakest one supported by Espresso.. Hence, it is up to Espresso team to decide which threshold is needed here..

**Comment by mkuron** [comment]
It's not related to compute capability. I tested on an M2050 (which has the same chip as yours) and on a GTX 780 Ti (which has compute capability 3.5) and always got
```
Energy difference to ref. value 0.007545170148659963. Is 148.93475031985133, should be 148.94229549
```
I get consistently that exact value, so the test or the code must be broken.

psci2195:
I agree. Either ```EWALD-GPU``` code is broken or test parameters are wrong (``` inter coulomb 1.0 ewaldgpu 2.70746 64 1.001 ```). And it does not relate to compute capability or threshold tuning.

RudolfWeeber:
In case it's really about numerical accuracy, Kahan summation may be an option.
But I did not look into the code.

mkuron:
It reproducibly fails right now, and I'm pretty sure it reproducibly worked back when Jenkins was still working and ran that test case every day. Must be a regression, though the Ewald code hasn't been touched in months.

psci2195:
Anyway, this pull request is not needed for the moment. Issue should be additionally investigated and fixed. I've opened the issue https://github.com/espressomd/espresso/issues/830 for further track.

PR #832, opened 2016-09-08
Anisotropy feature for a translational friction/noise

psci2195:
This is an anisotropy feature for a translational friction/fluctuation: see http://dx.doi.org/10.1080/02786828708959128 for a theoretical implication as an example. This feature was announced in the Espresso UG (thermostat section): `In the near future an anisotropic diffusion coefficient  tensor  will  be  made  available  to  simulate  anisotropic  colloids  (rods,  etc.) properly.`. Here, rotational (`ROTATIONAL_INERTIA`) and translational (`PARTICLE_ANISOTROPY`) anisotropies are developed for this purpose as fully independent. Most part of source code changes are quite straightforward: docs, core, tcl, python, tests, maintainer configurations. A conceptual new feature is just in the thermostat.hpp. In case of PARTICLE_ANISOTROPY (double checked via the `aniso_flag` to save some performance), one applies a friction and noise only in the body-fixed reference frame, where `gamma` tensor is diagonal. Then, we returning back to the lab reference frame. Generally, these are the same approach as it is already made for a rotation (torques conversion).

Additional feature is an ability to assign a vector global variable by a single scalar: the same value to all of the components. This possibility is enabled by the Datafield flag `scalar_default_allowed`.

Feel free to ask any question concerning the proposed feature.


src/core/thermostat.cpp
```diff
@@ -209,8 +247,17 @@ void thermo_heat_up()
 {
   int j;
   if(thermo_switch & THERMO_LANGEVIN) {
+#ifndef PARTICLE_ANISOTROPY
     langevin_pref2_buffer          = langevin_pref2;
     langevin_pref2 *= sqrt(3);
```
rempferg: Do you understand where this sqrt(3) comes from?

psci2195: no, please clarify..

rbardak: I believe it's related to the variance of the random number generator. How and why I don't know.

rbardak: I just ran a parameter sweep over kT and gamma... the diffusion constant is wrong by a factor of sqrt(3)...

arnolda: Just for the records: the Espresso variant of Espresso reuses forces in the Verlet algorithm - usually. Thus, you add two times the same randomly drawn number, and your variance is in fact 4*s, where s is the desired variance. This compensated for in the Langevin prefactor, if you ever wonder where the 24.0 in the prefactor comes from.
However, in the first step of the integration loop, Espresso may need to recompute forces (recalc_forces, e.g. because a particle was moved). In that case, we get two different random numbers of variances s' and s. To get the same variance 4s of the sum, you need to choose s' = 3s, or a prefactor of sqrt(3) to the actual random number.

psci2195:
Actually, most important part of these changes is already merged (https://github.com/espressomd/espresso/pull/937) to ```python``` branch. This branch now is a default one. I'm not sure which planning exists concerning the ```master``` branch? Depending on this we can conclude whether we still need the present pull request..

fweik:
New feature development should go into the python branch. I think we can do without porting this back. Thank you for your contribution.

PR #836, opened 2016-09-13
Revert initialization order in P3M Interface

KonradBreitsprecher:
Gone back to the old order because P3M wasn't tuning anymore with this one.
What doesn't work with this is sth like this:
```
p3m.validate_params()
p3m._set_params_in_es_core()
outParams = p3m.get_params() 
#this failes because python_p3m_set_params resets
#parameters in the core if l_bjerrum has not been set 
```
We also need a python testcase that involves tuning.

### DIFF
--- a/src/python/espressomd/electrostatics.pyx
+++ b/src/python/espressomd/electrostatics.pyx
@@ -215,8 +215,6 @@ IF P3M == 1:
             return params
 
         def _set_params_in_es_core(self):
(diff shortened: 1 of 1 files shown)


KonradBreitsprecher:
In non-periodic systems, it throws a seg fault before the sanity checks catch this illegal option for p3m.
The more I have to dig into the electrostatics initialization code, the more I think we need to refactor at some point.

Issue #831, opened 2016-09-07
Intermittent test failure of mass-and-rinertia_per_particle.tcl


mkuron:
`background_errors { ERROR: force on particle 174 was NAN.} { ERROR: force on particle 174 was NAN.} { ERROR: force on particle 174 was NAN.} 
    while executing
"integrate 100"
    (procedure "test_mass-and-rinertia_per_particle" line 132)
    invoked from within
"test_mass-and-rinertia_per_particle $i"
    ("for" body line 2)
    invoked from within
"for {set i 0} {$i < 4} {incr i} {
    test_mass-and-rinertia_per_particle $i
}"
    (file "mass-and-rinertia_per_particle.tcl" line 210)`

This test has failed in eddac1cfb21f3aca7ca6259afbf84519eec6829c for the first time and again in #817. Not sure what's causing it, but in both cases it only occured in one of the Travis jobs (once with the Debian image and once with Python3).

KonradBreitsprecher:
Done some testing, and it fails quite often. I think it spans a too large parameter range and has a fixed timestep for variing masses and intertias:

-Random masses are [0:20[, which cause _Force was NaN_ for small values.
-Rotational inertia components are also [0:20[, the testcase fails in cases like 20 0.1 20

A testing range of [0.5:2[ is much more stable and reduces the failures to an acceptable amount
(1 occurred out of >200 tests). Do we lose possible error coverage with this approach?

Alternatively, we could calculate the appropriate timestep and number of samples for a given set of parameters to reduce the failure rate below a certain threshold.

Same applies for _mass-and-rinertia.tcl_

mkuron:
@RudolfWeeber committed a few changes to that test case a couple weeks back to make it more robust, though it seems it has been failing more often since then. Perhaps he can chime in on this discussion.

fweik:
I think the more important question is why it is not deterministic. Looks
like undefined behavior to me...

KonradBreitsprecher:
The mass-and-rinertia_per_particle.tcl uses the `[pid] + Node_index` for setting the seed (mass-and-rinertia.tcl not). So I could reproduce failing tests..

fweik:
It shouldn't, I must have missed that one.

KonradBreitsprecher:
[...]

PR #839, opened 2016-09-18
Intermittent test stabilization

psci2195:
https://github.com/espressomd/espresso/issues/831


fweik:
I take it that this is the preferred solution to this?

PR #841, opened 2016-09-19
Py: Front-end to constraints based on script interface

RudolfWeeber:
* Adds ScriptInterfaceHelper-class wich allows to define
  py objects which will instanciate as a given base class of PScriptInterface
  class Wall(ScriptInterfaceHelper):
     _so_name="Shapes::Wall"

* Adds Constraints class with add and remove methods

* Added py Constraint base class
* Added py Shape classes
* makes naming "params" consistent with rest of py interface

Usage example:
from espresomd.shapes import Wall
c=system.constraints.add(
    shape=Wall(normal=[1,1,1],distance=1),particle_type=1)
print c.get_params()

This is still a preliminary solution.

### DIFF
--- a/src/python/espressomd/constraints.py
+++ b/src/python/espressomd/constraints.py
@@ -1,0 +1,25 @@
+from __future__ import print_function, absolute_import
+from .script_interface import ScriptInterfaceHelper
+
+
+class Constraints(ScriptInterfaceHelper):
+    _so_name = "Constraints::Constraints"
+
+    def add(self, *args, **kwargs):
+        if len(args) == 1:
+            if isinstance(args[0], Constraint):
+                constraint = args[0]
+            else:
+                raise TypeError(
+                    "Either a Constraint object or key-value pairs for the parameters of a Constraint object need to be passed.")
+        else:
+            constraint = Constraint(**kwargs)
+        self.call_method("add", constraint=constraint)
+        return constraint
+
+    def remove(self, constraint):
+        self.call_method("remove", constraint=constraint)
+
+
+class Constraint(ScriptInterfaceHelper):
+    _so_name = "Constraints::Constraint"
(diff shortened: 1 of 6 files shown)


RudolfWeeber:
Full usage example:


import espressomd
from espressomd.shapes import *


system=espressomd.System() 
box_l=10
system.box_l = [box_l, box_l, box_l]
system.periodicity = [1, 1, 1]
time_step = 0.01
system.time_step = 0.01
system.cell_system.skin = 0.4

c=system.constraints.add(shape=Wall(dist=1.0,normal=[0,0,1]),particle_type=1)

print c.get_params()
#s=c.get_params()["shape"]


system.part.add(id=0, type=0, pos=[0,0,2])

WCA_cut = 2.**(1. / 6.)
system.non_bonded_inter[0,1].lennard_jones.set_params(epsilon=1, sigma=1, cutoff=WCA_cut, 
shift="auto")
system.integrator.run(1)

print system.part[0].pos
print system.part[0].f

KonradBreitsprecher:
Could not test it right away, but the usage seems to be straightforward like that.
Only thing maybe the name _particle_type_ for the interaction type. I understand the reasoning, can't find sth better at the moment, but I don't like it.

RudolfWeeber:
> Could not test it right away, but the usage seems to be straightforward like that.
> Only thing maybe the name _particle_type_ for the interaction type. I understand the reasoning, can't find sth better at the moment, but I don't like it.
Constraints do mimic a particle type, don't they.
In the usage example, one sets up an interaction between particle types 0 and 1,
where 0 is the actual particle and 1 is the particle type used by the constraint.

Any suggestions?

If there is no urgent protests, I'd like to merge this preliminary solution, so we can work on the tutorials and lb boundaries tomorrow.

fweik:
That about interaction_type?

RudolfWeeber:
On Mon, Sep 19, 2016 at 12:49:47PM -0700, Florian Weik wrote:
> That about interaction_type?
I find that more confusing than particle_type, because the number
put there is later used like a particle_type.
I thought, particle_type was the obvious choice.
If this is not the case, we can just return to the "type" which was
there.

fweik:
I see your point. On the other hand it is only related to interactions, and not to particles... I'm fine with particle_type, let's stick to that for now.

PR #843, opened 2016-09-20
ScriptInterface: Abstraction for container-like objects

RudolfWeeber:
Introduces tempalte class ScriptObjectRegsitry
which abstracts the ScriptInterface-containers for
* constraints
* observables
* correlators
* lb boundaries
* ek boundaires

see script_interface/constraints/Constraints.hpp for an example


fweik:
Merged into the python branch

PR #847, opened 2016-09-20
Runtime errors

richter-t:
python error handling should now always display the runtime errors from the core before throwing an exception on the python level.


RudolfWeeber:
I think, this can be merged.

PR #851, opened 2016-09-21
Remove boost::mpi dependency in CUDA code

mkuron:
@fweik @RudolfWeeber @trau @rempferg 

_constraints.hpp_ included, by means of a very long dependency chain, _boost/mpi/communicator.hpp_.

### DIFF
--- a/src/core/EspressoSystemInterface_cuda.cu
+++ b/src/core/EspressoSystemInterface_cuda.cu
@@ -21,6 +21,10 @@
 #include "cuda_interface.hpp"
 #include "cuda_utils.hpp"
 
+#if defined(OMPI_MPI_H) || defined(_MPI_H)
+#error CU-file includes mpi.h! This should not happen!
+#endif
+
 // These functions will split the paritlce data structure into individual arrays for each property
 
 // Position and charge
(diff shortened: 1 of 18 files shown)


mkuron:
The underlying error was this one: https://svn.boost.org/trac/boost/ticket/10998, which is due to insufficiencies in @nvidia's template parser.

fweik:
Can you tell me what the actual include path was?

mkuron:
I can't because I didn't save it. However, `gcc -H` can tell you.

mkuron:
Here is the include chain:
src/core/electrokinetics_cuda.cu
. src/core/constraints.hpp
.. src/core/energy.hpp
... src/core/statistics.hpp
.... src/core/grid.hpp
..... src/core/communication.hpp
...... /usr/include/boost/mpi/communicator.hpp

fweik:
I'm not sure whether we need all those MPI warnings. Also NVCC and MPI should work fine since some time.

mkuron:
We already had them in a couple CUDA files, I just added them to the ones that didn't have them. And actually, CUDA and MPI don't get along on all systems if you use the Autotools build system instead of CMake -- we don't pass the MPI include path to nvcc (CMake's `FindMPI` internally does `mpicc -show` to get them and those do get passed on to nvcc).

PR #853, opened 2016-09-23
Travis: Build+test on Mac OS X and build CUDA code

mkuron:
This pull request adds the following build environments to Travis:
- Mac OS X 10.11 with LLVM 3.8 and Python 2.7
- Mac OS X 10.11 with LLVM 3.8 and Python 3
- Ubuntu 14.04 with GCC 4.8 and CUDA 7.5

@fweik will need to add ` maintainer/docker/ubuntu-cuda/Dockerfile` to the Espresso Docker Hub as `buildenv-espresso-ubuntu-cuda` before the CUDA build will work.

We obviously can't test CUDA code on Travis, but compiling it should help spot the biggest problems already. #851 would have never occurred if we had automatically built the CUDA code.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,60 @@
 language: generic
-sudo: required
 
-services: docker
+matrix:
+  include:
+    - os: linux
+      sudo: required
+      services: docker
+      env: myconfig=default
+    - os: linux
+      sudo: required
+      services: docker
+      env: myconfig=maxset
+    - os: linux
+      sudo: required
+      services: docker
+      env: myconfig=maxset image=ubuntu-python3
+    - os: linux
+      sudo: required
+      services: docker
+      env: myconfig=maxset image=debian
+    - os: osx
+      osx_image: xcode8
+      env: myconfig=maxset
+      cache:
+        directories:
+        - /usr/local/Cellar/boost
+    - os: osx
+      osx_image: xcode8
+      env: myconfig=maxset image=python3
+      cache:
+        directories:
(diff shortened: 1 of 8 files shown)


mkuron:
I have successfully run the CUDA build (https://travis-ci.org/espressomd/espresso/jobs/162200071 -- only difference to the current commit is that it used the image from my Docker Hub), so I think this pull request is in mergeable state now.

fweik:
The MacOS stuff seems to exeed the travis time limit

mkuron:
Hmm, it worked for me before. The trouble is that Boost is not available pre-built with MPI support, so it takes about 30 minutes to build. The build is then cached, but the remaining 20 minutes are not enough to complete the Espresso build.

How about this solution: if Boost is built from source, I abort the Espresso build and fail the Travis job. The newly-built Boost is still cached by Travis and will be used in subsequent builds. The only downside is that about once in six months (with every new Boost version), a build fails and needs to be manually retried.

mkuron:
I've fixed that, but now it appears as if the binary version of gcc (needed as a dependency of openmpi) is broken. Will retry in a few days and let you know if everything is ok to merge.

mkuron:
Ready to merge! The Mac builds have successfully completed after Travis fixed their Mac OS X workers last night (https://www.traviscistatus.com/incidents/b4xk2bybjc9d). After you merge, please create the CUDA docker image on Docker Hub. Once that is built, please retry the Travis build so that the CUDA job also runs.

PR #860, opened 2016-10-01
Rotational Predictor–Corrector method fix

psci2195:
Please correct me if I'm wrong. However, it looks like a very simple correction. According to the original [method](http://ciks.cbt.nist.gov/~garbocz/dpd1/node3.html), iterations (equations (13), (14), etc) should include a constant zero-th estimate of `omega(t+time_step)` defined once by the equation (12). Without proposed changes, both the zero-th estimate of `omega(t+time_step)` and  the velocity-dependent term `g_alpha` take this value from a previous step at each iteration. However, only `g_alpha` is a corrector, cause the zero-th estimate of `omega(t+time_step)` is estimated based on a velocity-independent right part of Euler equations (3) at the `time=t+time_step` as described in the method.


rbardak:
That fix makes sense, but I'm wondering why that was not detected by testcases. It looks like an error that should have noticeable consequences on the physics of the system, since Wd was added multiple times to omega.

psci2195:
Maybe I'm wrong, however, it seems, that there are no tests which really validate separate rotational degrees of freedom in case of an anisotropic particle. After all, this code fragment was equal to zero for a long time (I[0]-I[1] using fixed constants). And this is normal because there are not so much closed-form solution of these nonlinear equations. Hence, there are not so much conditions to compare with.

Robin, I agree that we need here a test which will resolve an existing _false positive_ situation before this specific merge. New test should be failing without the proposed here changes and vice versa.

I could try to create ```mass-and-rinertia*``` like tests per a degree of freedom. Probably, it should be a part of existing summing in order to avoid an overall test phase performance degradation.

rbardak:
Correct me if I am wrong, but one could test the robustness of this part of the code by spinning a particle around its unstable axis of rotation. If the iteration is accurate, the particle would keep its unstable rotation. If the algorithm would somehow add a bunch of terms to omega (which it does in the current master) then the particle would switch its rotational axis to a more stable one. Of course this would not test thermalization, but only the integration method of the EOM. 

@psci2195 Are you working on a test?

psci2195:
@rbardak, yes, I will complete a test implementation very soon.

psci2195:
Hi @rbardak 

I've implemented test ```rotational_inertia.tcl``` which validates an inertial motion around stable and unstable axes: 

1. An approximate constancy of the corresponding body reference frame angular velocity around stable axes
2. A preservation of the total angular momentum in the lab reference frame for such rotations (stable+unstable cases)

These tests fail (in terms of a tolerance: a difference is significant) on master branch and passing here. Hence, probably, the false positive issue discussed in previous comments is resolved now.

Concerning the unstable axis (your suggestion), maybe I am wrong, however checking of a stability of the unstable rotational motion will not help in this specific pull request. Because, in this case, we will have zero values in unlinear terms which we are trying to calculate. Hence, such test will be passed regardless of the given changes. However, I've got your general idea and prepared the test here. A rotation around unstable axis is interesting for above case №2 (already done). By the way, maybe I'm wrong, but in this case a rotation will not switch to stable axis, c.f. video [here](https://en.wikipedia.org/wiki/Tennis_racket_theorem) (Dzhanibekov effect). So the rotation will be infinitely quickly switched between 2 mirror orientations of the body frame keeping an unstable axis rotation for the most part of the time.

Colleagues, please advise. I'm not sure whether we need additional thermalization tests (kT per a degree of freedom). Our experience with an intermittent behaviour of the ```mass-and-rinertia*``` test means that we will need more steps and more particles for this. The overall test run duration will be perceivably more durable. I propose to not split thermalization tests per degrees of freedom. However, I can do it without commits and report results here to be sure that changes are consistent.

psci2195:
The maxset XCode build is failing. Is it a common problem now or one is related to this specific pull request?

psci2195:
> However, I can do it without commits and report results here to be sure that changes are consistent.

I've just run basic tests. They are passing.

psci2195:
fyi, XCode build failure is probabilistic. Now it is passed.. Is everyone experience this?

mkuron:
Looks like the Xcode build just got stuck. Not sure about the cause, @travis-ci was having trouble with their Mac builders several times over the past week, it might be related to that.

rbardak:
> These tests fail (in terms of a tolerance: a difference is significant) on master branch and passing here. Hence, probably, the false positive issue discussed in previous comments is resolved now.

I agree, your tests cover the issue adressed in this pull request quite nicely and provide the expected results for both the master branch and your fix.

> I'm not sure whether we need additional thermalization tests (kT per a degree of freedom).
I don't see the point of additional tests for thermalization, since a thermostat just needs to provide the necessary momenta. Nevermind the test duration, do you think the additional coverage would be beneficial to catch problems/regressions that other tests wouldn't notice?

psci2195:
> Nevermind the test duration, do you think the additional coverage would be beneficial to catch problems/regressions that other tests wouldn't notice?

Yes, I think that ```kT/2``` per each degree of freedom separately is physically very important taking into account the mutual influence of rotations around different axes that we, probably, have properly introduced through this pull request. Even a number of rotational predictor-corrector iterations (currently hardcoded as 5) required for a target precision and an overall numerical stability is of interest here. Currently, it is validated implicitly (cumulatively) as a rotational kinetic energy ```omega_x^2+ omega_y^2+omega_z^2```. Hence, we could keep existing summations in the ```mass-and-rinertia*``` and just to add additional ```kT/2``` validations separately for ```omega_x^2```, ```omega_y^2```, ```omega_z^2```. _Maybe_ it will produce some intermittent test behaviour (e.g. later on, after some new features or changes in the random number generation) and a need of 3 times more summation to compensate it.

rbardak:
I see, since the axis influence each other, checking for kT/2 for every axis seperately is necessary to provide ample coverage.

> we could keep existing summations in the mass-and-rinertia* and just to add additional kT/2 validations separately

I agree, we should keep the number of new tests to a minimum. Regarding the possibility of intermittent test behaviour: mass-and-rinertia* already takes up most of the time checking the maxset (~200s on travis). In this case, compensating for the slight possibility of intermittent test behaviour might be a bit excessive.

psci2195:
I will try to compensate this by rinertia ranges if one will be needed.
Hence, without extra summation.

On Oct 24, 2016 1:58 AM, "Robin Bardakcioglu" <notifications@github.com>
wrote:

I see, since the axis influence each other, checking for kT/2 for every
axis seperately is necessary to provide ample coverage.

we could keep existing summations in the mass-and-rinertia* and just to add
additional kT/2 validations separately

I agree, we should keep the number of new tests to a minimum. Regarding the
possibility of intermittent test behaviour: mass-and-rinertia* already
takes up most of the time checking the maxset (~200s on travis). In this
case, compensating for the slight possibility of intermittent test
behaviour might be a bit excessive.

—
You are receiving this because you were mentioned.
<https://github.com/espressomd/espresso/pull/860#issuecomment-255620833>,
or mute the thread

psci2195:
Hi Robin

> checking for kT/2 for every axis seperately is necessary to provide ample coverage

This is done, now thermalization test is also added per degrees of freedom and we have full bunch of required tests here (+new EOM test above). See changes in the ```mass-and-rinertia_per_particle.tcl```. No additional summing is added there (duration is the same). At the same time, no intermittent behaviour is detected with proposed test parameters changes: 300 runs of tests were passed.

Concerning Travis failure: it looks like pure Travis problem ```Unable to find image 'espressomd/buildenv-espresso-ubuntu:latest' locally``` in one of builds. Could you please restart the build in this pull request?

psci2195:
Done, now only XCode is failing which seems to be a common infrastructure problem now.

rbardak:
Alright, I believe this can be merged now. As @mkuron also stated, the failure of the XCode build can be ignored. 
Thank you for your continued contributions, Bogdan!

fweik:
Thank you for your contribution and for your work during the review. I'm sorry this took so long, but we were rather busy lately.

PR #875, opened 2016-10-13
Make the User guide compile again

hmenke:
**tl;dr** Don't use underscores in `\label`.

The user guide used to not compile on the first try with an error message like this
````
! Missing \endcsname inserted.
<to be read again> 
                   \unhbox 
l.23 ... on Ubuntu 16.04 LTS}{subsection.1.5.1}{}}
````
This is due to the fact that someone used an underscore in a `\label`.  The user guide does `\usepackage[strings]{underscore}`.  This package will make the underscore an active character, such that you can conveniently type text containing underscores without having to escape them using `\_` while retaining the meaning as the subscript operator in math mode.  Unfortunately, the definition of the macro behind the active underscore is not e-TeX protected which leads to it being expanded once written to an `aux` file (which is what `\label` does).  It then tries to expand assignments somewhere deep down in the expansion chain which is not possible at this point which then leads to the error.

### DIFF
--- a/doc/ug/introduction.tex
+++ b/doc/ug/introduction.tex
@@ -297,7 +297,7 @@ whole \es-script, there should be no problems.
 
 \section{Requirements}
 \label{sec:requirements}
-\index{requirements}
+\index{Installation requirements}
 
 The following libraries and tools are required to be able to compile
 and use \es:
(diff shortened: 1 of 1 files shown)


hmenke:
The only file touched is `introduction.tex`, so the testsuite should not complain.

PR #876, opened 2016-10-13
Sphinx

fweik:


fweik:
I would very much prefer to use only one documentation system. We have doxygen documentation,
and apparently sphinx does not have a doxygen importer.

KaiSzuttor:
Actually I don't think it's a bad idea to have a separate code documentation for the interface. And I also don't see a comfortable solution to use doxygen for documenting the python code.

PR #880, opened 2016-10-14
Fix Python include path detection

### LINKED ISSUE #869: CMake mixes Python 2 and Python 3 if both are installed
If both Python 2 and Python 3 are installed (possibly with one having only the interpreter and the library installed), CMake may end up using a mixture of Python 2 and 3 headers, libraries and executables. This is definitely a bug in the FindPython modules of CMake, but we need to work around it.

mkuron:
FindPythonLibs.cmake appears to be broken, even more so in older CMake versions. There are many situations where it uses libraries and headers that don't match the interpreter used. This pull request uses `distutils.sysconfig` to directly obtain the header path from Python.

We don't need the library path because we are not embedding a Python interpreter into a C++ program, we are only creating a Python module. We did link the library before, but it's not necessary and not even considered good practice.

Closes #869.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -107,13 +107,14 @@ if (WITH_CUDA)
   endif(CUDA_FOUND)
 endif(WITH_CUDA)
 
-find_package(PythonInterp 2)
+find_package(PythonInterp)
 
 if (WITH_PYTHON)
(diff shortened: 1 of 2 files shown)


mkuron:
Build succeeded, ready to merge.

PR #887, opened 2016-10-18
Add --gdb flag to pypresso to launch Python in GDB

mkuron:
@richter-t complained about the difficulty of running Espresso with Python in the debugger. This pull request adds a flag `--gdb` to pypresso to simplify that as discussed with @RudolfWeeber. Obviously, you need to set `CMAKE_BUILD_TYPE=Debug` via ccmake.

### DIFF
--- a/src/python/pypresso.cmakein
+++ b/src/python/pypresso.cmakein
@@ -13,7 +13,11 @@ if test -n "$PYTHONPATH"; then
 else
   PYTHONPATH=@PYTHON_DIR@
 fi
-
-
 export PYTHONPATH
+
+if [ "$1" = "--gdb" ]; then
+  shift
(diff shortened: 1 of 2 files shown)


fweik:
Good idea. I'd rather have it execute start.

mkuron:
I don't think it makes much of a difference for most people (except that they would need to explicitly run `continue` before the program starts). We don't care about what happens inside the Python interpreter before it reaches its `main()` function. Besides, breakpoints in Espresso functions cannot be set before all the Python modules have been loaded (I checked, there is no way to pre-load symbols for `dlopen`ed libraries in GDB), which means you need to run your program, interrupt it, set breakpoints, then rerun it.

Issue #886, opened 2016-10-17
Polymer command in the python interface is broken


richter-t:
Upon creating 2 polymer chains (or more) via the "Polymer" command/class of the python interface the program segfaults.
Tested with the attached mini script on the current state (46c5ec6917bd508f67b13572adcf185155af8a5d) of the python branch. 

[minimal_poly_test.txt](https://github.com/espressomd/espresso/files/533805/minimal_poly_test.txt)

richter-t:
fixed in pull request #920

PR #888, opened 2016-10-18
fixed segfaulting of polymerC command  #886 

richter-t:
see code snippet below. It seems a bad idea to free pointers that were given to a function externally. Could've seen that sooner. #886


fweik:
Can you please remove the comments?

richter-t:
now part of other pull request

PR #890, opened 2016-10-21
Fix travis cmake script

fweik:


src/python/espressomd/script_interface.pyx
```diff
@@ -2,7 +2,7 @@
 cdef class PScriptInterface:
     def __init__(self, name=None):
         if name:
-            self.sip = make_shared(name)
+            self.sip = make_shared(name.encode('UTF-8'))
```
mkuron: Don't do that, we have `espressomd.utils.to_char_pointer` for that purpose.

fweik:
@RudolfWeeber ready for review

RudolfWeeber:
Can be merged

PR #892, opened 2016-10-27
Relative import behaviour changed in Cython 0.25

hmenke:
Homebrew now ships version 0.25 of Cython.  There the behaviour of relative imports has changed which leads to an error when generating C++ code.  This commit merely replaces the offending line.

### DIFF
--- a/src/python/espressomd/reaction.pyx
+++ b/src/python/espressomd/reaction.pyx
@@ -1,7 +1,7 @@
 from __future__ import print_function, absolute_import
 include "myconfig.pxi"
+cimport globals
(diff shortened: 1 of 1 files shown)


mkuron:
This appears to build correctly on Cython 0.23, 0.24 and 0.25 both with Python 2 and Python 3, so I approve of the change.

fweik:
I suppose this should go to the python branch?

hmenke:
@fweik Actually, it should go to both branches because the master will start failing once Travis decides to pull cython-0.25 from homebrew.

PR #895, opened 2016-11-02
Added check for MASS feature in force property.

KaiSzuttor:


fweik:
What is the benefit of this, except for a minimal optimization?

KaiSzuttor:
The particle only has a mass if the feature is compiled in, therefore you get an error if you call the __get__ method of the f-property.

Issue #856, opened 2016-09-23
ScriptInterface: call_method should be able to return a VariantMap


RudolfWeeber:
Some core methods need to return more complex data structures than those supported by the variant used in the script interface.
E.g. prining resutls of correlators

This could be achieved
either by adding VariantMap to the supported parameter types (no idea, if this kind of cyiclical definition is possible)
or by adding a separate
variant_map_call_method()
to ScriptInterfaceBase.

fweik:
The plan is to make Variant recursive (which boost::variant). Together with the base types the can represent arbitrary complex data.

RudolfWeeber:
Resolved by implementation of recursive variants.

Issue #855, opened 2016-09-23
ScriptInterface: Parametrs of type Object aren't returned in python


RudolfWeeber:
Instead, a numeric id is returned

c=Constraint(shape=Wall(distance=1,normal=(1,0,0))
print c.get_params()["shape"]

fweik:
This is a regression, that worked earlier. I'll look into that.

RudolfWeeber:
Can this be closed?

Issue #854, opened 2016-09-23
ScriptInterface: Exception when trying to retrieve not set parameters of type OBJECT


RudolfWeeber:
from espressomd.constraints import Constraint
c=Constraint()
print c.get_params()
Because the shape property has not been set and the logic in Constraint::get_parameter() doesnt catch that.

RudolfWeeber:
Is this resolved?

PR #900, opened 2016-11-08
Recursive variant

### LINKED ISSUE #889: Script interface: VariantMap in call_method
- c++:
  - The method `T& operator[]` of `std::map` is not a const method. Using `parameters["foo"]` in `call_method` will not compile, because the input argument 'parameters' of `call_method` is a `const VariantMap&`.
- python:
  - When setting `parameters[name] = kwargs[name]` in `call_method` of the class `PScriptInterface`, we need to wrap `kwargs[name]` with `self.make_variant` with the appropriate type, otherwise this will result in a bad `boost::get` at runtime. (same way it is done in `set_params`). As far as I understand it, the supported `**kwargs` for call_method should be arbitrary (?) so we can't rely on the list of valid params to derive the type.

fweik:
- Made the variant type in the script interface recursive, so that functions can return structured data
- Corresponding translation functions in python
- some clean up of the ObjectId handling (needs further improvement)
- Cleanup and fix of the variant conversion

Fixes #889, #856, #855, #854,

### DIFF
--- a/src/core/Vector.hpp
+++ b/src/core/Vector.hpp
@@ -48,7 +48,7 @@ public:
   }
 
   template <typename InputIterator>
-  Vector(InputIterator const& begin, InputIterator const& end) {
+  Vector(InputIterator const &begin, InputIterator const &end) {
     assert(std::distance(begin, end) <= n);
     std::copy(begin, end, d.begin());
   }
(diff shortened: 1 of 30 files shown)


RudolfWeeber:
The python code looks good. 
However, just from reading it, it's hard to judge whether it really does what it should.
I'd suggest putting in a testcase in the python testsuite together with a c++ function returning a recursive variant to check that the conversion really happens as intended.

rbardak:
I'll take a look on friday. Since I'll probably be using this feature, I might as well design some test cases.

fweik:
Tests are included, but I just realized that this broke the Vector?d parameters. Will add a fix.

RudolfWeeber:
Looks good. Can be merged

PR #905, opened 2016-11-10
Add custom mapping for particle visualization.

pkreissl:
Usually, particles are visualized according to their Lennard-Jones radius.  This is a slight drawback because this requires a LJ interaction of particles with the same type.  If this is not desirable for physical reasons these particles cannot be visualized.  Therefore we overload the visualizer in such a way that the user can define a custom mapping function from type to particle radius.  If this overload is not used it defaults to the previous behavior.

Example:  Set the radius for all particles (all types) to 0.5
````
visualizer = visualization.mayaviLive(system, lambda x: 0.5)
````

### DIFF
--- a/src/python/espressomd/visualizationMayavi.pyx
+++ b/src/python/espressomd/visualizationMayavi.pyx
@@ -36,6 +36,7 @@ cdef class mayaviLive:
     interact with the GUI."""
 
     cdef object system
+    cdef object mapping
     cdef object points 
     cdef object box 
     cdef object arrows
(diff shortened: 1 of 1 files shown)


mkuron:
Good idea, but there are two things that need to be changed:
- The OpenGL visualizer has the same interface as the Mayavi visualizer and should thus support the same parameters. Therefore, please match the behavior between the two. The OpenGL visualizer takes a parameter `particle_sizes`, which can be set to _auto_ (giving it the same behavior as the Mayavi visualizer) or _type_ (where the radius is set to the type ID). 
- It should be possible to specify either a lambda function that calculates the radii (meaning that `mapping(t)` should be used as radius; use `if callable(mapping)` to detect this case), or a list or dictionary that lists them directly (meaning that `mapping[t]` should be used). For the latter, please make sure that the case of a missing entry is handled by falling back to some sensible default behavior.

PR #906, opened 2016-11-10
Script interface

fweik:


rbardak:
At this point I'd like to point out that the PScriptInterface class needs `__getstate__` and `__setstate__` methods which are called from `pickle.dump` and `pickle.load` respectively. This should ensure that checkpointing actually works with script interface objects, as I can imagine this cannot be handled with default behaviour.

PR #909, opened 2016-11-11
Fix Travis Mac builds after build environment changes

mkuron:
@fweik, this should fix it.

Once this builds, I will cherry-pick to the python branch and create another pull request.


mkuron:
Looks good, please merge.

PR #970, opened 2016-11-30
Rasterize visualization

KonradBreitsprecher:
Description of changes:
- Basic constraints are visualized with basic openGL primitives (Shpere, Wall, Cylinder)
- Other constraints are visualized by rasterization of the volume and drawing 2D points:

image

PR Checklist
------------
 - [x] Tests?
   - [:(] Interface
   - [x] Core 
 - [:(] Docs?
..will come after interface is finalized


KonradBreitsprecher:
Not sure if a pool billard game sample script is appropriate...

image

mkuron:
Ok, that's neat. I see you also fixed all my comments except for the one asking to introduce the `particle_sizes` kwargs argument to match the changes @pkreissl introduced in #905. Once you have done that, I'll approve the pull request.

KonradBreitsprecher:
Ok did it, particle_sizes now can be 
- 'auto' for LJ self-interaction parameter, 
- callable
- list
The list (as all type dependent lists in the visualizer) is indexed with [type % len(list)].
It was easy to do, so no drama, but I can't think of any real functional dependence between type and radius,
I only see: All radii constant, LJ sigmas or hardcoded list...

src/python/espressomd/visualizationOpenGL.pyx
```diff
@@ -327,15 +326,17 @@ class openGLLive:
...
+            if callable(self.specs['particle_sizes']):
+                radius = self.specs['particle_sizes'](ptype)
+            elif self.specs['particle_sizes'] == 'auto':
+                lj_sig = self.system.non_bonded_inter[ptype, ptype].lennard_jones.get_params()['sigma']
                 radius = lj_sig / 2.0
                 if radius == 0:
                     radius = self.sizeByType(ptype)
 
-            elif self.specs['particle_sizes'] == 'type':
+            elif isinstance(self.specs['particle_sizes'], list):
```
mkuron: This only applies to `list`, but you should accept anything that has a subscript operator (e.g. a Numpy array). Please see #905 for how to do that with a try-except. Also, I think the line below should be something like 
```
radius = self.specs['particle_sizes'][ptype]
```

src/python/espressomd/visualizationOpenGL.pyx
```diff
@@ -204,18 +326,18 @@ def drawSystemParticles(self):
...
-                if radius == 0:
-                    radius = self.sizeByType(ptype)
-
-            elif self.specs['particle_sizes'] == 'type':
-                radius = self.sizeByType(ptype)
+            radius = 0.5
+            try:
+                if callable(self.specs['particle_sizes']):
+                    radius = self.specs['particle_sizes'](ptype)
+                elif isinstance(self.specs['particle_sizes'], (list,tuple,np.ndarray)):
```
mkuron: Please don't explicitly check for whether it's an instance of only those classes. That violates Python's duck-typing approach. Just try subscripting it in try-except, like @pkreissl did. Your current try-except doesn't help much.

KaiSzuttor: @mkuron isn't it easier to propose the change directly instead of describing it?

mkuron: Just look at `_determine_radius` in #905.

KonradBreitsprecher: I see your point, but the try..except now prevents 
-corrupted lambda functions
-errors from `self.system.non_bonded_inter[ptype, ptype].lennard_jones.get_params()`
Duck-typing here would mean to check for __iter__ and types of list members?
I think it is pretty hard to catch all (user-induced) irregular parameters. e.g. the mayavi approach fails if the uses types 'autos', which would end up in radius = 'a' for type 0..

mkuron: You could `except TypeError`. If you don't like that, use `hasattr(obj, '__getitem__')` to detect whether an object is subscriptable instead of a try-except. Whichever you choose, please make the same change in the Mayavi visualizer.

mkuron:
Approved.

PR #940, opened 2016-11-24
utils: Enable native serialization for double arrays.

fweik:
Fixes #

Description of changes:
 - 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?

PR #979, opened 2016-12-08
communication: Speed up callback request transmittion to work around …

fweik:
Fixes #

Description of changes:
 - 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?

…some boost.mpi quirks

### DIFF
--- a/src/core/MpiCallbacks.cpp
+++ b/src/core/MpiCallbacks.cpp
@@ -42,7 +42,7 @@ void MpiCallbacks::call(int id, int par1, int par2) const {
   std::array<int, 3> request{id, par1, par2};
 
   /** Send request to slaves */
-  boost::mpi::broadcast(m_comm, request, 0);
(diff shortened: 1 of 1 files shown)


mkuron:
We now have 10 seconds (down from more than 40, see #940 for details on the test), which is still significantly more than the 4 seconds we get with Infiniband disabled.

fweik:
The boost::mpi code that is in now just results in MPI_Bcast(comm, int *, int, int), so the rest of the overhead should be due to the MPI implementation or the infiniband. Still that overhead looks somewhat high to me, isn't infiniband supposed to be low latency?!

mkuron:
I'm not running in parallel, so this overhead comes sloely from registering memory with the Infiniband driver. MPI_Alloc_mem/MPI_Free_mem is still called in some places during startup (and that is a slow operation because it registers memory with Infiniband). Your patch does get rid of all calls during particle creation however, which made up the vast majority of the performance issue.

PR #980, opened 2016-12-19
h5md: changed type dataset to integer dataset

jonaslandsgesell:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -212,7 +212,7 @@ list(APPEND LIBRARIES ${Boost_LIBRARIES})
 
 if(WITH_TESTS)
   enable_testing()
-  find_package(Boost COMPONENTS unit_test_framework timer chrono) 
+  find_package(Boost COMPONENTS unit_test_framework) 
   if(Boost_UNIT_TEST_FRAMEWORK_FOUND)
     set(WITH_UNIT_TESTS ON)
     list(APPEND LIBRARIES ${Boost_LIBRARIES})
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -212,7 +212,7 @@ list(APPEND LIBRARIES ${Boost_LIBRARIES})
 
 if(WITH_TESTS)
   enable_testing()
-  find_package(Boost COMPONENTS unit_test_framework) 
+  find_package(Boost COMPONENTS unit_test_framework timer chrono)
```
fweik: Why this?

jonaslandsgesell: Otherwise Espresso was not compiling on the institute machines.

fweik: I don't think this is true.

jonaslandsgesell:
There is a problem in travis when compiling boost. Seems not to have something to do with the fix for the integer dataset.

mkuron:
Just resubmit the job to travis. It was actually nice enough to tell you:
```=============================================================================
Boost was built from source, but this took so long that we won't have time to
finish the actual build process before the Travis timeout kills it.
Please manually retry the current Travis build!
=============================================================================```

fweik:
We need to find a solution to this.

mkuron:
Yeah, we do... I don't think it happens that often, but it's something that can be fixed by filing a pull request against https://github.com/Homebrew/homebrew-core/blob/master/Formula/boost.rb . I'll have a look at on the next coding day.

jonaslandsgesell:
This is already included in the other h5md pull request. So I close this PR.

PR #982, opened 2016-12-20
Fixing cython issue on second build (import local module io).

KaiSzuttor:


mkuron:
This only helps for out-of-source builds. It breaks in-source builds.

KaiSzuttor:
How should it break in source builds?

mkuron:
Ah sorry, that is only while `cython`ing, not while executing tests or the like. So it should be fine.

PR #985, opened 2016-12-27
H5md serial output

jonaslandsgesell:
Description of changes:
Allows to optionally write h5md files where positions etc. are ordered according to ids. The current implementation however only works in the single process mode since it relies on the existence of the global partCfg array. Still positions/types/charges/masses/images/etc. which have a fixed order remove the need to sort the h5md content and therefore this change eases the analysis of the obtained data.

### DIFF
--- a/src/core/io/writer/h5md/h5md_core.cpp
+++ b/src/core/io/writer/h5md/h5md_core.cpp
@@ -268,23 +268,27 @@ void File::Write(int write_dat)
     bool write_force = write_dat & W_F;
     bool write_mass = write_dat & W_MASS;
 
-    std::cout << "ordered " << m_write_ordered << std::endl;
-
+    int num_particles_to_be_written;
+    int remainder=0;
+    
     if (m_write_ordered==true) {
-    	int npart=max_seen_particle+1;
-    	int num_particles_to_be_written=std::floor(npart/n_nodes);
-    	int remainder=0;
+    	num_particles_to_be_written=std::floor(n_part/n_nodes);
     	if(this_node==n_nodes-1)
-    		remainder=npart%num_particles_to_be_written;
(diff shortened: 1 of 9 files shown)


src/core/io/writer/h5md/h5md_core.cpp
```diff
@@ -268,36 +268,25 @@ void File::Write(int write_dat)
...
-    {
-        local_cell = local_cells.cell[cell_id];
-        for (int local_part_id = 0;
-             local_part_id < local_cell->n; ++local_part_id)
-        {
-            const auto & current_particle = local_cell->part[local_part_id];
-            id[0][particle_index][0] = current_particle.p.identity;
+
+    if (m_write_ordered==true && n_nodes==1) {
+
```
fweik: Please find a way to avoid the code duplication here.

src/core/io/writer/h5md/h5md_core.cpp
```diff
@@ -257,73 +257,90 @@ void File::Close()
...
+        
+        //loop over all particles
+        for(int particle_index=0;particle_index<n_part;particle_index++){
+            Particle current_particle;
+	        get_particle_data(particle_index, &current_particle); //this function only works when run with one process
+            fill_arrays_for_h5md_write_with_particle_property(particle_index, id, typ, mass, pos, image, vel, f, &current_particle,  write_typ, write_mass, write_pos, write_vel, write_force );
+	        free_particle(&current_particle);
+        }
+
+    }else{
```
KaiSzuttor: if m_write_ordered == true && n_nodes != 1 this should throw an exception.

jonaslandsgesell:
Writing H5MD Files like here with a lot of independent writes is very expensive. Do not use this. I open another pull request which writes ordered data only from the master node. This is however much faster.

PR #989, opened 2017-01-11
script_interface: Make sure that parallel wrapper instances are kept …

### LINKED ISSUE #968: get_params() for constraint segfaults..
..only if the shape is created inside of constraints.add.

```
from espressomd import *
from espressomd.shapes import *

system = espressomd.System()

#Works:
shape_wall = Wall(dist=14, normal=[1,1,1])
constr_wall = system.constraints.add(shape=shape_wall,particle_type=1)
print constr_wall.get_params() 

#Nope:
constr_wall2 = system.constraints.add(shape=Wall(dist=14, normal=[1,1,1]),particle_type=1)
print constr_wall2.get_params()

#terminate called after throwing an instance of 'std::out_of_range'
#      what():  _Map_base::at
#               Segmentation fault
```

fweik:
…alive when their payload is still used. Fixes #968.

Fixes #968 

Description of changes:
 - When passing object parameters to the payload, the host instance keeps a shared_ptr to
   the host object of the passed object. This keeps it alive. After the call to the payload returned,
  the ref count of the payload of all stored host classes is check to see if they can be removed.

~~Arguably the should be unit tests for stuff like that, I'll look into that.~~

PR Checklist
------------
 - [x] Tests?
   - ~~Interface~~
   - [x] Core 
 - ~~Docs~~

### DIFF
--- a/src/core/unit_tests/ParallelScriptInterface_test.cpp
+++ b/src/core/unit_tests/ParallelScriptInterface_test.cpp
@@ -37,6 +37,18 @@ struct TestClass : public ScriptInterfaceBase {
 
   void set_parameter(const std::string &name, const Variant &value) override {
     last_parameter = make_pair(name, value);
+
+    if (name == "obj_param") {
+      obj_param = get_instance(boost::get<ObjectId>(value)).lock();
+    }
+  }
+
+  Variant get_parameter(std::string const &name) const override {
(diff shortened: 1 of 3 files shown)


KonradBreitsprecher:
It indeed fixes #968. Can't really comment on the code though.

PR #996, opened 2017-01-23
Three new features.  Number two will blow your mind!

hmenke:
# Observable Recorder

- [x] Sample script (rather than test case)
- [x] Documented

This PR mainly contain the observable recorder.  To verify whether the internal correlator was working properly and to compute correlations with a different algorithm we needed to write out the full trajectory of a particle for every time step.  This was straight-forwardly generalized to all observables and now an observable can be written to a file using
````python
p = espressomd.observables.ComPosition(ids=[0,1,2])
p.auto_write_to(filename="test_observable_writer.dat")
system.auto_update_observables.add(p)
````

# Fixes to the Checkpointing

- [x] Test case (old test case remains working)
- [x] Documented

It was previously not possible for a checkpoint to overwrite existing particles.  To restore the behaviour familiar from the old blockfiles, checkpoints can now overwrite particles.  The high-level interface using `system.part.add()` continues to throw if a particle exists.

The checkpointing module was always creating a new new file for everytime `checkpoint.save()` was called.  This might be disadvantageous if the data to be checkpointed are huge and disk space will run out quickly.  A small change allows to specify a specific checkpoint index also in `save()` (it was already available in `load()`).

The combination of
````python
````
will now almost behave like old blockfiles.

This raises the demand that checkpoint writing is exception safe.  The problem is that if the simulation is terminated mid-writing, you will end up with an unrestorable checkpoint.  Therefore we save the checkpoint to a temporary file and move this to the target file.  The move operation is atomic and even if this fails a restorable but older checkpoint remains.

# LB_GPU Momentum Correction

- [x] Test case
- [x] Documented

This is essentially a copy of #976 but for bare LB_GPU instead of EK (#976 might be unnecessary).

Given an `LBFluid_GPU` called `lbf` we can apply the momentum correction using
````python
lbf.remove_total_momentum()
````

### CODE REVIEW

File: src/core/lbgpu.cpp [hunk 1]
```diff
@@ -440,4 +442,33 @@ int lb_lbfluid_load_checkpoint_wrapper(char* filename, int binary)
...
+  {
+    double mass;
+#ifdef MASS
+    mass = particle_data_host[i].mass;
+#else
+    mass = 1.;
+#endif
+
+    double new_velocity[3] = {
+      particle_data_host[i].v[0] + velocity[0] / mass * time_step,
```
rempferg: why * time_step?

src/core/lbgpu.cpp
```diff
@@ -440,4 +442,33 @@ int lb_lbfluid_load_checkpoint_wrapper(char* filename, int binary)
   return lb_lbfluid_load_checkpoint(filename, binary);
 }
 
+void lb_lbfluid_particles_add_momentum ( float velocity[3] )
```
rempferg: shouldn't this parameter rather be called momentum[3]?

src/core/lbgpu_cuda.cu
```diff
@@ -4071,4 +4075,98 @@ int statistics_observable_lbgpu_velocity_profile(profile_data* pdata, double* A,
...
+  velocity[1] = -total_momentum[1]/(fluid_mass + particles_mass);
+  velocity[2] = -total_momentum[2]/(fluid_mass + particles_mass);
+
+  float momentum_fluid[3] = {
+    velocity[0]*fluid_mass,
+    velocity[1]*fluid_mass,
+    velocity[2]*fluid_mass
+  };
+
+  lb_lbfluid_particles_add_momentum( velocity );
```
rempferg: i would rather have these two be consistent, or rename the particle correction to something with velocity or galilei transform.

src/core/lbgpu.cpp
hmenke: Sure, why not.  In the core there is not really a difference because everything is made dimensionless.

src/core/lbgpu.cpp
hmenke: That’s also in the interface.  From `particle_data.pyx`:
````python
myv[i] *= time_step
````
The parameter `velocity` is the one returned from `calc_linear_momentum` which is already scaled back to interface units, so we have to multiply by time step again.  Trust me, I checked this.

src/core/lbgpu_cuda.cu
hmenke: This is also about naming?  If yes, then it’s fixed.

PR #1002, opened 2017-01-30
potential bugfix for bounce back in CPU lb

lahnerml:
Fixes #

Description of changes:
 - 

PR Checklist
------------
 - [x] Tests?
   - [x] Interface
   - [x] Core 
 - [x] Docs?


rbardak:
Off the top of my head I would say you need agrid^4. Three to transform lbpar.rho from md units to lb units and another one because the formula says so.

PR #1007, opened 2017-02-01
Async ghost comm

hirschsn:
This PR implements asynchronous ghost communication (only GHOST_SEND and GHOST_RECV, not broadcast or reduce since domain_decomposition doesn't make use of the latter two; support for GHOST_LOCL could easily be added).

Also, as an example and to allow testing, the implementation of domain_decomposition is adapted to make use of asynchronous communication.


fweik:
@hirschsn Do I remember correctly that there was a performance impact by this? Can you comment?

hirschsn:
First of all, for others that want to follow the discussion: The asynchronous communication is necessary if we want to support partitionings which are not box shaped.

@fweik Yes, you do remember correctly. For a very small number of processes it is slower simply because a) it does not support GHOST_LOCL and b) it communicates 26 times instead of the synchronous 6 times---in the extreme case of a simulation on a single process 26 times with itself. I did some quick tests on Hazel Hen with a totally homogeneous Lennard-Jones fluid of density 0.7 (particles arranged in a grid) with 1000 particles per process, weak scaling. The following numbers refer to the maximum(!) measured ghost communication time of an integrate_vv call with n_steps=1000.
- 1 process: The asynchronous version is about two times slower. (The communication time, however, is only roughly 10% of the total simulation time in this case.)
- Starting from 4 processes, the difference in both versions gets lower than 10% with the synchronous version still being faster.
- Starting from 2 nodes (24 processes per node) the asynchronous version is marginally faster (up to ~10%).

Use these numbers with caution. There are most likely scenarios and machines where this behaves totally different, e.g. big shared-memory machines probably.

In my fork I also have a branch (async-resort) which implements dd_exchange_and_sort_particles asynchronously. Thus, both changes together (branch async-dd) implement a fully asynchronous domain_decomposition. Note, however, that the asynchronous resorting does not support Lees Edwards boundary conditions.

RudolfWeeber:
> @fweik Yes, you do remember correctly. For a very small number of processes it is slower simply because a) it does not support GHOST_LOCL and b) it communicates 26 times instead of the synchronous 6 times---in the extreme case of a simulation on a single process 26 times with itself.
A single processor is, of course, a very common use case for coarse-grained simulations.

fweik:
If the reason is only the missing GHOST_LOCL, then this isn't actually a problem of the ghost comm. These should not exists in the first place.

hirschsn:
If I recall the synchronous code correctly, this is not easily testable because we cannot simply replace the GHOST_LOCLs in it. What we probably could do to get an idea of the impact of GHOST_LOCL is skipping all local ghost exchanges and only measure communications with other processes.

hirschsn:
I thought about the measurement and came to the conclusion that the measurement method I proposed does not work as it effectively measures nothing for a single process (all transfers are local).

Nevertheless, I measured it for 2 and 4 processes. If the problem was purely the missing GHOST_LOCL, I would expect the absolute runtimes of both versions to the same with the described adaptions. This is not the case. The asynchronous version is still slower.

PR #1009, opened 2017-02-03
Travis: Use boost-mpi from Homebrew

mkuron:
Fixes #991. Please also merge to python branch.

Note that this depends on https://github.com/Homebrew/homebrew-core/pull/9598, so wait until that one is merged.

PR #1010, opened 2017-02-03
Travis: use HDF5 from Homebrew

mkuron:
Please merge after #1009 because that commit re-enables macOS builds.


mkuron:
All tests passed. @fweik, please click the green button.

Issue #978, opened 2016-12-08
Seeding of the PRNG in tcl does not work


jonaslandsgesell:
However a new problem has shown up:
In the tcl interface running the following script fails to set the seed in the core to 3, instead the seed 1 is reported from the core.

```
t_random seed 3
puts [t_random seed]
```
(for the python interface seeding works)

RudolfWeeber:
No longer fixing tcl related issues

Issue #577, opened 2016-03-26
Checkpointing of the RNG is gone


hmenke:
Commit 9ccafbbd9308c1e8732a30af9225078083deb3f4 completely broke the blockfile mechanism of ESPResSo!  `t_random stat` is called from within the [`blockfile_support.tcl`](https://github.com/espressomd/espresso/blob/master/scripts/blockfile_support.tcl#L413) script.

jonaslandsgesell:
The bitrandom generator (and its checkpointing) have already gone before this commit.

Maybe we should think about the necessity to checkpoint the random number generator. For physically relevant observables the seed and the current state of the random number generator before checkpointing should not matter statistically (also since checkpointing is a rare event).
Checkpointing of the RNG is however relevant if there is a need to produce bit-identical simulations where checkpointing occurs. See also http://www.gromacs.org/Documentation/How-tos/Extending_Simulations#Exact_vs_binary_identical_continuation in the exact vs binary section.

However we can use the standard Cpp11 methods to obtain the internal state of the RNG.

hmenke:
I am not referring to `bit_random` but rather to `t_random`.  The relevant change is [this](https://github.com/espressomd/espresso/commit/9ccafbbd9308c1e8732a30af9225078083deb3f4#diff-c61b75d62735b82fb66f0891144e810fL129).

Also, if you remove any features, make sure they are purged completely from all helper scripts.  That is to say, please have a look at [`blockfile_support.tcl`](https://github.com/espressomd/espresso/blob/master/scripts/blockfile_support.tcl).  There are still some references to `bit_random` (and the offending `t_random stat`).

hmenke:
Here is also a simple script to reproduce the issue
````
set f [open "test.blk" "w"]
blockfile $f write random
````
One then receives
````
Unknown argument 'stat' requested!
Usage: 't_random [{ int <n> | seed [<seed(0)> ... <seed(0)>] }]'
````

fweik:
This is just an interface bug. The method you quote only deals with the bitrandom PRNG that was not used and is now removed. The obvious solution is to purge "t_random state" everywhere. If somebody needs checkpointing for the RNG that would not be hard to implement, but I don't think that came up yet.

tigoesnumb3rs:
This also affects the lennard jones tutorial..

fweik:
Ok, I somewhat misread that. Henri is right, we need to add the t_random stat mechanism again.

hmenke:
[...]

PR #1039, opened 2017-02-20
random: Initialize PRNG with random number from system.

fweik:
Due to popular demand and against my better judgement, initialize
the PRNG with std::random_device.

Fixes nothing

Description of changes:
 - Initialize PRNG with true random number from std::random_device.
- clang-format

PR Checklist
------------
 - [ ] Tests? Very hard to test, details are implementation specific.

### DIFF
--- a/src/core/random.hpp
+++ b/src/core/random.hpp
@@ -1,28 +1,28 @@
 /*
   Copyright (C) 2010,2012,2013,2014,2015,2016 The ESPResSo project
 
-  Copyright (C) 2002,2003,2004,2005,2006,2007,2008,2009,2010 
+  Copyright (C) 2002,2003,2004,2005,2006,2007,2008,2009,2010
     Max-Planck-Institute for Polymer Research, Theory Group
-  
+
   This file is part of ESPResSo.
-  
+
   ESPResSo is free software: you can redistribute it and/or modify
   it under the terms of the GNU General Public License as published by
   the Free Software Foundation, either version 3 of the License, or
   (at your option) any later version.
-  
+
   ESPResSo is distributed in the hope that it will be useful,
   but WITHOUT ANY WARRANTY; without even the implied warranty of
   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
   GNU General Public License for more details.
-  
+
   You should have received a copy of the GNU General Public License
-  along with this program.  If not, see <http://www.gnu.org/licenses/>. 
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 #ifndef RANDOM_H
 #define RANDOM_H
 
-/** \file random.hpp 
+/** \file random.hpp
 
     A random generator
 */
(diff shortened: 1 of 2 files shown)


RudolfWeeber:
I find this a quite dangerous interface change. And I find the current behaviour the correct one.

In my opinion any randomness in the results of a simulation script should be explicitly introduced in the script.

If there is a majority for merging this, there should be a very prominent warning in the release notes

mkuron:
I agree with @RudolfWeeber. @fweik, is there any specific problem this is supposed to address or did you just give in to @jonaslandsgesell's repeated requests about RNG checkpointing?

fweik:
The powers to be thought this is a good idea...

jonaslandsgesell:
We have a python function system . set_random_state_PRNG () which allows the user to set a fully random state of the PRNG. And we also have a property "seed" in the system class which allows the user to set a seed.

The only question is whether we want a random seed as default here. In my opinion it is a good idea to do so since a simulation result should not depend on the seed which is chosen. Therefore the seed is nothing a user should have to set.
The only time a seed is intended to be set is when trying to reproduce trajectories (e.g. for test cases). However in this case I think the developer who writes the test case is responsible for using the same seed.
I think in 99% of the runs of espresso a random seed is a sensible default value.

Looks good to me. The changes are fine to merge.

@mkuron I did I requested PRNG checkpointing! Have a look there #577 
PS: I talked to Floh and this was a wish from Christian. Maybe this wish is related to this bug: #978

RudolfWeeber:
>The only question is whether we want a random seed as default here. In my opinion it is a good idea to do so since a simulation result should not >depend on the seed which is chosen.

If the result does not depend on the seed, then 1 is just fine as choice.

 

 

>The only time a seed is intended to be set is when trying to reproduce trajectories 

This is precicely, why it should not be randomly chosen. Because in that case, it is not possible to reproduce existing simulation results unless whoever generated them wrote down the randomly chosen seed.

kosovan:
I fully agree with Rudolf here. With sufficient statistics, the result should not depend on the seed, then a fixed seed is just fine. If the user has reasons to use different seeds, there is an option to do so.

Reproducibility of results is an important issue. For example, if you encounter a rare bug which may depend on seed, it may become very hard to reproduce it.

KaiSzuttor:
Regarding the reproducibility in the hopefully unusual case of debugging please have a look at the gromacs docu addressing this issue: http://www.gromacs.org/Documentation/Terminology/Reproducibility
So having the same seed does not at all mean that the trajectory is reproducable.
I think the default use case still is having to perform several runs for better statistics.

kosovan:
Well, the gromacs docs explains that simulations may not be exactly reproducible mainly due to accumulation of rounding errors if you use different architectures or different versions of libraries. However, this only shows up after a big number of steps. And if you use the same computer, libraries, same random seed, it is 100% reproducible.

With randomly chosen random_seed the situation is different: the simulation becomes totally irreproducible from the very beginning. Should this be the preferred default behaviour? What is the gain? Saving one line in the script which sets the seed? Is that worth losing reproducible default behaviour?

I think that Kai's default use case is only the production stage. Before that, in the testing and tuning stage it happens that scripts crash unexpectedly, e.g. integrator crash with the "bond broken" message discussed repeatedly in the user mailing list. In the tuning stage it is very confusing if your simulation behaves randomly. And it takes a while to figure out that you need to explicitly set random seed to achieve reproducible behaviour.

src/core/random.hpp
```diff
@@ -78,51 +78,55 @@ void init_random_seed(int seed);
...
  */
-inline double gaussian_random_cut(void){
+inline double gaussian_random_cut(void) {
   using namespace Random;
-  const double random_number=1.042267973*normal_distribution(generator);
-  
-  if ( fabs(random_number) > 2*1.042267973 ) {
-    if ( random_number > 0 ) {
-      return 2*1.042267973;
+  const double random_number = 1.042267973 * normal_distribution(generator);
```
hmenke: What is this number? (`1.042267973`)  It keeps appearing here but there is no documentation whatsoever explaining the purpose and origin of it.  It was originally introduced in https://github.com/espressomd/espresso/commit/8185eae2d29463bbd7a649617ea9d662d2f821a8 also without docs.

KaiSzuttor:
I think this is a hack for too large random number leading to undesired behaviour.

On 1 Mar 2017, at 14:34, Henri Menke <notifications@github.com<mailto:notifications@github.com>> wrote:


@hmenke commented on this pull request.

________________________________

In src/core/random.hpp<https://github.com/espressomd/espresso/pull/1039#discussion_r103683395>:

>    using namespace Random;
-  const double random_number=1.042267973*normal_distribution(generator);
-
-  if ( fabs(random_number) > 2*1.042267973 ) {
-    if ( random_number > 0 ) {
-      return 2*1.042267973;
+  const double random_number = 1.042267973 * normal_distribution(generator);


What is this number? (1.042267973) It keeps appearing here but there is no documentation whatsoever explaining the purpose and origin of it. It was originally introduced in 8185eae<https://github.com/espressomd/espresso/commit/8185eae2d29463bbd7a649617ea9d662d2f821a8> also without docs.

hmenke:
This should be merged already.

RudolfWeeber:
I would like to discuss this in an Espresso meeting, please.

fweik:
Please don't merge. We will come to a decision in the next offline Espresso meeting on Tuesday.

davidsean:
We did not agree on the best default behaviour between automatic and manual seeding.

The option to REQUIRE a seed before ESPresSo starts using random numbers was brought up. This forces the user to take an explicit stance on what the seed is, and removes the dangers associated with (wrong) expected behaviour.

To make it easy on the user, the error message will display the python command to seed randomly.
The test scripts will all have to be modified to include the explicit seeding.

It just occurred to me that perhaps the error message could simply state that no manual seeding was performed and that ESPresSo has chosen a seed.

jonaslandsgesell:
There is literature on the langevin thermostat: If the seed is the same trajectories converge although initial conditions may differ. Maybe we dig a bit more into this. It was a group at LANL who described it:
http://pubs.acs.org/doi/abs/10.1021/ct800573m

Please use a random seed and not always the same seed.

kosovan:
Jonas, I guess you meant: "please use different seeds in otherwise identical production runs".
If you want to reproduce the result, you should use the same seed.

I agree with David that requiring a seed before espresso starts using random numbers removes the dangers of unexpected behaviour. The user should be aware what the seed does, and if he/she is not aware, it is not the fault of Espresso.

jonaslandsgesell:
@Peter As the Paper outlines the choice of an identical random seed for production runs introduces a synchronization bias in simulation data that are acquired via averaging over multiple simulation runs. This is true, even if the simulations are not identical in the beginning, for example because they differ in initial conditions (like atom velocities). As the paper states this is a risk when publishing data. We are in principle discussing whether we want espresso users to be at risk of suffering synchronization bias by default or not. Maybe you can do an experiment and base your decision on that: Ask your students after a lecture about the Layapunov instability whether they think the seed in a Langevin simulation matters if the atom positions are set randomly in the beginning. I suppose a majority of the students might vote for "No".

fweik:
Closed in favour of #1211

PR #1042, opened 2017-02-22
cmake: allow user to override PYTHON_INSTDIR

junghans:
This is useful on HPC machines, where the python install location does not necessarily follow the standard FHS conventions.


junghans:
This needs a build retrigger.

junghans:
`pypresso` isn't really the issue, I can control its installation with the `${BINDIR}` option. I wanted to install espresso in python's `<site package path>/{openmpi,mpich}` for two different mpi version.

If it seems more useful we could make PYTHON_INSTDIR a true cmake option (in cache), but it seems not really need at this point as not many users install espresso.

PR #1044, opened 2017-02-22
cmake: remove soversion

junghans:
Espresso is a python module now, so .so versioning is not needed
anymore. Move all libraries into python module path as well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,10 +25,6 @@ project(ESPResSo)
 enable_language(CXX)
 
 set(PROJECT_VERSION "4.0-dev")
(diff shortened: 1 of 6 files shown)


fweik:
IIRC there was a reason for this. @mkuron do you remember something?

junghans:
I added this in  #752, but without tcl there is no point in keeping it.

mkuron:
Why did it make sense for Tcl in the first place?
The Python modules obviously don't have an so version, only libEspressoCore etc. do, which the Python modules link to. So it's not obvious to me how the Python migration changes thing.

PR #1054, opened 2017-03-01
H5md proposol conform

KaiSzuttor:
PR Checklist
------------
 - [x] Tests?
   - [x] Interface

### DIFF
--- a/src/core/io/writer/h5md/h5md_core.hpp
+++ b/src/core/io/writer/h5md/h5md_core.hpp
@@ -107,10 +107,6 @@ private:
   MPI_Comm m_hdf5_comm;
   bool m_already_wrote_bonds = false;
 
-  bool check_file_exists(const std::string &name) {
-    std::ifstream f(name.c_str());
-    return f.good();
-  };
   /**
    * @brief Method to check if the H5MD structure is present in the file.
    * Only call this on valid HDF5 files.
--- a/src/core/io/writer/h5md/h5md_core.hpp
+++ b/src/core/io/writer/h5md/h5md_core.hpp
@@ -19,7 +19,6 @@
   along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-
 #ifndef ESPRESSO_H5MD_CORE_HPP
 #define ESPRESSO_H5MD_CORE_HPP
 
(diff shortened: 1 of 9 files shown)


src/core/io/writer/h5md/h5md_core.hpp
```diff
@@ -39,180 +38,187 @@
...
+  std::unordered_map<std::string, h5xx::dataset> datasets;
 };
 
-
 struct incompatible_h5mdfile : public std::exception {
-    const char* what () const throw ()
-    {
-        return "The given hdf5 file does not have a valid h5md structure!";
-    }
+  const char *what() const throw() {
```
KaiSzuttor: Isn't "throw()" already telling the compiler that this function does not throw anything, therefore "noexcept" is not needed here?

src/core/io/writer/h5md/h5md_core.hpp
```diff
@@ -39,180 +38,187 @@
...
+   */
+  void Write(int write_dat);
+
+  /**
+   * @brief Method to write the energy contributions to the H5MD file.
+   * @param Boolean values for total, kinetic.
+   * \todo Implement this method.
+   */
+  void WriteEnergy(bool total = true, bool kinetic = true);
+  std::string &filename() { return m_filename; };
```
KaiSzuttor: this is due to the scriptinterface access, I just adopted this structure from e.g. shapes/Wall.hpp.

src/core/io/writer/h5md/h5md_core.hpp
```diff
@@ -39,180 +38,187 @@
...
+  /**
+   * @brief Method to force flush to h5md file.
+   */
+  void Flush();
+
+private:
+  MPI_Comm m_hdf5_comm;
+  bool m_already_wrote_bonds = false;
+
+  bool check_file_exists(const std::string &name) {
```
hmenke: This function is unused now.  Perhaps remove?

src/core/io/writer/h5md/h5md_core.hpp
fweik: This is technically no necessary, you could e.g. return a const& and make the property read only (which a think it is?)

PR #1055, opened 2017-03-02
Reaction ensemble merge python

jonaslandsgesell:
This PR introduces the Reaction Ensemble, the Wang-Landau Reaction Ensemble as well as the Constant pH ensemble to Espresso. Further minor bugs in the particle counting feature are fixed.

Documentation is provided in Sphinx and the user guide.

Tests for both the reaction ensemble and the constant pH ensemble are provided.

An Example is provided for the reaction ensemble.


KaiSzuttor:
You can remove the tcl testcase.

jonaslandsgesell:
Please note that this pull request is independent of the implementation of the particle counting feature (which Tobi implemented in paricle_data.cpp and which is called "grandcanonical feature" in the user guide - apart from some memory fixes in the grandcanonical feature). Changes to the particle counting feature can bei made independently of this pull request. I would therefore suggest to seperate a refactoring of the grandcanonical feature and this reaction ensemble feature. The Reaction Ensemble feature further has no common base with the reaction feature in Espresso.

src/core/reaction_ensemble.cpp
```diff
@@ -0,0 +1,2061 @@
...
+
+//global variable
+bool system_is_in_1_over_t_regime=false;
+
+int number_of_times_in_the_same_bin=0;//for avoiding becoming trapped in HMC-Wang-Landau
+int last_bin_index=-10;//for avoiding becoming trapped in HMC-Wang-Landau
+int invalid_charge=-10000;
+
+double minimum_average_of_hisogram_before_removal_of_unused_bins=1000.0; //make this accessible from interface (use 90 for free polymer, 1000 for rod system, this is determined by trial and error)
+
```
fweik: What's with the magic numbers?

src/core/reaction_ensemble.cpp
```diff
@@ -0,0 +1,2061 @@
...
+#include "global.hpp" //for access to global variables
+#include "particle_data.hpp" //for particle creation, modification
+#include "statistics.hpp" //for distto
+#include "integrate.hpp" //for time_step
+#include <stdlib.h>  // for qsort()
+#include <stdio.h> //for getline()
+#include <iostream> //for std::cout
+#include <fstream> //for std::ifstream, std::ofstream for input output into files
+#include "utils.hpp" // for PI
+
```
fweik: Please put everything in a namespace. You have symbols like replace() that are globally visible and bound to cause collisions.

src/core/reaction_ensemble.cpp
```diff
@@ -0,0 +1,2061 @@
+//method according to smith94x for the reaction ensemble at constant volume and temperature, for the reaction ensemble at constant pressure additionally employ a barostat!
```
fweik: Please run clang-format on this file

src/core/reaction_ensemble.cpp
```diff
@@ -0,0 +1,2061 @@
...
+	generic_oneway_reaction(reaction_id);
+	
+	return 0;
+}
+
+
+//checks the reaction_ensemble struct for valid parameters
+int check_reaction_ensemble(){
+	int check_is_successfull =ES_OK;
+	if(current_reaction_system.standard_pressure_in_simulation_units<0 and not (std::abs(constant_pH-(-10)) >0.00001) ){
```
fweik: Replace magic number by numeric_limits<>::epsilon()

src/core/reaction_ensemble.cpp
```diff
@@ -0,0 +1,2061 @@
...
+		init_energies(&total_energy);
+		master_energy_calc();
+	}
+  	int num_energies=total_energy.data.n;
+	double kinetic_energy =total_energy.data.e[0];
+	double sum_all_energies=0;
+	for(int i=0;i<num_energies;i++){
+		sum_all_energies+= total_energy.data.e[i];
+	}
+	for (int i = 0; i < n_external_potentials; i++) {
```
fweik: This should go into master_energy_calc() if it is not included there.

src/core/reaction_ensemble.cpp
```diff
@@ -0,0 +1,2061 @@
...
+	#endif
+	//set type
+	int desired_type=previous_type+100;//+100 in order to assign types that are out of the "usual" range of types
+	int err_code_type=set_particle_type(p_id, desired_type);
+	return err_code_type;
+}
+
+
+//copies last particle to the particle with id p_id and deletes the then last one
+int delete_particle (int p_id) {
```
fweik: I'm not happy with this. This changes the id of the unrelated particle with id max_seen_particle. This is rather unexpected behavior.

src/core/reaction_ensemble.cpp
```diff
@@ -0,0 +1,2061 @@
...
+		sum_all_energies+= total_energy.data.e[i];
+	}
+	for (int i = 0; i < n_external_potentials; i++) {
+        	sum_all_energies += external_potentials[i].energy;
+        }
+
+	return sum_all_energies-kinetic_energy;
+}
+
+void make_reaction_attempt(single_reaction* current_reaction, double** changed_particles_properties, int* _len_changed_particles_properties, int** p_ids_created_particles, int* _len_p_ids_created_particles, double** hidden_particles_properties, int* _len_hidden_particles_properties) {
```
fweik: This function is to long and very hard to read. Please refactor. There seems to be considerable repetition in the loop bodies, this is potentially a good starting point.

src/core/reaction_ensemble.cpp
```diff
@@ -0,0 +1,2061 @@
...
+	}
+
+	int* p_ids_created_particles =NULL;
+	int len_p_ids_created_particles=0;
+	double* hidden_particles_properties=NULL;
+	int len_hidden_particles_properties=0;
+	double* changed_particles_properties=NULL;
+	int len_changed_particles_properties=0;
+	int number_of_saved_properties=3; //save p_id, charge and type of the reactant particle, only thing we need to hide the particle and recover it
+	make_reaction_attempt(current_reaction, &changed_particles_properties, &len_changed_particles_properties, &p_ids_created_particles, &len_p_ids_created_particles, &hidden_particles_properties, &len_hidden_particles_properties);
```
fweik: If you make a new type ReactionAttempt, you can make this much less messy. Also, since as far as I can see this is used only locally it should be easy to replace the pointer mess with vectors.

src/core/reaction_ensemble.cpp
```diff
@@ -0,0 +1,2061 @@
...
+	//do reaction
+	//save old particle_numbers
+	int* old_particle_numbers=(int*) calloc(1,sizeof(int) *(current_reaction_system.nr_different_types));
+	for(int type_index=0;type_index<current_reaction_system.nr_different_types;type_index++){
+		number_of_particles_with_type(current_reaction_system.type_index[type_index], &(old_particle_numbers[type_index])); // here could be optimized by not going over all types but only the types that occur in the reaction
+	}
+
+	int* p_ids_created_particles =NULL;
+	int len_p_ids_created_particles=0;
+	double* hidden_particles_properties=NULL;
```
fweik: Please use a vector of structs or similar instead of storing ints in doubles. This is cryptic and error prone.

fweik: Same of course for the changed particles

src/core/reaction_ensemble.cpp
```diff
@@ -0,0 +1,2061 @@
...
+	for(int i=0;i<current_reaction->len_product_types;i++) {
+		int nu_i=current_reaction->product_coefficients[i];
+		int N_i0= old_particle_numbers[find_index_of_type(current_reaction->product_types[i])];
+		factorial_expr=factorial_expr*factorial_Ni0_divided_by_factorial_Ni0_plus_nu_i(N_i0,nu_i); //zeta = 1 (see smith paper) since we only perform one reaction at one call of the function
+	}
+	return factorial_expr;
+}
+
+int generic_oneway_reaction(int reaction_id){
+	double volume = current_reaction_system.volume;
```
fweik: Please make things that are not supposed to change anymore const, here and evrywhere

src/core/reaction_ensemble.cpp
```diff
@@ -0,0 +1,2061 @@
...
+		return 0;
+	}
+	
+	//calculate potential energy
+	double E_pot_old=calculate_current_potential_energy_of_system(0); //only consider potential energy since we assume that the kinetic part drops out in the process of calculating ensemble averages (kinetic part may be seperated and crossed out)
+	
+	//find reacting molecules in reactants and save their properties for later recreation if step is not accepted
+	//do reaction
+	//save old particle_numbers
+	int* old_particle_numbers=(int*) calloc(1,sizeof(int) *(current_reaction_system.nr_different_types));
```
fweik: vector

src/core/reaction_ensemble.cpp
```diff
@@ -0,0 +1,2061 @@
...
+		//accept
+		
+		//delete hidden reactant_particles (remark: dont delete changed particles)
+		//extract ids of to be deleted particles and sort them. needed since delete_particle changes particle p_ids. start deletion from the largest p_id onwards
+		int to_be_deleted_hidden_ids[len_hidden_particles_properties];
+		for(int i=0;i<len_hidden_particles_properties;i++) {
+			int p_id = (int) hidden_particles_properties[i*number_of_saved_properties];
+			to_be_deleted_hidden_ids[i]=p_id;
+		}
+		qsort(to_be_deleted_hidden_ids, len_hidden_particles_properties, sizeof(int), intcmp);
```
fweik: Please use std::sort (more efficient). Preferably of course with a vector.

src/core/reaction_ensemble.cpp
```diff
@@ -0,0 +1,2061 @@
...
+		
+		for(int i=0;i<len_hidden_particles_properties;i++) {
+			int status=delete_particle(to_be_deleted_hidden_ids[i]); //delete particle
+		}
+		reaction_is_accepted= 1;
+	} else {
+		//reject
+		//reverse reaction
+		//1) delete created product particles
+		qsort(p_ids_created_particles, len_p_ids_created_particles, sizeof(int), intcmp); // needed since delete_particle changes particle p_ids. start deletion from the largest p_id onwards
```
fweik: Please use std::sort (more efficient). Preferably of course with a vector.

src/core/reaction_ensemble.cpp
```diff
@@ -0,0 +1,2061 @@
...
+		for(int i=0;i<len_p_ids_created_particles;i++){
+			delete_particle(p_ids_created_particles[i]);
+		}
+		//2)restore previously hidden reactant particles
+		for(int i=0;i<len_hidden_particles_properties*number_of_saved_properties;i+=number_of_saved_properties) {
```
fweik: Factor out repeated part with line 340

src/core/reaction_ensemble.cpp
```diff
@@ -0,0 +1,2061 @@
...
+		bool reactant_type_i_is_known=is_in_list(reactant_types[i],current_reaction_system.type_index,current_reaction_system.nr_different_types);
+		if (reactant_type_i_is_known==false){
+			current_reaction_system.type_index=(int*) realloc(current_reaction_system.type_index, sizeof(int)*(current_reaction_system.nr_different_types+1));
+			current_reaction_system.type_index[current_reaction_system.nr_different_types]=reactant_types[i];
+			current_reaction_system.nr_different_types+=1;
+			status_gc_init_temp=init_type_array(reactant_types[i]); //make types known in espresso
+			status_gc_init=status_gc_init || status_gc_init_temp;
+		}
+	}
+	for (int i =0; i<len_product_types;i++){
```
fweik: Code duplication from line 389, put into function

src/core/reaction_ensemble.cpp
```diff
@@ -0,0 +1,2061 @@
...
+
+
+int hide_particle(int p_id, int previous_type){
+	//remove_charge and put type to a non existing one --> no interactions anymore (not even bonds contribute to energy) it is as if the particle was non existing
+	#ifdef ELECTROSTATICS
+	//set charge
+	set_particle_q(p_id, 0.0);
+	#endif
+	//set type
+	int desired_type=previous_type+100;//+100 in order to assign types that are out of the "usual" range of types
```
fweik: No.

src/core/reaction_ensemble.cpp
```diff
@@ -0,0 +1,2061 @@
...
+		return -1;	
+	}
+	return p_id;
+}
+
+//the following 3 functions are directly taken from ABHmath.tcl
+double vec_len(double* vec){
+	double veclen=0;
+	for(int i=0;i<3;i++){
+		veclen+=pow(vec[i],2);
```
fweik: Use std::pow, std::sqrt (faster)

src/core/reaction_ensemble.cpp
```diff
@@ -0,0 +1,2061 @@
...
+double vec_len(double* vec){
+	double veclen=0;
+	for(int i=0;i<3;i++){
+		veclen+=pow(vec[i],2);	
+	}
+	return sqrt(veclen);
+}
+
+int vecnorm(double* vec, double desired_length){
+	double veclen=vec_len(vec);
```
fweik: const double scale =  desired_length / vec_len(vec);

src/core/reaction_ensemble.cpp
```diff
@@ -0,0 +1,2061 @@
...
+		}
+		if (vec_len(vecrandom)<=1)
+			break;
+	}
+	vecnorm(vecrandom,desired_length);
+	return 0;
+}
+
+//for sorting integer lists, derived from http://rosettacode.org/wiki/Sort_an_integer_array#C, function is licensed under GPL
+int intcmp(const void *aa, const void *bb){
```
fweik: aka std::less<int>

src/core/reaction_ensemble.cpp
```diff
@@ -0,0 +1,2061 @@
...
+//for sorting integer lists, derived from http://rosettacode.org/wiki/Sort_an_integer_array#C, function is licensed under GPL
+int intcmp(const void *aa, const void *bb){ 
+    const int *a = (int*) aa, *b =(int*)  bb;
+    return (*a > *b) ? -1 : (*a < *b);
+}
+
+
+bool is_in_list(int value, int* list, int len_list){
+	if(len_list==0){
+		return true;
```
fweik: Why should an empty list contain every value?

src/core/reaction_ensemble.cpp
```diff
@@ -0,0 +1,2061 @@
...
+	int p_id;
+	double E_pot_old=calculate_current_potential_energy_of_system(0);
+
+	int particle_number_of_type;
+	number_of_particles_with_type(type, &(particle_number_of_type));
+	if(particle_number_of_type==0){
+		bool got_accepted=false;
+		return got_accepted;
+	}
+	particle_number_of_type=1;
```
fweik: This is fixed to 1 but reused. Please remove the variable if this is correct or fix.

src/core/reaction_ensemble.cpp
```diff
@@ -0,0 +1,2061 @@
...
+	while(changed_particle_counter<particle_number_of_type){
+		if(changed_particle_counter==0){
+			find_particle_type(type, &p_id);
+		}else{
+			//determine a p_id you have not touched yet
+			while(is_in_list(p_id,p_id_s_changed_particles,changed_particle_counter)){
+				find_particle_type(type, &p_id); //check wether you already touched this p_id
+			}
+		}
+		Particle part;
```
fweik: Why do you copy the part instead of manipulating it directly? Also I don't believe that this code is correct. It deletes a lot of properties like old_pos, which potentially leads to incorrect short range forces, removes all bonds, exclusions ...

src/core/reaction_ensemble.cpp
```diff
@@ -0,0 +1,2061 @@
...
+	}
+		if(attempts==max_tries){
+		//reversing
+		//create particles again at the positions they were
+		for(int i=0;i<particle_number_of_type;i++){
+			double pos_x=particle_positions[3*i];
+			double pos_y=particle_positions[3*i+1];
+			double pos_z=particle_positions[3*i+2];
+			double pos_vec[3]={pos_x,pos_y,pos_z};
+			place_particle(p_id_s_changed_particles[i],pos_vec);
```
fweik: place_particle(p_id_s_changed_particles[i],&particle_positions[3*i]);

src/core/reaction_ensemble.cpp
```diff
@@ -0,0 +1,2061 @@
...
+		}
+	}
+	
+	
+	//change polymer conformation if start and end id are provided
+	double old_pos_polymer_particle[3*(end_id_polymer-start_id_polymer+1)];
+	if(start_id_polymer>=0 && end_id_polymer >=0 ){
+
```
fweik, line 802: Put into function

src/core/reaction_ensemble.cpp
jonaslandsgesell: Good idea!

src/core/reaction_ensemble.cpp
jonaslandsgesell: I remove the option to remove unused bins from the Wang-Landau grid. This was introduced for energy reweighting simulations of polymers which tend to get trapped in Wang-Landau simulations. This trapping was e.g. mentioned in:
Improving the Wang–Landau algorithm for polymers and proteins, DOI: 10.1002/jcc.21660

src/core/reaction_ensemble.cpp
jonaslandsgesell: Good idea!

src/core/reaction_ensemble.cpp
jonaslandsgesell: What is done here is to sum up all individual energy contributions. This is not done in master_energy_calc(). I agree with you that this could be done in a function in energy.cpp.
For now this kind of task always happened in the interface e.g. in analyze.pyx
```
        for i in range(c_analyze.total_energy.data.n):
            tmp += c_analyze.total_energy.data.e[i]

        e["total"] = tmp
```
Should I change this?

src/core/reaction_ensemble.cpp
jonaslandsgesell: done

src/core/reaction_ensemble.cpp
jonaslandsgesell: directly use function call

src/core/reaction_ensemble.cpp
fweik: Shouldn't this be chosen by the user if it depends on the system?

src/core/reaction_ensemble.cpp
fweik: No, i mean the external potential should be handled by master_energy_calc, not here

src/core/reaction_ensemble.cpp
jonaslandsgesell: Yes as I noted in the comment this should be accessible from the interface. I would rather drop the support for simulating polymers with Wang-Landau energy reweighting simulations for now, since this not trivial to test for correctness.

src/core/reaction_ensemble.cpp
jonaslandsgesell: I got that code from the old tcl interface:

https://github.com/espressomd/espresso/blob/master/src/tcl/energy_tcl.cpp around line 45
```
  value = total_energy.data.e[0];
  for (i = 1; i < total_energy.data.n; i++)
    value += total_energy.data.e[i];
  
  for (i = 0; i < n_external_potentials; i++) {
    value+=external_potentials[i].energy;
  }
```
I wonder why the code with the external potentials is missing in the python interface in analyze.pyx. I also wonder why the energy calculation happens in the interface.

src/core/reaction_ensemble.cpp
jonaslandsgesell: done

src/core/reaction_ensemble.cpp
jonaslandsgesell: done

src/core/reaction_ensemble.cpp
jonaslandsgesell: totally removed after this is not needed with std::sort()

src/core/reaction_ensemble.cpp
fweik: Yes, this should be in the core.

src/core/reaction_ensemble.cpp
jonaslandsgesell: Since this function is in principle capable of moving more than just one particle of this type, I renamed it and made the case that only one particle is to be moved the default parameter.

src/core/reaction_ensemble.cpp
jonaslandsgesell: yes! that s better since it avoids the creation of variables pos_x,...

src/core/reaction_ensemble.cpp
jonaslandsgesell: The concept of moving particles around in the id range and deleting always only the last particle avoids that the particle id range has holes. Further using this we prevent the biggest particle id to become extremely large: Otherwise each time a new particle is inserted we would get a new id.
When following the sphinx documentation for setting up reactions polymer particles keep their id, only free ions change ids. However this is OK since the reaction ensemble is derived from a grand canonical reaction scheme, you would not expect particle ids to be always present (or keep their meaning) since it is not guaranteed that a certain particle id is around at all. It is not unexpected that particle ids might change their meaning in a grand canonical simulation scheme.

src/core/reaction_ensemble.cpp
jonaslandsgesell: I tried that before. Unfortunately running clang-format makes the code unreadable since the variable/function names are long. Still I would not shorten the variable names since they are fully descriptive.

src/core/reaction_ensemble.cpp
jonaslandsgesell: removed the check for an empty list from this function and put it into the other places where it was needed

src/core/reaction_ensemble.cpp
jonaslandsgesell: good idea!

src/core/reaction_ensemble.cpp
jonaslandsgesell: good idea, done

src/core/reaction_ensemble.cpp
jonaslandsgesell: This code is now part of the function do_global_mc_move_for_particles_of_type() which is only to be used for changing particle positions. Particle part is only used to retrieve the data related to a certain particle. I use a copy of a particle to retrieve the data since get_particle_data() expects one. Since get_particle_data() only gets the properties of one particle using mpi_recv_part() this could be faster than updating the global partcfg struct every time I want to access the property of one of only a few particles. Is this reasoning correct?

The function place_particle() retains all other properties like all bonds and only places the particle to a new position. I do not see how properties could become deleted here (except of the position, which is intended).

src/core/reaction_ensemble.cpp
jonaslandsgesell: good idea! done

src/core/reaction_ensemble.cpp
jonaslandsgesell: Can you specify what you mean here?

src/core/reaction_ensemble.cpp
jonaslandsgesell: done

src/core/reaction_ensemble.cpp
jonaslandsgesell: done

src/core/reaction_ensemble.cpp
jonaslandsgesell: done

KaiSzuttor:
Just as a reminder: if you add doxygen documentation, you can use short variable names.

src/core/reaction_ensemble.cpp
fweik: This hinders composablity. Also max_seen_particle does not need to take part in any grand canonical stuff. There is no reason why you would only have polymers and ions in your system. Please come up with a different solution.

src/core/reaction_ensemble.cpp
jonaslandsgesell: done, moved calculate_current_potential_energy_of_system() to energy.cpp, included the usage in analyze.pyx. Now the python version of Espresso also includes external potentials in the energy calculation (like the tcl version before)

src/core/reaction_ensemble.cpp
jonaslandsgesell: the user now provides the type which he has to make sure has no interactions. This is documented in sphinx.

src/core/reaction_ensemble.cpp
jonaslandsgesell: Done

RudolfWeeber:
Concerning the Python interface, I think, the following should be addressed


Please use pep8 naming convention: Wang_Landau_has_converged -> WangLandauHasConverged

The ReactionEnsemble class itself lacks a docstring

The various print_... function should probably be replaced by getters returning re-usable values. 

Most setters have set_... prefix, but default_charge() has not. Please unify.

Some methods that do something have a do_... prefix, others don't, e.g., reaction(). Pleaze unify.

Since there are free() methods, probably the classes in py should have destructors, which call the frees
  once the py object goes out of scope or is deleted.

RudolfWeeber:
Regarding the Python test cases:

Please add tests for the various setters. There are nearly 500 lines of Python interface code, so it's not immediately apparant, whether it does what it should.

The test running code at the very bottom of the files differs from those in other test cases. 
If there is a reason for that, please put in a comment. Otherwise, please use the same code as elsewhere.

The doc strings of the test classes are wrong (sth about hdf5). Please put in a docstring mentioning, what exactly is tested.

RudolfWeeber:
Regarding the code in the core:

There are lots of global variables. Is there a good reason, why both, the methods and the data cannot be in a single class?
This wouldn't require all that much change to the code. Then, one would need only a single global, namely the instance of that class.

In the structs defined in the hpp file, there are still quite a few pointer arrays.
Please replace them by vectors, unless there is a very good reason not to. This makes memory leaks and segfaults much more unlikely, and one gets rid of the hard-to-verify allocation and freeing code.


A bunch of routines printf error messages and then return ES_ERR. Shouldn't they throw exceptions instead, so the error can be processed.
This needs 
except +
on the relevant imports in the pxd file.

Please provide doxygen comments at the very least for those methods and data structures exposed in the hpp file.
double calculate_current_potential_energy_of_system_wrap(int unimportant_int)
what is the parameter for?

jonaslandsgesell:
Rudolf please keep in mind that this was developed as C code. I started working on the code before the transition to C++ was clearly expressed to be desired. Before the recent introduction of vectors this code could be compiled with a C compiler.

There is an argument in the wrap function since I use a function pointer "current_collective_variable->determine_current_state_in_collective_variable_with_index" which can either point to "calculate_degree_of_association" or "calculate_current_potential_energy_of_system_wrap". Since I want to call later only the function "determine_current_state_in_collective_variable_with_index" independent of the collective variable both functions have to have the same signature (at least in C, which is the way the code was written).
Since this code was all written in C it also works a lot with pointers. Especially the pointers in the wang_landau_system struct I would not like to touch since there is a whole logic built on it and rewriting it in C++ would not bring more functionality. This is also the reason why the code is not encapsulated in a class. In C there are no classes.
The reason why the code does not throw exceptions is also easy: There are no exceptions in C so they are not used throughout the code. Where exceptions are used they were introduced later.

RudolfWeeber:
> Rudolf please keep in mind that this was developed as C code. I started working on the code before the transition to C++ was clearly expressed to be desired. Before the recent introduction of vectors this code could be compiled with a C compiler.
Its not about converting this into pure c++. 
However, a large number of global variables is very much in the way of testability and modularity. 
While we cannot easily get rid of all the globals already present in Espresso, we can, at least, not make it worse.
Converting the whle thing into 1-2 classes is probably the easiest and most clear way to do it.
The amount of refactoring is pretty small, because the code for accessing class member variables and vlobal variables is identical.
The same goes for function calls.


Concerning the arrays: 
As accessing elements has the same syntax, converting them into vectors is not a lot of work at all and is likely to save someone else a lot of time later debugging memory issues.

jonaslandsgesell:
I introduced Pep8 naming convention for the reaction ensemble python class, which now also has a docstring.
> lowercase with words separated by underscores as necessary to improve readability

print_ functions are changed to be getters.
There are no do_functions anymore. frees now do not have to be performed by the user anymore since the cpp destructors handle this now when an object goes out of scope.

Setters are now tested based on the get_status function in the reaction_ensemble.py test case. The docstring is fixed there too. 

> The test running code at the very bottom of the files differs from those in other test cases.
Actually it was the same code as in the h5md testcase. I changed it though.

I wrapped the C-Code into a cpp class so that there are no additional global variables anymore.
When errors are thrown the C-Code is now mixed and it now throws Cpp exceptions which get propagated to Cython.

kosovan:
The current implementation of filling the holes in the partCfg[ ] array breaks the analysis functions in chain_statistics.cpp because positions in partCfg[ ] array do not coincide with particle ids. Maybe this could work with the new implementation of partCfg - pending pull request of fweik? In any case, I think this is undesired. We are discussing with Jonas how to fix it.

RudolfWeeber:
On Tue, Jun 20, 2017 at 08:24:36AM -0700, Peter Kosovan wrote:
> The current implementation of filling the holes in the partCfg[ ] array breaks the analysis functions in chain_statistics.cpp because positions in partCfg[ ] array do not coincide with particle ids. Maybe this could work with the new implementation of partCfg - pending pull request of fweik? In any case, I think this is undesired. We are discussing with Jonas how to fix it.
I assume that the beds making up the chains have particle ids
0..N_c-1
and everything concerning the reaction has ids N_c..infty.
In that case, no holes will ever appear in the 0..N_c-range with
which analyze chains is concerned.
Hence, the analysis code will likely work, once the new PartCfg is in.
This will probably be soon, as there are only minor points left.

jonaslandsgesell:
There is no guarantee that the chain particles do not take part in the reactions. Actually this is the interesting case.

fweik:
While not perfect, this has considerably improved. Also as far as I can tell it is pretty much self contained and does not have unexpected side effects anymore. If nobody objects I would merge this.

kosovan:
Flo, I do object. If it breaks standard analysis routines, it should not be merged before it is fixed.

Currently it can happen that partCfg[index] and espressomd.System.part[index] refer to completely different data. I am afraid that the problem is deeper than just broken chain analysis routines and can cause trouble elsewhere.

Maybe it will work smoothly with the new implementation of partCfg we cannot be sure unless we understand the problem. The new partCfg is still a pending pull request, so if you merge now, we end up with a broken code.

The problem was introduced only recently, in commit dc3058763. But I have not yet fully understood what the code from this commit is supposed to do. Could you please explain what was the problem with non-contiguous particle ids? Before dc3058763 everything worked well from my point of view.

I will prepare a minimal example script and post it soon.

fweik:
Ok. I will finish the partCfg stuff today. So that should be merged within
the week. From my understanding this will fix the issue, and then we can
merge this PR.

kosovan:
Attached is the minimal working example with a reaction
HA <-> A- + H+
where A is an ionizable group on the chain.

In the earlier implementation, when HA was deleted in the reaction, it was replaced by newly created A-. Since dc30587 it seems like the hole created by deleting HA is filled by a random particle.

[ree_mwe.py.gz](https://github.com/espressomd/espresso/files/1091190/ree_mwe.py.gz)

jonaslandsgesell:
Let me look closer into this. I will check how to fix this.

fweik:
I've merged the particle config stuff.

jonaslandsgesell:
Hi Peter, the script seems to run fine now. Can you confirm this? Maybe Floh can then finally merge the PR. Best Regards Jonas

kosovan:
My original test script works well now. But I managed to get a segmentation fault when I added one line:
`esrg=system.analysis.calc_rg(chain_start=0, number_of_chains=1,chain_length=len(system.part))`

It works with the old implementation of partCfg and RE (commit a07b806) and also with the new partCfg but without RE on current master on github (commit 63f4c89a5e). I will come back to it tomorrow.

jonaslandsgesell:
The line is aready present in your script and there works OK. Can you provide more information on that?

kosovan:
Line 128 in my script was similar, but this one calculates Rg instead of Re and its last argument is not chain_l but len(system.part). Same happens with calc_re and this argument.

Try the modified script: [ree_mwe.txt](https://github.com/espressomd/espresso/files/1107927/ree_mwe.txt)

After the reaction, last particle in the system.part array does not exist but the analysis function tries to access the data. There is a check in analysis.pyx (lines 642-644) which should avoid such problem. But it assumes len(system.part) = actual number of particles which is not true anymore.

I know that such calculation of Rg physically does not make sense but for the test I wanted the analysis function to access all particles.

Sorry, I cannot work on it further today because I am busy with exams.

jonaslandsgesell:
Hey thank you Peter for your Feedback! i overlooked rg vs re. So (as you showed) the problem is, that len(system.part) is interpreted only once for setting up the analysis routine. This routine however is not updated later if the particle number changes. That s why there is a segmentation fault. This is not a problem of the Reaction Ensemble and floh can merge this.

fweik:
If this is an unrelated issue please open a ticket.

kosovan:
I am not convinced that this is a separate issue. It only happens with the reaction ensemble, while it works without it with the current master. So far the problem is localized in this branch and I see no point in opening a ticket. It is not because calc_rg is faulty but because the reaction ensemble breaks some conventions that were guaranteed before, and were described in the user guide.  I will try to look at it more closely tomorrow, today I am out of office.

jonaslandsgesell:
Hi Peter,
the problem here is, that in your new script the particle id range gets holes. This is due to your simulation setup: The first 4 particle ids have type A-, the next 4 particle ids have type H+. The reaction deletes correctly an H+. This creates a hole in the id range as it was wished by Floh and Rudolf (use print(system.part[:].id) to check). This makes the analysis routine throw an error (not a segmentation fault) "terminate called after throwing an instance of 'std::out_of_range'"
This is in my opinion not an error of the reaction ensemble but of the way the analysis routine is called here. The analysis routine was probably not designed to be called over an id range which has holes.
Thank you Peter for your checks on this PR!

fweik:
The rg calculation requires that the particles in the chain have continuous ids. This is not guaranteed they way you set up the RE. (Btw Jonas, in your original design the gap in the chain would have been filled with an unrelated particle, not leading to an error but to _very_ unexpected behavior.) To make this work the chain statistics calculation has to be changed. Currently the "chains" are defined by a start id and a length.
I think a better way would be to use something like the molid to group the particles into molecules, which is also more powerful than the current mechanism. I'm not willing to put in the work though.

kosovan:
Hi Jonas,

that is the reason why I opened this whole discussion. Not just calc_rg but ALL chain analysis routines require contiguous particle id range. If we want to break it, then we should ensure that the errors are caught. And maybe we should also check whether some other parts of Espresso, not just chain analysis, rely on paticle ID being contiguous.

A simple solution could be to modify the check_topology function in analyze.pyx to explicitly check for each particle. I also noticed that currently only some of the chain analysis functions check the topology, while others take it for granted. 

And finally you were right, this is not just RE issue. I could make it fail also on the current master if I manually delete a particle in the id range provided to calc_rg. But this seems to be a deeper problem and I would open a ticket for after I have understood it better.

I can implement the suggested fix but first want to have your opinion.

Btw: do we really need to handle holes in particle id range? Requiring contiguous particle ids would make the coding simpler and less susceptible to errors.

kosovan:
Since we have solved the issue with segfaults and chain analysis functions, I do not object against the merge anymore :-)

kosovan:
Is there anything that prevents the merge of this pull request?

fweik:
So I think we got some reasonable result here, even if it took some time. This was a very big PR, and things would have moved faster and smoother if you would have involved other devs earlier in the process.

PR #1056, opened 2017-03-03
cmake: install all shared libraries

junghans:
Currently `make install` forgets some shared libraries and hence renders any Espresso installation useless.


junghans:
Should we re-add the so numbering as well?

PR #1060, opened 2017-03-13
[python] Hopefully better comparison for time_step

### LINKED ISSUE #1051: Restoring checkpoint with LB_GPU fails
If one tries to restore a checkpointed simulation with an LB_GPU fluid, one tends to be greeted with
````
ValueError: Time Step (0.002) must be > LB_time_step (0.00200000009499)
````
This is due to the fact that the LB_GPU time step is stored as a float but is converted to a double in the interface which then loses precision.  A simple C program to understand the error is:
````c
#include <stdio.h>

int main()
{
  float f = 0.002;
  double d = f;
  printf("%.8f\n", f);  /* 0.00200000 */
  printf("%.16f\n", d); /* 0.0020000000949949 */
}
````

The following script reproduces the issue in ESPResSo.
````python
# Import the relevant features from ESPResSo

import espressomd
import espressomd.lb

# Set the regular checkpoint

outdir = "./CHK/"
checkpoint = espressomd.checkpointing.Checkpointing(checkpoint_id="checkpoint",checkpoint_path=outdir)

# Set up the simulation

system = espressomd.System()
system.time_step = 0.002
system.cell_system.skin = 0.1
system.box_l = [10,10,10]
checkpoint.register("system")

# Set up the LB fluid

lbf = espressomd.lb.LBFluid_GPU(agrid=1,dens=1,visc=1,fric=1,tau=0.002)
system.actors.add(lbf)

# Integrate for a bit and save the checkpoints

system.integrator.run(0)

lbf.save_checkpoint("lbf.chk", binary=False)

# Restart the simulation

for i in range(100):

    lbf.load_checkpoint("lbf.chk", binary=False)

    system.integrator.run(1)

    lbf.save_checkpoint("lbf.chk", binary=False)
````

### PR DESCRIPTION
**Pull request by hmenke**
Fixes #1051

### DIFF
--- a/src/python/espressomd/_system.pyx
+++ b/src/python/espressomd/_system.pyx
@@ -20,6 +20,7 @@ from __future__ import print_function, absolute_import
 include "myconfig.pxi"
 
 from globals cimport *
+from libcpp.limits cimport numeric_limits
 import numpy as np
 
 from . cimport integrate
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by fweik** [comment]
Would something like
```c++
bool less(double a, float b) {
      return (b - a) > std::numeric_limits<float>::epsilon();
}
```
be cleaner?

hmenke:
Not exactly, because we need the relative difference of the two digits.  For large numbers the absolute difference between them might grow much larger than `std::numeric_limits<float>::epsilon()`, here is an example:
````cpp
#include <iostream>
#include <limits>

int main()
{
  double d = 1000.2;
  float  f = d;
  std::cout << ( f-d > std::numeric_limits<float>::epsilon() ? "True" : "False" ) << "\n";
}
````
But using `numeric_limits` is indeed a good idea and replacing the `1e-8` prefactor by epsilon also fixes the bug.  I adjusted the PR.

fweik:
The cython version on travis does not have limits yet....

PR #1069, opened 2017-03-28
Remove unnecessary mutable specifiers

mkuron:
It turns out that none of the `mutable` specifiers were even needed.

@RudolfWeeber


hmenke (approved):
That was partly me (at least `write` and `do_write`).

PR #1071, opened 2017-03-29
Non shape based constraints

### LINKED ISSUE #1050: Re-introducing non-shape based constraints in Py branch
I'd suggest to use "Constraint" as the name for the base-class to avoid interface change
I.e.: a constraint provides a force and an energy contribution for a single particle
What's currently a constraint (i.e., is based on a distance function) would go in a sub-class of Constraint called ShapeBasedConstraint

Roughly, the following steps would be required.
In src/core/constraints/
Copy the current Constraint.cpp/hpp to ShapeBasedConstraint.cpp/hpp.
In Constraint.cpp/hpp:
  * remove everything related to shape, reflection type, penetrable
  * make the add_force() and add_energy() functions virtual and put in empty implementations

In SahepBasedConstraint.cpp/hpp
  * Rename the class to ShapeBasedConstraint and make it a derived class inheriting from Constraint
  * declare add_energy() and add_force() virtual and override. Leave the existing implementation

Add HomogeneousMagneticField.cpp/hpp containting a class derived from Constraint.
  * private member "field" as double vector of length 3
  * corresponding setter/getter
  * add_force() needs to add a torque in this case. See tcl implementation
  * add_energy() corresponding to tcl implementation

Corresponding modifications in src/script_interface/constraints:
  * copy Constraint.hpp to ShapeBasedConstraint.hpp
  * Remove shape-dependent stuff in Constraint.hpp
  * Change class names and inheritnce in ShapeBasedConstraint.hpp
  * Add script_interface class for HomogeneousMagneticField (see shapes/Wall.hpp for an example for set_parameter and get_parameter stuff for the field direction)
  * Add the HomogeneousMagneticField class in initialize.cpp

In src/python/espressomd/constraints.py
  * Add classes for ShapeBasedConstraint and HomogeneousMagneticField derived from Constraint. Probably no implementation except setting the correct _so_name is needed.

In src/core/unit_tests
  * Add a unit test for the HomogeneousMagneticField core class to check the add_force and add_energy methods
  * add the unit test to CMakeList

In testsuite/python:
  * Add a test case for the HomogeneousMagneticField python class to check 
    that correct forces and energies are applied and that the setter and getter
    for the field direction work

@fweik @pkreissl

pkreissl:
Fixes #1050 

Description of changes:
 - Makes "Constraint" a shape-independent base class
 - Adds "ShapeBasedConstraint" class which inherits from "Constraint"
 - Add python test for shape based constraints 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


RudolfWeeber:
I'll have a look.

mkuron (approved):

RudolfWeeber:
The C++ and ScriptInterface part looks ok

In src/python/espressomd/constraints.py:
* In the add() method of the Constraints class: this should either take an instance of Constraint (rather than ShapeBasedConstraint) or the parameter set of a ShapeBasedConstraint. Otherwise, you won't be able to add anything but a shape-based constraint.
* Update docstring accordingly
* Provide an explanation of what a ShapeBasedConstraint is in the class's docstring

pkreissl:
The add() method of Constraints now takes any instance of Constraint.
The default behavior for keyword-arguments is still the creation of a ShapeBasedConstraint.
Probably we should discuss in the next ESPResSo meeting on Tuesday, whether we want to keep this behavior for backward compatibility reasons -- or if this feature should go; that is only the addition of instances of Constraint should be allowed.

pkreissl:
Thanks for the comment, the test now uses the ut decorator.

PR #1073, opened 2017-03-29
I can't stop reading this PR 😂

### LINKED ISSUE #1072: P3M doesn't check for full periodicity (any more)
```
import espressomd
system = espressomd.System()
system.periodicity = [1, 1, 0]
p3m = electrostatics.P3M(bjerrum_length=1, accuracy=1e-3)
system.actors.add(p3m)
```

Returning a Segmentation fault...

hmenke:
Fixes #1072

If the periodicity is not met, P3M immediately return an error without writing a message in the `log` variable.  Afterwards this log is printed to the screen, however, if it is the NULL pointer, it will fail because `strlen` cannot determine the length of the NULL pointer.  If we properly initialize the message to empty string, the segfault is gone and we obtain a regular Python exception.

From the documentation of [`std::strlen`](http://en.cppreference.com/w/cpp/string/byte/strlen)

> The behavior is undefined if there is no null character in the character array pointed to by str.

### DIFF
--- a/src/python/espressomd/electrostatics.pxd
+++ b/src/python/espressomd/electrostatics.pxd
@@ -128,10 +128,11 @@ IF ELECTROSTATICS:
             return p3m_set_mesh_offset(mesh_offset[0], mesh_offset[1], mesh_offset[2])
 
         cdef inline python_p3m_adaptive_tune():
-            cdef char * log = ""
+            cdef char * log = NULL
             cdef int response
             response = p3m_adaptive_tune(& log)
-            print(log)
+            handle_errors("Error in p3m_adaptive_tune")
+            if log.strip(): print(log)
             return response
 
         cdef inline python_p3m_set_params(p_r_cut, p_mesh, p_cao, p_alpha, p_accuracy):
--- a/src/python/espressomd/electrostatics.pxd
+++ b/src/python/espressomd/electrostatics.pxd
@@ -128,7 +128,7 @@ IF ELECTROSTATICS:
             return p3m_set_mesh_offset(mesh_offset[0], mesh_offset[1], mesh_offset[2])
 
         cdef inline python_p3m_adaptive_tune():
-            cdef char * log = NULL
+            cdef char * log = ""
             cdef int response
             response = p3m_adaptive_tune(& log)
             print(log)


KonradBreitsprecher:
That does fix the seg fault, but it still doesn't write the correct error message..
I saw that `p3m_sanity_checks_system() ` is called on tuning p3m which checks a few things, but it writes it's error messages to the `runtimeErrorMsg()` instead of the tuning log. Not sure what's the proper way to fix this.. I feel like using the tuning log for an errormsg due to general misuse of p3m is wrong.

fweik:
I agree with Konrad. Also the log in itself is a questionable construction.

hmenke:
@KonradBreitsprecher Thanks for the hint.  `handle_errors` is the way to go there.

@fweik I think the tuning log should go into something like `runtimeInfo(...)` but in `errorhandling.hpp` there is only `runtimeWarning` and `runtimeError`.  The `runtimeInfo` can then be collected by `handle_errors`.  If this is cool with the other maintainers I will add something along those lines.

fweik:
Yeah, the macro is missing. Ideally you could give a log level...

hmenke:
I tried replacing the errors in the tuning log with `runtimeErrorMsg()` for a start.  However, the tuning loop becomes infinite then (???) and `strace` reports that ESPResSo is doing
````
open(".fftw3_1d_wisdom_back_n62.file", O_RDONLY) = 3
fstat(3, {st_mode=S_IFREG|0664, st_size=28350, ...}) = 0
read(3, "(fftw-3.3.4 fftw_wisdom #x458a31"..., 4096) = 4096
read(3, "vx 0 #x11048 #x11048 #x0 #x4b017"..., 4096) = 4096
read(3, "se2 0 #x10048 #x10048 #x0 #xd331"..., 4096) = 4096
read(3, "r 0 #x11048 #x11048 #x0 #xe2d5b5"..., 4096) = 4096
read(3, " #x0 #x57d7ce7a #x81642d8a #xc90"..., 4096) = 4096
read(3, " #x0 #xd5b45a23 #x5578823b #x32b"..., 4096) = 4096
read(3, "8 #x10048 #x0 #x64dbb4ce #x3b704"..., 4096) = 3774
read(3, "", 4096)                       = 0
close(3)
````
Where the number (here `n62`) keeps increasing.

fweik:
Hm, I still tend to merge this one, bc it still removes a bug?

hmenke:
Yes, merging it would be good, as I don't see how to resolve this problem on the fly.  I filed an issue (#1090) for that.

PR #1079, opened 2017-04-06
Make ScriptInterface more expressive and less repetitive

fweik:
Fixes: Lack of magic
Description of changes:
 - More magic

This change adds a more convenient way of binding parameters in the script interface.
If you have a class
```c++
struct A {
  int i;
};
```
the SI can be implementet by deriving from AutoParameters and just listing which name
should be bound to what in the constructor, e.g.

```c++
 struct A : public AutoParameters {
   A() : AutoParameters({"expressive_name", i});
};
```
For more complicated cases, getters and setters can be provided as functions.

PR Checklist
------------
 - [x] Tests?
   - [x] Core


src/core/utils/type_traits.hpp
```diff
@@ -0,0 +1,33 @@
...
+};
+
+/**
+ * @brief True iff T is an innstantiation of of Template.
+ */
+template <typename T, template <typename...> class Template>
+struct is_instance_of : public std::false_type {};
+
+template <typename... T, template <typename...> class Template>
+struct is_instance_of<Template<T...>, Template> : public std::true_type {};
```
hmenke, line 30: This only works for templates with `typename` parameters.  Perhaps a more expressive name is required.  Compile with `c++ -std=c++11 -c`

````cpp
#include <array>
#include <type_traits>

template <typename T, template <typename...> class Template>
struct is_instance_of : public std::false_type {};

template <typename... T, template <typename...> class Template>
struct is_instance_of<Template<T...>, Template> : public std::true_type {};

bool fail = is_instance_of< std::array<int,2> , std::array<int,2> >::value;
````

File: src/script_interface/Variant.hpp [hunk 2]
```diff
@@ -29,6 +31,64 @@ enum class VariantType {
...
+
+template <> struct infer_type_helper<std::string> {
+  static constexpr VariantType value{VariantType::STRING};
+};
+
+template <typename T>
+struct infer_type_helper<
+    T, typename std::enable_if<std::is_integral<T>::value, void>::type> {
+  static constexpr VariantType value{VariantType::INT};
+};
```
hmenke: This substitution allows down-conversion from `int` to `short` or even `char`.  I wouldn't make the assumption that people only pass sufficiently short values through this.

src/script_interface/Variant.hpp
```diff
@@ -29,6 +31,64 @@ enum class VariantType {
...
+
+template <typename T>
+struct infer_type_helper<
+    T, typename std::enable_if<std::is_floating_point<T>::value, void>::type> {
+  static constexpr VariantType value{VariantType::DOUBLE};
+};
```
hmenke: This substitution allows up-coversion to `long double` which results in bogus digits in the extra decimal places.  #1051 is a good example for this.

src/script_interface/h5md/h5md.hpp
```diff
@@ -25,23 +25,26 @@
 #include "config.hpp"
 #ifdef H5MD
 #include "ScriptInterface.hpp"
-#include <string>
+#include "auto_parameters/AutoParameters.hpp"
```
hmenke, line 28: Perhaps add `auto_parameters/` to the search path?  This here is not so nice.

hmenke (approved):

src/script_interface/Variant.hpp
fweik: You are right, I'm not sure what the right way to deal with this is. We could either only allow stuff that is large enough (excluding everything smaller than 32 bit), or do a runtime check. What do you think?

src/script_interface/Variant.hpp
fweik: The "bogus digits" are only a problem when comparisons are not done in a proper way. But this also allows for narrowing, which could be a problem. Runtime checks are not so easy for floats. So only allow float & double?

src/script_interface/Variant.hpp
hmenke: I would only offer specializations for types which are actually Variant members.  For everything else the conversion behaviour is too ambiguous and might lead to unexpected results (narrowing, precision loss, etc.).

I think it is reasonable to demand that classes compatible with ScriptInterface should use compatible types, i.e. those in Variant.  If clients are keen on messing with the types they can still do this with an explicit cast/conversion.
````cpp
struct infer_type_helper<
    T, typename std::enable_if<std::is_same<T,int>::value, void>::type> {
  static constexpr VariantType value{VariantType::INT};
};

struct infer_type_helper<
    T, typename std::enable_if<std::is_same<T,double>::value, void>::type> {
  static constexpr VariantType value{VariantType::DOUBLE};
};
````

**Comment by fweik** [comment]
Sounds reasonable.

File: src/core/utils/type_traits.hpp (hunk 1 above)
Comment (fweik) [reply src/core/utils/type_traits.hpp:30]: Suggestions?

**Comment by RudolfWeeber** [comment]
Conentwise this is very nice. 
But since there is a lot of hard to read template magic, please put more comments in the code, which explain what stuff is doing.

File: src/core/unit_tests/Variant_test.cpp [hunk 5]
```diff
@@ -72,19 +72,24 @@ BOOST_AUTO_TEST_CASE(transform_vectors_test) {
   vv.emplace_back(std::vector<Variant>{
       std::string("test"), std::vector<Variant>{1.1, 1.2, 1.3, 1.4}});
 
+  /* v = {{INT, INT, INT}, {STRING, {DOUBLE, DOUBLE, DOUBLE, DOUBLE}}} */
   auto v = Variant(vv);
 
-  transform_vectors(v);
-
+  transform_vectors(v)
```
hmenke: Semicolon missing.

hmenke (changes requested):

fweik:
@RudolfWeeber

hmenke (approved):

PR #1080, opened 2017-04-07
Typos

pkreissl:
Corrects  typos, and, in the files touched, consistently uses AE instead of BE -- since it is also used in the code base.


hmenke:
Wouldn't it make more sense to correct the typos in the Sphinx documentation?  The LaTeX user guide will eventually be abandoned.  Also it contains heaps of outdated TCL stuff.

pkreissl:
Sure, you're right. Today I just read the REAME.org in the script_interface directory and realized that there was a number of typos in there. While correcting it, I noticed that I had yet another local branch, where I had already made corrections in the ug some time ago. So I put these files in the second commit -- I think the corrections don't hurt, but if you think they are superfluous, just ignore the second commit.

hmenke (approved):

PR #1070, opened 2017-03-28
Thermo translational anisotropic test

bogdan-tanygin:
Description of changes:
- The typo fix in the langevin thermostat Python setter
- Making the translational thermo test anisotropic as the rotational one is already
- Proposed required feature for the whole test: `PARTICLE_ANISOTROPY`. It will stick the present test to fully (translationally and rotationally) anisotropic particles only. This is just a suggestion, I can be wrong
- Stability: 200 runs

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?

PR #1081, opened 2017-04-07
[Python] MS DDS GPU enable / test port / actor misc

bogdan-tanygin:
Description of changes:
- abstract actor removal method
- `DipolarDirectSumGpu` actor
- port of `dawaanr-and-dds-gpu.tcl` to Python. Validated on the Tesla C2070.

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


hmenke (approved):

fweik:
@RudolfWeeber please merge if you are happy.

testsuite/python/dawaanr-and-dds-gpu.py
```diff
@@ -10,7 +10,7 @@
 from numpy import linalg as la
 from numpy.random import random
 
-if "DIPOLES" in espressomd.features() and "CUDA" in espressomd.features() and "PARTIAL_PERIODIC" in espressomd.features():
+if ("DIPOLES" in espressomd.features()) and ("CUDA" in espressomd.features()) and ("PARTIAL_PERIODIC" in espressomd.features()) and ("ROTATION" in espressomd.features()):
```
hmenke: You can use `has_features` for that (see #1088 for an example).

hmenke (changes requested):

bogdan-tanygin:
done

testsuite/python/dawaanr-and-dds-gpu.py
bogdan-tanygin: done

testsuite/python/dawaanr-and-dds-gpu.py
```diff
@@ -0,0 +1,135 @@
...
+from espressomd.interactions import *
+from espressomd.magnetostatics import *
+from espressomd.analyze import *
+import math
+from tests_common import *
+from numpy import linalg as la
+from numpy.random import random
+from espressomd import assert_features, has_features, missing_features
+
+if has_features(["DIPOLES","CUDA","PARTIAL_PERIODIC","ROTATION"]):
```
hmenke: Sorry for requesting more changes, but ideally you'd use the `@ut.skipIf` decorator instead of wrapping the whole class definition into an `if` statement, i.e.
````python
@ut.skipIf(not has_features(["DIPOLES","CUDA","PARTIAL_PERIODIC","ROTATION"]),
           "Features not available, skipping test!")
class DDSGPUTest(ut.TestCase):
````

File: testsuite/python/dawaanr-and-dds-gpu.py [hunk 3]
```diff
@@ -0,0 +1,135 @@
...
+import numpy as np
+import espressomd
+from espressomd.interactions import *
+from espressomd.magnetostatics import *
+from espressomd.analyze import *
+import math
+from tests_common import *
+from numpy import linalg as la
+from numpy.random import random
+from espressomd import assert_features, has_features, missing_features
```
hmenke: Only import stuff you need.  `from espressomd import has_features` is sufficient.

hmenke (changes requested):

hmenke (approved):

testsuite/python/dawaanr-and-dds-gpu.py
bogdan-tanygin: Thanks for advices, @hmenke , no problem. Done.

testsuite/python/dawaanr-and-dds-gpu.py
bogdan-tanygin: Thanks, done.

RudolfWeeber:
please implement the _deactivate_method() in the python class.
It should:
1. call the deactivation method in src/core/actor/DipolarDirectSum.hpp
2. call the _deactivate_method() of the parent python class (MagnetostaticInteraction)

This is necessary, as the interaction is implemented in the Actor-framework in C++, and the actor needs to be removed from the list of active ones in the core.

bogdan-tanygin:
Done

RudolfWeeber:
What was the purpose of the change in integrate.cpp:python_integrate()?


Aside from that, the method deactivation was not quite correct yet.
* the __del__()-method is not supposed to do method deactivation, since
  it is not guaranteed that the method is even active in Espresso, when the instance of an object is deleted
  p3m=DipolarP3m()
  dds=DipolarDirectSumGpu()
  S.actors.add(p3m)
  del dds

  If this called MagnetostaticInteraction._deactivate_method(),
  it would set the dipolar method in the core to NONE, thereby deactivating the p3m
  Instead the _deactivate_method() is called when removing the actor in the
  Actors class (_remove_actor())
* The call of the _deactivate_method() from the parent class was wrong, this is
  why you had to define your method under a different name.
  I changed the name back to _deactivate_method() and used
  super(type/(self),self)._deactivate_method()
  to call the parent class method

I had to merge #1070 into #1081 to get the corrected test case.

I also speed up the mass_and_rinertia_per_particle test so that it runs in 45 secs.

Please merge the corrections_1070_1081 branch from github.com/RudolfWeeber/espresso and check whether everything you wanted to acheive still works.

bogdan-tanygin:
@RudolfWeeber thanks for useful advices. I agree with them and will merge/fix.

Concerning:
> What was the purpose of the change in integrate.cpp:python_integrate()?

Please find two .log files: the ones without and with such suggested fix, there are some additional traces:
https://gist.github.com/psci2195/540127ff523788a157a06e8492ca995d
https://gist.github.com/psci2195/b111d0a6f0708cf880a56cb69574aeb7
As you can see in both cases the test case is passed. However, without proposed fixes, the test passing is treated as correct wrongly because particles just keep forces unchanged (zero).

I think that this anomaly is related to the cython integration with the `python_integrate` call: the boolean parameter cannot be changed inside the .cpp routine as int.

Hence, probably, the proposed change of the `integrate.cpp:python_integrate()` is even more important than the current pull request: it makes a force recalculation enabling from the python possible.

bogdan-tanygin:
Done: I've merged changes of the corrections_1070_1081 branch from github.com/RudolfWeeber/espresso

RudolfWeeber:
Thanks a lot for the contribution!

PR #1083, opened 2017-04-11
Particle iterator

fweik:
Fixes # -

Description of changes:
 Added an ParticleIterator class and methods to get
  a range of such iterators to CellPList.

This allows e.g. to iterate over the local particles by
```c++
for(auto &p: local_cells.particles()) {
;
}
```

PR Checklist
------------
 - [ ] Tests?
   - [-] Interface
   - [x] Core 
 - [-] Docs?

### DIFF
--- a/src/core/utils/Range.hpp
+++ b/src/core/utils/Range.hpp
@@ -1,0 +1,43 @@
+#ifndef UTILS_RANGE_HPP
+#define UTILS_RANGE_HPP
+
+#include <utility>
+
+namespace Utils {
+
+/**
+ * @brief Helper class that wraps two iterators into
+          a range object.
+*/
+template <typename Iterator> class Range {
+  Iterator m_begin, m_end;
+
+public:
+  Range(Iterator begin, Iterator end)
+      : m_begin(std::forward<Iterator>(begin)),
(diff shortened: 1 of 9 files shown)


src/core/utils/Range.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+#include <utility>
+
+namespace Utils {
+
+/**
+ * @brief Helper class that wraps two iterators into
+          a range object.
+*/
+template <typename Iterator> class Range {
```
hmenke, line 12: To me this looks like a copy of [`boost::iterator_range`](http://www.boost.org/doc/libs/1_63_0/libs/range/doc/html/range/reference/utilities/iterator_range.html).  Why don't you use this?  It seems to be header-only as well.

fweik: I think the way to go if we want to use any range library would be ranges v3 from eric niebler, because I looks like some form of that one is going to be standarized (bosot ranges is ranges v2). Not non of this is clear yet, and ranges v3 needs C++14. So I think we can live with these 10 lines of trivial code for now.

hmenke (approved):

PR #1084, opened 2017-04-12
writevsf and writevcf

davidsean:
Added writevsf and writevcf functions, with corresponding Sphinx documentation.
one can write vtf files by running the previous two (in order) with the same fp argument.
 
These functions are located in particle_data.pyx which may seem like a funny place to put io functions. I placed them there since they are based off of writevtk, which is also there.

Unlike the TCL version, the bonds are displayed properly in VMD.
VMD atoms are defined with the same ESPreSo particle ID's.

The frames use "timestep indexed" so that selected types an be printed.

Eventually, all of these io functions (writevsf, writevcf and writevtk) should probably be in io/writer/

see example test script
[testVTF.py.txt](https://github.com/espressomd/espresso/files/916063/testVTF.py.txt)

Fixes #

Description of changes:
 - 

PR Checklist
------------
 - [x] Tests?
   - [x] Interface
   - [ ] Core 
 - [x] Docs?


src/python/espressomd/particle_data.pyx
```diff
@@ -1678,6 +1678,62 @@ cdef class ParticleList:
...
+
+        Parameters
+        ----------
+        types : str
+                Specifies the particle types. The string 'all' will write all particles
+        fp : file
+                   File pointer to write to.
+
+        """
+        global box_l
```
hmenke: What if the user doesn't name their box length variable `box_l`?  What if it is not an array?

src/python/espressomd/particle_data.pyx
```diff
@@ -1678,6 +1678,62 @@ cdef class ParticleList:
...
+        if not hasattr(types, '__iter__'):
+            types = [types]
+            
+        fp.write("unitcell {} {} {}\n".format(*(box_l)))
+        for p in self:
+            for t in types:
+                if (p.type == t or t == "all"):
+                    fp.write("atom {} radius 1 name {} type {} \n".format(p.id, p.type, p.type))
```
hmenke: Shouldn't radius be the LJ radius rather than 1 for all of them?

hmenke (dismissed):

hmenke:
You should add the test script to the testsuite to have it tested at `make check` (i.e. also on Travis and Jenkins).

src/python/espressomd/particle_data.pyx
davidsean: This is the global box_l variable from ESPresSo, not a user defined one.
see line 30: `from globals cimport box_l`

KaiSzuttor:
Shouldn't we think about transferring all io functions into the right place, i.e. the io namespace/folder?

src/python/espressomd/particle_data.pyx
davidsean: This is a generic value for VMD, the particles don't necessarily have a LJ potential so it is not obvious to find any meaningful value.

If we want to make it something more specific than just 1, perhaps:
- 0.5 since it corresponds  the size of a WCA bead: sigma.
- some radius from the friction coefficient gamma (although I think that will require LANGEVIN_PER_PARTICLE)

KaiSzuttor:
In addition, io should not depend on the interface, we can improve here and implement these simple functionality in the core.

src/python/espressomd/particle_data.pyx
hmenke: Just use the LJ radius (I'm not sure whether this is sigma in ESPResSo, there might be internal unit conversions) for the corresponding type of the particle and 1 if there is no active LJ interaction.

KaiSzuttor (commented):
I think this test does not test if the output is as expected. Please have a look at the h5md testcase.

src/python/espressomd/particle_data.pyx
```diff
@@ -1678,6 +1678,61 @@ cdef class ParticleList:
...
+ 
+        for p in self:
+            for t in types:
+                if (p.type == t or t == "all"):
+                    for b in p.bonds:
+                        if (self[b[1]].type == t or t == "all"):
+                            fp.write("bond {}:{}\n".format(p.id,self[b[1]].id))
+
+
+    def writevcf(self, fp, types='all'):
```
KaiSzuttor: Can you please move the io related stuff to a seperate module in the io folder.

src/python/espressomd/particle_data.pyx
```diff
@@ -1678,6 +1678,61 @@ cdef class ParticleList:
                     if (p.type == t or t == "all"):
                         vtk.write("{} {} {}\n".format(*p.v))
 
+
+    def writevsf(self, fp, types='all'):
+        """
+        writes a VST (VTF Structure Format) to a file.
```
KaiSzuttor: Can you please also add a documentation in the io part of the sphinx documentation and shortly describe what your function does?

testsuite/python/writevtf.py
```diff
@@ -0,0 +1,46 @@
+from __future__ import print_function
```
KaiSzuttor: As already mentioned, the testcase should be based on assertions and preferably inherit from the unittest python module. Please have a look at the other testcases in the testsuite.

KaiSzuttor (changes requested):
Some changes should be made before merging.

davidsean:
better tests in the testsuite and and description in the usersguide

KaiSzuttor (approved):

PR #1086, opened 2017-04-13
Check if LB boundary is compiled in.

KaiSzuttor:
Until now, the user would not recognize that the LB boundaries are not used since no error is thrown if the feature is not compiled in.

### DIFF
--- a/src/python/espressomd/lbboundaries.py
+++ b/src/python/espressomd/lbboundaries.py
@@ -1,25 +1,27 @@
 from __future__ import print_function, absolute_import
 from .script_interface import ScriptInterfaceHelper
+import espressomd.code_info
 
 
-class LBBoundaries(ScriptInterfaceHelper):
-    _so_name = "LBBoundaries::LBBoundaries"
+if any(i in espressomd.code_info.features() for i in ["LB_BOUNDARIES", "LB_BOUNDARIES_GPU"]):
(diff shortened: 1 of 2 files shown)


src/python/espressomd/lbboundaries.py
```diff
@@ -1,25 +1,27 @@
 from __future__ import print_function, absolute_import
 from .script_interface import ScriptInterfaceHelper
+import espressomd.code_info
 
 
-class LBBoundaries(ScriptInterfaceHelper):
-    _so_name = "LBBoundaries::LBBoundaries"
+if any(i in espressomd.code_info.features() for i in ["LB_BOUNDARIES", "LB_BOUNDARIES_GPU"]):
```
hmenke, line 6: How about throwing an `ImportError` instead of `if`-ing the whole class definition (which apparently fails the checks)?
````python
if not set(["LB_BOUNDARIES", "LB_BOUNDARIES_GPU"]) < set(espressomd.code_info.features()):
    throw ImportError("LB_BOUNDARIES not compiled in!")

class LBBoundaries(ScriptInterfaceHelper):
    ...
````

KaiSzuttor: The logic of your check is not quote right since you would have to have both features compiled in, if I'm getting it right. The check only fails because the system class does not check if the boundaries are compiled in.

fweik (approved):

PR #1087, opened 2017-04-13
Electromagnetic constraints feature

bogdan-tanygin:
Description of changes:
 - External electric and magnetic field have been added via constraints. Previously (in the `Tcl Espresso`), similar functionality was available via `constraint ext_magn_field Bx By Bz`;
- The functionality is physically fully correct only for a `Wall` constraint. For other shapes, the present feature is introducing external fields normal to the closest constraint surface in a correspondent vicinity. Corresponding "Warning" has been added to the Python file. Even though such field spatial distribution has no physical sense (for a non-`Wall` case) it can be useful for tests of complex field geometries (based on `Shape`s complexity and their arbitrary spatial combination) or in cases of some specific physical models;
- Other option: to introduce this feature via the `Wall` class with a zero function in a `Shape` virtual method for other shapes. However, this approach contradicts to present architecture: field and energy are calculated by the `Constraint`, not by `Shape`, cause shape, after all, is just an abstract shape. Please, provide your feedback: how external electric and magnetic field should be applied in current Espresso architecture? Maybe some `ExternalField` class should be defined fully separate to `Constraint` and `Shape` and introduced additionally to fields, energies and torques?
- Suggested test validates electric and magnetic constraints in a real physical task with DDS CPU + MMM2D inter-particle interactions and external field which drives the particles aggregate. The test finally validates that particles aggregate drift as a single entity;
- If overall approach will be approved then I will supplement docs to this pull request for a user guiding; 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?

### DIFF
--- a/src/core/constraints/Constraint.cpp
+++ b/src/core/constraints/Constraint.cpp
@@ -59,7 +59,7 @@ void Constraint::reflect_particle(Particle *p, const double *distance_vector,
 void Constraint::add_force(Particle *p, double *folded_pos) {
   double dist, vec[3], force[3], torque1[3], torque2[3];
   Particle part_rep;
-  Vector3d vec_dip({0., 0., 0.}), vec_field({0., 0., 0.}), torque_3d({0., 0., 0.});
+  Vector3d vec_dip({0., 0., 0.}), vec_field({0., 0., 0.}), torque_ext({0., 0., 0.});
   part_rep.p.type = m_type;
 
   IA_parameters *ia_params = get_ia_param(p->p.type, part_rep.p.type);
@@ -69,7 +69,7 @@ void Constraint::add_force(Particle *p, double *folded_pos) {
     force[j] = 0;
     vec_field[j] = 0;
 #ifdef ROTATION
-    torque1[j] = torque2[j] = torque_3d[j] = 0;
+    torque1[j] = torque2[j] = torque_ext[j] = 0;
 #endif
   }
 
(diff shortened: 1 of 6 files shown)


mkuron:
This seems to conceptually duplicate parts of #1071, which was done to allow for magnetic fields as constraints and seems to have the more elegant design.

RudolfWeeber:
A more general solution is being implemented (#1050) which allows for constraints not depending on a shape. Based on that external fields will follow.

The first part, splitting shape-based and non-shape-based constraints,. is already an open pr (#1071).

To my knowledge, work on external potentials based on grid data is also underway (@konradbreitsprecher)

bogdan-tanygin:
Yes, I've missed existing pull request.. Then, only the test proposed here could be useful later on after the pull request #1071 merging. The test will require minor adaptations, I could open separate pull request for it later. Propose to close current PR.

bogdan-tanygin:
OK, I see. The PR #1071 just splits shape-based and non-shape-based constraints. However, an external electric and magnetic field could still be implemented based on former and/or latter ones. An uniform field case is much more simpler (non-shape-based).

> To my knowledge, work on external potentials based on grid data is also underway (@konradbreitsprecher)

@KonradBreitsprecher could you please advise, do you work on electric/magnetic fields for shape-based or non-shape-based constraints? Could I implement a simple non-shape-based uniform case based on the PR #1071 architecture in the same way as it is done here in the present PR #1087: just a spatially uniform constant field without any special interactions methods? Or latter approach contradicts to your design?

RudolfWeeber:
A homogenewous magnetic field was implemented based on #1071 (@pkreissl). I assume, there will be a pr soon.

A homogeneous electric field can be modelled using the ext_force property of the particles.

Grid based constraints would be used for inhomogeneous fields of all kinds.

KonradBreitsprecher:
- I think any interaction with shape-based constraints should be introduced via a non-bonded interaction, not directly in `Constraint::add_force`.
- Based on recent discussions, we agreed on implementing tabulated external potentials and fields via the more general concept of a _LatticeCoupling_ class, which will be a non-shaped-based constraint. It can be populated with just a vector field for arbitrary force coupling or with a potential and gradient for non-rotational fields.
- When we have _LatticeCoupling_, I will also add _MeshCapacitor_. It aims for arbitrary electrode shapes with applied potential (I use it for capacitor systems). It takes electrode geometries from a stl-file (triangles+normals) and allows to set the applied potential for these electrodes. Then it solves Laplace equation and generates the potential grid data. The stl geometry is also used to place fixed ICC particles for the actual electrode boundary. With p3m and ICC, metallic boundary condition (constant potential) is maintained on the electrode particles. Finally the applied potential is superimposed with the generated vacuum solution via _LatticeCoupling_.

bogdan-tanygin:
Dear @KonradBreitsprecher and @RudolfWeeber ,
Thank you for the advices and clarifications. It will be very interesting to use the _LatticeCoupling_ and _MeshCapacitor_ classes. Meanwhile, for the simplest tasks of spatially homogeneous external fields, we've utilized the `ext_force` and `ext_torque` particle properties. Corresponding energy could be calculated directly in the user Python script.

PR #1089, opened 2017-04-16
Sphinx-ified Developer's Guide

hmenke:
Ported the developer's guide to Sphinx using Pandoc.  Some handwork was done to correct the hyperlinks.


fweik:
Not sure if it would be a better idea to start from scratch. The information in the DG is almost completely outdated.

KaiSzuttor:
Maybe put a warning in there notifying a user that the information is outdated, but maybe it's still a good starting point for s.o. (in the sense of a lower barrier to even start) to write the DG.

hmenke:
@KaiSzuttor Warning added.

@fweik Starting from scratch is inevitable but the current DG already gives a nice outline which can be adapted as is.

KaiSzuttor (approved):
Looks good to me, can be merged.

Issue #624, opened 2016-05-23
Consolidate integration loops integrate_vv() and integrate_sd() etc.


RudolfWeeber:

KaiSzuttor:
@dschwoerer  maybe can comment on that?

PR #1094, opened 2017-04-19
Remove SD

### LINKED ISSUE #233: Remove code duplication in integrate function
Integration loop is duplicated between integrate.cpp (velocity verlet/langevin, stepest descend) and integrate_sd.cpp (BD/SD). This will break within weeks. Needs some thoughts because one and two step integrators. Maybe new hooks (on_...).

### LINKED ISSUE #1091: Decide faith of Stokesian Dynamics (SD)
I'm pretty sure nobody ever used this.  It is not part of any predefined feature set (like `maxset.hpp`), has no tests, no Python interface, and it is unkown whether this even compiles.  Please remove.

hmenke:
Fixes #233, #624, and #1091


RudolfWeeber:
Please, also for this one, wait for the next Espresso meeting

fweik (changes requested):
Requires offline discussion

hmenke:
I heard that in the meeting a few weeks ago SD was scheduled for removal.  What is blocking the merge of this one?

fweik:
The duplicated main loop that comes with SD is very hard to maintain and has already diverged quiet a bit from the other stuff. If @dschwoerer or somebody else is willing to integrate the method with the normal main loop I will remove it.

dschwoerer:
I am slightly sad to see SD go. I am not using espresso any more, and if there is nobody else who uses it ...

If there would be use for the SD/BD integrators, I could be bothered, but I don't think that this is currently the case ...

hmenke:
Can this be merged?

fweik (approved):

fweik:
We would also be more happy if we had SD, but we can not fix it at the moment. The code is not lost, and if we find a student in the future we will pick this up again.

PR #1096, opened 2017-04-21
seed with 32bit ints

### LINKED ISSUE #1046: Using set_random_state_PRNG leads to stucked script.
After using the system.set_random_state_PRNG(), adding an LB fluid does not work.
[lbf.py.zip](https://github.com/espressomd/espresso/files/798676/lbf.py.zip)

 method adding

davidsean:
Fixes #1046 

Description of changes:
 - 
Use int32 instead of sys.maxint which can sometimes be 64 bits.

Seems to fix the (#1046) problem for me, my understanding is that this needs to be 32 bits anyways, see
http://www.cplusplus.com/reference/random/seed_seq/

I think there is still another issue concerning a 625 length vector, (instead of 626) but that's for another day. 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?

### DIFF
--- a/src/python/espressomd/_system.pyx
+++ b/src/python/espressomd/_system.pyx
@@ -245,7 +245,7 @@ cdef class System:
         for i in range(n_nodes):
             states_on_node_i = []
             for j in range(_state_size_plus_one + 1):
-                states_on_node_i.append(rng.randint(0, numeric_limits[int].max))
+                states_on_node_i.append(rng.randint(0, numeric_limits[int].max()))
             states[i] = " ".join(map(str, states_on_node_i))
         mpi_random_set_stat(states)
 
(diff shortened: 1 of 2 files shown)


src/python/espressomd/_system.pyx
```diff
@@ -245,7 +245,7 @@ cdef class System:
         for i in range(n_nodes):
             states_on_node_i = []
             for j in range(_state_size_plus_one + 1):
-                states_on_node_i.append(rng.randint(0, sys.maxint))
+                states_on_node_i.append(rng.randint(0, 2**31-1))
```
hmenke: Better extend `numeric_limits[T]` in `utils.pxd` by `max()`, i.e.
````python
cdef extern from "<limits>" namespace "std" nogil:
    cdef cppclass numeric_limits[T]:
        @staticmethod
        T epsilon()
        @staticmethod
        T max()
````
and then use
````
numeric_limits[int].max
````
This will translate to C++ code `std::numeric_limits<int>::max()` which will adjust to the `int` chosen by the compiler.

hmenke (changes requested):

hmenke (approved):

PR #1098, opened 2017-04-24
fix for lbboundaries force interface + test case

KunTares:
LBoundary scriptinterface compiles again. Flo still has to implement LBoundary get_force.

PR Checklist
------------
 - [x] Tests?
   - [x] Interface
   - [x] Core 
 - [ ] Docs?


fweik:
Rebased to #1099

PR #1100, opened 2017-05-03
Script interface class registration

RudolfWeeber:
This implements a mapping between script object names and corresponding python classes
The purpose is that object-typed parameters are returned as an instance of the correct python class

Consider a constraint
c=ShapeBasedConstraint(shape=Wall())
right now,
c.shape is of type PScriptInterface
After this is merged, it is of type Wall

Implementation:
* decorator script_interface_register stores the script object name and python class in a map at module-level in script_interface.py
* the varitant to python converter retrieves the class based on the script object name and creates the instance

Usage:
@script_interface_register
class Wall(ScriptInterfaceHelper):
    _so_name = "Shape::Wall"
i.e., the only new thing is the use of the decorator

Decorators were added to all (public) uses of script objects, i.e., constraints, observables,correlators,lbboundaries,ekboundaries

Test included

This also fixes the python comparisson between script objects


RudolfWeeber:
Blocked by #1101

hmenke (approved):

RudolfWeeber:
Block by 1101 no longer applies,
but now it will make half the sphinx documentation vanish, until #1154 is solved

src/python/espressomd/script_interface.pyx
```diff
@@ -1,4 +1,14 @@
-from espressomd.utils import to_char_pointer
+from espressomd.utils import to_char_pointer,to_str
+
+cdef class PObjectId:
```
mkuron: Should be new-style class: `cdef class PObjectId(object)`.

PR #1160, opened 2017-05-08
Dynamic properties for ParticleSlice

### LINKED ISSUE #958: ParticleSlice lacks support for particle properties
This should be resolved by a generic implementation of __setattr__ and __getattr__ in the ParticleSlice class
If ParticleSlice does not have an implementation for the given property, 
The getter and seetter should call the property with the given attribute name of the ParticleHandle class for all particles in the slice.
The getter additionally has to determine the number of values for the particle property by obtaining the result for the first particle in the slice and then creating the approprate numpy array.

Those properties of ParticleSlice already there, which only call properties of ParticleHandle should be removed, because it's lots of copied code.
The properties of ParticleSlice directly calling core methods should probably stay, because they might offer better performance.

mkuron:
Fixes #958.

@RudolfWeeber @fweik @hmenke

Any properties that are not defined for `ParticleSlice` are automatically drawn from `ParticleHandle`, including docstrings. To the setter, one can either provide one value that is set on all particles or a list of values where the order corresponds to that of the slice members.

This turned out to be surprisingly difficult:
- One cannot use `setattr` to add stuff to Cython classes, so I ended up wrapping the Cython class into a Python class that has all the autogenerated properties.
- The wrapping cannot be done in the loop I added to the end of particle_data.pyx. I had to introduce a class that holds the attribute name as a member. If I generated the functions directly in the loop (i.e. capturing the loop variable), the attribute name a setter/getter was operating on would vary in a non-deterministic fashion (seems like some kind of reference counting bug in Python itself...).

Further improvements made along the way:
 - Fix `ParticleSlice.__str__`, which previously didn't return anything as `int` is not the same as `numpy.int64`.
-  Improve `ParticleSlice.__str__` and `ParticleList.__str__` so that the representations look like Numpy arrays or Python lists. Previously there were newlines and the name was missing.
- Improve `utils.check_type_or_throw_except` so that it also allows `numpy.int64` where `int` is requested.

PR Checklist
------------
 - [x] Tests?
   - [x] Interface
   - [x] Core

### DIFF
--- a/src/python/espressomd/particle_data.pyx
+++ b/src/python/espressomd/particle_data.pyx
@@ -1148,10 +1148,11 @@ cdef class ParticleHandle:
             setattr(self, k, P[k])
 
 
-cdef class ParticleSlice:
+cdef class ParticleSliceImpl:
     """
     Handles slice inputs e.g. part[0:2]. Sets values for selected slices or returns values as a single list.
 
+    This is the base class that should not be used directly as it lacks some properties. Use :class:`espressomd.ParticleSlice` instead.
     """
 
     def __cinit__(self, slice_):
@@ -1479,6 +1480,13 @@ cdef class ParticleSlice:
             ParticleHandle(id).remove()
 
 
+class ParticleSlice(ParticleSliceImpl):
+    """
+    Handles slice inputs e.g. part[0:2]. Sets values for selected slices or returns values as a single list.
+
+    """
+    pass
+
 cdef class ParticleList:
     """
     Provides access to the particles via [i], where i is the particle id. Returns a ParticleHandle object.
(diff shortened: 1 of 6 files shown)


fweik (approved):

KaiSzuttor (changes requested):
Maybe add some more in-code documentation for future developers.

mkuron:
@KaiSzuttor, I'm currently refactoring `_InjectParticleProperty` to make it easier to understand. That should address your documentation request.

src/python/espressomd/particle_data.pyx
```diff
@@ -1430,9 +1219,9 @@ cdef class ParticleSlice:
         pl = ParticleList()
         for i in self.id_selection:
             if pl.exists(i):
-                res += str(pl[i]) + "\n"
-        # Remove final newline
-        return res[:-1]
+                res += str(pl[i]) + ", "
+        # Remove final comma
+        return "ParticleSlice([" + res[:-2] + "])"
```
hmenke, line 1225: No need for this awkward comma removal
````python
pl = ParticleList()
return "ParticleSlice([" + ", ".join(str(pl[i]) for i in self.id_selection if pl.exists(i)) + "])"
````

File: src/python/espressomd/particle_data.pyx [hunk 2]
```diff
@@ -1640,9 +1436,9 @@ cdef class ParticleList:
         res = ""
         for i in range(max_seen_particle + 1):
             if self.exists(i):
-                res += str(self[i]) + "\n"
-        # Remove final newline
-        return res[:-1]
+                res += str(self[i]) + ", "
+        # Remove final comma
+        return "ParticleList([" + res[:-2] + "])"
```
hmenke, line 1442: No need for this awkward comma removal
````python
return "ParticleList([" + ", ".join(str(self[i]) for i in range(max_seen_particle + 1) if self.exists(i)) + "])"
````

File: src/python/espressomd/particle_data.pyx (hunk 1 above)
Comment (mkuron) [reply src/python/espressomd/particle_data.pyx:1225]: It's nicer, and it's more in line with how Python and Numpy print their own data types. Also, if I remove that space at the end, why not remove the trailing comma along with it.

**Comment by mkuron** [comment]
I was able to get rid of the injector class by switching to `functools.partial`. It seems like the following two are different here, even though they look like they do the same:
```
functools.partial(geta, attribute=attribute_name)
lambda s: geta(s, attribute_name)
```
Explicitly capturing the loop variable like this
```
def geta(particle_slice, attribute=attribute_name):
```
didn't help either, even though that is the standard solution for this problem in pure Python. I really don't understand what is going on here, might be a Cython bug.

Anyway, once Travis passes, I think this can be merged.

hmenke (approved):

mkuron:
Small change to the getter so it writes straight to a numpy array instead of converting a list into one at the end. Also extend unit test so it covers all code paths.

hmenke (approved):

PR #1167, opened 2017-05-09
unit tests for VSF bonds and atom declaration

davidsean:
Fixes #
slightly more complete tests, a bit silly but at least it forces these functions to be used by Travis
+ a small fix from moving the writevsf function to io/writer

Description of changes:
checks if the declared bonds are written by writevsf
checks if the atoms declarations are written by writevsf  
 - 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


test/test.py
```diff
@@ -0,0 +1,45 @@
+
```
hmenke: Are you sure this file should go into the `git` repo?

testsuite/python/writevtf.py
```diff
@@ -43,64 +44,112 @@ class CommonTests(ut.TestCase):
...
+
+
 class VCFTestAll(CommonTests):
     """
     Test the writing VTF files.
     """
     @classmethod
     def tearDownClass(cls):
-        os.remove("test.vtf")
+        os.remove("test.vcf")
```
hmenke: Please use [`tempfile`](https://docs.python.org/3/library/tempfile.html) rather than creating and removing files in the working directory.  This might be dangerous.

hmenke (changes requested):

test/test.py
davidsean: Funny, don't know how it ended up here. I think I mixed up two branches earlier...

hmenke (approved):

PR #1175, opened 2017-05-11
Implement iterators over cells

hirschsn:
Following up the ParticleIterator, I implemented an iterator over cells and cell neighborhood. They are supposed to abstract away loops over local_cells and dd.cell_inter. Since these kinds of loops are everywhere in the code, an abstraction is definitely useful. Implemented features are:
- CellProxy and IndexedCellProxy encapsulate cells (and their indices in local_cells)
  and allow to iterate over the particles in a cell.
- CellNeighIterator provides an iterator over dd.cell_inter.
- CellIterator iterates over local_cells or ghost_cells.
- Extend ParticleIterator to be able to sell it as a Particle*

As an example, I show how calc_link_cell() could be transformed (which assumes another function ParticleIterator::start_after which is trivial to implement, just copy m_part_id and increment it once):
```
void new_calc_link_cell()
{
  for (auto c1: local_cells.cells()) {
    for (auto c2: c1.hsneigh()) {
      for (auto p1 = c1.pbegin(); p1 != c1.pend(); p1++) {
        auto p2 = c2.pbegin();

        if (c1 == c2) {
          // Within cell itself: bonded forces
          add_single_particle_force(p1);
          if (rebuild_verletlist)
            std::copy(p1->r.p, p1->r.p + 3, p1->l.p_old);

          // Half-Shell within a cell itself
          p2.start_after(p1);
        }
        for (; p2 != c2.pend(); p2++) {
#ifdef EXCLUSIONS
          if (do_nonbonded(p1, p2))
#endif
          {
            double vec21[3];
            double dist2 = distance2vec(p1->r.p, p2->r.p, vec21);
            add_non_bonded_pair_force(p1, p2, vec21, sqrt(dist2), dist2);
          }
        }
      }
    }
  }
  rebuild_verletlist = 0;
}
```

Using these abstractions to change calc_link_cell is probably not a good idea before measuring what the cost of them are. Runtime uncritical code could definitely benefit from abstractions.

Things which can (and should) be improved:
- CellProxy and IndexedCellProxy: Two classes are needed because for neighbors cells we don't have an index to local_cells
- Currently, CellPList::cells() allows to call hsneigh() on ghost_cells which is not desirable.
- ...

### DIFF
--- a/src/core/ParticleIterator.hpp
+++ b/src/core/ParticleIterator.hpp
@@ -16,6 +16,10 @@ struct ParticleIterator : public boost::iterator_facade<
     }
   }
 
+  operator Particle *() const {
+    return &((*m_cell)->part[m_part_id]);
+  }
+
 private:
   friend class boost::iterator_core_access;
 
--- a/src/core/ParticleIterator.hpp
+++ b/src/core/ParticleIterator.hpp
@@ -3,10 +3,29 @@
 
 #include <boost/iterator/iterator_facade.hpp>
 
-template <typename BidirectionalIterator, typename Particle>
+namespace __particle_iterator_util {
+template <typename ReturnType, typename It, typename Particle>
+struct CellDeref;
+
+template <typename It, typename Particle>
+struct CellDeref<Particle&, It, Particle> {
+  static inline Particle& get_particle(It m_cell, int m_part_id) {
+    return (*m_cell)->part[m_part_id];
+  }
+};
+
+template <typename It, typename Particle>
+struct CellDeref<Particle*, It, Particle> {
+  static inline Particle* get_particle(It m_cell, int m_part_id) {
+    return &((*m_cell)->part[m_part_id]);
+  }
+};
+}
+
+template <typename BidirectionalIterator, typename Particle, typename DerefType = Particle&>
 struct ParticleIterator : public boost::iterator_facade<
-                              ParticleIterator<BidirectionalIterator, Particle>,
-                              Particle, boost::forward_traversal_tag> {
+                              ParticleIterator<BidirectionalIterator, Particle, DerefType>,
+                              Particle, boost::forward_traversal_tag, DerefType> {
   ParticleIterator(BidirectionalIterator cell, BidirectionalIterator end,
                    int part_id)
       : m_cell(cell), m_end(end), m_part_id(part_id) {
@@ -14,10 +33,6 @@ struct ParticleIterator : public boost::iterator_facade<
     if ((m_cell != m_end) && (*m_cell)->n == 0) {
       increment();
     }
-  }
-
-  operator Particle *() const {
-    return &((*m_cell)->part[m_part_id]);
   }
 
 private:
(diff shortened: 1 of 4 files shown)


src/core/ParticleIterator.hpp
```diff
@@ -16,6 +16,10 @@ struct ParticleIterator : public boost::iterator_facade<
     }
   }
 
+  operator Particle *() const {
```
hmenke: Why should the iterator behave like a pointer to `Particle`?  Usually you dereference an iterator to get the contents.

hmenke (changes requested):
So far this PR only implements the cell iterators.  They are not used anywhere.  Why should we merge unused code?

src/core/ParticleIterator.hpp
hirschsn: ParticleIterator has already an operator* which returns Particle&. You cannot overload it by return type. ESPResSo, however, uses functions which take pointers to particles everywhere (e.g. above: do_nonbonded, add_single_particle_force, ...).
But you are right. It would be nicer if you could iterate over the pointers themselves by dereferencing the iterator. This requires another iterator class which dereferences to Particle*.

hirschsn:
I wanted to show the code here and discuss its usefulness. If it is considered useful than we can go ahead an use it.

src/core/ParticleIterator.hpp
hmenke: How about a `get()` function (or `ptr()` if you want)?  If you don't think these suit you, then please at least make the conversion `explicit`.  Otherwise the iterator might decay to a `Particle*` in unexpected places.

hirschsn:
I have updated the code:
- ParticleIterator now takes a the dereference return type (Particle* or Particle&) as template parameter.
- The Iterators now have associated exactly one cell proxy object which they update. For dereference, they return const& to it. This avoids the creation of cell proxy objects for each dereference.

The updated version of new_calc_link_cell would now look more natural due to the iterator changes:
```
void new_calc_link_cell()
{
  for (auto c1: local_cells.cells()) {
    for (auto c2: c1.hsneigh()) {
      int halfshellbegin = 0;
      for (auto p1: c1.particle_ptrs()) {
        if (c1 == c2) {
          // Within cell itself: bonded forces
          add_single_particle_force(p1);
          if (rebuild_verletlist)
            std::copy(p1->r.p, p1->r.p + 3, p1->l.p_old);

          // Half-Shell within a cell itself - count the number of particles
          // to start after p1
          halfshellbegin++;
        }
        for (auto p2: c2.particle_ptrs(c1 == c2? halfshellbegin: 0)) {
#ifdef EXCLUSIONS
          if (do_nonbonded(p1, p2))
#endif
          {
            double vec21[3];
            double dist2 = distance2vec(p1->r.p, p2->r.p, vec21);
            add_non_bonded_pair_force(p1, p2, vec21, sqrt(dist2), dist2);
          }
        }
      }
    }
  }
  rebuild_verletlist = 0;
}
```

hirschsn:
Made obsolete by PR #1197.

Issue #1127, opened 2017-05-05
thermostat.pyx


RudolfWeeber:

jonaslandsgesell:
I have a look on it

jonaslandsgesell:
Documentation is there. This can be closed. There is only the problem with the decorators which break documentation, see #1154

mkuron:
Please close, fixed by #1176.

PR #1176, opened 2017-05-11
Use new-style classes everywhere

### LINKED ISSUE #1154: Sphinx documentation lost for decorated classes
Will affect all script_interface based classes after #1100 is merged

mkuron:
@RudolfWeeber, please give this a try. If it works, you can proceed with #1100.

Fixes #1154 and #1127.

### DIFF
--- a/maintainer/travis/build_cmake.sh
+++ b/maintainer/travis/build_cmake.sh
@@ -71,6 +71,8 @@ else
     echo ""
     exit $ec
 fi
+# enforce style rules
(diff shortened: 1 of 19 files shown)


hmenke:
We could also run Sphinx with Python3 where new-style classes are the default.

mkuron:
We could, but there's no reason to keep a random subset of our classes old-style when using Python 2.7. After all, they have been deprecated since Python 2.2, released 15 years ago... Python 3 automatically upgrades everything to new-style. For consistency between Python 2.7 and 3, all classes should be new-style, perhaps we could even enforce that.

mkuron:
Fix verified with @RudolfWeeber, please merge along with #1100.

PR #1186, opened 2017-05-23
[python] Convert non-list types to lists in has_features

### LINKED ISSUE #1185: espressomd.has_featues silently fails when passing a single feature
has_features("EXCLUSIONS") -> False
has_features(("EXCLUSIONS",)) -> True

In my opinion, the 1st syntax should be supported. If it isn't, it should throw an exception.

hmenke:
Fixes #1185

### DIFF
--- a/src/python/espressomd/__init__.py
+++ b/src/python/espressomd/__init__.py
@@ -27,20 +27,26 @@ espressomd._init.setup()
 from espressomd.system import System
 from espressomd.code_info import features
 
-def has_features(arg):
+def has_features(*args):
     """Tests whether a list of features is a subset of the compiled-in features"""
 
-    return set(arg) < set(features())
+    if len(args) == 1 and type(args[0]) is not str and hasattr(args[0], "__iter__"):
+        return set(args[0]) < set(features())
+
+    return set(args) < set(features())
 
 
-def missing_features(arg):
+def missing_features(*args):
     """Returns a list of the missing features in the argument"""
 
-    return list(set(arg) - set(features()))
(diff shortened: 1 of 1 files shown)


mkuron (changes requested):
Better use `type(arg) is not str and hasattr(arg, "__iter__")` to check whether the argument is iterable. Only comparing against known containers (`tuple`, `list`) doesn't cover if somebody uses `set` or `numpy.ndarray(..., dtype=str)`.

mkuron:
Also, while you're at it: could you change it to use `*args` so that the following works too: `has_features("EXCLUSIONS", "ELECTROSTATICS")`? This would remove the need for brackets.

hmenke:
@mkuron These fixes should capture everything.

src/python/espressomd/__init__.py
```diff
@@ -27,20 +27,26 @@
 from espressomd.system import System
 from espressomd.code_info import features
 
-def has_features(arg):
+def has_features(*args):
     """Tests whether a list of features is a subset of the compiled-in features"""
 
-    return set(arg) < set(features())
+    if len(args) == 1 and type(args[0]) is not str and hasattr(args[0], "__iter__"):
+        return set(args[0]) < set(features())\
```
mkuron: unnecessary backslash at the end

mkuron (approved):
Looks good, except for that one unneeded backslash.

src/python/espressomd/__init__.py
hmenke: Well, English keyboard...

PR #1188, opened 2017-05-29
script_interface: Make ScriptInterfaceHelper forward the creation pol…

fweik:
…icy to PScriptInterface. Create observables and correlators only on the master as they should.

I'm not totally happy with the way this is done, but this is what we already started....

Description of changes:
 - Let objects derived from ScriptInterfaceHelper decided whether they should be created on
  all nodes, or only the caller.
- Make observabled and correlaters local.

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?

### DIFF
--- a/src/python/espressomd/script_interface.pyx
+++ b/src/python/espressomd/script_interface.pyx
@@ -198,9 +198,10 @@ cdef class PScriptInterface(object):
 class ScriptInterfaceHelper(PScriptInterface):
     _so_name = None
     _so_bind_methods =()
+    _policy = "GLOBAL"
 
-    def __init__(self,**kwargs):
-        super(ScriptInterfaceHelper,self).__init__(self._so_name)
+    def __init__(self, **kwargs):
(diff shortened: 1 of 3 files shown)


RudolfWeeber:
Functionallity-wise this is fine.
The other script object related class level variables are prefixed with _so_. Should it be _so_policy or _so_creation_policy?

hmenke (approved):

src/python/espressomd/script_interface.pyx
```diff
@@ -198,10 +198,10 @@ cdef class PScriptInterface(object):
 class ScriptInterfaceHelper(PScriptInterface):
     _so_name = None
     _so_bind_methods =()
-    _policy = "GLOBAL"
+    _so_creation_policy = "GLOBAL"
```
hmenke: Should they really all default to global?

fweik:
Yes, this was the default before.

PR #1190, opened 2017-05-31
New implementation of partCfg

### LINKED ISSUE #362: Replace partcfg[] by a hash-map
The 
Particle\* partcfg 
data structure is used for non-parallelized analysis routines.
This is right now an array. 
For some algorithms (notably virtual sites at center of mass), and a few analysis routines, the index of the array has to be the particle's id. 
This is accomplished by sortPartCfg(). 
There are two problems:
- The sorting only works with consecutive particle numbering
- During the "sorting", the particle data exists twice in memory, in addition to the particles stored in the cells of the domain decomposition. For systems with 10000 particles, this is a problem.

Both problems would go away, if partcfg[] was a hash map with particle id as key and the particle data structure as value.

fweik:
Fixes #362

Description of changes:
 - Replace Particle **partCfg by an implementation
   that can reasonably accommodate holes in the particle ids.
 - Updates of the Particle data are now implicit on access,
 - bond info is still only updated on request.
 - Removed GHOST_FLAG feature, the property is now always
   there as this does not affect the size due to padding.

PR Checklist
------------
 - [x] Tests?
   - [x] Core 
 - [x] Docs?


RudolfWeeber:
````
diff --git a/src/core/ParticleCache.hpp b/src/core/ParticleCache.hpp
new file mode 100644
index 000000000..ad5f9b601
--- /dev/null
+++ b/src/core/ParticleCache.hpp
@@ -0,0 +1,378 @@
[...]
+  Merge(Compare &&comp = Compare{}) : m_comp(comp) {}
+  Container operator()(Container const &a, Container const &b) const {
+    Container ret;
+    ret.reserve(a.size() + b.size());
+
+    auto first1 = a.begin();
+    auto last1 = a.end();
+    auto first2 = b.begin();
+    auto last2 = b.end();
+
+    while (first1 != last1) {
+      if (first2 == last2) {
+        /* The first range has no more elements, so we can
+           just copy the rest of range 2. */
````
The comment should refer to 2nd range, or did I miss th?
````
+class ParticleCache {
+  using map_type = boost::container::flat_set<Particle, detail::IdCompare>;
+
+  std::unordered_map<int, int> id_index;
+  map_type remote_parts;
+  std::vector<int> bond_info;
+  bool m_valid, m_valid_bonds;

+
+  Utils::Parallel::Callback update_cb;
+  Utils::Parallel::Callback update_bonds_cb;
+
+  GetParticles parts;
+  UnaryOp m_op;
````
Please put comments explaining what above members contain
````
+
+  /**
+   * @brief Implementation of bond update.
+   *
+   * Particle ids, and the bond info is packed
+   * into a linear buffer and gathet to the master.
+   */
+  void m_update_bonds() {
+    std::vector<int> local_bonds;
+
+    for (auto &p : parts()) {
+      local_bonds.push_back(p.identity());
+      std::copy(p.bl.begin(), p.bl.end(), std::back_inserter(local_bonds));
+    }
+
+    Utils::Mpi::gather_buffer(local_bonds,
+                              Communication::mpiCallbacks().comm());
+  }
+
+  /**
+   * @brief Implementation of bond update.
+   *
+   * Master version of m_update_bonds().
+   */
+  void m_recv_bonds() {
+    bond_info.clear();
+
+    Utils::Mpi::gather_buffer(bond_info, Communication::mpiCallbacks().comm());
+
+    for (auto it = bond_info.begin(); it != bond_info.end();) {
+      auto &p = remote_parts.begin()[id_index[*it++]];
+      p.bl.e = &(*it);
+      p.bl.max = p.bl.size();
+
+      it += p.bl.size();
+    }
````
Please insert a comment explaining what above loop does
````
+  }
+
+  /**
+   * @brief Actual update implementation.
+   *
+   * This gets a new particle range, packs
+   * the particles into a buffer and then
+   * merges these buffers hierachicaly to the
+   * master node
+   */
+  void m_update() {
+    remote_parts.clear();
+
+    for (auto const &p : parts()) {
+      typename map_type::iterator it;
+      /* Add the particle to the map */
+      std::tie(it, std::ignore) = remote_parts.emplace(p);
````
What's the role of the std::tie here?
````
diff --git a/src/core/partCfg.hpp b/src/core/partCfg.hpp
new file mode 100644
index 000000000..d6dcc0cce
--- /dev/null
+++ b/src/core/partCfg.hpp
@@ -0,0 +1,51 @@
[...]
+#ifndef CORE_PART_CFG_HPP
+#define CORE_PART_CFG_HPP
+
+#include <boost/iterator/indirect_iterator.hpp>
+
+#include "ParticleCache.hpp"
+#include "cells.hpp"
+#include "grid.hpp"
+#include "particle_data.hpp"
+#include "utils/Range.hpp"
+#include "utils/SkipIterator.hpp"
+#include "utils/serialization/Particle.hpp"
+
+/**
+ * @brief Proxy class that gets a particle range from
+ *        from the global local_particles.
+ */
+class GetLocalParts {
+  class SkipIfNull {
+  public:
+    bool operator()(Particle const *p_ptr) const {
+      return (p_ptr == nullptr) or (p_ptr->l.ghost);
+    }
+  };
````
Should be SkifIfNullOrGhost, probably
````
+
+  using skip_it = Utils::SkipIterator<Particle **, SkipIfNull>;
+  using iterator = boost::indirect_iterator<skip_it>;
+  using Range = Utils::Range<iterator>;
+
+public:
+  Range operator()() const {
+    auto begin = skip_it(local_particles,
+                         local_particles + max_seen_particle + 1, SkipIfNull());
+    auto end = skip_it(local_particles + max_seen_particle + 1,
+                       local_particles + max_seen_particle + 1, SkipIfNull());
````
What happens if the last particle in local_particles is a ghost and therefore skipped? Shouldn't the end be decreased then?
````
diff --git a/src/core/utils/SkipIterator.hpp b/src/core/utils/SkipIterator.hpp
new file mode 100644
[...]
+template <typename ForwardIterator, typename Predicate,
+          typename ValueType =
+              typename std::iterator_traits<ForwardIterator>::value_type>
+class SkipIterator : public boost::iterator_facade<
````
Please put doxygen comment explaining what this class does.

**Comment by RudolfWeeber** [comment]
Since travis does not do it automatically right now, did you run funcitonal tests depending on partcfg in parallel?
For example observables.py

**Comment by fweik** [comment]
Yes, I tested correlator, observable and h5md in parallel.

**Comment by fweik** [comment]
> What's the role of the std::tie here?

The return value of insert is a pair<Iterator, bool>, since only the first one is needed, std::tie is used to unpack it, using std::ignore to ignore the bool value.

**Comment by fweik** [comment]
> What happens if the last particle in local_particles is a ghost and therefore skipped? Shouldn't the end be decreased then?

I think the behaviour is correct. The case you are describing is checked in the unit test: In the "gap" test, the base range ends with rejected values.

File: src/core/utils/SkipIterator.hpp [hunk 1]
```diff
@@ -21,6 +31,7 @@ class SkipIterator : public boost::iterator_facade<
   SkipIterator(ForwardIterator const &it, ForwardIterator const &end,
                Predicate const &pred)
       : m_it(it), m_end(end), m_p(pred) {
+    /* increment until the first element that is not skipped. */
```
hmenke: You do `increment` recursively in a much more concise fashion.
````cpp
void increment() {
  ++m_it;
  if ( (m_it == m_end) || (!m_p(*m_it)) ) return;
  increment();
}
````

hmenke (approved):

RudolfWeeber:
Can be merged.

PR #1191, opened 2017-06-01
Remove explicit cells loop

fweik:
Description of changes:
 -  Removed for(cells) for(particles) loops by for(local_cells.particle()) where
    this was possible.

The remaining occurrence are mostly related to pair iteration or in the cell systems.
These are in maggs, mmm2d, iccp3m, collision, reaction, force, pressure, energy
and the cell systems.

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


src/core/forces.cpp
```diff
@@ -26,105 +26,68 @@
...
      set torque to zero for all and rescale quaternions
   */
-  for (c = 0; c < local_cells.n; c++) {
-    cell = local_cells.cell[c];
-    p  = cell->part;
-    np = cell->n;
-    for (i = 0; i < np; i++)
-      init_local_particle_force(&p[i]);
+  for (auto &p : local_cells.particles()) {
+    init_local_particle_force(&p);
```
hmenke, line 53: (Applies to many other instances of this pattern in the changeset)
````cpp
std::for_each(std::begin(local_cells.particles()),
              std::end(local_cells.particles()),
              [](Particle &p){ init_local_particle_force(&p); });
````
The advantage is that with C++17 you can add an execution policy to this, possibly performing the unary function out of order.

File: src/core/lees_edwards_domain_decomposition.cpp [hunk 2]
```diff
@@ -1,902 +1,995 @@
...
   int extra_cells = 0;
 
   /* initialize cell neighbor structures */
-  dd.cell_inter = (IA_Neighbor_List *) Utils::realloc(dd.cell_inter,local_cells.n*sizeof(IA_Neighbor_List));
-  for(m=0; m<local_cells.n; m++) { 
-    dd.cell_inter[m].nList = NULL; 
-    dd.cell_inter[m].n_neighbors=0; 
+  dd.cell_inter = (IA_Neighbor_List *)Utils::realloc(
+      dd.cell_inter, local_cells.n * sizeof(IA_Neighbor_List));
+  for (m = 0; m < local_cells.n; m++) {
```
hmenke, line 57: Missed this one?

hmenke (approved):
Can we maybe agree that in the future you run clang-format after review?  Otherwise the changeset is just so huge and most of it is only whitespace,obfuscating the actual changes.

src/core/lb.cpp
```diff
@@ -3269,109 +3266,84 @@ void calc_particle_lattice_ia() {
...
+        lb_viscous_coupling(&p, force);
+
+        /* add force to the particle */
+        p.f.f[0] += force[0];
+        p.f.f[1] += force[1];
+        p.f.f[2] += force[2];
+
+        ONEPART_TRACE(if (p->p.identity == check_id) {
+          fprintf(stderr, "%d: OPT: LB f = (%.6e,%.3e,%.3e)\n", this_node,
+                  p->f.f[0], p->f.f[1], p->f.f[2]);
```
hmenke: Should be `p.p.identity` and `p.f.f[0]` since `p` is no longer pointer type.

src/core/lb.cpp
```diff
@@ -3269,109 +3266,84 @@ void calc_particle_lattice_ia() {
...
-                          );
-          }
+        {
+          lb_viscous_coupling(&p, force);
+        }
+
+        /* ghosts must not have the force added! */
+        ONEPART_TRACE(if (p->p.identity == check_id) {
+          fprintf(stderr, "%d: OPT: LB f = (%.6e,%.3e,%.3e)\n", this_node,
+                  p->f.f[0], p->f.f[1], p->f.f[2]);
```
hmenke: Should be `p.p.identity` and `p.f.f[0]` since `p` is no longer pointer type.

hmenke (changes requested):

src/core/forces.cpp
fweik: Yes, but this has requirements on the body that must be checked individually. So you can't make this parallel without touching it anyway.

src/core/lees_edwards_domain_decomposition.cpp
fweik: Nope, not a loop over particles (only over the cells)

hmenke (approved):

src/core/virtual_sites_com.cpp
```diff
@@ -481,28 +468,10 @@ double get_mol_dist_partcfg(Particle *p1, Particle *p2) {
...
-        mol_id2);
-    dist[0] = dist[1] = dist[2] = 0.0;
-    return;
-  }
-#endif
-  get_mi_vector(dist, p1_com->r.p, p2_com->r.p);
+  auto &p1_com = get_mol_com_particle_from_molid_cfg(mol_id1);
+  auto &p2_com = get_mol_com_particle_from_molid_cfg(mol_id2);
+
+  get_mi_vector(dist, p1_com.r.p, p2_com.r.p);
```
hmenke: Why did you remove the debugging snippet?

fweik:
Because the new code will throw and not return NULL.

hmenke (approved):

RudolfWeeber:
I went through the changes.
Due to the size of changes and the un-obvious nature of teh diff, it is sometimes hard to ensure that things are doing the same as before. In particular for algorithms such as Maggs, Ghmc and the dipolar stuff.
In the end, one can only make sure with integration tests.
The question is, should we merge this independently of the short range pr, or do we do that in one go, as the short range pr has a lot of extra testing?
Weh reviewing the short range pr, can we take this one as basis for the comparrison?

fweik:
Sorry for the messy pull request. Maybe when we have a freeze for the release we can also run clang-format over the whole repo to avoid this kind of problems in the future.
The short range stuff is mostly orthogonal, as the short range calculations are always related to cells, and not to all particles. I don't think it has any additional tests in it that would help, and would merge it independently.  
Unfortunately I can not easily add tests for MEMD and GHMC because they are not accessible from python, the magnetostatics has tests iirc.

RudolfWeeber:
To me, it is clear that this needs to be merged, eventually. By
getting rid of the c-style array access and the repetetive loop
code, everything gets much more safe. 
Since two people have looked through it already, I think, we better
merge now, so if there are regressions, we have a good chance of
finding them before the school.

PR #1193, opened 2017-06-06
Physical testcase for ICC*

KonradBreitsprecher:
Description of changes:
 - Test system is a parallel plate capacitor with electrodes of separation `L` made up of ICC charges.
 - Sets up a dipole moment `q*d` with two test charges which should be equal to the dipole moment of the induced charges `0.5*abs(Q_L)+abs(Q_R)*L`.

PR Checklist
------------
 - [x] Tests?
   - [ ] Interface
   - [x] Core 
 - [ ] Docs?

### DIFF
--- a/testsuite/python/icc.py
+++ b/testsuite/python/icc.py
@@ -7,10 +7,9 @@ import espressomd
 
 class test_icc(ut.TestCase):
            
-    from espressomd.electrostatics import P3M
-    from espressomd.electrostatic_extensions import ICC
-
     def runTest(self):
+        from espressomd.electrostatics import P3M
+        from espressomd.electrostatic_extensions import ICC
 
         S=espressomd.System()
         #Parameters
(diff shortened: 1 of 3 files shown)


testsuite/python/icc.py
```diff
@@ -0,0 +1,80 @@
...
+        nicc_tot = 2*nicc_per_electrode
+        iccArea = box_l*box_l/nicc_per_electrode
+
+        iccNormals=[]
+        iccAreas=[]
+        iccSigmas=[]
+        iccEpsilons=[]
+
+        l=box_l/nicc
+        for xi in xrange(nicc):
```
hmenke: There is no `xrange` in Python3.  Just use `range` instead.

hmenke (changes requested):

hmenke (approved):

PR #1201, opened 2017-06-14
Ladd boundaries

hmenke:
This PR introduces moving boundaries in the LB_GPU code which were developed as part of a Bachelor thesis by Robin Bardakcioglu (@rbardak) supervised by Florian Weik (@fweik).  I first merged the code into the `master` branch and from there into the `python` branch.  The TCL user interface was removed and replaced by a Python version.  It was Christian's wish to include this feature into ESPResSo.

So far there doesn't exist a testcase nor documentation.  I would be very happy if I could get some assistance from ICP there, especially because I no longer have access to a machine with a GPU. 

 - [ ] Testcase
 - [ ] Documentation


src/core/lbgpu_cuda.cu
```diff
@@ -1020,6 +1061,67 @@ __device__ void calc_n_from_modes_push(LB_nodes_gpu n_b, float *mode, unsigned i
...
+     pop_to_bounce_back =  n_curr.vd[(population+component*LBQ)*para.number_of_nodes + index ]; 
+     to_index_x = (x+c[0]+para.dim_x)%para.dim_x; 
+     to_index_y = (y+c[1]+para.dim_y)%para.dim_y; 
+     to_index_z = (z+c[2]+para.dim_z)%para.dim_z; 
+     to_index = to_index_x + para.dim_x*to_index_y + para.dim_x*para.dim_y*to_index_z; 
+     if ( n_curr.boundary[to_index] == 0) 
+     { 
+       boundary_force[0] += (2.0f*pop_to_bounce_back+shift)*c[0]/para.tau/para.tau/para.agrid; 
+       boundary_force[1] += (2.0f*pop_to_bounce_back+shift)*c[1]/para.tau/para.tau/para.agrid; 
+       boundary_force[2] += (2.0f*pop_to_bounce_back+shift)*c[2]/para.tau/para.tau/para.agrid;
```
mkuron, line 1119: Shouldn't this also cause a torque?

src/core/lbgpu_cuda.cu
```diff
@@ -2795,22 +2897,505 @@ __global__ void set_rho(LB_nodes_gpu n_a,  LB_rho_v_gpu *d_v, int single_nodeind
...
+  unsigned int index = blockIdx.y * gridDim.x * blockDim.x + blockDim.x * blockIdx.x + threadIdx.x;
+  if(!index) printf("\n init velocity %f ", current_boundary->velocity[0]);
+  unsigned int xyz[3];
+  float dummy_delta[3] = {0, 0, 0};
+  index_to_xyz(index, xyz);
+  if(!n_in.boundary[index])
+    set_equilibrium_from_u_rho(index, xyz, dummy_delta, current_boundary, d_v, n_in);
+}
+  
+/**inits moving boundaries
```
mkuron, line 3319: This comment is useless -- what does initialization consist of?

mkuron (approved):
Looks good overall, though I still don't like the idea of having `lb_integrate_moving_boundaries` completely separate from the main `integrate_vv` function. Right now, I think there is no way of having a moving boundary that also participates in a Lennard-Jones interaction, for example.

src/core/lbgpu_cuda.cu
hmenke: This is just the expansion of the old [`BOUNCEBACK()`](https://github.com/espressomd/espresso/pull/1201/files/80b4d00c71a44fe04e25db0dbfa7187383d6c394#diff-4e94214b9a5c2058eeed5947f6e317b7L1072) macro.  Torque is probably added elsewhere.

src/core/lbgpu_cuda.cu
hmenke: Welcome to ESPResSo.  Looks to me like you've been in waLBerla for too long.

PR #1204, opened 2017-06-15
Silence deprecated NumPy API warning

### LINKED ISSUE #1202: Silence numpy deprecation warning during compilation
it is my understanding that this is a Cython issue, not an Espresso one.

hmenke:
Fixes #1202 

Unfortunately, we cannot just follow the advice in the warning because using
````
#define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION
````
disables the old NumPy API which is used by Cython and the code will stop working.  Therefore we simply disable the warnings emitted by the C preprocessor for these files.  I believe this is a sensible approach as the person writing the Cython code has little influence on how the C++ code is actually generated and therefore compiler warnings are kind of meaningless in this case anyway.


RudolfWeeber:
Can be merged

mkuron:
This doesn't work on Clang and just produces warnings. The correct flag there is `-Wno-#warnings`.

PR #1211, opened 2017-06-27
Require Seed

davidsean:
Fixes #1039 

Description of changes:
 - 
Requires that the user seeds from the script, displays how to do it as part of the runtime error message

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


RudolfWeeber:
The tutorial scripts and corresponding tex files need to be changed, too.

hmenke:
Great!  More global variables.

mkuron:
This is broken on Python 3: https://travis-ci.org/espressomd/espresso/jobs/247486394#L1369

jonaslandsgesell:
This PR seems to have a lot of merge conflicts. Identical seeds may result in synchronization even when the system is setup in  different start configurations. This is due to stochastic thermostats like the langevin thermostat acting as an attractor and the Lyapunov exponent(s) therefore having a negative sign such that trajectories exponentially converge. The Lyapunov exponent is system dependent.

This attracting behaviour has for example been shown in 1D Lennard-Jones systems and I can reproduce this behaviour with espresso.

The remaining question is whether this synchronization may be present in systems which are of interest for us. In this case it is important to seed with different seeds when using multiple simulations to estimate the variance of observables. With same seeds (but different start configurations
) the estimated variance of the observables are wrongly too small due to the correlation of different runs with same seeds of the PRNG in espresso.

Synchronization has for example also been reported in a 3d alanine peptide system:
http://pubs.acs.org/doi/abs/10.1021/ct800573m

Although synchronization might not always happen I would be careful here and rather use a random seed. I would say to be on the safe side we should look at the alanine peptide example and see wether something like this can happen in our systems. If we decide it s not possible we can keep the fixed seeds and maybe just close this PR?

kosovan:
If the user is required to provide a seed, what prevents him from providing a random seed?

davidsean:
@jonaslandsgesell The question is not whether we want a random seed or not. I agree that for production runs a random seed is preferred. There are pros and cons in both directions, for example, the opposite is true in the development stage (to reproduce a crash after an attempted fix).

The issue is: do we want ESPResSo to make that decision silently for the user? The "REQUIRE SEED" trick enables ESPResSo to side-step the issue and to force the user to make a decision. ESPResSo does not know if the user is in a development vs production stage, and should not make that call.

davidsean:
@kosovan 
No problem with the user providing a random seed, the user can also explicitly ask espresso to find its own random seed it if is desired. The point is to force a line somewhere in the script that makes seeding explicit---and necessary.

KaiSzuttor:
i think that here we should have a default behaviour that catches  most of the users. Therefore, i think for most of the users a random seed is safe, while the people who need a fixed seed should really know that they need a fixed seed and can manually fix it. So i think we don’t need this PR.

RudolfWeeber:
> i think that here we should have a default behaviour that catches  most of the users. Therefore, i think for most of the users a random seed is safe, while the people who need a fixed seed should really know that they need a fixed seed and can manually fix it. So i think we don’t need this PR.
We had this discussion several times and the decision made at the
Espresso meeting was to make seeding explicit. 
system.seed=1
system.seed=np.random.random()
Let's stick with it.

KaiSzuttor:
you are right rudolf, still this is my opinion. if we stick to the decision (what we should do), then we should make the seed parameter a System constructor argument, just as it has been done for the box length.

jonaslandsgesell:
+1 for the constructor. Probably the cleanest solution for this behaviour

davidsean:
+1 for the constructor, will thus not require those extra global variables I added.

RudolfWeeber:
On Sat, Feb 03, 2018 at 09:02:43PM +0000, David Sean wrote:
> +1 for the constructor, will thus not require those extra global variables I added.
I'm not sure about the constructor. In constrast to the box
geometry which is always needed when you have particles, quite a few simulations may not require random numbers.

KaiSzuttor:
So how do you want to only make the seed required for these scenarios? Handle them explicitely and hope not to forget one?

davidsean:
sounds like we are back to` if (!user_has_seeded) throw Random::unseeded_error;` in places where random numbers generated.

doc/sphinx/advanced_methods.rst
```diff
@@ -270,54 +270,21 @@ current implementation with the ``COLLISION_DETECTION`` feature.
...
+
 Please contact the Biofluid Simulation and Modeling Group at the
 University of Bayreuth if you plan to use this feature.
 
-This section describes an alternative way to include soft elastic
-objects somewhat different from the previous chapter. In the Immersed
 Boundary Method (IBM), soft particles are considered as an infinitely
 thin shell filled with liquid (see e.g. :cite:`Peskin2002,Crowl2010,KruegerThesis`). When the
 shell is deformed by an external flow it responds by elastic restoring
+>>>>>>> 86ca06ad32a3dfa71547de702e9933fd6b7f6037
```
hmenke: Oops!

hmenke (changes requested):

RudolfWeeber:
@fweik, we need to merge this before the QA day. Is there sth left to do, here?

testsuite/constraint_shape_based_cylinder.py
```diff
@@ -0,0 +1,58 @@
+from __future__ import division, print_function
```
KaiSzuttor: shape based constraints are tested in ```constraint_shape_based.py```. This and the ellipsoid test file should be removed.

testsuite/constraint_shape_based_ellipsoid.py
```diff
@@ -0,0 +1,113 @@
+from __future__ import division, print_function
```
KaiSzuttor: remove this file

testsuite/constraint_shape_based_wall.py
```diff
@@ -0,0 +1,84 @@
+from __future__ import division, print_function
```
KaiSzuttor: remove this file

testsuite/lb_test.py
```diff
@@ -0,0 +1,44 @@
+from __future__ import print_function
```
KaiSzuttor: remove this file

testsuite/lb_test_gpu.py
```diff
@@ -0,0 +1,44 @@
+from __future__ import print_function
```
KaiSzuttor: remove this file

KaiSzuttor (changes requested):

KaiSzuttor:
there might be some more merge regressions to fix...

RudolfWeeber:
github and CI seem to be happy. Any specific thing @KaiSzuttor?

doc/tutorials/01-lennard_jones/scripts/lj_tutorial.py
```diff
@@ -65,7 +65,9 @@
 
 # System setup
 #############################################################
-system              = espressomd.System(box_l=[1.0, 1.0, 1.0])
+system              = espressomd.Systembox_l=([1.0, 1.0, 1.0])
```
fweik: This looks wrong.

fweik (changes requested):

src/core/random.hpp
```diff
@@ -83,6 +89,12 @@ void init_random_seed(int seed);
 
 inline double d_random() {
   using namespace Random;
+
+  if (!user_has_seeded && !unseeded_error_thrown) {
+    unseeded_error_thrown = true;
```
KaiSzuttor: there is a lot of code duplication here for every function that returns random numbers... also there is no need to throw if temperature is 0.0, e.g. the user only wants the damping part of a langevin thermostat

KaiSzuttor (changes requested):
@davidsean please remove code duplication and only throw if temperature != 0.0

testsuite/npt.py
```diff
@@ -73,7 +74,7 @@ def test_npt(self):
         print(avp,compressibility)
 
         self.assertAlmostEqual(2.0, avp, delta=0.02)
-        self.assertAlmostEqual(0.2, compressibility, delta=0.01)
+        self.assertAlmostEqual(0.2, compressibility, delta=0.013)
```
KaiSzuttor, line 77: Why did you change the accepted deviation?

RudolfWeeber:
On Tue, Jul 24, 2018 at 06:29:24AM -0700, Kai Szuttor wrote:
> 
> 
> 
> > @@ -73,7 +74,7 @@ def test_npt(self):
>          print(avp,compressibility)
>  
>          self.assertAlmostEqual(2.0, avp, delta=0.02)
> -        self.assertAlmostEqual(0.2, compressibility, delta=0.01)
> +        self.assertAlmostEqual(0.2, compressibility, delta=0.013)
> 
> Why did you change the accepted deviation?
Blew up for a certain number of cores.

KaiSzuttor:
I broke the drude testcase, no idea how. Thermalized bond and harmonic bond tests pass.

KaiSzuttor:
@fweik , ready for review i think

KaiSzuttor (approved):

src/core/random.hpp
```diff
@@ -1,40 +1,58 @@
...
 #include <string>
 #include <vector>
+#include <stdexcept>
 
 namespace Random {
 extern std::mt19937 generator;
 extern std::normal_distribution<double> normal_distribution;
 extern std::uniform_real_distribution<double> uniform_real_distribution;
+extern bool user_has_seeded;
+extern bool unseeded_error_thrown;
```
fweik: This should not be global, I think the best solution is to make it a static local variable in check_user_has_seeded.

fweik (changes requested):

src/core/random.hpp
KaiSzuttor: hmm, but how to switch the user_has_seeded variable?

jonaslandsgesell:
What was the exact reason for not requiring a user provided seed when initializing a random number generator (and preventing a random number generator to exist without a user provided seed or state)? That would be the correct dependency inversion right? Then we would not need to check for a seed to be present all the time.

src/core/random.hpp
fweik: I'm only talking about `unseeded_error_thrown`.

KaiSzuttor:
good idea, but i think that would be some more work. having our manpower in mind, i think this PR serves our current needs. But you can open an issue.

KaiSzuttor:
@fweik

PR #1224, opened 2017-07-12
Bond classes test

timi1995:
Fixes #

Description of changes:
 - new bond classes added 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


mkuron:
There are a lot of #something.cpp# files in your commit. They look like backups from your text editor. Please remove them. Also, the build is failing due to what looks like an undefined macro.

hmenke (commented):
The quality of this code is rather bad.  First of all, right now all the bonds are essentially duplicated, because I do not see any removal of the old code.  Second, as far as I can see, the `class`-ification of the bond types is intended to deprecate the `enum BondedInteraction` but it's still there. Third, this code will leak *as hell*:
```cpp
extern std::vector<Bond *> bonds_ia;
// ...
FENE* new_fene_bond = new FENE(r0, drmax, drmax2, 1.0/drmax2, k) ;
bonds_ia.push_back(new_fene_bond);
```
Better make this something like
```cpp
extern std::vector< std::shared_ptr<Bond> > bonds_ia;
// ...
bonds_ia.emplace_back(std::make_shared<FENE>(r0, drmax, drmax2, 1.0/drmax2, k));
```
if you need pointer semantics at all.  Otherwise make it a vector of `BondVariant` where `BondVariant` is a variant type which can hold any bond type.

fweik:
First of all, Timur is a new hiwi that is working on Espresso. This PR is only for review as the final design for the bonds is not fixed yet.

Regarding the content:
Please adhere to the Espresso naming standards, as found in the wiki [1].

The current plan is to use dynamic polymorphism here, this is why the container stores pointers (to the base class), but of course @hmenke is right that this should be done in a non-leaking manner. I think the right way to achieve this here is to use unique_ptr, because there is no actual shared ownership, and unique_ptr is zero overhead (shared_ptr isn't, it has a counter). See inline comments.

Please remove the commented out code.

Base class:
Since the derived classes will be destroyed thru base pointers, the base class should have a virtual destructor. For a in-depth discussion of this see [2].
As an abstract interface, it should not have data members. Please make Bond::get_interaction_code virtual
and move the implementation to the derived classes. (The use of protected is usual a sign that there is some design issue, because this breaks encapsulation. There are exceptions to this, but they do not apply here.)

Derived classes:
There is no need to use virtual and override for the implementations, please remove the virtual.

[1]: https://github.com/espressomd/espresso/wiki/Naming-conventions-and-folder-structure
[2]: http://www.gotw.ca/publications/mill18.htm

src/core/bond_exist.cpp
```diff
@@ -0,0 +1,22 @@
+#include"bond_exist.hpp"
+#include"interaction_data.hpp"
+#include"bond.hpp"
+bool make_bond_exist_bond_class(int bond_type){
```
fweik: Wouldn't a simple bonds_ia.resize(bond_type+1) suffice here? It is perfectly find to initialize the new bond to nullptr, it should not be accessed w/o being initialized anyway. I think a more sensible solution would be something like

```c++
void set_bond_by_type(int type, std::unique_ptr<Bond> && bond) {
  if(type >= bonds_ia.size()) {
    bonds_ia.resize(type + 1);
 }

 bonds_ia[type] = std::move(bond);
}
```

fweik: This is not only simpler, it is also a step in the direction of abstracting away the implementation of bonds_ia, which will probably be changed in the future.

src/core/bonded_coulomb.cpp
```diff
@@ -41,6 +42,15 @@ int bonded_coulomb_set_params(int bond_type, double prefactor)
   /* broadcast interaction parameters */
   mpi_bcast_ia_params(bond_type, -1); 
 
+  //bond classes part
```
fweik: If make_bond_exist_bond_class is replaced as suggested, this simplifies to 
```c++
#include "utils/make_unique.hpp"

set_bond_by_type(bond_type, Utils::make_unique<BONDED_COULOMB>(prefactor));
```

src/core/bonded_interaction.hpp
```diff
@@ -0,0 +1,68 @@
+#ifndef BONDED_INTERACTION_H
+#define BONDED_INTERACTION_H
+enum BondedInteraction {
```
fweik: If possible this should be an enum class. This has the advantage that it is scoped and does not pollute the namespace, and it does not decay into an int at first chance. You could then drop the BONDED_IA_ part from the names (because they are now qualified by the enum name, it is clear from the context that these are bond types. e.g. BondedInteraction::FENE ...)

I would also opt for renaming this to BondType, because this is a clearer description of what it is.

src/core/fene.cpp
```diff
@@ -46,6 +49,15 @@ int fene_set_params(int bond_type, double k, double drmax, double r0)
 
   /* broadcast interaction parameters */
   mpi_bcast_ia_params(bond_type, -1); 
-  
+
```
fweik: See BONDED_COULOMB

src/core/harmonic.cpp
```diff
@@ -41,5 +42,13 @@ int harmonic_set_params(int bond_type, double k, double r,double r_cut)
   /* broadcast interaction parameters */
   mpi_bcast_ia_params(bond_type, -1); 
 
+  //bond classes part
```
fweik: See BONDED_COULOMB

src/core/harmonic_dumbbell.cpp
```diff
@@ -44,6 +45,15 @@ int harmonic_dumbbell_set_params(int bond_type, double k1, double k2, double r,
   /* broadcast interaction parameters */
   mpi_bcast_ia_params(bond_type, -1); 
 
+  //bond classes part
```
fweik: See BONDED_COULOMB.

src/core/integrate.cpp
```diff
@@ -25,6 +25,28 @@
  *  see \ref integrate.hpp "integrate.hpp".
 */
 
+/*
+#########################################
+Fragen:
+-rescale vel/force: warum rescaling?
```
fweik: Good question :-)

src/core/integrate.cpp
```diff
@@ -25,6 +25,28 @@
...
+-bei calc_forces: warum jedes mal erneut init_forces?
```
fweik: init_forces sets the forces to zero (or to whatever comes out of the thermostat if it is turned on). Everything else just adds to them. Otherwise the particles would accumulate forces over time?!

src/core/integrate.cpp
```diff
@@ -25,6 +25,28 @@
  *  see \ref integrate.hpp "integrate.hpp".
 */
 
+/*
+#########################################
+Fragen:
+-rescale vel/force: warum rescaling?
+-bei calc_forces: warum jedes mal erneut init_forces? 
+-warum inline fkt für inline force???
```
fweik: Performance

src/core/integrate.cpp
```diff
@@ -25,6 +25,28 @@
...
+-wo werden forces hingeschrieben?
```
fweik: Particle::f::f and Particle::f::torque

src/core/integrate.cpp
```diff
@@ -25,6 +25,28 @@
...
+-warum bond forces bei single particle force?
```
fweik: I agree that the name is misleading here, what it is supposed to mean is that it does not act on particle interaction pairs like the short-range IAs (e.g. Lennard-Jones).

src/core/integrate.cpp
```diff
@@ -25,6 +25,28 @@
...
+-was macht ONEPART_TRACE z.B. bei fene.hpp?
```
fweik: This is a debugging facility that let's you track a specific particle through the complete calculation. It is often not feasible to print all the information for all parts because that can be a huge amount.

src/core/integrate.cpp
```diff
@@ -25,6 +25,28 @@
...
+-was macht die actor list?
+-Particle classe wo ist diese?
```
fweik: src/core/particle_data.hpp:311

src/core/integrate.cpp
```diff
@@ -25,6 +25,28 @@
...
+/*
+#########################################
+Fragen:
+-rescale vel/force: warum rescaling?
+-bei calc_forces: warum jedes mal erneut init_forces? 
+-warum inline fkt für inline force???
+-wo werden forces hingeschrieben?
+-warum bond forces bei single particle force?
+-was macht ONEPART_TRACE z.B. bei fene.hpp?
+-was macht die actor list?
```
fweik: Actors are things that map from a complete particle configuration to forces for all particles.

src/core/integrate.cpp
```diff
@@ -25,6 +25,28 @@
...
+-Particle classe wo ist diese?
+- wo werden observablen berechnet in integrate.cpp?
```
fweik: They aren't. They are either calculated on demand, or in the case of auto update this is triggered in mpi_integrate in src/core/communication.cpp:1291

src/core/interaction_data.hpp
```diff
@@ -1176,6 +1113,10 @@ extern int n_bonded_ia;
 /** Field containing the paramters of the bonded ia types */
 extern Bonded_ia_parameters *bonded_ia_params;
 
+/*vector of pointers because bond is an abstact class*/
+extern std::vector<Bond *> bonds_ia;
```
fweik:
```c++
#include <memory>

extern std::vector<std::unique_ptr<Bond>> bonds_ia;
```

src/core/quartic.cpp
```diff
@@ -45,5 +46,14 @@ int quartic_set_params(int bond_type, double k0, double k1, double r,double r_cu
   /* broadcast interaction parameters */
   mpi_bcast_ia_params(bond_type, -1); 
 
+  //bond classes part
```
fweik: See BONDED_COULOMB.

src/core/bond_exist.cpp
fweik: The move semantics is needed because unique_ptr can not be copied. (it would not be very unique otherwise...)

fweik (changes requested):

fweik:
The tests on travis fail because TABULATED is actually an existing macro (because it is a feature, defined from config,.hpp) Just rename it to TABULATED_IA or so

src/core/bond/Harmonic.cpp
```diff
@@ -0,0 +1,55 @@
...
+  } else { 
+     fac=0;
+  }
+  
+  for(i=0;i<3;i++)
+    force[i] = fac*dx[i];
+  ONEPART_TRACE(if(p1->p.identity==check_id) fprintf(stderr,"%d: OPT: HARMONIC f = (%.3e,%.3e,%.3e) with part id=%d at dist %f fac %.3e\n",this_node,p1->f.f[0],p1->f.f[1],p1->f.f[2],p2->p.identity,dist2,fac));
+  ONEPART_TRACE(if(p2->p.identity==check_id) fprintf(stderr,"%d: OPT: HARMONIC f = (%.3e,%.3e,%.3e) with part id=%d at dist %f fac %.3e\n",this_node,p2->f.f[0],p2->f.f[1],p2->f.f[2],p1->p.identity,dist2,fac));
+
+#ifdef CONFIGTEMP
```
hmenke: Wasn't this removed in #1206?

jonaslandsgesell:
What is the status here?

timi1995:
Currently, we are testing the performance. 
After that the old code will be removed and then it‘s finished.

KaiSzuttor:
Whats the status here?

fweik:
Waiting until somebody wants do do the bond refactoring? This has all the bonds put into classed, which can probably be reused.

PR #1225, opened 2017-07-13
Particle copy fix

fweik:
Fixes segfaults with partCfg under some circumstances.

Description of changes:
 - move/ (deep) copy for Utils::List
 - Particle::flat_copy() that copies Particles w/o bonds and exclusions
 - Some adjustments to ParticleCache to keep the particles in a valid (read destructible) state if they do not 
   carry bonds/exclusions.
 - Removed fcs_acf from the correlator because it was broken (and affected by the changes on List)
 - Some fixes to Utils::List
 - Getter Particle::exclusions() that may show a way to get rid of all the pointer_to_... stuff (will throw if the 
   feature is not enabled.)

### DIFF
--- a/src/core/ParticleCache.hpp
+++ b/src/core/ParticleCache.hpp
@@ -126,8 +126,6 @@ class ParticleCache {
   std::unordered_map<int, int> id_index;
   /** The particle data */
   map_type remote_parts;
-  /** Bond info for remote_parts */
-  std::vector<int> bond_info;
   /** State */
   bool m_valid, m_valid_bonds;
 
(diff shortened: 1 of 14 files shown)


kosovan:
Hi Flo, could you please explain why you removed fcs_acf from the correlator? If it was broken, I would like to fix it. We are using this function quite a lot and it would be a pity to lose it.

fweik:
As far is I could tell, args is not initialized. This is then cast to a DoubleList * and used. Are you sure that this works?! If you can show me where this comes from I'll fix it. I don't think it is hard to, but it was not clear to me how this could work in the first place.

kosovan:
Well, I know exactly how this was done in Espresso 3.x The original idea was that `void *args` is a generic pointer to pass arbitrary additional arguments to the correlation operation. The TCL parser allocates and initializes the data for the given correlation operation, then casts it to `(void *)` which is then passed to the correlator.  The correlator casts it back to the original type. Eventually, we only used it for the `fcs_acf`, and all other correlation operations set `args=NULL`.

In the release version 3.3.1, the corresponding lines in statistics_correlator.cpp read
`dim_corr = dim_A/3;`
`self->corr_operation = &fcs_acf;`
`self->args = args;`

The last line was deleted in Correlators.cpp.  But just adding the line won't solve the problem.  We also need to implement the passing of three doubles from the script interface to `fcs_acf`. I looked at the corresponding code but am not sure how this should be properly done.

In 3.x the `args` was initialized by the TCL parser (see lines 571+ in the attached file
[statistics_correlation_tcl_cpp.txt](https://github.com/espressomd/espresso/files/1145605/statistics_correlation_tcl_cpp.txt))

Since the `args` is not used by anything else, maybe the generic solution is not necessary, and it would be enough to turn it to an array of 3 doubles which is passed to `fcs_acf()`.

fweik:
So I will revert the change, then it will be as broken as before. We can
then just make the the args settable from python. We can just take it as
three floats because it is not used anywhere else.

kosovan:
I agree.

PR #1227, opened 2017-07-15
Particle copy fix

fweik:
Fixes segfaults with partCfg under some circumstances.

Description of changes:
 - move/ (deep) copy for Utils::List
 - Particle::flat_copy() that copies Particles w/o bonds and exclusions
 - Some adjustments to ParticleCache to keep the particles in a valid (read destructible) state if they do not 
   carry bonds/exclusions.
 - Removed fcs_acf from the correlator because it was broken (and affected by the changes on List)
 - Some fixes to Utils::List
 - Getter Particle::exclusions() that may show a way to get rid of all the pointer_to_... stuff (will throw if the 
   feature is not enabled.)

### DIFF
--- a/src/core/utils/List.hpp
+++ b/src/core/utils/List.hpp
@@ -56,6 +56,7 @@ public:
   size_type size() const { return n; }
   bool empty() const { n == 0; }
   size_type capacity() const { return max; }
+  void clear() { resize(0); }
 
 private:
   /**
--- a/src/core/utils/List.hpp
+++ b/src/core/utils/List.hpp
@@ -5,6 +5,8 @@
 #include <cstdint>
 #include <limits>
 #include <type_traits>
+#include <iterator>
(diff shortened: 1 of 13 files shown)


src/core/utils/List.hpp
```diff
@@ -1,31 +1,126 @@
...
 #include "memory.hpp"
 
 namespace Utils {
-
 /** List.
     Use the functions specified in list operations. */
 template <typename T> struct List {
+  static_assert(std::is_pod<T>::value, "");
+
+  using size_type = uint32_t;
```
hmenke, line 19: Why not just `std::size_t`?  (and corresponding `size_t` in Cython)

hmenke (approved):

fweik:
List is used in places where the empty size is important, e.g. for bonds
and exclusions. Otherwise we could just use a vector.

PR #1231, opened 2017-07-25
added minimalistic MDAnalysis support

Marcello-Sega:
Description of changes:
 - adds some basic support for online analysis with MDAnalysis


Marcello-Sega:
Hi, as discussed at the LMC, here's the patch about MDAnalysis. Mind that I tested it only with MDAnalysis==0.15, not with newer ones. 
There's an example on how to use it here: 

[samples/python/MDAnalysisIntegration.py](https://github.com/espressomd/espresso/pull/1231/files#diff-05e8aa9a94c141701489a4fbbc3e3db0)

If you like the patch I will add tests / support for newer versions, and automate the exposure of espresso data to MDAnalysis, so that it becomes transparent to the user.

fweik:
Thank you, I will fix the remaining macOS issues.

fweik:
Sorry for the delay, we had some issues with updates in our test environment.
Ok, so this looks good. We'd like to have this. It would be nice if you could provide some minimal documentation on how to use it. If it would work with newer versions of MDA, this would of course also be very much appreciated.

Marcello-Sega:
Hi, 
I ported the thing to the latest MDAnalysis version, and dropped support for the older one: the topology system of the two versions is so much different that it is not worth supporting both.

I condensed everything in the `__init__.py` of the MDA_ESP module, adding docstrings, and I polished the usage example in `samples/python/MDAnalysisIntegration.py` 

Let me please know if you need some other kind of documentation (manual, ..., ?)

fweik:
Thank you, that looks good to me. I think the provided examples and in-code documentation is sufficient, but maybe you could mention the existence of this in the IO section of the sphinx docs and link to the appropriate inline docs, just so that people can find out about it.

Marcello-Sega:
Ok will do that asap

hmenke (approved):
The example should be made compatible with Python 3 by doing
```python
```
and replaceing all istances of `print ...` by `print(...)`.

Marcello-Sega:
Added the documentation to io.rst, merged last changes, fixed print fcuntion

fweik (approved):

fweik:
Thank you!

PR #1158, opened 2017-05-08
Sphinx documentation for online visualization with mayavi/openGL

KonradBreitsprecher:
Fixes #1129 

Description of changes:
- Added docu for online visualization
 - Deleted VMD/IMD code/doc 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [x] Docs?

PR #1232, opened 2017-07-25
IMD cleanup. Removed both transfer_rate fields from globals. Updated …

KonradBreitsprecher:
…io sphinx docu

Makes PR #1158 obsolete.

Description of changes:
- Deleted imd code
- Removed transfer_rate from globals and interface
- Updated sphinx docu


KonradBreitsprecher:
Somehow critical is the removal of the two global transfer_rate vars. I removed those and updated the indices. I hope nobody used direct numerical field indices for the globals.

fweik (approved):

fweik:
Great PR: +139 −1,000

mkuron (approved):

fweik:
The mac issue is the same as the one fixed in b1379a1, why this only strikes on macOS is beyond me...

PR #1234, opened 2017-07-25
Remove switchable noise

### LINKED ISSUE #1137: What to do about different thermostat noise distributions?
By means of compile time switches, one can chose from uniform, Gaussian and cutoff Gaussian distributions.
This is not documented nad requires separate CI builds.
Is this actually relevant for anyone? Do we keep it?

fweik:
Fixes #1137 

Description of changes:
 - Removed switchable noise type from MD

Note that the feature remains and that the noise is still switchable for LB.


RudolfWeeber:
I think, this can be merged

mkuron:
Sorry I didn't review this despite @fweik's request. It seems like for some reason Github only sends me a random subset of the email notifications it should send. The review request wasn't among them.

PR #1236, opened 2017-07-25
espressomd: Removed left-over occurences of updatePartCfg

fweik:
Fixes #

Description of changes:
 - 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?

### DIFF
--- a/src/core/unit_tests/ParticleCache_test.cpp
+++ b/src/core/unit_tests/ParticleCache_test.cpp
@@ -25,6 +25,13 @@
 #include <random>
 #include <vector>
 
+#include <boost/version.hpp>
+/* Work around bug in boost, see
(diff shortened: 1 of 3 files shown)


fweik:
The macOS stuff was updated to Cython 0.26, and everything is broken.

mkuron:
It's unclear to me what change in 0.26 causes that. The [changelog](https://github.com/cython/cython/blob/master/CHANGES.rst) doesn't mention anything backwards-incompatible that looks like the error you are seeing.

fweik:
I can reproduce this on ubuntu with 0.26, and one of the errors is an
compiler crash. I will find out.

PR #1237, opened 2017-07-25
particle_data: Reset local_particles properly on remove_all_particles…

fweik:
…(), and correctly initialize local_cells after resize.

@RudolfWeeber please check if this fixes your issue.


RudolfWeeber:
Can be merged and solves the issue we worked on yesterday.

PR #1271, opened 2017-07-28
Gitlab ci

KaiSzuttor:
Introduces CUDA builds and documentation push to the website.


maintainer/cuda_build.sh
```diff
@@ -0,0 +1,31 @@
+GIT_COMMIT=$(git rev-parse HEAD)
+URL=$(echo "https://gitlab.icp.uni-stuttgart.de/kai/espresso-build/-/jobs/${CI_JOB_ID}")
+STATUS="pending"
+curl "https://api.github.com/repos/kaiszuttor/espresso/statuses/$GIT_COMMIT?access_token=$GITHUB_TOKEN" \
+      -H "Content-Type: application/json" \
+      -X POST \
+      -d "{\"state\": \"$STATUS\", \"context\": \"ICP CUDA build\", \"target_url\": \"$URL\"}"
```
mkuron: Replace "ICP CUDA build" with "ICP Gitlab CI". We'll soon be using this as a general-purpose Travis-CI alternative. And rename "cuda_build.sh" to "gitlab_build.sh".

maintainer/cuda_build.sh
```diff
@@ -0,0 +1,31 @@
+GIT_COMMIT=$(git rev-parse HEAD)
+URL=$(echo "https://gitlab.icp.uni-stuttgart.de/kai/espresso-build/-/jobs/${CI_JOB_ID}")
```
mkuron: Create a public namespace "espressomd" on our Gitlab server and make a new public project "espresso" there. Use that instead.

maintainer/cuda_build.sh
```diff
@@ -0,0 +1,31 @@
...
+STATUS="pending"
+curl "https://api.github.com/repos/kaiszuttor/espresso/statuses/$GIT_COMMIT?access_token=$GITHUB_TOKEN" \
```
mkuron: This will need to be _espressomd_, and you need to have @fweik create a new token in the project settings.

maintainer/cuda_build.sh
```diff
@@ -0,0 +1,31 @@
...
+      -H "Content-Type: application/json" \
+      -X POST \
+      -d "{\"state\": \"$STATUS\", \"context\": \"ICP CUDA build\", \"target_url\": \"$URL\"}"
+
+whoami;
+nvidia-smi;
+cd ${CI_PROJECT_DIR};
+cp maintainer/configs/maxset.hpp myconfig.hpp;
+mkdir build && cd build;
+cmake .. -DCUDA_NVCC_FLAGS="-D_FORCE_INLINES";
```
mkuron: why these flags?

maintainer/cuda_build.sh
```diff
@@ -0,0 +1,31 @@
...
+STATUS="pending"
+curl "https://api.github.com/repos/kaiszuttor/espresso/statuses/$GIT_COMMIT?access_token=$GITHUB_TOKEN" \
+      -H "Content-Type: application/json" \
+      -X POST \
+      -d "{\"state\": \"$STATUS\", \"context\": \"ICP CUDA build\", \"target_url\": \"$URL\"}"
+
+whoami;
+nvidia-smi;
+cd ${CI_PROJECT_DIR};
+cp maintainer/configs/maxset.hpp myconfig.hpp;
```
mkuron: Can we make this script more generic like the Jenkins and Travis scripts where they can run different myconfigs?

maintainer/cuda_build.sh
```diff
@@ -0,0 +1,31 @@
...
+URL=$(echo "https://gitlab.icp.uni-stuttgart.de/kai/espresso-build/-/jobs/${CI_JOB_ID}")
+STATUS="pending"
+curl "https://api.github.com/repos/kaiszuttor/espresso/statuses/$GIT_COMMIT?access_token=$GITHUB_TOKEN" \
+      -H "Content-Type: application/json" \
+      -X POST \
+      -d "{\"state\": \"$STATUS\", \"context\": \"ICP CUDA build\", \"target_url\": \"$URL\"}"
+
+whoami;
+nvidia-smi;
+cd ${CI_PROJECT_DIR};
```
mkuron: Why so many semicolons? This is bash...

maintainer/cuda_build.sh
```diff
@@ -0,0 +1,31 @@
+GIT_COMMIT=$(git rev-parse HEAD)
```
mkuron, line 3: Add a shebang, just for completeness.

.gitlab-ci.yml
```diff
@@ -0,0 +1,35 @@
...
+    - bash maintainer/cuda_build.sh
+
+
+sphinx_documentation:
+  stage: deploy
+  only:
+    - doc
+  before_script:
+    # Install ssh-agent if not already installed, it is required by Docker.
+    # (change apt-get to yum if you use a CentOS-based image)
```
mkuron: Comment not relevant. The check probably also isn't relevant.

.gitlab-ci.yml
```diff
@@ -0,0 +1,35 @@
...
+    # Run ssh-agent (inside the build environment)
+    - eval $(ssh-agent -s)
+
+    # Add the SSH key stored in SSH_PRIVATE_KEY variable to the agent store
+    - ssh-add <(echo "$SSH_PRIVATE_KEY")
+
+    # For Docker builds disable host key checking. Be aware that by adding that
+    # you are suspectible to man-in-the-middle attacks.
+    # WARNING: Use this only with the Docker executor, if you use it with shell
+    # you will overwrite your user's SSH config.
```
mkuron: There is a safety check (`[[ -f /.dockerenv ]]`) that prevents just that scenario.

mkuron (commented):
Looks good to me, but needs more work before it can be merged, see my comments.

.gitlab-ci.yml
```diff
@@ -0,0 +1,35 @@
...
+build_and_test:
+  stage: build
+  script:
+    - bash maintainer/cuda_build.sh
+
+
+sphinx_documentation:
+  stage: deploy
+  only:
+    - doc
```
mkuron, line 18: Shouldn't this be only: python? We don't want deployments of pull requests or stable branches, but there is no doc branch.

mkuron (changes requested):

maintainer/cuda_build.sh
KaiSzuttor: nice idea, but thats already done, just forgot to change the script. The script itself is actually just a proof of concept, not a polished version for production testing.

maintainer/cuda_build.sh
KaiSzuttor: I'm not sure if we want to run all tests on travis.

maintainer/cuda_build.sh
KaiSzuttor: I like them...

maintainer/cuda_build.sh
KaiSzuttor: See my previous comment.

.gitlab-ci.yml
KaiSzuttor: That's internal stuff, happening on the endpoint site. We only want to update the doc if pushes are received. Therefore I force push to doc branch for pull requests.

maintainer/cuda_build.sh
KaiSzuttor: They are needed for CUDA 7.5 build on xenial.

.gitlab-ci.yml
KaiSzuttor: Sorry, i meant push to python branch for PR and to doc for merging.

maintainer/cuda_build.sh
mkuron: Not all, but definitely more than just a single CUDA test. In fact, the doc deployment is not CUDA, but if it fails, "ICP CUDA build" will show up with a red X. So the name already need to be more generic.

KaiSzuttor: I dont want to discuss on details such as names at this early stage. Actually the purpose of the test is running CUDA code. The status is set by the build stage, a failing doc will not cause a red cross.

maintainer/cuda_build.sh
hmenke: That is even very important!  The `$(...)` command execution is Bash specific.  POSIX only mandates `` `...` ``.

maintainer/cuda_build.sh
mkuron: Hmm, a failing doc deploy should cause a red cross though. Otherwise changes that break the documentation and other failures may go unnoticed for weeks.

maintainer/cuda_build.sh
KaiSzuttor: yes, but it is explicitly called with bash

maintainer/cuda_build.sh
KaiSzuttor: yes but we can name the context differently for the doc stage and inform on github with a second status.

hmenke (approved):

PR #1273, opened 2017-07-30
Barnes-Hut octree dipole sum GPU implementation

bogdan-tanygin:
This is an adaptation of the original method https://doi.org/10.1016/j.cpc.2013.01.016 to the Espresso architecture as close as possible. Original scientific team had shared this source code. They are not my colleagues, I've found their research through a public Internet search and contacted them by an email. Dr. Alexander (Polyakov) Peletskyi has contacted all co-authors and they kindly gave a permission to move the source code to an open source according to the GPL license conditions. I've added a corresponding Copyright notice to the actor implementation:
```
  Copyright (C) 2016,2017 The ESPResSo project
  Copyright (C) 2012 Alexander (Polyakov) Peletskyi
```
(last 2 years correspond to my work on this adaptation)

Documentation and test have been added.

**Performance**: I've run the Espresso simulation of 10^6 magnetic nanoparticles VV integration. The given method is 230-300 times faster than an existing DDS GPU summation for such number of particles. I've obtained the 1M-nanoparticle Stockmayer polymer structure quite fast (an image is attached below). The theoretical complexity of the "classical" BH method is O(N*log(N)). After a comparison of 10^4, 10^5, and 10^6 nanoparticles I see a precise dependence O(N) which could be related to this specific GPU parallel implementation of the BH octree handling.

agg1e6p

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


hmenke (approved):

RudolfWeeber:
Please do not merge yet. I still need some time for this.

RudolfWeeber:
Barnes Hut would be a very nice addition to the dipolar methdos.

At this stage, I had a bit of a hard time understanding the code, though.

If possible, could you please do the following:
* Remove commented out stuff, unless there is a good reason to keep it
* Put in comments explaining the meaning of all the variables and functions
* Put in a comment explaining the data storage, in particular, how the tree is represented
* Put in a comment explaining the distribution of work to the threads
* Can the thrust library be used for the reduction?
* It might be advantageous to use 
    float[3] 
  rather than 
    float x, y, z;
  Then, it is more obvious what is done without carefully checking indices for all three Cartesian directions

bogdan-tanygin:
Hi @RudolfWeeber 
Thanks for a feedback. OK, I will do a cleanup and some refactoring. However, a detailed commenting of this code would be problematic. As I noted, this code has been developed by other team (with limited comments). I do not cooperate with them and they are hardly available now. I've just integrated it and checked tests. I have the same issues with its understanding as you have :-) If such "blind" approach is undesirable then it's OK, I will understand it.

bogdan-tanygin:
@RudolfWeeber 
I've made changes according to your comments.

> Remove commented out stuff, unless there is a good reason to keep it

Done

> Put in comments explaining the meaning of all the variables and functions

Done. Either at the declaration or inside the function at a place of usage.

> Put in a comment explaining the data storage, in particular, how the tree is represented

Done. Same.

> Put in a comment explaining the distribution of work to the threads

Done, there is some explanation concerning an organization of SMs, blocks, warps, threads in all 5 kernels. Actually, an algorithm, as I said, is quite sophisticated and related to a low-level CUDA work "under the hood". Also, there is a lot of technical optimizations which makes at glance view quite non-obvious. I've tried to fix this by comments.

Also, further detailed description of the method and its optimization with identical 5-6 kernels and pseudo-code (which almost correspond to the code of this pull request) could be referred here:

http://iss.ices.utexas.edu/Publications/Papers/burtscher11.pdf

Actually, this is an original BH method which has been implemented now to dipolar systems.

> Can the thrust library be used for the reduction?

Yes, actually, I've used it in the function `void energyBH()`. I've followed the same approach as it takes place already in Espresso for `DipolarDirectSum_cuda.cu`: the `thrust` is using for a sum the energy results of all blocks.

> It might be advantageous to use float[3] rather than float x, y, z; Then, it is more obvious what is done without carefully checking indices for all three Cartesian directions

Done

RudolfWeeber:
Sorry for the long wait.
Thanks a lot for providing all the additional clarification and commentary in the code.
I'll prepare this for merge now.

RudolfWeeber:
The current state of affairs is in the branch pr1273 on my github.

I merged the current Python branch. I also renamed the feature to DIPOLAR_BARNES_HUT to avoid confusion.
Unfortunately, the test for the energy does not pass. I'll look into it.

I noticed that there is Banres Hut specific code in EspressoSystemInterface.
I don't think, something method specific should go in there.
Do I understand it correctly that BH uses the position and dipole arrays, to store both, particles' and cells' positions and dipole moments. Is this the reason why the arrays are elongated if BH is compiled in?

Do I further understand it correctly that in these elongated arrays, the particles come first, then the BH nodes?

In that case I will move the relevant code out of EspressoSystemInterface and just copy the arrays provided by the system interface into new ones within the BH method. The effort of the copy is probably negligible compared to the calculations themselves.

Could you please also clarify the meaning of the two method parametes? How do I have to change them to approximate the exact result?

bogdan-tanygin:
@RudolfWeeber thanks for your support with merging. No problems concerning a delay at all.

> Unfortunately, the test for the energy does not pass. I'll look into it.

I'll take a look as well..

> Do I understand it correctly that BH uses the position and dipole arrays, to store both, particles' and cells' positions and dipole moments. Is this the reason why the arrays are elongated if BH is compiled in?

Correct.

> Do I further understand it correctly that in these elongated arrays, the particles come first, then the BH nodes?

Correct.

>> Could you please also clarify the meaning of the two method parametes? How do I have to change them to approximate the exact result?

It is shortly described in the documentation.

`...I_tol2 and epsilon2 which define a fraction of the cell and an additive distance respectively.`

Distances proportional to roots of these parameters are added and define that a cell is far enough to be considered as a single cumulative dipole. Hence, an increase of these parameters (one of them or both) will lead to higher results approximation and less performance.

I've selected these parameters within our test for enough accuracy and a good performance:

`epssq = 100.0, itolsq = 4.0`

It means 2 of the cell size plus 10 arbitrary units distance.

RudolfWeeber:
> It is shortly described in the documentation.
> 
> `...I_tol2 and epsilon2 which define a fraction of the cell and an additive distance respectively.`
> 
> Distances proportional to roots of these parameters are added and define that a cell is far enough to be considered as a single cumulative dipole. Hence, an increase of these parameters (one of them or both) will lead to higher results approximation and less performance.
> 
> I've selected these parameters within our test for enough accuracy and a good performance:
> 
> `epssq = 100.0, itolsq = 4.0`
> 
> It means 2 of the cell size plus 10 arbitrary units distance.
Sorry, I don't get that.
Given, eps and itol, a cell size c and a particle-particle distance
d, what is the expression which determins, whether the interaction
between the particle and the cell or the on between the two
particles is taken?

bogdan-tanygin:
an expression that entities are close enough to split the cell(s): `d<sqrt(itolsq*c*c+epssq)`

bogdan-tanygin:
Hi @RudolfWeeber 
I've just aligned the new magnetostatics prefactoring.
https://github.com/RudolfWeeber/espresso/pull/28
Now, the test is passing:
https://gist.github.com/psci2195/31244c71e07fc5436ba43b4e7a4b41af

bogdan-tanygin:
I've also made the test `dawaanr-and-bh-gpu.py` deterministic and increased its energy validation accuracy to 1E-4.

bogdan-tanygin:
> In that case I will move the relevant code out of EspressoSystemInterface and just copy the arrays provided by the system interface into new ones within the BH method. The effort of the copy is probably negligible compared to the calculations themselves.

Could I do it or you will?

> I also renamed the feature to DIPOLAR_BARNES_HUT to avoid confusion.

I see that you'd described both features in the `doc/sphinx/features.rst`. Does it mean that we should have BARNES_HUT feature enabling the elongated arrays meaning that the BH octree could be reused by the electrostatics and other long-range interactions? We could separate code fragments which are purely magnetic dipole related ones and designate them by DIPOLAR_BARNES_HUT.

RudolfWeeber:
> > In that case I will move the relevant code out of EspressoSystemInterface and just copy the arrays provided by the system interface into new ones within the BH method. The effort of the copy is probably negligible compared to the calculations themselves.
> 
> Could I do it or you will?
If you feel comfrtable doing it, that would be great. There is still
a buch of other stuff I need to do for the release.

bogdan-tanygin:
OK, I'm proceeding then..

bogdan-tanygin:
@RudolfWeeber 
I've done the requested changes within the PR: https://github.com/RudolfWeeber/espresso/pull/28

RudolfWeeber:
I merged the current Python branch and adapted the test to the Bjerrum length removal.
Once CI testing passes, its ready to merge.
Thanks a lot!

fweik:
Please remove the `register` storage specifier as it is deprecated and will be removed in c++17. (See failing gitlab ci tests)

bogdan-tanygin:
Thanks!
However, the `maxset` gitlab test is failed as well even though it has been passed on my GPU:
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/8342

I will troubleshoot the reason.

I think that `dawaanr-and-dds-gpu.py` is failing there as well. It is just disabled by:

`python_test(FILE dawaanr-and-dds-gpu.py MAX_NUM_PROC 4)`
`python_test(FILE dawaanr-and-bh-gpu.py MAX_NUM_PROC 1)`

in the `CMakeLists.txt` of the testsuite.

It could be an issue related to the bjerrum length disabling. However, I'm still is not sure why, the corresponding changes looks correct. I need to troubleshoot further.

bogdan-tanygin:
The energy test failure discrepancy does not equal to the temperature quantitatively. Hence, this issue, probably, is not related to the Bjerrum length removal PR.

Most probably, this is a GPU device-specific issue. My computational capability is smaller. It is something different.

However, I see this new warning in the log:

`More than one GPU detected, please note Espresso uses device 0 by default regardless of usage or capability`.

Colleagues, please confirm: is zeros GPU a correct one in the gitlab for the force and energy actors? Which one is used within other actors? I think the same cause the `cuda_set_device()` is not called from other *hpp or python files.. Hence, a further troubleshoot is needed.

My GPU is Tesla C2070 with the computational capability 2.0: I've enabled it in the `espresso/CMakeLists.txt`  as `set(CUDA_NVCC_FLAGS "${CUDA_NVCC_FLAGS} -gencode=arch=compute_20,code=sm_20`.

Maybe, the `register` variables could impact this failed test indeed for such the computational capability difference.

bogdan-tanygin:
How could I troubleshoot the gitlab tests, just using commits here :-) ? Or some limited access to the same VM is possible or some fake PRs between other forks?

bogdan-tanygin:
> How could I troubleshoot the gitlab tests, just using commits here :-) ? Or some limited access to the same VM is possible or some fake PRs between other forks?

I will allocate some Tesla machines here: https://cloud.google.com/gpu/ to troubleshoot the issue.
It will also help to compare tests with the Gitlab GeForce ones.

However, the 1st step - I will fix compilation warnings and make the introduction of the useful architecture advices represented by the above comments of @fweik and the ones referenced from @hmenke (#1766).

I'll feedback the results later (several days+) unless someone already has an understanding of the energy test issue root cause.

RudolfWeeber:
On Tue, Jan 16, 2018 at 02:20:49PM +0000, Dr. Bogdan Tanygin wrote:
> I will allocate some Tesla machines here: https://cloud.google.com/gpu/ to troubleshoot the issue.
> It will also help to compare tests with the Gitlab GeForce ones.
Before doing that, please note that I reduced the tolerance of teh test, which worked fine on my gpu, but might not on an other one. It might be that that has to be reverted.

I'm a bit surprised by the low per-particle accuracy achieved (5 digits).
Based on the previous discussion, I increased the two method parameters (by up to 6 orders of magnitude), which did not improve things.

To restart the CI tests, just push to the branch of the pull request.

mkuron:
> `More than one GPU detected, please note Espresso uses device 0 by default regardless of usage or capability.`

That is fine. If you look at the top of the Gitlab CI output, you'll see that nvidia-smi reports it as being a GeForce GTX 780 Ti.

> My GPU is Tesla C2070 with the computational capability 2.0

Math is slightly different on the different compute capabilities. I thought 2.0 was the first one to have full IEEE 754 compliance, but maybe you need 3.0 for that. The rounding and the use of fused multiply-add operations may be different. Espresso doesn't officially support 2.x anymore because CUDA 9.0 and higher cannot compile for 2.x.

> Maybe, the register variables could impact this failed test indeed for such the computational capability difference.

Probably not as this is not much more than a hint to the compiler. CUDA only has a certain number of registers available and if you use less than that, all your variables are in registers anyway. If you use more, things will get slower because the remaining variables will be stored in global memory.

RudolfWeeber:
I re-examined the accuracy issue. Both, the direct sum on the gpu and Banres Hut with high accuracy parameters give comparable rms force errors and energy errors. On the order of 10^-4 with quite large variations.
With regards to the CI tests failing due to the accuracy, you can revert the increase in accuracy target for the forces/torques on single particles.

In the long term, we might consider Kahan summation, but that's a different story.

bogdan-tanygin:
@RudolfWeeber, thanks! However, maybe specifically the GitLab environment (the latest maxset build) has another issue:

https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/8342
`AssertionError: False is not true : Energies for dawaanr -2.65532693338 and dds_gpu -1.56711867849 do not match.`

It is related to the whole energy, not to the specific particles.. And the error is quite large, at least larger than 1E-4. Do you refer to another CUDA environment? Maybe I misunderstood something..

RudolfWeeber:
OK. That's too much of a deviation. I think, the most probable cause might be the use of un-initialized memory.
You could try to run the test with
pypresso --cuda-memcheck

src/core/actor/DipolarBarnesHut.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include "../forces.hpp"
+#include "EspressoSystemInterface.hpp"
+#include "forces.hpp"
+#include "energy.hpp"
+
+#ifdef DIPOLAR_BARNES_HUT
+
+void activate_dipolar_barnes_hut(float epssq, float itolsq)
+{
+    if (dipolarBarnesHut)
```
hmenke: This check is not necessary, cf. [[expr.delete]](http://eel.is/c++draft/expr.delete#2) of the C++ standard.

> [...] the value of the operand of delete may be a null pointer value [...]

You only have to make sure that *after* calling `delete` you set the value to `nullptr` so that calling `delete` twice will not crash the program.

src/core/actor/DipolarBarnesHut.cpp
```diff
@@ -0,0 +1,59 @@
...
+        delete dipolarBarnesHut;
+
+    // also necessary on 1 CPU or GPU, does more than just broadcasting
+    mpi_bcast_coulomb_params();
+    dipolarBarnesHut =new DipolarBarnesHut(espressoSystemInterface, epssq, itolsq);
```
hmenke: Is it maybe possible to refactor this into a `std::shared_ptr`?  Then you don't need the global variable and explicit `new` and `delete`.

RudolfWeeber:
> hmenke commented on this pull request.
[...]
> +    dipolarBarnesHut =new DipolarBarnesHut(espressoSystemInterface, epssq, itolsq);
> 
> Is it maybe possible to refactor this into a `std::shared_ptr`?  Then you don't need the global variable and explicit `new` and `delete`.
It probably is.  This code is taken from the dipolar direct sum
which also doesn't use shared_ptr. 
To my understanding, the enabling/disabling of the method works
without segfaults, though.
So, once the energy accuracy issue is resolved, I think, we can go ahead
with this one.

src/core/actor/DipolarBarnesHut.cpp
bogdan-tanygin: Thanks for the review. Done.

src/core/actor/DipolarBarnesHut.cpp
bogdan-tanygin: I've started some drafting of the `shared_ptr` based architecture minor updates.. However, according to the @RudolfWeeber's [comment](https://github.com/espressomd/espresso/pull/1273#issuecomment-359282882).  I'm waiting for the Gitlab build of the current revision to isolate possible GPU-specific issues clearly if any.

bogdan-tanygin:
Waiting for the Gitlab build.. Is it automatic, daily/nightly?

KaiSzuttor:
it wont start for non-espressomd members

bogdan-tanygin:
I will investigate the issues and feedback.

bogdan-tanygin:
There are 2 environment specific issues:
docker image `espresso-ubuntu-clang-cuda`: the stalled `dawaanr-and-bh-gpu` issue
docker image `espresso-ubuntu-cuda`: the BH energy calculation issue

My Tesla C2070 env does not reproduce both. I agree with an official support [remark](https://github.com/espressomd/espresso/pull/1273#issuecomment-357986237) even though the issue is opposite now; it can be also the Tesla vs GeForce case. I will setup the new cloud environment to debug both types of an equipment.

bogdan-tanygin:
To have a fully clear experiment, could I download the above-mentioned docker images?

mkuron:
Sure, you can try the images on your own machine. Install Docker and [nvidia-docker](https://github.com/NVIDIA/nvidia-docker), then do
```
nvidia-docker --volumes=~/espresso:/espresso:rw run -it espressomd/espresso-ubuntu-cuda`
```
and build and test inside the container. The`--volumes` flag mounts the espresso directory from your home directory to /espresso in the container, so change it as needed.

RudolfWeeber:
I had a look at this. Actually, the deviation in energy is not architecture dependent. The reason it always occurs on gitlba ci is that the numpy random number generator used to generate the particle configuration is seede3d in the test.
You can reproduce this in other environments by removing the seed() call from the script and running it in a loop.
Probably introduce a numpy.savtxt() on s.part[:].pos and s.part[:].dip to save the particle config for which it occurs.
A possible cause might be that some particle configurations contain particles slightly outside the box. The direct summation wouldn't care, but a cell-based method like BH might.

bogdan-tanygin:
Thanks for the advice! I've already fixed the energy issue in other branch. It is related to a lack of 0 initial value assignment within the energy array.

Also, I've fixed the stalled `clang` build. It was related to the bottom BH line aligned to the WARP size.

Also, I've improved the error handling: both functional errors and sync threads originated CUDA errors (from previous steps which could be miss-reported in further kernel runs).

Now, I'm fixing a regression of the 5K+ particles work (also will check 1E6 particles :-) ). It might be related to a number of BH nodes reserve.

Each of my runs I perform on Espresso cuda dockers (both clang and gcc) through the Google cloud platform with NVIDIA Tesla K80.

bogdan-tanygin:
The above-mentioned issues are, probably, fixed now for 1E3 - 1E5 particles on different dockers.

The environment: _Tesla K80 in the Google cloud_.

The epoch time duration of the execution of `es.integrator.run(steps = 0,recalc_forces = True)` with the corresponding dipolar actor enabled is reported below with a comparison of the BH and DDS GPU dipolar actors performance and the number of particles dependence. The calculations are done using the test `dds-and-bh-gpu-perf.py` which is committed, however, not added to the `CMakeLists.txt`. If such "manual" long-run tests are needed then please keep it or advise a correct place.

_The docker `espresso-ubuntu-clang-cuda`:_

26487 particles
=== Performance comparison ===
dt_dds_gpu = 2.98461008072 s
dt_bh_gpu = 0.193454027176 s
147543 particles
=== Performance comparison ===
dt_dds_gpu = 91.4903919697 s
dt_bh_gpu = 0.692107915878 s

_The docker `espresso-ubuntu-cuda`:_

26487 particles
=== Performance comparison ===
dt_dds_gpu = 3.30993294716 s
dt_bh_gpu = 0.193707942963 s
147543 particles
=== Performance comparison ===
dt_dds_gpu = 102.147951126 s
dt_bh_gpu = 0.69380903244 s

bogdan-tanygin:
Thanks for the GitLab run. Concerning the requested smart pointers refactoring. Do you have some specific architecture design notes, or I could select a design? For instance, to init the smart pointer within the python actor class and to store the real pointer from it to the force and energy actors? Or the actor pointer list (cpp one, ActorList) should become a list of smartpointers now? Also, do we need to relate the python actor list and the cpp actor list (currently, probably, they have no relation)? Also, almost all actors (including CUDA ones) which currently are referenced from the python interface (at least ones really called from the testsuite) are integrated with an avoidance of a usage of the cpp `class ActorList : public std::vector<Actor*>`; is this cpp list deprecated or this is just a happenstance? An exception is just the DDS GPU which I've cloned.

RudolfWeeber:
Since the energy-related issue is sorted out and the tests pass, I suggest to merge.

fweik:
Tank you for all the work you've put in. This is a nice new feature!

PR #1285, opened 2017-08-03
Parallel testsuite

### LINKED ISSUE #1195: Python testsuite not run in parallel
The test suite seems to be run only on a single core on Travis. It is also not clear to me, whether the CMAke based testsuite even allows for running the tests in parallel.

fweik:
Fixes #1195 

Description of changes:
 - 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


hmenke (approved):

KaiSzuttor:
Hangs on coulomb_cloud_wall with maxset.hpp (and maybe other featuresets as well) if I test locally.

fweik:
That seems to be the P2NFFT tuning, which is out of our control.... Fixing tha cao and grid in the test did not help.

fweik:
I'm not sure if testing p2nfft and the p3m from scafacos makes sens, the methods are tested within scafacos, and to test the bindings one method should be enough. @RudolfWeeber what do you think about that?

fweik:
Aaaaand there is another mpi deadlock in the energy calc, will fix.

fweik:
@KaiSzuttor should now work (w/o scafacos)

RudolfWeeber:
> I'm not sure if testing p2nfft and the p3m from scafacos makes sens, the methods are tested within scafacos, and to test the bindings one method should be enough. @RudolfWeeber what do you think about that? 
If both of them either do or don't use short range delegation, only one
needs to be tested. If they do the short range differently, both.

PR #1294, opened 2017-08-08
Getting rid of the _system construct.

### LINKED ISSUE #1284: Hide private variables and methods in sphinx docs

KaiSzuttor:
Fixes #1284

Description of changes:
 - Removed the python module 'system' and modified the cython class _system, now named system appropriately


hmenke (approved):
Now we got rid of the stupid `@highlander` but there is still a singleton for the `system` class.  Is there a reason we can't have several instances in one script?

KaiSzuttor:
There is still a lot of core code that would cause trouble if we'd allow multi instances of system.
I.e. i think we cant until everything is object oriented in the core.

fweik (approved):

PR #1296, opened 2017-08-08
Some sphinx fixes and removed private and special members from sphinx…

### LINKED ISSUE #1284: Hide private variables and methods in sphinx docs

KaiSzuttor:
… documentation.

Fixes #1284

### DIFF
--- a/doc/sphinx/io.rst
+++ b/doc/sphinx/io.rst
@@ -150,13 +150,15 @@ re-registered when the same checkpoint id is used later.
 Following the example above, the next example loads the last checkpoint,
 restores the state of all checkpointed objects and registers a signal.
 
-.. code:: python
+.. code::
+
     import espressomd from espressomd import checkpointing import signal
 
-    checkpoint = checkpointing.Checkpointing(checkpoint\_id=“mycheckpoint”)
+    checkpoint = checkpointing.Checkpointing(checkpoint_id=“mycheckpoint”)
     checkpoint.load()
 
-    system = espressomd.System() system.cell\_system.skin = skin
+    system = espressomd.System()
+    system.cell_system.skin = skin
     system.actors.add(p3m)
 
     #signal.SIGINT: signal 2, is sent when ctrl+c is pressed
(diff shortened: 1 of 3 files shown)


mkuron:
How does this deal with `ParticleSlice`? Since #1160, most properties are dynamically created, only `pos_folded`, `add_exclusion` and `delete_exclusion` are statically declared in the base class, `_ParticleSliceImpl`.

KaiSzuttor:
@mkuron   I don't quite get your point, this PR does not change anything of the documentation behavior besides removing private and special members.

mkuron:
Isn't `_ParticleSliceImpl` private with that underscore at the beginning?

KaiSzuttor:
But aren't these properties documented in the ParticleHandle class?

mkuron:
Only the ones that are automatically added, not the statically-declared ones in the base class. Just look at the sphinx documentation to see whether they show up with documentation or not.

KaiSzuttor:
As far as I can see they show up in the documentation. Please counter-check yourself @mkuron .

KaiSzuttor:
Since sphinx also includes inherited members, the ones @mkuron was mentioning are included.

doc/sphinx/io.rst
```diff
@@ -150,13 +150,15 @@ re-registered when the same checkpoint id is used later.
 Following the example above, the next example loads the last checkpoint,
 restores the state of all checkpointed objects and registers a signal.
 
-.. code:: python
+.. code::
+
     import espressomd from espressomd import checkpointing import signal
 
-    checkpoint = checkpointing.Checkpointing(checkpoint\_id=“mycheckpoint”)
+    checkpoint = checkpointing.Checkpointing(checkpoint_id=“mycheckpoint”)
```
mkuron: There are smart quotes used here (`“”`) instead of straight ones (`""`), which makes this invalid Python code. Since you're touching the line anyway, please fix this.

mkuron (changes requested):

hmenke (approved):

fweik:
So we're good now?

PR #1299, opened 2017-08-09
Virtual sites relative integration test

RudolfWeeber:
Tests placement, velocity of vs, back-transfer of forces from vs to governing particles,
Fluid of dumbells made up of 2 virtual lj spheres per dumbell. Verify forces by manual summation, to make sure nothing is lost/outdated in paallel simulations.

The lj force verification code is in tests_common.py and can be re-used by other tests. (@rempferg)

ParticleHandle.vs_auto_relate_to() now makes the particle virtual auomatically.
Variant of System.distance(part1,part2) that returns a vectorial distance (System.distance_vec())


fweik:
Please resolve merge conflict

fweik (approved):

PR #1308, opened 2017-08-11
Enlarge your DG with one simple trick

hmenke:
:joy: :ok_hand: :100:

### DIFF
--- a/doc/sphinx/dg.rst
+++ b/doc/sphinx/dg.rst
@@ -1,3 +1,4 @@
+=================
 Developer's Guide
 =================
 .. warning::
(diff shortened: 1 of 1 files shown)


doc/sphinx/dg.rst
```diff
@@ -1,3 +1,4 @@
+=================
```
KaiSzuttor, line 1: not needed, see http://docutils.sourceforge.net/docs/user/rst/quickref.html#section-structure

doc/sphinx/dg.rst
```diff
@@ -195,7 +196,7 @@ developers.
 Source code structure
 ---------------------
 The source tree has the following structure:
-* src: The actual source code
+  * src: The actual source code
```
KaiSzuttor: that indentation was on purpose:
src
  core
  python/espressomd
.
.
.

KaiSzuttor (changes requested):

doc/sphinx/dg.rst
hmenke: I see.  Sphinx was complaining.

doc/sphinx/dg.rst
hmenke: Totally needed.  Compare the screenshot to http://espressomd.org/html/doc/index.html

screenshot from 2017-08-11 22-45-04

KaiSzuttor: The current online documentation does not yet correspond to the current pyhton branch. I havent checked if the overline changes sth but if so the doc of sphinx is wrong.

PR #1309, opened 2017-08-15
[tutorials] Preliminary Python version of 06-active_matter

### LINKED ISSUE #1242: No Python version of active matter tutorial

hmenke:
Fixes #1242


jdegraaf:
I've worked my way through your pull request, however, there is still something wrong. Running pypresso on the script:
```
rectification_simulation.py 0.0
```
or using any other value of the velocity, results in the following error during integration.
https://gist.github.com/anonymous/c1f77ed855ca1ca64584764739710ad1
I'll give a full description via another route.

hmenke:
I can reproduce with the following minimal example
```python
import espressomd
s = espressomd.System()
s.part.add(pos=[0,0,0])
s.part.writevtk("abc.vtk")
```
but only with `maxset.hpp` and running in `gdb`.  The debugger points to these lines in `ParticleProperties` https://github.com/espressomd/espresso/blob/ee9bbbc4576713d082b90a8002ad4751e3b31940/src/core/particle_data.hpp#L212-L215
In the destructor of `ParticleProperties` their allocator segfaults on freeing memory (why?).

See issue #1327

jdegraaf:
So, can we conclude that the issue is not the way the tutorials are implemented, hence, the pull request for new tutorials can be put through? We'll have to

- double check the tutorials for the school
- make a new Active_Matter git label

at some point in the future anyway. And, as far as I gather, the first point can only be done once the above problem is fixed. Still, from a review point, all of the tutorials put together by Henri look convincing and I expect the last one to run smoothly too, whenever the bug is fixed.

PR #1328, opened 2017-08-16
[tutorials] Preliminary version of 09-catalytic_reactions

hmenke:
Everything seems to work fine now.


jonaslandsgesell:
Hi,
unfortunately there is a fundamental misunderstanding involved in the "feature" which is called catalytic reactions in Espresso (since always?). A catalyst does not influence where the equilibrium lies. A catalyst only influences the reaction dynamics. Do not merge until this fundamental Problem in this "feature" (description) is removed.

jdegraaf:
I see what you mean. I think the problem is the phrasing, rather than the actual implementation. I'll rewrite the tutorial accordingly.

fweik:
The ParticleProperty issue is fixed in #1334

fweik:
So can this be merged now?

jdegraaf:
Hello Florian,

Yes, I think so. I want to test it once more two weeks before the school
starts and then have an Active_Matter label put in to a specific git hash,
such that the students can use it. Or, will everything be merged into a new
version of ESPResSo, before then?

Best Wishes,

Joost

hmenke:
I have tested the tutorial again and everything seems to work fine.  The output is visualized below.  I'd say that this can be merged now.  Since we will be working on the `python` branch or a new 4.0 release during the school there is no need to tag a new Active_Matter label and the old one can be removed.

Averaged MSD over five runs with standard error on the error bars for both, the active and the passive system:
msd

The AVACF for the same system as in the previous figure.  Note that the activity does not influence the rotational behaviour.
avacf

jonaslandsgesell:
Is the old (described) behaviour still there, that "H2O2 particles" are split with a certain rate when in vicinity to a catalyst but also recrated in the "Bulk" with another rate?
Now the description does not say so anymore but this behaviour would still be unphysical in a thermodynamic sense. It implies that in the "bulk" "H2O2" is created spontaneously and H2O2 would also be consumed and split by the catalyst spontaneously. However only one of those two reactions can happen spontaneously.

jdegraaf:
Dear Jonas,

No it is not. The source code is still present -- this may be reused for
something else, besides it being the standard (albeit nonphysical) way of
handling this kind of catalytic reactive modelling for self-propelled
particles. However, this source code can only be accessed by changing the
interface code, so for a regular user it is inaccessible. Nor is it
mentioned in the documentation, so that a regular user would be unaware of
the existence of this piece of source code. Only the swap move remains
directly accessible and is the one documented in the UG and tutorials. It
is a model for the specific surface reaction on a self-propelled Janus
particle, as is now clearly detailed in the two pieces of documentation
(tutorial and UG).

Best Wishes,

Joost

PR #1334, opened 2017-08-16
Part copy fix

### LINKED ISSUE #1327: Segmentation fault due to std::vector in ParticleProperties
The following minimal example produces a segmentation fault
```python
import espressomd
s = espressomd.System()
s.part.add(pos=[0,0,0])
s.part.writevtk("abc.vtk")
```
but only with `maxset.hpp` and running in `gdb`.  The debugger points to these lines in `ParticleProperties` which were introduced in #1201 https://github.com/espressomd/espresso/blob/ee9bbbc4576713d082b90a8002ad4751e3b31940/src/core/particle_data.hpp#L212-L215
In the destructor of `ParticleProperties` their allocator segfaults on freeing memory (why?).

Here is a full backtrace https://gist.github.com/anonymous/a563be9c515b2a1e3913a1ff68d9dcc3

I'm building on Ubuntu 16.04 with compilers and libraries from the official repos.

fweik:
Fixes #1327 

Description of changes:
 - 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


RudolfWeeber:
can be merged

PR #1344, opened 2017-08-16
Make system.part.add return the ParticleHandle; document some particle stuff

hmenke:
Returning the `ParticleHandle` is useful because then you don't have to do tricks like
```python
pid = len(system.part)
system.part.add(id=pid,...)
```
anymore to get the id of a new particle.  Sometimes you need the id, e.g., when you want to compute observables like
```python
p = system.part.add(...)
pos_id = ParticlePositions(ids=[p.id])
```


RudolfWeeber:
> Returning the particle id is useful because then you don't have to do tricks like
Can we please return the ParticleHandle instead?

Actually, I implemented that already. No idea, where that went
missing...

hmenke:
@RudolfWeeber Yes, we can.  I'll update the PR tomorrow (NZST).

hmenke:
@RudolfWeeber Can this be merged now?

KaiSzuttor:
@hmenke can you merge the current python branch and push, so that ICP CUDA build gives it's OK too?

fweik:
@KaiSzuttor you can push to his branch if you want to.

hmenke:
I have rebased my branch onto python.  The CUDA and Travis builds succeed.

KaiSzuttor (approved):

PR #1346, opened 2017-08-17
core: Removed LJ_ANGLE

fweik:
Should not be implemented like this, not properly documented, not accessible from the interface.


fweik:
This can be implemented with virtual sites an a directional interaction, which is cleaner and more flexible. The LJ_ANGLE thing is more a hackish one-time thing.

pkreissl (approved):

PR #1349, opened 2017-08-21
Integration test for Rattle (BOND_CONSTRAINT)

RudolfWeeber:
Also allow BOND_CONSTRAINT together with ROTATION, include into maxset, and fix Py interface for RigidBond
The test simulates a bead-rigid-bond chain under Langevin thermostat. Then checks distance between beads aand projection of velocity differences onto the bond vector.


src/python/espressomd/interactions.pyx
```diff
@@ -958,7 +958,7 @@ IF BOND_CONSTRAINT == 1:
                             "vtol": 0.001}
 
         def _get_params_from_es_core(self):
-            return {"r": bonded_ia_params[self._bond_id].p.rigid_bond.r, "ptol": bonded_ia_params[self._bond_id].p.rigid_bond.ptol, "vtol": bonded_ia_params[self._bond_id].p.rigid_bond.vtol}
+            return {"r": bonded_ia_params[self._bond_id].p.rigid_bond.d2**0.5, "ptol": bonded_ia_params[self._bond_id].p.rigid_bond.p_tol, "vtol": bonded_ia_params[self._bond_id].p.rigid_bond.v_tol}
```
hmenke, line 961: `sqrt(x)` is much faster than `x**0.5`.

fweik (approved):

PR #1350, opened 2017-08-21
Langevin Thermostat test

RudolfWeeber:
Tests the Langevin thermostat by comparing the obtained velocity distribution to the single-component Maxwell distribution.
Also covers LANGEVIN_PER_PARTICLE, if compiled in


fweik (approved):

fweik:
@RudolfWeeber new test has issues on travis

RudolfWeeber:
On Thu, Aug 31, 2017 at 08:05:43PM +0000, Florian Weik wrote:
> @RudolfWeeber new test has issues on travis
Looks like particles getting lost in parallel simulations:
On one core it passes.

For the global test (all particles analyzed at once)
1 cores: passes
2 cores:
  File "langevin_thermostat.py", line 78, in test_global_langevin
      v_stored[i*N:(i+1)*N,:]=s.part[:].v
ValueError: could not broadcast input array from shape (107,3) into
shape (200,3)

3 cores:
ValueError: could not broadcast input array from shape (135,3) into
shape (200,3)

4 cores:
ValueError: could not broadcast input array from shape (104,3) into
shape (200,3)

So it seems like s.part[:] misses some of the particles in the system.

The underlying code uses max_seen_particle and particle_exists(),
which in turn uses the particle_node[] structure.

hmenke (approved):

fweik (approved):

fweik:
The tolerances are too tight. There also seems to be an other problem, which I'm currently investigating.

PR #1352, opened 2017-08-22
Travis builds for SUSE and CentOS

mkuron:
After we found that Espresso didn't build on RedHat at the group retreat and since we are currently switching our main development environment away from openSUSE, I have added Travis jobs for these two distributions. I only do builds and don't run the testsuite (only a single test instead) as that seems to be sufficient to ensure that the build system remains portable. We're not looking for compiler or library bugs, after all.


mkuron:
This pull request also fixes two bad regressions @fweik introduced (and both @hmenke and myself overlooked) in #1336: all Travis builds used the default myconfig because out-of-source builds were broken and the `docker run` exit code did not reflect build success because the coverage upload happened after it.

.travis.yml
```diff
@@ -2,50 +2,56 @@ language: generic
 
 matrix:
   include:
-    - os: linux
```
fweik: That does not look correct ^^

mkuron: That's why there was a _WIP_ in the title...

mkuron:
Ready to review now... The Travis build will not go green until @fweik creates the autobuild on Docker Hub. But it succeeded on the previous commit, where it was still using my own Docker Hub.

maintainer/travis/build_docker.sh
```diff
@@ -19,7 +19,7 @@ fi
 
 if [ "$TRAVIS_OS_NAME" = "linux" ]; then
 	image=espressomd/buildenv-espresso-$image:python
-	docker run -u espresso --env-file $ENV_FILE -v ${PWD}:/travis -it $image /bin/bash -c "git clone --recursive /travis && cd travis && maintainer/travis/build_cmake.sh"
+	docker run -u espresso --env-file $ENV_FILE -v ${PWD}:/travis -it $image /bin/bash -c "cp -r /travis .; cd travis && maintainer/travis/build_cmake.sh" || exit 1
```
mkuron: I had to switch this to `cp` because CentOS contains a Git version that does not support cloning shallow repositories (Travis appears to have switched to shallow some time ago). The `&&` needed to be replaced with `;` because there is a hidden file created by travis in there that cannot be copied.

maintainer/travis/build_cmake.sh
```diff
@@ -74,17 +74,21 @@ fi
...
-if [ ! $insource ]; then
+if ! $insource; then
     if [ ! -d $builddir ]; then
         echo "Creating $builddir..."
         mkdir -p $builddir
     fi
 fi
 
-if [ ! $insource ]; then
+if ! $insource; then
```
mkuron: This fixes the myconfig copying for out-of-source builds.

maintainer/travis/build_cmake.sh
```diff
@@ -74,17 +74,21 @@ fi
...
     cd $builddir
 fi
 
+# load MPI module if necessary
+grep -q suse /etc/os-release && source /etc/profile.d/modules.sh && module load gnu-openmpi
+grep -q rhel /etc/os-release && source /etc/profile.d/modules.sh && module load mpi
```
mkuron: Manual `source`ing of the profile seems to be necessary because `docker run` does not produce a login shell.

maintainer/travis/build_docker.sh
```diff
@@ -45,10 +45,10 @@ elif [ "$TRAVIS_OS_NAME" = "osx" ]; then
 	brew install homebrew/science/hdf5 --with-mpi --without-cxx
 
 	export TMPDIR=/tmp
-	maintainer/travis/build_cmake.sh
+	maintainer/travis/build_cmake.sh || exit 1
```
mkuron: These `exit`s are needed because the build script is no longer the last thing called in the script. Without it, all Travis builds would seemingly pass.

maintainer/docker/centos/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM centos:7
+MAINTAINER Florian Weik <fweik@icp.uni-stuttgart.de>
```
hmenke: Better put your own name there.

maintainer/docker/opensuse/Dockerfile
```diff
@@ -1,21 +1,19 @@
-FROM opensuse:42.1
+FROM opensuse:42.3
```
hmenke: `FROM opensuse:leap`

maintainer/docker/opensuse/Dockerfile
```diff
@@ -1,21 +1,19 @@
...
 && zypper -n --gpg-auto-import-keys install \
-  patterns-openSUSE-devel_C_C++ \
+  patterns-openSUSE-devel_C_C++ gcc-c++ \
   cmake \
-  openmpi-devel \
+  openmpi-devel Modules \
   fftw3-devel \
   boost-devel libboost_serialization1_54_0 libboost_mpi1_54_0  libboost_filesystem1_54_0 libboost_test1_54_0 \
-  python python-numpy python-PyVTK python-h5py\
-  tcl-devel \
```
hmenke: Without `tcl-devel` we can no longer build the master branch with this image.

hmenke (approved):

fweik:
I'm going to move the docker files to a separate repository tomorrow and adapt the autobuilds so that the the docker update can be triggered independently. I've been meaning to do that anyway and it makes stuff easier.

maintainer/docker/opensuse/Dockerfile
mkuron: We don't need it here as the opensuse image was never used with master. But you're right, it needs to stay in the Ubuntu images. I think @fweik removed it from there a while ago.

maintainer/docker/opensuse/Dockerfile
mkuron: I did that on purpose so the version is fixed, just like with ubuntu (16.04), debian (8) and centos (7).

maintainer/docker/opensuse/Dockerfile
fweik: Can also be solved by a repository for the docker files. Than we can branches mirroring those of the code repro that go to different container tags.

fweik:
There appears to be a problem with OpenSuSE:
https://hub.docker.com/r/espressomd/espresso-opensuse/builds/bew95sesyyscvon4dtud8yu/

mkuron:
@fweik, you need to merge this pull request first before that will work. It was still building the old Dockerfile, which contained packages like python-PyVTK that no longer exist.

mkuron:
~@fweik, you need to merge this pull request first before that will work. It was still building the old Dockerfile, which contained packages like python-PyVTK that no longer exist.~

Just noticed you're already building from https://github.com/espressomd/docker. You might just need to restart the build so that it picks up the correct Dockerfile from there. I can see that it was still using the old one.

mkuron:
Also, can you please get rid of the two branches on https://github.com/espressomd/docker ? There is no need for that, we can easily create Dockerfiles that can build both Tcl and Python.

fweik (approved):

PR #1354, opened 2017-08-23
tests: use absolute paths to data.

KaiSzuttor:
Makes it possible to directly call the tests.


mkuron (approved):

RudolfWeeber:
Could you please outsource the code generating the test data pth to tests_copmmon.py, since this will be needed in more tests in the future.

Beyond that, it can be merged.

fweik (approved):

PR #1363, opened 2017-09-01
Nsquare load

fweik:
Fixes #1350 

Description of changes:
 - Append particles to destination PartiicleList in recv_particles instead of overwriting
   it, which is the needed behavior for the nsqure cell system.

PR Checklist
------------
 - [x] Tests?
   - [x] Interface
   - [x] Core 
 - [x] Docs?

### DIFF
--- a/src/core/communication.cpp
+++ b/src/core/communication.cpp
@@ -195,7 +195,8 @@ static int terminated = 0;
   CB(mpi_thermalize_cpu_slave)                                                 \
   CB(mpi_scafacos_set_parameters_slave)                                        \
   CB(mpi_scafacos_free_slave)                                                  \
-  CB(mpi_mpiio_slave)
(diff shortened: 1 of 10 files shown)


RudolfWeeber:
This indeed fixes the thermostat test with nsquare. 
It seems like the performance is degraded by a factor of at least 2,
probably as a consequence of the boost serialization used in
producing the send buffers. 

On a different note:
I merged this and the langevin thermostat into the short range pr.
For 1 core, everything passes.
for 2 cores, everything nsquare related segfaults, and the pairs.py
domain decomposition test fails.

The merge is as my github repo in the branch
shortrange_langevin_test_nsquare

RudolfWeeber:
The statement about timimngs was wrong. Could not reproduce on an
other system. Probably interaction with previous build. Sorry

fweik:
The updated version should be a little bit faster.

PR #1364, opened 2017-09-04
Cellsystems tests

fweik:
Fixes resorting bug in layered cs with pbc in z direction

Description of changes:
 - 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


RudolfWeeber:
Did you check that setting a new position does not already move the particle? New particles are directly created on the correct node.

Since the domain decomposition and layered tests only differ in a single place, but share the test code, could you merge them and put the testing code into a separate function which is then used by both test_... functions?

Beyond that, it can be merged.

fweik:
Yes I've checked that. Only new particles are directly placed on the correct node, for existing particles the position is just folded and updated.

PR #1366, opened 2017-09-05
mention ways to modify build process in documentation

jonaslandsgesell:


mkuron (changes requested):
Please remove all `$` signs (there are two more in other sections) and replace the C++-style comments (`//`) with proper Bash-style comments (`#`).

Also, the myconfig line needs to be more elaborate to be useful, e.g. like this:
```bash
cp myconfig-default.hpp myconfig.hpp # use the default configuration as template
nano myconfig.hpp                    # edit to add/remove features as desired
```

PR #1372, opened 2017-09-06
Fix Mayavi visualizer.

pkreissl:
The visualizer was not able to get the particle type. Fixed.

### DIFF
--- a/doc/tutorials/python/08-visualization/scripts/simulation.py
+++ b/doc/tutorials/python/08-visualization/scripts/simulation.py
@@ -1,5 +1,5 @@
 from __future__ import print_function
-import espressomd._system as es
+
 import espressomd
 from espressomd import visualization
 import numpy
--- a/doc/tutorials/python/08-visualization/scripts/visualization.py
+++ b/doc/tutorials/python/08-visualization/scripts/visualization.py
@@ -1,5 +1,5 @@
 from __future__ import print_function
-import espressomd._system as es
+
 import espressomd
 from espressomd import visualization
 import numpy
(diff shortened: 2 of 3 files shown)


doc/tutorials/python/08-visualization/scripts/simulation.py
```diff
@@ -1,5 +1,5 @@
 from __future__ import print_function
-import espressomd._system as es
+
```
KaiSzuttor, line 2: Can you please just remove that empty lines?

pkreissl: I usually leave a blank line after `__future__ `imports, as they are required to appear before any code (except docstrings) and in contrast to other imports are not script-specific -- just a part of the preamble.
I think this is also the way pep8 suggests (see https://www.python.org/dev/peps/pep-0008/#imports section 'Module level dunder names').

fweik (approved):

doc/tutorials/python/08-visualization/scripts/simulation.py
fweik: We have bigger fish to fry...

RudolfWeeber:
It seems to me like a matter of taste and therefore maybe beyoud the scope of QA.

doc/tutorials/python/08-visualization/scripts/simulation.py
KaiSzuttor: Thats why I just use autopep8 -i

PR #1387, opened 2017-09-10
Cluster analysis framework

RudolfWeeber:
This introduces a cluster analysis framework to Es. 
* Clusters are assembled based on the decision of a pair criterion, which tells if two particles are neighbors (pair energy, distance, presence of a bond, etc.)

* Works for parallel simulations, although the analysis is run on the head node

* A bunch of analysis routines for individual clusters is provided (radius of gyration, center of mass, ...)

* Integration tests for the cluster analysis and the pair criteria are provided.

* Docs are provided.

Collaboration with Milena Smiljanic (@smiljanic).


src/core/Vector.hpp
```diff
@@ -143,6 +145,24 @@ template <size_t n, typename Scalar> class Vector {
...
+Vector<n, Scalar> operator*(const Vector<n,Scalar> &v, const Scalar s) {
+    Vector<n,Scalar> res({0,0,0});
+    for (int i=0;i<n;i++){
+      res[i] =v[i] * s;
+    }
+    return res;
+}
+
+template<size_t n, typename Scalar>
+Vector<n, Scalar> operator/(const Vector<n,Scalar> v, const Scalar rhs) {
```
hmenke: Why do you not pass by ref?

src/core/Vector.hpp
```diff
@@ -203,12 +223,12 @@ bool operator!=(Vector<N, T> const &a, Vector<N, T> const &b) {
 }
 
 template <size_t N, typename T>
-Vector<N, T> operator+(Vector<N, T> const &a, Vector<N, T> const &b) {
+Vector<N, T> operator+(Vector<N, T> const a, Vector<N, T> const b) {
```
hmenke: Why do you not pass by ref?

src/core/Vector.hpp
```diff
@@ -203,12 +223,12 @@ bool operator!=(Vector<N, T> const &a, Vector<N, T> const &b) {
...
   return detail::binary_op(a, b, std::plus<T>());
 }
 
 template <size_t N, typename T>
-Vector<N, T> operator-(Vector<N, T> const &a, Vector<N, T> const &b) {
+Vector<N, T> operator-(Vector<N, T> const a, Vector<N, T> const b) {
```
hmenke: Why do you not pass by ref?

src/core/Vector.hpp
```diff
@@ -143,6 +145,24 @@ template <size_t n, typename Scalar> class Vector {
   }
 };
 
+template<size_t n, typename Scalar>
+Vector<n, Scalar> operator*(const Vector<n,Scalar> &v, const Scalar s) {
+    Vector<n,Scalar> res({0,0,0});
+    for (int i=0;i<n;i++){
```
hmenke: For that loop you could do some code deduplication.
```cpp
namespace detail {
    template<size_t N, typename T, typename Op>
    Vector<N,T> binary_scalar_op(Vector<N,T> v, T const s, Op const& op)
    {
        for (int i=0; i<n; ++i)
            v[i] = op(v[i],s);
        return v;
    }
}

template<size_t N, typename T>
class Vector {
    //...
    Vector operator*(Vector const& v, T const s)
    {
        return detail::binary_scalar_op(v, s, std::multiplies<T>());
    }

    Vector operator/(Vector const& v, T const s)
    {
        return detail::binary_scalar_op(v, s, std::divides<T>());
    }
    //...
};
```

testsuite/python/cluster_analysis.py
```diff
@@ -0,0 +1,188 @@
...
+
+    # 2nd cluster
+    es.part.add(id=4,pos=(0.5,0.5,0.5))
+    es.part.add(id=5,pos=(0.55,0.5,0.5))
+    
+
+    cs=ClusterStructure()
+
+    def test_00_fails_without_criterion_set(self):
+        self.assertRaises(self.cs.run_for_all_pairs())
```
hmenke: The usage of `assertRaises` is wrong.  You ought to use it like this:
```python
with self.assertRaises(Exception):
    self.cs.run_for_all_pairs()
```

testsuite/python/pair_criteria.py
```diff
@@ -0,0 +1,106 @@
...
+    p2=es.part[1]
+    # Setup purely repulsive lj
+    es.non_bonded_inter[0,0].lennard_jones.set_params(sigma=0.11,epsilon=1,cutoff=2**(1./6.)*0.11,shift="auto")
+    epsilon=1E-8
+
+    def test_distance_crit(self):
+        dc=DistanceCriterion(cut_off=0.1)
+        # Interface
+        self.assertTrue(dc.get_params().keys()==["cut_off",])
+        self.assertTrue(abs(dc.get_params()["cut_off"]-0.1)<self.epsilon)
```
hmenke: Please use
```python
self.assertAlmostEqual(a, b, places=7)
```
instead of
```python
self.assertTrue(abs(a-b)<self.epsilon)
```
Applies to all other comparisons with `epsilon` as well.

testsuite/python/pair_criteria.py
```diff
@@ -0,0 +1,106 @@
...
+    p1=es.part[0]
+    p2=es.part[1]
+    # Setup purely repulsive lj
+    es.non_bonded_inter[0,0].lennard_jones.set_params(sigma=0.11,epsilon=1,cutoff=2**(1./6.)*0.11,shift="auto")
+    epsilon=1E-8
+
+    def test_distance_crit(self):
+        dc=DistanceCriterion(cut_off=0.1)
+        # Interface
+        self.assertTrue(dc.get_params().keys()==["cut_off",])
```
hmenke: Instead of
```python
self.assertTrue(a==b)
```
please use
```python
self.assertEqual(a, b)
```
In the first case we only get the absolutely meaningless error message `AssertionError: False is not true`.  This applies to almost all uses of `assertTrue` in your code.

src/script_interface/pair_criteria/3
```diff
@@ -0,0 +1,62 @@
+/*
```
hmenke: Should this file have been added?  The filename is only the digit `3`.

fweik:
Please cherry-pick 0ed167b77981659a684ad271f4a612fae1b4dea7 for the vector ops, which also contains tests.

RudolfWeeber:
> Why do you not pass by ref?
Because otherwise
v =constant*v. 
and v=v+w
will segfault, as the operator result will overwrite the operands.
The - operator I overlooked.

fweik:
That is not the reason for the segfault (as operator+ returns by value), but a bug in detail::binary_op which is fixed in the mentioned commit that which you should cherry-pick :-)

src/core/pair_criteria/pair_criteria.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+
+
+namespace PairCriteria {
+
+/** @brief Criterion which provides a true/false for a pair of particles */
+class PairCriterion {
+  public: 
+    /** @brief Make a decision based on two Particle objects */
+    virtual bool decide(const Particle& p1, const Particle& p2) const {
```
fweik: Use 
```c++
 virtual bool decide(const Particle& p1, const Particle& p2) const = 0;
```
to make this a compile time error.

src/core/pair_criteria/pair_criteria.hpp
```diff
@@ -0,0 +1,94 @@
...
+      auto p2 =get_particle_data(id2);
+      const bool res =decide(*p1,*p2);
+      return res;
+    }
+};
+
+/** @brief True if two particles are closer than a cut off distance, respecting minimum image convention */
+class DistanceCriterion : public PairCriterion {
+  public: 
+    virtual bool decide(const Particle& p1, const Particle& p2) const {
```
fweik: Remove virtual, add override.

src/core/pair_criteria/pair_criteria.hpp
```diff
@@ -0,0 +1,94 @@
...
+    }
+    private:
+      double m_cut_off;
+};
+
+
+/** True if the short range energy is largern than a cut_off */
+class EnergyCriterion : public PairCriterion {
+  public: 
+    virtual bool decide(const Particle& p1, const Particle& p2) const {
```
fweik: Remove virtual, add override.

src/core/pair_criteria/pair_criteria.hpp
```diff
@@ -0,0 +1,94 @@
...
+      m_cut_off =c;
+    }
+    private:
+      double m_cut_off;
+};
+
+/** True if a bond of given type exists between the two particles */
+class BondCriterion : public PairCriterion {
+  public: 
+    virtual bool decide(const Particle& p1, const Particle& p2) const {
```
fweik: Remove virtual, add override.

src/core/cluster_analysis/Cluster.cpp
```diff
@@ -0,0 +1,132 @@
...
+ Vector3d com{};
+  
+  // The distances between the particles "folded", such that all distances
+  // are smaller than box_l/2 in a periodic system. The 1st particle
+  // of the cluster is arbitrarily chosen as reference.
+  
+  Vector3d reference_position=folded_position(partCfg()[particles[0]]);
+  Vector3d dist_to_reference;
+  double total_mass=0.;
+  for (int pid : particles)  //iterate over all particle ids within a cluster
```
fweik: I think this can be made less verbose by replacing the body by
```c++
auto const& p = partCfg()[pid];
com += p.p.mass * get_mi_vector(folded_position(p), reference_position);
total_mass += p.p.mass;
```

src/core/cluster_analysis/Cluster.cpp
```diff
@@ -0,0 +1,132 @@
...
+    com[i] = fmod(com[i],box_l[i]);
+  }
+  return com;
+}
+
+
+
+double Cluster::longest_distance() {
+  double ld=0.;
+  for (auto a=particles.begin();a!=particles.end();a++) {
```
fweik: The a==b case should be excluded, e.g. by using for_each_pair from #1391. I think something like
```c++
for_each_pair(particles, [&ld](int a, int b) {
 auto const& p1 = partCfg()[a];
 auto const& p2 = partCfg()[b];

  ld = std::max(ld, get_mi_vector(p1.r.p,p2.r.p).norm());
});
```
would me clearer. If this section is performance relevant, you should compare the squares (.norm2()).

src/core/cluster_analysis/Cluster.cpp
```diff
@@ -0,0 +1,132 @@
...
+  }
+  return ld;
+}
+
+
+//Radius of gyration
+double Cluster::radius_of_gyration() {
+  // Center of mass
+  Vector3d com=center_of_mass();
+  double sum_sq_dist=0.;
```
fweik: Maybe
```c++
auto const sum_sq_dist = std::accumulate(particles.begin(), particles.end(), 0, [](double sum, int pid) {
    auto const& p = partCfg()[pid];

   return sum + get_mi_vector(com, p.r.p).norm2();
});
```

src/core/cluster_analysis/Cluster.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+double Cluster::fractal_dimension(double dr, double& mean_sq_residual) {
+#ifdef GSL
+  Vector3d com = center_of_mass();  
+// calculate Df using linear regression on the logarithms of diameters [__std::vector<double> diameters__] and num of particles [__std::vector<int> pcounts__] within the diameters
+  
+  std::vector<double> distances;
+
+  for (auto const& it : particles) {
+    double dist[3];
```
fweik: I think
```c++
distances.push_back(get_mi_vector(com, partCfg()[it].r.p).norm());
```
would be clearer.

src/core/cluster_analysis/Cluster.hpp
```diff
@@ -0,0 +1,47 @@
...
+    double longest_distance();
+    /** @brief Calculate radius of gyration of the cluster */
+    double radius_of_gyration();
+    /** @brief Calculate the fractal dimension 
+    *  N(r) via r^d, where N(r) counts the number of particles in a sphere
+    *  of radius n, and d denotes the fractal dimension.
+    *  The fitting is done by the Gnu Scientific Library. 
+    *  @param dr:   increment for when constructing the discrete version of N(r)
+    *  @param mean_sq_residual:  Mean square residual returned by the fit */
+    double fractal_dimension(double dr, double& mean_sq_residual);
```
fweik: I'd prefer returning a pair over an output parameter.

src/core/cluster_analysis/ClusterStructure.cpp
```diff
@@ -0,0 +1,205 @@
...
+}
+
+// Analyze the cluster structure of the given particles
+void ClusterStructure::run_for_all_pairs()
+{
+  // clear data structs
+  clear();
+  
+  // Iteraeover pairs
+  for (int i=0;i<=max_seen_particle;i++) {
```
fweik: If you end up using partCfg anyway you can directly iterate over it, instead of checking if the ids exists,
e.g.
```c++
for_each_pair(partCfg, [](Particle const& a, Particle const& b) {
    add_pair(a, b);
});
```

src/core/cluster_analysis/ClusterStructure.cpp
```diff
@@ -0,0 +1,205 @@
...
+      add_pair(partCfg()[i],partCfg()[j]); 
+    }
+  }
+  merge_clusters();
+}
+
+void ClusterStructure::run_for_bonded_particles() {
+clear();
+partCfg().update_bonds();
+for (int i=0;i<=max_seen_particle;i++) {
```
fweik: for(auto const& p : partCfg)...

src/core/cluster_analysis/ClusterStructure.cpp
```diff
@@ -0,0 +1,205 @@
...
+  }
+}
+
+void ClusterStructure::merge_clusters() {
+  // Relabel particles according to the cluster identities map
+  // Also create empty cluster objects for the final cluster id
+  
+  // Collect needed changes in a separate map, as doing the changes on the fly
+  // would screw up the iterators
+  std::map<int,int> to_be_changed;
```
fweik: If I understand this correctly you only need lookup by id, and no order. Then you can use std::unordered_map, which has O(N) insert and lookup (std::map is ordered and has O(N*log(N)). I *think* this is also true for clusters.

src/core/cluster_analysis/ClusterStructure.cpp
```diff
@@ -0,0 +1,205 @@
...
+ int tmp=x;
+ while (m_cluster_identities.find(tmp)!=m_cluster_identities.end())
+ {
+  tmp =m_cluster_identities[tmp];
+ }
+ return tmp;
+}
+
+int ClusterStructure::get_next_free_cluster_id(){
+  //iterate over cluster_id'
```
fweik, line 181:
```c++
auto const max_seen_cluster = std::accumulate(cluster_id.begin(), cluster_id.end(), 0, [](int max_id, std::pair<int, int> &kv) {
   return std::max(max_id, kv.second);
});
```

fweik (changes requested):

src/script_interface/pair_criteria/3
```diff
@@ -0,0 +1,62 @@
...
+#include "../auto_parameters/AutoParameters.hpp"
+#include "core/pair_criteria.hpp"
+#include <string>
+
+namespace ScriptInterface {
+namespace PairCriteria {
+
+class PairCriterion : public AutoParameters {
+public:
+  PairCriterion() : m_c(new ::PairCriterion()) {};
```
fweik: If you turn this into
```c++
PairCriterion(std::shared_ptr<PairCriterion *ptr) : m_c(ptr) {}
```
PairCriterion can be abstract.

src/script_interface/pair_criteria/initialize.cpp
```diff
@@ -0,0 +1,37 @@
...
+*/
+
+#include "initialize.hpp"
+#include "ScriptInterface.hpp"
+#include "pair_criteria.hpp"
+
+namespace ScriptInterface {
+namespace PairCriteria {
+void initialize() {
+  ScriptInterface::register_new<ScriptInterface::PairCriteria::PairCriterion>(
```
fweik: It's not clear to me why the base class should be exposed to python.

src/script_interface/pair_criteria/pair_criteria.hpp
```diff
@@ -0,0 +1,110 @@
...
+#define SCRIPT_INTERFACE_PAIR_CRITERIA_PAIR_CRITERIA_HPP
+
+#include "../auto_parameters/AutoParameters.hpp"
+#include "core/pair_criteria/pair_criteria.hpp"
+#include <string>
+
+namespace ScriptInterface {
+namespace PairCriteria {
+
+class PairCriterion : public AutoParameters {
```
fweik, line 32: Please remove the data memeber from the base class, and make pair_criterion() pure.
This should then be implemented in the derived classes, and this class should not be exposed to python.
(This is the same solution as with the observables)

fweik:
Apart from my comments the core and script_interface parts look good to me.

src/core/cluster_analysis/Cluster.hpp
hmenke: Even better, return a `std::tuple`.  Then you can do
```cpp
double df, mean_sq_residual;
std::tie(df, mean_sq_residual) = c.fractal_dimension(dr);
```

src/core/cluster_analysis/ClusterStructure.cpp
hmenke: Since the indexing property of `std::map` is used nowhere, it might be better to use
```cpp
std::vector< std::tuple<int,int> > to_be_changed;
```
Then insertion is done via
```cpp
to_be_changed.emplace_back(std::make_tuple(a,b));
```
and iteration is done via
```cpp
for (auto const& it : to_be_changed)
{
    std::get<0>(it);
    std::get<1>(it);
}
```
For more info on why to prefer `std::vector`, watch [Taming the performance Beast - Klaus Iglberger - Meeting C++ 2015](https://www.youtube.com/watch?v=LrVi9LHP8Bk).

fweik: @hmenke The ids can have gaps...

src/core/cluster_analysis/Cluster.hpp
fweik: Also works with a pair. ^^

src/core/pair_criteria/pair_criteria.hpp
```diff
@@ -0,0 +1,92 @@
...
+    bool decide(const Particle& p1, const Particle& p2) const override {
+      // Distnace between particles
+      double vec21[3];
+      get_mi_vector(vec21,p1.r.p, p2.r.p); 
+      const double dist_betw_part =sqrt(sqrlen(vec21));
+      
+      // Interaction parameters for particle types
+      IA_parameters *ia_params = get_ia_param(p1.p.type, p2.p.type);
+      
+      return (calc_non_bonded_pair_energy(const_cast<Particle*>(&p1), const_cast<Particle*>(&p2), ia_params,
```
fweik: Please merge #1788 and remove const_cast.

RudolfWeeber:
Something still missing here?

fweik:
Yes. there is a sphinx problem:
```
/builds/espressomd/espresso/build/doc/sphinx/analysis.rst:658:u'any' reference target not found: espressomd.cluster_analysis.ClusterStructure.run_for_all_pairs
```
and needs merge. Otherwise this looks good to me.

fweik (approved):

PR #1390, opened 2017-09-11
script_interface: Added None type to Variant to indicate no value.

### LINKED ISSUE #1389: Script interface: add nullptr to Variant
such that it can be safely returned from call_method(). 
Right now, it seems to be converted to one of the other types in Variant. On OpenSuse, it doesn't result in an error, but in the Ubuntu docker image, it gets converted to a string, and results in a runtime error, because there is no conversion from nullptr.

fweik:
Fixes #1389 , Closes #1376.

Description of changes:
 - Added a None type to Variant to indicate no value
 - Helper functions and tests


testsuite/python/minimize_energy.py
```diff
@@ -30,21 +28,15 @@ def runTest(self):
...
- 
+
         self.system.integrator.run(10000)
         self.system.integrator.run(10000)
 
-        energy = self.system.analysis.energy() 
+        energy = self.system.analysis.energy()
 
-        assert( energy["total"] == 0 )
+        self.assertTrue( energy["total"] == 0 )
```
hmenke:
```python
self.assertEqual(energy["total"], 0)
```
The `self.assertTrue` function only gives the absolutely useless error message `AssertionError: False is not true`.  Please only `self.assertTrue(var)` where `var` is boolean.

src/script_interface/Variant.hpp
```diff
@@ -11,10 +11,28 @@ class ScriptInterfaceBase;
 using ObjectId = Utils::ObjectId<ScriptInterfaceBase>;
 
 /**
+ * @brief Type to indicate no value in Variant.
+ */
+class None {
```
hmenke: Why not just `using None = std::nullptr_t`?  It has the same properties as your `None` class.
```cpp
#include <cstddef>
#include <iostream>

using None = std::nullptr_t;

int main()
{
    None a{};
    None b{};
    std::cout << std::boolalpha
              << static_cast<void*>(a) << '\n' // 0
              << static_cast<void*>(b) << '\n' // 0
              << (a == b) << '\n'  // true
              << (a != b) << '\n'; // false
}
```
This will also allow you to just return `nullptr` from a function returning a `Variant` (which is what Rudolf wanted I think).
```cpp
#include <boost/variant.hpp>
#include <cstddef>

using None = std::nullptr_t;
using Variant = boost::variant<None, int, float>;

Variant foo() { return nullptr; }

int main() { Variant v = foo(); }
```

hmenke (changes requested):

src/script_interface/Variant.hpp
fweik: Because stuff is implicitly convertible to std::nullptr_t, e.g. 0

hmenke: That's irrelevant because for the literal `0` the constructor for `int` is an exact match, whereas the one for `std::nullptr_t` needs the conversion and therefore disregarded.
```cpp
#include <boost/variant.hpp>
#include <cstddef>
#include <iostream>

using None = std::nullptr_t;
using Variant = boost::variant<None, int, float>;

int main() {
    Variant a = 0;
    Variant b = nullptr;

    std::cout << a.which() << '\n'  // 1
              << b.which() << '\n'; // 0

    a = nullptr;
    b = 0;
    std::cout << a.which() << '\n'  // 0
              << b.which() << '\n'; // 1
}
```
Assignment does not trigger implicit conversion either but changes the type of the variant.  I see no problems here.

fweik: ~~http://en.cppreference.com/w/cpp/language/integer_literal
Observe how 0 is not an int literal.~~

hmenke: It *is* an integer literal.  It's even explained on the site you linked.

> *octal-literal* is the digit zero (0) followed by zero or more octal digits

fweik: Ok, but I don't completely follow why returning nullptr_t is needed. On the other hand it has the added advantage that this is convertible to bool and can be directly tested.

fweik: There was an unanticipated issue since boost.serialization marks nullptr_t as non-serializable, but Variant has to be serializable. So I think we'll just keep the None class, but I made it behave like nullptr_t, it can be implicitly converted from nullptr_t so that return nullptr; works, and it converts implicitly to bool so that if (None{}) works.

hmenke: Looks good.  The ability to return `nullptr` was what Rudolf requested in #1389 if I understood it correctly.

hmenke (approved):

src/script_interface/Variant.hpp
hmenke: I retract my earlier comment.  This does not compile:
```cpp
#include <boost/variant.hpp>
#include <boost/archive/text_oarchive.hpp>
#include <cstddef>

class None {
public:
  None() = default;
  None(std::nullptr_t &) {}

  bool operator==(None const &) const { return true; }
  bool operator!=(None const &) const { return false; }

  operator bool() const { return false; }
private:
  friend boost::serialization::access;
  template <typename Archive>
  void serialize(Archive &, long int /* version */) {}
};

using Variant = boost::variant<None, int, float>;

Variant foo() { return nullptr; }

int main() {
    Variant a = foo();
}
```

hmenke: I retract my earlier comment.  This does not compile.  It can be easily resolved by taking `std::nullptr_t` by value rather than reference.  Also, I made everything `constexpr` (because it is), made the `nullptr_t` constructor delegating (just in case), and would propose to rename `None` to `NoneType` and introduce a global compile-time variable `None` of `NoneType`, such that the pythonic C++ programmer can write `return None;` without having to worry about braces.
```cpp
#include <boost/variant.hpp>
#include <boost/archive/text_oarchive.hpp>
#include <cstddef>

class NoneType {
public:
  constexpr NoneType() = default;
  constexpr NoneType(std::nullptr_t) : NoneType{} {}

  constexpr bool operator==(NoneType const &) const { return true; }
  constexpr bool operator!=(NoneType const &) const { return false; }

  constexpr operator bool() const { return false; }
private:
  friend boost::serialization::access;
  template <typename Archive>
  void serialize(Archive &, long int /* version */) {}
};

using Variant = boost::variant<NoneType, int, float>;
constexpr NoneType None{};

Variant foo() { return nullptr; }

int main() {
    Variant a = foo();
}
```

hmenke (changes requested):

src/script_interface/Variant.hpp
fweik: The constructor should be const&, missed that.
~~Constexpr member functions are c++14 iirc, which is ok but then cmake and the docs have to be adjusted.~~ We can use the global, but it should adhere to the naming convention, so 'none' (also True vs. true). This also avoids name collisions in cython.

hmenke (approved):

PR #1392, opened 2017-09-12
These assertions are bolder than Donald Trump's speeches.  You won't believe it!

hmenke:
It occurred to me that a lot of testcases were using the `assertTrue` function for comparisons, like
```python
self.assertTrue(a == b)
```
This of course works but in case of failure this only produces the absolutely meaningless error message
```
AssertionError: False is not true
```
without telling you what the values to be compared actually were.  This makes debugging kind of tedious and especially you cannot quickly compare the values you got locally to what, e.g., Travis CI got.

Therefore I went over all the testcases and revised many of the assertions for easier understanding.  For more info on available assertions, please see the [`unittest.TestCase` documentation](https://docs.python.org/3.6/library/unittest.html#test-cases).


RudolfWeeber:
Can be merged

PR #1395, opened 2017-09-12
travis: More coverage.

fweik:
Do coverage during normal tests, and collect data for all configs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,11 @@ matrix:
     - os: linux
       sudo: required
       services: docker
-      env: myconfig=default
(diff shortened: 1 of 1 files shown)


hmenke (approved):
It's certainly sensible to cover more tests but the debug builds needed for coverage reports have longer build times and slower runtime performance when executing the tests.

mkuron:
@hmenke, it looks like the time increases by less than 20%, and that's hardly noticable in the build times that vary quite a bit on Travis anyway.

fweik:
The net effect is to speed up the test, because the number of tests is reduced.

PR #1401, opened 2017-09-13
Rotation implies rotation per particle

RudolfWeeber:
This pr aims at keeping Es' behavior unchanged when activating the ROTATION feature.
That should always have been the case, but wasn't.

* ROTATION now always controlled is per particle (removed ROTATION_PER_PARTICLE as separate 
feature) 
* Default is NOT to rotate.
* The tests are adapted

This constitutes an interface change.


fweik (approved):

fweik:
Looks good to me. Needs this change somehow be reflected in the documentation?

testsuite/particle.py
```diff
@@ -115,6 +115,13 @@ def func(self):
         test_mass = generateTestForScalarProperty("mass", 1.3)
 
     if espressomd.has_features(["ROTATION"]):
+        
+        for x in 0,1:
```
hmenke, line 119: Consider
```python
from itertools import product
for x,y,z in product([0,1],repeat=3):
```
as an alternative to nested loops.

hmenke (approved):

hmenke (approved):

src/core/rotation.cpp
```diff
@@ -448,6 +451,22 @@ void convert_omega_body_to_space(Particle *p, double *omega) {
              A[2 + 3 * 2] * p->m.omega[2];
 }
 
+Vector3d convert_vector_body_to_space(const Particle& p, Vector3d vec) {
```
hmenke:
```cpp
Vector3d convert_vector_body_to_space(const Particle& p, const Vector3d& vec) {
```

src/core/rotation.hpp
```diff
@@ -29,6 +29,12 @@
 #include "particle_data.hpp"
 #include "thermostat.hpp"
 #include "utils.hpp"
+#include "Vector.hpp"
+
+
+const int ROTATION_X =2;
+const int ROTATION_Y =4;
+const int ROTATION_Z =8;
```
hmenke:
```cpp
constexpr const int ROTATION_X = 2;
constexpr const int ROTATION_Y = 4;
constexpr const int ROTATION_Z = 8;
```

src/core/rotation.hpp
```diff
@@ -51,6 +57,9 @@ void convert_initial_torques();
 void convert_omega_body_to_space(Particle *p, double *omega);
 void convert_torques_body_to_space(Particle *p, double *torque);
 
+
+Vector3d convert_vector_body_to_space(const Particle& p, Vector3d v);
```
hmenke:
```cpp
Vector3d convert_vector_body_to_space(const Particle& p, const Vector3d &v);
```

hmenke (approved):

PR #1403, opened 2017-09-13
Fix SpheroCylinder.

pkreissl:
Description of changes:
 - Parameters of the SpheroCylinder shape can now be set via the promised keywords.
 - Member `direction` of SpheroCylinder can be set and setting `direction=-1` actually yields the expected result. The SpheroCylinder inherits from the Cylinder shape, which had no option to ignore the top and bottom caps. The caps thus also appeared inside the SpheroCylinder, dividing the cylinder part and the hemispherical cap parts. Fixed by this PR.
 - Adds the shape visualization of SpheroCylinders to the openGL visualizer.

### DIFF
--- a/src/core/shapes/Cylinder.cpp
+++ b/src/core/shapes/Cylinder.cpp
@@ -53,10 +53,10 @@ int Cylinder::calculate_dist(const double *ppos, double *dist,
   d_par = fabs(d_par);
 
   if (m_direction == -1) {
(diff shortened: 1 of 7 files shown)


fweik (approved):

pkreissl:
Last commit also fixes the length of the visualization of `Cylinders` in the openGL visualizer, which was twice the correct length.

hmenke (approved):

PR #1404, opened 2017-09-13
Thermostat cleanup

fweik:
Description of changes:
 - Removed some of the confusion regarding
   ROTATIONAL_INERTIA and PARTICLE_ANISOTROPY
   in the langevin thermostat.
 - The vector gamma and gamma_rot (global or per particle if
   LANGEVIN_PER_PARTICLE) is now turned on by PARTICLE_ANISOTROPY. 
 - Made per particle gamma{_rot} and langevin_gamma{_rot} Vector3d.

### DIFF
--- a/src/core/Vector.hpp
+++ b/src/core/Vector.hpp
@@ -55,7 +55,7 @@ public:
     std::copy(std::begin(v), std::end(v), d.begin());
   }
 
-  template <typename T> explicit Vector(T const(&v)[n]) {
+  explicit Vector(Scalar const(&v)[n]) {
     std::copy(std::begin(v), std::end(v), d.begin());
   }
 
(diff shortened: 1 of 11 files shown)


src/core/Vector.hpp
```diff
@@ -208,7 +214,85 @@ Vector<N, T> operator+(Vector<N, T> const &a, Vector<N, T> const &b) {
...
+Vector<N, T> operator*(Vector<N, T> const &b, T const &a) {
+  Vector<N, T> ret;
+
+  std::transform(b.begin(), b.end(), ret.begin(),
+                 [a](T const &val) { return a * val; });
+
+  return ret;
+}
+
+template <size_t N, typename T> void operator*=(Vector<N, T> &b, T const &a) {
```
hmenke: The assignment operators should return a reference to the result, see http://en.cppreference.com/w/cpp/language/operator_assignment

hmenke (approved):
I like the reduction of ifdef-ed code by introducing a single ifdef-ed typedef.

hmenke (approved):

src/core/particle_data.cpp
```diff
@@ -796,11 +796,11 @@ int set_particle_gamma(int part, double gamma[3])
   return ES_OK;
 }
 #ifdef ROTATION
-#ifndef ROTATIONAL_INERTIA
+#ifndef PARTICLE_ANISOTROPY
```
hmenke, line 799: This is clearly wrong.  I had initially corrected this in #995.

hmenke (commented):
I should have never approved this...

PR #1408, opened 2017-09-15
maintainer: Use build_cmake script in icp cuda builds.

fweik:
Fixes #

Description of changes:
 - 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?

### DIFF
--- a/maintainer/cuda_build.sh
+++ b/maintainer/cuda_build.sh
@@ -9,11 +9,7 @@ curl "https://api.github.com/repos/espressomd/espresso/statuses/$GIT_COMMIT?acce
       -d "{\"state\": \"$STATUS\", \"context\": \"ICP CUDA build\", \"target_url\": \"$URL\"}"
 
 nvidia-smi
-cd ${CI_PROJECT_DIR}
-cp maintainer/configs/maxset.hpp myconfig.hpp
-mkdir build && cd build
-cmake .. -DCUDA_NVCC_FLAGS="-D_FORCE_INLINES"
-make -j 4 && make check_python
+srcdir="${CI_PROJECT_DIR}" cmake_params="-D_FORCE_INLINES" with_coverage="true" maintainer/travis/build_cmake.sh
 result=$?
 echo $result
 if (( $result == 0 )); then
--- a/maintainer/cuda_build.sh
+++ b/maintainer/cuda_build.sh
@@ -9,7 +9,7 @@ curl "https://api.github.com/repos/espressomd/espresso/statuses/$GIT_COMMIT?acce
       -d "{\"state\": \"$STATUS\", \"context\": \"ICP CUDA build\", \"target_url\": \"$URL\"}"
 
 nvidia-smi
-srcdir="${CI_PROJECT_DIR}" cmake_params='-DCUDA_NVCC_FLAGS="-D_FORCE_INLINES"' with_coverage="true" maintainer/travis/build_cmake.sh
+srcdir="${CI_PROJECT_DIR}" cmake_params='-DCUDA_NVCC_FLAGS="-D_FORCE_INLINES"' with_coverage="true" build_procs=4 maintainer/travis/build_cmake.sh
 result=$?
 echo $result
 if (( $result == 0 )); then
--- a/maintainer/cuda_build.sh
+++ b/maintainer/cuda_build.sh
@@ -1,27 +1,4 @@
 #!/usr/bin/env bash
 
-GIT_COMMIT=$(git rev-parse HEAD)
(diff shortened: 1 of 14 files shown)


fweik:
Coverage is not yet working, this needs adding the codecov token to the secrets in gitlab, which I lack permission for.

src/python/espressomd/lb.pyx
```diff
@@ -40,7 +40,6 @@ IF LB_GPU or LB:
     cdef class LBFluid(HydrodynamicInteraction):
         """
         Initialize the lattice-Boltzmann method for hydrodynamic flow using the CPU.
-
```
KaiSzuttor, line 43: these empty lines are part of the numpydoc style recommondations

src/python/espressomd/lb.pyx
```diff
@@ -209,7 +208,6 @@ IF LB_GPU:
     cdef class LBFluid_GPU(LBFluid):
         """
         Initialize the lattice-Boltzmann method for hydrodynamic flow using the GPU.
-
```
KaiSzuttor, line 212: same here

src/core/communication.cpp
```diff
@@ -2076,12 +2076,6 @@ void mpi_bcast_lb_params(int field, int value) {
 }
 
 void mpi_bcast_lb_params_slave(int field, int value) {
-#if defined(LB) || defined(LB_GPU)
```
KaiSzuttor, line 2079: are these changes intended to be part of this PR?

src/core/unit_tests/ParallelScriptInterface_test.cpp
```diff
@@ -96,35 +83,37 @@ BOOST_AUTO_TEST_CASE(ctor_dtor) {
   }
 
   /* Check that ctor and dtor were run on all nodes */
-  Testing::reduce_and_check(callbacks->comm(), TestClass::constructed);
```
KaiSzuttor, line 99: Are these changes intended to be part of this PR?

KaiSzuttor (changes requested):
~~You can also delete the old build file:
https://github.com/espressomd/espresso/blob/8673af53e19202ed05e1f50010f21c7fa57eff75/maintainer/cuda_build.sh~~

Overall looks good to me, let's wait for the builds after the secret is placed into the gitlab project

KaiSzuttor:
You now have permission to add a secret variable to the gitlab project

src/core/unit_tests/ParallelScriptInterface_test.cpp
fweik: Yes, I turned out that the test was broken.

KaiSzuttor:
why is ICC failing? There is also a problem with the artifacts but that seems to be a problem of the gitlab server

fweik:
The problem with the artifacts is that they are now too big, because I switched from default to maxset.

fweik:
ICC fails some times because of the P3M tuning. The test should use appropriate fixed P3M parameters, ~~will fix.~~ Fixed in #1409.

fweik:
Apparently the max artifact size has to be configured globaly (c.f. https://gitlab.com/gitlab-org/gitlab-ce/issues/20612)

KaiSzuttor:
I just set it to 1 GB, just tested your branch and the build dir has approx. 400 MB

fweik:
Jup, works (https://gitlab.icp.uni-stuttgart.de/fweik/espresso/-/jobs/595)

KaiSzuttor:
lcov was missing in the docker image, just PR'ed to docker repo

KaiSzuttor (approved):

maintainer/cuda_build.sh
```diff
@@ -9,11 +9,7 @@ curl "https://api.github.com/repos/espressomd/espresso/statuses/$GIT_COMMIT?acce
       -d "{\"state\": \"$STATUS\", \"context\": \"ICP CUDA build\", \"target_url\": \"$URL\"}"
 
 nvidia-smi
-cd ${CI_PROJECT_DIR}
-cp maintainer/configs/maxset.hpp myconfig.hpp
-mkdir build && cd build
-cmake .. -DCUDA_NVCC_FLAGS="-D_FORCE_INLINES"
-make -j 4 && make check_python
+srcdir="${CI_PROJECT_DIR}" cmake_params='-DCUDA_NVCC_FLAGS="-D_FORCE_INLINES"' with_coverage="true" build_procs=4 maintainer/travis/build_cmake.sh
```
mkuron: I think `-DCUDA_NVCC_FLAGS="-D_FORCE_INLINES` can be removed. We now have that in CMakeLists.txt for NVCC version 8 and GCC version 5.

mkuron (approved):

fweik:
Concurrent builds seem to work too...

PR #1412, opened 2017-09-19
cuda: added cuda compile flags for Pascal compatability

KaiSzuttor:


fweik:
I'm pretty sure we can remove some of the older compute cap's. CUDA compilation already takes forever. If somebody really wants to build for 5 year old hardware, the flags can be added by hand.

fweik:
Also, are really both -gencode flags required?
```
-gencode=arch=compute_52,code=sm_52 -gencode=arch=compute_52,code=compute_52
```

KaiSzuttor:
This has to be decided by so. more experienced with these flags, I just copied missing flags from the doc:
http://docs.nvidia.com/cuda/pascal-compatibility-guide/index.html

fweik:
The last device with compute capability 2.0 was the GTX 5xx series, released in spring 2011.
(For reference: CC: https://developer.nvidia.com/cuda-gpus, release dates: https://en.wikipedia.org/wiki/List_of_Nvidia_graphics_processing_units)

rempferg:
Why do we specify that at all? Does nvcc not use some reasonable default, possibly even what the local machine needs?

rempferg:
Maybe we want to avoid the JIT compilation time, in case someone runs lots of super short simulations?

fweik:
default used to be
```
'compute_10' and 'sm_10'
```
which is horrible.
For nvcc 7.5 it's apparently sm_20 (https://stackoverflow.com/a/35657430/3198615)

mkuron:
Nowadays the default is 2.0. unfortunately though Maxwell and higher sometimes segfault when running 2.0 binaries, so we need 2.0 and 5.0 as a bare minimum, and 3.0 and 5.2 to make full use of Kepler and Pascal.

mkuron:
> This has to be decided by so. more experienced with these flags

As far as I know, the `code=compute_XX` flags are there to embed the PTX assembly for just-in-time compilation to run on newer hardware. So the latest architecture should always have its PTX embedded in addition to the binary (`code=sm_XX`).

PR #1415, opened 2017-09-20
Lb test

KaiSzuttor:
Cleaning up LB testcase and making the CUDA init handle a member of the system class.


fweik:
Can you please remove the LENNARD_JONES guards from the tests? They are no longer needed...

fweik (approved):

PR #1417, opened 2017-09-20
Adapt Warning in virtual_sites_relative code to python interface

pkreissl:


KaiSzuttor:
Shouldn't the core warnings be independent from the interface?

pkreissl:
I guess ideally they should. However, the suggestion to set the minimal cutoff via `setmd min_global_cut` doesn't make sense anymore. Thus the update.
If desired I could just delete the suggestion part, only leaving the warning itself.

RudolfWeeber:
Above all, core warnings should be helpful to the user. We are also currently only shipping one interface.
I'd keep the solution suggestion to avoid user frustration.

fweik:
LGTM

PR #1418, opened 2017-09-20
Forcecap

### LINKED ISSUE #625: Rework force capping mechanism to get rid of if-blocks in the individual potentials

fweik:
Fixes #625.

Description of changes:
 - Removed individual force capping mechanisms from potentials
 - Introduced global force capping that actually caps the force
 - Moved force capping from NonBondedInteractions to System in
   interface to reflect that it acts on all forces.

PR Checklist
------------
 - [x] Tests?
   - [x] Interface
   - [x] Core 
 - [x] Docs?


hmenke (commented):
You forgot to adapt this signature:

https://github.com/fweik/espresso/blob/64cd2ae962b6102285d48848f953243bda2696b4/src/python/espressomd/interactions.pyx#L395-L405

RudolfWeeber:
This is good to have.
Before merging, please document the force_cap property in the System class.
Also, the old code call on_short_range_ia_change() after setting a cap.
Your code doesn't, as far as I can tell. 
This is correct for your implementation. However, you probably still need to call
invalidate_obs();

Generally speaking, please do not apply clang-format and the like on changed code. It is very difficult to judge, whether something changed by accident in the force and energy functions.

fweik:
@RudolfWeeber ready for review.

RudolfWeeber:
Can be merged

PR #1420, opened 2017-09-20
Disable codecov comments

fweik:


hmenke (commented):
Maybe make all the tests deterministic rather than disabling the coverage comments.  Until then you might want to lower the threshold for failure due to a reduction of coverage.

fweik:
This just disables the comments posted by the bot, not the codecov gh intergration.

hmenke:
What happened to this one?  Why was it closed?

RudolfWeeber:
> What happened to this one?  Why was it closed?
Presumably, because there wasn't agreement on whether it should be
done.
What plan to do is to make the tests fully deterministic, to get rid
of the coverage fluctutations. It would be relevant to know, which
test is responsible for them, because that indicates that there are
two situations which need to be tested explicitly.

fweik:
Turned out that the change did disable all codecov reporting to github, which wasn't what I intended.

PR #1425, opened 2017-09-21
openGL Visualizer: Ignore non shape-based Constraints.

pkreissl:
Without this fix, the visualizer crashes if non ShapeBasedConstraint type constraints are part of the system


src/python/espressomd/visualization_opengl.pyx
```diff
@@ -220,13 +220,14 @@ class openGLLive(object):
         # Collect shapes and interaction type (for coloring) from constraints
         coll_shape_obj = collections.defaultdict(list)
         for c in self.system.constraints.call_method('get_elements'):
```
fweik:
```python
for c in self.system.constraints:
```
should work now.

PR #1431, opened 2017-09-21
sphinx: fixing issue of cython functions not being recognized as pyth…

KaiSzuttor:
…on code, thermostat decorator fix


fweik:
We could remove the python modules from the docker containers and run pip install requirements.txt to check that they are complete and sufficient to build. (decorator module is missing in docker)

mkuron:
@fweik, then we'd give up testing whether Espresso runs with the distribution (older) package versions. Since our requirements are rather trivial, we shouldn't require everyone to use requirements.txt.

mkuron:
@KaiSzuttor, `decorator.decorator`'s job is pretty trivial. We shouldn't need to add a module and intoduce a dependency just to do that, you can probably do it in five lines yourself.

hmenke (approved):

src/python/espressomd/thermostat.pyx
```diff
@@ -17,13 +17,14 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 from __future__ import print_function, absolute_import
+from decorator import decorator
```
mkuron: `decorator.decorator`'s job is pretty trivial. We shouldn't need to add a module and intoduce a dependency just to do that, you can probably do it in five lines yourself. [`decorator`](https://github.com/micheles/decorator/blob/master/src/decorator.py) uses `compile()`/`exec()` inside, so it's a pretty terrible hack anyway!

mkuron (changes requested):

fweik (approved):

src/python/espressomd/thermostat.pyx
```diff
@@ -17,37 +17,25 @@
...
-
-    def __call__(self, f):
-        def __f(*args, **kwargs):
-            if (not (thermo_switch in self.thermo_type) and
-                  (thermo_switch != THERMO_OFF)):
-                raise Exception("A different thermostat is already set!")
-            f(*args, **kwargs)
-        return __f
+def AssertThermostatType(*allowedthermostats):
+    def decoratorfunction(function):
```
hmenke: Perhaps you could have *not* removed the docstring.

hmenke (approved):

src/python/espressomd/thermostat.pyx
hmenke: Sorry, GitHub showed me an outdated file.

mkuron (approved):
Does this also work in Python 2.7? I think `functools.wraps` didn't treat docstrings until Python ~3.5.

KaiSzuttor:
yes

KaiSzuttor:
done, can be merged

PR #1441, opened 2017-09-26
Top ten reasons why you should be parsing PDB.

### LINKED ISSUE #1438: Python support for pdb parser
Is probably rather useful to many users and uncomplicated since it is implemented in the core.

hmenke:
```python
import espressomd
system = espressomd.System()
system.part.(pdb_file="ortho_dimer.pdb", type=0, first_id=0)
```
Fixes #1438

### DIFF
--- a/src/python/espressomd/particle_data.pyx
+++ b/src/python/espressomd/particle_data.pyx
@@ -1898,8 +1898,87 @@ cdef class ParticleList(object):
                     continue
                 yield (self[i], self[j])
 
-    def readpdb(self, pdb_file=None, itp_file=None, type=-1, first_id=-1, first_type=0,
-                lj_rel_cutoff=2.5, rescale_box=False, lj_internal=False, lj_diagonal=False, lj_with=None):
+    def readpdb(self,
+                # Required arguments
+                pdb_file=None, type=None, first_id=None,
+                # Optional arguments
+                itp_file='', first_type=0, lj_rel_cutoff=2.5, rescale_box=False,
+                lj_internal=False, lj_diagonal=False, lj_with=[]):
+        """
+        Read the positions of particles from a PDB file, optionally
+        with charges, types, and Lennard-Jones parameters from an ITP
+        Gromacs topology file.  The topology file must contain the
+        atoms and atomtypes sections, it may be necessary to use the
+        Gromacs preprocessor to obtain a complete file from a system
+        configuration and a force field.  Any offset of the particle
+        positions if removed, such that the lower left corner bounding
+        box of the particles is in the origin.
+
+        Notes
+        -----
+
+        Reading bonded interactions and dihedrals is currently not
+        supported.
+
+        Parameters
+        ----------
+
+        'pdb_file': string
+            filename of the PDB source
+        'type': integer
+            Sets the particle type for the added particles. If there
+            is a topology file give that contains a types for the
+            particles, the particles get types by the order in the
+            topology file plus firstype.  If the corresponding type in
+            the topology file has a charge, it is used, otherwise the
+            particle charge defaults to zero.
+        'first_id': integer
+            The particles get consecutive id’s in the order of the pdb
+            file, starting at firstid. Please be aware that existing
+            particles get overwritten by values from the file.
+        'itp_file': string, optional (default: empty)
+            filename of the ITP source
+        'first_type': integer, optional (default: 0)
+            Different species can be assigned different types.  The
+            types are numbered ascending from this value.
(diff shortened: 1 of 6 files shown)


mkuron (approved):

src/python/espressomd/particle_data.pyx
```diff
@@ -1896,6 +1898,128 @@ cdef class ParticleList(object):
...
+
+        >>> import espressomd
+        >>> system = espressomd.System()
+        >>> system.part.(pdb_file="ortho_dimer.pdb", type=0, first_id=0)
+
+        """
+
+        check_type_or_throw_except(pdb_file     , 1, str  , "pdb_file must be string")
+        check_type_or_throw_except(itp_file     , 1, str  , "itp_file must be string")
+        check_type_or_throw_except(type         , 1, int  , "type must be an integer")
```
mkuron, line 1985: `type` is a bad name for a Python variable as it's a keyword. You can overwrite the keyword like you've done, but that can be confusing at times.

mkuron (approved):

src/python/espressomd/particle_data.pyx
hmenke: True, but I didn't want to break compatibility in terms of naming.  You can also easily restore the original meaning by simply doing `del type`.

hmenke: We want to be consistent, right?
https://github.com/espressomd/espresso/blob/db8b559b756dd1617021df88364a9b3f3928f091/src/python/espressomd/particle_data.pyx#L85-L88

mkuron: Isn't this a new function? So you can choose the signature freely without compatibility concerns. I assume you did this to mimic the signature of `system.particle.add` -- though that does not have the overwriting problem because it only accepts a `type` argument through `**kwargs`.

mkuron: Properties named `type` don't have that problem because they don't pollute the local namespace with a `type` variable. They are only accessible through `self.type`. So if you want to name an rgument `type` for consistency, you should go through `**kwargs`.

fweik:
The python interface looks good to me. This should have a test, and it should be referenced in the io section of the docs, so that people can actually find it. In general I' not so sure if we should implement that ourself, as pdb parser are available e.g. from MDAnalysis. (The pdb parser in Espresso can also (partly) parse gromacs force fields, which there are also python packages for, e.g. ACPYPE).

On a more general note, I think the IO stuff should be implemented as a class hierarchy in c++,
which would make it e.g. possible to use the same unit tests and script interface for all methods.
This was something @KaiSzuttor was thinking about doing, but I don't know that the status on
that is. There is a couple of other io methods that could be integrated into that too (vtf, vtk, ...)

KaiSzuttor:
I haven't done anything on that yet, but I also like the idea of not reinventing the wheel and use external tools if available.

fweik (approved):

RudolfWeeber:
Do we still aim to re-activate the PDB-support in the Espresso core, or do we fully rely on the md analysis bindings?
@fweik, @jngrad?

jngrad:
I wrote a PDB parser during my PhD. Parsing my own PDB files was easy. The trouble started when other users tried using my tool with their PDB files. I spent countless hours improving the code to read custom PDB formats used by various modeling softwares. A lot of time also went into tech support, providing scripts to post-process PDB files with regexes (simple cases) or Python (desperate cases) and explaining the history behind specific details in the PDB standard (e.g. the iCode field). I'm not sure providing a PDB parser in ESPResSo is worth the inevitable tech support on the mailing list. We could instead use well-established tools like MDAnalysis or Biopython. The C++ PdbParser class also seems to expect [whitespace-separated values](https://github.com/espressomd/espresso/blob/a91dfd767fadb958df04b0917eb8bf48ff111f42/src/core/PdbParser.cpp#L63-L69), which is not guaranteed in the PDB standard, meaning some users will have to use MDAnalysis or Biopython anyway to strip extra data and transform HETATM into ATOM lines to comply to the "ESPResSo PDB" format. Konrad and I once proposed to have a tutorial explaining the general procedure to load a topology in ESPResSo, using the Gromacs .gro/.itp as example, but we couldn't find the time to actually do it.

fweik:
I did write the the PDB parser and I agree with @jngrad.

RudolfWeeber:
So, the conclusion from @jngrad and @fweik's comments is to remove the pdb parser from the core and close this PR, correct?

jngrad:
Yes. I will take care of removing the current parser code from the python branch.

PR #1442, opened 2017-09-26
h5md: Make sure coords are folded.

### LINKED ISSUE #1437: H5md does not write folded positions as required by the standard
Particle positions are stored unfolded although they should be stored folded (pos in the central box + image). This behaviour is required e.g. by the H5MD writer to follow the H5MD specification with the current implementation.

The h5md writer showed wrong behaviour regarding writing folded positions and their images after the following commit:

git bisect bad
8a90942879d98716bf8bb387c9c704619320ca89 is the first bad commit
commit 8a90942879d98716bf8bb387c9c704619320ca89
Date:   Thu May 18 17:14:27 2017 +0200

    h5md: Use partCfg to account for id gaps

:040000 040000 047349ede6a5bb6d7f54244c474b70e0956176a8 241b59f5159a646e840f869e019238741483d5f0 M	src

Using the following line in PartCfg.hpp fixes the unwanted behaviour and the behaviour from before May 18th is recovered (particle positions are stored folded in partCfg):
using PartCfg = ParticleCache<GetLocalParts, PositionFolder>;

However then the Correlation test seem to break.

fweik:
Fixes #1437 

Description of changes:
 - Always write folded positions in h5md writer.

### DIFF
--- a/src/core/grid.hpp
+++ b/src/core/grid.hpp
@@ -306,6 +306,10 @@ template <typename Particle> Vector3d folded_position(Particle const &p) {
   return pos;
 }
 
+inline void fold_position(Vector3d &pos, Vector<3, int> &image_box) {
+  fold_position(pos.data(), image_box.data());
+}
+
 /** unfold coordinates to physical position.
     \param pos the position
     \param pos the velocity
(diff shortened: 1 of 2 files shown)


jonaslandsgesell:
Looks good to me. I wrote a test (#1444) for the images which will work as soon as this PR is merged.

PR #1443, opened 2017-09-26
python: autopep8

KaiSzuttor:


RudolfWeeber:
This will make diffs difficult to read. Should we do this last before release?

KaiSzuttor:
I agree

fweik:
Let's do this when the chnages fir the release are complete. We have other large changesets scheduled for that phase, and it makes sense to do all of that at the same time. Will close for now.

PR #1444, opened 2017-09-26
add test for images in h5md

jonaslandsgesell:
Fixes # Introduces test for writing images in h5md. After #1442 is merged this test won't fail anymore.


fweik:
Is there a reason why the new the new tests are static again?

KaiSzuttor (commented):
The setUpClass methods are run only once per testclass and are by definition classmethods, whereas the setUp methods are run for every test (not needed here). So this should be okay.

KaiSzuttor (approved):

PR #1445, opened 2017-09-26
doxylink: added a hacked version of doxylink

KaiSzuttor:
With that hacked version, we can link from sphinx to doxygen. However what is still open to be discussed is the way I added the python module doxylink (it seems that it's not developed anymore).


mkuron:
I like the `pylint` solution for old-style class detection; I didn't know that existed, so I had used that regex solution instead.

Regarding doxylink: do you think it's a good idea to rely on something that hasn't been updated in five years? It could fail with a future Python 3 update, and then we're stuck maintaining it ourselves. Also, what's the "hacked" part about? Can't you monkey-patch whatever you needed instead of pulling in the entire code?

fweik:
In general I agree with Michael on the flaky external dependencies. But in this case I think we could do it, because the only thing that happens if this breaks is that links from sphinx to doxygen in the docs would no longer work. I don't think that this is a critical feature.

KaiSzuttor:
It's just a proof of concept... I checked the opensource world and there doesn't seem to be any project doing what I want here.

fweik:
Could you please put the pylint stuff in an extra PR. Also, pylint is missing on macOS on travis.

PR #1448, opened 2017-09-27
Update shapes.rst

jonaslandsgesell:
Fixes #

Description of changes:
 - 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?

### DIFF
--- a/doc/sphinx/shapes.rst
+++ b/doc/sphinx/shapes.rst
@@ -11,7 +11,9 @@ module. This module provides classes for the different available shapes::
 
 Shapes define geometries which can be used in |es| either as
 constraints in particle interactions or as a boundary for a
-Lattice-Boltzmann fluid.
+Lattice-Boltzmann fluid. If you just want to observe "normal" behaviour make
+sure all parts of your shape are within the central box since the distance to the
+shape is calculated only within the central box.
 
 Creating different shapes
 =========================
--- a/doc/sphinx/shapes.rst
+++ b/doc/sphinx/shapes.rst
@@ -11,9 +11,13 @@ module. This module provides classes for the different available shapes::
 
 Shapes define geometries which can be used in |es| either as
 constraints in particle interactions or as a boundary for a
-Lattice-Boltzmann fluid. If you just want to observe "normal" behaviour make
-sure all parts of your shape are within the central box since the distance to the
-shape is calculated only within the central box.
+Lattice-Boltzmann fluid. 
+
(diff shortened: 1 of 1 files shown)


hmenke (approved):

fweik:
I think you should better explain what you mean by "normal".

jonaslandsgesell:
In absence of a better wording I chose the word "normal" behaviour.

An example which shows what I mean is the following.
Try to constrain your particles in a cylinder. Then you might have the idea to put the cylinder center at the position (0,0,length/2). Of course you would place particles at the positions (x,y,z), with sqrt(x^2+y^2)<R.  If the cylinder is inpenetrable this will however result in constraint violated errors which is somehow not "normal". Since you placed all particles within a cylinder around the z-axis.
This comes from espresso only calculating the distance within the central box.

fweik:
I think you should better explain what you mean by "normal" in the document.

jonaslandsgesell:
I m open to ideas about that.

KaiSzuttor:
maybe just leave out “normal” and just describe what interaction is calculated

jonaslandsgesell:
Well the exact values depend on the Implementation. What more to say than
the statement that only the central box is used?

Am 28.09.2017 9:00 Vorm. schrieb "Kai Szuttor" <notifications@github.com>:

maybe just leave out “normal” and just describe what interaction is
calculated

—
You are receiving this because you authored the thread.
<https://github.com/espressomd/espresso/pull/1448#issuecomment-332746454>,
or mute the thread

KaiSzuttor:
It's just about nobody knowing what is ment with "normal". If you think the remaining explanation is sufficient, then just remove the "normal" stuff ;)

jonaslandsgesell:
Well a user might want to have exactly this "unusual" behaviour. Not in my
example... but if you want a cylinder without top or bottom you will make
it longer than the simulation box. Feel free to phrase that ;)

Am 28.09.2017 09:58 schrieb "Kai Szuttor" <notifications@github.com>:

davidsean:
You can reverse the sentence and caution against "unexpected" behaviour by placing the entire shape in the box.

Perhaps specify how the shape surfaces are truncated by the box boundaries and that care must be taken for when (sections of surfaces) end up in a periodic box.

jonaslandsgesell:
Thanks David. To mention the truncation is a good point. I just wanted to see this somewhere in the documentation!

PR #1452, opened 2017-09-27
Fix GPU LB with multiple MPI ranks

mkuron:
Fixes https://lists.nongnu.org/archive/html/espressomd-devel/2017-09/msg00000.html reported by @jdegraaf.


jdegraaf:
that was trivial \;-)

fweik (approved):

hmenke (approved):

PR #1453, opened 2017-09-27
fix set volume

jonaslandsgesell:
Fixes #1453.


fweik (approved):

fweik:
Please apply the patch again from a current master, the sample scripts have moved and git isn't taking it too well. Otherwise LGTM.

KaiSzuttor:
please also change testcase accordingly

fweik (approved):

PR #1457, opened 2017-09-29
Fixed build w/o PARTICLE_ANISOTROPY

### LINKED ISSUE #1430: Feature ROTATIONAL_INERTIA does not build without PARTICLE_ANISOTROPY enabled

### LINKED ISSUE #1440: Clash of ROTATIONAL_INERTIA and PARTICLE_ANISOTROPY
Steps to reproduce:

1. Copy `maintainer/configs/maxset.hpp` as `myconfig.hpp`.
2. Comment out the line `#define PARTICLE_ANISOTROPY` in `myconfig.hpp`.
3. Try to compile.
4. ???
5. Profit.

fweik:
Fixes #1440. Fixes #1430.

Description of changes:
 - Replaced ROTATIONAL_INTERTIA by PARTICLE_ANISOTROPY in some places
 - Let Thermostat.set_langevin accept scalars for gamma{,_rotation} if PARTICLE_ANISOTROPY
   is on, so that the behaviour does not change by just turning on the feature
- Adapted some tests
- Two small bugfixes in the core code

### DIFF
--- a/src/python/espressomd/particle_data.pyx
+++ b/src/python/espressomd/particle_data.pyx
@@ -1501,16 +1501,17 @@ cdef class ParticleHandle(object):
         for k in P.keys():
             setattr(self, k, P[k])
 
-    def convert_vector_body_to_space(self,vec):
-        """Converts the given vector from the particle's body frame to the space frame"""
-        cdef Vector3d res
-        cdef Vector3d _v
-        _v[0]=vec[0]
-        _v[1]=vec[1]
-        _v[2]=vec[2]
-        self.update_particle_data()
-        res= convert_vector_body_to_space(self.particle_data.get()[0],_v)
-        return np.array((res[0],res[1],res[2]))
+    IF ROTATION:
+        def convert_vector_body_to_space(self,vec):
+            """Converts the given vector from the particle's body frame to the space frame"""
+            cdef Vector3d res
+            cdef Vector3d _v
+            _v[0]=vec[0]
+            _v[1]=vec[1]
+            _v[2]=vec[2]
+            self.update_particle_data()
+            res= convert_vector_body_to_space(self.particle_data.get()[0],_v)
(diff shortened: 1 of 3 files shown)


hmenke (commented):
I don't think that's correct either.  `ROTATIONAL_INERTIA` introduces different `gamma_rot` per rotational axis and `PARTICLE_ANISOTROPY` introduces different `gamma` per translational axis on top of that.  Right now it seems to be all mingled together with no clear distinction what each feature does.  I wouldn't be too surprised if it now somehow fails at runtime or at least has unexpected behaviour.

fweik:
No, ROTATIONAL_INERTIA activates the inertia tensor. PARTICLE_ANISOTROPY
makes the gammas anisotropic. That is how it is supposed to be and what the
docs say, it just was not reflected in the code before.

hmenke:
I'm not so sure about that.  As you can see the test `mass-and-rinertia_per_particle` now fails.  I won't have time to dig deeper into it this week, but I will take a look again during the Summer School.

Also when looking at the old `master` branch, the feature `ROTATIONAL_INERTIA` needs the anisotropic `gamma_rot` (this was before `PARTICLE_ANISOTROPY` was introduced).
https://github.com/espressomd/espresso/blob/6cbef0e6b2e102f0fa75c06c6288aea87d620371/src/tcl/particle_data_tcl.cpp#L56-L73

Furthermore, talking about physics for a second, the rotational inertia for isotropic `gamma_rot` would be pretty trivial (it would always be the one of a sphere).

---

For my own reference, the PR which introduced `PARTICLE_ANISOTROPY` #937

fweik:
The test broke by the merge commit 0d2779b6a28a6d8f5659303eea27438e53f299df, because I changed the behaviour of kT. I need to look into this, because the new bevaviour makes more sens to me, but seems to be not correct. But this is unrelated.

PR #1471, opened 2017-10-04
Rdf

fweik:
Fixes '<rdf>' calculation.

Description of changes:
 - Fixed average rdf calculation
 - accept single type list for rdfs
 - simple test for rdf


KonradBreitsprecher (approved):
Can't really see which of the changes actually fixes the rdf averaging, but I could successfully reproduce the NaCl crystal rdf from the tutorial.

fweik:
https://github.com/espressomd/espresso/pull/1471/commits/a3b4bd549d2ca03ea67fc6a0e37fe7f3aae50edd
the added i and j indices that I wrongly dropped when I switch to the new partCfg.

PR #1488, opened 2017-10-05
dissto(id) now ignores own ID from particle list

davidsean:
Fixes 
distto(id) always returns a distance of 0,
this was due to checking ID with all others (including ID), thus always zero minimum distance.

Description of changes:
 - found a bug, fixed it.
 - doc string describes a bit better what it does

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?

### DIFF
--- a/src/python/espressomd/analyze.pyx
+++ b/src/python/espressomd/analyze.pyx
@@ -108,7 +108,7 @@ class Analysis(object):
 
     def distto(self, id=None, pos=None):
         """
-        Calculates the minimum distance to a point or particle.
+        Calculates the distance to a point or particle.
 
         Parameters
         ----------
--- a/src/python/espressomd/analyze.pyx
+++ b/src/python/espressomd/analyze.pyx
@@ -139,13 +139,16 @@ class Analysis(object):
         if id != None:
             if not isinstance(id, int):
                 raise ValueError("Id has to be an integer")
+            if id >= len(self._system.part):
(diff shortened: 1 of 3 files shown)


src/python/espressomd/analyze.pyx
```diff
@@ -139,13 +139,16 @@ class Analysis(object):
         if id != None:
             if not isinstance(id, int):
                 raise ValueError("Id has to be an integer")
+            if id >= len(self._system.part):
```
KaiSzuttor: That's not a good test... you can set a single particle with whatever id you want it to have

KaiSzuttor: Better use sth like: if not id in system.part[:].id

davidsean: good suggestion, I somehow didn't make the realization that the ID property and index were different.

davidsean:
I am not sure why the ICP CUDA build failed

KaiSzuttor:
Did you try to restart it by hand? Thats actually not possible

davidsean:
no ( I wouldn't know how to), it's quite weird

davidsean:
should I try a dummy commit?

KaiSzuttor:
You can try but it's a general problem we have to deal with

fweik (approved):

PR #1490, opened 2017-10-06
Switching from deprecated property syntax to decorator based.

KaiSzuttor:


fweik:
LGTM, but this should than also be done in particle_data.pyx.

KaiSzuttor:
cython version partially too old (0.24 needed)

fweik:
I'm not sure that we can bump the cython version at this point.

RudolfWeeber:
Since the old synatx still works with py2 and py3, I don't think, this is a sufficient reason to require a newer Cython version.
Should we de-milestone this?

fweik:
I agree.

PR #1510, opened 2017-10-10
Extend thermostat and rotation test

RudolfWeeber:
Verify diffusion coeff via Green Kubo per component for rotation and translation.
All combinations of ROTATION, MASS, RINERTIA, LANGEVIN_PER_PARTICLE and PARTICLE_ANISOTROPY should be handled.
Extended test for rotational inertia: Test momenum and energy conservation for non-isotropic inertia tensor

Further changes:
* Coverage maxset build with -O1 rather than -O0. Otherwise the tests never complete.
Correlator: Prevent double finalize() as well as get_data() after finalize().
py thermostat interface: Replace isinstance(... List) by hasattr(.., @__iter___@) so it works with tuplesp and np arrays.

### DIFF
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -4,7 +4,7 @@ file(GLOB EspressoCore_SRC
 
 if( WITH_COVERAGE )
   set(CMAKE_BUILD_TYPE "Debug" CACHE STRING "Choose the type of build, options are: Debug Release RelWithDebInfo MinSizeRel." FORCE)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -O0 --coverage -fprofile-arcs -ftest-coverage")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -O1 --coverage -fprofile-arcs -ftest-coverage")
   link_libraries(gcov)
 endif()
 
--- a/src/core/correlators/Correlator.cpp
+++ b/src/core/correlators/Correlator.cpp
@@ -380,7 +380,8 @@ void Correlator::initialize() {
 
 int Correlator::get_data() {
   if (finalized) {
-     throw std::runtime_error("No data can be added after finalize() was called.");
+     runtimeErrorMsg() << "No data can be added after finalize() was called.";
+     return 0;
   }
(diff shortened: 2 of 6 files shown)


src/core/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ file(GLOB EspressoCore_SRC
 
 if( WITH_COVERAGE )
   set(CMAKE_BUILD_TYPE "Debug" CACHE STRING "Choose the type of build, options are: Debug Release RelWithDebInfo MinSizeRel." FORCE)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -O0 --coverage -fprofile-arcs -ftest-coverage")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -O1 --coverage -fprofile-arcs -ftest-coverage")
```
hmenke, line 7: That is not a good idea.  `-O1` includes for example `-finline-functions-called-once` which will mess up coverage.

hmenke (changes requested):

src/core/CMakeLists.txt
fweik: From the numbers that does not seem to be a big issue.

fweik (approved):

RudolfWeeber:
> > @@ -4,7 +4,7 @@ file(GLOB EspressoCore_SRC
>  
>  if( WITH_COVERAGE )
>    set(CMAKE_BUILD_TYPE "Debug" CACHE STRING "Choose the type of build, options are: Debug Release RelWithDebInfo MinSizeRel." FORCE)
> -  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -O0 --coverage -fprofile-arcs -ftest-coverage")
> +  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -O1 --coverage -fprofile-arcs -ftest-coverage")
> 
> That is not a good idea.  `-O1` includes for example `-finline-functions-called-once` which will mess up coverage.
It was simply necessary to get the Green Kubo integrals to converge
in less than 10 minutes on Travis. I think, we can trade good
testing for all aspects of the thermostat (and apparently the correlator) against a bit of accuracy
in coverage.

PR #1515, opened 2017-10-11
Lb interpol

KaiSzuttor:
Added kernel function + interface for getting fluid velocities at all particle positions


hmenke (approved):

fweik (changes requested):
Test missing

KaiSzuttor:
The three-point coupling is not yet working and I can not see why.
Two-point works just fine.

src/core/lbgpu_cuda.cu
```diff
@@ -4275,22 +4275,32 @@ void lb_lbfluid_get_population( int xyz[3], float population_host[LBQ], int c )
...
+        interpolation_three_point_coupling(n_a, position, node_index, nullptr, delta, &u_gpu[3*index]);
+    }
+#endif
   }
 }
 
 /** interface to get fluid velocities at particle positions
 */
-std::vector<float> lb_lbfluid_get_fluid_velocity_at_particle_positions() {
+std::vector<float> lb_lbfluid_get_fluid_velocity_at_particle_positions(std::string coupling="twopoint") {
```
hmenke: Please be consistent with the rest of LB and use the strings `"2pt"` and `"3pt"`.

src/core/lbgpu_cuda.cu
```diff
@@ -4275,22 +4275,32 @@ void lb_lbfluid_get_population( int xyz[3], float population_host[LBQ], int c )
...
+    unsigned int node_index[8];
+    float mode[19*LB_COMPONENTS];
+    float delta[8];
     position[0] = particle_data[index].p[0];
     position[1] = particle_data[index].p[1];
     position[2] = particle_data[index].p[2];
     // Do the velocity interpolation
-    get_interpolated_velocity(n_a, position, &u_gpu[3*index], node_f);
+#ifndef SHANCHEN //because SHANCHEN uses dv in the two point coupling and we only give NULL pointer.
+    if (coupling == 0) {
```
hmenke: Please use an `enum`
```cpp
enum class particle_coupling {
    two_point,
    three_point,
};
```

src/core/lbgpu_cuda.cu
```diff
@@ -4301,13 +4311,22 @@ std::vector<float> lb_lbfluid_get_fluid_velocity_at_particle_positions() {
   int blocks_per_grid_particles_x = (lbpar_gpu.number_of_particles + threads_per_block_particles * blocks_per_grid_particles_y - 1) /
                                     (threads_per_block_particles * blocks_per_grid_particles_y);
   dim3 dim_grid_particles = make_uint3(blocks_per_grid_particles_x, blocks_per_grid_particles_y, 1);
-  KERNELCALL( lb_lbfluid_get_fluid_velocity_at_particle_positions_kernel, dim_grid_particles, threads_per_block_particles,
-              ( *current_nodes, gpu_get_particle_pointer(), u_gpu, node_f)
-            );
-  std::vector<float> u_host(3*lbpar_gpu.number_of_particles);
+  if (coupling.compare("twopoint") == 0) {
```
hmenke: Why???
```cpp
if (coupling == "2pt")
```

src/core/lbgpu_cuda.cu
```diff
@@ -4301,13 +4311,22 @@ std::vector<float> lb_lbfluid_get_fluid_velocity_at_particle_positions() {
...
+                );
+  }
 
+  std::vector<float> u_host(3*lbpar_gpu.number_of_particles);
   cuda_safe_mem(cudaMemcpy(u_host.data(), u_gpu, 3*lbpar_gpu.number_of_particles*sizeof(float), cudaMemcpyDeviceToHost));
   cuda_safe_mem(cudaFree(u_gpu));
+  // Multiply all values of u_host by agrid/tau to get MD units.
+  std::transform(u_host.begin(), u_host.end(), u_host.begin(),
+                         std::bind1st(std::multiplies<float>(),lbpar_gpu.agrid/lbpar_gpu.tau));
   return u_host;
```
hmenke: Please add an assertion for SHANCHEN somewhere.
```cpp
#ifdef SHANCHEN
assert(false && "get_fluid_velocity_at_particle_positions is not compatible with SHANCHEN");
#endif
```

hmenke (changes requested):

src/core/lbgpu_cuda.cu
```diff
@@ -4265,4 +4268,65 @@ void lb_lbfluid_get_population( int xyz[3], float population_host[LBQ], int c )
...
+        interpolation_two_point_coupling(n_a, position, node_index, mode, nullptr, delta, &u_gpu[3*index]);
+    } else if (coupling == 1) {
+        interpolation_three_point_coupling(n_a, position, node_index, nullptr, delta, &u_gpu[3*index]);
+    }
+  }
+}
+
+/** interface to get fluid velocities at particle positions
+*/
+std::vector<float> lb_lbfluid_get_fluid_velocity_at_particle_positions(std::string coupling="twopoint") {
```
fweik: Please move the string parsing to a sep function or to python and raise an error if it is not one of the allowed values. The current impl will just return garbage if you do lb_lbfluid_get_fluid_velocity_at_particle_positions("not_a_coupling").

src/core/lbgpu_cuda.cu
```diff
@@ -4265,4 +4268,65 @@ void lb_lbfluid_get_population( int xyz[3], float population_host[LBQ], int c )
...
+                  ( *current_nodes, gpu_get_particle_pointer(), u_gpu, node_f, device_rho_v, 1)
+                );
+  }
+
+  std::vector<float> u_host(3*lbpar_gpu.number_of_particles);
+  cuda_safe_mem(cudaMemcpy(u_host.data(), u_gpu, 3*lbpar_gpu.number_of_particles*sizeof(float), cudaMemcpyDeviceToHost));
+  cuda_safe_mem(cudaFree(u_gpu));
+  // Multiply all values of u_host by agrid/tau to get MD units.
+  std::transform(u_host.begin(), u_host.end(), u_host.begin(),
+                         std::bind1st(std::multiplies<float>(),lbpar_gpu.agrid/lbpar_gpu.tau));
```
fweik: std::bind1st is deprecated in C++11 (and the use of std::bind is generally disencouraged.) This can always be replaced by a lambda like so:
```c++
std::transform(_, _, _, [&lbpar_gpu](float &f) { return f * lbpar_gpu.agrid/lbpar_gpu.tau });
```
Please do so.

fweik (changes requested):

fweik:
~~It's very well possible that the 3-pt coupling is just broken. I don't think that anybody has used that recently and there is no test. There was some testing in engin_lbgpu.py but that did not work when I tried.~~ I just checked the lb thermostat test with 3pt and that seems to work.

src/core/lbgpu_cuda.cu
KaiSzuttor: error: a variable with static storage duration cannot be captured in a lambda.
Not capturing lbpar_gpu compiles but the test then fails.

fweik: Ah sorry, you don't need to capture it because it's global. But then it should just work...

fweik: Sorry, should be
```c++
  std::transform(u_host.begin(), u_host.end(), u_host.begin(),
                 [](float u) { return u * lbpar_gpu.agrid / lbpar_gpu.tau; });
```
, the value has to be returned.

KaiSzuttor:
@hmenke can you see why three point cpuling does not work here?

testsuite/lb_get_u_at_part.py
```diff
@@ -0,0 +1,78 @@
...
+        self.lb_fluid = lb.LBFluid_GPU(
+            visc=self.params['viscosity'],
+            dens=self.params['dens'],
+            agrid=self.params['agrid'],
+            tau=self.params['tau'],
+            fric=self.params['friction']
+        )
+        self.system.actors.add(self.lb_fluid)
+        self.vels = np.zeros((self.n_nodes_per_dim, 3))
+        self.vels[:, 0] = np.arange(self.n_nodes_per_dim, dtype=float)
```
hmenke: If I use
```python
self.vels[:, 0] = np.ones(self.n_nodes_per_dim, dtype=float)
```
then the result with 3pt coupling is off by a constant factor of exactly `2.25`.  I have still no idea why.

KaiSzuttor:
Threepoint coupling seems not to work correct but is unrelated to this PR, maybe we should have seen sth like that in previous PRs if we had tested it

KaiSzuttor:
Ready for review again

fweik (approved):

KaiSzuttor:
Test fails for parallel execution

fweik:
The particles on the GPU are in an arbitrary order, this just happens to work on a single node. This actually can not be fixed easily, because the information is not available on the head node.

KaiSzuttor:
@fweik please have a look, implemented the thrust version, but I'm sure code can be improved.

src/core/lbgpu_cuda.cu
```diff
@@ -4252,5 +4257,42 @@ void lb_lbfluid_get_population( int xyz[3], float population_host[LBQ], int c )
...
+        float3 _u = make_float3(u[0], u[1], u[2]);
+		return _u;
+	} 
+};
+
+void lb_lbfluid_get_interpolated_velocity_at_positions(double *positions, double *velocities, int length) {
+    thrust::host_vector<float3> positions_host(length);
+    for (int p=0; p < 3 * length;) {
+        // Cast double coming from python to float.
+        positions_host[p/3].x = (float)positions[p];
```
fweik: Please use static_cast instead of c-style casts

src/core/lbgpu_cuda.cu
```diff
@@ -4252,5 +4257,42 @@ void lb_lbfluid_get_population( int xyz[3], float population_host[LBQ], int c )
...
+        positions_host[p/3].y = (float)positions[p+1];
+        positions_host[p/3].z = (float)positions[p+2];
+        p += 3;
```
fweik: move to for statement

src/core/lbgpu_cuda.cu
```diff
@@ -4252,5 +4257,42 @@ void lb_lbfluid_get_population( int xyz[3], float population_host[LBQ], int c )
...
+    }
+    thrust::device_vector<float3> positions_device = positions_host;
+    thrust::device_vector<float3> velocities_device(length);
+    thrust::transform(positions_device.begin(), positions_device.end(), velocities_device.begin(), two_point_interpolation(*current_nodes, device_rho_v));
+    thrust::host_vector<float3> velocities_host = velocities_device;
+    int index = 0;
+    for (auto v : velocities_host) {
+        velocities[index] = (double)v.x * lbpar_gpu.agrid/lbpar_gpu.tau;
```
fweik: static_cast

src/core/lbgpu_cuda.cu
```diff
@@ -4252,5 +4257,42 @@ void lb_lbfluid_get_population( int xyz[3], float population_host[LBQ], int c )
   cuda_safe_mem(cudaFree(population_device));
 }
 
+struct two_point_interpolation { 
+    LB_nodes_gpu current_nodes_gpu;
+    LB_rho_v_gpu *d_v_gpu;
+    two_point_interpolation(LB_nodes_gpu _current_nodes_gpu, LB_rho_v_gpu *_d_v_gpu) : current_nodes_gpu(_current_nodes_gpu), d_v_gpu(_d_v_gpu) {};
+	__device__ float3 operator()(const float3 &position) {
```
fweik: should be const

src/core/lbgpu_cuda.cu
```diff
@@ -4252,5 +4257,42 @@ void lb_lbfluid_get_population( int xyz[3], float population_host[LBQ], int c )
...
+        unsigned int node_index[8];
+        float delta[8];
+        float u[3];
+        float mode[19*LB_COMPONENTS];
+        float _position[3] = {position.x, position.y, position.z};
+        interpolation_two_point_coupling(current_nodes_gpu, _position, node_index, mode, d_v_gpu, delta, u);
+        float3 _u = make_float3(u[0], u[1], u[2]);
```
fweik: Directly return the value

fweik (changes requested):

fweik (approved):

fweik:
Looks good to me. Did you discuss the unit change in lb_lbnode_set_u with @rempferg ?

KaiSzuttor:
yes that was simply wrong before

PR #1520, opened 2017-10-12
lbgpu: Made remove momentum respect particle masses.

fweik:


testsuite/lbgpu_remove_total_momentum.py
```diff
@@ -37,8 +41,7 @@ def test(self):
 
         p = np.array(s.analysis.analyze_linear_momentum())
 
-        self.assertTrue(np.all(np.abs(p) < 1e-3))
-
+        self.assertAlmostEqual(np.max(p), 0., places=3)
```
hmenke, line 45: This doesn't catch negative outliers.  Just add another test
```python
self.assertAlmostEqual(np.min(p), 0., places=3)
```

hmenke (changes requested):
Rest looks good.

hmenke (approved):

PR #1531, opened 2017-10-17
maintainer: Parse git repository info and expose to core and python.

### LINKED ISSUE #1529: Make git commit hash and state accessible from Py interface
This existed already for tcl.

fweik:
Fixes #1529.

Description of changes:
 - Expose version and repository info to python

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,6 +50,13 @@ if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE "Release" CACHE STRING "Choose the type of build, options are: Debug Release RelWithDebInfo MinSizeRel." FORCE)
 endif(NOT CMAKE_BUILD_TYPE)
 
+# Parse repository info from git if available
+include(version)
+
+configure_file(
+  ${CMAKE_SOURCE_DIR}/src/version.hpp.in
+  ${CMAKE_BINARY_DIR}/src/version.hpp
+  )
 
 # On Mac OS X, first look for other packages, then frameworks
 set(CMAKE_FIND_FRAMEWORK LAST)
--- a/src/python/espressomd/version.pxd
+++ b/src/python/espressomd/version.pxd
@@ -1,0 +1,5 @@
+cdef extern from "version.hpp":
+    cdef int ESPRESSO_VERSION
+    cdef const char * GIT_BRANCH
+    cdef const char * GIT_COMMIT_HASH
+    cdef const char * GIT_STATE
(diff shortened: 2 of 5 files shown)


src/python/espressomd/version.pxd
```diff
@@ -0,0 +1,5 @@
+cdef extern from "version.hpp":
+    cdef int ESPRESSO_VERSION
+    cdef const char * GIT_BRANCH
+    cdef const char * GIT_COMMIT_HASH
```
mkuron, line 4: In version.hpp, these are macros and not strings. Does importing them really work like that?

CMakeLists.txt
```diff
@@ -50,6 +50,13 @@ if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE "Release" CACHE STRING "Choose the type of build, options are: Debug Release RelWithDebInfo MinSizeRel." FORCE)
 endif(NOT CMAKE_BUILD_TYPE)
 
+# Parse repository info from git if available
+include(version)
+
+configure_file(
+  ${CMAKE_SOURCE_DIR}/src/version.hpp.in
+  ${CMAKE_BINARY_DIR}/src/version.hpp
+  )
```
mkuron, line 59: For release tarballs, this will need special treatment, e.g. shipping version.hpp inside the tarball and skipping this configure_file.

src/python/espressomd/version.pxd
fweik: Apparently, but now that I think of it they don't need to be macros in the first place.

CMakeLists.txt
fweik: I think an install statement for the generated header should do it.

src/python/espressomd/version.pyx
```diff
@@ -0,0 +1,35 @@
+def major():
+    """Prints the major version of Espresso.
+    """
+    return ESPRESSO_VERSION // 100000
```
hmenke, line 4:
```python
from __future__ import division
```

hmenke (changes requested):

src/python/espressomd/version.pyx
fweik: Why? // works in python 2 and 3 and does what I want.

CMakeLists.txt
fweik: Also at the moment it will just work, the version.hpp is not needed at runtime.

src/python/espressomd/version.pyx
hmenke: Sorry, I didn't know that.

hmenke (approved):

mkuron (approved):

PR #1539, opened 2017-10-19
 added test for catalytic reactions, thanks to henri

### LINKED ISSUE #1329: Integration test for catalytic reactions

### LINKED ISSUE #1536: REACTION_ENSEMBLE sort out ifdefs
Currently, 
REACTION_ENSEMBLE 
is defined as a feature in features.def, but the code in the core is compiled in unconditionally.
Only the Python interface is not added, if the features is not defined.

Probably, the feature should be removed from features.def and the Python stuff also compiled unconditionally. Furthermore the skipIf decorator needs to be removed from the test.

jonaslandsgesell:
Fixes #1329
Fixed #1536


fweik:
Why do you need ROTATION if you then do not turn it on?

jonaslandsgesell:
Errors occur with e.g. the rest1 config. The needed module is not imported since it is not built. 

  File "catalytic_reaction.py", line 2, in <module>
    from espressomd.reaction import Reaction
ImportError: cannot import name Reaction

How do I fix this in an elegant way, instead of introducing try, except during the import?

fweik:
Python supports conditional imports:
```python
if espressomd.has_features("CATALYTIC_REACTION"):
    from espressomd.reaction import Reaction
```

jonaslandsgesell:
Yeah I would have done try: except but that of course also works.

jonaslandsgesell:
Removed conditional inclusion for reaction ensemble #1536

fweik (approved):

jonaslandsgesell:
Langevin Thermostat tests fail on mac os although untouched.

PR #1548, opened 2017-10-19
Com fixed

### LINKED ISSUE #1464: Implement Python supoort for com_fixed
Probably as an actor, since it is a collective interaction.
The feature seems to be of general use and implementing support is probably not complicated.

fweik:
Fixes #1464

Description of changes:
 - Reimplementation of comfixed
   (this fixes the center of mass of a type by subtracting
    the total force on the type from the parts after each
    force calcculation).
 - Python interface via script interface.


src/python/espressomd/comfixed.py
```diff
@@ -0,0 +1,22 @@
...
+from .script_interface import ScriptInterfaceHelper, script_interface_register
+
+@script_interface_register
+class ComFixed(ScriptInterfaceHelper):
+    """Fix the center of mass of specific types.
+
+    Subtracts mass-weighted fraction of the total
+    force action on all particles of the type from
+    the particles after each force calculation. This
+    keeps the center of mass of the type fixed iff
```
jonaslandsgesell, line 11: The name of the function is comfix so why not remove all the momentum of the particles of the given type, even if the momentum for this type is nonzero initially? It should remove any drift in order to fix the center of mass.

fweik:
Hm I see your point. On the other hand there are other functions to remove the com velocity (or a numpy one-liner), and this is more flexible, as it also allows constant com motion systems.

jonaslandsgesell (approved):

PR #1552, opened 2017-10-20
cmake: Made EspressoConfig into a library containing config-features.…

### LINKED ISSUE #1538: Feature requirements get checked late if at all
e.g., putting ELECTROKINETICS into myconfig.hpp and compiling without cuda.

ELECTROKINETCS requires CUDA as stated in features.def.

The checks are put in the build dir in src/core/config-features.cpp.
So
* either, config-features.cpp needs to be compiled before everything else
* or, the requirement checking should go into config-features.hpp. In that case, it would also be possible to output a warning and undef the feature rather than to prevent compilation (if thats wat we want).

src/core/gen_feature_config.py

fweik:
…cpp to force early feature consistency check.

Fixes #1538

### DIFF
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -25,15 +25,14 @@ add_custom_command(OUTPUT config-version.cpp
 )
 
 # Custom target for everything that needs to be triggered when the myconfig changes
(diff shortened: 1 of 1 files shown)


RudolfWeeber:
Can be merged.

PR #1553, opened 2017-10-20
Build CUDA code with Clang

mkuron:
Clang has supported [compiling CUDA](https://llvm.org/docs/CompileCudaWithLLVM.html) for a while now. This allows us to bypass `nvcc`'s terrible C++ frontend and use Clang's instead, making .cu files behave pretty much exactly like .cpp files. I've been wanting to try this for a some time and finally got around to doing it.

By default, of course we still use `nvcc`, but you can switch to Clang by specifying it as the host and CUDA compiler:
```
cmake .. -DCMAKE_C_COMPILER=$(which clang) -DCMAKE_CXX_COMPILER=$(which clang++) -DCUDA_NVCC_EXECUTABLE=$(which clang++)
```

The minimum required Clang version is 3.8, but newer is probably better. Of course, you also need to have CUDA installed (for the driver and the PTX assembler) and the version needs to be supported by your Clang version. Note that [Thrust](https://github.com/thrust/thrust) 1.8, which is currently shipped with CUDA, doesn't fully support Clang, so you need to manually install the 1.9 beta first.


fweik (approved):

hmenke (approved):

mkuron:
Just added a CI build. This depends on https://github.com/espressomd/docker/pull/7.

mkuron:
I have added a CI build. This depends on https://github.com/espressomd/docker/pull/7.

There is a successful build at https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/pipelines/918. After that, I switched the build to the espressomd the Docker Hub, where the image hasn't been created yet.

fweik (approved):

PR #1557, opened 2017-10-21
maintainer: Prevent user from declaring external FEATURES in the myco…

### LINKED ISSUE #1554: Prevent users from putting external dependencies in myconfig.hpp
#define SCAFACOS
in myconfig.hpp whith scafacos not found/installed results in a compilation error late in the build process.
?Defining CUDA when compiling without cuda support will compile successfully (everything except cuda stuff) and fail on module import.

fweik:
…nfig.

Fixes #1554.

Description of changes:
 - Add guards for external features in config-features.hpp
 - Include cmake_config.hpp from there to control the order.


RudolfWeeber:
Can be merged

PR #1561, opened 2017-10-23
python: Added missing initialization to local_stress_tensor interface.

### LINKED ISSUE #1545: local_stress_tensor segfault
calling analysis.local_stress_tensor() results in a segfault
```python
import espressomd
system = espressomd.System()
system.box_l = [100, 100, 100]

    epsilon=1.0, sigma=1.0,
    cutoff=1.0**(1.0/6.0), shift="auto")

for i in range(9):
    system.part.add(id=i, pos=[1.0, 1.0, i], type=0)

res=system.analysis.local_stress_tensor(periodicity=(1, 1, 1),
                                        range_start=(1.0, 1.0, 1.0),
                                        stress_range=(50.0, 50.0, 50.0),
                                        bins=(5, 5, 5))
```

fweik:
Fixes #1545.

Description of changes:
 - 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?

### DIFF
--- a/src/python/espressomd/analyze.pyx
+++ b/src/python/espressomd/analyze.pyx
@@ -558,7 +558,7 @@ class Analysis(object):
         """local_stress_tensor(periodicity=(1, 1, 1), range_start=(0.0, 0.0, 0.0), stress_range=(1.0, 1.0, 1.0), bins=(1, 1, 1))
         """
 
-        cdef double_list * local_stress_tensor = NULL
+        cdef double_list local_stress_tensor
         cdef int[3] c_periodicity, c_bins
         cdef double[3] c_range_start, c_stress_range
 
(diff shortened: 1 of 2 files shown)


jonaslandsgesell:
we could add some more Inline documentation here and remove that Info from analyze.rst

fweik:
Yes you could ;-)

KaiSzuttor (approved):

hmenke:
This fixes the segfault but the behaviour of the function is still broken.  Looking at the core code you will find that the argument is simply unused and therefore the function will always be a no-op.  Also it is very unclear from reading the function definition what it is even supposed to do.  In my opinion it should be removed or at least rewritten.

https://github.com/espressomd/espresso/blob/c0adc640087ecd6441483f45aa45f3a887451a77/src/core/pressure.cpp#L1099-L1125

https://github.com/espressomd/espresso/blob/c0adc640087ecd6441483f45aa45f3a887451a77/src/core/communication.cpp#L1396-L1479

src/python/espressomd/analyze.pyx
```diff
@@ -558,7 +558,7 @@ class Analysis(object):
         """local_stress_tensor(periodicity=(1, 1, 1), range_start=(0.0, 0.0, 0.0), stress_range=(1.0, 1.0, 1.0), bins=(1, 1, 1))
         """
 
-        cdef double_list * local_stress_tensor = NULL
```
hmenke, line 561: This was “correct”, `local_stress_tensor` is a list of DoubleLists.

PR #1564, opened 2017-10-24
core: Implemented subt_lj via regular lj + test

fweik:
Fixes subt_lj with lj offset.

Description of changes:
 - Calc subt_lj via regular lj
 - Test


RudolfWeeber:
Please check that forces/energy are there before the bond is added.
Or did I miss that?

PR #1567, opened 2017-10-25
Maintainer

KaiSzuttor:
Fixed all current sphinx warning, moved status pending post to the webhook script, turned sphinx warnings into errors and added a job for sphinx check.


fweik (approved):

mkuron:
It seems like the webhook script didn't post the pending status. At the bottom of this page, the "Details" button next to "ICP CUDA build" is still missing, even though the build is running (https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/pipelines/961).

KaiSzuttor:
I removed the link because at the time the hook is received the pipeline id is not yet known. After the job finishes, the link will be available

KaiSzuttor:
Maybe it's better to have the link as soon as the job started, so I'll reremove the status pending within the job again.

mkuron:
The link should be there as soon as the job starts. You could set it to pending without link when the hook is called and set it to pending again with link when the job starts. The pipeline is tied to a commit ID though (and not a build) and should thus be created (almost) immediately after the hook is called, right?

> I'll reremove the status pending within the job again.

I don't get what you're trying to say here... Does "reremove" mean adding it again?

KaiSzuttor:
dont mind, wip

mkuron (approved):

KaiSzuttor:
Get warnings in the docker, not locally, don't know why.

fweik:
@KaiSzuttor I fixed the key issue, but the user seem to lack permissions. I tried the private key stored in gitlab by hand, and it would not let me log in w/o a password. Can you please check that the key is authorized?

mkuron:
@davidsean, this is what introduced
```
Warning, treated as error:
WARNING: the sphinxcontrib.bibtex extension is not safe for parallel
reading, doing serial read
```

Issue #1559, opened 2017-10-23
2d electrostatic methods don't agree with each other 


RudolfWeeber:
@KonradBreitsprecher @arnolda 

Discovered while writing tests (#1324).
I tested: p3m+elc, mmm2d, scafacos p2nfft against each other and against reference data obtained by direct summation with a circle-cutoff with a radius of 400 periodic images.
* scafacos p2nfft agreed with the direct summation result (Deviation: forces: abs 1.6E-3, abs energy 0.08, rel=1E-3). Energy should converge much slower. I'm generating test data with 2000 images and will update, once finished.
* mmm2d forces match reference data also up to 1.6E-3. The forces seem to agree with p2nfft. Energy is off completely (deviation of abs 55, rel 25%)
* P3m+elc somewhat agrees with reference data and p2nfft, but is an order of magnitude worse even though I tuned p3m to 1E-7 and elc pair wise error to 1E-9. Force difference is 0.05 and energy difference 0.6.

The system is the coulomb_cloud_wall test from scafacos with the gap region from z=9 to z=10=box_l cleared for elc.

In conclusion, I think, something is wrong with elc forces+energy and mmm2d energy. It is not clear to me whether this is in the interface or the core.

Can someone with experience in these methods please take this over?

To reproduce:
The test + reference data is on my github in the branch electrostatic_mixed_periodicity.
Test is in testsuite/coulomb_mixed_periodicity.py. Not included in CMakeList and the assertions are turned off at this point.
I put the reference data and the script to produce it in the testsuite/data folder.

To use the scafacos methods on ICP Ubuntu machines:
export PKG_CONFIG_PATH=/usr/local/stow/scafacos/lib/pkgconfig/scafacos.pc:$PKG_CONFIG_PATH
and 
export LD_LIBRARY_PATH=/usr/local/stow/scafacos/lib/:$LD_LIBRARY_PATH
before cmake.

KonradBreitsprecher:
I found that in the calculation of the self-energy, the sum over local particles changed to a std::accumulate, which is broken:
 
```
 double seng = coulomb.prefactor * calc_mmm2d_copy_pair_energy(dv);
 auto parts = local_cells.particles();
 self_energy = std::accumulate(
     parts.begin(), parts.end(), 0,
     [seng](double sum, Particle const &p) { return sum + seng * SQR(p.p.q); });
```
There is a subtle `0` to initialize the sum, but also sets the type to int. 
[...]

PR #1568, opened 2017-10-26
Fix mmm2d energy calculation

KonradBreitsprecher:
Fixes mmm2d energy calculation, partly #1559 .

Description of changes:
 - changed initial value to 0.0 in std::accumulate in mmm2d.cpp and mdlc_correction.cpp.


fweik:
There is another one in src/core/actor/EwaldGPU.cpp, could you change that too please. Thank's for fixing my bugs.

fweik (approved):

PR #1570, opened 2017-10-26
Pythonic functionality for array properties

### LINKED ISSUE #1550: Don't allow attempted change of vector components
A pitfall for many users might be that `system.part[0].pos[2] = 123` will run but don't change the z component of the position. This only changes the returned numpy.array and not the underlying property's component. A way out would be:
```
ret = np.array(..)
ret.flags.writable = False
return ret
```
in all getters that return arrays, which then throws an error if the users attempts the component change.

KonradBreitsprecher:
Fixes #1550 

I changed the following properties:

PARTICLES
- pos
- v
- f
- pos_folded
- omega_lab
- quat
- rotation
- omega_body
- torque_lab
- rinertia
- director
- gamma_rot
- gamma
- dip
- ext_force
- fix
- ext_torque

SYSTEM
- box_l
- periodicity

CELLSYSTEM
- node_grid

LB NODES
- velocity
- density
- pi
- pi_neq
- population


fweik:
interactions_non-bonded fails in test_lj_generic_softcore, which uses s.part[1].pos += self.step.
Should't this throw an error? Instead it seems so fail silently, which breaks the test.

KonradBreitsprecher:
Yes it throws `ValueError: output array is read-only` if I try a `p.pos += [1,1,1]`.
I successfully ran the tests locally and adapted the interactions_non-bonded and interactions_bonded already in the original PR.

testsuite/interactions_non-bonded.py
```diff
@@ -148,7 +148,8 @@ def test_lj_generic_softcore(self):
             epsilon=lj_eps, sigma=lj_sig, cutoff=lj_cut, offset=lj_off, b1=lj_b1, b2=lj_b2, e1=lj_e1, e2=lj_e2, shift=lj_shift, delta=lj_delta, lam=lj_lam)
 
         for i in range(231):
-            self.system.part[1].pos += self.step
+            self.system.part[1].pos = self.system.part[1].pos + self.step
+            self.system.part[1].pos = self.system.part[1].pos + self.step
```
fweik: This line should not be duplicated.

mkuron:
How about a solution that actually makes `system.part[0].pos[2] = 123` work? Just subclass `numpy.ndarray` and overwrite the `__setitem__` method with one that sets the value in the core too. That would be much more pythonic behavior.

KonradBreitsprecher:
That sounds good. Could you provide an example in the context of property _get_ and _set_ ?

mkuron:
It's pretty simple actually. Let's start with this simplified mockup of `system.part`:

```python
import numpy

class particle(object):
	def __init__(self):
		self.pos_core = [1,2,3] # let's pretend this member variable is the variable in Espresso core
	
	@property
	def pos(self):
		return numpy.array(self.pos_core)
	
	@pos.setter
	def pos(self, value):
		self.pos_core[0] = value[0]
		self.pos_core[1] = value[1]
		self.pos_core[2] = value[2]

p = particle()
assert numpy.all(p.pos == [1,2,3])
p.pos = [4,5,6] # this uses the setter
assert numpy.all(p.pos == [4,5,6])
p.pos[0] = 0 # this doesn't change the value because it uses the getter
assert numpy.all(p.pos == [4,5,6])
```

Now we subclass the ndarray, add a reference to the core variable and overwrite the `__setitem__` method:

```python
class my_array(numpy.ndarray):
	def __new__(cls, input_array, source):
		obj = numpy.asarray(input_array).view(cls)
		obj.source = source
        	return obj
	def __array_finalize__(self, obj):
		if obj is None: return
		self.source = getattr(obj, 'source', None)
        
	def __setitem__(self, i, val):
		newval = numpy.asarray(self.source.pos)
		newval[i] = val
		self.source.pos = newval

class particle2(object):
	def __init__(self):
		self.pos_core = [1,2,3]
	
	@property
	def pos(self):
		p = my_array(self.pos_core, self)
		return p
	
	@pos.setter
	def pos(self, value):
		self.pos_core[0] = value[0]
		self.pos_core[1] = value[1]
		self.pos_core[2] = value[2]
		

p2 = particle2()
assert numpy.all(p2.pos == [1,2,3])
p2.pos = [4,5,6] # this still uses the setter
assert numpy.all(p2.pos == [4,5,6])
p2.pos[0] = 0 # this now uses our custom __setitem__ method
assert numpy.all(p2.pos == [0,5,6])
```

KonradBreitsprecher:
It's working!
I extended my_array to access the property by string, so it can be used for all np.array returns.
It works (for pos) but it seems a bit unorthodox:

```python
class my_array(np.ndarray):

    def __new__(cls, input_array, source, attr):
        obj = np.asarray(input_array).view(cls)
        obj.source = source
        obj.attr = attr
        return obj

    def __array_finalize__(self, obj):
        if obj is None: return
        self.source = getattr(obj, 'source', None)

    def __setitem__(self, i, val):
        attr = self.source.__class__.__dict__[self.attr]
        newval = np.asarray(attr.__get__(self.source))
        newval[i] = val
        attr.__set__(self.source, newval)

```
All one has to do to in property `__get__` is to return:

```python
return my_array(ppos, self, 'pos')
```

@mkuron is this valid? Do I have to extend `__array_finalize__` in some way?

KonradBreitsprecher:
closing this until implemetation is finished..

fweik:
Please change the name to something more telling like array_ref or so.

mkuron:
@KonradBreitsprecher, looks good. You need one line in `__array_finalize__` for each instance variable that you have added, so you need `self.attr = getattr(obj, 'attr', None)` too.

RudolfWeeber:
Do I understand it correctly that 
s.part[0].pos[2] =5
will call the the setter fo the pos property of ParticleHanlde rather than just modifying the memory in the instance of ParitcleHandle.

Does this pr imply that
x=part[0].pos
and then changing x will modeify the live particle?

If so, I'd find that quite dangerous and likely to produce unindented consequences. 
Then, I'd probably prefer returning non-writable arrays or leaving the current state.

RudolfWeeber:
An other issue is particle slices, which reutn n-by-m arrays, where n is number of particles and m is number of entries for the property.

Merging this would result in different behaviour for slices and single particles.

mkuron:
> Does this pr imply that `x=part[0].pos` and then changing x will modeify the live particle?

That is correct, and that is expected pythonic behavior, similar to this:
```
import numpy as np
x = np.zeros(3)
y = x
y[1] = 1
assert x[1] == 1
```

> Merging this would result in different behaviour for slices and single particles.

Good point. @KonradBreitsprecher, can you add a `__setitem__` method for ParticleSlice as well?

RudolfWeeber:
Florian and I would like to discuss this issue at the upcoming Espresso meeting.

RudolfWeeber:
Offline discussion yielded that it is unlikely that we can implement this in a consistent fashion while still allowing fast numpy operations with, e.g., s.part[:].pos. 

We will therefore go with returning by value but as non-writable numpy arrays to prevent things like
s.part[:].pos[2]=0

fweik:
So this can be closed?

mkuron:
No, we are waiting for Konrad to revert this pull request to read-only attributes.

KonradBreitsprecher:
New PR #1610 due to lots of experimental commits here.

PR #1572, opened 2017-10-26
Lennard-Jones tutorial converted to iPython notebook

ashreyaj:
Fixes #

Description of changes:
 - 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


fweik:
Should the IPython notebooks replace the old tutorials? @KaiSzuttor @RudolfWeeber

KaiSzuttor:
IMHO there are a few reasons against that:
 - we wanted to automatically test for syntax errors in the tutorials, don't know if that is easily possible with IPYNB
 - to me it's not clear what are the big advantages compared to the current tutorials
 - we can not have both (not maintainable)
 - if we are not sure that IPYNB  will bring a lot of advantages we should not put too much effort into this since we have "bigger fish to fry" (as a smart guy once told me)
 - another dependency: ipython-notebook

RudolfWeeber:
So my memory is that we discussed this in the ESPResSo meeting and
the outcome was that we wanted to switch all tutorials to IPython
notebooks to avoid the duplication of code between Tex files and
scripts.

RudolfWeeber:
Re-milestoning for 4.0 due to use in summer school

jonaslandsgesell:
Can you please merge upstream python ? And check for runnability again. You might need to provide a seed if you did not do it yet.

jonaslandsgesell (approved):

PR #1573, opened 2017-10-26
Deceleration test

bogdan-tanygin:
Improvement of the per-particle anisotropic deceleration test.

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


bogdan-tanygin:
@fweik could you please clarify whether I should cleanup the non-`PARTICLE_ANISOTROPY` case from the test `mass-and-rinertia_per_particle.py` according to the commit https://github.com/espressomd/espresso/commit/dfdf77ea3a6f47b74572df9be928660429f1ac32#diff-c0d7dc25fbfb88170e8c08cd5a6bdcc0 ?

Also, I could try to fix this case. It depends on a reason of the above-mentioned commit. If it is just a technical fix without some implications then I could work on a return of non-`PARTICLE_ANISOTROPY` test cases. There could be some defects caused a build failure before.

bogdan-tanygin:
My last changes have been introduced with an assumption that non-`PARTICLE_ANISOTROPY` case is deprecated in this test. Again, I could either cleanup or fix this feature switch throughout the test depending on a reason of this change in the commit dfdf77e:
```
@ut.skipIf(not espressomd.has_features(["MASS",
 +                                        "PARTICLE_ANISOTROPY",
                                          "ROTATIONAL_INERTIA",
                                          "LANGEVIN_PER_PARTICLE"]),
             "Features not available, skipping test!")
```

fweik:
So the plan is that PARTICLE_ANISOTROPY enable the directional diffusion constants (linear and rotation iff ROTATION), and ROTATIONAL_INERTIA enable rinertia without having an effect on diffusion. The mentioned changes where to reflect that in the core.

bogdan-tanygin:
OK, I see, logically. Hence, within new approach - I should cleanup the `PARTICLE_ANISOTROPY` code in the test as far as it is anyway now is skipped. I'll report in the next comment once my changes will be final.

bogdan-tanygin:
Done..

fweik:
Ok, thank you. I'll have a look.

fweik (approved):

PR #1574, opened 2017-10-27
core: Disable LB

### LINKED ISSUE #1461: LB Fluid deactivation code missing
_deactivate_method() in lb.pyx for cpu and gpu lb needs to be implemented
@rempferg

fweik:
Fixes #1461.

Description of changes:
 - lb_set_lattice_swithc(0) turns off LB

Can please somebody who knows more about LB check that this is sufficient.
Any ideas how to test this are also welcome. @rempferg @KaiSzuttor


hmenke:
This is definitely sufficient for LB and LB_GPU, see below.  I'm not sure about EK though.
https://github.com/espressomd/espresso/blob/37dd602d01b73109804f2fdb14c70763768767ee/src/core/integrate.cpp#L493-L509

hmenke:
Okay, it doesn't suffice.  You have to deactivate EK as well because `ek_integrate()` calls to `lb_integrate_GPU()`.

KaiSzuttor:
This should not be a problem because you can not use LB and EK at the same time, see
https://github.com/espressomd/espresso/blob/29189a154185be99d2cfcb18fcebb041de29aa04/src/core/electrokinetics_cuda.cu#L2470-L2473

fweik:
I think we should merge this anyhow as it is an improvement.

RudolfWeeber:
Are you really sure, this is save? Will reactivating the lbf really provide a resting fluid when the dimensions of the lattice arne't changed? 
Did you try it out by running, e.g., lb.py and lbgpu.py tests from the testsuite from a single file?

fweik:
I'll try to come up with a test. I think it's no feasible to restore the fields after reactivation, this would mean storing the population field in the python objects while they are deactivated. This would be nice to have but nothing I'm willing to do atm.

RudolfWeeber:
>I think it's no feasible to restore the fields after reactivation, this would mean storing the population field in the python objects while they are deactivated. This would be nice to have but nothing I'm willing to do atm.
That's not what I meant:

s=System(box_l=(10,10,10)
lbf=LbFluid(agrid=1,...)
s.actors.add(lbf)
# I assume, there is a resting fluid now.
stuff that creates a flow field
s.actors.remove(lbf)

newlbf=LbFluid(agrid=1)
s.actors.add(newlbf)
Do I get a resting fluid or the flow field from before?

fweik:
Resting

KaiSzuttor:
thats what i would expect

fweik:
This is what you wanted to see?

hmenke (approved):

PR #1578, opened 2017-10-29
Draft for a contributors guide

hmenke:
I got the idea of a contributors guide when browsing the Rust repository because it greeted me with these helpful messages:

- On the PR page
    test

- On the Issues page
    test

It seems to me that these automatically appear as soon as there is a `CONTRIBUTING.md` in the repo, so I borrowed the structure of Rust's contributors guide and put some hints about ESPResSo.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,80 @@
+# Contributing to ESPResSo
+[contributing-to-espresso]: #contributing-to-espresso
+
+Thank you for your interest in contributing to ESPResSo! There
+are many ways to contribute and we appreciate all of them. This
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,81 @@
+# Contributing to ESPResSo
+[contributing-to-rust]: #contributing-to-rust
```
fweik: Rust?

hmenke: I copy-pasted the structure from the Rust contributors guide.

fweik (approved):

PR #1582, opened 2017-10-30
core: Don't exclude DPD and short-range electrostatics via exclusion…

fweik:
…s + test

Fixes #

Description of changes:
 - 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


RudolfWeeber:
Looks good.

If possible without a big effort, please extend the test for pressure/stress

Beyond that, it can be merged

PR #1583, opened 2017-10-30
Scafacos related fixes

RudolfWeeber:


src/core/scafacos.cpp
```diff
@@ -172,9 +173,31 @@ double pair_energy(Particle *p1, Particle *p2, double dist) {
...
+      }
+    }
+  }
+  return true;
+}
+
 void add_long_range_force() {
   particles.update_particle_data();
 
+  if (!check_position_validity(particles.positions)) {
```
fweik, line 198: Couldn't we just use folded coordinates instead of failing? Otherwise scafacos can only be used with strict particle sorting, which has a considerable performance impact.

src/core/initialize.cpp
```diff
@@ -219,6 +219,19 @@ void on_integration_start() {
   check_global_consistency();
 #endif
 
+#ifdef ADDITIONAL_CHECKS
+#ifdef ELECTROSTATICS
+  if (!Utils::Mpi::all_compare(comm_cart,coulomb.method))
+    runtimeErrorMsg() << "Nodes disagree about Coulomb long range method";
```
fweik, line 223: I'd prefer an exception here, if this condition fails there is almost certainly going to be a MPI deadlock. Alternatively we could check the runtime errors on the end of the function (if ADDITIONAL_CHECKS).

RudolfWeeber:
> fweik commented on this pull request.
> >  void add_long_range_force() {
>    particles.update_particle_data();
>  
> +  if (!check_position_validity(particles.positions)) {
> 
> Couldn't we just use folded coordinates instead of failing? Otherwise scafacos can only be used with strict particle sorting, which has a considerable performance impact.

I'll put in folding, but the check is still needed for a system with
open boundary conditions.

> 
> > @@ -219,6 +219,19 @@ void on_integration_start() {
>    check_global_consistency();
>  #endif
>  
> +#ifdef ADDITIONAL_CHECKS
> +#ifdef ELECTROSTATICS
> +  if (!Utils::Mpi::all_compare(comm_cart,coulomb.method))
> +    runtimeErrorMsg() << "Nodes disagree about Coulomb long range method";
> 
> I'd prefer an exception here, if this condition fails there is almost certainly going to be a MPI deadlock. Alternatively we could check the runtime errors on the end of the function (if ADDITIONAL_CHECKS).
The runtime errors are collected from integrate_vv() after calling
on_integration_start, so this should be safe.

fweik (approved):

PR #1584, opened 2017-10-31
Immersed Boundary with python interface

### LINKED ISSUE #1139: Implement Python support for IMMERSED_BOUNDARIES

sgekle:
Fixes #1139.

Description of changes:
 Python interface for Immersed Boundary

PR Checklist
------------
 - [X] Tests?
   - [ ] Interface
   - [ ] Core 
 - [X] Docs?


samples/immersed_boundary/sampleImmersedBoundary.py
```diff
@@ -0,0 +1,70 @@
...
+from writeVTK import WriteVTK
+WriteVTK(system, str(outputDir + "/cell_" + str(0) + ".vtk"))
+
+stepSize = 1000
+numSteps = 20
+
+for i in range(0, numSteps):
+
+    system.integrator.run(stepSize)
+    WriteVTK(system, str(outputDir + "/cell_" + str(i+1) + ".vtk"))
```
hmenke, line 70: There is [`system.part.writevtk`](http://espressomd.org/html/doc/espressomd.html#espressomd.particle_data.ParticleList.writevtk).

samples/immersed_boundary/writeVTK.py
```diff
@@ -0,0 +1,19 @@
+def WriteVTK(system, outFile):
```
hmenke, line 1: There is `system.part.writevtk`.

https://github.com/espressomd/espresso/blob/37dd602d01b73109804f2fdb14c70763768767ee/src/python/espressomd/particle_data.pyx#L1838-L1867

src/core/immersed_boundary/ibm_cuda.cu
```diff
@@ -129,20 +129,38 @@ void InitCUDA_IBM(const int numParticles)
     // Copy boundary velocities to the GPU
     // First put them into correct format
 #ifdef LB_BOUNDARIES_GPU
-    float* host_lb_boundary_velocity = new float[3*(n_lb_boundaries+1)];
+//    float* host_lb_boundary_velocity = new float[3*(n_lb_boundaries+1)];
```
hmenke, line 132: Please remove commented out code.  Don't worry about losing it, it will still be present in the history.

src/python/espressomd/interactions.pxd
```diff
@@ -312,6 +344,12 @@ cdef extern from "object-in-fluid/oif_global_forces.hpp":
     int oif_global_forces_set_params(int bond_type, double A0_g, double ka_g, double V0, double kv)
 cdef extern from "object-in-fluid/oif_local_forces.hpp":
     int oif_local_forces_set_params(int bond_type, double r0, double ks, double kslin, double phi0, double kb, double A01, double A02, double kal)
+cdef extern from "immersed_boundary/ibm_triel.hpp":
+    int IBM_Triel_SetParams(const int bond_type, const int ind1, const int ind2, const int ind3, const double max, const tElasticLaw elasticLaw, const double k1, const double k2)
+cdef extern from "immersed_boundary/ibm_tribend.hpp":
+    int IBM_Tribend_SetParams(const int bond_type, const int ind1, const int ind2, const int ind3, const int ind4, const double kb, const bool flat)
+cdef extern from "immersed_boundary/ibm_volume_conservation.hpp":
+    int IBM_VolumeConservation_SetParams(const int bond_type, const int softID, const double kappaV);
```
hmenke: These maybe need to be ifdef'ed to avoid linking errors?  What happens when you build without `#define IMMERSED_BOUNDARY`?

hmenke (changes requested):

RudolfWeeber:
Thanks for the pr, Stephan. It's good to have IBM support for Python. 
Also, I think, the lb tracer virtual sites could be very useful for other applications.

Going through the pr, I found the following points, mostly docs/testing:

* Could you please add some comments and/or docstrings to the sample scripts, explaining what each of the scripts does and what is the expected outcome of the script. Also, what is the content of the respective data files.

* Would it be possible to provide a test script, with which one can test the virtual sites behaviour, so we can be sure, it still works in the future. 
  Ideas might be:
  * Put a virtual site without external forces into a known flow field and check that it is advected correctly
  * Put a virtual site with an external force in a resting fluid and check the expected increase of the fluid momentum
  We will likely make virtual sites switchable at runtime before the release, so IMMERSED_BOUNDARIES can be included into the automatic CI testing and regressions from other changes will be noticed immediately. 
  
* Please reference the relevant Python bond classes for the bending rigidity and volume conservation etc. in the ibm.rst documentation via
```rst
  :class:`espressomd.interactions.IBMTribend`
```
  Additionally, please mention the sample scripts in the doc.

* Please mention in the docs whith which lb version(s) this works and if there are any limitations/incompatible features you are aware of.

* With regards to the bond classes in src/python/espressomd/interactions.pyx:
  * Please use pep8 naming conventions for the class names, i.e., CamelCase with capital 1st letter and no underscores
  * Please add docstrings to the classes, as e.g., for HarmonicBond
  * IBM_Triel:
    * I think, there is an extra {} in the return statement of _get_params_from_es_core
    * The back_translation (from int to string) of the elasticLaw seems to be missing in _get_params_from_es_core()
  * IBM_Tri_Ben: the _get_params_from_es_core() implementation seems to be missing?
  * Same goes for volume conservatoin  
  The easiest way to ensure consistency is to add your bonds in testsuite/interactions_bonded_interface.py. Not having consistency can lead to very surprising results in the simulation :-(

RudolfWeeber:
I adapted this to the new VirtualSites which are switchable at runtime.
This involved separating the two independent parts (lb tracer particles, volume conservation).
The code for the lb tracer particles is now in the src/core/virtual_sites folder, so I included it in the maxset configuration for CI testing.
To get the tracers, VIRTUAL_SITES_INERTIALESS_TRACERS needs to be defined in myconfig.hpp.
Just compinling that in only has a moderate performance penalty (temporarily storing the force densities on the lb nodes).
To activate the tracers, the virtual site implementation has to be set accordingly:
system.virtual_sites=VirtualSitesInertialessTracers()
There is a testcase for the advection part, both on cpu and gpu. There is not yet a test for tra\nsfer of forces from the particle to the lb fluid.

With regards to the volume conservation part:
In the initialization part (which sotres the reference volumes in the bonds marking the individaul soft objects), I removed a call to mpi_bcast_interaction_params(). This cannot (and never could) be called savely in parallel code and results in random outcomes. However, if I understand the code correctly, due to the mpi_allreduce, all modes are aware of the correct volumes anyway, and so no broadcasting of interaction parameters should be needed. @sgekle, could you please check?
This is in src/core/immersed_boundary/ImmersedBoundaries.cpp.

I adapted the sample script provided by @sgekle to run. Stephan, could you please check the outcome?
We'd also need the doc and py interface fixes mentioned in the earlier comments on this pr, since the release is approaching fast.

RudolfWeeber:
I had a look ad terminal velocities of a single particle with an ext_force in a channel with 2 walls (same system as the advection test).
* For LB GPU, just activating th VirtualSitesInertialessTracers changes the terminal velocity of a non-virtual particle (i.e., one coupled with inertia). This behvior is also present in the original pull request (commit d51a93), as far as I can tell.
* For LB CPU this is not the case.
* For virtual particles, the terminal velocities obtained from s.part[].v do not agree with the difference in positions divided by the difference in time before/after integration. 

I also thought, the terminal velocities should be identical for viscous and non-viscous coupling. This doesn't seem to be the case. Or is this assumption incorrect?

@sgekle could you take a look at this issues?

The test script is below. To test this with your original pr, you need to comment out the 2 virtual_sites-related lines and change LBFluidGPU to LBFLUID_GPU
The myconfig was the current maintainer/configs/maxset.hpp minues VIRTUAL_SITES_RELATIVE and everything electrokinetics related


import unittest as ut
import espressomd
from espressomd import System, lb, shapes, lbboundaries
import numpy as np
from espressomd.interactions import FeneBond
from espressomd.virtual_sites import VirtualSitesInertialessTracers, VirtualSitesOff

from tests_common import verify_lj_forces
from numpy import random


#@ut.skipIf(not espressomd.has_features("VIRTUAL_SITES_INERTIALESS_TRACERS","LB"),
#           "Test requires VIRTUAL_SITES_INERTIALESS_TRACERS")
class VirtualSitesTracers(ut.TestCase):
    s = espressomd.System()
    s.seed = range(s.cell_system.get_state()["n_nodes"])
    
    def test_advection(self):

        # System setup
        system = self.s
        system.virtual_sites=VirtualSitesInertialessTracers()
        system.time_step = 0.02
        system.cell_system.skin = 0.1
        
        box_height = 16. 
        box_lw=16
        system.box_l = box_lw,box_lw,box_height
        
        lbf = lb.LBFluidGPU(agrid=1, dens=1, visc=2, tau= system.time_step, fric = 1)
        system.actors.add(lbf)
        
        system.thermostat.set_lb(kT=0)
        
        # Setup boundaries
        walls = [lbboundaries.LBBoundary() for k in range(2)]
        walls[0].set_params(shape=shapes.Wall(normal=[0,0,1], dist = 0.5))
        walls[1].set_params(shape=shapes.Wall(normal=[0,0,-1], dist = -box_height-0.5))
        
        for wall in walls:
            system.lbboundaries.add(wall)
        
        # Establish steady state flow field
        system.part.add(id=0, pos=(0,5.5,5.5),virtual=0,ext_force=(10,0,0))
        for i in range(100):
            last_t=system.time
            last_x=system.part[0].pos
            system.integrator.run(500)
            print(system.part[0].v,(system.part[0].pos-last_x)/(system.time-last_t))
if __name__ == "__main__":
    #print("Features: ", espressomd.features())
    ut.main()

PR #1591, opened 2017-11-03
core: Enable verlet lists for all cellsystems. Some cleanup of no lon…

fweik:
…ger needed dependency on dd.

This should have been in the short-range PR.

Description of changes:
 - Enable verlet lists for all cellsystems by default
 - Don't force exact sorting if VL are disabled. As before
   skin == 0  can be used to get that.


RudolfWeeber:
Looks good to me.
One question:
Why are you replacing bool -> int for use_vlist?

Outside the scope of the pr:
Are we checking cell system flag/param consistency between nodes somewhere in the core? I didn't find it at first glance. If we don't, should we?

fweik:
Ah, the int thing is a regression (dd.use_vList was int), will fix.

The cell system consistency is at least partially checked due to ADDITIONAL_CHECKS and the CS specific tests that test the resorting. IIRC is the flag not checked, but this could (and should) be done in the same way as for coulomb.method.

fweik:
@RudolfWeeber int -> bool and check.

Issue #1558, opened 2017-10-22
Eliminate all compiler warnings before releasing Espresso 4.0


mkuron:
Our CI builds currently show hundreds of compiler warnings. We should take care of all of them before Espresso 4.0 is released. To ensure that no new warnings are ever introduced after that, we might want to add `-Werror` to the compiler flags, which makes all warnings fatal, and `-pedantic`, which shows warnings when the C++ standard is violated.

hmenke:
Some thoughts:
- We'd have to exclude `-pedantic` and `-Werror` from the Cython-generated code.
- We should aim to compile with `-Wall`.  Currently this triggers heaps of the following
  * `-Wsign-compare`
  * `-Wmissing-field-initializers`
  * `-Wreorder`
  * `-Wmaybe-uninitialized`
  * `-Wunused-variable`

  Apart from `-Wunused-variable` all of those indicate the possibility for undefined behaviour.  We should *really* fix all of those.

mkuron:
I‘ll try to do this at next week‘s coding day.

Fixing the `-Wall` Warnings you mention is going to be a huge amount of work, so I‘ll add them one by one after all other warnings have been fixed.

hmenke:
#1616 fixes all regular and `-pedantic` warnings from GCC 5.4.1 and all regular and `-pedantic` warnings for non-CUDA code from Clang 6.0.0.

PR #1594, opened 2017-11-07
Implementation of cylindrical observable

KaiSzuttor:
PR Checklist
------------
 - [x] Tests?
   - [x] Interface
   - [ ] Core 
 - [ ] Docs?


KaiSzuttor:
Binning in phi seems to have some problems... have to look into that.

src/core/observables/CylindricalFluxDensityProfile.hpp
```diff
@@ -0,0 +1,88 @@
...
+        double v_z = partCfg[id].m.v[2] / time_step;
+        // Write a flat histogram.
+        // index calculation: using the following formula for N dimensions:
+        //   ind = ind_{N-1} + sum_{j=0}^{N-2} (ind_j * prod_{k=j+1}^{N-1} n_k),
+        // where ind is the flattened index, ind_i is the ith unflattened index
+        // and n_i is the size of the ith dimension.
+        int ind =
+            3 * (r_bin * n_phi_bins * n_z_bins + phi_bin * n_z_bins + z_bin);
+        // The velocity transformation might yield nan if the denominator is 0.0 (Particle exactly on the center).
+        if (!(v_r != v_r)) {last_value[ind + 0] += v_r / bin_volume;}
```
fweik: I think `not std::isnan` or `std::isfinite` would make this so clear that you can even remove the comment.

src/core/observables/CylindricalProfileObservable.hpp
```diff
@@ -0,0 +1,36 @@
...
+  // Number of bins for each coordinate.
+  int n_r_bins, n_phi_bins, n_z_bins;
+  double r_bin_size() { return (max_r - min_r) / n_r_bins; }
+  double phi_bin_size() { return (max_phi - min_phi) / n_phi_bins; }
+  double z_bin_size() { return (max_z - min_z) / n_z_bins; }
+  virtual int n_values() const override {
+    return n_r_bins * n_phi_bins * n_z_bins;
+  };
+  ::Vector<3, double>
+  transform_to_cylinder_coordinates(::Vector<3, double> const &pos) const {
```
fweik: This is not specific to you class and of general use. Please make this a free function and put it into utils.

src/core/utils.hpp
```diff
@@ -1091,7 +1091,20 @@ void vecsub(T const *const a, T const *const b, T *const c) {
 template <typename T> int sign(T value) {
   return (T(0) < value) - (value < T(0));
 }
+
+/**
+*Returns the unraveled index of the provided flattened index.
+*/
+inline void unravel_index(const int* len_dims, const int ndims, const int flattened_index, int* unravelled_index_out){
```
fweik: Please move to separate file and write a unit test. Also the namespace should be Utils. Can you also please extent the doc to explain what this function does, the name makes this not clear to me.

src/script_interface/observables/CylindricalProfileObservable.hpp
```diff
@@ -0,0 +1,115 @@
...
+namespace ScriptInterface {
+namespace Observables {
+
+class CylindricalProfileObservable : public Observable {
+public:
+  const std::string name() const override {
+    return "Observables::CylindricalProfileObservable";
+  };
+
+  VariantMap get_parameters() const override {
```
fweik, line 42: This can be made more compact by using AutoParameters.

fweik (changes requested):

src/script_interface/observables/CylindricalProfileObservable.hpp
KaiSzuttor: Like how?

src/core/utils/Histogram.hpp
```diff
@@ -0,0 +1,20 @@
...
+* \brief Returns the unravelled index of the provided flat index.
+*        Therefore is the inversion of flattening an ndims dimensional index.
+* @param len_dims an int array of length ndims containing the lengths of the dimensions. (Input)
+* @param ndims int denoting the number of dimensions. (Input)
+* @flattened_index an int denoting the flat index. (Input)
+* @unravelled_index_out an int array with length ndims where the unflat indices are written to. (Output)
+*/
+inline void unravel_index(const int* const len_dims, const int ndims, const int flattened_index, int* unravelled_index_out){
+	//idea taken from http://codinghighway.com/2014/02/22/c-multi-dimensional-arrays-part-2-flattened-to-unflattened-index/
+	int mul[ndims];
```
hmenke: I'm sure this is some compiler extension.  Please use `std::vector`.

src/script_interface/observables/CylindricalProfileObservable.hpp
```diff
@@ -0,0 +1,115 @@
...
+    std::shared_ptr<::Observables::CylindricalProfileObservable>                    \
+    cylindrical_profile_observable() const override {                               \
+      return m_observable;                                                     \
+    }                                                                          \
+                                                                               \
+  private:                                                                     \
+    std::shared_ptr<::Observables::obs_name> m_observable;                     \
+  };
+
+NEW_RADIAL_PROFILE_OBSERVABLE(CylindricalFluxDensityProfile);
```
hmenke, line 110: Brrr!  Please make this a template.

hmenke (changes requested):

RudolfWeeber:
> +NEW_RADIAL_PROFILE_OBSERVABLE(CylindricalFluxDensityProfile);
> 
> Brrr!  Please make this a template.
I was forced to introduce that for the other observable types, as
well. So this is not new. 
I couldn't convince C++ to take the string literal for the
observable name as template parameter. 
If you are aware of the solution, we can implement it for all
observables, but I suggest to decouple this from this particular pr.

fweik:
template<const char *name>

src/core/utils/Histogram.hpp
KaiSzuttor: I wont touch that function since it's used in the reaction ensemble (where I actually ripped it off).

mkuron: That's a variable-length array, which is not legal syntax in C++. It is a GNU extension and supported by Clang, Intel and possibly others.
You can safely replace this line with `std::vector<int> mul(ndims);` to get the exact same functionality. If you don't do it now, somebody else will have to as we want to eliminate all compiler warnings before the release (#1558).

hmenke:
Strings are not allowed as template parameters because they do not have a fixed size at compile-time.  On the other hand it is definitely possible to achieve a nicer architecture using [CRTP](https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern), see the little example program below.  Bonus: No more virtual function overhead!  This construct is polymorphic **at compile-time**.
```cpp
#include <iostream>
#include <memory>
#include <string>

namespace Observables {
class ParticlePositions {};
} // namespace Observables

namespace ScriptInterface {
namespace Observables {

// Name method
template < typename Observable >
std::string get_name() = delete;

// Base class
template < typename CoreObservable >
class Observable {
    std::shared_ptr<CoreObservable> m_observable;
public:
    std::string name() const { return get_name<CoreObservable>(); }
};

// First derived
template < typename CoreObservable >
class PidObservable : public Observable<CoreObservable> {};

// Second derived
class ParticlePositions : public PidObservable<::Observables::ParticlePositions> {};
template <> std::string get_name<::Observables::ParticlePositions>() {
    return "Observables::ParticlePositions";
};

} // namespace Observables
} // namespace ScriptInterface

int main()
{
    ScriptInterface::Observables::ParticlePositions ppos;
    std::cout << ppos.name() << '\n';
}
```

RudolfWeeber:
I think, the discussion about refactoring the creation of some 50 observables without a large amount of code duplication should be split off from this pr.

hmenke (approved):

src/script_interface/observables/CylindricalProfileObservable.hpp
hmenke: This would require to refactor `Observable` to use `AutoParameters` (which is not a bad idea in and of itself).  However this would also require to adapt *all* the other observables to use `AutoParameters` which is a daunting task and most probably beyond the scope of this PR.

fweik: Ah that is true. Than we leave this for know, it is also easily replaced later, I already did several of those.

fweik (approved):

src/python/espressomd/utils.pyx
```diff
@@ -191,3 +192,40 @@ cdef handle_errors(msg):
...
+                      ``n_dims`` dimensional index.
+
+    Returns
+    -------
+    unravelled_index : array_like :obj:`int`
+                       An array containing the index for each dimension.
+
+    """
+    cdef int *c_len_dims
+    c_len_dims = <int *>malloc(len(len_dims) * cython.sizeof(int))
```
fweik: Use a vector instead of malloc+free.

src/python/espressomd/utils.pyx
```diff
@@ -191,3 +192,40 @@ cdef handle_errors(msg):
...
+    for i in range(len(len_dims)):
+        c_len_dims[i] = len_dims[i]
+    cdef int c_n_dims = n_dims
+    cdef int c_flattened_index = flattened_index
+    cdef int *unravelled_index_out
+    unravelled_index_out = <int *>malloc(n_dims * cython.sizeof(int))
```
fweik: Use a vector instead of malloc+free.

fweik (changes requested):

fweik (approved):

hmenke (approved):

PR #1595, opened 2017-11-07
py: script interface variant conversion handles numpy's int and float iqypes

### LINKED ISSUE #1174: Python object to variant conversion does not handle numpy float types
e.g. `np.sqrt(0.1**2+0.11**2+0.12**2)` is not converted to a double

RudolfWeeber:
Fixes #1174

### DIFF
--- a/src/python/espressomd/script_interface.pyx
+++ b/src/python/espressomd/script_interface.pyx
@@ -1,4 +1,5 @@
 from espressomd.utils import to_char_pointer,to_str
+import numpy as np
 
 cdef class PObjectId(object):
     cdef ObjectId id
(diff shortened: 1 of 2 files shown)


fweik (approved):

src/python/espressomd/script_interface.pyx
```diff
@@ -127,12 +128,15 @@ cdef class PScriptInterface(object):
...
-        elif type(value) == int:
-            return make_variant[int](value)
-        elif type(value) == float:
-            return make_variant[float](value)
+        
+        # Note: Bool has to be checked before int/float because
+        # numpy considers True to be an instance of np.int
         elif type(value) == type(True):
             return make_variant[bool](value)
+        elif type(value) == int or isinstance(value,np.int):
```
hmenke: Probably it should be
```python
elif np.issubdtype(np.dtype(type(value)), int):
```
This interactive session at least looks good (output in Python 2 and 3 is the same).  It also solves the problem that `bool` is falsely recognized as `np.int`.
```python
>>> value = 2
>>> np.issubdtype(np.dtype(type(value)), int)
True
>>> value = np.int64(2)
>>> np.issubdtype(np.dtype(type(value)), int)
True
>>> value = 2.0
>>> np.issubdtype(np.dtype(type(value)), int)
False
>>> value = np.float64(2.0)
>>> np.issubdtype(np.dtype(type(value)), int)
False
>>> value = True
>>> np.issubdtype(np.dtype(type(value)), int)
False
>>> value = "Definitely not a number"
>>> np.issubdtype(np.dtype(type(value)), int)
False
```

src/python/espressomd/script_interface.pyx
```diff
@@ -127,12 +128,15 @@ cdef class PScriptInterface(object):
...
+            return make_variant[int](value)
+        elif type(value) == float or isinstance(value,np.float):
```
hmenke: See other comment.
```python
elif np.issubdtype(np.dtype(type(value)), float):
```

hmenke (changes requested):

hmenke (approved):

PR #1601, opened 2017-11-08
Update mmm2d.cpp

KonradBreitsprecher:
Added minus to fulfill F = -dE/dz/.

### DIFF
--- a/src/core/mmm2d.cpp
+++ b/src/core/mmm2d.cpp
@@ -674,9 +674,9 @@ static double z_energy()
 
     MPI_Allreduce(&lcl_dm_z, &gbl_dm_z, 1, MPI_DOUBLE, MPI_SUM, comm_cart);
     // zero potential difference contribution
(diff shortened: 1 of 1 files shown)


jonaslandsgesell:
This change makes particle moves with monte carlo (energy dependent) with mmm2d and a capacitor and MD (using forces) consistent.

fweik:
Test is missing

fweik (changes requested):
Test is missing

KonradBreitsprecher:
I suggest to test E vs F via integrating forces of a particle pair for the following combinations (that are allowed):
- dielectric contrasts: off, 100, inf
- neutral, non-neutral
- const.pot w/wo applied potential
Same for ELC. I'm a bit scared doing this though..

fweik:
Do it, it's still better to know what is broken.

hmenke (approved):

KonradBreitsprecher:
Changes and tests are part of #1638

PR #1607, opened 2017-11-10
Fix #1557, which caused GPU testcases to silently be skipped in CI

mkuron:
This ~reverts~ fixes #1557. myconfig-final.hpp must be included after cmake_config.hpp because the sample myconfigs we have in the maintainer directory (and use for CI testing) have things like `#ifdef CUDA` so that the CUDA tests are only run on CUDA machines.

~This obviously means that the original aim of "Prevent[ing] user from declaring external FEATURES" cannot be achieved.~ We can just include cmake_config.hpp multiple times and undef the external features in here.


mkuron:
I've updated this pull request to use a few `#undef`s to achieve the goal of #1557 while still allowing myconfig to contain `#ifdef`s. Of course, this only works because `cmake_config.hpp` does not contain a ´#pragma once´.

fweik:
Will merge for now so that the tests run again, but this does not solve #1557.

mkuron:
Actually, it does fix it. It just undefs the external features, so people can add them to their myconfig and it won't harm. Of course, that's not the original intention of #1557 (which wanted to throw an error), but it's better than nothing for now.

PR #1610, opened 2017-11-13
Array properties return non-writable arrays with special message

### LINKED ISSUE #1550: Don't allow attempted change of vector components

KonradBreitsprecher:
Fixes #1550, replaces #1570 

Description of changes:
 - Added array_locked subclass of ndarray in utils.pyx, that throws the following error message upon `__setitem__`  or in-place operators:
_"Array properties return non-writable arrays and can only be modified as a whole, not in-place or componentwise. Use numpy.copy(<ESPResSo array property>) to get a writable copy."_
- Added testcase array_properties.py, testing for 
     - the exception for componentwise and in-place operators manipulation
     - unit test for allowed setters "as a whole"
     - unit test for allowed modification of a copy
- Added subsection in docu part.rst "Vectorial properties"


doc/sphinx/part.rst
```diff
@@ -63,6 +63,16 @@ Similarly, the position can be set::
 
 Note that the index and the property ID are not necessasirly the same.
 
+Vectorial properties
+~~~~~~~~~~~~~~~~~~~~
+
+For vectorial properties, componentwise manipulation like ``system.part[0].pos[0]
+= 1`` or in-place operators like ``+=`` or ``*=`` are not allowed and result in an error.
+This behavior is inherited, so the same accounts for ``a`` after ``a =
```
mkuron: accounts for => applies to

testsuite/array_properties.py
```diff
@@ -0,0 +1,167 @@
+from __future__ import print_function
+import unittest as ut
+import numpy as np
+import espressomd
+from espressomd import lb
+
+@ut.skipIf(not espressomd.has_features(["ROTATION", "PARTICLE_ANISOTROPY", "LANGEVIN_PER_PARTICLE", "DIPOLES", "EXTERNAL_FORCES", "LB", "EXCLUSIONS"]),
+           "Features not available, skipping test!")
```
mkuron: Please only skip the individual operations and not the entire test if a required features is not available.

src/python/espressomd/utils.pyx
```diff
@@ -141,42 +141,110 @@ cdef check_range_or_except(D, name, v_min, incl_min, v_max, incl_max):
...
+        obj = np.asarray(input_array).view(cls)
+        obj.flags.writeable = False
+        return obj
+
+    def __repr__(self):
+        return repr(np.array(self))
+
+    def __array_finalize__(self, obj):
+        if obj is None:
+            return
```
mkuron: This method is unnecessary because your child class doesn't have any data members. Your implementation is a no-op anyway.

src/python/espressomd/utils.pyx
```diff
@@ -141,42 +141,110 @@ cdef check_range_or_except(D, name, v_min, incl_min, v_max, incl_max):
...
+
+    def __setitem__(self, i, val):
+        raise UserWarning(array_locked.ERR_MSG)
```
mkuron: Should be ValueError, not UserError. Numpy raises `ValueError("assignment destination is read-only")`.

src/python/espressomd/utils.pyx
```diff
@@ -141,42 +141,110 @@ cdef check_range_or_except(D, name, v_min, incl_min, v_max, incl_max):
...
+    """
+    Returns a non-writable numpy.ndarray with a special error message upon usage
+    of __setitem__  or in-place operators. Cast return in __get__ of array
+    properties to array_locked to prevent these operations. 
+
+    """
+
+    ERR_MSG = "Array properties return non-writable arrays \
+    and can only be modified as a whole, not in-place or componentwise. \
+    Use numpy.copy(<ESPResSo array property>) to get a writable copy."
```
mkuron: Array properties => particle properties. componentwise => component-wise

mkuron (approved):

RudolfWeeber:
> +class array_locked(np.ndarray):
Can and/or should this be done in Cython to speed things up?
I.e., something like

cimport numpy as c_np
and use a cdef class inheriting from c_np.ndarray?

mkuron (approved):

KonradBreitsprecher:
Added the requested changes. 
@RudolfWeeber Not sure about the cython idea. Using ``cimport numpy as np`` and a ``cdef class ...`` results in ``'np.pxd' not found`` on the ``cdef class``.

fweik:
analyze_chains needs to be adapted.

PR #1611, opened 2017-11-13
Update analyze.pyx

jonaslandsgesell:
make check for none with "is" instead of "==". This allows pos to be a ndarray. The comparison of an ndarray with "==" None fails here since it produces an array again which cannot be used in the if condition.

Fixes #

Description of changes:
 - 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?

### DIFF
--- a/src/python/espressomd/analyze.pyx
+++ b/src/python/espressomd/analyze.pyx
@@ -124,10 +124,10 @@ class Analysis(object):
 
         """
 
-        if id == None and pos == None:
+        if id is None and pos is None:
             raise Exception("Either id or pos have to be specified\n" + __doc__)
 
-        if id != None and pos != None:
+        if (id is not None) and (pos is not None):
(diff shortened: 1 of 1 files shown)


fweik (approved):

src/python/espressomd/analyze.pyx
```diff
@@ -124,10 +124,10 @@ class Analysis(object):
 
         """
 
-        if id == None and pos == None:
+        if id is None and pos is None:
             raise Exception("Either id or pos have to be specified\n" + __doc__)
 
-        if id != None and pos != None:
+        if (id is not None) and (pos is not None):
```
KaiSzuttor, line 130: brackets not needed here

hmenke (approved):

PR #1612, opened 2017-11-13
testsuite: Enabled NPT test.

fweik:
There are at least two issues here: 
- Even on single core the test does not reach the expected accuracy (but still 4 significant places). 
- With several cores there is a problem with the globals.


hmenke (approved):

KaiSzuttor (approved):

KaiSzuttor:
this can be closed since @KonradBreitsprecher opened #1614

PR #1613, opened 2017-11-13
Observable lb

KaiSzuttor:
Description of changes:
 - contains #1594
 - adds an observable and test for the LB fluid velocity in polar coordinates at particle positions

PR Checklist
------------
 - [x] Tests?
   - [x] Interface
   - [x] Core 
 - [x] Docs?


src/core/observables/CylindricalLBFluxDensityProfileAtParticlePositions.hpp
```diff
@@ -0,0 +1,109 @@
+#ifndef OBSERVABLES_CYLINDRICALLBFLUXDENSITYPROFILEATPARTICLEPOSITIONS_HPP
+#define OBSERVABLES_CYLINDRICALLBFLUXDENSITYPROFILEATPARTICLEPOSITIONS_HPP
+
+#include "CylindricalProfileObservable.hpp"
+#include "utils.hpp"
+#include "utils/Histogram.hpp"
+#include "lbgpu.hpp"
+
+namespace Observables {
+class CylindricalLBFluxDensityProfileAtParticlePositions : public CylindricalProfileObservable {
```
fweik: Please move the implementation to a .cpp file. Then you don't have to drag in the other headers here, which are not actually part of the interface. That also decreases the build time.

src/core/utils/Histogram.hpp
```diff
@@ -1,3 +1,5 @@
+#ifndef HISTOGRAM_HPP
```
fweik: Should correspond to the file path, e.g. CORE_UTILS_HISTOGRAM_HPP, so that there can not be collisions.

fweik (changes requested):

KaiSzuttor:
just found there is still sth wrong...

KaiSzuttor:
Ready for review @fweik

fweik (approved):

PR #1619, opened 2017-11-14
Fix and test analysis.moment_of_inertia_matrix()

RudolfWeeber:
Fix docstring (accepts exactly one particle type)
Fix return value (3x3 matrix ready for np.linalg.eig)
Guard against missing keyword arg
Testcase with 2 partilce types


testsuite/analyze_itensor.py
```diff
@@ -0,0 +1,62 @@
...
+
+    @classmethod
+    def setUpClass(self):
+        box_l = 50.0
+        self.system.box_l = [box_l, box_l, box_l]
+        self.system.cell_system.skin = 0.4
+        self.system.time_step = 0.01
+        self.system.thermostat.turn_off()
+        for i in range(10):
+            self.system.part.add(id=i, pos=np.random.random(3) * box_l,type=0)
```
KaiSzuttor, line 21: ~~shouldnt we try to keep tests deterministic, i.e. setting a fixed seed for numpy random?~~

testsuite/analyze_itensor.py
```diff
@@ -0,0 +1,62 @@
...
+        return I
+        
+  
+        
+    
+    def test(self):
+        # Partilces of type 0
+        I0=self.i_tensor(range(0,10,1))
+        
+        self.assertTrue(np.allclose(I0, self.system.analysis.momentofinertiamatrix(p_type=0),atol=1E-9))
```
KaiSzuttor: please use numpy testing asserts for array comparison

KaiSzuttor (changes requested):

testsuite/analyze_itensor.py
KaiSzuttor: resolved by offline discussion: this test should pass for all given positions therefore it does not harm to use random positions

testsuite/analyze_itensor.py
KaiSzuttor:
```python
numpy.testing.assert_allclose(I0, self.system.analysis.momentofinertiamatrix(p_type=0), atol=1E-9))
```

PR #1620, opened 2017-11-14
pep8: correction for naming convention violations

### LINKED ISSUE #1590: Py: pep8 class name violations
Angle_Harmonic, Angle_..., Subt_Lj

should probably be fixed before the release.

KaiSzuttor:
Fixes #1590


RudolfWeeber:
nb_hood should not be underscored (neighberhood = single word)
Feature MMM1D_GPU referes to myconfig.hpp-feature (which wasn't renamed). I suggest to leave the feature names alone to avoid a ton of regression potential in the core.

Once those two are fixed, I think this should be merged quickly to avoid collisions with other pr.
Furthermore, we need to write to the lists, once this is merged.

KaiSzuttor:
But the class name can still be different from the featurename just as for LBGPU?!

fweik (approved):

RudolfWeeber:
On Tue, Nov 14, 2017 at 02:43:47PM +0000, Kai Szuttor wrote:
> But the class name can still be different from the featurename just as for LBGPU?!
Sure. I took the text fragment to mean the MMM1D_GPU feature name in
myconfig.hpp, though. (Not sure, why this needs to be a feature in
the 1st place...)

fweik:
@RudolfWeeber Because it has a short-range part iirc.

Issue #1360, opened 2017-08-30
Do we need a Brownian Dynamics (BD) integrator ?


bogdan-tanygin:
Dear ESPResSo team,

My team always had been using the BD integrators for ferrofluid research: regular magnetite ferrofluid, conventional parameters, nothing special. We just need more particles and larger _in silico_ durations.

Main question: are other users requesting this integrator? I'm asking because the SD and BD has been removed: https://github.com/espressomd/espresso/pull/1094

Hence, before moving forward, I've decided to ask an opinion.

**pros**:  at least, in case of the magnetite ferrofluid, BD provides 10-100 times speed up of _in silico_ timeframes which could be simulated compare to Velocity Verlet Langevin Dynamics (VV LD). Stability is also much improved. The reason is simple: BD does not have anymore the restriction `time_step < mass / gamma`. Also, it is worth to mention, that if we are using _real_ friction coefficients, i.e. `gamma_trans` and `gamma_rot` which are calculated according to the translational and rotational Stokes' law and the parameters are real material parameters (not ones selected for a simulation optimization) then VV LD usually is spending most part of computational resources to either translation or rotational motion while keeping an overall computational problem to be not well balanced. The BD does not have these issues.

**cons**: BD requires just position updates. Velocities are already analytically integrated. Hence, this approach does not organically fits to Velocity Verlet integrator. I've integrated BD directly into ESPResSo VV integrator for my personal usage. Architecturally it looks a bit artificial even though it is well theoretically derived. I could show my branch if it is of interest. Another cons: due to larger time_step and significant particle rotations, BD is not fully compatible with ESPResSo quaternion rotation mechanics. Again, I've integrated it a bit artificially to the existing architecture and all tests are passed.

Yours faithfully,
Dr. Theo Tanygin

hmenke:
You are correct that both, SD and BD, were removed in #1094.  While the removal of SD was deliberate, the removal of BD was a mere consequence of the interwoven nature of implementation of the two methods.

In a past ESPResSo team meeting we discussed that we would want to keep BD, so if you have a proper implementation which does not do nonsensical things like duplicating the integration loop, as the removed variant did, we are happy to receive your pull request.

Regarding the conflicting features you mentioned, it is all about documentation in my opinion.  If it is properly documented that quaternions cannot be used with BD and a user decides to ignore that note, well, bad luck.

mkuron:
[...]

PR #1622, opened 2017-11-14
Anisotropic rotational diffusion test

bogdan-tanygin:
Description of changes:
 - I've added the anisotropic rotational diffusion test. It is similar to a translational one which we already have in the `mass-and-rinertia_per_particle.py`. I've taken some measures to avoid a performance impact: limited loops for this specific calc and only 1 test case. Compare to overall bunch of python tests run, probably, there is not significant increase now: 17 minutes vs 15 minutes before. However, of course, up to the team to decide whether it is worth that. In my opinion: the rotational diffusion is a critically important physical process in Langevin Dynamics and its variance should be validated.
- Refactoring in the lab<->body frame of reference transformation: now it is tests common based script.

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


bogdan-tanygin:
Maybe, it makes sense to have some tests (like this one) commented out or removed from the file `testsuite/CMakeLists.txt`. This test is performance consuming. However, the existing rotational motion algorithm is well developed and barely will fail some tests. Such tests are probably needed for some long period validations during significant MD changes. Particularly, I've planned to use it to show that the new Brownian Dynamics passes the same tests as existing LD. This is why I've opened this pull request (in a test driven way).

bogdan-tanygin:
This is one of tests which validate both Velocity Verlet LD and future BD. Hence, it indirectly relates to https://github.com/espressomd/espresso/issues/1360.

bogdan-tanygin:
It is already a part of #1842.
I've not closed it only because I was not sure about the future of #1842. If it is already approved for the release 4.1 then we can close this PR.

RudolfWeeber:
The mass and rinertia test reached a complexity so that I don't really understand it anymore.
Could you please clarify in which way this pr goes beyond what is already tested in the Langevin thermostat testcase?
That  tests roational and translatioal diffusion coefficients with global and per particle gamma and temperature via Green-Kubo and also it tests Maxwell-Boltzmann velocity distributions.

For rotation, only isotropic inertia tensors are tested.
Does this test go beyond that?

bogdan-tanygin:
@RudolfWeeber thanks for the feedback. I agree regarding the complexity, I already have the corresponding task #1905. The tests improvement here are tightly connected to the BD and LD results match comparison. I also need to backport the latest changes from the #1842.

> Does this test go beyond that?

1. Yes it does, this is an anisotropic rotational diffusion test. However, it requires too much of the test runtime. This is why I've relocated this part to the non-automically run test mass-and-rinertia_per_particle_rotdiff-longrun.py in the #1842.

2. Except the anisotropic rotational diffusion, by a physical nature, you are right. The Langevin thermostat testcase, indeed, validates fundamentally the same things. The mass and rinertia test does the same but as an integrated (/summed/averaged) practical result (variance = 2 * D * time).

Shortly, a relation is like this:

[langevin_thermostat.py] + [Es. VV integrator] + statistical rules = [mass-and-rinertia_per_particle]

If we are fully sure that the VV integrator is properly tested by other tests then, indeed, we could consider a removal of the test mass-and-rinertia_per_particle with a check that all test parameters are ported to the langevin_thermostat.

bogdan-tanygin:
This PR depends on #2021.

bogdan-tanygin:
For more durable runs I've detected the instability of the test. It could be either a technical or a conceptual issue. In the latter case extra theoretical/bibliography study is needed. Anyway, the present PR is not consistent anymore. I'll work on this separately (#2029).. Propose to close it.

PR #1626, opened 2017-11-14
reaction ensemble refactoring

jonaslandsgesell:
Fixes #1620 in reaction ensemble, removes global cython object. Adds test for wang-landau reaction ensemble so it does not break by accident.


fweik:
You have to use the conversion functions from utils (to_char_pointer), because python 3.

```
File "wang_landau_reaction_ensemble.py", line 34, in <module>
    class ReactionEnsembleTest(ut.TestCase):
  File "wang_landau_reaction_ensemble.py", line 90, in ReactionEnsembleTest
    filename="energy_boundaries.dat", delta=0.05)
``

File: testsuite/reaction_ensemble.py [hunk 1]
```diff
@@ -148,14 +148,19 @@ def test_reaction_system(self):
...
             places=9,
-            msg="reaction ensemble temperature not set correctly.")
+            msg="reaction ensemble exclusion radius not set correctly.")
 
         self.assertAlmostEqual(
             ReactionEnsembleTest.volume,
             ReactionEnsembleTest.RE.get_volume(),
             places=9,
-            msg="reaction ensemble temperature not set correctly.")
+            msg="reaction ensemble golume not set correctly.")
```
hmenke: volume?

src/python/espressomd/reaction_ensemble.pyx
```diff
@@ -37,10 +38,18 @@ cdef class reaction_ensemble(object):
                        gradients from the short ranged interaction potential
                        when using the Reaction ensemble together with a MD
                        scheme.
-        
+
     """
-    def __init__(self,*args,**kwargs):
-        self._params=self._default_params()
+    cdef object _params
+    cdef cReactionEnsemble* RE
```
fweik: Why do you use a pointer here? Can't you just do cdef cReactionEnseble RE, and drop the explicit memory management?

jonaslandsgesell: Well, if we write it directly like this I get a comiler crash in line 52:
       self.RE = cReactionEnsemble()


/tikhome/jlandsgesell/src/espresso/src/python/espressomd/reaction_ensemble.pyx:52:35: Compiler crash in AnalyseExpressionsTransform

ModuleNode.body = StatListNode(reaction_ensemble.pyx:1:0)
StatListNode.stats[2] = CClassDefNode(reaction_ensemble.pyx:12:5,
    as_name = u'ReactionEnsemble',
    class_name = u'ReactionEnsemble',
    doc = u'\n    ...
    module_name = u'',
    visibility = u'private')
CClassDefNode.body = StatListNode(reaction_ensemble.pyx:13:4)
StatListNode.stats[0] = DefNode(reaction_ensemble.pyx:46:4,
    modifiers = [...]/0,
    name = u'__init__',
    num_required_args = 1,
    py_wrapper_required = True,
    reqd_kw_flags_cname = '0',
    used = True)
File 'Nodes.py', line 430, in analyse_expressions: StatListNode(reaction_ensemble.pyx:48:8)
File 'Nodes.py', line 4775, in analyse_expressions: SingleAssignmentNode(reaction_ensemble.pyx:52:35)
File 'Nodes.py', line 4887, in analyse_types: SingleAssignmentNode(reaction_ensemble.pyx:52:35)

Compiler crash traceback from this point on:
  File "/usr/lib/python2.7/dist-packages/Cython/Compiler/ExprNodes.py", line 4568, in analyse_as_type_constructor
    self.function = RawCNameExprNode(self.function.pos, constructor.type)
AttributeError: 'NoneType' object has no attribute 'type'
src/python/espressomd/CMakeFiles/reaction_ensemble.dir/build.make:92: recipe for target 'src/python/espressomd/reaction_ensemble.cpp' failed
make[2]: *** [src/python/espressomd/reaction_ensemble.cpp] Error 1
make[2]: *** Deleting file 'src/python/espressomd/reaction_ensemble.cpp'
CMakeFiles/Makefile2:4246: recipe for target 'src/python/espressomd/CMakeFiles/reaction_ensemble.dir/all' failed
make[1]: *** [src/python/espressomd/CMakeFiles/reaction_ensemble.dir/all] Error 2
Makefile:127: recipe for target 'all' failed
make: *** [all] Error 2

jonaslandsgesell: It looks to me like this reported bug in cython: https://github.com/cython/cython/issues/1791

fweik:
Does it help to define a constructor as suggested in the issue?

jonaslandsgesell:
No, I had the constructor declared in reaction_ensemble.hpp and defined in reaction_ensemble.cpp (there as ReactionEnsemble::ReactionEnsemble(){} ). The constructor was also exposed to cython in the pxd file as shown in https://cythoncython.readthedocs.io/en/latest/src/userguide/wrapping_CPlusPlus.html.

Turns out it was about the naming clash in the cython code (cdef class ReactionEnsemble and the cpp class ReactionEnsemble)

jonaslandsgesell:
Nope recompiled and the error was back. It s not about the naming clash...

KaiSzuttor:
The naming should not be a problem if you declare the cpp names in a pxd according to the documentation.

KaiSzuttor:
see kaiszuttor@39ee61f792f06e569fb4e6a0a9fe160add1c4f9f

fweik:
Please put the particle type stuff into an extra PR as this is independent of the RE refactoring.

kosovan:
@jonaslandsgesell Is the reaction ensemble refactoring done, so that I can work on the documentation of python functions?

jonaslandsgesell:
Hey Peter it s not yet merged but you can introduce chnages in this branch
directly without introducing merge conflicts.

Am 03.12.2017 9:53 Nachm. schrieb "Peter Kosovan" <notifications@github.com
>:

@jonaslandsgesell <https://github.com/jonaslandsgesell> Is the reaction
ensemble refactoring done, so that I can work on the documentation of
python functions?

—
You are receiving this because you were mentioned.

<https://github.com/espressomd/espresso/pull/1626#issuecomment-348813653>,
or mute the thread

samples/reaction_ensemble.py
```diff
@@ -23,7 +23,6 @@
 from espressomd import integrate
 from espressomd.interactions import *
 from espressomd import reaction_ensemble
-from espressomd import grand_canonical
 import numpy as np
 
 dev = "cpu"
```
KaiSzuttor: can be removed

samples/reaction_ensemble.py
```diff
@@ -23,7 +23,6 @@
 from espressomd import integrate
 from espressomd.interactions import *
 from espressomd import reaction_ensemble
-from espressomd import grand_canonical
 import numpy as np
```
KaiSzuttor: external imports before internal

samples/reaction_ensemble.py
```diff
@@ -62,20 +61,19 @@
...
 RE = reaction_ensemble.ReactionEnsemble(
     standard_pressure=0.00108, temperature=1, exclusion_radius=1)
 RE.add(equilibrium_constant=K_diss, reactant_types=[0], reactant_coefficients=[
        1], product_types=[1, 2], product_coefficients=[1, 1])
 RE.set_default_charges(dictionary={"0": 0, "1": -1, "2": +1})
 print(RE.get_status())
-grand_canonical.setup([0, 1, 2])
+system.setup([0, 1, 2])
 
 # RE.pH=2
```
KaiSzuttor: remove commented out code

samples/reaction_ensemble.py
```diff
@@ -62,20 +61,19 @@
...
 
 for i in range(10000):
     RE.reaction()
 #	RE.reaction_constant_pH()
```
KaiSzuttor: remove

src/core/particle_data.cpp
```diff
@@ -520,32 +519,33 @@ int set_particle_mu_E(int part, double mu_E[3]) {
...
 
-int set_particle_type(int part, int type) {
-  auto const pnode = get_particle_node(part);
+int set_particle_type(int p_id, int type) {
+  auto const pnode = get_particle_node(p_id);
   make_particle_type_exist(type);
 
-  if (pnode == -1)
+  if (pnode == -1){
+    throw std::runtime_error("pnode not found\n");
```
KaiSzuttor: shouldnt the get_particle_node function throw instead

src/core/particle_data.cpp
```diff
@@ -61,12 +62,9 @@
  * variables
  ************************************************/
 // List of particles for grandcanonical simulations
-TypeOfIndex Type;
-IndexOfType Index;
-TypeList *type_array;
 int number_of_type_lists;
-int GC_init;
-int Type_array_init;
+bool GC_init;
```
KaiSzuttor: Are those globals necessary?

src/core/particle_data.cpp
```diff
@@ -741,27 +741,28 @@ void remove_all_particles() {
...
-  auto cur_par = get_particle_data(part);
-
+int remove_particle(int p_id) {
+  auto cur_par = get_particle_data(p_id);
   if (cur_par) {
-    int type = cur_par->p.type;
-    if (remove_id_type_array(part, type) == ES_ERROR)
-      return ES_ERROR;
+    if(GC_init==true){
+        //remove particle from its current type_list
```
KaiSzuttor: Why is the function not named remove_id_from_type_list? Then you wouldn't need the comment

src/core/particle_data.cpp
```diff
@@ -1105,7 +1106,7 @@ void auto_exclusions(int distance) {
   /* setup bond partners and distance list. Since we need to identify particles
      via their identity, we use a full sized array */
   partners =
-      (IntList *)Utils::malloc((max_seen_particle + 1) * sizeof(IntList));
+      (IntList *)calloc((max_seen_particle + 1), sizeof(IntList));
```
KaiSzuttor: Why do you switched from Utils::malloc?

src/core/particle_data.cpp
```diff
@@ -1170,310 +1171,54 @@ void auto_exclusions(int distance) {
...
-  int l_err = 1;
-  // type i not indexed, so no list for this particle exists
-  for (int i = 0; i < Type.max_entry; i++) {
-    if (Type.index[i] == type) {
-      l_err = 0;
-      break;
+int get_random_p_id(int type){
+    if(ParticleTypeMap.at(type).size()==0)
+        throw std::runtime_error("No particles of given type could be found");
+    int i=0;
```
KaiSzuttor: please give "i" a more descriptive name

src/core/particle_data.cpp
```diff
@@ -741,27 +741,28 @@ void remove_all_particles() {
...
-    return ES_ERROR;
+    throw std::runtime_error("particle could not be retrieved during remove");
   }
 
-  auto const pnode = get_particle_node(part);
+  auto const pnode = get_particle_node(p_id);
 
   if (pnode == -1)
-    return ES_ERROR;
+    throw std::runtime_error("particle node could not be retrieved");
```
KaiSzuttor: To be consistent, start with the Exception message with capital letter

src/core/particle_data.cpp
```diff
@@ -1170,310 +1171,54 @@ void auto_exclusions(int distance) {
...
-      l_err = 0;
-      break;
-    }
-  }
-  if (l_err) {
-    return ES_ERROR;
-  }
-  if (type_array[Index.type[type]].max_entry == 0)
-    return ES_ERROR;
+void find_particle_type(int type, int *id) {
```
KaiSzuttor: Please don't use output variables

src/core/particle_data.cpp
```diff
@@ -1170,310 +1171,54 @@ void auto_exclusions(int distance) {
...
-      indexed = 1;
-      break;
-    }
-  }
-  if (indexed) {
-    *number = type_array[Index.type[type]].max_entry;
-    return ES_OK;
-  }
-  return NOT_INDEXED;
+void number_of_particles_with_type(int type, int *number) {
```
KaiSzuttor: Please don't use output variables

src/core/particle_data.hpp
```diff
@@ -929,70 +929,18 @@ void try_add_exclusion(Particle *part, int part2);
...
 // flag indicating init_gc was called
-extern int GC_init;
-
-// flag that indicates that the function init_type_array was called already
-extern int Type_array_init;
-
-int init_gc(void);
+extern bool GC_init;
 
 /** init particle lists		*/
```
KaiSzuttor: Is it a particle list or is it a particle map. Remove the comment if the name of the function already makes clear what happens

src/core/particle_data.hpp
```diff
@@ -929,70 +929,18 @@ void try_add_exclusion(Particle *part, int part2);
...
-int reallocate_global_type_list(int size);
-
-/** free particle lists		*/
-int free_particle_lists(void);
-
-// update particle list
-int update_particle_array(int type);
+void init_type_map(int type);
 
 /* find a particle of given type and return its id */
```
KaiSzuttor, line 958: This comment is also wrong and useless, please remove

src/core/particle_data.hpp
```diff
@@ -929,70 +929,18 @@ void try_add_exclusion(Particle *part, int part2);
...
-int find_particle_type(int type, int *id);
-
-/** return an array with real particle id and the corresponding index of
- * typelist */
-int find_particle_type_id(int type, int *id, int *in_id);
+void find_particle_type(int type, int *id);
```
KaiSzuttor: don't use output variables

src/core/reaction_ensemble.cpp
```diff
@@ -1,1724 +1,1970 @@
...
-#include <vector>
-#include "utils.hpp" // for PI and random vectors
-#include "utils/Histogram.hpp"
 #include "partCfg_global.hpp"
+#include "particle_data.hpp" //for particle creation, modification
+#include "random.hpp"        //for random numbers
+#include "statistics.hpp"    //for distto
+#include "utils.hpp"         // for PI and random vectors
+#include "utils/Histogram.hpp"
+#include <fstream> //for std::ifstream, std::ofstream for input output into files
```
KaiSzuttor: the comments are common knowledge, can be removed

src/core/reaction_ensemble.hpp
```diff
@@ -1,312 +1,376 @@
...
+
+struct StoredParticleProperty {
   int p_id;
   double charge;
   int type;
-} stored_particle_property;
+};
 
-typedef struct collective_variable {
+struct collective_variable {
```
KaiSzuttor: Naming convention CamelCase

src/core/reaction_ensemble.hpp
```diff
@@ -1,312 +1,376 @@
...
-  bool do_not_sample_reaction_partition_function;
-} wang_landau_system;
-
-double calculate_degree_of_association(int index_of_current_collective_variable,
-                                       void *m_wang_landau_system);
-double calculate_current_potential_energy_of_system_wrap(
-    int unimportant_int, void *unimportant_wang_landau_system);
-
-class ReactionEnsemble {
+  virtual double determine_current_state() {
```
KaiSzuttor: override

src/core/reaction_ensemble.hpp
```diff
@@ -1,312 +1,376 @@
...
+  virtual double determine_current_state() = 0; // use pure virtual, otherwise
+                                                // this will be used in vector
+                                                // of collective variables
+};
+
+class WangLandauReactionEnsemble;
+
+struct EnergyCollectiveVariable : public collective_variable {
+  std::string energy_boundaries_filename;
+  virtual double determine_current_state() {
```
KaiSzuttor: override

src/core/reaction_ensemble.hpp
```diff
@@ -1,312 +1,376 @@
...
+private:
+  /**
+  * Returns the degree of association for the current collective variable. This
+  * is needed since you may use multiple degrees of association as collective
+  * variable for the Wang-Landau algorithm.
+  */
+  double calculate_degree_of_association() {
+    int total_number_of_corresponding_acid = 0;
+    for (int corresponding_type_i = 0;
+         corresponding_type_i < this->corresponding_acid_types.size();
```
KaiSzuttor: instead of using "this", you should use the convention of "m_" prefix for class members

src/core/reaction_ensemble.hpp
```diff
@@ -1,312 +1,376 @@
...
+
+private:
+  /**
+  * Returns the degree of association for the current collective variable. This
+  * is needed since you may use multiple degrees of association as collective
+  * variable for the Wang-Landau algorithm.
+  */
+  double calculate_degree_of_association() {
+    int total_number_of_corresponding_acid = 0;
+    for (int corresponding_type_i = 0;
```
KaiSzuttor: if the scope of the variable resides within the for loop, you can choose a shorter name to increase readability

src/core/reaction_ensemble.hpp
```diff
@@ -1,312 +1,376 @@
...
+    if (total_number_of_corresponding_acid == 0) {
+      throw std::runtime_error("Have you forgotten to specify all "
+                               "corresponding acid types? Total particle "
+                               "number of corresponding acid type is zero\n");
+    }
+    int num_of_associated_acid;
+    number_of_particles_with_type(this->associated_type,
+                                  &num_of_associated_acid);
+    double degree_of_association =
+        double(num_of_associated_acid) /
```
KaiSzuttor: use static_cast<double>() instead of c-style cast

src/core/reaction_ensemble.hpp
```diff
@@ -1,312 +1,376 @@
...
-  // is an input parameter for the reaction
-  // ensemble;
-    
+  ReactionAlgorithm(){};
+  virtual ~ReactionAlgorithm(){};
+
+  int nr_single_reactions = 0;
+  std::vector<SingleReaction> reactions = std::vector<SingleReaction>();
+  std::vector<int> type_index = std::vector<int>();
+  int nr_different_types = 0; // is equal to length type_index
```
KaiSzuttor: equal to size of type_index, but why do I need a variable for that then?

src/core/reaction_ensemble.hpp
```diff
@@ -1,312 +1,376 @@
...
-  // the standard_pressure_in_simulation_units
-  // is an input parameter for the reaction
-  // ensemble;
-    
+  ReactionAlgorithm(){};
+  virtual ~ReactionAlgorithm(){};
+
+  int nr_single_reactions = 0;
+  std::vector<SingleReaction> reactions = std::vector<SingleReaction>();
+  std::vector<int> type_index = std::vector<int>();
```
KaiSzuttor: std::vector<int> type_index;

src/core/reaction_ensemble.hpp
```diff
@@ -1,312 +1,376 @@
...
-  };
-  // the standard_pressure_in_simulation_units
-  // is an input parameter for the reaction
-  // ensemble;
-    
+  ReactionAlgorithm(){};
+  virtual ~ReactionAlgorithm(){};
+
+  int nr_single_reactions = 0;
+  std::vector<SingleReaction> reactions = std::vector<SingleReaction>();
```
KaiSzuttor: std::vector<SingleReaction> reactions;

src/core/reaction_ensemble.hpp
```diff
@@ -1,312 +1,376 @@
...
-    false, // bool box_has_wall_constraints
-    -10.0, // double slab_start_z
-    -10.0, // double slab_end_z
-    100 // int non_interacting_type
-  };
-  // the standard_pressure_in_simulation_units
-  // is an input parameter for the reaction
-  // ensemble;
-    
+  ReactionAlgorithm(){};
```
KaiSzuttor, line 83: Why so many public members? Are you sure about using a single class here?

src/core/reaction_ensemble.hpp
```diff
@@ -1,312 +1,376 @@
...
+class ReactionEnsemble : public ReactionAlgorithm {
+public:
+  double calculate_boltzmann_factor(
+      SingleReaction &current_reaction, double E_pot_old, double E_pot_new,
+      std::vector<int> &old_particle_numbers, int dummy_old_state_index,
+      int dummy_new_state_index,
+      bool dummy_only_make_configuration_changing_move);
+};
+
+class WangLandauReactionEnsemble : public ReactionAlgorithm {
```
KaiSzuttor: Maybe another file?

src/core/reaction_ensemble.cpp
```diff
@@ -1,1724 +1,1970 @@
...
+  char *line = nullptr;
+  size_t len = 0;
+  ssize_t length_line;
+  const char *delim = "\t ";
+  length_line = getline(&line, &len, pFile); // dummy call of getline to get rid
+                                             // of header line (first line in
+                                             // file)
+  while ((length_line = getline(&line, &len, pFile)) != -1) {
+    int counter_words_in_line = 0;
+    for (char *word = strtok(line, delim); word != nullptr;
```
KaiSzuttor: c-code

src/core/reaction_ensemble.cpp
```diff
@@ -1,1724 +1,1970 @@
...
-            if(status==ES_ERROR)
-                 throw std::runtime_error("could not initialize gc particle list for types\n");
-
+void ReactionAlgorithm::add_reaction(double equilibrium_constant,
+                                     std::vector<int> _reactant_types,
+                                     std::vector<int> _reactant_coefficients,
+                                     std::vector<int> _product_types,
+                                     std::vector<int> _product_coefficients) {
+  SingleReaction new_reaction;
+  int len_reactant_types = (int)_reactant_types.size();
```
KaiSzuttor: c-style cast, use static_cast

src/core/reaction_ensemble.cpp
```diff
@@ -1,1724 +1,1970 @@
...
+  int len_product_types = (int)_product_types.size();
+  new_reaction.len_reactant_types = len_reactant_types;
+  new_reaction.len_product_types = len_product_types;
+
+  new_reaction.equilibrium_constant = equilibrium_constant;
+  new_reaction.reactant_types = _reactant_types;
+  new_reaction.len_reactant_types = len_reactant_types;
```
KaiSzuttor: use .size() of the container

src/core/reaction_ensemble.cpp
```diff
@@ -1,1724 +1,1970 @@
...
+    // particles of reactant_types(i) to product_types(i)
+    for (int j = 0; j < std::min(current_reaction.product_coefficients[i],
+                                 current_reaction.reactant_coefficients[i]);
+         j++) {
+      append_particle_property_of_random_particle(
+          current_reaction.reactant_types[i], changed_particles_properties);
+      replace_particle(changed_particles_properties.back().p_id,
+                       current_reaction.product_types[i]);
+    }
+    // create product_coefficients(i)-reactant_coefficients(i) many product
```
KaiSzuttor, line 230: i don't understand the comment

src/core/reaction_ensemble.cpp
```diff
@@ -1,1724 +1,1970 @@
...
-	const double beta =1.0/m_current_reaction_system.temperature_reaction_ensemble;
-	const double standard_pressure_in_simulation_units=m_current_reaction_system.standard_pressure_in_simulation_units;
-	//calculate boltzmann factor
-	return std::pow(volume*beta*standard_pressure_in_simulation_units, current_reaction->nu_bar) * current_reaction->equilibrium_constant * factorial_expr * exp(-beta * (E_pot_new - E_pot_old));
+double ReactionEnsemble::calculate_boltzmann_factor(
+    SingleReaction &current_reaction, double E_pot_old, double E_pot_new,
+    std::vector<int> &old_particle_numbers, int dummy_old_state_index,
+    int dummy_new_state_index,
+    bool dummy_only_make_configuration_changing_move) {
+  /**calculate the acceptance probability in the reaction ensemble */
```
KaiSzuttor, line 346: if the comment is correct, why don't you call it acceptance_probability?

src/core/reaction_ensemble.cpp
```diff
@@ -1,1724 +1,1970 @@
...
+         current_reaction.equilibrium_constant * factorial_expr *
+         exp(-beta * (E_pot_new - E_pot_old));
 }
 
 /**
 *generic one way reaction
 *A+B+...+G +... --> K+...X + Z +...
-*you need to use 2A --> B instead of A+A --> B since in the last case you assume distinctness of the particles, however both ways to describe the reaction are equivalent in the thermodynamic limit
-*further it is crucial for the function in which order you provide the reactant and product types since particles will be replaced correspondingly! If there are less reactants than products, new product particles are created randomly in the box. Matching particles simply change the types. If there are more reactants than products, old reactant particles are deleted.
+*you need to use 2A --> B instead of A+A --> B since in the last case you assume
```
KaiSzuttor: Why is a reaction described right before the save_old_particle_numbers function?

src/core/reaction_ensemble.cpp
```diff
@@ -1,1724 +1,1970 @@
...
+#endif
+  // set type
+  int err_code_type =
+      set_particle_type(p_id, non_interacting_type);
+  return err_code_type;
 }
 
 /**
-* Deletes the particle with the given p_id. This method is intended to only delete unbonded particles since bonds are coupled to ids. The methods is aware of holes in the particle id range and fills them if a particle gets created.
+* Deletes the particle with the given p_id. This method is intended to only
```
KaiSzuttor: If you write function documentation, please use the doxygen syntax in all places.

src/core/reaction_ensemble.cpp
```diff
@@ -1,1724 +1,1970 @@
...
-		random_radius=m_current_reaction_system.cyl_radius*d_random();
-		out_pos[0]=random_radius*cos(phi);
-		out_pos[1]=random_radius*sin(phi);		
-	}
-	out_pos[0]+=m_current_reaction_system.cyl_x;
-	out_pos[1]+=m_current_reaction_system.cyl_y;
-	out_pos[2]=box_l[2]*d_random();
+void ReactionAlgorithm::
+    get_random_position_in_box_enhanced_proposal_of_small_radii(
+        double *out_pos) {
```
KaiSzuttor: don't use output variables

src/core/reaction_ensemble.cpp
```diff
@@ -1,1724 +1,1970 @@
...
-	return p_id;
+  } else {
+    get_random_position_in_box(pos_vec);
+    place_particle(p_id, pos_vec);
+    // set type
+    set_particle_type(p_id, desired_type);
+    // set velocities
+    set_particle_v(p_id, vel);
+#ifdef ELECTROSTATICS
+    //
```
KaiSzuttor: the next three functions are unrelated to the reaction_ensemble and should be moved to an appropriate utils header

KaiSzuttor: remove commented out code

KaiSzuttor: remove commented code

KaiSzuttor: after #1613 is merged, please use the histogram implementation therein

KaiSzuttor: The histogram itself should not be part of the WLRE

KaiSzuttor: don't use output variable, this function is generic and does not belong here

KaiSzuttor: this function does also not belong here and uses output variable

KaiSzuttor: please remove the printf

src/python/espressomd/reaction_ensemble.pyx
```diff
@@ -376,39 +470,31 @@ cdef class reaction_ensemble(object):
         """
         for k in kwargs:
             if k in self._valid_keys_add_collective_variable_potential_energy():
-                self._params[k]=kwargs[k]
-            else: KeyError("%s is not a valid key" %k)
-            
+                self._params[k] = kwargs[k]
+            else:
+                KeyError("%s is not a valid key" % k)
```
KaiSzuttor: you meant raise KeyError

testsuite/wang_landau_reaction_ensemble.py
```diff
@@ -0,0 +1,131 @@
...
+    #
+    system = espressomd.System()
+    system.time_step = 0.01
+    system.cell_system.skin = 0.4
+
+    #
+    # Setup System
+    #
+
+    # Interaction setup
```
KaiSzuttor: outdated comment

KaiSzuttor (changes requested):

src/core/particle_data.cpp
jonaslandsgesell: for big particle numbers we don't always want to build the map which contains the ids for each type. therefore we leave this in

src/python/espressomd/reaction_ensemble.pyx
jonaslandsgesell: k

testsuite/wang_landau_reaction_ensemble.py
jonaslandsgesell: k

src/core/reaction_ensemble.hpp
jonaslandsgesell: k

src/core/reaction_ensemble.hpp
jonaslandsgesell: k

src/core/reaction_ensemble.hpp
jonaslandsgesell: k

src/core/reaction_ensemble.hpp
jonaslandsgesell: k

src/core/reaction_ensemble.hpp
jonaslandsgesell: k

src/core/reaction_ensemble.hpp
jonaslandsgesell: k

src/core/reaction_ensemble.hpp
jonaslandsgesell: k

src/core/reaction_ensemble.cpp
jonaslandsgesell: k

src/core/reaction_ensemble.cpp
jonaslandsgesell: k

src/core/reaction_ensemble.cpp
jonaslandsgesell: k

src/core/reaction_ensemble.hpp
jonaslandsgesell: k

src/core/reaction_ensemble.cpp
jonaslandsgesell: the comment is valid and describes the logic of the replacement, deletion or creation of particles within the reaction ensemble algorithm.

src/core/reaction_ensemble.cpp
jonaslandsgesell: k

src/core/particle_data.hpp
jonaslandsgesell: replaced with direct call of get_random_pid(type)

src/core/particle_data.hpp
jonaslandsgesell: k

src/core/particle_data.hpp
jonaslandsgesell: k

src/core/particle_data.cpp
jonaslandsgesell: k

src/core/particle_data.cpp
jonaslandsgesell: k

src/core/particle_data.cpp
jonaslandsgesell: k

src/core/particle_data.cpp
jonaslandsgesell: k

src/core/particle_data.cpp
jonaslandsgesell: because it also removes the particle from the system via calling mpi_remove_particle(...)

src/core/particle_data.cpp
jonaslandsgesell: in principle yes. However in place_particle() in its current form depends on get_particle_node() returning -1 instead of throwing an exception.

samples/reaction_ensemble.py
jonaslandsgesell: k

samples/reaction_ensemble.py
jonaslandsgesell: k

samples/reaction_ensemble.py
jonaslandsgesell: this was an example for the constant pH method. Now the initialization of the constant pH ensemble or the reaction ensemble happens conditionally in the example script.

samples/reaction_ensemble.py
jonaslandsgesell: this was an example for the constant pH method. Now the initialization of the constant pH ensemble or the reaction ensemble happens conditionally in the example script.

src/core/reaction_ensemble.cpp
jonaslandsgesell: moved this comment to the reaction function

src/core/particle_data.cpp
fweik: Why are you using *alloc?

src/core/reaction_ensemble.hpp
jonaslandsgesell: modification via cython requires quite some members to be public. made more members protected.

src/core/particle_data.cpp
jonaslandsgesell: This code is gone. It s shown outdated here.

src/core/reaction_ensemble.cpp
fweik: In doxygen comments you can actually use TeX ^^

jonaslandsgesell: I am not sure how a comment like this would improve readability:
/**a(i)=product_coefficients(i)-reactant_coefficients(i)
*b(i)=product_coefficients(i)-reactant_coefficients(i)
*create a(i) many product particles if and only if a(i)>0,
* if and only if b(i)<0, hide b(i) many reactant particles */

src/core/particle_data.cpp
fweik: We decided to leave this in, as we should not introduce new scaling bottlenecks. This needs O(N) memory on the head node.

src/core/particle_data.cpp
```diff
@@ -1258,310 +1258,48 @@ void auto_exclusions(int distance) {
...
-  for (int i = 0; i < Type.max_entry; i++) {
-    if (Type.index[i] == type) {
-      l_err = 0;
-      break;
+int get_random_p_id(int type){
+    if(ParticleTypeMap.at(type).size()==0)
+        throw std::runtime_error("No particles of given type could be found");
+    int returned_p_id=0;
+    int rand_index = i_random(ParticleTypeMap.at(type).size());\
+    int actual_i_th_p_id_in_map=0;
```
fweik: What?!

src/core/particle_data.cpp
```diff
@@ -1258,310 +1258,48 @@ void auto_exclusions(int distance) {
...
-  }
-  if (l_err) {
-    return ES_ERROR;
-  }
-  int in_type = Index.type[type];
-  for (int i = 0; i < type_array[in_type].max_entry; i++) {
-    printf("%d\n", type_array[in_type].id_list[i]);
-  }
-  return ES_OK;
+    if(returned_p_id > max_seen_particle)
```
fweik: How is this possible?

src/core/particle_data.cpp
```diff
@@ -64,12 +65,9 @@
...
  ************************************************/
 // List of particles for grandcanonical simulations
-TypeOfIndex Type;
-IndexOfType Index;
-TypeList *type_array;
 int number_of_type_lists;
-int GC_init;
-int Type_array_init;
+bool GC_init;
+std::unordered_map<int, std::unordered_set<int>> ParticleTypeMap{};
```
fweik: Please adhere to the naming convention. Variable names should be lower case and not camel.

src/core/particle_data.cpp
KaiSzuttor: then SRP is violated

src/core/particle_data.cpp
```diff
@@ -1258,310 +1258,48 @@ void auto_exclusions(int distance) {
...
-  // fill in array type_index_of_type
-  for (int i = 0; i < Type.max_entry; i++) {
-    Index.type[Type.index[i]] = i;
-  }
-
-  int t_c = 0; // index
-  type_array[Index.type[type]].id_list =
-      (int *)Utils::malloc(sizeof(int) * n_part);
+  //fill particle map
+  if(ParticleTypeMap.count(type)==0)
```
fweik: This is only needed if you need the sets to exist for types that have no particles.
Otherwise you can just do
```c++
ParticleTypeMap[type].insert(id);
```

src/core/particle_data.cpp
```diff
@@ -1258,310 +1258,48 @@ void auto_exclusions(int distance) {
...
-    if (Type.index[i] == type) {
-      l_err = 0;
-      break;
+int get_random_p_id(int type){
+    if(ParticleTypeMap.at(type).size()==0)
+        throw std::runtime_error("No particles of given type could be found");
+    int returned_p_id=0;
+    int rand_index = i_random(ParticleTypeMap.at(type).size());\
+    int actual_i_th_p_id_in_map=0;
+    for( int p_id :ParticleTypeMap[type]){
```
fweik:
```c++
return *std::advance(ParticleTypeMap[type].begin(), rand_index);
```

src/core/particle_data.cpp
```diff
@@ -1258,310 +1258,48 @@ void auto_exclusions(int distance) {
...
-    return ES_OK;
-  }
-  for (int i = 0; i < Type.max_entry; i++) {
-    free(type_array[i].id_list);
-  }
-  free(type_array);
-  free(Type.index);
-  free(Index.type);
-  return ES_OK;
+void add_particle_to_list(int part_id, int type) {
```
fweik: add_id_to_type_list ?

src/core/particle_data.hpp
```diff
@@ -950,70 +950,17 @@ void try_add_exclusion(Particle *part, int part2);
...
- * typelist */
-int find_particle_type_id(int type, int *id, int *in_id);
+int get_random_p_id(int type);
 
-int remove_id_type_array(int part_id, int type);
-int add_particle_to_list(int part_id, int type);
-// print out a list of currently indexed ids
-int gc_status(int type);
-int number_of_particles_with_type(int type, int *number);
+void remove_id_from_map(int part_id, int type);
```
fweik: I think the add and remove functions need not be exported.

fweik (changes requested):

src/python/espressomd/system.pyx
```diff
@@ -442,3 +443,59 @@ cdef class System(object):
...
+        if not hasattr(type_list, "__iter__"):
+            raise ValueError("type_list has to be iterable.")
+
+        for current_type in type_list:
+            if (max_seen_particle < 0):
+                raise ValueError(
+                    "The system contains no particles. Create one particle with arbitrary type first!")
+            init_type_map(current_type)
+
+    def number_of_particles(self, current_type=None):
```
fweik: `current_type` -> `type`

src/python/espressomd/system.pyx
```diff
@@ -442,3 +443,59 @@ cdef class System(object):
...
+        -------
+        :obj:`int`
+            The number of particles which share the given type.
+
+        """
+        self.check_valid_type( current_type)
+        number=number_of_particles_with_type(current_type)
+        return int(number)
+
+    def find_particle(self, current_type=None):
```
fweik: `current_type` -> `type`

src/core/particle_data.cpp
```diff
@@ -64,12 +65,9 @@
  * variables
  ************************************************/
 // List of particles for grandcanonical simulations
-TypeOfIndex Type;
-IndexOfType Index;
-TypeList *type_array;
 int number_of_type_lists;
-int GC_init;
-int Type_array_init;
+bool GC_init;
```
fweik: Rename to `type_list_enable` or similar.

src/python/espressomd/system.pyx
```diff
@@ -442,3 +443,59 @@ cdef class System(object):
...
+        drawn. If you want Espresso to keep track of particle ids of a certain type
+        you have to initialize the method by calling the setup function. After that
+        Espresso will keep track of particle ids of that type.
+
+        """
+        if not hasattr(type_list, "__iter__"):
+            raise ValueError("type_list has to be iterable.")
+
+        for current_type in type_list:
+            if (max_seen_particle < 0):
```
fweik: Your core code can perfectly handle types w/o parts, so this check is not needed.

fweik (changes requested):

src/core/particle_data.cpp
jonaslandsgesell: yes because then the map will create a default constructed unordered set and will insert id into it. However I need to have sets existing where no particles exist (yet) in order to be able to get a particle count of zero (via calling type_set.size()). Particles might come into existance because of this zero count in the reaction ensemble.

src/core/particle_data.cpp
jonaslandsgesell: This is a way to draw a random element from an unordered set.
According to http://en.cppreference.com/w/cpp/iterator/advance std::advance needs a Bidirectional Iterator. However the unordered set only is guaranteed to have a forward iterator (http://www.cplusplus.com/reference/unordered_set/unordered_set/). Therefore I cannot use std::advance on the unordered set.

src/core/particle_data.cpp
jonaslandsgesell: Indeed impossible. This was a check for the old code where this could happen (as a bug)

src/core/particle_data.cpp
fweik: Just use `ParticleTypeMap[type]` when getting the particle count.

src/core/particle_data.cpp
fweik: No

jonaslandsgesell: True, the problem was std::advance returns void and advances the iterator itself...
    std::unordered_set<int>::iterator it=particle_type_map[type].begin();
    std::advance(it,rand_index);
    return *it;
works

fweik:
Sorry, std::next

src/core/particle_data.cpp
jonaslandsgesell: When using ParticleTypeMap[type] on an unitialized type this will result in a zero count then although the system might contain particles of the given type. In the current implementation 
int number_of_particles_with_type(int type) {
  return static_cast<int>(particle_type_map.at(type).size());
}
at least tells you that something went wrong since at checks for the type already existing in the particle_type_map. Here it is the users responsibility to initialize the set of ids for the types he is interested in.

src/core/reaction_ensemble.cpp
KaiSzuttor: This for sure should go into a function with a descriptive name, actually maybe several functions...

jonaslandsgesell: Feel free to do it

fweik (approved):

KaiSzuttor (approved):

PR #1629, opened 2017-11-14
Best-effort myconfig check

fweik:
Fixes #1557.

Description of changes:
 - Only do FindBOOST once
 - check_myconfig target
 - Implementation thereof

The implementation uses the compiler to parse the myconfig, so that
ifdefs work as expected. Because this is not supported by all compilers
the test is skipped if the parsing fails. As far as I can tell this works
on all major compilers except for MSVC. Before checking, the defined
external features are extracted from cmake_config.hpp, so that conditional
defines depending on external features (as in maxset.hpp) are correctly
evaluated.


mkuron:
Python 3 support is missing.

hmenke:
I don't think this will be ready before the merge, but recently I've been putting some efforts into replacing the `myconfig.hpp` with a CMake-based approach.  You can see the relevant changes in the comparison https://github.com/espressomd/espresso/compare/python...hmenke:features
So far it is all very centralised in the file `src/featuredefs.cmake` and features marked as `equal` are not yet handled, so it does not compile.  Nevertheless the approach reduces the current abomination of a bunch of Python scripts, scattered around the source directories, to one single point in the CMake workflow.

fweik:
@hmenke this is great, I think in the long run we should replace the header config by some cmake cnostruction, but I think for now this is an improvement an should be merged.

RudolfWeeber:
@mkuron, could you review this one?

mkuron (approved):
I checked the logic and believe it's all correct. I also verified that it works with GCC, Clang and Intel and ensured that it correctly deals with CUDA being present or not present.

I think this is safe to merge. Unlike the previous solution, it can never cause the CUDA tests to silently be skipped. If a compiler isn't supported, the check will be skipped and in all other cases an invalid myconfig is a hard error.

PR #1633, opened 2017-11-15
More scafacos related fixes

RudolfWeeber:
Fixes #

Description of changes:
 - 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


src/python/espressomd/scafacos.pyx
```diff
@@ -38,8 +45,45 @@ IF SCAFACOS == 1:
...
+                pname =p[i]
+                i+=1
+                
+                # The first item after the parameter name is alsways a value
+                # But in the case of array-like properties, there can also
+                # follow several values. Therefore, we treat the next
+                # words as part of the value, if they begin with a digit
+                pvalues=[p[i]]
+                i+=1
+                if i>=len(p):
```
KaiSzuttor, line 62: ~~this is already checked in the while condition~~

RudolfWeeber:
On Thu, Nov 16, 2017 at 07:37:53AM +0000, Kai Szuttor wrote:
> 
> 
> 
> > -                res[p[2 * i + 1]] = p[2 * i + 2]
> +            
> +            
> +            method_params={}
> +            i=1
> +            while i<len(p):
> +                pname =p[i]
> +                i+=1
> +                
> +                # The first item after the parameter name is alsways a value
> +                # But in the case of array-like properties, there can also
> +                # follow several values. Therefore, we treat the next
> +                # words as part of the value, if they begin with a digit
> +                pvalues=[p[i]]
> +                i+=1
> +                if i>=len(p):
> 
> this is already checked in the while condition
But `i` is incremented since the head of the outer while loop, as
the first parameter value is consumed.

KaiSzuttor:
You are right, i missed that

src/core/scafacos.cpp
```diff
@@ -100,6 +100,8 @@ int ScafacosData::update_particle_data() {
 
 void ScafacosData::update_particle_forces() const {
   int it = 0;
+  if (positions.size()==0)
```
fweik: `positions.empty()`

src/core/scafacos.cpp
```diff
@@ -148,6 +150,7 @@ void ScafacosData::update_particle_forces() const {
   if (!dipolar()) {
     assert(it == fields.size());
   } else {
+    int tmp=positions.size()/3;
     assert(it = positions.size() / 3);
```
fweik: not yours, but this should be `it == positions.size() / 3`.

fweik (approved):

PR #1634, opened 2017-11-16
Accumulator for observables

KaiSzuttor:
Description of changes:
 -  added an accumulator for observable to get the mean and the variance.


hmenke:
This should use `AutoParameters`.

fweik:
To fix the boost error define BOOST_RESULT_OF_USE_DECLTYPE before including the boost headers. This is a boost bug that we have hit before (see Cells.hpp).

fweik:
Looks good to me. Very terse. If you fix the boost issue, this can be merged.

src/core/accumulators/Accumulator.hpp
```diff
@@ -0,0 +1,61 @@
...
+#define BOOST_RESULT_OF_USE_DECLTYPE
+#include <boost/accumulators/accumulators.hpp>
+#include <boost/accumulators/numeric/functional/vector.hpp>
+#include <boost/accumulators/statistics/mean.hpp>
+#include <boost/accumulators/statistics/stats.hpp>
+#include <boost/accumulators/statistics/variance.hpp>
+
+#include "observables/Observable.hpp"
+
+namespace ba = boost::accumulators;
```
fweik: Please move this into the namespace, otherwise this has global visibility.

fweik (approved):

PR #1635, opened 2017-11-16
Eliminate some more warnings

mkuron:
More fixes for #1558, eliminating some warnings that @hmenke missed in #1616.


KaiSzuttor:
Since your looking at the errors... If we don't want to see errors from external libs such as boost: add the SYSTEM keyword in CMakeLists:
include_directories(SYSTEM ${Boost_INCLUDE_DIRS})
Maybe there are more includes that we should declare system...

fweik (approved):

PR #1639, opened 2017-11-17
Fetch cache

fweik:
Description of changes:
 - Cache particle access via get_particle_data


src/core/grid.hpp
```diff
@@ -346,6 +352,25 @@ inline void unfold_position(double pos[3], double vel[3], int image_box[3]) {
 #endif
 }
 
+template<typename Particle>
+Vector3d unfolded_position(const Particle * p) {
+  Vector3d pos{p->r.p};
+#ifdef LEES_EDWARDS
+  auto const y_img_count = (int)floor(pos[1] * box_l_i[1] + p->l.i[1]);
```
hmenke: Why the explicit (C-style!) cast?
```cpp
int const y_img_count = std::floor(pos[1] * box_l_i[1] + p->l.i[1]);
```

RudolfWeeber:
The PR looks good to me.

How will the size of the cache be limited?

How will this integrate with particle slicing? In that case, the cache should, in the future, have an interface for telling ahead of the individual fetches, which particles will be needed. Then the cache can decide to use PartCfg (with bond information) or fetch from several nodes at once.

fweik:
The size is not limited at the moment, I will fix that. As far as I can see the only way to implement that in an overhead free way is to drop a random element if the maximal size is reached. I'll also add an explicit put method to the cache, so that it can be filled before access.

fweik:
Ok, so I made everything a bit clearer. It's now possible to avoid a copy for the particles that are local to the master (because the access is now const), which should give considerable speedup in the single node case. The size of the cache is now limited, and it will drop a random particle if the maximal size is reached and a new value is supposed to be put. The maximal size is currently hardcoded to 100 MiB which is about 500k particles with minimal size, 120k with max_set. I think this is sufficient for now, in the long run the should be controllable by the user. 

I think the next step with regard to ParticleSlice would be to implement a method that gets multiple particles via mpi in parallel, to speed up the collecting too. I will look into that tomorrow.

fweik:
@RudolfWeeber I"ve implemented the prefetching. On a single node the cached versions are faster than the current master by about 10%, but it's kind of hard to measure these small differences against the background of the gigantic python overhead.

RudolfWeeber:
The core sie looks good.
In Python, you are only prefetching when iterating over particle slice, but not for ghe getters as in
s.part[:].pos
if I read geta() correctly.
One probably needs the same kind of loop as in ParticleSlice.__iter__()

fweik:
As far as I understand the ParticelHandle class it always returns a ParticleSlice if no specific id is given,
see ParticleList.__getitem__(self, key) (particle_data.pyx: 1638). But I am not at all sure that I understand this construction. Maybe @mkuron can comment on that.

fweik:
@RudolfWeeber now the attribute access should be prefetched too.

RudolfWeeber:
This pr can be merged

PR #1644, opened 2017-11-23
Matheval

hmenke:
### What?

This adds a mathematical expression parser to the ESPResSo core.  With this it is possible to pass a mathematical expression as a string all the way into a tight loop, like the integration loop and have it evaluated in every iteration with, say, the time step.

### Why?

This is intended to serve as a building block for constraints or LB boundaries with time-dependent parameters.  It was motivated by the Lees-Edwards boundary conditions for LB as the Lees-Edwards shift has to be adjusted every time step.

### How?

The parser uses Boost.Spirit X3 and Boost.Fusion under the hood.  The compilation time of projects using Boost.Spirit is unfortunately quite excessive, which is why I hide all those nasty details behind an opaque pointer (also known as pImpl).

---

A unit test and inline Doxygen documentation is supplied alongside.
- [x] Tests
- [x] Docs

---

Right now some of the checks are failing because the Boost version in some of the containers is too old.  The oldest Boost version with which it seems to work fine is 1.65.1 (maybe it works with an even older version, but that is the oldest version in the Docker containers for which it works).


KaiSzuttor:
Do we want to depend on a self maintained lib?

mkuron:
Considering it consists of just one header file and isn't likely to have a high update frequency we might as well copy that into Espresso.

KonradBreitsprecher:
The obvious use case would be expression-based non-bonded and bonded interactions.
Aside from putting some fatal responsibility to the user, would it be possible to pass an expression via the python interface?

hmenke:
It is very easy to pass an expression through the Python interface.  Instead of a number you simply pass the expression as a string.  If you can give me a mockup of what you would imagine expression-based interactions to look like, I could try to implement it and add it to this PR.

KonradBreitsprecher:
Something like 
```python

#Non-bonded IA

system.non_bonded_inter[0, 0].generic.set_params(
distance_var = "r", 
potential = "-g**2 * exp(-k*m*r)/r", 
gradient = "...", 
cutoff=cut)

#Same for bonded IA

genbond = GenericPairBond(
distance_var = "r", 
potential = "-g**2 * exp(-k*m*r)/r", 
gradient = "...",
r_cut=cut )
system.bonded_inter.add(genbond)

```

hmenke:
@KonradBreitsprecher 
```python
import espressomd

system = espressomd.System(box_l=[2,2,2])
system.time_step = 0.01
system.cell_system.skin = 0.1

system.part.add(pos=[0.9,1,1],type=0)
system.part.add(pos=[1.1,1,1],type=0)

# Why do I have to do this?  I probably forgot to set some switch.
system.non_bonded_inter[0,0].lennard_jones.set_params(
    epsilon=0.0, sigma=0.0, cutoff=1.0, shift=0.0)

system.non_bonded_inter[0,0].generic.set_params(
    cutoff=1.0,
    energy="4*{epsilon}*(({sigma}/r)**12 - ({sigma}/r)**6)".format(epsilon=0.1,sigma=0.1),
    force="-48*{epsilon}*(({sigma}/r)**12 - .5*({sigma}/r)**6)/r".format(epsilon=0.1,sigma=0.1))

system.integrator.run(10)
print(system.part[:].pos)
```
Without setting the Lennard-Jones interaction with non-zero cutoff, ESPResSo never enters the `add_non_bonded_pair_force` function.  I have no idea why that is the case.  Help is much appreciated.

I also need a good name for the feature guard.  How about `EXPRESSION`?

Bonded interactions require some more work.

KonradBreitsprecher:
That was quick.. I think you have to adjust the maximum cutoff in interaction_data.cpp analogous to other non-bonded Interactions.

hmenke:
@KonradBreitsprecher I have implemented the bonded interactions as well, but a testcase is missing so far.  I also chose to not implement the dihedral bond, because the implementation for tabulated dihedral bonds, from which I wanted to borrow, appeared quite obscure to me.  Also it has the following comment:
https://github.com/espressomd/espresso/blob/110ca4c202525fd301ecd79fd0858ce21004aa43/src/core/tab.hpp#L315

hmenke:
@KonradBreitsprecher For the bonded interactions there is a slight problem.  I wanted to make a testcase with the `HarmonicBond` and `AngleHarmonic` but the results do not match.  However, `Tabulated` and `GenericBond` match.  I'm not sure where the error is or whether this is just roundoff (albeit quite large for that).

### Distance

```python
import sys
import numpy as np
import espressomd
from espressomd.interactions import GenericBond, HarmonicBond, Tabulated

system = espressomd.System(box_l=[10,10,10])
system.time_step = 0.01
system.cell_system.skin = 0.1

system.part.add(pos=system.box_l/2 - [.1,0,0],type=0)
system.part.add(pos=system.box_l/2 + [.1,0,0],type=0)

params = { 'k': 1.0, 'r_0': 1.0, 'r_cut': 1.0 }

if sys.argv[1] == "harm":
    bond = HarmonicBond(**params)
elif sys.argv[1] == "expr":
    bond = GenericBond(type='distance', cutoff=params['r_cut'],
                       energy="{k}/2*(r-{r_0})**2".format(**params),
                       force="-{k}*(r-{r_0})".format(**params))
elif sys.argv[1] == "tab":
    r = np.linspace(0,params['r_cut'],10000)
    def V(r):
        return params['k']/2*(r-params['r_0'])**2
    def F(r):
        return -params['k']*(r-params['r_0'])
    bond = Tabulated(type='distance', min=0, max=params['r_cut'],
                     energy=V(r), force=F(r))
else:
    raise RuntimeError("BOOM!")

system.bonded_inter.add(bond)
system.part[0].add_bond((bond,1))

system.integrator.run(10)
print(system.part[:].pos)
```
Result
```
$ ./pypresso distance.py harm
[[ 4.8960066  5.         5.       ]
 [ 5.1039934  5.         5.       ]]
$ ./pypresso distance.py expr
[[ 4.9040066  5.         5.       ]
 [ 5.0959934  5.         5.       ]]
$ ./pypresso distance.py tab
[[ 4.9040066  5.         5.       ]
 [ 5.0959934  5.         5.       ]]
```

### Angle

```python
import sys
import numpy as np
import espressomd
from espressomd.interactions import GenericBond, AngleHarmonic, Tabulated

system = espressomd.System(box_l=[10,10,10])
system.time_step = 0.01
system.cell_system.skin = 0.1

system.part.add(pos=system.box_l/2 - [.1,0,0])
system.part.add(pos=system.box_l/2)
system.part.add(pos=system.box_l/2 + [.1,0,0])

params = { 'bend': 1.0, 'phi0': np.pi/2 }

if sys.argv[1] == "harm":
    bond = AngleHarmonic(**params)
elif sys.argv[1] == "expr":
    bond = GenericBond(type='angle', cutoff=np.pi,
                       energy="{bend}/2*(r-{phi0})**2".format(**params),
                       force="-{bend}*(r-{phi0})".format(**params))
elif sys.argv[1] == "tab":
    r = np.linspace(0,np.pi,10000)
    def V(r):
        return params['bend']/2*(r-params['phi0'])**2
    def F(r):
        return -params['bend']*(r-params['phi0'])
    bond = Tabulated(type='angle', min=0, max=np.pi,
                     energy=V(r), force=F(r))
else:
    raise RuntimeError("BOOM!")

system.bonded_inter.add(bond)
system.part[1].add_bond((bond,0,2))

system.integrator.run(10)
print(system.part[:].pos)
```
Result
```
$ ./pypresso angle.py harm
[[ 4.90000056  5.          5.        ]
 [ 5.          5.          5.        ]
 [ 5.09999944  5.          5.        ]]
$ ./pypresso angle.py expr
[[ 4.9  5.   5. ]
 [ 5.   5.   5. ]
 [ 5.1  5.   5. ]]
$ ./pypresso angle.py tab
[[ 4.9  5.   5. ]
 [ 5.   5.   5. ]
 [ 5.1  5.   5. ]]
```

hmenke:
@KaiSzuttor @mkuron There are two jobs in the CI which are repeatedly failing for no apparent reason.

GitLab CI fails in `check_sphinx` even though I didn't even touch any Sphinx related files.
```
/usr/local/lib/python2.7/dist-packages/sphinx/util/requests.py:65: UserWarning: Some links may return broken results due to being unable to check the Server Name Indication (SNI) in the returned SSL cert against the hostname in the url requested. Recommended to install "requests[security]" as a dependency or upgrade to a python version with SNI support (Python 3 and Python 2.7.9+).
  'Some links may return broken results due to being unable to '

Warning, treated as error:
autodoc: failed to import module u'espressomd.visualization_mayavi'; the following exception was raised:
No module named pyface.api
```

Travis CI fails in `autopep8` but *only* in the configuration `xcode8 myconfig=maxset image=python3`
```
maintainer/travis/build_cmake.sh: line 66: pep8: command not found
Error: Python files are not indented the right way. Please use 4 spaces per indentation level!
```

mkuron:
@KaiSzuttor and I fixed the issues you reported, @hmenke. Once the fixes are merged, you should be able to build again. Note that
```
Warning, treated as error:
docstring of espressomd.interactions.GenericNonBonded.set_params:10:Inline interpreted text or phrase reference start-string without end-string.
```
on the last Sphinx build is an actual error that you need to fix.

KaiSzuttor:
is this PR still considered for merging? if so, please make clang happy @hmenke

RudolfWeeber:
Should we discuss this at the school?

hmenke:
Since ESPResSo now uses C++14, I would even be tempted to redo this using Spirit X3 instead of Spirit Qi, which compiles a thousand times faster.

mkuron (approved):

jngrad:
@KaiSzuttor @fweik  the [interaction kernel refactoring](https://github.com/espressomd/espresso/projects/6) should remain on hold until this PR is merged.

@hmenke to facilitate review, I think it would be helpful for this PR to be structured in 2 commits: a definitive one which passes CI and contains the feature (can be anywhere in the commit history between 4.0 and 4.1), and a merge commit which can be force-pushed any time the master branch changes. This way the PR can be approved based on the first commit, and the merge commit can be maintained by core developers.

hmenke:
> This branch is 1 commit ahead of espressomd:python.

Just push any merge commits you like.

RudolfWeeber:
This could also be used for virtual sites which follow an analytically given trajectory.
The question we have to figure out, I think, is whether we copy in the code or pull it in via git submodule. 
Personally, I'd prefer the latter.

KaiSzuttor:
Why do you prefer the submodule?

RudolfWeeber:
On Mon, Oct 14, 2019 at 06:38:35AM -0700, Kai Szuttor wrote:
> Why do you prefer the submodule?
Because code placed in the Espresso repo tends to be understood to be the Espresso team's responsibility, 
and because the version history of the library stays consistent with the git repo where it is originally hosted.

hmenke:
Interesting, @RudolfWeeber suggests a submodule which it originally was, however I pulled it into the ESPResSo repo because of the earlier comment by @KaiSzuttor: https://github.com/espressomd/espresso/pull/1644#issuecomment-346543852
Maybe you two can discuss what is the preferred strategy and give me a unanimous decision.

KaiSzuttor:
```Unknown feature 'EXPRESSION'```

**Comment by RudolfWeeber** [comment]
I guarded all uses of the matheval library in Espresso, so it should build without it, in case the need arises.

I used git subtree to integrate the library, in the end, because unless we upgrade to cmake 3.11 and fetch_content, some changes to mathevalś cmake were needed.

From my side, we can go ahead with this pr. However, if someone could push better cmake in the root CMakeList.txt to detect the presence of the library, that would be helpful.

**Review by RudolfWeeber — APPROVED** [review]

File: testsuite/python/generic.py [hunk 1]
```diff
@@ -0,0 +1,143 @@
...
+# (at your option) any later version.
+#
+# ESPResSo is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import print_function
```
jngrad:
```suggestion
```
not required anymore

jngrad (changes requested):
I cannot compile this PR on maxset:
```sh
cp ../maintainer/configs/maxset.hpp myconfig.hpp
cmake ..
make
```
```
External feature 'MATHEVAL' can not be defined in myconfig.
There were errors in '/work/jgrad/espresso-fork/build/src/config/myconfig-final.hpp'
```

fweik:
@jngrad you potentially need #3446 for this to work.

KaiSzuttor:
ping

KaiSzuttor:
pong

RudolfWeeber:
After a renewed discussion in the core team, we have come to the conclusion that we are not able to maintain the code introduced by this PR.
Therefore, we are, unfortunately, not going to merge this PR.

I'd like to emphasize that this is not a matter of code quality.
However, merging this would imply keeping some 800 lines of math parsing code and a dependency on the boost spirit library working for the forseeable future.
Given the rate at which OS, library, and compiler versions evolve, this is a promise we don't feel we can make for code which does not directly relate to Espresso's core purpose.

The functionality provided by this Pr is of interest but a very similar result can be obtained using tabulated potentials.

hmenke:
> we are not able to maintain the code introduced by this PR

Of course not.

Have a good day.

PR #1646, opened 2017-11-23
Resort flag

### LINKED ISSUE #1637: interactions_non-bonded.py broken with 3 cores
with 3 cores. To reproduce:
```bash
(set -e; while true; do mpirun -np 3 ./pypresso ../testsuite/interactions_non-bonded.py; done)
```
The repetition is only needed because the testcase picks a random starting point. If the interaction crosses a periodic boundary in the direction of the node grid it always fails. The reason seems to be
that the positions in the ghost cell that takes part in the interaction are not shifted as they should be.
This leads to a wrong distance and the pair is discarded, resulting in a missing interaction. This is actually pretty severe, as it means that Espresso can not do MD correctly with this node layout.

fweik:
Fixed #1637.

Description of changes:
 - Introduce different levels for resort_particles.
 - Do global resort if the user has moved particles.

### DIFF
--- a/src/core/cells.cpp
+++ b/src/core/cells.cpp
@@ -61,6 +61,8 @@ CellStructure cell_structure = {/* type */ CELL_STRUCTURE_NONEYET,
 
 double max_range = 0.0;
 
+/** On of Cells::Resort, annouces the level of resort needed.
+ */
 unsigned resort_particles = Cells::RESORT_NONE;
 int rebuild_verletlist = 1;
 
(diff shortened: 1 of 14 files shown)


fweik:
Ready for review.

RudolfWeeber:
Can we prevent write acces to resort particles, so no-one reduces the resort level by accident by using
= instad of |=?
i.e., have functions
void need_local_resort()
void need_global_resort()
or the like?

Apart from that:

announce_resort_particles():
  please rename `sum` which no longer is a sum

cells_update_ghosts():
  Why is the behavior the same, if LEES_EDWARDS is compiled in, eventhough the ifdef block was removed?

fweik:
> Can we prevent write acces to resort particles, so no-one reduces the resort level by accident by using
> = instad of |=?
> i.e., have functions
> void need_local_resort()
> void need_global_resort()
> or the like?

Done. There is now a setter for resort_particles that can only increase the level.

> announce_resort_particles():
> please rename sum which no longer is a sum

Done.

> cells_update_ghosts():
> Why is the behavior the same, if LEES_EDWARDS is compiled in, eventhough the ifdef block was 
> removed?

Because the proper resort level is now set in integrate_vv.

RudolfWeeber:
Can be merged

PR #1647, opened 2017-11-23
Icc fix

fweik:
Description of changes:
 - Moved icc ghost comm into right place
 - Don't trigger a resort if only particle charges have changed.
 - Removed unneeded initialization pass form icc force calc.


KonradBreitsprecher:
I think I got the problem:
`add_non_bonded_pair_force_iccp3m()` won't initialize the forces of 'isolated' particles that don't appear in the loop. Putting back in the force reset for particles and ghosts fixes it. Not sure if there's a way out of doing so.
Apart from that, the speedup for a plate cap test system (with the quick init forces fix) with 100 ions and 200 icc particles is definetly **>2** (in ns/day)!

fweik:
Yes, the initializations have to go back in, is just realized that. Will
do, then it should be correct again.

fweik:
@KonradBreitsprecher should work now.

fweik:
@KonradBreitsprecher

KonradBreitsprecher (changes requested):
We should throw an error if convergence is not reached after the given max iterations.

KonradBreitsprecher (approved):

PR #1651, opened 2017-11-24
Eliminate compiler warnings

### LINKED ISSUE #1558: Eliminate all compiler warnings before releasing Espresso 4.0

mkuron:
Fixes #1558


hmenke (commented):
Please don't use C-style casts.

mkuron:
The only problems that are currently left are:
- GCC <4.9's `-Wno-pedantic` is broken, so we can't disable the pedantry on the Cython-generated code: https://travis-ci.org/mkuron/espresso/jobs/306767862
- MPI headers are not correctly treated as system headers, so warnings there are not ignored: https://travis-ci.org/mkuron/espresso/jobs/306767860
- Boost headers are not correctly treated as system headers, so warnings there are not ignored: https://travis-ci.org/mkuron/espresso/jobs/306767858

These need to be fixed on the CMake level. @hmenke, if you're bored over the weekend, you can try to fix these. Otherwise I'll do it next week.

Also, `-Wsign-compare` is not activated because this causes a huge number of extra errors. I might turn it on in a future pull request.

hmenke:
The solutions to those problems are:

1. Do not enable all warnings in `CMAKE_CXX_FLAGS` but do it per target using `set_source_file_properties`.

2. The attribute `SYSTEM` has to be prefixed, i.e. `include_directories(SYSTEM ${MPI_INCLUDE_PATH})`.  Then it works.

3. This error is *not* from Boost but from here: https://github.com/espressomd/espresso/blob/e6695586cd05cdf6e4518023a475ebb5d21435ad/src/core/utils/mpi/all_compare.hpp#L41-L57
`root_value` is indeed used possibly uninitialized.

mkuron:
Ah, thanks. Will update the pull request accordingly.

mkuron:
All green. @fweik, can we get this merged soon, to avoid conflicts?

@hmenke, I decided to disable pedantic on older GCCs because we now have so many subdirectories under src/core that there is no central place where it could be enabled per-file.

fweik (approved):

PR #1653, opened 2017-11-24
observables: made the observables stateless.

KaiSzuttor:
Removed data member last_value from observables. The observables call operator now performs the calculation.


src/core/observables/Observable.hpp
```diff
@@ -33,22 +33,15 @@ class Observable {
   friend class CylindricalFluxDensityProfile;
   Observable();
   virtual ~Observable() = default;
-  int calculate();
-
+  // In the call operator the calculation is performed.
+  virtual std::vector<double> operator()(PartCfg &partCfg) = 0;
```
fweik: This should be const.

fweik (changes requested):

fweik (approved):

PR #1608, opened 2017-11-13
Le testcase

bindgens1:
Fixes #1602.

PR #1658, opened 2017-11-24
LE CI

### LINKED ISSUE #1657: Lees Edwards not testet in CI
There are now tests...

fweik:
Fixes #1657. Closes #1608.

Description of changes:
 - Added a testing stage with Lees-Edwards to gitlab-ci.

There is still a lot to do from the looks of it. But the LE test works.


fweik:
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/5197

fweik:
The best course of action is probably to make LE into it's own cell system, so that it only has an effect if activated.

PR #1667, opened 2017-11-28
Docfix

### LINKED ISSUE #1649: correlator docs incomplete
passing of args to fcs_acf needs to be explained

possible values of compress1 and compress2 are not quite clear.

I can fix this on Monday November 27

kosovan:
Fixes #1649 

Description of changes:
 - added description of `correlator args`, modified description of the compression functions.
 - added docs for `complex_conjugate_product`, and removed redundant variable from the cpp code of this correlation operation

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


fweik (approved):
Do you hapoen to know what the complex conjugate is for? As far as I know there are only real observables in the code.

kosovan:
I think that when Stefan first implemented the `componentwise_product` he called it `complex_conjugate_product.` Then he realized that it was not the right name, so he renamed the function and implemented true `complex_conjugate_product` although there was no such observable.

kosovan:
@jonaslandsgesell , please review and comment on my changes. I decided to remove the section about the water autoprotolysis because it is still an open debate about how it should be done. In such case it is not a good idea to put it in the user guide.

doc/sphinx/run.rst
```diff
@@ -171,157 +171,219 @@ The equilibrium constant of the reaction is then given as
...
+at which the standard chemical potential :math:`\Delta_{\mathrm{r}}G^{\ominus}` was determined.
+In practice, this constant is often used with the dimension of :math:`(c^{\ominus})^{\bar\nu}`
+
+.. math::
+
+   K_c(c^{\ominus}) = K(c^{\ominus})\times (c^{\ominus})^{\bar\nu}
+
+A simulation in
+the reaction ensemble consists of two types of moves: the *reaction move*
+and the *configuration move*. The configuration move is performed by a
```
jonaslandsgesell: I would call it a "configuration changing move"

doc/sphinx/run.rst
```diff
@@ -171,157 +171,219 @@ The equilibrium constant of the reaction is then given as
...
 suitable molecular dynamics or a Monte Carlo scheme. The
 ``reacton_ensemble`` command of takes care only of the reaction moves.
```
jonaslandsgesell: of too much

doc/sphinx/run.rst
```diff
@@ -171,157 +171,219 @@ The equilibrium constant of the reaction is then given as
...
+products is inserted into the system. In the *backward* reaction,
 reactants and products exchange their roles. The acceptance probability
 :math:`P^{\xi}` for move from state :math:`o` to :math:`n` reaction
-ensemble is given by the criterion ::cite:`smith94a`
+ensemble is given by the criterion :cite:`smith94a`
 
 .. math::
 
-   P^{\xi} = \text{min}\biggl(1,V^{\bar\nu\xi}\Gamma^{\xi}e^{-\beta\Delta E_\mathrm{pot}}\prod_{i=1}\frac{N_i^0!}{(N_i^0+\nu_{i}\xi)!}
+   P^{\xi} = \text{min}\biggl(1,V^{\bar\nu\xi}\Gamma^{\xi}e^{-\beta\Delta E}\prod_{i=1}\frac{N_i^0!}{(N_i^0+\nu_{i}\xi)!}
```
jonaslandsgesell, line 219: I would rather leave \Delta E_\mathrm{pot] here. It makes the equation self explanatory without further needing to read the description of the symbols.

doc/sphinx/run.rst
```diff
@@ -171,157 +171,219 @@ The equilibrium constant of the reaction is then given as
...
 
-   \Gamma = K \biggl(\frac{p^{\ominus}}{k_\mathrm{B}T}\biggr)^{\bar\nu},
-         \label{eq:Gamma}
-
-where :math:`p^{\ominus}=1 atm` is the standard pressure. It is often
-convenient and in some cases even necessary (dissociation reaction of
-polyelectrolytes) that some particles representing reactants are not
-removed from or placed at randomly in the system (think about reacting monomers in a polymer), but their identity is changed to that of the
-products (and vice versa in the reverse direction). The replacement rule is that for any given reactant type it is replaced by the corresponding product type (corresponding means here in terms of order in the reaction equation that was provided) as long as the corresponding coefficients allow it.
+   \Gamma = \prod_i \Bigl(\frac{N_i}{V} \Bigr)^{\bar\nu} = V^{-\bar\nu} \prod_i N_i^{\bar\nu_i}
```
jonaslandsgesell: I would rather state \Gamma=K_c(c^{\ominus})

doc/sphinx/run.rst
```diff
@@ -171,157 +171,219 @@ The equilibrium constant of the reaction is then given as
...
        \label{eq:Pacc}
        \biggr),
 
-where :math:`\Delta E_\mathrm{pot}=E_\mathrm{pot new}-E_\mathrm{pot old}` is the interaction energy change,
-:math:`\beta=1/k_\mathrm{B}T`, :math:`V` is the simulation box volume,
-:math:`\bar\nu = \sum_i
-\nu_i`. The extent of reaction, :math:`\xi=1` for the forward, and
-:math:`\xi=-1` for the reverse direction. The parameter :math:`\Gamma`
-is related to :math:`K` as
+where :math:`\Delta E=E_\mathrm{new}-E_\mathrm{old}` is the change in potential energy,
```
jonaslandsgesell, line 223: E_pot,new and E_pot,old here please. I had to do the derivation of the reaction ensemble to be convinced which energy goes here

doc/sphinx/run.rst
```diff
@@ -171,157 +171,219 @@ The equilibrium constant of the reaction is then given as
...
-derivation compared to the Reaction Ensemble. The constant pH method can
-be used by initializing the reactions of interest with the commands
-mentioned for the reaction ensemble. However with the difference that
+serious artifacts. 
+The constant pH method can be used within the reaction ensemble module by
+initializing the reactions with the standard commands of the reaction ensemble. 
+
+However with the difference that
 you do not provide the dimensionless reaction constant but directly the
 *apparent reaction constant* (from the law of mass action) :math:`K_a`
```
jonaslandsgesell: can you pleas write K_c instead of K_a, after you introduced it above.

jonaslandsgesell (commented):
So far this looks good.

doc/sphinx/run.rst
kosovan: I see no reason for giving it a longer name. But I will modify the subsequent explanation to make it clearer.

doc/sphinx/run.rst
kosovan: In my opinion, this is a common knowledge in statistical mechanics. Momenta can be integrated out analytically to yield the translational partition function, and only the potential energy contributes to the configuration integral. That is why MC or RE do not work with momenta at all. Literature which deals with Monte Carlo or Reaction ensemble uses Delta E without the subscript "pot", and I suggest to conform to the established notation.

doc/sphinx/run.rst
kosovan: I forgot to change this one. However, this note is a consequence of the previous misunderstanding. Now, when we both agree that RE uses the dimensional equilibrium constant, this note should be dropped. But we need to define the reaction which is implicitly used in the constant pH method.

doc/sphinx/run.rst
kosovan: In addition, the calculation of Delta E is implemented in the core, so there is no risk that the user does something wrong even if he/she misunderstands our description.

testsuite/constant_pH.py
```diff
@@ -88,7 +88,8 @@ def test_ideal_titration_curve(self):
         RE = ReactionEnsembleTest.RE
         """ chemical warmup in order to get to chemical equilibrium before starting to calculate the observable "degree of association" """
         for i in range(40 * N0):
-            r = RE.reaction_constant_pH()
+            #r = RE.reaction_constant_pH()
+            r = RE.reaction(3)
```
jonaslandsgesell: In the present branch this won't call the constant pH method. Only in the upcoming branch jonaslandsgesell:test_wang_landau_reaction_ensemble it will call the constant_pH method (iff you created a constant ph object)

doc/sphinx/run.rst
```diff
@@ -228,7 +229,7 @@ The parameter :math:`\Gamma` proportional to the reaction constant. It is define
 
 .. math::
 
-   \Gamma = \prod_i \Bigl(\frac{N_i}{V} \Bigr)^{\bar\nu} = V^{-\bar\nu} \prod_i N_i^{\bar\nu_i}
```
jonaslandsgesell: Maybe we add an explanation for N_i: N_i is here the number of particles of type i *after* the system has equilibrated. Otherwise one might confuse it with the current number of particles which are in the simulation box.

jonaslandsgesell (approved):

PR #1672, opened 2017-11-29
Creteria-based particle selection

RudolfWeeber:
system.part.select(q=0,v=(1.,0.,0.))
system.part.select(lambda p: p.pos[0]<0.5)
Returns an instance of ParticleSlice.


fweik (approved):

KonradBreitsprecher:
np.testing.assert acutally does a |= on the input, which fails due to the locked array.
Just change in testsuite/particle.py: 

l148: 
res = np.copy(self.system.part[1].vs_relative)
l180: 
np.testing.assert_allclose((0.2, 0.3, 0.4), np.copy(p.pos), atol=1E-12)

PR #1676, opened 2017-12-01
Enabled range checking on STL containers in Debug mode

### LINKED ISSUE #1671: Incude _GLIBCXX_DEBUG flag in default debug flags for CXX compiler

mkuron:
Also remove use of a CMake variable that didn't exist before 3.6.

Note that CMake doesn't detect which standard library is used, so we can't just define the applicable flag. And just because libc++ is standard on AppleClang and libstdc++ is standard on GNU doesn't mean that these are the only combinations that can be used -- both both libc++ and libstdc++ can be used with Clang and GCC.

Fixes #1671


CMakeLists.txt
```diff
@@ -321,11 +321,15 @@ if(WARNINGS_ARE_ERRORS)
   endif()
 endif(WARNINGS_ARE_ERRORS)
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wno-sign-compare -Wno-unused-variable -Wno-unused-function -Wno-unused-private-field")
-if (NOT CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_VERSION VERSION_GREATER "4.8.5")
+if (NOT CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR CMAKE_CXX_COMPILER_VERSION VERSION_GREATER "4.8.5")
   # older versions don't support -Wno-pedantic which we need in src/python
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pedantic")
 endif()
 
+if (${CMAKE_BUILD_TYPE} STREQUAL "Debug")
```
fweik: I think you can do just
```cmake
set(CMAKE_CXX_FLAGS_Debug "${CMAKE_CXX_FLAGS_Debug} -D_GLIBCXX_DEBUG -D_LIBCPP_DEBUG")
```

mkuron: Thanks, done.

fweik (approved):

PR #1679, opened 2017-12-04
coverage: ignoring testsuite folders.

KaiSzuttor:


fweik (approved):

fweik:
Are you sure that this works? The unit test files seem to be still in the coverage report:
https://codecov.io/gh/espressomd/espresso/pull/1679/tree/src/core/unit_tests

KaiSzuttor:
The reason seems to be a bug of codecov: see https://github.com/codecov/support/issues/426

fweik:
The benefit of having the unit tests in the coverage is that it can be easily seen if they are executed completely. I tend to keep them in.

RudolfWeeber:
I'd also keep them in. In the end, test coverage reporting serves to help us identify holes, which this does.

PR #1680, opened 2017-12-04
tcl: removing left tcl files.

KaiSzuttor:


hmenke (commented):
The cup logo should be ported to Python before removal.

fweik:
#1684.

PR #1683, opened 2017-12-04
Clang Static Analysis on Gitlab CI

### LINKED ISSUE #1681: Add static analysis to CI
This should be done on gitlab, since this takes some time. Cmake has to be run first, to get the compile commands, so we can split the jobs in multiple parts, forwarding the buid dir as artifact between stages for each config, starting with cmake, then static anlysis, build and finaly running the tests. This would also give more detailed and potentially quicker feedback in case of failure. Tool-wise I'd use clang-tidy for this, as this seems to have the most momentum at the moment and has so MPI related analysis in newer releases.

mkuron:
Fixes #1681


maintainer/gh_post_status.sh
```diff
@@ -8,4 +8,4 @@ STATUS="$1"
 curl "https://api.github.com/repos/espressomd/espresso/statuses/$GIT_COMMIT?access_token=$GITHUB_TOKEN" \
      -H "Content-Type: application/json" \
      -X POST \
-     -d "{\"state\": \"$STATUS\", \"context\": \"ICP CUDA build\", \"target_url\": \"$URL\"}"
+     -d "{\"state\": \"$STATUS\", \"context\": \"ICP GitLab CI\", \"target_url\": \"$URL\"}"
```
KaiSzuttor, line 11: Did you also change the status message set by the php script? If not, we should do that before...
My opinion... we have bigger fish....

mkuron: I did not change it yet. ~Is that /web/kai/status_pending.sh?~ /web/espresso_gitlab/status_pending.sh

We also have bigger fix to fry than fixing typos, yet we still do that. If "ICP CUDA build" continues becoming more general CI and less CUDA, eventually we'll have to change it, so why not do it now.

hmenke (approved):

.gitlab-ci.yml
```diff
@@ -26,6 +26,13 @@ maxset:
     paths:
     - build/
 
+clang-static:
+  stage: build
+  image: espressomd/espresso-ubuntu-clang-cuda:latest
+  script:
+    - export myconfig=maxset with_coverage=false with_static_analysis=true
+    - bash maintainer/cuda_build.sh
```
hmenke, line 41: We need to save the log as an artefact.  Otherwise we cannot access the result of the static analysis which makes it a bit pointless.

mkuron: Actually we can’t do that. If errors are found, the build fails, meaning we can’t upload artifacts. If there are no errors, there is nothing to upload.

mkuron:
By the way, any suggestions how I could test this? I couldn’t come up with any bug that doesn’t cause a compiler warning but is caught by the analyzer.

mkuron:
Ah, turns out analysis wasn't even working. It's fixed now and here is a failed build, complete with artifacts (thanks to `on_failure`): https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/5984. Great, now I need to fix hundreds of warnings, which could easily take me another whole coding day...

hmenke:
Perhaps you shouldn't set it to `-Werror` immediately and suppress the `Value stored to '...' is never read` warning.

mkuron:
> Perhaps you shouldn't set it to -Werror immediately

That is a compiler flag. The analyzer is independent of that.

> suppress the `Value stored to '...' is never read warning.`

Clang Analyzer can't suppress individual warnings: http://clang-analyzer.llvm.org/faq.html#suppress_issue . So we need to fix them all before we can start doing automated static analysis in CI.

mkuron:
> Perhaps you shouldn't set it to `-Werror` immediately

That is a compiler flag. The analyzer is independent of that. Its equivalent is `--status-bugs`, but turning that off makes it kind of pointless.

> suppress the `Value stored to '...' is never read warning.`

Clang Analyzer can't suppress individual warnings: http://clang-analyzer.llvm.org/faq.html#suppress_issue . So we need to fix them all before we can start doing automated static analysis in CI.

KaiSzuttor:
I suggest dividing the work among espresso people

RudolfWeeber:
Looking at the output, I don't think most of that needs to be fixed before the release.
In particular, the nullptr derefs would probably cause visible erros (segfaults) if they were actually a problem. 
A lot of it should go away anyway with the new particle interface and c-syle heap arrays being replaced by c++ containers.

RudolfWeeber:
I filtered the list accordingly. Here is what's left

Probably will go away via Flo's upcoming pr.
/builds/espressomd/espresso/src/core/correlators/Correlator.cpp:266:16: warning: Potential memory leak
/builds/espressomd/espresso/src/core/correlators/Correlator.cpp:318:31: warning: Result of 'malloc' is converted to a pointer of type 'unsigned int', which is incompatible with sizeof operand type 'int'


/builds/espressomd/espresso/src/core/io/writer/h5md/h5md_core.cpp:530:13: warning: Assigned value is garbage or undefined

Upcoming external potential refactor:
/builds/espressomd/espresso/src/core/external_potential.cpp:176:19: warning: Potential leak of memory pointed to by 'line'


/builds/espressomd/espresso/src/core/gb.hpp:128:17: warning: The left expression of the compound assignment is an uninitialized value. The computed value will also be garbage

/usr/include/boost/mpi/collectives/all_reduce.hpp:55:5: warning: Address of stack memory associated with local variable 'op' is still referred to by the global variable 'op_ptr' upon returning to the caller.  This will be a dangling reference

/usr/include/boost/mpi/collectives/all_reduce.hpp:55:5: warning: Address of stack memory associated with local variable 'op' is still referred to by the global variable 'op_ptr' upon returning to the caller.  This will be a dangling reference

/builds/espressomd/espresso/src/core/lbboundaries.cpp:367:12: warning: Potential leak of memory pointed to by 'forces'

/builds/espressomd/espresso/src/core/lb.cpp:1274:29: warning: The left operand of '*' is a garbage value
/builds/espressomd/espresso/src/core/lb.cpp:2582:32: warning: The left operand of '-' is a garbage value
/builds/espressomd/espresso/src/core/lb.cpp:2582:32: warning: The right operand of '-' is a garbage value

/builds/espressomd/espresso/src/core/magnetic_non_p3m_methods.cpp:388:40: warning: The right operand of '*' is a garbage value

/builds/espressomd/espresso/src/core/mdlc_correction.cpp:222:31: warning: The left operand of '*' is a garbage value
/builds/espressomd/espresso/src/core/mdlc_correction.cpp:233:20: warning: The left expression of the compound assignment is an uninitialized value. The computed value will also be garbage
/builds/espressomd/espresso/src/core/mdlc_correction.cpp:274:22: warning: The right operand of '*' is a garbage value
/builds/espressomd/espresso/src/core/mdlc_correction.cpp:291:11: warning: The left expression of the compound assignment is an uninitialized value. The computed value will also be garbage
/builds/espressomd/espresso/src/core/mdlc_correction.cpp:494:38: warning: The right operand of '*' is a garbage value

/builds/espressomd/espresso/src/core/molforces.cpp:290:19: warning: The right operand of '<' is a garbage value
/builds/espressomd/espresso/src/core/molforces.cpp:322:19: warning: The right operand of '<' is a garbage value

/builds/espressomd/espresso/src/core/p3m-dipolar.cpp:1209:10: warning: Assigned value is garbage or undefined
/builds/espressomd/espresso/src/core/p3m-dipolar.cpp:1435:24: warning: Division by zero
/builds/espressomd/espresso/src/core/p3m-dipolar.cpp:1522:24: warning: Division by zero

/builds/espressomd/espresso/src/core/particle_data.cpp:1077:17: warning: The right operand of '<' is a garbage value
/builds/espressomd/espresso/src/core/particle_data.cpp:1138:21: warning: The right operand of '<' is a garbage value
/builds/espressomd/espresso/src/core/particle_data.cpp:1163:19: warning: The right operand of '<' is a garbage value


/builds/espressomd/espresso/src/core/polymer.cpp:421:3: warning: Potential leak of memory pointed to by 'bond'
/builds/espressomd/espresso/src/core/polymer.cpp:741:41: warning: The left operand of '==' is a garbage value
/builds/espressomd/espresso/src/core/polymer.cpp:799:17: warning: The left operand of '>' is a garbage value
/builds/espressomd/espresso/src/core/polymer.cpp:890:14: warning: Potential leak of memory pointed to by 'bond'
/builds/espressomd/espresso/src/core/polymer.cpp:890:14: warning: Potential leak of memory pointed to by 'bonds'
/builds/espressomd/espresso/src/core/polymer.cpp:890:14: warning: Potential leak of memory pointed to by 'cross'
/builds/espressomd/espresso/src/core/polymer.cpp:890:14: warning: Potential leak of memory pointed to by 'link'
/builds/espressomd/espresso/src/core/polymer.cpp:890:14: warning: Potential leak of memory pointed to by 'links'

/builds/espressomd/espresso/src/core/lj.hpp:80:16: warning: The left expression of the compound assignment is an uninitialized value. The computed value will also be garbage

/builds/espressomd/espresso/src/core/pressure.cpp:771:153: warning: Potential leak of memory pointed to by 'occupiedybins'
/builds/espressomd/espresso/src/core/pressure.cpp:771:153: warning: Potential leak of memory pointed to by 'occupiedzbins'
/builds/espressomd/espresso/src/core/pressure.cpp:771:153: warning: Potential leak of memory pointed to by 'starty'
/builds/espressomd/espresso/src/core/pressure.cpp:771:153: warning: Potential leak of memory pointed to by 'startz'
/builds/espressomd/espresso/src/core/pressure.cpp:986:16: warning: Function call argument is an uninitialized value

/builds/espressomd/espresso/src/core/reaction_ensemble.cpp:65:49: warning: Call to 'malloc' has an allocation size of 0 bytes
/builds/espressomd/espresso/src/core/reaction_ensemble.cpp:77:48: warning: Call to 'malloc' has an allocation size of 0 bytes
/builds/espressomd/espresso/src/core/reaction_ensemble.cpp:1062:2: warning: Assigned value is garbage or undefined
/builds/espressomd/espresso/src/core/reaction_ensemble.cpp:1064:13: warning: The left operand of '<' is a garbage value

/builds/espressomd/espresso/src/core/lb.hpp:328:10: warning: Dereference of null pointer (loaded from variable 'rho')
/builds/espressomd/espresso/src/core/lb.hpp:334:8: warning: Dereference of null pointer (loaded from variable 'rho')

/builds/espressomd/espresso/src/core/statistics_fluid.cpp:212:7: warning: Use of memory after it is freed
/builds/espressomd/espresso/src/core/statistics_fluid.cpp:216:7: warning: Potential leak of memory pointed to by 'result'
/builds/espressomd/espresso/src/core/statistics_fluid.cpp:311:7: warning: Use of memory after it is freed
/builds/espressomd/espresso/src/core/statistics_fluid.cpp:316:7: warning: Potential leak of memory pointed to by 'result'

/builds/espressomd/espresso/src/core/statistics_chain.cpp:720:20: warning: The right operand of '<' is a garbage value
/builds/espressomd/espresso/src/core/statistics_chain.cpp:730:17: warning: The left expression of the compound assignment is an uninitialized value. The computed value will also be garbage
/builds/espressomd/espresso/src/core/statistics_chain.cpp:734:12: warning: Assigned value is garbage or undefined
/builds/espressomd/espresso/src/core/statistics_chain.cpp:741:14: warning: Potential leak of memory pointed to by 'min_d'
/builds/espressomd/espresso/src/core/statistics_chain.cpp:751:12: warning: Potential leak of memory pointed to by 'cm'

/builds/espressomd/espresso/src/core/statistics.cpp:707:26: warning: The left operand of '!=' is a garbage value

/usr/include/boost/variant/detail/forced_return.hpp:42:12: warning: Called function pointer is null (null dereference)
/usr/include/boost/variant/detail/forced_return.hpp:42:12: warning: Called function pointer is null (null dereference)

/builds/espressomd/espresso/build/src/python/espressomd/utils.cpp:2027:94: warning: Potential leak of memory pointed to by '__pyx_v_il'

/usr/include/boost/variant/detail/forced_return.hpp:42:12: warning: Called function pointer is null (null dereference)

/builds/espressomd/espresso/build/src/python/espressomd/grand_canonical.cpp:2589:33: warning: The left operand of '==' is a garbage value

/builds/espressomd/espresso/build/src/python/espressomd/analyze.cpp:4088:135: warning: Potential leak of memory pointed to by '__pyx_v_il'

RudolfWeeber:
I followed some of the warnings.

E.g.,
/builds/espressomd/espresso/src/core/mdlc_correction.cpp:222:31: warning: The left operand of '*' is a garbage value
This seems to be caused by the analyzer not being able to tell that the number of particles doesn't change within the function. The c-style array seems to be filled by a range-based loop over local_cells.particles() and used in a similar loop.
Replacing the c-style array by a vector (which initializes values at instancing) would hide the error message but not really affect the reliability of the code.

Probably in many cases, the only real solution would be considerable refactoring.
E.g. in the case of gb.hpp:128. The analyzer might understand the value to be garbage because pointers are passed in but not checked (as they are known to be valid in the calling funciton). Or it might be that one of the variables iss really not initialized.
One can get rid of that possibility by declaring the variables at the time of assignment.

fweik:
Cython generated code should be excluded.

mkuron:
It should be excluded, but there is no way to do that with `scan-build`. We would need to write a parser for the HTML output, discard the reports for Cython-generated files and extract the success status from the remaining ones.

fweik:
New versions of cmake have per-target clang-tidy support. clang-tidy would have the added benefit that checks can be enabled and disabled individually.

mkuron:
Ah, I wasn‘t aware of clang-tidy. LLVM just has way too many tools nowadays... That would certainly be a better solution.

mkuron:
I switched to clang-tidy now, though for some reason`-warnings-as-errors=*` is being ignored. So for now there are only warnings and I still need to silence a significant number of them.

mkuron:
Turns out Clang-Tidy is also unsuitable for automated invocation. You can't silence warnings coming from specific header files (especially Boost::MPI). I already fixed a significant fraction of the errors found by clang-tidy, so this PR is worth merging nonetheless.

I disabled all the style checks and the nullptr checks because Espresso fails them in so many places that I can't fix them all in reasonable time. All the undefined behavior checks should still be on though.

maintainer/gh_post_status.sh
mkuron: I have now modified the script on the server to automatically pull the context name from this file in the pull request.

hmenke (approved):

mkuron:
After some more thought, I was able to fix all warnings with only minor patching of boost. One was an actual bug (https://github.com/boostorg/mpi/pull/52), while the other two are because Clang-Tidy is a bit too sensitive. I've fixed all three with a simple patch: https://github.com/espressomd/docker/pull/15.

Once https://github.com/espressomd/docker/pull/14 and https://github.com/espressomd/docker/pull/15 are merged, we can trigger the CI build again and then this pull request can be merged.

src/core/readpdb.cpp
```diff
@@ -41,7 +41,7 @@ static void add_lj_internal(std::set<PdbParser::itp_atomtype, PdbParser::itp_ato
 	continue;
       if(only_diagonal && (it->espresso_id != jt->espresso_id))
 	continue;
-      const double epsilon_ij = sqrt(it->epsilon * jt->epsilon);
+      const double epsilon_ij = sqrtf(it->epsilon * jt->epsilon);
```
hmenke, line 44: `std::sqrt`?

mkuron: Same thing. I just applied the fixes that Clang suggested.

fweik (approved):

mkuron:
Merge conflicts in the reaction ensemble resolved (thanks @jonaslandsgesell for the refactoring in #1626, which has significantly improved its code quality!) and h5xx updated (this required a modification to the submodule code so that it updates the submodule if it was already cloned).

mkuron:
Please merge, @fweik. CI is green now.

jonaslandsgesell:
Where can I see the results of the static analysis?

KaiSzuttor:
in the artifacts of the gitlab ci pipeline

mkuron:
No, it's not an artifact anymore. The results look like compiler warnings, so you can see them when you click on the CI job to see the log. If there are no warnings, then everything is fine.
  

See https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/7533 for an example:
```
/builds/espressomd/espresso/src/core/unit_tests/Variant_test.cpp:197:24: error: the parameter #1 is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param,-warnings-as-errors]
    int mem(std::string) { return 12; }
                       ^
            const      &
```

jonaslandsgesell:
Ahh cool! Thank you

PR #1685, opened 2017-12-05
Collision detection / dynamic bonding support

### LINKED ISSUE #942: Python interface for dynamic bonding

### LINKED ISSUE #1077: Collision detection possibly wrong
The three particle bonding code for cellsystem domain_decomposition in collision.cpp (function three_particle_binding_domain_decomposition) in the inner-most loop looks as follows:
```
if (P->l.ghost) {
  continue;
}
coldet_do_three_particle_bond(P, p1, p2);

if (p1->l.ghost) {
  continue;
}
coldet_do_three_particle_bond(p1, P, p2);

if (p2->l.ghost) {
  continue;
}
coldet_do_three_particle_bond(p2, P, p1);
```
While the first `continue` might make sense, the second and third seem wrong since the formed bonds depend on the ordering of the particles. This code should be reviewed.

RudolfWeeber:
Fixes #942 
Fixes #1077 

This provides a Python interface to the collision detection
Also, all collision modes now work in parallel

Notes:
* The interface is based on the script interface framework. Some extra code is needed to provide type conversion (e.g., bond types) and to make sure, a complete set of parameters for a certain collision mode is passed at once.

* The use of the script interface makes the collision detection internal parameter broadcast obsolete, but leads to some changes in the core. In particular, in the validation method.

* I also replaced the c-stype malloc'd arrays by std::vector 

* Loops over the index in the collision que are mostly replace dby raged loops

* the detect_collision() method was inlined as it runs in the hot loop

* The point of collision is no longer calculated in the hot loop to reduce the size of the code executed there

@hirschsn, since you now the collision detection well and it affects your project, would you be willing to look at this?


src/python/espressomd/collision_detection.py
```diff
@@ -0,0 +1,181 @@
...
+
+    # Override to call validat after parameter update
+    def set_params(self, **kwargs):
+        """Set the parameters for the collision detection
+           
+           See :ref:`Creating bonds when particles collide` for detailed instructions.
+
+           Parameters
+           ----------
+           mode: One of "off", "bind_centers", "bind_at_point_of_collision", "bind_three_particles", "glue_to_surface"
```
KaiSzuttor: there has to be a blank between parameter ant the colon

src/python/espressomd/collision_detection.py
```diff
@@ -0,0 +1,181 @@
...
+               Collision deteciton mode
+          
+           distance: float
```
KaiSzuttor: please use the :obj:`float` etc. notation

src/script_interface/collision_detection/CollisionDetection.hpp
```diff
@@ -0,0 +1,68 @@
+/*
```
KaiSzuttor, line 1: please clang-format this file

testsuite/analyze_energy.py
```diff
@@ -24,12 +24,15 @@ def setUpClass(self):
         self.system.non_bonded_inter[1, 1].lennard_jones.set_params(
             epsilon=1.0, sigma=1.0,
             cutoff=2**(1. / 6.), shift="auto")
-        self.system.part.add(id=0, pos=[1, 2, 2], type=0)
-        self.system.part.add(id=1, pos=[5, 2, 2], type=0)
         self.system.thermostat.set_langevin(kT=0., gamma=1.)
         self.system.bonded_inter.add(self.harmonic)
 
 
+    def setUp(self):
```
KaiSzuttor, line 31: how is this related to the collision detection?

src/python/espressomd/collision_detection.py
```diff
@@ -0,0 +1,181 @@
...
+        raise Exception("Mode not hanled: "+mode.__str__())
+            
+            
+    
+    _int_mode={
+        "off":0,
+        "bind_centers":2,
+        "bind_at_point_of_collision":4,
+        "glue_to_surface":8,
+        "bind_three_particles":16}
```
hirschsn: It's probably better to import these magic constants from collision.hpp.

src/python/espressomd/collision_detection.py
```diff
@@ -0,0 +1,181 @@
...
+            
+            Bond types: int -> BondedInteraction
+            mode: int -> string
+
+            """
+        # Py3: Cast from binary to normal string. Don't understand, why a
+        # binary string can even occur, here, but it does.
+        name=to_str(name)
+        # Convert mode parameter into python enum
+        res=value
```
hirschsn: Is this intended? If the `name` is unknown to this method, it returns the `value` parameter.

src/core/rotation.cpp
```diff
@@ -500,13 +500,6 @@ void convert_vec_space_to_body(Particle *p, double *v, double *res) {
 }
 
 /** Multiply two quaternions */
```
hirschsn: Also remove the function comment.

src/core/interaction_data.hpp
```diff
@@ -943,6 +943,33 @@ int virtual_set_params(int bond_type);
...
 
+
+/** @brief Checks if particle has a pair bond with a given partner  
+*  Note that bonds are stored only on one of the two particles in Espresso
+* 
+* @param P
+* @param p          particle on which the bond may be stored
+* @param partner    bond partner 
+* @param bond_type  numerical bond type */ 
+inline bool bond_exists(const Particle* const p, const Particle* const partner, int bond_type)
```
hirschsn: Although your PR does not change the name of the function but only moves it, the fact that it only works for pair bonds should probably be reflected in its name.

src/core/collision.hpp
```diff
@@ -84,27 +91,69 @@ extern Collision_parameters collision_params;
...
+
+#ifdef VIRTUAL_SITES_RELATIVE
+  // Ignore virtual particles
+  if ((p1->p.isVirtual) || (p2->p.isVirtual))
+    return;
+#endif
+
+
+  // Check, if there's already a bond between the particles
+  if (bond_exists(p1,p2, collision_params.bond_centers))
```
hirschsn: You could get rid of some of the empty lines here and merge the conditionals. Also, the ghost check could be done before the bond_exists and even the glue to surface checks because it is cheaper. However, it is unlikely that the ghost check will rule out particle pairs.

src/core/collision.cpp
```diff
@@ -43,310 +44,255 @@ using namespace std;
...
 // particles
-static collision_struct *collision_queue = 0;
-static collision_struct *gathered_queue = 0;
+static std::vector<collision_struct> local_collision_queue;
 
-// Number of collisions recoreded in the queue
-static int number_of_collisions, total_collisions;
 
 /// Parameters for collision detection
 Collision_parameters collision_params = { 0, };
```
hirschsn: This leaves uninitialized fields. Are they guaranteed to be initialized (e.g. by the script interface) before accessed?

src/core/collision.cpp
```diff
@@ -654,110 +580,165 @@ void three_particle_binding_domain_decomposition()
...
-    for (int i=0;i<number_of_collisions;i++) {
-	// Create virtual site(s) 
+  // Gather the global collision queue, because only one node hsa a collision across node boundaries in its cueue.
+  // The other node might still have to change particle properties on its non-ghost particle
+  auto gathered_queue = gather_global_collision_queue();
+
+  // Sync max_seen_part
+  int tmp;
+  MPI_Allreduce(&max_seen_particle, &tmp, 1, MPI_INT, MPI_MAX, comm_cart);
+  max_seen_particle=tmp;
```
hirschsn: Could use `MPI_IN_PLACE` here.

src/core/collision.cpp
```diff
@@ -400,97 +345,87 @@ void coldet_do_three_particle_bond(Particle* p, Particle* p1, Particle* p2)
...
   
     // Initialize number of collisions gathered from all processors
-    for (int a=0;a<n_nodes;a++)
-        counts[a]=0;
-    
+    std::vector<int> counts(n_nodes);
     // Total number of collisions
-    MPI_Allreduce(&number_of_collisions, &total_collisions, 1, MPI_INT, MPI_SUM, comm_cart);
+    int total_collisions;
+    int tmp=local_collision_queue.size();
```
hirschsn: Should name this appropriately as it keeps being used in this function.

src/core/collision.cpp
```diff
@@ -654,110 +580,165 @@ void three_particle_binding_domain_decomposition()
...
-  gather_collision_queue(counts.data());
+  auto gathered_queue = gather_global_collision_queue();
 
-    if (counts[this_node]>0) {
 
       // If we don't have domain decomposition, we need to do a full sweep over all
       // particles in the system. (slow)
       if (cell_structure.type!=CELL_STRUCTURE_DOMDEC) {
-        three_particle_binding_full_search();
+        three_particle_binding_full_search(gathered_queue);
```
hirschsn: If the full search only works for 1 process and aborts if it is called on several, why do we need to gather the collision queue at all?

hirschsn (commented):
Changes are good. Collision.cpp should probably be reformatted (not in particular because of your changes but because its formatting has been misleading ever since).

I was just thinking of the gathering of the collision queue: We could optimize the three_particle_binding_domain_decomposition because (if at all) a process is only going to need information about collisions on neighboring (not all) processes if its own particles (ghost particles on the neighboring process) are involved. Or am I wrong here?

CMakeLists.txt
```diff
@@ -188,6 +189,8 @@ if (WITH_PYTHON)
     set(CYTHON_FLAGS "-3" CACHE STRING "Flags used by the Cython compiler during all build types.")
   endif()
 endif(WITH_PYTHON)
+find_python_module(enum REQUIRED)
```
fweik: Please add enum to `requirements.txt`

src/script_interface/collision_detection/CollisionDetection.hpp
```diff
@@ -0,0 +1,68 @@
...
+      {"part_type_to_be_glued",collision_params.part_type_to_be_glued},
+      {"part_type_to_attach_vs_to",collision_params.part_type_to_attach_vs_to},
+      {"part_type_after_glueing",collision_params.part_type_after_glueing}
+    });
+  };
+  Variant call_method(const std::string& name, const VariantMap& params) override {
+      if (name == "validate") {
+          return validate_collision_parameters();
+      };
+    return nullptr;
```
fweik: `return {};` or `return None{};` if you want to be more explicit. Then it will even compile.

fweik (changes requested):
I only looked at the ScriptInterface part.

src/script_interface/collision_detection/CollisionDetection.hpp
hmenke: Or `return none;`

RudolfWeeber:
> I was just thinking of the gathering of the collision queue: We could optimize the three_particle_binding_domain_decomposition because (if at all) a process is only going to need information about collisions on neighboring (not all) processes if its own particles (ghost particles on the neighboring process) are involved. Or am I wrong here?
Yes, for the three particle binding, a more narrow queue gathering is possible. For the virtual sites based schemes, the full queue is needede to execute particle creation book keeping on all nodes synchronously.
If it turns out to be necessary, a specialized version for three particle binding can be added, later.

> > +            res[k]=self._convert_param(k,res[k])
> +        return res
> +
> +
> +    def _convert_param(self,name,value):
> +        """Handles type conversion core -> python
> +            
> +            Bond types: int -> BondedInteraction
> +            mode: int -> string
> +
> +            """
> +        # Py3: Cast from binary to normal string. Don't understand, why a
> +        # binary string can even occur, here, but it does.
> +        name=to_str(name)
> +        # Convert mode parameter into python enum
> +        res=value
> 
> Is this intended? If the `name` is unknown to this method, it returns the `value` parameter.
Yes. The function converts only a subset of the parameters (for types that the script interface cannot handle), and returrns the remaining ones as they re.

>  /// Parameters for collision detection
>  Collision_parameters collision_params = { 0, };
> 
> This leaves uninitialized fields. Are they guaranteed to be initialized (e.g. by the script interface) before accessed?
Yes, for no collision detection, no further parametes are needed, and for all other method, the py interface ensures that a complete set of parameters is passed in.

RudolfWeeber:
I think, I addressed all points from the review.
In the end, a bit of re-shuffling was needed, because the Verlet criterion seems to be more narrow than in the old days. The collision cutoff is now explicitly included, there. This made some more initialization necessary, after all.

The three-particle binding wihtout domain decompositoin (full search) is now possible with more than one node (by including ghost cells). Tested with n_square.

Also, the return value of CollisionDetection.get_params() should now be consistent with what can be set.

fweik (approved):

PR #1688, opened 2017-12-05
Utils cleanup

fweik:
Description of changes:
 - Removed unused code from utils.
 - Removed intlist resizing functions.
 - Some modernization of bond and exclusion code.
 - Fixed removing all exclusions.
 - Added a probably working interface for local_stress_tensor stuff.


src/core/ghosts.cpp
```diff
@@ -313,17 +313,18 @@ void put_recv_buffer(GhostCommunication *gc, int data_parts)
...
+	  memmove(&n_exclusions, retrieve, sizeof(int));
 	  retrieve +=  sizeof(int);
-          if (n_bonds) {
-            realloc_intlist(&pt->el, pt->el.n = n_bonds);
-            std::copy(bond_retrieve, bond_retrieve + n_bonds, pt->el.e);
-            bond_retrieve += n_bonds;
+          if (n_exclusions) {
+	    pt->el.resize(n_exclusions);
+            std::copy_n(bond_retrieve, n_exclusions, pt->el.begin());
+            bond_retrieve += n_exclusions;
```
KonradBreitsprecher, line 327: This looks fishy, but I guess bond lists and exclusion lists can be treated similarly.

src/core/ghosts.cpp
```diff
@@ -450,11 +451,9 @@ void cell_cell_transfer(GhostCommunication *gc, int data_parts)
 	if (data_parts & GHOSTTRANS_PROPRTS) {
 	  memmove(&pt2->p, &pt1->p, sizeof(ParticleProperties));
 #ifdef GHOSTS_HAVE_BONDS
```
KonradBreitsprecher, line 453: So GHOSTS_HAVE_BONDS is actually GHOSTS_HAVE_BONDS_AND_EXCLUSIONS

src/core/initialize.cpp
```diff
@@ -84,12 +83,6 @@ static int reinit_particle_comm_gpu = 1;
 void on_program_start() {
   EVENT_TRACE(fprintf(stderr, "%d: on_program_start\n", this_node));
 
-/* tell Electric fence that we do realloc(0) on purpose. */
-#ifdef EFENCE
-  extern int EF_ALLOW_MALLOC_0;
-  EF_ALLOW_MALLOC_0 = 1;
-#endif
-
```
KonradBreitsprecher, line 92: Can't tell if this removal is valid.

src/core/particle_data.cpp
```diff
@@ -992,9 +984,10 @@ void remove_all_bonds_to(int identity) {
 void local_change_exclusion(int part1, int part2, int _delete) {
   if (part1 == -1 && part2 == -1) {
     for (auto &p : local_cells.particles()) {
-      realloc_intlist(&p.el, p.el.n = 0);
-      return;
+      p.el.clear();
     }
+
+    return;
```
KonradBreitsprecher, line 990: The return inside the particle loop seemed like a crucial bug..

src/core/topology.cpp
```diff
@@ -26,99 +26,62 @@
...
     }
   }
 
   topo_part_info_synced = 1;
-
 }
 
-
-int set_molecule_trap(int mol_num, int trap_flag,DoubleList *trap_center,double spring_constant, double drag_constant, int noforce_flag, int isrelative) {
+int set_molecule_trap(int mol_num, int trap_flag, DoubleList *trap_center,
```
KonradBreitsprecher, line 60: Could not find any places where this method is actually used..
No interface yet?

KonradBreitsprecher (approved):
Found nothing too crucial. I guess a testcase for combined usage of exclusions and bonds might be useful. Also the molecule trap stuff still needs interface.

src/core/ghosts.cpp
fweik: Why? I just don't reuse n_bonds for the exclusions anymore?

src/core/initialize.cpp
fweik: We don't have EF (Electric Fence) support in the build system anymore, so no need for this in the code. There are anyway better and more modern tools than EF.

src/core/particle_data.cpp
fweik: Yes, that was broken, but I'm not so sure this function is needed anymore anyhow.

src/core/topology.cpp
fweik: That's not interfaced but we decided to keep it for now.

PR #1701, opened 2017-12-11
Correlator checkpointing

### LINKED ISSUE #509: Correlation checkpointing broken with 2 nodes
The correlation checkpointing is broken (according to the testcase) for two nodes. Could not find an obvious reason. Testcast removed, because the correlator is undergoing an overhaul at the moment, and we won't fix in the old code. This is just a gentle remainder that there is some sort of problem.

### LINKED ISSUE #806: Pickle support for Script_Interface based classes
-> README.org lesen
-> script_interface/shapes/Wall.hpp und Wall.cpp lesen
-> script_interface/constraints/Constraint.hpp, Constraint.cpp und
Constraints.cpp lesen
Der interessante Punkt ist wie man beim Pickeln damit umgeht wenn sich
mehrere Python-
objecte ein C++ Objekt teilen. Das kann zubeispiel passieren wenn man ein
object als parameter
uebergibt und sich spaeter zurueck holt.

### LINKED ISSUE #878: Use c++ containers in Correlator class and re-implement checkpointing via boost serialization
Todo for individual member vars of the class
_args: Arguments of the correlation operation. Convert to vector<double>
*tau: Lag time for each result in units of dt
*__A and *__B: [level][tau_i][component]. There are twice as mnay elements in lvel 0 as in higher levels. 
*_result: [index in tau][component of correlation output]
*n_sweeps: Number of correlation outputs per index in tau vector<int>
*n_vals: Number of elements per level already present vector<int>
*newest index of newest entry in each hierarchy level vector<int>
*a_data, b_data: Linear storeage of data. To be removed
*result_data: Linearly stored results. To be removed
*accumlated_average and variance. vector<double> with dima and dimb

### LINKED ISSUE #1241: Random segfaults on correlator.finalize()
e.g. in the lj tutorial msd.py

fweik:
Fixes #1241, fixes #878, fixes #509, fixes #806.

# Description of changes:
## Core
 - Use STL/boost containers in core correlator
 - State serialization for core correlator
 - Removed pointless complex conjugate correlation op.

## ScriptInterface
 - Constructor support for SI objects
 - Removed the name member from the ScriptInterfaceBase interface,
   SI objects now always have the name they were created with.
 - Serialization for SI objects
 - Pickle support in the interface.

The desired functionality is implemented with this PR, but there is still some
room for improvement. We need to find a way to document and check the
constructor arguments. The error messages need improvement. I'd like to have
some input on this. 

I'll still provide extended tests as the checkpointing and serialization should
be tested independent of a specific implementation.


src/core/accumulators/Accumulator.hpp
```diff
@@ -43,16 +43,15 @@ class Accumulator {
 public:
   // The accumulator struct has to be initialized with the correct vector size,
   // therefore the order of init is important.
-  Accumulator() : m_initialized(false), m_autoupdate(false){};
-  std::shared_ptr<Observables::Observable> m_obs = nullptr;
-  void initialize();
+  Accumulator(std::shared_ptr<Observables::Observable> obs)
+      : m_obs(obs), m_acc(std::vector<double>(obs->n_values())) {}
```
hmenke: Maybe use `std::move` for `obs`?
```cpp
Accumulator(std::shared_ptr<Observables::Observable> obs)
    : m_obs(std::move(obs)), m_acc(std::vector<double>(obs->n_values())) {}
```

src/core/correlators/Correlator.cpp
```diff
@@ -1,891 +1,684 @@
...
+  if (!(A.size() == B.size())) {
+    throw std::runtime_error(
+        "Error in square distance componentwise: The vector sizes do not "
+        "match.");
+  }
+
+  auto const C_size = A.size() / 3;
+  if (3 * C_size == A.size()) {
+    throw std::runtime_error("Invalid dimensions.");
+  }
```
hmenke: Why must `A.size()` not be a multiple of 3?  I think you meant
```cpp
if (3 * C_size != A.size()) {
```

src/core/correlators/Correlator.cpp
```diff
@@ -1,891 +1,684 @@
...
+  std::transform(C.begin(), C.end(), C.begin(),
+                 [](double c) -> double { return std::exp(c); });
+
+  return C;
 }
 
+/* global variables */
 
+/* Error codes */
+const char init_errors[][64] = {
```
hmenke: `constexpr`

src/core/correlators/Correlator.cpp
```diff
@@ -1,891 +1,684 @@
...
+    "dt is smaller than the MD timestep",                            // 15
+    "dt is not a multiple of the MD timestep",                       // 16
+    "cannot set compress2 for autocorrelation",                      // 17
+    "diA must be divisible by 3 for fcs_acf",                        // 18
+    "fcs_acf requires 3 additional parameters"                       // 19
+};
 
-int Correlator::get_correlation_time(double* correlation_time) {
-  // We calculate the correlation time for each dim_corr by normalizing the correlation,
+const char file_data_source_init_errors[][64] = {
```
hmenke: `constexpr`

src/core/correlators/Correlator.cpp
```diff
@@ -1,891 +1,684 @@
...
+    "", "No valid filename given.", "File could not be opened.",
+    "No line found that was not commented out"};
+
+const char double_correlation_get_data_errors[][64] = {
```
hmenke: `constexpr`

src/core/correlators/Correlator.cpp
```diff
@@ -1,891 +1,684 @@
...
-  for (unsigned i = 0; i < dim_corr; i++ ) 
-    C[i] = 0;
-  for (unsigned i = 0; i < dim_A; i++ ) {
-    C [i/3] -= ( (A[i]-B[i])*(A[i]-B[i]) ) / wsquare[i%3];
-  }
-  for (unsigned i = 0; i < dim_corr; i++ ) 
-    C[i] = exp(C[i]);
-  return 0;
+void Correlator::set_internal_state(std::string const &state) {
+  std::stringstream ss(state);
```
hmenke: You could use
```cpp
boost::iostreams::array_source asource(state.c_str(), state.size());
boost::iostreams::stream<boost::iostreams::array_source> ss(asource);
boost::archive::binary_iarchive ar(ss);
```
to prevent copying `state`.  Needs
```cpp
#include <boost/iostreams/device/array.hpp>
#include <boost/iostreams/stream.hpp>
```

src/core/correlators/Correlator.hpp
```diff
@@ -113,263 +116,227 @@
...
-#include "utils.hpp" 
-#include "Vector.hpp"
-
 
 namespace Correlators {
-#define MAXLINELENGTH 2048
-
-#define MIN(a,b) ((a)>(b)?(b):(a))
-
+#define MIN(a, b) ((a) > (b) ? (b) : (a))
```
hmenke: Maybe refactor to `std::min` when you're at it?

src/core/correlators/Correlator.hpp
```diff
@@ -113,263 +116,227 @@
...
-  int n_columns;
-  char last_line[MAXLINELENGTH];
-  int data_left;
-};
-
 /********** The error codes ********************************/
 extern const char init_errors[][64];
 extern const char init_from_checkpoint_errors[][64];
 extern const char file_data_source_init_errors[][64];
 extern const char double_correlation_get_data_errors[][64];
```
hmenke: Are those even used elsewhere?  If not, please remove them from the header.

src/core/unit_tests/ParallelScriptInterface_test.cpp
```diff
@@ -17,10 +17,7 @@ std::unique_ptr<Communication::MpiCallbacks> callbacks;
 using namespace ScriptInterface;
 
 struct TestClass : public ScriptInterfaceBase {
-  TestClass() {
-    constructed = true;
-  }
-  
+  TestClass() { constructed = true; }
   ~TestClass() { destructed = true; }
```
hmenke, line 21: ~destructed~ destroyed
inb4 grammar nazi

src/python/espressomd/script_interface.pyx
```diff
@@ -11,15 +11,24 @@ cdef class PObjectId(object):
...
+    def __init__(self, name=None, policy="GLOBAL", oid=None, **kwargs):
+        cdef CreationPolicy policy_
+        cdef map[string, Variant] ctor_args
+
+        if policy == "GLOBAL":
+            policy_ = GLOBAL
+        elif policy == "LOCAL":
+            policy_ = LOCAL
+        else:
+            raise Exception("Unknown policy '{}'.".format(policy))
```
hmenke, line 24:
```python
if policy not in ["LOCAL", "GLOBAL"]:
    raise Exception("Unknown policy '{}'.".format(policy))
policy_ = policy
```

src/python/espressomd/script_interface.pyx
```diff
@@ -202,16 +222,29 @@ cdef class PScriptInterface(object):
 
         return odict
 
+def _unpickel_so_class(so_name, state):
```
hmenke: ~`_unpickel_so_class`~ `_unpickle_so_class`

src/python/espressomd/script_interface.pyx
```diff
@@ -202,16 +222,29 @@ cdef class PScriptInterface(object):
...
     _so_creation_policy = "GLOBAL"
 
     def __init__(self, **kwargs):
-        super(ScriptInterfaceHelper,self).__init__(self._so_name, self._so_creation_policy)
-        self.set_params(**kwargs)
+        super(ScriptInterfaceHelper,self).__init__(self._so_name, policy=self._so_creation_policy, **kwargs)
         self.define_bound_methods()
 
+    def __reduce__(self):
+        return (_unpickel_so_class , (self._so_name, self._serialize()))
```
hmenke: ~`_unpickel_so_class`~ `_unpickle_so_class`

src/script_interface/ScriptInterfaceBase.cpp
```diff
@@ -53,4 +64,75 @@ std::weak_ptr<ScriptInterfaceBase> &
...
+  return ss.str();
+}
+
+/**
+ * @brief Creates a new instance from a binary state,
+ *        as returned by @f serialize().
+ */
+std::shared_ptr<ScriptInterfaceBase>
+ScriptInterfaceBase::unserialize(std::string const &state) {
+  std::stringstream ss(state);
```
hmenke: You could use
```cpp
boost::iostreams::array_source asource(state.c_str(), state.size());
boost::iostreams::stream<boost::iostreams::array_source> ss(asource);
boost::archive::binary_iarchive ar(ss);
```
to prevent copying `state`.  Needs
```cpp
#include <boost/iostreams/device/array.hpp>
#include <boost/iostreams/stream.hpp>
```

hmenke (changes requested):

src/script_interface/get_value.hpp
```diff
@@ -96,6 +95,96 @@ template <typename T> T get_value(Variant const &v) {
...
+  }
+}
+
+/**
+ * @brief Make a new T with arguments extracted from a VariantMap.
+ */
+template <typename T, typename... Types, typename... ArgNames>
+T make_from_args(VariantMap const &vals, ArgNames... args) {
+  return T{get_value<Types>(vals, args)...};
+}
```
hmenke:
```cpp
template <typename T, typename... Types, typename... ArgNames>
T make_from_args(VariantMap const &vals, ArgNames&&... args) {
  return T{get_value<Types>(vals, std::forward<ArgNames>(args))...};
}
```

src/script_interface/get_value.hpp
```diff
@@ -96,6 +95,96 @@ template <typename T> T get_value(Variant const &v) {
...
+
+/**
+ * @brief Make a new std::shared_ptr<T> with arguments extracted from a
+ * VariantMap.
+ */
+template <typename T, typename... Types, typename... ArgNames>
+std::shared_ptr<T> make_shared_from_args(VariantMap const &vals,
+                                         ArgNames... args) {
+  return std::make_shared<T>(get_value<Types>(vals, args)...);
+}
```
hmenke:
```cpp
template <typename T, typename... Types, typename... ArgNames>
std::shared_ptr<T> make_shared_from_args(VariantMap const &vals,
                                         ArgNames&&... args) {
  return std::make_shared<T>(get_value<Types>(vals, std::forward<ArgNames>(args))...);
}
```

src/script_interface/get_value.hpp
```diff
@@ -96,6 +95,96 @@ template <typename T> T get_value(Variant const &v) {
...
+ *
+ * (In R (T::*m)(Args...), m is a pointer to member function of a T,
+ * returning R and taking Args as parameters.)
+ */
+template <typename T, typename R, typename... Args, typename... ArgNames>
+auto call_with_args(T &this_, R (T::*m)(Args...), VariantMap const &vals,
+                    ArgNames... args) -> R {
+  return (this_.*m)(get_value<Args>(vals, args)...);
+}
+
```
hmenke:
```cpp
template <typename T, typename R, typename... Args, typename... ArgNames>
auto call_with_args(T &this_, R (T::*m)(Args...), VariantMap const &vals,
                    ArgNames&&... args) -> R {
  return (this_.*m)(get_value<Args>(vals, std::forward<ArgNames>(args))...);
}
```

src/script_interface/get_value.hpp
```diff
@@ -96,6 +95,96 @@ template <typename T> T get_value(Variant const &v) {
...
+ *
+ * (In R (*m)(Args...), m is a function pointer,
+ * returning R and taking Args as parameters.)
+ */
+template <typename R, typename... Args, typename... ArgNames>
+auto call_with_args(R (*m)(Args...), VariantMap const &vals, ArgNames... args)
+    -> R {
+  return (*m)(get_value<Args>(vals, args)...);
+}
+
```
hmenke:
```cpp
template <typename R, typename... Args, typename... ArgNames>
auto call_with_args(R (*m)(Args...), VariantMap const &vals, ArgNames&&... args)
    -> R {
  return (*m)(get_value<Args>(vals, std::forward<ArgNames>(args))...);
}
```

hmenke (changes requested):

src/core/accumulators/Accumulator.hpp
fweik: obs is a pointer, so copy is cheap an there is no point in moving.

src/core/correlators/Correlator.hpp
fweik: Needs cleanup of the int types, as there are comparisons of different int types (It's `template<typename T> std::min(T,T)`, with a single `T`).

src/script_interface/get_value.hpp
fweik: Yes, but in practice ArgNames are `const char *`, so there is not much gain here. You missed that `get_value` always forces a copy, which is a bigger isssue. Will fix both of course.

src/core/accumulators/Accumulator.hpp
hmenke: It's a shared pointer, so copying means constructing a new control block, increasing the reference count, copying, decreasing the reference, destroying the control block.

src/core/unit_tests/ParallelScriptInterface_test.cpp
fweik: Buuuuut: https://www.merriam-webster.com/dictionary/destruct

src/python/espressomd/script_interface.pyx
fweik: policy_ is a (c) int, policy is a python string, GLOBAL and LOCAL are enum constants from the core. So your construction does not work without a helper function.

src/script_interface/ScriptInterfaceBase.cpp
fweik: That's a better solution. It would be nice to have zero-copy also for the output direction (::serialize()), but that seems not possible with either std or boost streams.

src/python/espressomd/script_interface.pyx
hmenke: Oops ¯\_(ツ)_/¯

hmenke (approved):

RudolfWeeber:
Correlator.cpp
```c++
  if (((t - ((m_tau_lin + 1) * ((1 << (i + 1)) - 1) + 1)) % (1 << (i + 1)) == 0)
```
If possible, please provide a comment explaining this expression

script_interface.pyx
```python
# We accept ints for floats (but not the other way round)
if <int> type == <int> DOUBLE and isinstance(in_params[pname], int):
```
Needs to tolerate np.int. (as in Konrads is_valid_type()))
```c++
+                    if isinstance(in_params[pname][i], int):
```
Same here

ScriptObjectRegistry.hpp:
```c++
-      if (obj_ptr == nullptr)
-        throw std::runtime_error("Wrong type");
```
Why is this no longer needed?


Otherwise, I think, it can be merged.

fweik:
> if (((t - ((m_tau_lin + 1) * ((1 << (i + 1)) - 1) + 1)) % (1 << (i + 1)) == 0)

Not mine, no idea. For parts of the code I have a hard time believing that
it is correct. It's also not tested well enough to catch subtle bugs. In the
long run this has to be replaced by something better.

fweik:
```c++
      if (obj_ptr == nullptr)
        throw std::runtime_error("Wrong type");
```
get_value does already check the type.

fweik:
the int stuff was already fixed in the master.

PR #1703, opened 2017-12-11
Fix rotation flag getter

mkuron:
Bug reported by @iammmiru while reading in checkpoints:
```
ValueError: The rotation flag has to be given as a tuple of 3 integers. -- Item 0 was of type float64
```


iammmiru:
Thank you :)

fweik (approved):

PR #1712, opened 2017-12-12
core: Fixed h5md bond writing.

### LINKED ISSUE #1303: Bonds with more than one partners in H5MD will fail
https://github.com/espressomd/espresso/blob/aecc6622ea2b3082185b4c5910222761e63bd1e8/src/core/io/writer/h5md/h5md_core.cpp#L350

The bond list format is e.g. bondindex1 Id1 Id2 bondindex2 Id1 Id2 Id3 ...
The current implementation assumes single partner bonds. This will cause errors, types and Ids will be mixed.

If the H5MD format doesn't support that, skip the bond and throw a warning.

fweik:
Fixes #1303.

Description of changes:
 - Correctly parse bond lists

Bonds with more than 1 partner are ignored, this would need a larger refactoring.

### DIFF
--- a/src/core/io/writer/h5md/h5md_core.cpp
+++ b/src/core/io/writer/h5md/h5md_core.cpp
@@ -21,6 +21,7 @@
 
 #include <vector>
 #include "h5md_core.hpp"
+#include "core/interaction_data.hpp"
 
 namespace Writer {
 namespace H5md {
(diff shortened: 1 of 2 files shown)


fweik:
Feel free to add. At least it does not write garbage anymore.

jonaslandsgesell (approved):

KaiSzuttor (approved):

PR #1720, opened 2017-12-15
close threads if window is closed with x-button

KonradBreitsprecher:
Calls `os._exit()`  if window is closed via x-button.

### DIFF
--- a/src/python/espressomd/visualization_opengl.pyx
+++ b/src/python/espressomd/visualization_opengl.pyx
@@ -22,8 +22,8 @@ class openGLLive(object):
     """This class provides live visualization using pyOpenGL.
     Use the update method to push your current simulation state after
     integrating. Modify the appearance with a list of keywords.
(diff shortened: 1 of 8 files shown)


src/python/espressomd/visualization_opengl.pyx
```diff
@@ -852,6 +852,9 @@ class openGLLive(object):
             self.specs['window_size'][1] = 1.0 * h
             # glPushMatrix()
 
+        def closeWindow():
```
KaiSzuttor: can you please rename pep8 conform? maybe also other functions in that module

KaiSzuttor (changes requested):

KonradBreitsprecher:
@KaiSzuttor pep8 it is

fweik (approved):

KaiSzuttor (approved):

PR #1721, opened 2017-12-15
changed definition of the hydrodynamic radius

davidsean:
Fixes #

Not sure if this is a good idea since it changes older behaviour.
Calculating the chain hydrodynamic radius involves finding the mean <1/r> distances.
The old prefactor of N^2 in
old
is now N(N-1)
new

Description of changes:
 - change the prefactor in calculating the average

PR Checklist
------------
 - [x] Tests?
   - [ ] Interface
   - [x] Core 
 - [x] Docs?

### DIFF
--- a/doc/sphinx/analysis.rst
+++ b/doc/sphinx/analysis.rst
@@ -530,7 +530,7 @@ The following formula is used for the computation:
 .. math::
 
    \label{eq:Rh}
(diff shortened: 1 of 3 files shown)


hmenke (commented):
Please remove code rather than commenting it out.
What is the motivation behind this change?

davidsean:
The current definition is an approximation valid at large N. This form is fine for analytical work in that limit (not to mention easier to write on paper and aesthetically pleasing). My suggestion is valid for all N. For the numerical work we do, there is no reason to use an approximation if we can have the exact result.

If you are not convinced, see attached plots of the RH and the diffusion coefficient from Zimm theory (using the two forms).
rh-n
diffusion-n

fweik (approved):

PR #1723, opened 2017-12-18
Tab refactor

### LINKED ISSUE #1075: Tabulated bond interpolates incorrectly. State of tabulated non-bonded unclear
The tabulated bond interpolates for 1/r rather than r. This is especially relevant for r close to 0.
A tentative fix is here. https://github.com/RudolfWeeber/espresso/commit/c84832b64936abe077358d7367f96e519b4bcb92

The state of the tabulated non-bonded IA seems to be unclear 
* the tcl test does not actually test the erngy of the tabulated potential against a stored result. It just checks that Espresso correctly adds different non-bonded energies.
* The tcl testcase does check forces against stored resutls
* The Python test also contains the non-test for energies, but no test for forces

Conceptually, tabulated bonded and non-bonded force share the file format and interpolation scheme. In practice, the code exists twice, in slightly different fashion.
This needs serious refactoring. Also, we need to decide whether to continue to support the file format in Python or rather just let the user pass in numpy arrays or a python function for the energy, for which derivatives are calculated automatically.

At the very least, both interaction sshould share the file reader as well as the table lookup and interpolation code.

Also, there needs to be a testcase which compares the tabulated interaction to one for which an analytical version is available. E.g., LJ, Fene.

### LINKED ISSUE #1695: Tabulated potentials should not do file io
but take the potential table as parameter. Also there is no good reason why the tables are stored in a global and not the corresponding IA_Params for the type pair.

fweik:
Fixes #1695, fixes #1075.

Description of changes:
 - Tables are now stored in the interactions
 - Forces and energies are directly interpolated from
   the tables, nor 1/r factor.
 - In corner cases the bonded interaction tables can leak,
   there is now way doing this clearly with the current structure.

PR Checklist
------------
 - [x] Tests?
   - [x] Interface
   - [x] Core 
 - [x] Docs?


src/core/tab.cpp
```diff
@@ -27,188 +27,67 @@
...
-                                char *filename) {
-  int i, token = 0, size;
-  double dummr;
-  FILE *fp;
-
+                                TabulatedBondedInteraction tab_type, double min,
+                                double max, std::vector<double> const &energy,
+                                std::vector<double> const &force) {
   if (bond_type < 0)
     return 1;
```
KaiSzuttor, line 55: throw?

src/python/espressomd/interactions.pyx
```diff
@@ -2007,11 +2008,20 @@ IF TABULATED == 1:
 
             Parameters
             ----------
-            type : :obj:`str`
-                   Specifies the type of bonded interaction. Possible inputs:
-                   'distance', 'angle' and 'dihedral'.
-            filename : :obj`str`
-                       Filename of the tabular.
+
```
KaiSzuttor, line 2011: blank not needed

src/python/espressomd/interactions.pyx
```diff
@@ -2114,31 +2113,55 @@ IF TABULATED == 1:
...
             """
-            return ["filename", ]
+            return ["min", "max", "energy", "force"]
+
+        def set_params(self, **kwargs):
+            """ Set parameters for the TabulatedNonBonded interaction.
+
+            Parameters
+            ----------
+
```
KaiSzuttor, line 2129: blank not needed

testsuite/tabulated.py
```diff
@@ -17,1616 +17,73 @@
...
+    def check(self):
+        # Below cutoff
+        self.assertTrue(np.allclose(self.s.part[:].f, 0.0))
 
-        self.system.box_l = [10.0, 10.0, 10.0]
-        self.system.time_step = 0.01
+        for z in np.linspace(0, self.max_ - self.min_, 200, endpoint=False):
+            self.s.part[1].pos = [5., 5., 6. + z]
+            self.s.integrator.run(0)
+            self.assertTrue(np.allclose(self.s.part[0].f, [0., 0., 5. + z * 2.3] ))
```
KaiSzuttor: please use https://docs.scipy.org/doc/numpy-dev/reference/generated/numpy.testing.assert_allclose.html

KaiSzuttor (commented):
LGTM

KaiSzuttor (approved):

PR #1725, opened 2017-12-18
Fixing histogram normalization

KaiSzuttor:


src/core/observables/FluxDensityProfile.hpp
```diff
@@ -19,9 +19,9 @@ class FluxDensityProfile : public ProfileObservable {
     Utils::Histogram<double> histogram(n_bins, 3, limits);
     for (int id : ids) {
       auto const ppos = ::Vector<3, double>(folded_position(partCfg[id]));
-      histogram.update(ppos, ::Vector<3, double> {{partCfg[id].m.v[0],
-                                                   partCfg[id].m.v[1],
-                                                   partCfg[id].m.v[2]}});
+      histogram.update(ppos, ::Vector<3, double> {{partCfg[id].m.v[0] / time_step,
+                                                   partCfg[id].m.v[1] / time_step,
+                                                   partCfg[id].m.v[2] / time_step}});
```
mkuron, line 24: Are you sure they should be divided by the time step and not multiplied? The units on these weights look a bit odd.

src/core/utils/Histogram.hpp
```diff
@@ -230,24 +240,28 @@ template <typename T> void Histogram<T>::normalize() { do_normalize(); }
...
+template <typename T>
+class CylindricalHistogram : public Histogram<T> {
 public:
-  using Histogram::Histogram;
+  using Histogram<T>::Histogram;
+  using Histogram<T>::get_n_bins;
+  using Histogram<T>::get_limits;
+  using Histogram<T>::get_bin_sizes;
+  using Histogram<T>::m_hist;
+  using Histogram<T>::m_tot_count;
```
mkuron, line 258: I don't think you need to `using` all these member functions. They should all come through inheritance already, the only exception being the constructor.

src/core/observables/FluxDensityProfile.hpp
KaiSzuttor: Actually this is correct and tested in observable_profile.

src/core/utils/Histogram.hpp
KaiSzuttor: Not true for template inheritance.

mkuron (approved):

PR #1733, opened 2017-12-19
Checking int, float against various numpy types

### LINKED ISSUE #1630: py: Type checking for int and float needs to cope with numpy types
E.g., access to system.non_bonded_inter[i,j] with i,j being output of ParticleSlice[:].type

KonradBreitsprecher:
Fixes #1630

Added helper function:
```
def is_valid_type(value, t):
    if t == int:
        return isinstance(value, (int, np.integer, np.long))
    elif t == float:
        return isinstance(value, (float, np.float16, np.float32, np.float64, np.float128, np.longdouble))
```
Replaced all instances of `isinstance(.. , int)` and `isinstance(.. ,float)` with `is_valid_type`.

int returns true for:
- int      
- np.short    
- np.long     
- np.longlong 
- np.int   
- np.int_  
- np.intc  
- np.intp  
- np.int8  
- np.int16 
- np.int32 
- np.int64 
- np.uint8 
- np.uint16
- np.uint32
- np.uint64

float returns true for:
- float        
- np.float     
- np.float_    
- np.float16   
- np.float32   
- np.float64   
- np.float128  
- np.double    
- np.longdouble

### DIFF
--- a/src/python/espressomd/analyze.pyx
+++ b/src/python/espressomd/analyze.pyx
@@ -35,7 +35,7 @@ cimport numpy as np
 from globals cimport n_configs, min_box_l
 from collections import OrderedDict
 from .system import System
-
+from espressomd.utils import is_valid_type
 
 class Analysis(object):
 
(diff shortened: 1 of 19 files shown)


RudolfWeeber:
On Tue, Dec 19, 2017 at 02:12:19PM +0000, KonradBreitsprecher wrote:
> Fixes #1630
> 
> Added helper function:
> ```
> def is_valid_type(value, t):
>     if t == int:
>         return isinstance(value, (int, np.integer, np.long))
>     elif t == float:
>         return isinstance(value, (float, np.float16, np.float32, np.float64, np.float128, np.longdouble))
> ```
else:
    return isinstance(value,t)

Then, it can be used in all cases.

fweik (approved):

PR #1741, opened 2017-12-28
Corrected reaction ensemble documentation, small changes in RE syntax…

### LINKED ISSUE #1650: reaction ensemble docs need to be corrected
Sections 6.6 - 6.8 in the user guide need changes. Some statements are incorrect due to previous misunderstandings, others could be simplified.

Some parameters of the reaction_ensemble module need to be fixed:
pressure - its meaning is incorrectly described. 
default_charges - it is unclear from the docs how it is used

The order of types in the list of reactant_types and product_types matters for the algorithm although it does not matter for the thermodynamics. This point is presumably not mentioned.

I can plan to work on this on Monday November 27th.

kosovan:
…, and corrections to teststuite and samples to comply with the current syntax.

Partly Fixes #1650 Some issues still need to be clarified.

Description of changes:
 - fixed the description of the standard_pressure
 - rephrased documentation of several other functions and parameters
 - changed RE.add() to RE.add_reaction() to make the semantics consistent with RE.add_collective_variable..()
 - adopted the samples to the current state of Reaction Ensemble code

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


hmenke (approved):

src/python/espressomd/reaction_ensemble.pyx
```diff
@@ -11,34 +11,32 @@ class WangLandauHasConverged(Exception):
...
     standard_pressure : :obj:`float`
-                        The pressure in simulation units where the reactions
-                        should occur. This is an input parameter of the
-                        reaction ensemble.
+                        The standard pressure (in simulation units)
+                        at which the reaction constant was determined.
+                        Note that if you provide the reaction constant :math:`K_c`
+                        as described in the user guide, this parameter should be kept at its default value.
+                        This paramater is retained for backward compatibility with earlier
+                        implementation, where the pressure-based constaht :math:`K_p` was
```
jonaslandsgesell: The user guide only deals with a dimensionless reaction constant K (K_p is not mentioned there). Please also name it this way here.

src/python/espressomd/reaction_ensemble.pyx
```diff
@@ -400,12 +397,12 @@ cdef class WangLandauReactionEnsemble(ReactionAlgorithm):
 
     def add_collective_variable_degree_of_association(self, *args, **kwargs):
         """
-        Adds a reaction coordinate of the type degree of association.
+        Adds degree of association as the collective variable (reaction coordinate) for the Wang-Landau Reaction Ensemble.
```
jonaslandsgesell: You can have multiple reaction coordinates of this type. Therefore it was previsously "adds a reaction coordinate of the type..." implying that you can have more than one reaction coordinates of this type.

Therefore I would explicitly state that you can have more than one reaction coordinate of this type.

jonaslandsgesell (changes requested):

src/python/espressomd/reaction_ensemble.pyx
jonaslandsgesell: Sorry i overlooked parts. But how is K=Exp(...) related to K_p?

jonaslandsgesell: I don't see that there is a statement K=K_p although it is used in that way. After all this only holds for ideal gases but all the reaction constants which are provided in the reaction ensemble are ideal Gas reaction constants. Adding Interactions to the system effectively alters reaction constant K_c which is observed in the system then. Can we introduce a statement like this or so you see Problems with it?

kosovan: You are right, K_p is not defined anywhere. At the same time, though K_p = prod_i (p_i / p_standard)^{nu_i} is well defined only if the system is a gas, and it is equal to K only if the system is an ideal gas. Regarding the message of this note, saying that standard_pressure is needed only in gas-phase reactions, I would leave it as is. I expect that most users need not bother, and those who need to study gas-phase reactions should be familiar with K_p.

src/python/espressomd/reaction_ensemble.pyx
kosovan: Agreed. I will add a note that several collective variables can be set simultaneously.

kosovan:
@jonaslandsgesell and @helvrud please check my changes and comment if any further changes are needed.

I think that together with #1767 this is the final fix for #1650 After these two PRs are merged, this issue can be closed.

doc/sphinx/run.rst
```diff
@@ -225,7 +225,15 @@ where :math:`\Delta E=E_\mathrm{new}-E_\mathrm{old}` is the change in potential
...
-The parameter :math:`\Gamma` proportional to the reaction constant. It is defined as
+The parameter :math:`\Gamma`  is defined as
+
+.. math::
+
+   \Gamma = \prod_i \Bigl(\frac{q_i}{\Lambda_i} \Bigr)^{\bar\nu} 
+
+where :math:`q_i` and :math:`\Lambda_i` are the single-particle partition function and
+thermal wavelength of species :math:`i`. In the absence of intermolecular interactions
+(ideal gas or ideal solution), :math:`\Gamma` proportional to the concentration-based reaction constant
```
jonaslandsgesell: Gamma is the concentration based reaction constant in units of 1/(sigma^3)^nu_bar, if no interactions are present. Proportional is too weak for my taste.

src/core/reaction_ensemble.hpp
```diff
@@ -89,7 +89,8 @@ class ReactionAlgorithm {
 
   std::vector<SingleReaction> reactions;
   std::map<int, double> charges_of_types;
-  double standard_pressure_in_simulation_units = -10.0;
+  //double standard_pressure_in_simulation_units = -10.0;
+  //double standard_pressure_in_simulation_units = 1.0;
```
jonaslandsgesell: Please remove the comment

testsuite/constant_pH.py
```diff
@@ -36,14 +36,18 @@ class ReactionEnsembleTest(ut.TestCase):
     type_A = 1
     type_H = 5
     temperature = 1.0
-    standard_pressure_in_simulation_units = 0.00108
     # avoid extreme regions in the titration curve e.g. via the choice
     # (np.random.random()-0.5)
+#    target_alpha=0.5; # choose target alpha close to 0.5 to get good statistics in a small number of steps
+#    Ka=target_alpha*target_alpha/(1.0-target_alpha)*c0; # determine Ka which corresponds to the desired c0 and alpha
+#    pKa=-np.log10(Ka);
```
jonaslandsgesell: I did not choose 0.5 because this value was quite robust regarding errors from my experience for typical errors which happened during coding. But of course in principle any point on the curve should be suitable for testing, so I don't have a preference.

testsuite/constant_pH.py
```diff
@@ -36,14 +36,18 @@ class ReactionEnsembleTest(ut.TestCase):
...
     pKa_minus_pH = 1
     pH = 2  # or randomly via: 4*np.random.random()
     pKa = pKa_minus_pH + pH
     # could be in this test for example anywhere in the range 0.000001 ... 9
     K_HA_diss_apparent = 10**(-pKa)
+    #K_HA_diss_apparent = 10**(-pKa)*0.00108; # first, ensure that by removing standard pressure we obtain the same hard-coded numbers as before
+    #print("pKa:", pKa, "pk_pKa:", -np.log10(K_HA_diss_pk));
```
jonaslandsgesell: Please remove

testsuite/reaction_ensemble.py
```diff
@@ -29,17 +29,25 @@
...
     type_A = 1
     type_H = 2
     temperature = 1.0
-    standard_pressure_in_simulation_units = 0.00108
     exclusion_radius = 1.0
     # could be in this test for example anywhere in the range 0.000001 ... 9,
     # chosen for example like 8.8*np.random.random()+0.2
-    K_HA_diss = 8.8 * 0.5 + 0.2
+    # FIXME choose desired alpha and then calculate Ka based on the desired alpha
+    # use alpha close to 0.5 to get good statistics in few steps,
```
jonaslandsgesell: Please don't leave comments.

jonaslandsgesell:
Looks good to me

doc/sphinx/run.rst
kosovan: I agree but I will not specify the units because these are defined by the simulation script. The units have to be consistent with other units in the simulation script. Can be mol/L or 1/sigma^3 or anything else.

jonaslandsgesell: It all depends on the unit of the volume that is provided that is correct. However for the case that the user does not provide a volume (which is standard and V=box_l_x*box_l_y*box_l_z, V is in units of sigma^3) and then as a standard the unit of gamma is 1/(sigma^3)^nubabr

kosovan: The sentence before says: "Note that the dimension of :math:`\Gamma` is :math:`V^{\bar\nu}`, therefore its units must be consistent with the units in which Espresso measures the box volume, i.e. :math:`\sigma^3`." Isn't this precisely the same thing as you wrote, just spelled differently?

jonaslandsgesell (approved):

jonaslandsgesell (approved):

testsuite/reaction_ensemble.py
```diff
@@ -93,28 +96,43 @@ def test_ideal_titration_curve(self):
...
-        K_apparent_HA_diss = K_HA_diss * standard_pressure_in_simulation_units / temperature
-        pK_a = -np.log10(K_apparent_HA_diss)
-        print(average_degree_of_association)
-        real_error_in_degree_of_association = abs(
-            average_degree_of_association - ReactionEnsembleTest.ideal_degree_of_association(
-                pK_a, pH)) / ReactionEnsembleTest.ideal_degree_of_association(
-            pK_a, pH)
+        pK_a = -np.log10(K_HA_diss)
+        #ideal=ReactionEnsembleTest.ideal_degree_of_association( pK_a, pH);
+        ideal=0.05080885682460401; # hard-coded value from the original setup by Jonas
```
jonaslandsgesell: I would not hard code these values but use the function that determines these values

testsuite/reaction_ensemble.py
```diff
@@ -93,28 +96,43 @@ def test_ideal_titration_curve(self):
...
+        target=0.05024999999999979; # hard-coded value from the original setup by Jonas
```
jonaslandsgesell: what is target used for? It could be removed. That s where the Mac Os test is currently failing.

testsuite/constant_pH.py
```diff
@@ -64,14 +64,14 @@ def setUpClass(cls):
...
-            Gamma=cls.K_HA_diss_apparent, reactant_types=[
+            Gamma=cls.Ka, reactant_types=[
                 cls.type_HA], reactant_coefficients=[1], product_types=[
                 cls.type_A, cls.type_H], product_coefficients=[
                 1, 1], default_charges={cls.type_HA: 0, cls.type_A: -1, cls.type_H: +1})
         cls.RE.constant_pH = cls.pH
 
     @classmethod
-    def ideal_degree_of_association(cls, pH):
+    def ideal_alpha(cls, pH):
```
jonaslandsgesell: this is the formula for the degree of association= 1-degree of dissociation=1-alpha, please dont use alpha but a more descriptive name!

testsuite/constant_pH.py
```diff
@@ -38,15 +38,15 @@ class ReactionEnsembleTest(ut.TestCase):
     temperature = 1.0
     # avoid extreme regions in the titration curve e.g. via the choice
     # choose target alpha not too far from 0.5 to get good statistics in a small number of steps
-    pKa_minus_pH = 1
-    pH = pKa_minus_pH  # or randomly via: 4*np.random.random()
-    pH = 2  # or randomly via: 4*np.random.random()
+    pKa_minus_pH = 0
+    pH = pKa_minus_pH
```
jonaslandsgesell: this line is overwritten by the next line

testsuite/reaction_ensemble.py
```diff
@@ -108,42 +109,34 @@ def test_ideal_titration_curve(self):
...
+        for i in range(num_samples):
+            RE.reaction(10); 
+            average_NH += system.number_of_particles( type=type_H)
+            average_NHA += system.number_of_particles( type=type_HA)
+            average_NA += system.number_of_particles( type=type_A)
+        average_NH /= num_samples
+        average_NA /= num_samples
+        average_NHA /= num_samples
+        average_alpha = average_NA / float(N0)
+        print("average_NH:", average_NH,
```
jonaslandsgesell: please dont print in tests

testsuite/reaction_ensemble.py
```diff
@@ -81,7 +81,8 @@ def setUpClass(cls):
             reactant_types=cls.reactant_types,
             reactant_coefficients=cls.reactant_coefficients,
             product_types=cls.product_types,
-            product_coefficients=cls.product_coefficients, default_charges={cls.type_HA: 0, cls.type_A: -1, cls.type_H: +1}, check_for_electroneutrality=True)
+            product_coefficients=cls.product_coefficients, 
+            default_charges={cls.type_HA: 0, cls.type_A: -1, cls.type_H: +1}, check_for_electroneutrality=True)
 
     @classmethod
     def ideal_alpha(cls, Gamma,N0,V,nubar):
```
jonaslandsgesell: same here

testsuite/constant_pH.py
jonaslandsgesell: Looks to me like exchanging the degree of association and the degree of dissociation works here because both are 0.5 here. Therefore I would choose a point in the parameter space where both are distinct.

jonaslandsgesell:
The reaction ensemble method was moved by @KonradBreitsprecher to the appendix since it was not obvious that the code is documented. If you add a minimal example to the documentation on how to setup a reaction in the rst you can move it to the advanced methods section.

jonaslandsgesell (approved):
Looks good to me

fweik:
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/pipelines/1843

PR #1745, opened 2018-01-08
Add more tools to the pypresso script

### LINKED ISSUE #1744: Debugging needs to be better documented
The docs should point out that running tools on the pypresso script does not work as expected for most tools. It should also list the `--TOOL` options for the script.

hmenke:
Fixes #1744.

### DIFF
--- a/src/python/pypresso.cmakein
+++ b/src/python/pypresso.cmakein
@@ -17,13 +17,19 @@ export PYTHONPATH
 
 if [ "$1" = "--gdb" ]; then
   shift
-  exec gdb -ex run --args @PYTHON_FRONTEND@ $@
+  exec gdb --args @PYTHON_FRONTEND@ $@
 elif [ "$1" = "--lldb" ]; then
   shift
-  exec lldb -o "run $@" @PYTHON_FRONTEND@
+  exec lldb -- @PYTHON_FRONTEND@ $@
+elif [ "$1" = "--valgrind" ]; then
+  shift
+  exec valgrind --leak-check=full @PYTHON_FRONTEND@ $@
 elif [ "$1" = "--cuda-gdb" ]; then
   shift
-  exec cuda-gdb -ex run --args @PYTHON_FRONTEND@ $@
+  exec cuda-gdb --args @PYTHON_FRONTEND@ $@
+elif [ "$1" = "--cuda-memcheck" ]; then
+  shift
+  exec cuda-memcheck @PYTHON_FRONTEND@ $@
 fi
 
 exec @PYTHON_FRONTEND@ $@
--- a/src/python/pypresso.cmakein
+++ b/src/python/pypresso.cmakein
@@ -7,8 +7,6 @@
 # without any warranty.
 #
 
-set -x
-
 if test -n "$PYTHONPATH"; then
   PYTHONPATH=@PYTHON_DIR@:$PYTHONPATH
 else
--- a/src/python/pypresso.cmakein
+++ b/src/python/pypresso.cmakein
@@ -7,6 +7,7 @@
 # without any warranty.
 #
 
+set -x
 
 if test -n "$PYTHONPATH"; then
   PYTHONPATH=@PYTHON_DIR@:$PYTHONPATH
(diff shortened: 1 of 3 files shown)


src/python/pypresso.cmakein
```diff
@@ -17,13 +17,19 @@ export PYTHONPATH
...
   shift
-  exec gdb -ex run --args @PYTHON_FRONTEND@ $@
+  exec gdb --args @PYTHON_FRONTEND@ $@
 elif [ "$1" = "--lldb" ]; then
   shift
-  exec lldb -o "run $@" @PYTHON_FRONTEND@
+  exec lldb -- @PYTHON_FRONTEND@ $@
+elif [ "$1" = "--valgrind" ]; then
+  shift
+  exec valgrind --leak-check=full @PYTHON_FRONTEND@ $@
```
mkuron: Need to be able to optionally add extra arguments to `valgrind` as it is not interactive like the debuggers.

mkuron (changes requested):

src/python/pypresso.cmakein
hmenke: To be honest, if someone needs to add extra arguments to `valgrind`, this person is probably smart enough to open the script and extract the relevant stuff.  And I think `--leak-check=full` is most common.

hmenke:
Actually we could replace the `pypresso` script by a Python script as well.
```python
#!/usr/bin/env python

import sys

# Append Espresso to Python path
es_path = "@PYTHON_DIR@"
if es_path not in sys.path:
    sys.path.append(es_path)

# Remove the starter from the command line
del sys.argv[0]

if len(sys.argv) != 0:
    # There is a script present. Run it!
    execfile(sys.argv[0])
else:
    # No script :( Drop into REPL
    import code
    code.interact()
```
But that can also not be run directly in the debugger.

src/python/pypresso.cmakein
mkuron: Smart people can be lazy too. pypresso is all about convenience. That's why I'd like to be able to do something like this: `./pypresso --valgrind='--leak-check=full --verbose --trace-children=yes --time-stamp=yes' mysim.py 1.0 1.0 3.1 0.7`. If no arguments are specified, your default of `--leak-check=full` should be used.

hmenke: The syntax you requested should be feasible now.  I'm not a Bash wizard so my solution comes with quite a bit of code duplication.

src/python/pypresso.cmakein
```diff
@@ -15,15 +16,53 @@ else
...
-
-exec @PYTHON_FRONTEND@ $@
+case $1 in
+    --gdb)
+        shift
+        exec gdb --args @PYTHON_FRONTEND@ $@
+        ;;
+    --lldb)
+        shift
+        exec lldb -- @PYTHON_FRONTEND@ $@
```
mkuron, line 24: Have you tested whether this works? Previously we supplied the arguments to the `run` command.

 Is there any reason why you no longer have the debuggers run the program directly? You can‘t even set breakpoints before the first run because we use dynamic libraries, so you might as well run immediately.

hmenke: This form of the `lldb` command is what I found [online](https://lldb.llvm.org/lldb-gdb.html).

No, you can set breakpoints without shared libraries loaded (only tab-completion won't work).
```
(gdb) break integrate_vv
Function "integrate_vv" not defined.
Make breakpoint pending on future shared library load? (y or [n]) y
Breakpoint 1 (integrate_vv) pending.
(gdb) run
[...]
Thread 1 "python" hit Breakpoint 1, 0x00007fffed9749a0 in integrate_vv(int, int)@plt ()
```
*EDIT:* I also tested the `lldb` command.  It works as is.
```
(lldb) breakpoint set --name integrate_vv
Breakpoint 1: no locations (pending).
WARNING:  Unable to resolve breakpoint to any actual locations.
(lldb) run
[...]
* thread #1, name = 'python', stop reason = breakpoint 1.1
    frame #0: 0x00007fffeda0ed80 libEspressoCore.so.4`integrate_vv(int, int)
```

mkuron (approved):

fweik (approved):

PR #1748, opened 2018-01-08
doc: switch from numpydoc to napolean extension

KaiSzuttor:
Description of changes:
 - remove numpydoc dependency by using napolean extension to sphinx which is part of sphinx since version 1.3

### DIFF
--- a/doc/sphinx/zrefs.bib
+++ b/doc/sphinx/zrefs.bib
@@ -438,24 +438,6 @@ pages = {1305--1320}
   timestamp = {2007.08.29}
 }
 
-@ARTICLE{hofling11a,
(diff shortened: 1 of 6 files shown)


doc/sphinx/zrefs.bib
```diff
@@ -438,24 +438,6 @@ @BOOK{hockney88
   timestamp = {2007.08.29}
 }
 
-@ARTICLE{hofling11a,
```
pkreissl, line 441: Is there a reason for removing this entry?

pkreissl (approved):
Looks good to me.

doc/sphinx/zrefs.bib
pkreissl: OK, offline discussion: This article isn't cited anywhere, which produces a warning. Thus removed.

PR #1749, opened 2018-01-08
Virtual sites switchable at runtime

### LINKED ISSUE #1287: Improve performance of virtual sites relative
The presence of VIRTUAL_SITES_RELATIVe in myconfig.hpp slows down ES by a factor of 2 as compared to a simulation with ROTATION_PER_PARTICLE and all rotations turned off.
This is true, even if no virtual sites are present.

The reason is probably additional ghost communicatoin. For some of them (particularly the one in integrate_vv()), I don't understnad the origin. Also, they can be skipped if no vs are in the system.
Furthermore, on a single node, the additional ghost force communicator after distribution of vs forces to their controlling particles can be skipped, as the update is done via local_particles lookup, which on a single node should point to the non-ghost particle.

RudolfWeeber:
This makes the virtual sites implementation switchable at runtime.
By default, VirtualSitesOff, which does not touch virtual sites during the integration, is active.
VirtualSitesRelative is also provided.
Whether or not virtual sites have velocity is controlled at runtime via the have_velocity attribute of the implementations.
Virtual sites implementations decide on the need for ghost communication before/after vs updates and force/torque back transfers.
Ghosts_have_v flag is updated based on the have_velocity attribute of the vs implementation. 
Contains communicator data parts fix from @fweik.
Docs are adjusted.
Testcase Virtual_sites_relative checks virtual sites scheme and velocity switching as well.
Raspberry electrophoresis tutorial adapted.
VIRTUAL_SITES_THERMOSTAT and THERMOSTAT_IGNORE_NON_VIRTUAL features are remofed, as I'm not aware of a physically useful application.
VIRTUAL_SITES_COM (at center of mass of a set of particles) is postponed to ESPResSo 4.1, unless someone needs it urgently.
CI: Rest1 and rest2 configs are removed, as they tested VS stuff no longer relevant. METADYNAMICS is added to the maxset config.

This PR constitutes an interface change.

Fixes #1287 (in that un-needed ghost comms are removed. The performance penalty from the larger particle size will stay, though)


src/core/initialize.cpp
```diff
@@ -737,10 +735,12 @@ void on_ghost_flags_change() {
...
-  ghosts_have_v = 1;
+  // If they have velocities, VIRUTAL_SITES need v to update v of virtual sites
+  if (virtual_sites()->have_velocity()) {
+    ghosts_have_v = 1;
+  };
 #endif
 
-  if (old_have_v != ghosts_have_v)
-    cells_re_init(CELL_STRUCTURE_CURRENT);
+//  if (old_have_v != ghosts_have_v)
```
fweik: Remove

src/core/virtual_sites.hpp
```diff
@@ -35,54 +34,54 @@
...
- #include "virtual_sites_com.hpp"
-#endif
-
-// Virtual particles relative to position and orientation of a real particle
-#ifdef VIRTUAL_SITES_RELATIVE
- #include "virtual_sites_relative.hpp"
+#include <memory>
+
+/** @brief Base class for virtual sites implementations */
+class VirtualSites {
```
fweik: The `need_*` construction seems artificial to me. These are implementation details of the VS method and should be handled internally. If you think that they should be handled generically, you can make the actual update impl private, like update_impl(), and add a update() implementation to the base class that runs the ghost updates if needed before and after the call to update_impl. If you think this is specific to a certain implementation you can also move it there. As there is currently only one real implementation this is  hard to tell for me.
  
The same applies to `back_transfer_forces_and_torques()`.

src/script_interface/virtual_sites/ActiveVirtualSitesHandle.hpp
```diff
@@ -0,0 +1,61 @@
...
+namespace ScriptInterface {
+namespace VirtualSites {
+
+class ActiveVirtualSitesHandle : public AutoParameters {
+public:
+  ActiveVirtualSitesHandle() {
+      add_parameters({
+          {"implementation", 
+                     [this](Variant const &value) {
+                       m_active_implementation =
```
fweik, line 42: The type check is not needed as `get_value` will throw if value can not be cast into the correct type, you can just do `set_virtual_sites(get_value<std::shared_ptr<VirtualSites>>(value))`.

src/core/forces.cpp
```diff
@@ -107,8 +109,9 @@ void force_calc() {
   // to reassemble the cell lists after all position updates, also of virtual
   // particles.
   if ((lattice_switch & LATTICE_LB) &&
-      cell_structure.type == CELL_STRUCTURE_DOMDEC && (!cell_structure.use_verlet_list))
-    cells_update_ghosts();
+      cell_structure.type == CELL_STRUCTURE_DOMDEC && (!cell_structure.use_verlet_list) &&
```
fweik: Can you explain to me why this is needed? This seems strange to me as verlet lists have no influence on either VS or LB.

src/core/pressure.cpp
```diff
@@ -123,7 +123,10 @@ void pressure_calc(double *result, double *result_t, double *result_nb, double *
   calc_long_range_virials();
 
 #ifdef VIRTUAL_SITES_RELATIVE  
-  vs_relative_pressure_and_stress_tensor(virials.vs_relative,p_tensor.vs_relative);
+  std::shared_ptr<VirtualSitesRelative> vsr =std::dynamic_pointer_cast<VirtualSitesRelative>(virtual_sites());
```
fweik: I don't like this as it breaks dependency inversion. The client should not have to know about specific implementations of VS. Please add this as virtual function to the VS class, with a default implementation that does either nothing or throws an error/warning as you see fit. While you're on it you can also replace the output params and return by value.

fweik (changes requested):
The overall design looks sensible to me, but some changes are needed to better separate the interface from the implementations. You should also consider moving the classes to their own files, and move them into a subfolder of core. At the end, the core should only use the interface definition, and specific implementations should only referenced in the script bindings.

fweik (approved):

PR #1757, opened 2018-01-10
Drag terminal velocity tests

bogdan-tanygin:
One of tests which validate both Velocity Verlet LD and future BD. Hence, it indirectly relates to https://github.com/espressomd/espresso/issues/1360.

Description of changes:
- Constant force/torque driven viscous drag motion: translational and rotational kinematics validation
- The same set of the possible thermostat parameters' setting test cases is kept
- A performance impact:
`mass-and-rinertia_per_particle ..........   Passed   26.20 sec`
->
`mass-and-rinertia_per_particle ..........   Passed   58.29 sec`
c.f. https://github.com/espressomd/espresso/pull/1622#issuecomment-356453946

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


bogdan-tanygin:
It is already a part of #1842.
I've not closed it only because I was not sure about the future of #1842. If it is already approved for the release 4.1 then we can close this PR.

RudolfWeeber:
Same question here. The Langevin test tests the decay of v/omega for T=0 and the diffusion coefficient.
Could you please clarify what this test adds?

bogdan-tanygin:
It tests a viscous acceleration. It focuses not on the exponential details (already validated by the decay tests), rather that the terminal velocity is stable for a long run. Hence, the BD and LD will match. However, it requires also too much of the test runtime. This is why I've relocated this part to the non-automically run test mass-and-rinertia_per_particle_rotdiff-longrun.py in the #1842.

RudolfWeeber:
On Sun, May 27, 2018 at 03:38:56AM -0700, Bogdan Tanygin wrote:
> It tests a viscous acceleration. It focuses not on the exponential details (already validated by the decay tests), rather that the terminal velocity is stable for a long run. Hence, the BD and LD will match. However, it requires also too much of the test runtime. This is why I've relocated this part to the non-automically run test mass-and-rinertia_per_particle_rotdiff-longrun.py in the #1842.
The runtime issues for this mass-and-rinertia tests will probably go away, once an
upcoming pr disables the short-range loop entirely for cases, where
it is not needed.
So the main issue for the entire mass-and-rinertia test for me is
clarity.

bogdan-tanygin:
> once an upcoming pr disables the short-range loop entirely for cases, where
it is not needed.

@RudolfWeeber thanks, sounds like an important PR which I need to take into account in further development/refactoring. Could you please clarify corresponding PR or the issue number with some extra details? Just to understand the idea..

RudolfWeeber:
On Mon, May 28, 2018 at 12:30:08PM -0700, Bogdan Tanygin wrote:
> > once an upcoming pr disables the short-range loop entirely for cases, where
> it is not needed.
> 
> @RudolfWeeber thanks, sounds like an important PR which I need to take into account in further development/refactoring. Could you please clarify corresponding PR or the issue number with some extra details? Just to understand the idea..
No PR yet. This may or may not make it into 4.0. So, if the short
range cutoff is 0, the short range loop will not
do anything, because, effectively, no short range potential is
active. Still, a pair search is executed. The PR would disable the
short range loop in this case.  This will not result in user-visible
changes (except performance).

This is mostly a corner
case for things such as the thermostat and mass and rinertia tests,
but can also occur for tracer particles in a fluid.

bogdan-tanygin:
This PR depends on #2021.

bogdan-tanygin:
I've ported the corresponding validations to the BD tests within #1842 . Similar checks of the LD viscous properties without a testing performance degradation will be possible after the [above-mentioned PR](https://github.com/espressomd/espresso/pull/1757#issuecomment-392592867). I suggest to close the current PR and to return to it (#2028) later.

PR #1764, opened 2018-01-12
observable: throw if axis not z for cylindrical lb obs.

### LINKED ISSUE #1763: CylindricalLBFluxDensityProfileAtParticlePositions only works with axis = 'z'

KaiSzuttor:
Fixes #1763 

Description of changes:
 - throw error if not supported observable specs are used


hmenke (commented):
Better remove the axis parameter altogether?

KaiSzuttor:
it's just not yet implemented, I'll change the doctoring and mention that it is not yet implemented

fweik:
If only one axis is supported, then the axis should not be a parameter. Also the case where axis is neither of the three is not handled, maybe a string is not the right datatype for this...

PR #1767, opened 2018-01-15
Interface change Reaction Algorithms

jonaslandsgesell:
*added default_charges as mandatory argument.
*added check for electroneutrality of the reaction system. Can be switched off via check_for_electroneutrality=False
*removed polymer configuration changing moves -- better moves would be needed for production in wang-Landau reaction ensemble with energy reweighting enabled in any case. wang-landau reaction ensemble without energy reweighting can make use of MD, which changes the polymer configurations and is therefore not affected by the removal of the code.


jonaslandsgesell:
This can be merged. @fweik

jonaslandsgesell:
@kai the line is deleted because:
```int n_part = max_seen_particle + 1;```
the local variable n_part hides the global variable n_part. This is unwanted since indeed here the number of particles in the system is what is needed.
Since @fweik introduced that ids can be consecutive or not max_seen_particle is not guaranteed to be n_part-1 anymore. 
Imagine a system with the ids {0,1,2,6} where the hole in the range is {3,4,5}.
H5MD now should just output properties of 4 particles. max_seen_particle is however the largest particle_id which in this case would be 6. 
Previously the code tried to write out 7 particles (max_seen_particle+1). This was a good idea as long as the particle id range was not allowed to have holes ({0,1,2,3,4} => n_part=5, max_seen_particle=4 => try to output 4+1=5 particles. With the change now the code only outputs n_part=4 particles.

PR #1769, opened 2018-01-16
Travis optimization

RudolfWeeber:
* Compile with build_type=Debug to have assertions in runs with make check and faster compile times in the other runs
* Combine OpenSuSe run with nocheck-maxset run, to avoid one run
* Reduce duration of dpd and lb (cpu) tests, without touching accuracy targets

In total, this should be build-minute neutral


fweik (approved):

mkuron:
The CUDA/Docker issue is fixed now. There is one CUDA warning with a few thousand lines left that I'll fix, but it doesn't trigger an error. @RudolfWeeber needs to fix two actual issues before this pull request can be merged:
```
src/core/lees_edwards_domain_decomposition.cpp: In function 'void le_dd_prepare_comm(le_dd_comms_manager*, GhostCommunicator*, int)':
src/core/lees_edwards_domain_decomposition.cpp:604:55: error: 'thisCommCount' may be used uninitialized in this function [-Werror=maybe-uninitialized]
           comm->comm[cnt].n_part_lists = thisCommCount;
```
```
FAIL: test_p3m_gpu (__main__.CoulombCloudWallTune)
  File "coulomb_tuning.py", line 92, in test_p3m_gpu
    self.compare("p3m_gpu")
  File "coulomb_tuning.py", line 73, in compare
    method_name)
AssertionError: Asbolute force difference 0.0204302267017 too large for method p3m_gpu
```

RudolfWeeber:
I can reproduce the p3m issue on the icp machines when running with 2 cpu cores.
This seems to be, because the tuning avoids as much cpu short-0range calculation as possible. Due to the slow debug build,
the resulting real space cutoff is 2 whereas it is 4 for the cpu p3m.
The closet particle distance in the system is 0.9.

fweik:
How did you reproduce this? The parameters that are used in the maxset test on gitlab (and fail) do work perfectly on my machine, independent of the number of ranks and with Debug and Release builds. This is strange...

mkuron:
I've seen the p3m test fail independently of this pull request on https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/8700 .

fweik:
Maybe there is some sort of race condition. The error looks suspiciously similar to the other case, despite the parameters being totally different....

maintainer/travis/build_cmake.sh
```diff
@@ -49,7 +49,7 @@ function cmd {
 [ -z "$build_procs" ] && build_procs=2
 [ -z "$make_check" ] && make_check="true"
 
-cmake_params="-DWARNINGS_ARE_ERRORS=ON -DTEST_NP:INT=$check_procs $cmake_params"
+cmake_params="-D CMAKE_BUILD_TYPE=Debug -DWARNINGS_ARE_ERRORS=ON -DTEST_NP:INT=$check_procs $cmake_params"
```
fweik, line 52: If the whitespace between the D and the variable name allowed here?

mkuron: Yes, that is optional.

PR #1771, opened 2018-01-17
fix stress tensor components python

jonaslandsgesell:
Fixes indexing of stress tensor contributions thanks to @fweik.
Introduce function that returns the minimum image distance between two positions. The function which was already there took particle types as argument and is not equivalent therefore. The previous function returns the minimal distance between sets of particles with given types.

### DIFF
--- a/src/python/espressomd/analyze.pyx
+++ b/src/python/espressomd/analyze.pyx
@@ -66,7 +66,7 @@ class Analysis(object):
     #
 
     def min_dist2(self, p1, p2):
-        """Minimal distance between two points p1 and p2.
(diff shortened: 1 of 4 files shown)


src/python/espressomd/analyze.pyx
```diff
@@ -65,6 +65,21 @@ class Analysis(object):
     # Minimal distance between particles
     #
 
+    def min_dist2(self, p1, p2):
+        """Minimal distance between two points p1 and p2.
```
KaiSzuttor: maybe three dimensional coordinates

fweik (approved):

fweik (approved):

PR #1775, opened 2018-01-22
Gitlab osx

KaiSzuttor:
adding osx job to gitlab


mkuron (approved):
Looks good. Can you also move over the Python3 Mac build and create a CUDA build (though with `make_check=false` because we can't use GPUs in Virtualbox)?

hmenke:
I think the Travis OS X build should not be removed.  If it was removed, OS X wouldn't be tested for people who are not espressomd members.

mkuron:
Travis should be removed altogether as it has been quite unreliable in recent weeks. Mac builds still have like a six-hour waiting time and Linux was flaky too on occasion. We’re currently considering dedicated build machines with appropriate firewall rules so we can open up builds to everyone.

fweik (approved):

KaiSzuttor:
We wait with merging until I added the tests @mkuron suggested.

mkuron (approved):

fweik:
There is sth messed up with the compiler options for NVCC...

KaiSzuttor:
under investigation... can also see the Red Cross

CMakeLists.txt
```diff
@@ -171,10 +171,6 @@ if (WITH_CUDA)
         list(APPEND CUDA_NVCC_FLAGS "-D_MWAITXINTRIN_H_INCLUDED -D_FORCE_INLINES")
       endif()
       SET(CUDA_PROPAGATE_HOST_FLAGS OFF)
-
-      if (APPLE)
-        set(CUDA_NVCC_FLAGS "${CUDA_NVCC_FLAGS} -Xcompiler -isysroot -Xcompiler ${CMAKE_OSX_SYSROOT}")
-      endif()
```
mkuron: Don’t remove this. The sysroot is needed in certain cases, e.g. with Xcode 9 on macOS 10.12.

mkuron (changes requested):

CMakeLists.txt
KaiSzuttor: okay, but it breaks the build for the current clang version. How to make this future save?

mkuron: I wasn't able to reproduce your issue on a real Mac with the current python branch, but I came up with a simple fix that checks whether `CMAKE_OSX_SYSROOT` is even set. Just merge https://github.com/KaiSzuttor/espresso/pull/17

maintainer/travis/build_cmake.sh
```diff
@@ -96,12 +98,10 @@ if ! $insource; then
     if [ ! -d $builddir ]; then
         echo "Creating $builddir..."
         mkdir -p $builddir
+        cd $builddir
```
mkuron: Move this out of the `if [ ! -d $builddir ]`, otherwise the script cannot be safely rerun. It doesn't matter for the CI builds, but if somebody tries to run this script locally, it will fail on the second run.

mkuron (changes requested):

mkuron (approved):

KaiSzuttor:
~~cant see why centos build fails, works locally~~

KaiSzuttor:
we merge after we have installed dedicated test machines

PR #1780, opened 2018-01-24
LB: make LB boundaries require the respective CPU/GPU feature.

KaiSzuttor:


fweik:
Is it really true that the LB boundaries need CONSTRAINTS? As far as I can tell they use the shapes directly. Please check and if so remove the requirement.

PR #1781, opened 2018-01-24
Manual trigger for external PRs

KaiSzuttor:
Added a manual job for external PR's (and will adapt the webhook endpoint).


mkuron:
Using stages for this is super ugly. Once we have the dedicated build machines and firewall rules in place, none of this will be needed anymore. Other than that, I think your solution is fine, but the message you `echo` is printed after the job is manually started, so it is kind of pointless; just replace it with `exit 0`.

PR #1788, opened 2018-01-29
core: Made calc_non_bonded_pair_energy const-correct.

fweik:
Arguably the same should be done for the forces, making the forces on the particles `mutable`.


RudolfWeeber:
I think, you lost some multi timestep and possibly npt code

fweik:
Oh, that was not intended... (Really!)

fweik:
Should be good now...

RudolfWeeber:
Can be merged

PR #1791, opened 2018-01-30
Travis to gitlab

KaiSzuttor:
Description of changes:
 -  moved testjobs from travis to gitlab
 - removed all travis related files


KaiSzuttor:
successor of #1775

RudolfWeeber:
On Tue, Jan 30, 2018 at 04:04:52PM +0000, Kai Szuttor wrote:
> Description of changes:
>  -  moved testjobs from travis to gitlab
>  - removed all travis related files
Can we please keep the std, maxset (py2), and maxset (py3) Travis checks, so people can use it with ther own branches on github?

KaiSzuttor:
ok, i'll add a travis configuration file. We can then just deactivate the travis build for the espressomd/espresso repository on travis.

mkuron:
Actually, just leave Travis enabled on the repo. With only three Linux builds, the wait times should be low enough (at least when Travis is not broken). That way, we'll see if anything breaks on Travis that doesn't break on Gitlab.

KaiSzuttor:
That is not necessary since the builds are exactly the same. If we decide we want to be independent of travis, we should also don't use it in the main repo anymore.

mkuron:
The environments are not the same, Docker versions are different, Gitlab has some caching that Travis doesn't, etc.. We still need to ensure that the Travis builds remain working, so we should continue testing them.

We are independent as long as we have Gitlab with the same builds. If Travis goes down, we can just switch it off.

KaiSzuttor:
You are right, but we don't want to test for different docker version, nor we want to test gitlab. We decided not to use travis anymore because it is unreliable, so there is no point in keeping the builds for every PR. Maybe we should use a daily cron job on travis (https://docs.travis-ci.com/user/cron-jobs/).

RudolfWeeber:
If there are only one std and two maxset builds with tetts, travis uilds will not be slower than gitlab, probably. As it costs us nothing, I'd leave it on.

mkuron (approved):

PR #1794, opened 2018-01-30
core: Removed pair-exclusion rule from dpd.

fweik:


KaiSzuttor:
how do we deal with dpd interactions across the LE boundary condition then?

fweik:
It's now at least less wrong than before.

KaiSzuttor:
you mean, lees false 🤪

KaiSzuttor (approved):

PR #1800, opened 2018-02-01
Get rid of SQR macro and SQR definition outside Utils namespace.

pkreissl:
There were two identical Template definitions for square calculation (`sqr` and `SQR`) in `src/core/utils/math/sqr.hpp`---one within, the other outside the `Utils` namespace.
This PR gets rid of the the latter (`SQR`) using `Utils::sqr()` instead.
`SQR` was also defined as macro several times. These were also replaced.

### DIFF
--- a/src/core/actor/EwaldGPU.cpp
+++ b/src/core/actor/EwaldGPU.cpp
@@ -2,16 +2,16 @@
 
 #ifdef EWALD_GPU
 
-#include <iostream>
 #include <algorithm>
+#include <iostream>
 #include <vector>
 
 #include "cells.hpp"
 #include "grid.hpp"
 #include "integrate.hpp"
 #include "interaction_data.hpp"
-#include "tuning.hpp"
 #include "partCfg_global.hpp"
+#include "tuning.hpp"
 
 #define DEBUG(A) std::cout << #A << ": " << (A) << std::endl;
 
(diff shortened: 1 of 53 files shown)


jonaslandsgesell (approved):

src/core/actor/EwaldGPU.cpp
```diff
@@ -28,12 +28,13 @@ void EwaldgpuForce::compute_num_k() {
         if (ix * (2 * m_num_ky + 1) * (2 * m_num_kz + 1) +
                     iy * (2 * m_num_kz + 1) + iz >=
                 0 // Half m_k-space
-            and
-            SQR(ix) * SQR(m_num_ky) * SQR(m_num_kz) +
-                    SQR(iy) * SQR(m_num_kx) * SQR(m_num_kz) +
-                    SQR(iz) * SQR(m_num_kx) * SQR(m_num_ky) <=
-                SQR(m_num_kx) * SQR(m_num_ky) *
-                    SQR(m_num_kz)) // m_k-space ellipsoid
+            and Utils::sqr(ix) * Utils::sqr(m_num_ky) * Utils::sqr(m_num_kz) +
```
KaiSzuttor: maybe here: ```using Utils::sqr```

KaiSzuttor (approved):

src/core/actor/EwaldGPU.cpp
pkreissl: Agreed. Added this in the two methods with lots of `sqr`.

Issue #1802, opened 2018-02-01
Doc part.rst


RudolfWeeber:
* grand canonical (incompatibility with collision detection)
* Mention criteria based particle selection (ParticleList.select()). See docstring

jonaslandsgesell:
Can be closed

PR #1826, opened 2018-02-02
Electroneutrality

### LINKED ISSUE #1786: Suggestion for Default Behaviour: Throw an exception for non-neutral systems during tuning of electrostatic methods unless the user explicitly states differently
Typically simulations of charge are carried out such that the system is neutral within some limit (e.g. 1e-8 charges).
Sometimes the user fails in setting up the system  as charge neutral. Electrostatic methods like P3M can be used in a non-neutral system since they automatically introduce a background charge which neutralizes the system. However I assume that this default behaviour can lead to unexpected behaviour.

I suggest to throw an exception in the tuning routine of the electrostatics methods unless the user specifies a flag "simulate_non_neutral_system" that states e.g.
"The system is non-neutral. Please neutralize your system or explcitly state that you want to simulate a non neutral system via setting the flag simulate_non_neutral_system in the tuning routine."

jonaslandsgesell:
Fixes #1786 and #1802


fweik (approved):

src/python/espressomd/electrostatic_extensions.pyx
```diff
@@ -27,7 +27,13 @@ from espressomd.utils cimport handle_errors
 
 IF ELECTROSTATICS and P3M:
     cdef class ElectrostaticExtensions(actors.Actor):
-        pass
+        def _check_for_electroneutrality(self):
```
KaiSzuttor: please autopep8 this file

src/python/espressomd/electrostatics.pyx
```diff
@@ -38,7 +38,15 @@ IF ELECTROSTATICS == 1:
         def _set_params_in_es_core(self):
             raise Exception(
                 "Subclasses of ElectrostaticInteraction must define the _set_params_in_es_core() method.")
-    
+
+        def _check_for_electroneutrality(self):
```
KaiSzuttor: this function is duplicate to the one in electrostatic_extensions

jonaslandsgesell: Autopep8 will be applied anyway with the release of 4.0 right?

src/python/espressomd/electrostatic_extensions.pyx
jonaslandsgesell: autopep8 will be applied with the 4.0 release

fweik:
The parameter name seems to be awfully long, how about "check_neutrality"? After all it's on a electrostatics method, so it is pretty clear this is about electric charge, and the `for` adds nothing but five characters.
Why are the mmm?d methods not checked by default, but p3m is?

jonaslandsgesell:
> Why are the mmm?d methods not checked by default, but p3m is?

I don't see what you mean. Can you please specify the problem? Using e.g.

```
    def setUp(self):
        if len(self.system.part) == 0 :
            self.system.periodicity=[0,0,1]
            self.system.cell_system.set_n_square()
            self.system.box_l = 10, 10, 10
            self.system.part.add(id=0, pos = [0, 0, 0])
            self.system.part.add(id=1, pos = [0.1, 0.1, 0.1])
            self.system.part[0].q = 1
            self.system.part[1].q = 0
```
as a non-neutral setup in testsuite/mmm1d.py an exception is thrown as expected since the function ``_check_for_electroneutrality(system, _params)`` is called in the method _activate_method(). I also checked that for the mmm2d setup in elc_vs_mmm2d_neutral.py if you provide a non-neutral setup.

fweik:
Ah sorry you're right, I misread that.

PR #1836, opened 2018-02-07
Python-shifts

CompRhys:
Fixes #

Description of changes:
 - 
added potential shifts to Smooth Step, Soft Sphere and Gaussian non-bonded interactions. 

had to define the value of 'e', this was taken as 2.718281828459045 from https://oeis.org/A001113.

PR Checklist
------------
 - [x] Tests? - no tests need to be updated 
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs? - documentation will need to be amended to state that potential shifts have been added to the Smooth Step, Soft Sphere and Gaussian non bonded interactions. I couldn't find where to do this.

### DIFF
--- a/src/core/gaussian.cpp
+++ b/src/core/gaussian.cpp
@@ -28,7 +28,8 @@
 #ifdef GAUSSIAN
 
 int gaussian_set_params(int part_type_a, int part_type_b,
-			double eps, double sig, double cut)
+			double eps, double sig, double cut,
+		        double shift)
 {
   IA_parameters *data = get_ia_param_safe(part_type_a, part_type_b);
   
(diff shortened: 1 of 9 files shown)


fweik:
Euler's constant is defined in python in `math.e`

pkreissl:
There exists a python test for non-bonded interactions (`testsuite/interactions_non-bonded.py`). It checks the energy and force (between ~to~ two particles) calculated by the ESPResSo-core against the analytical expressions for all non-bonded interactions.

All checks have passed successfully. However, that only indicates that your adding a shift for the Smooth Step, Soft Sphere, and Gaussian interactions does not change their behavior when not taking the shift into account.

Your new feature still needs to be tested!
So please update the analytical expressions given in `testsuite/tests_common.py` to take the shift correctly into account, and add a non-zero shift in the respective test-methods in `testsuite/interactions_non-bonded.py`.

pkreissl:
Concerning the documentation: for our sphinx documentation, the file containing information about non-bonded interactions is `doc/sphinx/inter_non-bonded.rst`.

PR #1841, opened 2018-02-09
Samples tested and modified wherever necessary

ashreyaj:


samples/debye_hueckel.py
```diff
@@ -18,7 +18,7 @@
 #
 from __future__ import print_function
 import numpy
-import espressomd._system as es
+import espressomd.system as es
```
pkreissl: `es.*` is not used anywhere in that script. It only is part of outdated comments.
So please just delete this line. Also the commented lines containing the outdated commands.

pkreissl (changes requested):
Running the `samples/constraints.py` script fails every third time or so. Error is something like:
```
ERROR: particle coordinate out of range, pos = 6.75621e+13, image box = -2124923835
ERROR: Constraint violated by particle 29 dist -0.1
```
Please check that.

PR #1842, opened 2018-02-10
Brownian Dynamics within the existing VV loop

### LINKED ISSUE #1360: Do we need a Brownian Dynamics (BD) integrator ?

bogdan-tanygin:
Fixes #1360, also it contains the integrated changes from PRs #1622, #1757 which were required to validate both existing Langevin- (LD) and new Brownian Dynamics (BD) viscous and fluctuation dynamics including the rotational one.

Description of changes:
- Generally, this is a conventional simplest BD [schlick2010].
- Corresponding Sphinx documentation section is populated with the required details to reveal the concept.
- A velocity, even though it is optional for a simplest implementation of the BD, has been added for a consistency with the existing VV loop (c.f. https://github.com/espressomd/espresso/issues/1360#issuecomment-326118840) and all the existing tests and new tests. It does not impact the positional mechanics though which is natural for the BD [schlick2010].
- The velocity and position random walk has been added according to the classical diffusion equations [schlick2010,Pottier2014].
- Rotational motion BD is implemented analogously. Existing quaternion approach has been reused where possible.
- New test `brownian_thermostat.py` is based on the existing `langevin_thermostat.py`. It contains some commented out fragments which are inconsistent for the simplest BD cause ones correspond to more fine dynamical structure compare to the BD temporal relation `time_step >> mass / gamma`. The Maxwell distribution test is passing well and the Gaussian noise type is crucial for this which has been done in the `thermostat.hpp`.
- The `mass-and-rinertia_per_particle.py` contains changes described within the PRs #1622, #1757. Probably, it is of interest for both the LD and BD independently. This is why I've opened that PRs in advance.
- Existing automatic tests execution timeline is minimized.
- Additional "heavy" test has been added without an automatic run: `mass-and-rinertia_per_particle_rotdiff-longrun.py`. It requires ~2 hours of a run and provides the intensive validation of the existing LD and new BD dynamics in the same way.

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?

_References_
schlick2010: https://link.springer.com/book/10.1007%2F978-1-4419-6351-2
Pottier2014: https://link.springer.com/article/10.1007/s10955-010-0114-6


hmenke:
I'll review this at the end of next week.

doc/sphinx/system_setup.rst
```diff
@@ -421,6 +421,69 @@ Be aware that this feature is neither properly examined for all systems
...
+In order to activate the Brownian thermostat the member function
+:py:attr:`~espressomd.thermostat.Thermostat.set_brownian` of the thermostat
+class :class:`espressomd.thermostat.Thermostat` has to be invoked.
+Best explained in an example:::
+    
+    import espressomd
+    system = espressomd.System()
+    therm  = system.Thermostat()
+
+    therm.set_brownian(kT=1.0, gamma=1.0)
```
hmenke: The `System` class has a built-in thermostat handle.
https://github.com/espressomd/espresso/blob/ac9cdfe4f84d83034c13ad4d2147de6734e3b228/src/python/espressomd/system.pyx#L119
Use instead
```python
import espressomd
system = espressomd.System()
system.thermostat.set_brownian(kT=1.0, gamma=1.0)
```

doc/sphinx/system_setup.rst
```diff
@@ -421,6 +421,69 @@ Be aware that this feature is neither properly examined for all systems
...
+In terms of the Python interface and setup, Brownian thermostat derives most 
+properties of the :ref:`Langevin thermostat`. Same feature
+``LANGEVIN_PER_PARTICLE`` is using to be able to control the per-particle
+temperature and the friction coefficient setup. Major differences are
+its internal integrator implementation and other temporal constraints.
+The integrator is still a symplectic Velocity Verlet-like integrator.
+It is implemented via a viscous drag part and a random walk of both the position and
+velocity. Due to a nature of the Brownian Dynamics method, ``time_step``
+should be large enough compare to the relaxation time
+:math:`\text{MASS}/\text{gamma}`. This requirement is just a conceptual one
```
hmenke:
```rst
:math:`m/\gamma` where :math:`m` is the particle mass.
```
Please also introduce sensible variables in the equations further below, e.g. `\delta t` for `time_step`.

.travis.yml
```diff
@@ -10,10 +10,18 @@ matrix:
       sudo: required
       services: docker
       env: myconfig=maxset with_coverage=true
+    - os: linux
+      sudo: required
+      services: docker
+      env: myconfig=maxset-bd with_coverage=true
```
hmenke: Does your implementation of BD change the integrator even when the thermostat is disabled?  If not, then why do we need another configuration?  Just add BD to the regular `maxset.hpp`.

src/core/integrate.cpp
```diff
@@ -135,6 +135,20 @@ void force_and_velocity_display();
...
+
+/** Propagate position: viscous drag driven by conservative forces.*/
+void bd_drag(Particle *p, double dt);
+/** Set the terminal velocity driven by the conservative forces drag.*/
+void bd_drag_vel(Particle *p, double dt);
+
+/** Propagate position: random walk part.*/
+void bd_random_walk(Particle *p, double dt);
+/** Thermalize velocity: random walk part.*/
+void bd_random_walk_vel(Particle *p, double dt);
```
hmenke: Take `Particle &p`.  Value semantics are easier to read than pointers.

src/core/integrate.cpp
```diff
@@ -1230,3 +1317,190 @@ int integrate_set_npt_isotropic(double ext_pressure, double piston, int xdir,
   mpi_bcast_nptiso_geom();
   return (ES_OK);
 }
+
+#ifdef BROWNIAN_DYNAMICS
+
+/** Propagate position: viscous drag driven by conservative forces.*/
+void bd_drag(Particle *p, double dt) {
```
hmenke: This function is very performance critical.  Mark it as `inline`.

(Applies to the other functions as well.)

src/core/integrate.cpp
```diff
@@ -1230,3 +1317,190 @@ int integrate_set_npt_isotropic(double ext_pressure, double piston, int xdir,
...
+  int j;
+  double scale_f;
+  Thermostat::GammaType local_gamma;
+
+  scale_f = 0.5 * time_step * time_step / p->p.mass;
+
+  if(p->p.gamma >= Thermostat::GammaType{}) local_gamma = p->p.gamma;
+  else local_gamma = langevin_gamma;
```
hmenke: Horrible formatting.  Use something like
```cpp
if(p->p.gamma >= Thermostat::GammaType{}) {
  local_gamma = p->p.gamma;
} else {
  local_gamma = langevin_gamma;
}
```

(Applies to the other functions as well.)

src/core/integrate.cpp
```diff
@@ -1230,3 +1317,190 @@ int integrate_set_npt_isotropic(double ext_pressure, double piston, int xdir,
...
+
+  for (j = 0; j < 3; j++) {
```
hmenke: [ES.6: Declare names in for-statement initializers and conditions to limit scope](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-cond)

(Applies to the other functions as well.)

src/core/integrate.cpp
```diff
@@ -1230,3 +1317,190 @@ int integrate_set_npt_isotropic(double ext_pressure, double piston, int xdir,
...
+
+  if(p->p.gamma >= Thermostat::GammaType{})
+  {
+    // Is a particle-specific temperature also specified?
+    if(p->p.T >= 0.)
+    {
+      if (p->p.T > 0.0)
+        brown_sigma_pos_temp_inv = sqrt(p->p.gamma / (langevin_temp_coeff * p->p.T));
+      else
+        brown_sigma_pos_temp_inv = -1.0 * sqrt(p->p.gamma); // just an indication of the infinity; negative sign has no sense here
```
hmenke: A temperature less than zero does not make sense anyway, so you could just set
```cpp
brown_sigma_pos_temp_inv = NAN;
```
Signalling `NAN` will also speed up all subsequent “calculations” and you can fail earlier.

(Also applies below.)

src/core/integrate.cpp
```diff
@@ -1230,3 +1317,190 @@ int integrate_set_npt_isotropic(double ext_pressure, double piston, int xdir,
...
+
+/** Propagate the positions: random walk part.*/
+void bd_random_walk(Particle *p, double dt) {
+  int j;
+  extern Thermostat::GammaType brown_sigma_pos_inv;
+  Thermostat::GammaType brown_sigma_pos_temp_inv;
+#ifdef PARTICLE_ANISOTROPY
+  double delta_pos_body[3] = { 0.0, 0.0, 0.0 }, delta_pos_lab[3] = { 0.0, 0.0, 0.0 };
+#endif
+  int aniso_flag = 1; // particle anisotropy flag
```
hmenke: `bool aniso_flag = false;`  That's only one bit instead of eight.

src/core/integrate.cpp
```diff
@@ -1230,3 +1317,190 @@ int integrate_set_npt_isotropic(double ext_pressure, double piston, int xdir,
...
+    for (j = 0; j < 3; j++) {
+#ifdef EXTERNAL_FORCES
+      if (!(p->p.ext_flag & COORD_FIXED(j)))
+#endif
+      {
+        p->r.p[j] += delta_pos_body[j];
+      }
+    }
+  }
+}
```
hmenke: Lots of duplication.  Why not invert control flow?
```cpp
if (aniso_flag) {
    convert_vec_body_to_space(p, delta_pos_body, delta_pos_lab);
}

for (int j = 0; j < 3; ++j) {
#ifdef EXTERNAL_FORCES
    if (!(p->p.ext_flag & COORD_FIXED(j)))
#endif
    {
        p->r.p[j] += aniso_flag ? delta_pos_lab[j] : delta_pos_body[j];
    }
}
```
The compiler will notice that `aniso_flag` is not touched in the loop and the branch predictor will hence discard the branch not taken.

src/core/integrate.cpp
```diff
@@ -1230,3 +1317,190 @@ int integrate_set_npt_isotropic(double ext_pressure, double piston, int xdir,
...
+    if (!(p->p.ext_flag & COORD_FIXED(j)))
+#endif
+    {
+      // velocity is added here. It is already initialized in the terminal drag part.
+      p->m.v[j] += brown_sigma_vel_temp * Thermostat::noise_g() / sqrt(p->p.mass);
+    }
+  }
+}
+
+#endif // BROWNIAN_DYNAMICS
```
hmenke: The implementation looks safe and sound from a programming point of view.  Could you please add bibliography references to the code for the individual equations similar to how it is done in LB?
https://github.com/espressomd/espresso/blob/ac9cdfe4f84d83034c13ad4d2147de6734e3b228/src/core/lbgpu_cuda.cu#L1546-L1566
It is much easier to debug when all the factors are annotated with references, so you can check.

src/core/rotation.cpp
```diff
@@ -499,6 +507,15 @@ void convert_vec_space_to_body(Particle *p, double *v, double *res) {
   res[2] = A[2 + 3 * 0] * v[0] + A[2 + 3 * 1] * v[1] + A[2 + 3 * 2] * v[2];
 }
 
+void convert_vec_body_to_space(Particle *p, double *v,double* res)
```
hmenke: Add `const`
```cpp
void convert_vec_body_to_space(Particle *p, double const *v,double* res)
```

src/core/rotation.cpp
```diff
@@ -550,4 +567,269 @@ void rotate_particle(Particle *p, double *aSpaceFrame, double phi) {
...
+    a[i]/=l;
+
+#endif
+
+  double q[4];
+  q[0]=cos(phi/2);
+  double tmp=sin(phi/2);
+  q[1]=tmp*a[0];
+  q[2]=tmp*a[1];
+  q[3]=tmp*a[2];
```
hmenke: This unreadable version has zero performance advantage in the presence of optimizations.  Just use
```cpp
double q[] = {
    cos(phi/2),
    sin(phi/2)*a[0],
    sin(phi/2)*a[1],
    sin(phi/2)*a[2]
};
```
This produces identical assembly, see [Compiler Explorer](https://godbolt.org/g/asJvpS).

src/core/rotation.cpp
```diff
@@ -550,4 +567,269 @@ void rotate_particle(Particle *p, double *aSpaceFrame, double phi) {
...
+  // When dipoles are enabled, update dipole moment
+  convert_quatu_to_dip(p->r.quatu, p->p.dipm, p->r.dip);
+#endif
+}
+
+/** Rotate the particle p around the j-th body axis by amount phi */
+void rotate_particle_body_j(Particle* p, int j, double phi)
+{
+  double u_dphi[3] = {0.0, 0.0, 0.0};
+  if (phi) {
```
hmenke: Did you mean
```cpp
if (phi != 0.0) {
```

src/python/espressomd/thermostat.pyx
```diff
@@ -306,13 +309,13 @@ cdef class Thermostat(object):
...
-        global thermo_switch
-        thermo_switch = (thermo_switch | THERMO_LANGEVIN)
-        mpi_bcast_parameter(FIELD_THERMO_SWITCH)
         mpi_bcast_parameter(FIELD_TEMPERATURE)
         mpi_bcast_parameter(FIELD_LANGEVIN_GAMMA)
         IF ROTATION:
             mpi_bcast_parameter(FIELD_LANGEVIN_GAMMA_ROTATION)
+        global thermo_switch
+        thermo_switch = (thermo_switch | THERMO_LANGEVIN)
+        mpi_bcast_parameter(FIELD_THERMO_SWITCH)
```
hmenke, line 540: Pointless change

src/core/integrate.hpp
```diff
@@ -114,5 +114,4 @@ int python_integrate(int n_steps, bool recalc_forces, bool reuse_forces);
 void integrate_set_nvt();
 int integrate_set_npt_isotropic(double ext_pressure, double piston, int xdir, int ydir, int zdir, bool cubic_box);
 
-
```
hmenke: Pointless change

src/core/thermostat.cpp
```diff
@@ -48,14 +48,20 @@ Vector3d sentinel(Vector3d) { return {-1.0, -1.0, -1.0}; }
 }
 
 /* LANGEVIN THERMOSTAT */
-
```
hmenke: Pointless change

testsuite/CMakeLists.txt
```diff
@@ -78,6 +78,7 @@ python_test(FILE ek_eof_one_species_y_nonlinear.py MAX_NUM_PROC 1)
 python_test(FILE ek_eof_one_species_z_nonlinear.py MAX_NUM_PROC 1)
 python_test(FILE exclusions.py MAX_NUM_PROC 2)
 python_test(FILE langevin_thermostat.py MAX_NUM_PROC 1)
+python_test(FILE brownian_thermostat.py MAX_NUM_PROC 1)
```
hmenke: Why only single core?

testsuite/brownian_thermostat.py
```diff
@@ -0,0 +1,413 @@
...
+#                if espressomd.has_features("PARTICLE_ANISOTROPY"):
+#                    self.assertAlmostEqual(s.part[0].omega_body[j],o0*np.exp(-gamma_r_a[j]/s.part[0].rinertia[j]*s.time),places=2)
+#                else:
+#                    self.assertAlmostEqual(s.part[0].omega_body[j],o0*np.exp(-gamma_r_i/s.part[0].rinertia[j]*s.time),places=2)
+
+
+
+
+if __name__ == "__main__":
+    ut.main()
```
hmenke: You have mixed indentation (4 spaces and 3 spaces).  Please make it 4 spaces throughout.

testsuite/python/CMakeLists.txt
```diff
@@ -0,0 +1,77 @@
+if(NOT DEFINED TEST_NP)
```
hmenke: This file has been removed.

hmenke (changes requested):
Overall great work!  Only some minor tweaks.
I see, there are some TODOs in the test cases.  I guess you want to complete those before a merge can be considered.

bogdan-tanygin:
@hmenke thanks for the useful review! I'm finalizing the changes accordingly.

bogdan-tanygin:
@hmenke 
> I see, there are some TODOs in the test cases. I guess you want to complete those before a merge can be considered.

I suppose you mean the test `brownian_thermostat.py`. Actually, these fragments are important, however, they can be passed only within the modified _inertial BD_ (IBD, schlick2010) with more complex extra implementations. I'm planning to prepare corresponding pull requests later. The present PR is dedicated to the simplest BD adapted only to the VV loop. I'm keeping these test fragments as is because they takes place for the Langevin Dynamics similar test. Hence, I'm trying to keep the BD to be fully functional as LD, at least, potentially.

doc/sphinx/system_setup.rst
bogdan-tanygin: Agree / done.

doc/sphinx/system_setup.rst
bogdan-tanygin: Agree / done.

.travis.yml
bogdan-tanygin: Agree / done.

src/core/integrate.cpp
bogdan-tanygin: Agree / done.

src/core/integrate.cpp
bogdan-tanygin: Agree. However, I've done these for some `bd_` functions. Others: I've kept in a previous way, cause their inline change will introduce transforming a lot of other functions to inline and also the linking issue (multiple definitions).

src/core/integrate.cpp
bogdan-tanygin: Agree / done.

src/core/integrate.cpp
bogdan-tanygin: Thanks, agree / done.

src/core/integrate.cpp
bogdan-tanygin: Thanks, agree / done.

src/core/integrate.cpp
bogdan-tanygin: Agree / done.

src/core/integrate.cpp
bogdan-tanygin: Agree / done.

src/core/integrate.cpp
bogdan-tanygin: Makes sense / done.

src/core/rotation.cpp
bogdan-tanygin: Done.

src/core/rotation.cpp
bogdan-tanygin: Agree / done.

src/core/rotation.cpp
bogdan-tanygin: Yes, I did. Done.

src/python/espressomd/thermostat.pyx
bogdan-tanygin: @hmenke could you please clarify this. Maybe I've missed how correctly `thermo_switch` should set for the given thermostat within the latest architecture. Similar code snippets exist for other thermostat types in the same file.

src/core/integrate.hpp
bogdan-tanygin: Agree / I've reverted this.

src/core/thermostat.cpp
bogdan-tanygin: Agree / I've reverted this.

testsuite/CMakeLists.txt
bogdan-tanygin: Agree, increased to 4.

testsuite/brownian_thermostat.py
bogdan-tanygin: Agree, I've fixed this after the pep8 check.

testsuite/python/CMakeLists.txt
bogdan-tanygin: Thanks, I've cleaned up this.

bogdan-tanygin:
Almost all requested changes has been implemented. See comments in the review.

Latest sphinx gitlab CI issue, probably, is not related to my changes and will be fixed after merge of the Espresso head (already made): https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/pipelines/2107

At least, sphinx docs are building locally fine.

Concerning https://travis-ci.org/espressomd/espresso/jobs/352620609
<- it was well few minutes before; the merge would not impact this. Probably, the Travis interim performance issue. I'll take a look tomorrow.

src/python/espressomd/thermostat.pyx
hmenke: Well, it looks like you just moved those three lines a little down.

src/core/thermostat.cpp
```diff
@@ -46,6 +46,10 @@ namespace {
      set yet. */
 constexpr double sentinel(double) { return -1.0; }
 Vector3d sentinel(Vector3d) { return {-1.0, -1.0, -1.0}; }
+#ifdef BROWNIAN_DYNAMICS
+constexpr double set_nan(double) { return NAN; }
+Vector3d set_nan(Vector3d) { return {NAN, NAN, NAN}; }
+#endif // BROWNIAN_DYNAMICS
```
hmenke: No need to `#ifdef` function declarations.  These will not have any runtime overhead.

src/python/espressomd/thermostat.pyx
bogdan-tanygin: Oops, reverted.

src/core/thermostat.cpp
bogdan-tanygin: Agree, thanks.

bogdan-tanygin:
I'm fixing the performance issue of the modified here test `mass-and-rinertia_per_particle.py` in case of the `espresso-ubuntu-python3` docker only.

bogdan-tanygin:
Please trigger the GitLab CI at your convenience (not urgent).

bogdan-tanygin:
I've completed an implementation of changes according to each requested point.

bogdan-tanygin:
Probably, the GitLab failure does not relate to the last commit 2744188. Changes in the sphinx documentation cannot impact an accuracy of the ek_eof_one_species_y_nonlinear test... Also, the BD implementation should not produce this. Actually, I've already seen this failure few times, it seems to be a probabilistic one with a low probability.

Please restart the GitLab test.

bogdan-tanygin:
The issue #1917 which I've referenced above is, probably, not a defect, rather a conceptual task which could to lead to new tweaks to the BD feature. Current implementation has enough parameters for a moment of inertia and viscous friction tensors. Hence, the BD feature could be already used properly in terms of a theoretical correctness: isotropic or almost isotropic particles. I will analyze and perform additional bibliographic research concerning the BD of anisotropic particles. If a functional design will be non-straightforward then I will email espressomd-devel beforehand. I'm still not fully sure that that task is doable in a general case.

I suggest to separate the enhancement #1917 to other PR and to merge the current one; however, it is up to you to decide.

bogdan-tanygin:
@hmenke
I've investigated the issue #1917 using the ESPResSo VV LD as a reference to validate the BD viscous rotation. Results are following. Indeed, in case of rinertia[0]!=rinertia[1] or rinertia[1]!=rinertia[2], the viscous rotation is not stable:
direction of both the lab angular momentum and the lab angular velocity are oscillating over the time.
Hence, we have the "viscous precession", there is no terminal angular velocity driven (dragged) by the external torque in this case at all. I have doubts that there is some closed-form solution for this motion: the system of 3 strongly nonlinear differential equations. Hence, probably, the rotational BD is not consistent for anisotropic particles.

Oppositely, when I set the isotropic particle, according to the VV LD simulation, the terminal lab angular velocity was constant under the rotational viscout drag. Hence, all this situation just requires a proper documentation: I've added corresponding fragment to the sphinx doc. Also, corresponding testing has been originally placed to the mass-and-rinertia_per_particle.py:
see a part starting from the `if test_case >= 4 + self.rot_flag:`...
Note, there is also test mass-and-rinertia_per_particle_rotdiff-longrun.py which is not running automatically. It contains a comparison of the LD and BD simulation in the same code fragment which making sure that the ESPResSo is self-consistent and matches the theory as well:

https://gist.github.com/psci2195/1add4a7c680291443678d2b4eec08b36

It also validates the rotational diffusion variance.

Probably, I've completed the work in this PR. Please review, when you will have time.

hmenke:
Thanks for your efforts.  I will review it again, hopefully soon.  However, I doubt that it will be merged before the 4.0 release (as indicated by the 4.1 milestone).

bogdan-tanygin:
Understood, thanks. Sure then.

testsuite/mass-and-rinertia_per_particle.py
```diff
@@ -14,11 +14,55 @@
 class ThermoTest(ut.TestCase):
     longMessage = True
     # Handle for espresso system
-    es = espressomd.System(box_l=[1.0, 1.0, 1.0])
+    es = espressomd.System(box_l=[1.0E5, 1.0E5, 1.0E5])
```
hmenke: Wow!  That is a *massive* box.  Does it really have to be that big?

hmenke (approved):

testsuite/mass-and-rinertia_per_particle.py
bogdan-tanygin: @hmenke thanks for catching this, I will revert this fragment. Such box has been needed in a previous release of this test when I've kept a viscous terminal velocity and corresponding position validation for both BD (big time_step) and LD (small time_step). In a latter case we need a lot of steps/time/space to reach the terminal velocity depending on visous coefficients and mass. In a former case it is reaching by a single step by a nature of the method which obviously has a physical consistency only at time_step >> mass/gamma. Finally, I've removed this part of test for the LD due to too much a temporal impact on overall the ESPResSo build.

However, this test case is kept full in mass-and-rinertia_per_particle_rotdiff-longrun.py which is not running automatically. Hence, all aspects could be validated there upon a specific need.

bogdan-tanygin:
@hmenke thanks for approving the PR. I suppose that I need to fix the conflicting changes only after the 4.0.x releases.

hmenke:
It is definitely required that the conflicts are removed before this can be merged.  This will not happen before the 4.0 release, so until then it is up to you when to remove the conflicts.  My personal advice is to remove conflicts as soon as you become aware of them, because as they pile up it becomes harder and harder to fix them.

src/core/brownian_inline.hpp
```diff
@@ -0,0 +1,146 @@
...
+
+  if (p.p.gamma >= Thermostat::GammaType{}) {
+    local_gamma = p.p.gamma;
+  } else {
+    local_gamma = langevin_gamma;
+  }
+
+  for (int j = 0; j < 3; j++) {
+#ifdef EXTERNAL_FORCES
+    if (!(p.p.ext_flag & COORD_FIXED(j)))
```
hmenke: I realised that you also have to check whether the particle is virtual.

hmenke (dismissed):

src/core/brownian_inline.hpp
bogdan-tanygin: thanks, I agree. Working on that.

bogdan-tanygin: @hmenke thanks, I've probably done this now. I've modified the file integrate.cpp, keeping the same approach as it is done for the Langevin dynamics. Corresponding changes however are implemented in the end functions for the Brownian thermostat acting on the virtual sites (same logic as for Langevin dynamics is kept). See commits "Virtual sites propagation only for rotations" and "Virtual sites translational thermostat control".

bogdan-tanygin:
@hmenke actually, I've introduced all the requested changes and introduced additional adaptations related to the python branch recent notions (philox seeding, normal distribution removal, etc) within the conflicts resolution. Up to you: is it better to proceed with further actions (final review) soon/now or after this PR will reach the backlog top (https://github.com/espressomd/espresso/projects/4) ? Both options fits to me. Maybe the second one just can lead to an extra release delay cause I can work on any further comments (if any) before the backlog come to this PR.

RudolfWeeber:
Unfortunately, we are not going to make it for the code freeze of 4.1 which is next week.
There is ongoing work into simplifying the integration loop (#3062) which should go in first.
I'll then take a closer look at this PR.

bogdan-tanygin:
OK, understood. Let's do one by one then. Indeed, otherwise the main loop refactoring will be too tricky.

RudolfWeeber:
My understanding of conclusions from offline discussions during the school:
The propagation kernels (which already are separated functions) need to be moved to src/core/integrators and called from the hooks in integrate.cpp.
Then, this PR can go ahead.

The scheme, like SD, would be a single step scheme.

The BD-related code in rotation.cpp (which just skips the Velocity-Verlet logic) will go away, once the integration kenrels for rotation are pulled out like the ones for translation.
I'll do this next coding day.

bogdan-tanygin:
> My understanding of conclusions from offline discussions during the school:
> The propagation kernels (which already are separated functions) need to be moved to src/core/integrators and called from the hooks in integrate.cpp.
> Then, this PR can go ahead.
> 
> The scheme, like SD, would be a single step scheme.
> 
> The BD-related code in rotation.cpp (which just skips the Velocity-Verlet logic) will go away, once the integration kenrels for rotation are pulled out like the ones for translation.
> I'll do this next coding day.

Hi @RudolfWeeber , I've introduced these changes towards new inline hooks architecture, cleaned up the rotational specific code, fixed conflicts and tests. However, perhaps, an additional round of review is preferred cause I changed surrounding architecture a bit: I've moved some rotational functions from .cpp to .hpp and wrapped out one of them in order to avoid a code duplication. I believe, that this change will be of use for whole Es. platform: e.g. an ability to call both body or lab frame based rotation, to call these routines from other .hpp inlines, etc. In short, this PR is ready for a final review. However, no rush: feel free to select a correct order of SD and BD features merge. Some similarities are possible. Also, it depends whether we would like to keep the SD and BD architecture pattern aligned.

RudolfWeeber:
I'll take a final look.

testsuite/python/tests_common.py
```diff
@@ -173,6 +173,31 @@ def transform_vel_from_cartesian_to_polar_coordinates(pos, vel):
         (pos[0] * vel[1] - pos[1] * vel[0]) / (pos[0]**2 + pos[1]**2), vel[2]])
 
 
+def define_rotation_matrix(system, part):
```
jngrad: 1. Why is this function identical to `rotation_matrix_quat()`?
2. Can `define_rotation_matrix()`, `rotation_matrix_quat()`, `convert_vec_body_to_space()` be simplified down to the following code?
    ```python
    from scipy.spatial.transform import Rotation
    
    def rotation_matrix_quat(quat):
        return Rotation.from_quat(np.roll(quat, -1)).as_dcm()
    
    def convert_vec_body_to_space(system, part, vec):
        R = define_rotation_matrix(system[part].quat)
        return np.dot(R, vec)
    ```

jngrad: Nevermind point 2, `scipy.spatial.transform` is not available in the `scipy` version packaged in the Ubuntu 18.04 repository.

RudolfWeeber:
On Mon, Nov 11, 2019 at 04:54:42AM -0800, Jean-Noël Grad wrote:
> 
> 
> 
> > @@ -173,6 +173,31 @@ def transform_vel_from_cartesian_to_polar_coordinates(pos, vel):
>          (pos[0] * vel[1] - pos[1] * vel[0]) / (pos[0]**2 + pos[1]**2), vel[2]])
>  
>  
> +def define_rotation_matrix(system, part):
> 
> 1. Why is this function identical to `rotation_matrix_quat()`?
> 2. Can `define_rotation_matrix()`, `rotation_matrix_quat()`, `convert_vec_body_to_space()` be simplified down to the following code?
>     ```python
>     from scipy.spatial.transform import Rotation
We need to make a decision about depending on scipy (or is this de facto already the case?)
I'd not mind the dependency.

jngrad:
@RudolfWeeber At the moment we only depend on scipy for tutorials. The solution I proposed would simplify tremendously the code, but the corresponding scipy module was only introduced last May. The Ubuntu 18.04 repository has the outdated scipy version 0.19 while 18.10 has version 1.1. Since we support the two last Ubuntu LTS, we're stuck here until mid-2023

RudolfWeeber:
On Mon, Nov 11, 2019 at 07:17:17AM -0800, Jean-Noël Grad wrote:
> @RudolfWeeber At the moment we only depend on scipy for tutorials. The solution I proposed would simplify tremendously the code, but the corresponding scipy module was only introduced last May. The Ubuntu 18.04 repository has the outdated scipy version 0.19 while 18.10 has version 1.1. Since we support the two last Ubuntu LTS, we're stuck here until mid-2023
Let's discuss minimum versions to support elsewhere, but I agree, 18.04 packages should work for a while longer.

RudolfWeeber:
I think, we can remove the BROWNIAN_DYNAMICS compile time feature as it does not have performance impacts. The 'unifdef'tool should do it for the core, the Python stuff has to be done by hand unfortunately.
Skipping virtual sites does not have to be ifdef'ed either, I think. ParticleProperties.is_virtual should be
constexpr const is_virtual = false
if VS are not compiled in.

I'll sort it out next week.

src/core/integrators/brownian_inline.hpp
```diff
@@ -0,0 +1,510 @@
...
+#ifdef EXTERNAL_FORCES
+    if (!(p.p.ext_flag & COORD_FIXED(j)))
+#endif
+    {
+      // Second (deterministic) term of the Eq. (14.39) of Schlick2010.
+      // Only a conservative part of the force is used here
+#ifndef PARTICLE_ANISOTROPY
+      p.r.p[j] += p.f.f[j] * dt / (local_gamma);
+#else
+      p.r.p[j] += p.f.f[j] * dt / (local_gamma[j]);
```
fweik: Are you sure that this is correct? The gamma's are in the body-fixed frame while the positions and forces are in the space-fixed frame. There are several other place like this.

RudolfWeeber: I also think, the gamma needs to be converted body->space. Same for the vel update a few lines down. Bogdan?

bogdan-tanygin: @fweik @RudolfWeeber you are right, I need to correct this. An origin of this confusion: the current BD is consistent for spherical particles only. However, the first: it relates only to a rotational inertia tensor; and, the second: I need to keep this code architecturally consistent.

testsuite/python/tests_common.py
bogdan-tanygin: @jngrad for point 1: agree, I will deduplicate this. Perhaps, this PR is just very old is come routines have been added.

bogdan-tanygin:
> I'll sort it out next week

@RudolfWeeber thanks, understood. Just let me know once I could proceed with a removal of the BROWNIAN_DYNAMICS compile time feature. As I understood, you are going to update something in the base.

RudolfWeeber:
> @RudolfWeeber thanks, understood. Just let me know once I could proceed with a removal of the BROWNIAN_DYNAMICS compile time feature. As I understood, you are going to update something in the base.
That's not what I meant. I already removed the ifdef BROWNIAN_DYNAMICs from the PR and pushed to your branch. You might have to pull from your github repo. You should have commit 45088428a35131a8d0.
You can apply the frame transformation at any time.

bogdan-tanygin:
> @RudolfWeeber thanks, understood. Just let me know once I could proceed with a removal of the BROWNIAN_DYNAMICS compile time feature. As I understood, you are going to update something in the base.
> That's not what I meant. I already removed the ifdef BROWNIAN_DYNAMICs from the PR and pushed to your branch. You might have to pull from your github repo. You should have commit [4508842](https://github.com/espressomd/espresso/commit/45088428a35131a8d0b3824672454923bfbbcf3d). You can apply the frame transformation at any time.

Understood, clear, thanks!

src/core/integrators/brownian_inline.hpp
bogdan-tanygin: @fweik @RudolfWeeber I've corrected this and similar fragments where needed.

testsuite/python/tests_common.py
bogdan-tanygin: @jngrad point 1: done.

bogdan-tanygin:
hi @RudolfWeeber , I've introduced changes according to the comments above / fixed the latest base conflicts.

RudolfWeeber:
I merged the Python branch up to (not including) the Python integrator clean up done by @jngrad.
I tried to dis-entangle the Gaussian random number generation, such that the genertaor does not increment the counter. Otherwise, the counters on different MPI canks can diverge.
The "solution" is to allow a slight weakness in the random numbers should the Box Muller corner case of one of the numbers being exactly zero ever occur (Once every 2^20 tires, approximat4ely).

Unfortunately, this is still not ready, because Brownian dynamics is entangled with the Langevin thermostat in serveral places:
* in Python, the brownian thermostat setup calls the langevin thermostat setup
* the rng for Brownian dynamics uses the Langevin counter
* Brownian dynamics re-uses Langevin globals
* the tests were entangled as well

I removed the Brownian dynamics stuff from the langevin test and added a separate brownian dynamics test. It is probably not yet complete.
The velocity autocorrelation based diffusion testing does not work for Brownian dynamics, as consecutive velocities are uncorrelated.
I put in a basic MSD-based diffusioun test for now.

I do not have time to work on this more, at present.

jngrad:
@RudolfWeeber Let me know if ede6e7e1a is what you had in mind. Don't forget to add `#define BROWNIAN_PER_PARTICLE` in your myconfig.hpp.

RudolfWeeber:
@jngrad, yes. Could you please gather all the Brownian-dynamics related globals in a sinlge struct? Then, it's ready, I think.

jngrad:
The Brownian globals are now in a struct.

In [jngrad/therm-refactor](https://github.com/jngrad/espresso/tree/therm-refactor) I also moved the Langevin and NPT globals in structs and moved the Langevin code to `integrators/langevin_inline.hpp` for consistency with `integrators/brownian_inline.hpp`. Feel free to do `git merge --ff-only`. If you consider it's out of scope, I'll open a dedicated PR once this one is merged.

RudolfWeeber (approved):

RudolfWeeber:
@hmenke, from my side, this PR is ready. Please check your open review for things not addressed. If there aren't any, please approve the PR.

src/core/integrators/brownian_inline.hpp
```diff
@@ -0,0 +1,570 @@
...
+    // Only a conservative part of the force is used here
+#ifdef PARTICLE_ANISOTROPY
+    if (aniso_flag) {
+      delta_pos_body[j] = force_body[j] * dt / (local_gamma[j]);
+    } else {
+#ifdef EXTERNAL_FORCES
+      if (!(p.p.ext_flag & COORD_FIXED(j)))
+#endif
+      {
+        p.r.p[j] += p.f.f[j] * dt / (local_gamma[j]);
```
KaiSzuttor, line 77: still missing a check for virtual particles here and probably at other places too

RudolfWeeber:
@KaiSzuttor, the virtual sites are excluded for translational degrees in 
void brownian_dynamics_propagator(const ParticleRange &particles)

For rotational degrees, VS are not excluded, which is correct. If the virtual sites have_quanterion feature is used, the virtual sites update will assign the correc orientation after the propagation step.

RudolfWeeber:
I found a different but related issue. The testcase for checking thermo_virtual with Brownian dynamics was incorrect (copy/paste error from the Langevin test). 
Fixing the test revealed that thermo_virtual=True was ignored by the virtual sites guards. Pushing the fix now.

src/core/integrators/brownian_inline.hpp
```diff
@@ -0,0 +1,571 @@
...
+    // further on top of it
+#ifdef PARTICLE_ANISOTROPY
+    if (aniso_flag) {
+      vel_body[j] = force_body[j] * dt / (local_gamma[j]);
+    } else {
+#ifdef EXTERNAL_FORCES
+      if (!(p.p.ext_flag & COORD_FIXED(j)))
+#endif
+      {
+        p.m.v[j] = p.f.f[j] / (local_gamma[j]);
```
jngrad, line 152: @psci2195 @RudolfWeeber Is this expression of the velocity intentionally missing the time step `dt`?

src/core/thermostat.cpp
```diff
@@ -134,6 +151,47 @@ void thermo_init_npt_isotropic() {
...
+   */
+  /* If gamma_rotation is not set explicitly,
+     we use the translational one. */
+  if (brownian.gamma_rotation < GammaType{}) {
+    brownian.gamma_rotation = brownian.gamma;
+  }
+  brownian.sigma_vel_rotation = sqrt(temperature);
+  if (temperature > 0.0) {
+    brownian.sigma_pos_rotation_inv =
+        sqrt(brownian.gamma / (2.0 * temperature));
```
jngrad: Are we intentionally using `gamma` instead of `gamma_rotation` in the expression of `sigma_pos_rotation_inv`?

src/core/integrators/brownian_inline.hpp
bogdan-tanygin: @jngrad yes, the BD velocity does not depend on the time step. It is equal to the terminal viscous velocity.

src/core/thermostat.cpp
bogdan-tanygin: looks like a typo, thanks, should be rotational.

src/core/integrators/brownian_inline.hpp
jngrad: I still have a hard time understanding the code. When `PARTICLE_ANISOTROPY` is disabled, that function returns a velocity multiplied by the time step.

bogdan-tanygin: thanks for catching! This is a mistake. The dt should be deleted there. The viscous test did not catch this because it covers the translationally isotropic particle only. I can fix this and increase a coverage.

jngrad: I've actually written C++ unit tests for Brownian and Langevin already. They caught that discrepancy, and I wasn't sure if it was intentional. Once the `dt` is removed on the line above and the `gamma_rotational` gets fixed in the other comment, my tests should pass. It would be nice to have an additional Python test to cover the anisotropic case, but we also shouldn't delay the merging of this PR for too long. I've already carried out a significant refactoring of the Brownian/Langevin thermostats thanks to @fweik to make them non-global and testable with Boost, but this PR is blocking me.

fweik:
@jngrad I think we should merge this now, or you should push your tests onto this branch.

PR #1843, opened 2018-02-10
Mindist for constraints

davidsean:
Fixes #

Description of changes:
 - wrote a core mindist for constraints
 - python interface
 - updated documentation
 - fixed a warmup problem for the constraints sample
 - CI tests for wall and cylinder

PR Checklist
------------
 - [x] Tests?
   - [x] Interface
   - [x] Core 
 - [x] Docs?

### DIFF
--- a/src/core/constraints/ShapeBasedConstraint.cpp
+++ b/src/core/constraints/ShapeBasedConstraint.cpp
@@ -20,25 +20,23 @@ Vector3d ShapeBasedConstraint::total_force() const {
 
 
 double ShapeBasedConstraint::min_dist() {
-  double pt[3];
-  double vec[3];
-  double dist;
-  double mindist = std::numeric_limits<double>::infinity();
-  Particle part_rep;
-  part_rep.p.type = m_type;
-  IA_parameters *ia_params ;
-  auto & parts = partCfg();
-  for (auto jt = parts.begin(); jt != (parts.end()); jt++) {
-    pt[0] = jt->r.p[0];
-    pt[1] = jt->r.p[1];
-    pt[2] = jt->r.p[2];
-    ia_params = get_ia_param(jt->p.type, part_rep.p.type);
+  double global_mindist;
+  auto parts = local_cells.particles();
+
+  auto const local_mindist = std::accumulate(
+        parts.begin(), parts.end(), std::numeric_limits<double>::infinity(),
+        [this](double min, Particle const& p) {
+    IA_parameters *ia_params;
+    ia_params = get_ia_param(p.p.type, m_type);
     if (checkIfInteraction(ia_params)) {
-        m_shape->calculate_dist(pt, &dist, vec);
-        mindist = std::min(mindist, dist);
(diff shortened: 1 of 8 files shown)


testsuite/constraint_shape_based_wall.py
```diff
@@ -45,14 +45,15 @@ def test(self):
         # (1)
         constraint_wy = espressomd.constraints.ShapeBasedConstraint(
             shape=wy, particle_type=1)
-        wall_xz = system.constraints.add(constraint_wy)
+        wall_xy = system.constraints.add(constraint_wy)
```
jonaslandsgesell: Why did you exchange the naming of the wall variables? The old convention was the following: if the normal vector is in y direction (0,1,0) then the wall lies parallel to the xz plane and therefore the wall is named wall_xz

src/core/constraints/ShapeBasedConstraint.cpp
```diff
@@ -17,6 +18,30 @@ Vector3d ShapeBasedConstraint::total_force() const {
     return total_force;
 }
 
+
+double ShapeBasedConstraint::mindist() {
```
fweik: This is parallel code (it runs on all nodes in synchronous mode), so you can't use `partCfg`, as it exists only on the head node. There is also no need, as your algorithm does not need global information, you can run it on each node locally and then reduce the local results to the master. Also folded coordinates should be used with the shape, but particles in partCfg have folded coordinates. So you could do something along the lines of

```c++
auto & parts = local_cells.particles();

auto const local_mindist = std::accumulate(parts.begin(), parts.end(), std::numeric_limits<double>::infinity(), [this](double min, Particle const& p) {
  if(!checkIfInteraction(...))
    return min;
  else  {
     double dummy[3], dist;
     m_shape->calculate_dist(folded_position(p).data(), &dist, dummy); 

     return std::min(min, dist);
   });

   double global_min;
   boost::mpi::reduce(comm_cart, local_min, global_min, std::min, 0);

   return global_min;
}
```  
}

fweik (changes requested):
There are several issues with this:
- It uses the shape contained in the constraint with unfolded coordinates where the shape expects folded positions. (positions in partCfg are unfolded)
- The tests apparently did not catch this.
- It uses partCfg that can only be used on the head node on all nodes.

RudolfWeeber:
The python method in analysis.pyx is called min_dist.
I'm also not sure, min_dist is actually what it does. For a sphere,
e.g., I think it takes the vector which is perpendicular to the
surface of the sphere (@pkreissl?).

testsuite/constraint_shape_based_wall.py
davidsean: I had it follow the naming convention for walls and shape where the last character gives the normal. I did not realize xz meant the xz plane. I don't care which convention is chosen, however constraint_wy should then be constraint_xz and the shape wy should be xz to be consistent.

fweik (approved):

PR #1845, opened 2018-02-12
consistent style in samples

davidsean:
Fixes #

changed samples such they all have the same style, so that users could copy/paste code blocks from different samples into their own.
for example, all sampls now have
system=espresomd.System()
instead of sys=espresomd.System() or S=espresomd.System()

all concerned scripts now also use
import numpy as np

Description of changes:
 - added seeding in samples
 - import numpy as np
 - system=espresomd.System() 
 - widom and grandcanonical had older RE syntax
 - most samples run

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


jonaslandsgesell (approved):

pkreissl (approved):
Resolve conflicting file issue, can be merged, then.

PR #1853, opened 2018-02-15
fix Python 3 string/map issues

pdebuyl:
Description of changes:
1. explicitly set encoding in set_random_state_PRNG in
   src/python/espressomd/system.pyx
2. use (c)StringIO from six in src/python/espressomd/MDA_ESP/__init__.py
3. "listify" a map in src/python/espressomd/MDA_ESP/__init__.py

There were a few Python 3 incompatibilities I encountered.

I used https://stackoverflow.com/questions/42678779/how-to-pass-string-from-python3-to-cythonized-c-function for the Python 3 `libcpp.string` vector issue.

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?

### DIFF
--- a/src/python/espressomd/MDA_ESP/__init__.py
+++ b/src/python/espressomd/MDA_ESP/__init__.py
@@ -30,7 +30,12 @@ A minimal working example is the following:
 <Universe with 1 atoms>
 
 """
-from six import StringIO
+
+try:
+    import cStringIO as StringIO
+except ImportError:
+    from io import StringIO
+
 import numpy as np
 import MDAnalysis
 
--- a/src/python/espressomd/MDA_ESP/__init__.py
+++ b/src/python/espressomd/MDA_ESP/__init__.py
@@ -30,7 +30,7 @@ A minimal working example is the following:
 <Universe with 1 atoms>
 
 """
-import cStringIO
+from six import StringIO
 import numpy as np
 import MDAnalysis
 
@@ -102,7 +102,7 @@ class Stream(object):
             _xyz+=str(_p.v)+'\n'
         for _p in self.system.part:
             _xyz+=str(_p.f)+'\n'
-        return  NamedStream(cStringIO.StringIO(_xyz), "__.ESP")
+        return  NamedStream(StringIO(_xyz), "__.ESP")
 
 class ESPParser(TopologyReaderBase):
     """
@@ -202,7 +202,7 @@ class ESPReader(SingleFrameReaderBase):
                     self.ts = ts = self._Timestep(self.n_atoms, **self._ts_kwargs)
                     self.ts.time = time
                 elif(pos==-1):
-                    self.ts._unitcell[:3] = np.array(map(float,line[1:-2].split()))
+                    self.ts._unitcell[:3] = np.array(list(map(float,line[1:-2].split())))
                 elif(pos < n_atoms):
                     positions[pos] = np.array(list(map(float, line[1:-2].split())))
                 elif(pos < 2*n_atoms):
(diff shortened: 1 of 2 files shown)


src/python/espressomd/MDA_ESP/__init__.py
```diff
@@ -30,7 +30,7 @@
 <Universe with 1 atoms>
 
 """
-import cStringIO
+from six import StringIO
```
hmenke: `six` is not part of the Python standard library.  I would prefer if the code wasn't introducing additional dependencies.

src/python/espressomd/system.pyx
```diff
@@ -330,7 +330,7 @@ cdef class System(object):
             for j in range(_state_size_plus_one + 1):
                 states_on_node_i.append(
                     rng.randint(0, numeric_limits[int].max()))
-            states[i] = " ".join(map(str, states_on_node_i))
+            states[i] = (" ".join(map(str, states_on_node_i))).encode('utf-8')
```
hmenke, line 333: Wouldn't it be easier to use a Unicode literal string? (untested)
```python
states[i] = u" ".join(map(str, states_on_node_i))
```

hmenke (changes requested):

src/python/espressomd/system.pyx
pdebuyl: It does not work with either of `u` or `b` as a prefix. As we store string in a `std::vector[std::string]` at the Cython level for Python 2/3 compatibility, I did proceed by trial and error and then found the SO answer I mention in the PR.

src/python/espressomd/MDA_ESP/__init__.py
pdebuyl: There is already an import of six in `src/python/espressomd/io/writer/h5md.py` and MDAnalysis requires it anyway. If you prefer, I can replace this by a try/except block or a conditional import that check the version of Python.

hmenke (approved):

src/python/espressomd/MDA_ESP/__init__.py
pdebuyl: Also, if needed, [six](https://github.com/benjaminp/six)

> (...) is contained in only one Python file, so it can be easily copied into your project.

But it is on PyPI and does not carry further dependencies.

pdebuyl:
@hmenke I realized that two similar changes were needed in system.pyx, I pushed them on the same PR, I hope this is not an issue.

pdebuyl:
@hmenke is there any extra task needed for this PR on my side?

fweik:
I'm not thrilled about the extra dependency. h5md and MDAnalysis are not required dependencies of Espresso so far, so six is not yet required to build or run it.
Can you please explain why the changes you are introducing here are required? As far as I can tell the code was working fine with python 3 before (the tests are currently run on python 3.5.2).

src/python/espressomd/MDA_ESP/__init__.py
mkuron: We don't need to use six for this. What it does is trivial and you can do the same using something like:
```
try:
    import cStringIO as StringIO
except ImportError:
    from io import StringIO
```

mkuron (changes requested):

pdebuyl:
Ok, as the PR was already approved I believed this was ok. I made the change as requested.

Note however that anyone actually using the `MDA_ESP` module will require mdanalysis that requires six anyway (and about ten thousands other packages of course).

pdebuyl:
@fweik with respect to testing, the issues originated is some of the samples which are not part of the testsuite I suppose. Should I consider a failure to run a test a good reason to submit other PRs?

hmenke (approved):

fweik:
Yes, fixes for the samples are also very welcome. They often slip thru because they are not automatically tested at the moment, but of course they should work.

PR #1857, opened 2018-02-19
Actor

### LINKED ISSUE #1228: P3M forgetts parameters
@kosovan  was pointing out to me that If P3M parameters are set before adding the actor,
they are not saved:

```python
import espressomd as es

bjerrum=2.0;
accuracy=0.001
p3m = electrostatics.P3M(bjerrum_length=bjerrum, accuracy=accuracy, tune=False, mesh=6, cao=5, r_cut=4.64152e-01, alpha=4.53518e+00, )
print ("p3m params:", str(p3m.get_params()));
```

fweik:
Fixes #1228 

Description of changes:
 - Allow non-highlander actors (for test)
 - Only fetch/write params from core if the actor is active,
   otherwise use params in the python class.
 - Test


RudolfWeeber:
Can be merged

PR #1858, opened 2018-02-19
cmake: More robust version detection for sphinx.

fweik:


KaiSzuttor (commented):
LGTM, i thought about sth like ```python -c “import sphinx; print sphinx.__version__”``` which would work in both cases but we would depend on the python executable then.

fweik:
In principle the version of the sphinx executable can differ. Could be e.g. in a different env. I think this is good enough for now.

PR #1859, opened 2018-02-19
Decouple Domain Decomposition and Collision Detection

hirschsn:
In collision.cpp there is a collection of three raw for-loops iterating over the full shell neighborhood of a domain decomposition. This PR slightly increases the decoupling between the two modules by offering a function dd_full_shell_neigh which can be used. This also makes the function dd_position_to_cell_indices obsolete which exists only for this piece of code in the collision detection.

Description of changes:
 - Remove dd_position_to_cell_indices
 - Offer dd_full_shell_neigh
 - Get rid of irritating j-loop in three_particle_binding_domain_decomposition by moving the actual search code to a separate function
 - Use a 1d loop and calls to dd_full_shell_neigh instead of a 3d iteration directly over cell indices
 - Remove some useless comments
 - Convert pointers to references in the updated code

The three particle binding collision detection test still exits successfully and the pairs (basecell, neighborcell) traversed during this test are exactly the same and in the same ordering as before.

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?

### DIFF
--- a/src/core/collision.cpp
+++ b/src/core/collision.cpp
@@ -509,17 +509,44 @@ void three_particle_binding_full_search(
   }
 }
 
+static void three_particle_binding_dd_do_search(
+    Cell *basecell, Particle &p1, Particle &p2) {
+  int basecellno = std::distance(&cells[0], basecell);
+  for (int n = 0; n < 27; ++n) {
+    Cell &cell = cells[dd_full_shell_neigh(basecellno, n)];
+    for (int pno = 0; pno < cell.n; ++pno) {
+      Particle &P = cell.part[pno];
+      // Skip collided particles themselves
+      if ((P.p.identity == p1.p.identity) ||
+          (P.p.identity == p2.p.identity)) {
+        continue;
+      }
+
+      // We need all cyclical permutations, here (bond is placed on 1st
+      // particle, order of bond partners does not matter, so we don't need
+      // non-cyclic permutations).
+      // coldet_do_three_particle_bond checks the bonding criterion and if
+      // the involved particles are not already bonded before it binds them.
+      if (!P.l.ghost) {
+        coldet_do_three_particle_bond(P, p1, p2);
+      }
+
+      if (!p1.l.ghost) {
+        coldet_do_three_particle_bond(p1, P, p2);
+      }
+
+      if (!p2.l.ghost) {
+        coldet_do_three_particle_bond(p2, P, p1);
+      }
+    }
+  }
+}
+
 // Goes through the collision queue and for each pair in it
 // looks for a third particle by using the domain decomposition
 // cell system. If found, it performs three particle binding
 void three_particle_binding_domain_decomposition(
     const std::vector<collision_struct> &gathered_queue) {
-  // We have domain decomposition
-
-  // Indices of the cells in which the colliding particles reside
-  int cellIdx[2][3];
-
-  // Iterate over collision queue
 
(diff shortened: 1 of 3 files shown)


fweik:
This sure is an improvement, but it is a somewhat dangerous interface as the neighbor list is not complete in case of node boundaries (the lower ghosts, that do not interact are not communicated). In the collision detection this still works, because the collision queue is synchronized, and the detection is run on both sides of a node boundary, which guarantees that one of the neighbors will pick up every collision.

I think this should at least be clearly documented, maybe we can even come up with a better name.

hirschsn:
Are you sure about the incomplete boundary communication? I did a quick test: Take 27 (to get rid of any LOCL-magic) nodes (3x3 grid) and at the end of dd_prepare_comm output the communication partners, i.e. `comm->comm[i]` for all `i`. In the synchronized scheme a node has 6 surface neighbors. To fully communicate the whole boundary it, therefore, needs 12 communications.

The test output for all nodes gives `comm->num == 12`. Exemplary for the middle node (same for all the different communicators before order change):
```
Ncomm = 12: Recv (22) Send (4) Recv (4) Send (22) Recv (16) Send (10) Recv (10) Send (16) Recv (14) Send (12) Recv (12) Send (14)
```
Exactly as expected for a fully communicated boundary (all have `n_part_lists > 0`).

fweik:
That would mean that there is excess communication in the normal case? `dd_prepare_comm` is unfortunately completely incomprehensible to me.
Ah there is a comment:
```c++
    /* here we could in principle build in a one sided ghost
       communication, simply by taking the lr loop only over one
       value */
```

fweik (approved):

hirschsn:
I noticed a problem with this patch which I was not aware of before, sorry for that.
This patch uses cell_structure.position_to_cell to resolve the position of particles. However, the collision code is also executed for ghost particles (collision queue is communicated and `local_particles[x]` is set for ghost particles with id x).

`cell_structure.position_to_cell` cannot resolve ghost particles (as this would lead to problems inside the domain decomposition itself). And so it returns a nullptr. The original code (`dd_position_to_cell_indices`) did a projection to the nearest boundary cell in this case. I'm afraid, this patch needs to be reverted if nobody has an idea how to easily fix this.

What we could do: If cell_structure.position_to_cell returns null it is clear that the particle is a ghost particle (since local_particles[x] is set). We could search for the nearest ghost cell by adding/subtracting the cell width from the particle's coordinates.

hirschsn:
Since the PR is closed: @fweik did you get notified about the above comment?

fweik:
Yes, but we did not have time to deal with it. I think this can easily
fixed though.

hirschsn:
In a way that works for arbitrary decompositions? I'd be interested in your fix (even the idea) because I need it.

hirschsn:
If resolving the cell of particle.r.p fails but the particle is stored on this node, particle.l.p_old must still be a valid position inside the subdomain of the node and not too far from the current position of the particle? This should work if Verlet lists are enabled.

hirschsn:
Something like this: https://github.com/hirschsn/espresso/commit/7a371d2f4d3275694e98384684a9c4e4e515a483

fweik:
@RudolfWeeber kannst du dir das bitte auch mal anschauen. Ich denke das funktioniert.

PR #1864, opened 2018-02-21
doc: Corrected exclusions doc.

### LINKED ISSUE #1839: Exclusions Documentation still valid?
Is the following documentation (http://espressomd.org/html/doc/particles.html?highlight=exclusion) still valid? I remember discussions about that:

```
exclusions also remove the short range part of electrostatics and dipolar interactions. Hence, exclusions should not be applied to pairs of particles which are charged or carry a dipole
```

fweik:
Fixes #1839.

Corrected the exclusion docs to reflect that in fact the short range parts of electrostatics and magnetostatics methods are not excluded. This is already tested in the exclusion test.

### DIFF
--- a/doc/sphinx/particles.rst
+++ b/doc/sphinx/particles.rst
@@ -169,7 +169,7 @@ Exclusions
 Particles can have an exclusion list of all other particles where non-bonded interactions are ignored.
 This is typically used in atomistic simulations, 
 where nearest and next nearest neighbor interactions along the chain have to be omitted since they are included in the bonding potentials.
-Exclusions do not apply to the short range part of electrstatics and magnetostatics methods, e.g. to P3M.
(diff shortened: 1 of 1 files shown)


doc/sphinx/particles.rst
```diff
@@ -169,8 +169,7 @@ Exclusions
 Particles can have an exclusion list of all other particles where non-bonded interactions are ignored.
 This is typically used in atomistic simulations, 
 where nearest and next nearest neighbor interactions along the chain have to be omitted since they are included in the bonding potentials.
-Be aware that currently, exclusions also remove the short range part of electrostatics and dipolar interactions. Hence, exclusions should not be applied to pairs of particles which are charged or carry a dipole.
-
+Exclusions do not apply to the short range part of electrstatics and magnetostatics methods, e.g. to P3M.
```
KaiSzuttor: electrostatics

PR #1870, opened 2018-02-21
warmup and force type

davidsean:
Fixes #

create_polymer() creates type 1 monomers with type_poly_uncharged=0. This is a bit confusing when one wants uncharged monomers and ends up with what seems to be the wrong type.
The example script now forces all monomer types to be type=0. and the Docs have a better explaination of particle type assignment from create_polmyer()

create_polymer()  did not have a warmup steps, now it does.

Description of changes:
 - warmup in minimal-polymer,
 - writes trajectory file
 - forces particle type
 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [x] Docs?

### DIFF
--- a/samples/minimal-polymer.py
+++ b/samples/minimal-polymer.py
@@ -22,6 +22,7 @@ import espressomd
 from espressomd import thermostat
 from espressomd import interactions
 from espressomd import polymer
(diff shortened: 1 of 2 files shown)


davidsean:
dpd test failed?

fweik (approved):

PR #1873, opened 2018-02-22
Thermo virtual

fweik:
Description of changes:
 - Introduced a switch on the langevin and lb thermos to indicate if they should act on virtual sites.
 - Made ParticleProperty::is_virtual always exist (as static if VIRTUAL_SITES is off)
 - Some cleanup of globals
 - Some header cleanup
 - Finer control of what goes into GPU particles depending on FEATURES.
 - Removed unused langevin_{trans,rot}_switch


RudolfWeeber:
If I understand this correctly, you can choose per thermostat whether it acts on virtual sites, but they use a common global.
I not sure, that's save. You might even get an outcome depending on the order of setup when you combine an lb thermostat with a Langevin thermostat for the rotational degrees of freedom, as they have different defaults for acting on virtual sites.
Otherwise, it looks good.

fweik:
Yes, but I did only put the property on Langevin and LB, which are mutually exclusive. Also we can not put it on the Thermostat class, because it is not implemented for all thermostats.

RudolfWeeber:
Based on further off-line discussion, this can be merged.
We need to keep in mind that changes are needed, if the guard against using lb and Langevin thermostats together are ever removed for some reasons.

PR #1874, opened 2018-02-22
core: Removed unused code.

fweik:
Description of changes:
 - Removed unused code.


hmenke:
Also remove `src/core/utils/Timer.cpp` and `src/core/utils/Timer.hpp`.  The `Timer.cpp` is not compiled in any `CMakeLists.txt` and `Timer.hpp` is used nowhere except `Timer.cpp`.  (In #1644 I fixed the errors in those files to make them compilable but they don't seem to be used at all)

PR #1875, opened 2018-02-23
Virtual site orientation update

KaiSzuttor:
Description of changes:
 - updating the virtual site orientation
 - code improvement for rotation.cpp

PR Checklist
------------
 - [x] Tests?
   - [x] Interface
   - [ ] Core 
 - [x] Docs?


src/core/particle_data.cpp
```diff
@@ -171,14 +171,20 @@ void build_particle_node() { mpi_who_has(); }
...
+    throw std::runtime_error("Invalid particle id!");
 
   if (particle_node.empty())
     build_particle_node();
 
   auto const needle = particle_node.find(id);
 
-  return (needle != particle_node.end()) ? needle->second : -1;
+  // Check if particle has a node, if not, we assume it does not exist.
+  int pnode = (needle != particle_node.end()) ? needle->second : -1;
```
fweik:
```c++
if(needle != particle_node.end()) {
    return needle->second;
} else {
    throw std::runtime_error("Particle node not found!");
}
```

src/core/particle_data.cpp
```diff
@@ -1411,4 +1361,11 @@ void pointer_to_rotational_inertia(Particle const *p, double const *&res) {
 }
 #endif
 
-bool particle_exists(int part) { return get_particle_node(part) != -1; }
+bool particle_exists(int part_id) noexcept {
+  try {
```
fweik: Please refrain from using exceptions for regular control flow, you can just do 
```c++
return particle_node.count(part_id);
```

src/core/particle_data.cpp
```diff
@@ -316,9 +322,10 @@ Utils::Cache<int, Particle> particle_fetch_cache(max_cache_size);
 void invalidate_fetch_cache() { particle_fetch_cache.invalidate(); }
 
 const Particle *get_particle_data(int part) {
-  auto const pnode = get_particle_node(part);
-  /* Check if particle exists at all. */
-  if (-1 == pnode) {
+  int pnode;
```
fweik: If you want to be consequent, you should do
```c++
return particle_exists(part) ? get_particle_node(part) : nullptr;
```
which is somewhat more idiomatic, but results in two lookups. Alternatively you can get the iterator and compare it to end.

src/core/virtual_sites/VirtualSitesRelative.hpp
```diff
@@ -20,38 +20,45 @@
...
-#include "virtual_sites.hpp"
-#include "particle_data.hpp"
 #include "communication.hpp"
+#include "particle_data.hpp"
+#include "virtual_sites.hpp"
+
+/** @brief Virtual sites implementation for rigid bodies */
+class VirtualSitesRelative : public VirtualSites {
+public:
+  VirtualSitesRelative(){};
```
fweik: Not needed.

fweik (changes requested):

src/core/virtual_sites/VirtualSitesRelative.hpp
hmenke: Even if it was necassary it should have been
```cpp
VirtualSitesRelative() = default;
```

src/core/particle_data.cpp
hmenke: Especially not with a `catch`-em-all.  This will shadow exceptions like `std::bad_alloc` for which there is no way of recovering.

src/core/virtual_sites/VirtualSitesRelative.hpp
fweik: I guess this is a matter of style as the class has no user-defined special functions. If kept, the `= default` version makes the intent clearer.

hmenke: [C.80: Use `=default` if you have to be explicit about using the default semantics](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c80-use-default-if-you-have-to-be-explicit-about-using-the-default-semantics)

KaiSzuttor: i haven't touched that part besides formatting

src/core/particle_data.cpp
```diff
@@ -169,16 +171,18 @@ void build_particle_node() { mpi_who_has(); }
...
     build_particle_node();
 
   auto const needle = particle_node.find(id);
 
-  return (needle != particle_node.end()) ? needle->second : -1;
+  // Check if particle has a node, if not, we assume it does not exist.
+  if (needle != particle_node.end()) {
+    return boost::optional<int>{needle->second};
+  } else {
+    return boost::optional<int>{};
```
fweik: `boost::none`

src/script_interface/virtual_sites/VirtualSites.hpp
```diff
@@ -40,8 +40,18 @@ class VirtualSites : public AutoParameters {
...
+         {"have_quaternion",
+          [this](const Variant &v) {
+            virtual_sites()              // NOLINT, bug in clang-tidy-4.0
+                ->set_have_quaternion(   // NOLINT, bug in clang-tidy-4.0
+                    get_value<bool>(v)); // NOLINT, bug in clang-tidy-4.0
+          },
+          [this]() {
+            return virtual_sites()       // NOLINT, bug in clang-tidy-4.0
+                ->get_have_quaternion(); // NOLINT, bug in clang-tidy-4.0
           }}});
```
mkuron, line 55: Are you sure this is necessary? You just doubled the number of `NOLINT`s in Espresso. There is probably some design weakness here if Clang-Analyzer trips over this code so much.

KaiSzuttor: Maybe there is a design weakness, ~~I don't know who put the NOLINTS there before me~~, I just adapted. @fweik maybe knows more

KaiSzuttor:
Tests seem to fail because of a bug in cython. I checked and it compiles from cython version 0.26 on. Maybe we should use the required cython version in all docker containers.

mkuron:
Currently we only require Cython 0.23 ([CMake](https://github.com/espressomd/espresso/blob/d90ed296af960023c309519b1893639070b1d4f5/CMakeLists.txt#L185), [documentation](https://github.com/espressomd/espresso/blob/e656d9ad749860c2f83f9776fa6b0f6b057c0463/doc/sphinx/installation.rst#installing-requirements-on-ubuntu-1604-lts)). If that is too buggy to support, we need to upgrade, but since that is the version shipped with Ubuntu 16.04, I think we should continue supporting it for a few more months. Ubuntu 18.04 is going to ship with 0.26, maybe higher.

KaiSzuttor:
you are right, but what i meant was the fact that some docker containers install cython via pip which currently installs cython 0.27. We should consistently use 0.23 for the CI tests.

mkuron:
Actually, we should test both 0.23 and the latest version. That way, we can ensure to remain forward and backward compatible. Ubuntu has 0.23 and everything else has the latest version right now.

src/script_interface/virtual_sites/VirtualSites.hpp
fweik: That is a known false-positive in clang-tidy (4.0), which is fixed in later releases. As far as I can tell the code is correct, as it is not actually evaluated in the constructor (this would be UB), but merely used to define a lambda.

KaiSzuttor:
that would almost double the tests, i think thats not feasible

mkuron:
I am not saying we need more builds. We just need at least one build that uses 0.23 and at least one build that uses the most recent version. We have that, so there is nothing we need to change about the Docker images.

So we shouldn't do this:

>  We should consistently use 0.23 for the CI tests.

KaiSzuttor:
But then we only check parts of the interface (although with maxset the largest) for both versions.
What I meant with 

> We should consistently use 0.23 for the CI tests.

is, that if we require version 0.23, we should perform *all* checks for that version.

KaiSzuttor:
So we wait with merging until Ubuntu 18.04 is out?

mkuron:
You'd need to until the end of the year so that everybody has time to upgrade their machines... So you better try to find a 0.23-compatible workaround.

mkuron:
https://github.com/cython/cython/blob/20c3022aec0951c1809f5030e1ce64e240de5df7/Cython/Utility/ModuleSetupCode.c#L333-L349 shows a correctly-implemented `operator&` in the latestet Cython version. https://github.com/cython/cython/blob/0.23.x/Cython/Utility/ModuleSetupCode.c#L252-L264 is the same code for 0.23, which does not have it (but it has `operator->` and `operator T&`, which you might be able to make use of instead).

Replacing `&get_particle_data(self.id)` with something like `&(<const Particle>(get_particle_data(self.id)))` might work.

KaiSzuttor:
Your proposed replacement does not work, gives me
```
Taking address of non-lvalue
```

KaiSzuttor:
ready for review @fweik

src/core/particle_data.hpp
```diff
@@ -1028,6 +1035,6 @@ void pointer_to_swimming(Particle const *p,
 void pointer_to_rotational_inertia(Particle const *p, double const *&res);
 #endif
 
-bool particle_exists(int part);
+bool particle_exists(int part) noexcept;
```
fweik: The `noexcept` is a lie, as `particle_exists` potentially calls `build_particle_node`, which can throw for various reasons.

fweik (changes requested):

fweik:
Looks good to me. The particle stuff is a nice improvement. If you want to make it perfect to can add a test to `particle.py` that verifies that trying to access a non-existent id will actually throw. I think you can make a `particle_data.ParticleHandle` with a non-existing id, and then accessing any property should throw.

fweik (approved):

KaiSzuttor:
```autodoc: failed to import module u'espressomd.visualization_mayavi'; the module executes module level statement and it might call sys.exit().``` @mkuron

mkuron:
@KaiSzuttor, looking at the autodoc source code, it seems like it is only triggered when something actually calls `sys.exit()`. I have no idea why that should happen.

I think this error only just appeared because you installed pyface in https://github.com/espressomd/docker/pull/24. Now Mayavi can actually be imported, but probably fails at some later point because it doesn't have access to a screen.

KaiSzuttor:
~~You may have only checked the current sphinx version ```1.7.0``` but in version ```1.6.6``` which is used in the test-container and which we require every exception is caught and leeds to an errormessage
https://github.com/sphinx-doc/sphinx/blob/739022730295c4968ecc212bbb80b03981eeced3/sphinx/ext/autodoc.py#L673-L691~~, I was wrong, in the container we even have version ```1.7.1```

KaiSzuttor:
Seems like we are not able to build sphinx since they released version ```1.7.1``` 11 days ago.

PR #1883, opened 2018-02-26
Test CodeCov threshold

hmenke:


hmenke:
Hm, this test was unsuccessful (increasing coverage).  Well, I hope it works anyway...

PR #1884, opened 2018-02-26
Minimize energy

### LINKED ISSUE #1779: Forces are not scaled after energy minimization
Steepest descent uses unscaled forces, and they are not rescaled if the integrator is switched back to VV,
the way to go is probably to set `recalc_forces` after switching the integrator.

fweik:
Fixes #1779.

Description of changes:
 - Rescale forces after steepest descent integration,
   as steepest descent works with unscaled forces.


RudolfWeeber:
Can be merged

PR #1886, opened 2018-02-28
Lees edwards

fweik:
Description of changes:
 - Correct distance calc for Lees Edwards with bonds
 - Interface improvement.


pdebuyl:
Hi @fweik  should this remain open?

PR #1893, opened 2018-03-07
Add signal handler for Ctrl+C (SIGINT)

### LINKED ISSUE #1892: Ctrl-C does not interrupt integration loop
Waiting for an integrator.run(10000) to complete...

hmenke:
Fixes #1892 

Little test script which takes forever:
```python
import numpy as np
import espressomd

system = espressomd.System(box_l=[100,100,100])
system.time_step = 0.01
system.cell_system.skin = 0.1

for i in range(10000):
    system.part.add(pos=np.random.random()*system.box_l)

system.integrator.run(10000)
print("Integration done!")
```

### DIFF
--- a/src/core/integrate.cpp
+++ b/src/core/integrate.cpp
@@ -53,6 +53,7 @@
 #include "pressure.hpp"
 #include "rattle.hpp"
 #include "rotation.hpp"
+#include "signalhandling.hpp"
 #include "swimmer_reaction.hpp"
 #include "thermostat.hpp"
 #include "utils.hpp"
@@ -97,6 +98,11 @@ double verlet_reuse = 0.0;
 double db_max_force = 0.0, db_max_vel = 0.0;
 int db_maxf_id = 0, db_maxv_id = 0;
 #endif
+
+bool set_py_interrupt = false;
+namespace {
+volatile static std::sig_atomic_t ctrl_C = 0;
+}
 
 /** \name Private Functions */
 /************************************************************/
@@ -462,6 +468,18 @@ void integrate_vv(int n_steps, int reuse_forces) {
 
     if (check_runtime_errors())
       break;
+
+    // Check if SIGINT has been caught.
+    if (ctrl_C == 1) {
+      // Reset the flag.
+      ctrl_C = 0;
+
+      // Set global to notify Python of signal.
+      set_py_interrupt = true;
+
+      // Break the integration loop
+      break;
+    }
   }
 // VIRTUAL_SITES update vel
 #ifdef VIRTUAL_SITES
@@ -840,6 +858,9 @@ void force_and_velocity_display() {
 /** @todo This needs to go!! */
 
 int python_integrate(int n_steps, bool recalc_forces, bool reuse_forces_par) {
+  // Override the signal handler so that the integrator obeys Ctrl+C
+  SignalHandler sa(SIGINT, [](int) { ctrl_C = 1; });
+
   int reuse_forces = 0;
   reuse_forces = reuse_forces_par;
   INTEG_TRACE(fprintf(stderr, "%d: integrate:\n", this_node));
--- a/src/python/espressomd/integrate.pyx
+++ b/src/python/espressomd/integrate.pyx
@@ -17,9 +17,11 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 from __future__ import print_function, absolute_import
+from cpython.exc cimport PyErr_CheckSignals, PyErr_SetInterrupt
 include "myconfig.pxi"
 import espressomd.code_info
 from espressomd.utils cimport *
+cimport globals
 
 cdef class Integrator(object):
     """
@@ -79,6 +81,12 @@ cdef class Integrator(object):
                 reuse_forces, 1, bool, "reuse_forces has to be a bool")
 
             _integrate(steps, recalc_forces, reuse_forces)
+
+            if globals.set_py_interrupt:
+                PyErr_SetInterrupt()
+                globals.set_py_interrupt = False
+                PyErr_CheckSignals()
+
         elif self._method == "STEEPEST_DESCENT":
             minimize_energy_init(self._steepest_descent_params["f_max"],
                                  self._steepest_descent_params["gamma"],
--- a/src/core/integrate.hpp
+++ b/src/core/integrate.hpp
@@ -65,6 +65,9 @@ extern int recalc_forces;
     used. */
 extern double verlet_reuse;
 
+/** Communicate signal handling to the Python interpreter */
+extern bool set_py_interrupt;
+
 /*@}*/
 
 /** \name Exported Functions */
--- a/src/core/signalhandling.hpp
+++ b/src/core/signalhandling.hpp
@@ -1,0 +1,72 @@
+/*
+  Copyright (C) 2010,2012,2013,2014,2015,2016 The ESPResSo project
+  Copyright (C) 2002,2003,2004,2005,2006,2007,2008,2009,2010
+    Max-Planck-Institute for Polymer Research, Theory Group
+
+  This file is part of ESPResSo.
+
+  ESPResSo is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  ESPResSo is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+#ifndef SIGNAL_HANDLER_HPP
+#define SIGNAL_HANDLER_HPP
+
+#include <csignal>
+
+#include "errorhandling.hpp"
+
+/** @brief RAII guard for signal handling
+ *
+ * This object saves the current signal handler for @p signal,
+ * replaces it with a custom handler @p handler and restores it on
+ * destruction.
+ */
+class SignalHandler {
+  struct sigaction old_action;
+
+  // Delete all copy and move constructors
+  SignalHandler(SignalHandler &&) = delete;
+  SignalHandler &operator=(SignalHandler &&) = delete;
+  SignalHandler(SignalHandler const &) = delete;
+  SignalHandler &operator=(SignalHandler const &) = delete;
+
+public:
+  /** @brief Constructor
+   *
+   * @param[in] signal  Number of signal to replace
+   * @param[in] handler Function to handle the signal
+   */
+  SignalHandler(int signal, void (*handler)(int)) {
+    struct sigaction new_action;
+    new_action.sa_handler = handler;
+    sigemptyset(&new_action.sa_mask);
+    new_action.sa_flags = 0;
+
+    if (sigaction(SIGINT, &new_action, &old_action) < 0) {
+      runtimeErrorMsg() << "Failed to replace signal handler!";
+    }
+  }
+
+  /** @brief Destructor
+   *
+   * Restores the handler which was active at the time of
+   * construction.
+   */
+  ~SignalHandler() {
+    if (sigaction(SIGINT, &old_action, nullptr) < 0) {
+      runtimeErrorMsg() << "Failed to restore signal handler!";
+    }
+  }
+};
+
+#endif // SIGNAL_HANDLER_HPP
--- a/src/python/espressomd/globals.pxd
+++ b/src/python/espressomd/globals.pxd
@@ -54,6 +54,7 @@ cdef extern from "integrate.hpp":
     extern double sim_time
     extern double verlet_reuse
     extern double skin
+    extern bool set_py_interrupt
 
 cdef extern from "lattice.hpp":
     extern int lattice_switch
--- a/src/python/espressomd/integrate.pxd
+++ b/src/python/espressomd/integrate.pxd
@@ -33,7 +33,6 @@ cdef inline int _integrate(int nSteps, int recalc_forces, int reuse_forces):
     with nogil:
         return python_integrate(nSteps, recalc_forces, reuse_forces)
 
-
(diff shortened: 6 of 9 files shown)


src/core/initialize.cpp
```diff
@@ -88,6 +90,8 @@ void on_program_start() {
   cuda_init();
 #endif
 
+  std::signal(SIGINT, [](int) { std::abort(); });
```
hmenke: We could also call `std::exit(EXIT_FAILURE)` here, but it doesn't really matter, does it?

KaiSzuttor:
Does not work for me

src/core/initialize.cpp
mkuron: We can only do `abort`. `exit` is not signal-safe.

mkuron (changes requested):
Not sure if this is a good idea. People may want to have Python signal handlers in their script. Your solution bypasses them.

 Ideally, you would set a `volatile` variable to `true` from the handler, check for that from within `integrate` and stop integrating after the current time step, raising an exception to Python.

fweik:
`volatile` -> `atomic`

hmenke:
~Alternatively we could have a call to [`PyErr_CheckSignals`](https://docs.python.org/2.7/c-api/exceptions.html#c.PyErr_CheckSignals) in the integration loop if you want ESPResSo to properly interact with Python signal handlers.~

EDIT: This cannot work because `python_integrate` releases the GIL.

EDIT 2:  I think @mkuron's proposal could work.  Is it something like this you had in mind?
```cpp
// --- initialize.hpp ---
bool signal_caught() noexcept;

// --- initialize.cpp ---
namespace { volatile std::sig_atomic_t ctrl_C; }

bool signal_caught() noexcept {
    return ctrl_C == 1;
}

void on_program_start() {
    //...
    std::signal(SIGINT, [](int) {
            // Indicate that signal was caught
            ctrl_C = 1;
            // Reset signal handler
            std::signal(SIGINT, SIG_DFL);
            // Re-raise the signal for Python to catch
            std::raise(SIGINT);
        });
    //...
}

// --- integrate.cpp ---
void integrate_vv(int n_steps, int reuse_forces) {
    //...
    for (int step = 0; step < n_steps; step++) {
        //...
        if (check_runtime_errors() || signal_caught())
            break;
    }
    //...
}
```

EDIT 3:  This doesn't work either.  The only combination where Python signal handler is not overwriting the internal one is when the Python signal handler is registered before `import espressomd` but then the Python signal handler is overwritten of course.
```python
import numpy as np
import signal

def handler(signum, frame):
    # Is never called :(
    print("Caught signal", signum)

signal.signal(signal.SIGINT, handler)

import espressomd

system = espressomd.System(box_l=[100,100,100])
system.time_step = 0.01
system.cell_system.skin = 0.1

for i in range(10000):
    system.part.add(pos=np.random.random()*system.box_l)

system.integrator.run(10000)
print("Integration done!")
```

mkuron:
What about
```
PyGILState_STATE gstate = PyGILState_Ensure();
PyErr_CheckSignals();
PyGILState_Release(gstate);
```
in the integrate function?

hmenke:
That comes with a massive performance penalty.  I tried that before when I wanted to have callback functions in the integrate loop.

This project might do what we are looking for: https://github.com/sagemath/cysignals

mkuron:
Cysignals looks like it‘s worth a try. The website doesn‘t really explain how it works or how fast it is, but if it doesn‘t require a GIL, it should be fine.

All the other solutions proposed here have the major disadvantage that they cause Ctrl-C to behave differently during integrate than at any other time.

hmenke:
Have you tried https://cysignals.readthedocs.io/?

mkuron:
Yes, but there are no high-level technical details in it and I don’t want to read the code. I presume it was developed as a high-performance alternative to acquiring a GIL and calling the Python API to check the signal, so it’s absolutel worth a try.

Since this is not yet packaged on most distributions, you should include it as a Git submodule.

hmenke:
Actually Cysignals does not solve our problem.  It takes a different approach:

> When the user presses `CTRL-C` inside `sig_on()`, execution will jump back to `sig_on()` (the first one if there is a stack) and `sig_on()` will raise `KeyboardInterrupt`.

That is to say it also overwrites the signal handler and registers its own one for Ctrl+C but after catching restores the original one and raises an exception.  So Python signal handlers continue to be ignored in Cython functions.  Furthermore I cannot build it on Ubuntu 16.04, I suspect Cython is too old.

We could just implement something similar to Cysignals by replacing the signal handler in the integration loop and restoring it after: https://www.gnu.org/software/libc/manual/html_mono/libc.html#Advanced-Signal-Handling

mkuron:
> That is to say it also overwrites the signal handler and registers its own one for Ctrl+C but after catching restores the original one and raises an exception.

To me that sounds exactly like what we need. What am I misunderstanding here?

Also, calling `std::abort` is definitely not what we want — that would kill an interactive interpreter too and prevent any kind of custom signal handling (for checkpointing etc.). If we can‘t use cysignals or something like it, we should leave everything as-is and document that any signal handling will be postponed until after `integrate` returns, so one should not integrate for too many steps at once.

hmenke:
Hm... I don't think there is an easy way to feed signals back to the Python interpreter if they were intercepted in a library call.

There is a way to signal `SIGINT` to the Python interpreter by calling `PyErr_SetInterrupt()` followed by `PyErr_CheckSignals()` but the latter one requires a GIL.  I don't think this would be signal-safe:
```cpp
SignalHandler sa(SIGINT, [](int) {
    PyGILState_STATE gstate = PyGILState_Ensure();
    PyErr_SetInterrupt(); // indicate SIGINT to the interpreter
    PyErr_CheckSignals();
    PyGILState_Release(gstate);
});
```

hmenke:
Now this works alright for me:
```python
import numpy as np
import espressomd
import signal
import time

def handler(signum, frame):
    print("Caught signal", signum)

signal.signal(signal.SIGINT, handler)

system = espressomd.System(box_l=[100,100,100])
system.time_step = 0.01
system.cell_system.skin = 0.1

print("Setting up particles")
for i in range(1000):
    system.part.add(pos=np.random.random()*system.box_l)
time.sleep(4)

print("Starting integration")
system.integrator.run(1000)
print("Integration done!")

time.sleep(4)
print("Bye :-)")
```
Output
```
$ ./pypresso test.py
Setting up particles
^CCaught signal 2
Starting integration
^CCaught signal 2
Integration done!
^CCaught signal 2
Bye :-)
```

hmenke:
I think it works on a satisfactory level now.  They only downside we have still is that if one single integration step takes very long it seems as if Ctrl+C doesn't do anything because the signal is only propagated back once at the end of the integration loop.

I could adapt the signal handler such that pressing Ctrl+C twice (or any other number of times) calls `std::abort`.

CMakeLists.txt
```diff
@@ -180,6 +180,9 @@ if (WITH_CUDA)
 endif(WITH_CUDA)
 
 find_package(PythonInterp REQUIRED)
+find_package(PythonLibs REQUIRED)
+include_directories(${PYTHON_INCLUDE_PATH})
+link_libraries(${PYTHON_LIBRARIES})
```
mkuron: Don‘t use `find_package(PythonLibs)`. There are too many situations where it picks the wrong libraries for the interpreter. Use `distutils`, as is done below.

src/core/integrate.cpp
```diff
@@ -64,6 +65,7 @@
 #include <cstdlib>
 #include <cstring>
 #include <mpi.h>
+#include <Python.h>
```
mkuron: Direct usage of the Python header breaks builds with `-DWITH_PYTHON=OFF`. I don‘t know who needs them, but they are possible. So guards are needed.

testsuite/sigint.py
```diff
@@ -0,0 +1,16 @@
...
+import time
+
+class SigintTest(ut.TestCase):
+    def setUp(self):
+        self.process = subprocess.Popen(['@CMAKE_BINARY_DIR@/pypresso',  '@CMAKE_CURRENT_BINARY_DIR@/sigint_child.py'])
+
+    def test_signal_handling(self):
+        self.process.send_signal(signal.SIGINT)
+        time.sleep(1) # Wait for the signal to arrive and one integration step to be finished
+        self.assertFalse(self.process.poll() == None)
```
mkuron: Please have the signal handler print a message and assert that it has been printed. The test case would pass with your old `std::abort` solution too.

mkuron (changes requested):
Looks like a nice solution.

Don‘t bother with `integrate_vv` calls that take longer than a few seconds. They can‘t be aborted without leaving the simulation in an inconsistent state and are not very common anyway.

CMakeLists.txt
mkuron: This issue is actually what you are seeing happen on Travis right now.

testsuite/sigint.py
hmenke: That is tricky.  I cannot capture stdout because that is a blocking operation and I have to send the signal at the same time.  The return code is also not useful because it is simply set to `-2` because `SIGINT = -2` and the process ended after sending `SIGINT`.  This is what I have tried:

In the child process:
```python
try:
    while True:
        system.integrator.run(1000)
except KeyboardInterrupt:
    sys.exit(42)
```
In the testcase:
```python
self.assertFalse(self.process.poll() == None)
self.assertEqual(self.process.returncode, 42) # AssertionError: -2 != 42
```

hmenke: To be completely honest, I don't see why we need a testcase for something like this.  It's not impacting the rest of the program in any way.  Even if this feature breaks the only thing that happens is that the old behaviour is restored.

mkuron: You could have the child process write a file in the signal handler and check for its presence in the test. Of course, this testcase is only nice to have, but it ensures that nobody ever breaks this feature by accident.

hmenke: I realised that this testcase does not work at all.  Apparently `subprocess` installs its own signal handler (like `mpirun`).  I'll investigate this a little further.

KaiSzuttor: Once people rely on the new behaviour we should also make shure it works in the future.

hmenke: Then we need a different type of testcase.  The Python `subprocess` module simply does not work here as verified by this simple example.
```cpp
#include <csignal>
#include <cstdlib>

volatile static std::sig_atomic_t ctrl_C = 0;

int main() {
    std::signal(SIGINT, [](int) { ctrl_C = 1; });
    for (;;) {
        if (ctrl_C == 1) {
            std::exit(42);
        }
    }
}
```
```
~ $ ./a.out 
^C
~ $ echo $?
42
```
```python
import subprocess
import signal
import time

process = subprocess.Popen(['./a.out'])
process.send_signal(signal.SIGINT)
time.sleep(0.1)
assert(process.poll() is not None)
print(process.returncode)
```
```
~ $ python test.py 
-2
```

hmenke: Related Stack Overflow post: [Python subprocess.Popen return code after killing with signal
](https://stackoverflow.com/questions/34136378)

KaiSzuttor: What about this hacky version?
``` python
#!/usr/bin/env python

import os
import signal

os.system("pypresso sigint_child.py & echo $! > /tmp/myprocessid")
with open('/tmp/myprocessid') as f:
    pid = f.read()

os.kill(pid, signal.SIGTERM)
```

hmenke: That's a great opportunity for arbitrary code execution in the ESPResSo testsuite.  Also how would you obtain the return code this way?

KaiSzuttor: Everyone can add arbitrary python scripts to the testsuite, so this is not actually any worse! You maybe have to fork manually.

src/core/integrate.cpp
```diff
@@ -25,6 +25,10 @@
  *  see \ref integrate.hpp "integrate.hpp".
 */
 
+#ifdef WITH_PYTHON
+#include <Python.h>
+#endif // WITH_PYTHON
```
KaiSzuttor: this comment does not provide much information here

fweik:
I really do not like the python dependence in the core code. I think it would be a much cleaner solution to just exit `integrate_vv`  with an error code or exception if the signal has been caught. Please note that you want to break from the integration in anything resembling a consistent state you have to broadcast the abort to all nodes, and exit on the next synchronization point (announce_resort), you can't just leave on a random point on the master.

fweik (changes requested):
Don't make the core depend on python.

KaiSzuttor:
Whats the status?

fweik (approved):

fweik:
I now it's been a long time, but I still think we want this. I'm happy with this solution.

src/core/integrate.cpp
```diff
@@ -462,6 +469,21 @@ void integrate_vv(int n_steps, int reuse_forces) {
...
+    // Check if SIGINT has been caught.
+    if (ctrl_C == 1) {
+      // Reset the flag.
+      ctrl_C = 0;
+
+      // This function simulates the effect of a SIGINT signal
+      // arriving — the next time PyErr_CheckSignals() is called,
+      // KeyboardInterrupt will be raised. It may be called without
+      // holding the interpreter lock.
+      set_py_interrupt = true;
```
hmenke, line 478: The comment above is nonsense because the function call it refers to was removed earlier.

src/python/espressomd/integrate.pyx
```diff
@@ -78,7 +80,14 @@ cdef class Integrator(object):
...
                 reuse_forces, 1, bool, "reuse_forces has to be a bool")
 
-            _integrate(steps, recalc_forces, reuse_forces)
+            if (_integrate(steps, recalc_forces, reuse_forces)):
+                handle_errors("Encoutered errors during integrate")
+
+            if globals.set_py_interrupt:
+                PyErr_SetInterrupt()
+                globals.set_py_interrupt = False
+            PyErr_CheckSignals()
```
hmenke: `PyErr_CheckSignals()` could probably go into the `if` block as well.  Might save a few cycles.

hmenke:
I can review and rebase this tomorrow.

hmenke:
Rebase complete, `make check` passes locally with `maxset.hpp`.

fweik (approved):

PR #1895, opened 2018-03-08
warned about the usage of the word catalyst in feature swimmer_reaction

jonaslandsgesell:


mkuron:
We just discussed this with @jdegraaf. Please rename everything to "reactive swimmer". It does model a catalyst in a certain abstract way. It certainly is not chemically or thermodynamically correct, so it should only be used for its intended purpose.

This paragraph is a bit too defamatory for my taste and not entirely true:

> Be careful the model makes far-fetched usage of the word catalyst. This usage of the word cannot be brought into agreement with the correct usage of the word catalyst.

It is fine to point out that this feature has only a very narrow application (i.e. a specific catalytic swimmer model). But it is definitely not as useless as that paragraph suggests.

fweik:
Currently there are two implementations, the old one by Georg and the newer default one by @hmenke and @jdegraaf. Do we need to keep both? In the code there are comments that suggest that the original one is flawed.

jdegraaf:
Both have their weaknesses, but the former more so than the latter. If you
wish to clean up the code, I think getting rid of the former is acceptable,
as I will not be using it anymore and I don't know anyone who does, except
maybe Pierre?

pdebuyl:
I don't use ESPResSo's swimmer features myself. I tend to use the purpose-built [RMPCDMD](https://github.com/pdebuyl-lab/RMPCDMD) code for that :-)

I found easily the catalytic feature that has the tutorial. Where is the other one in the docs? I would consider it more consistent for end-users to have only one of the features present. But it also depends on the testing and on the use of the features. Do you have an idea about this?

If it is of interest, I know well the MPCD-type techniques to make chemical nanomotors move in the case you'd like me to have a look at the docs. (after clearing up what stays and what doesn't, of course).

jdegraaf:
Hi Pierre,

After much discussion with Jonas the last time, we moved the other
catalytic feature out of the documentation and made it the standard to
invoke the swap move. You can still access it by uncommenting bits of the
interface, I believe. Though Henri might know more, as he is the one that
implemented this. The original behavior is, I believe, similar to the one
you are using.

Best Wishes,

Joost

fweik:
@jonaslandsgesell please apply the name-change everywhere.

jonaslandsgesell:
Is the name swimmer_reaction ok? or do you want the feature to be named reactive_swimmer?

jdegraaf:
Hello Jonas,

I think swimmer_reaction is better, since reactive_swimmer might be
associated with the idea that this naturally has self-propulsion in it,
which it does not.

Best Wishes,

Joost

jonaslandsgesell:
test_checkpoint fails here although not touched

fweik (approved):

src/core/particle_data.cpp
```diff
@@ -812,6 +812,9 @@ int remove_particle(int p_id) {
   particle_node[p_id] = -1;
   mpi_remove_particle(pnode, p_id);
 
+  auto particle_node_iterator=particle_node.find(p_id);
+  particle_node.erase(particle_node_iterator);
```
fweik: Do `erase(p_id)`, your code is invalid if `p_id` is not in the map.

fweik (changes requested):

fweik (approved):

PR #1900, opened 2018-03-10
Lb cleanup

fweik:
Removed some unused code from LB. There is still more (D3Q19, PULL/PUSH).


KaiSzuttor (approved):

KaiSzuttor:
why does removing code *decrease* coverage?

hmenke:
Because the statistical tests are non-deterministic and the fluctuation is much larger than the impact of your removal.

PR #1907, opened 2018-03-15
Script params

### LINKED ISSUE #1880: Script interface doesn't terminate simulation due to unknown parameters
s.constraints.add(shape=Wall(normal=(-1,0,0),distance=1),only_positive=0,penetrable=True)
Wall doesn't know "distance". But while an exception is printed, the simulation is not terminated

KaiSzuttor:
Fixes #1880 

Description of changes:
 - generate a call to PyErr_Occurred() by using except *

### DIFF
--- a/src/python/espressomd/script_interface.pxd
+++ b/src/python/espressomd/script_interface.pxd
@@ -98,5 +98,5 @@ cdef class PScriptInterface:
     cdef set_sip(self, shared_ptr[ScriptInterfaceBase] sip)
     cdef variant_to_python_object(self, Variant value)  except +
     cdef Variant python_object_to_variant(self, value)
-    cdef map[string, Variant] _sanitize_params(self, in_params)
+    cdef map[string, Variant] _sanitize_params(self, in_params) except *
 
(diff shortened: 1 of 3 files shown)


fweik (approved):

hmenke:
Did you perhaps mean `except +` instead of `except *`?  If I understand the [documentation](https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#error-return-values) correctly, then `*` always raises an exception, regardless of the return value, whereas `+` catches C++ exceptions and rethrows them as Python exceptions.

fweik:
No that is correct `except +` catches C++ exceptions, but here this is about a python exception raised in a cdef function. `exept  *` does not always raise an error, it just invokes the python error checking code after each call. `PyErr_Occurred` checks the error indicator.

KaiSzuttor:
No, see

> If you don’t do anything special, a function declared with cdef that does not return a Python object has no way of reporting Python exceptions to its caller. If an exception is detected in such a function, a warning message is printed and the exception is ignored

The problem is that we raise a python exception within a ```cdef``` function. So the python interpreter needs to check for the exception for every call. ```except +``` only works for exceptions that are thrown from external defined c functions.

KaiSzuttor:
damn, @fweik was faster

PR #1910, opened 2018-03-15
Lb cleanup

fweik:
Fixes #

Description of changes:
 - 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


src/core/lb.cpp
```diff
@@ -2512,78 +2380,41 @@ inline void lb_relax_modes(Lattice::index_t index, double *mode) {
 }
 
 inline void lb_thermalize_modes(Lattice::index_t index, double *mode) {
-  double fluct[6];
+  const double rootrho = std::sqrt(
+      std::fabs(mode[0] + lbpar.rho * lbpar.agrid * lbpar.agrid * lbpar.agrid));
 #ifdef GAUSSRANDOM
-  double rootrho_gauss =
```
KaiSzuttor, line 2517: Why is this not needed anymore?

KaiSzuttor (approved):

src/core/lb.cpp
fweik: This enters thru the calculation of `pref` in L 2398. They only differ in the variance, the rest is identical.

PR #1911, opened 2018-03-15
core: remove duplicate call to correlator update.

KaiSzuttor:
Description of changes:
 - remove duplicate correlator update
 
Tests that use the correlator auto_update feature still pass which indicates that they do not test for the double update.


KaiSzuttor:
codecov is in a bad mood today... whatever I do, it tells me that I have decreased coverage :pensive:

PR #1912, opened 2018-03-16
Use Gitlab Container registry for CI

mkuron:
Depends on https://github.com/espressomd/docker/pull/26


KaiSzuttor (changes requested):
You could use a global variable and prepend that to the current image names, that makes it easier to change the path in the future

KaiSzuttor (approved):

mkuron:
@KaiSzuttor, we can't use variables there: https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/14332

KaiSzuttor:
I don't understand why...
https://docs.gitlab.com/ce/ci/yaml/README.html#variables

mkuron:
There is some new stuff in the EspressoScriptInterface that crashes the Intel compiler: https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/14526, somehow related to `boost::is_constructible`. No idea where that is used.

mkuron:
Ready to merge.

fweik (approved):

CMakeLists.txt
```diff
@@ -292,10 +292,20 @@ if(WITH_TESTS)
...
+else()
+  set(BOOST_MINIMUM_VERSION "1.53.0")
+endif()
+
+find_package(Boost ${BOOST_MINIMUM_VERSION} REQUIRED ${BOOST_COMPONENTS})
 include_directories(SYSTEM ${Boost_INCLUDE_DIRS})
 list(APPEND LIBRARIES ${Boost_LIBRARIES})
 
+if(Boost_VERSION VERSION_GREATER 106399 AND Boost_VERSION VERSION_LESS 106500)
+  message(FATAL_ERROR "Boost 1.64 is unsupported")
```
KaiSzuttor, line 306: Why is Boost 1.64 not supported? It's currently the standard version in fedora...

mkuron: Its Boost.MPI is incompatible with its Boost.Serialization. Boost 1.64 simply cannot be compiled if MPI is enabled. There is no way to make it work besides upgrading or downgrading. See https://svn.boost.org/trac10/ticket/12723 .

pkreissl: Removing these lines, I can compile on my fedora 27, with boost version 1.64.
All tests run successfully (in parallel), except for `testsuite/thermalized_bond.py`, which only succeeds when run single-core.

mkuron: Then Fedora must have backported that patch but left the Boost version unchanged. Can you check Fedora's source package repository for that? I guess we would need some way to detect whether we are using a vanilla Boost 1.64 or a patched one if we want to warn the user that their Boost version may be broken. If we don't do that, they will just get hard-to-understand error messages.

PR #1914, opened 2018-03-19
Fix Boost CMake on Fedora 27

mkuron:
Reported by @KaiSzuttor and @pkreissl. Issue introduced in #1912.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -303,7 +303,12 @@ include_directories(SYSTEM ${Boost_INCLUDE_DIRS})
 list(APPEND LIBRARIES ${Boost_LIBRARIES})
 
 if(Boost_VERSION VERSION_GREATER 106399 AND Boost_VERSION VERSION_LESS 106500)
-  message(FATAL_ERROR "Boost 1.64 is unsupported")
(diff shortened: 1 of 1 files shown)


KaiSzuttor (approved):

KaiSzuttor:
Maybe we should add a fedora docker image to the CI

pkreissl (approved):

mkuron:
@KaiSzuttor, at some point we decided to only test on long-term supported distributions. That means Debian, Ubuntu LTS, OpenSUSE (for SLES) and CentOS (for RedHat). If we added all other common ones (Debian unstable, Ubuntu 17.10, Arch Linux, Fedora 26 and 27, Gentoo), we would need to double the number of build machines. And considering that FreeBSD has more marketshare than some of them, we would have to start testing on that too.

I think testing on CentOS ensures that we are at least compatible with things like the file layout of the distribution (CentOS 7 is a mix of Fedora 19 and 20). We test on different Boost versions between 1.53 and 1.66, so we cover it reasonably well. It just so happens that Fedora 27 made an unfortunate choice of Boost version and then patched it up to not be completely broken.

PR #1923, opened 2018-03-22
DD resort refactor

fweik:
Description of changes:
 - DD resort refactored into multiple functions and simplified.
 - Fixed layered with full periodic boundaries (hopefully)
 - Extended pair finding test
 - Resort now works with folded coords
 - Moved local sort to cells level
 - Moved particle position reset from short-range loop (?) to sorting code.
 - Minor performance improvements.


src/core/cells.cpp
```diff
@@ -302,6 +302,73 @@ int cells_get_n_particles() {
...
 
+namespace {
+void fold_and_reset(Particle &p) {
+  fold_position(p.r.p, p.l.i);
+
+  for (int i = 0; i < 3; i++) {
+    p.l.p_old[i] = p.r.p[i];
+  }
+}
+
```
KaiSzuttor, line 309: could you please add docstrings to the new functions?

src/core/cells.cpp
```diff
@@ -311,24 +378,40 @@ void cells_resort_particles(int global_flag) {
...
 
 #ifdef ADDITIONAL_CHECKS
   /* at the end of the day, everything should be consistent again */
+  /* These have to be run before we put the particles that we already
+   * know are displaced back, otherwise they will trigger different errors.*/
   check_particle_consistency();
   check_particle_sorting();
 #endif
 
+  if (0 != displaced_parts.n) {
```
KaiSzuttor, line 406: this could be a function

src/core/domain_decomposition.cpp
```diff
@@ -904,243 +841,117 @@ void dd_topology_release() {
...
+    } else {
+
+      append_unindexed_particle(&rest, std::move(src.part[i]));
+    }
+  }
+
+  realloc_particlelist(&src, src.n = 0);
+}
+
+void move_left_or_right(ParticleList &src, ParticleList &left,
```
KaiSzuttor, line 775: please add a docstring here too

KaiSzuttor (changes requested):

fweik:
@KaiSzuttor

fweik:
@RudolfWeeber, we maybe should also look at some benchmarks to check that I did not screw up too much. Also I think we should find some permanent solution for the NPT test, it has a false positive every time the order the PRNG is applied changes. This is not what it is supposed to check.

fweik:
@RudolfWeeber ready from my side. The NPT issue remains, but I think it's a false positive.

RudolfWeeber:
I think, this PR can be merged.

PR #1927, opened 2018-03-23
CI: do not perform checks if merge.

KaiSzuttor:
Description of changes:
 - remove check jobs from merge branch "doc"

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -39,6 +39,8 @@ status_pending:
 
 no_cuda_default:
   stage: build
+  except:
+    - doc
   script:
     - export with_cuda=false
     - export myconfig=default with_coverage=true
@@ -46,6 +48,8 @@ no_cuda_default:
 
 no_cuda_maxset:
   stage: build
+  except:
+    - doc
   script:
     - export with_cuda=false
     - export myconfig=maxset with_coverage=true
@@ -53,6 +57,8 @@ no_cuda_maxset:
 
 no_cuda_maxset_python3:
   stage: build
+  except:
+    - doc
   image: gitlab.icp.uni-stuttgart.de:4567/espressomd/docker/ubuntu-python3:latest
   script:
     - export with_cuda=false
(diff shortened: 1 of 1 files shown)


KaiSzuttor:
why should the registry *not* be public?

mkuron:
Because it contains the Intel image, which we cannot redistribute. Your revert commit didn‘t work, by the way. Just do `git push -f 967cda1:check_skip` to delete it, it isn‘t even needed.

mkuron (approved):

fweik:
Hm, the merge version should be tested somehow. I'd prefer to merge the PRs before mergin, as travis does.

KaiSzuttor:
We somehow have to distinguish a merge (which triggers a documentation deployment) and pull requests on ~~travis~~gitlab. Therefore, we push merges to a special branch named ``doc``. So how do you want to trigger a job that only runs for merge?

KaiSzuttor:
Btw. travis does not merge, it checks out the PR hash directly.

mkuron:
> We somehow have to distinguish a merge (which triggers a documentation deployment) and pull requests on travis. Therefore, we push merges to the a special branch named doc.

You probably meant Gitlab. Travis doesn't know about the doc branch.

> Btw. travis does not merge, it checks out the PR hash directly.

Well, Travis looks like it builds the merged commit (301d68b) because that is what it shows in the top box. What it actually does though is this, which is identical to what Gitlab does:
```
git fetch origin +refs/pull/1927/merge:
git checkout -qf FETCH_HEAD
```

By the way, what was the original reason why you didn't want to build the documentation on pull requests?

fweik:
The doc build includes deployment of the docs, which you would not want run on PRs.

KaiSzuttor:
Yes, it's only the deployment.

mkuron:
Ok, so let's get this merged.

KaiSzuttor:
@fweik please merge if you are happy

Issue #1916, opened 2018-03-19
`testsuite/thermalized_bond` test fails when run in parallel (4, 5 cores)


pkreissl:
```
mpiexec -n 4 ./pypresso ../testsuite/thermalized_bond.py
```
yields
```
AssertionError: 0.016005163491671776 not less than or equal to 0.015
```
(also for `-n 5` cores).
Works again for `6` or `7` cores.
For 8 cores, it fails with different error, see #1915

KonradBreitsprecher:
The problem here was the decreased sample length.. should be fixed with #1930

Issue #1915, opened 2018-03-19
`testsuite/thermalized_bond` test fails when run in parallel on 8 cores


pkreissl:
```
mpiexec -n 8 ./pypresso ../testsuite/thermalized_bond.py
```
yields
```
.4: transfer data type wrong
6: transfer data type wrong
2: transfer data type wrong
0: transfer data type wrong
--------------------------------------------------------------------------
MPI_ABORT was invoked on rank 4 in communicator MPI_COMMUNICATOR 4
with errorcode 1.
```

KonradBreitsprecher:
This is vague, but it seems to be related to n_squared, ghosts, ghosts_have_v.
Same error happens with 8 cores and the DPD testcase using n_squared.
Both set ghosts_have_v. Will investigate further...

fweik:
I'm looking into this right now.

PR #1930, opened 2018-03-28
original loop length for thermalized bond test

KonradBreitsprecher:
Back to 16000 configurations for the thermalized bond test. 
Runs in approx. 16 seconds.

### DIFF
--- a/testsuite/thermalized_bond.py
+++ b/testsuite/thermalized_bond.py
@@ -94,7 +94,7 @@ the single component Maxwell distribution. Adapted from langevin_thermostat test
         self.system.integrator.run(50)
         
         # Sampling
-        loops = 160
+        loops = 16000
         v_stored = np.zeros((N2 * loops, 3))
         for i in range(loops):
             self.system.integrator.run(2)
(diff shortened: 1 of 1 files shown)


KonradBreitsprecher:
Also should fix #1916 #1915 .

KaiSzuttor:
Should or does fix?

KonradBreitsprecher:
It does fix #1916, not #1915

pkreissl:
True, it does fix #1916 (#1915 was closed via #1936).
However, runtime of the test seems to increase significantly when executed on more than one core. By default, the test is performed on up to 4 cores, which results in ~ **35 s** runtime (vs. 16 s single core).
Is it possible to speed the test up somehow? 
Also, why does this fix work? That is, why are 100 x sampling steps necessary for 4/5 cores? Shouldn't the results be the same independent of number of cores (at least almost)?

KonradBreitsprecher:
The testcase by itself is non-deterministic, so it's a matter of setting the random generator seed to get it stable on all cores. This should be assured by:
```
system.seed = range(system.cell_system.get_state()["n_nodes"])
np.random.seed(42)
```
I could reduce the number of configurations to 1600 with successful test runs up to 8 cores, also reducing the runtime significantly (~4s).

fweik:
I think 4s is acceptable, so if it passed I'll merge.

fweik:
Btw. the correlation time for Langevin is 1 / gamma, so about 12.5 time steps here. So the samples are all seriously correlated...

fweik:
The checkpointing issue in the other pr is fixed somewhere else.

PR #1941, opened 2018-04-03
observables: extended LB observables for usage with CPU LB.

KaiSzuttor:
Description of changes:
 - Cylindrical LB profile extended for CPU LB plus test


KaiSzuttor:
@fweik ready for review

testsuite/observable_cylindricalLB.py
```diff
@@ -79,6 +76,7 @@ def set_particles(self):
         self.system.part.add(pos=self.positions)
 
     def set_fluid_velocity(self):
+        del self.positions[:]
```
fweik, line 79: What is that supposed to do?

KaiSzuttor: it removes positions stored as a member

fweik: Ah sorry, i thought this was on the system. Never mind.

fweik:
@KaiSzuttor this can be merged even without the new lb interpolation code right?

KaiSzuttor:
With the restriction that it only works for 1 node in case of CPU LB, id rather wait till we have a working function for getting the interpolated velocity on multiple nodes

KaiSzuttor:
I think we should merge this. The fixes for the observables are quite important.

PR #1948, opened 2018-04-13
core: Removed MULTI_TIMESTEP feature.

fweik:
While this is an interesting feature, it is not sustainable in the current form. This has no tests, but wide and unclear side effects. It is not at all clear e.g. how it interacts with ROTATION, and many others.


hmenke (approved):

KaiSzuttor (approved):

RudolfWeeber:
I agree with @fweik's assessment. Final decision will be made at the next off-line meeting on Wed Apr 25th.

The feature can be easily re-introduced, once the internal scaling of forces and velocities is removed from Es.

If someone is actively using MULTI_TIMESTEP right now, please speak up before next Wednesday.

fweik:
The offline meeting opted for removal.

PR #1952, opened 2018-04-17
interfaced bounce back boundary condition for simulating hard walls with MD

jonaslandsgesell:
also applied clang-format....


jonaslandsgesell:
I will fix this tomorrow.

KaiSzuttor (commented):
Could you please add a documentation?

src/core/constraints/ShapeBasedConstraint.cpp
```diff
@@ -7,74 +7,75 @@
...
-    break;
+  if ((norm) > 0) {
+    p->r.p[0] = p->r.p[0] - 2 * vec[0];
+    p->r.p[1] = p->r.p[1] - 2 * vec[1];
+    p->r.p[2] = p->r.p[2] - 2 * vec[2];
+
+    /* vec seems to be the vector that points from the wall to the particle*/
+    /* now normalize it */
+    switch (std::abs(m_reflection_type)) {
+    case (int)ReflectionType::NORMAL: {
```
fweik: Please don't use c-style casts.

src/core/constraints/ShapeBasedConstraint.cpp
```diff
@@ -7,74 +7,75 @@
...
+      vec[2] /= norm;
+      /* calculating scalar product */
+      double normal_velocity =
+          vec[0] * p->m.v[0] + vec[1] * p->m.v[1] + vec[2] * p->m.v[2];
+      /* now add twice the normal component to the velcity */
+      p->m.v[0] = p->m.v[0] - 2 * vec[0] * normal_velocity;
+      p->m.v[1] = p->m.v[1] - 2 * vec[1] * normal_velocity;
+      p->m.v[2] = p->m.v[2] - 2 * vec[2] * normal_velocity;
+    } break;
+    case (int)ReflectionType::NORMAL_TANGENTIAL:
```
fweik: Please don't use c-style casts.

src/core/constraints/ShapeBasedConstraint.cpp
```diff
@@ -7,74 +7,75 @@
...
+      /* if bounce back, invert velocity */
+      p->m.v[0] = -p->m.v[0];
+      p->m.v[1] = -p->m.v[1];
+      p->m.v[2] = -p->m.v[2];
+      break;
+    case (int)ReflectionType::NONE:
```
fweik: Please don't use c-style casts.

src/core/constraints/ShapeBasedConstraint.cpp
```diff
@@ -105,19 +106,20 @@ void ShapeBasedConstraint::add_force(Particle *p, double *folded_pos) {
...
+    } else if (dist <= 0) {
+      if (m_penetrable && !m_only_positive && (dist != 0.0)) {
         calc_non_bonded_pair_force(p, &part_rep, ia_params, vec, -1.0 * dist,
                                    dist * dist, force, torque1, torque2);
       }
-    } else {
-      if (m_reflection_type != ReflectionType::NONE) {
-        reflect_particle(p, vec, folded_pos);
-      } else {
+      if (m_reflection_type != (int)ReflectionType::NONE) {
```
fweik: Please don't use c-style casts.

src/core/constraints/ShapeBasedConstraint.cpp
```diff
@@ -105,19 +106,20 @@ void ShapeBasedConstraint::add_force(Particle *p, double *folded_pos) {
...
+        reflect_particle_from_flat_wall(p, vec, folded_pos);
```
fweik: Please leave the name as it is. This is now a lie, as this can be activated on any shape.

src/core/constraints/ShapeBasedConstraint.hpp
```diff
@@ -17,7 +17,7 @@ class ShapeBasedConstraint : public Constraint {
 
   ShapeBasedConstraint()
       : m_shape(std::make_shared<Shapes::NoWhere>()),
-        m_reflection_type(ReflectionType::NONE), m_penetrable(false),
+        m_reflection_type((int)ReflectionType::NONE), m_penetrable(false),
```
fweik: Please don't use c-style casts.

src/core/constraints/ShapeBasedConstraint.hpp
```diff
@@ -41,7 +41,7 @@ class ShapeBasedConstraint : public Constraint {
 
   Shapes::Shape const &shape() const { return *m_shape; }
 
-  ReflectionType const &reflection_type() const;
+  int &reflection_type(){return m_reflection_type;}
```
fweik: Please leave the type as it is, and convert the parameters at the python/si level from a string parameter to the corresponding enum val.

src/core/constraints/ShapeBasedConstraint.hpp
```diff
@@ -57,13 +57,13 @@ class ShapeBasedConstraint : public Constraint {
...
 private:
   /** Private methods */
-  void reflect_particle(Particle *p, const double *distance_vector,
+  void reflect_particle_from_flat_wall(Particle *p, const double *distance_vector,
                         const double *folded_pos) const;
 
   /** Private data members */
   std::shared_ptr<Shapes::Shape> m_shape;
 
-  ReflectionType m_reflection_type;
```
fweik: Please leave the type as it is, and convert the parameters at the python/si level from a string parameter to the corresponding enum val.

src/core/constraints/ShapeBasedConstraint.hpp
```diff
@@ -57,13 +57,13 @@ class ShapeBasedConstraint : public Constraint {
...
+  int m_reflection_type;
```
fweik: Please leave the type as it is, and convert the parameters at the python/si level from a string parameter to the corresponding enum val.

fweik (changes requested):

jonaslandsgesell:
Documentation is given in http://espressomd.org/html/doc/constraints.html?highlight=bounce%20back

src/core/constraints/ShapeBasedConstraint.cpp
jonaslandsgesell: Not on any shape the implemented behaviour corresponds to the correct reflection. The implemented behaviour is only correct for flat walls.

fweik: Yes, but this is also not what I said.

src/core/reaction_ensemble.hpp
```diff
@@ -309,7 +309,8 @@ class ConstantpHEnsemble : public ReactionAlgorithm {
 class WidomInsertion : public ReactionAlgorithm {
 public:
     double measure_excess_chemical_potential(int reaction_id);
-
+    int do_reaction(int reaction_steps)override {return 0;};
```
fweik, line 312: This seems unrelated, was this on purpose?

fweik (approved):

fweik:
I still dislike the fact that reflecting and interacting constraints are the same thing.

src/core/reaction_ensemble.hpp
jonaslandsgesell: yep this is unrelated. do_reaction for the widom insertion method has no meaning. therefore i made it an empty function.

jonaslandsgesell:
Closing this because we decided to remove it. It is not energy conserving and moving particles (via reflecting them) may happen without communicating the move, nor updating cell lists. Rewrite of the feature needed. In the meantime one can use Monte-Carlo moves (contained in the reaction_ensemble module) to simulate hard walls if desired.

PR #1961, opened 2018-04-27
Accumulator

KaiSzuttor:
Description of changes:
 - added checkpointable average and variance calculator
 - moved correlator to accumulators

PR Checklist
------------
 - [x] Tests?
   - [x] Interface
   - [x] Core 
 - [x] Docs?


KaiSzuttor:
Checkpoint test will follow soon

src/core/utils/Accumulator.hpp
```diff
@@ -0,0 +1,111 @@
+#ifndef CORE_UTILS_ACCUMULATOR
+#define CORE_UTILS_ACCUMULATOR
+
+#include <boost/archive/binary_iarchive.hpp>
```
fweik: Why are those needed?

src/core/utils/Accumulator.hpp
```diff
@@ -0,0 +1,111 @@
...
+#include <boost/archive/binary_oarchive.hpp>
+#include <boost/serialization/string.hpp>
+#include <boost/serialization/vector.hpp>
+#include <boost/iostreams/device/array.hpp>
+#include <boost/iostreams/stream.hpp>
+
+namespace Utils {
+
+template <typename T> struct AccumulatorData {
+  AccumulatorData()
```
fweik: Just do `AcumulatorData() = default;`, that is more flexible.

src/core/utils/Accumulator.hpp
```diff
@@ -0,0 +1,111 @@
...
+    auto acc_data_it = m_acc_data.begin();
+    auto data_it = data.begin();
+    for (; acc_data_it != m_acc_data.end() and data_it != data.end();
+         ++acc_data_it, ++data_it) {
+      (*acc_data_it).m_mean_old = (*acc_data_it).m_mean_new = *data_it;
+      (*acc_data_it).m_variance_old = (*acc_data_it).m_variance_new = 0.0;
+    }
+  } else {
+    // Calculate mean value.
+    auto acc_data_it = m_acc_data.begin();
```
fweik: This is  very noisy, I think you can easily be simplified by replacing this loop and the next one by one call to `std::transform`, which would also be more efficient. I think the equivalent call is

```c++
std::transform(acc_data.begin(), acc_data.end(), data.begin(), acc_data.begin(), 
    [](AccumulatorData<double> a, double d) -> AccumulatorData<double> {
      auto const new_mean = a.mean + (d - a.mean) / m_n;
      auto const new_variance = ((m_n-1) * a.variance + (d - a.mean ...
     ....

    return {new_mean, new_variance};
   });
```
This actually allows to make `AccumulatorData` immutable. Also think that you do not need to store two values. After the last loop `m_mean_old == m_mean_new`, the same for the variance.

src/core/utils/Accumulator.hpp
```diff
@@ -0,0 +1,111 @@
...
+#include <boost/iostreams/device/array.hpp>
+#include <boost/iostreams/stream.hpp>
+
+namespace Utils {
+
+template <typename T> struct AccumulatorData {
+  AccumulatorData()
+      : m_mean_old(0.0), m_mean_new(0.0), m_variance_old(0.0),
+        m_variance_new(0.0) {}
+  T m_mean_old;
```
fweik: If think you only need to store one mean and one variance, see below. Also don't call the members `m_` if they are not private.

src/core/utils/Accumulator.hpp
```diff
@@ -0,0 +1,111 @@
...
+
+inline void Accumulator::operator()(const std::vector<double> &data) {
+  if (data.size() != m_acc_data.size())
+    throw std::runtime_error(
+        "The given data size does not fit the initialized size!");
+  ++m_n;
+  if (m_n == 1) {
+    auto acc_data_it = m_acc_data.begin();
+    auto data_it = data.begin();
+    for (; acc_data_it != m_acc_data.end() and data_it != data.end();
```
fweik:
```c++
std::transform(data.begin(), data.end(), m_acc_data.begin(), 
   [] (double d) -> AccumulatorData<double> {
       return {d, 0.0};
  });
```

fweik (changes requested):
Please introduce a common base class for `Correlator` and `ObservableAccumulator`, and give the `ObservableAccumulator` a more telling name. Then you can also get away with only one auto update mechanism, one list and so on...

src/core/accumulators/Correlator.cpp
```diff
@@ -395,10 +393,9 @@ void Correlator::initialize() {
     }
 }
 
-int Correlator::get_data() {
+void Correlator::update() {
   if (finalized) {
     runtimeErrorMsg() << "No data can be added after finalize() was called.";
-    return 0;
```
fweik: I think you have to keep the return statement, when you go on you might corrupt the data.

src/core/accumulators.cpp
```diff
@@ -21,13 +21,12 @@
 #include "integrate.hpp"
 
 namespace Accumulators {
-std::vector<std::shared_ptr<Accumulators::Accumulator>> auto_update_accumulators;
+std::vector<std::shared_ptr<Accumulators::AccumulatorBase>>
+    auto_update_accumulators;
```
KaiSzuttor, line 25: this is a single container for the autoupdates right? I just have to distinguish in the interface.

fweik: Why do you need two classes in the interface?

src/core/accumulators/AccumulatorBase.hpp
```diff
@@ -0,0 +1,25 @@
...
+namespace Accumulators {
+
+class AccumulatorBase {
+public:
+  AccumulatorBase(int delta_N)
+      : delta_N(delta_N){};
+  void auto_update();
+  virtual void update() = 0;
+  // Number of timesteps between automatic updates.
+  int delta_N;
```
fweik: These should be private.

src/python/espressomd/accumulators.py
```diff
@@ -3,8 +3,17 @@
 import numpy as np
 
 
+
+@script_interface_register
+class AccumulatorBase(ScriptInterfaceHelper):
```
fweik: This should not be exposed, you can not even register `AccumulatorBase`, because it is abstract. If you want this on the python level just define an empty python class to derive from.

fweik (changes requested):

KaiSzuttor:
@fweik do you know why it does not build on some distributions?

fweik (approved):

PR #1962, opened 2018-05-02
Gamma rot deadlock

RudolfWeeber:
Fix gamma_rot setter mpi deadlock. Add testcase, which should detect any future deadlocks in particle property setters.

### DIFF
--- a/testsuite/particle.py
+++ b/testsuite/particle.py
@@ -33,8 +33,7 @@ class ParticleProperties(ut.TestCase):
     tol = 1E-9
 
     # Handle for espresso system
-    system = espressomd.System(box_l=[1.0, 1.0, 1.0])
-    system.box_l = [10.0] * 3
+    system = espressomd.System(box_l=[100.0, 100.0, 100.0])
 
     f1 = FeneBond(k=1, d_r_max=5)
     system.bonded_inter.add(f1)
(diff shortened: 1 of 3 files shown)


hmenke (approved):

testsuite/particle.py
```diff
@@ -225,6 +224,22 @@ def get_pos_prop(handle):
...
+        s.part.clear()
+        s.part.add(pos=s.box_l*np.random.random((100,3)))
+
+        # Copy individual properties of particle 0
+        print("If this test hangs, there is an mpi deadlock in a particle property setter." )
+        for p in espressomd.particle_data.particle_attributes:
+            if p in ["director","image_box","node","vs_relative"  ]: continue
+            # Uncomment to identify guilty property
+            #print( p)
+            setattr(s.part[:],p,getattr(s.part[0],p))
```
fweik: This does not work because there are read-only properties.

RudolfWeeber:
On Wed, May 02, 2018 at 06:41:32AM -0700, Florian Weik wrote:
> fweik commented on this pull request.
> 
> 
> 
> > @@ -225,6 +224,22 @@ def get_pos_prop(handle):
>              return handle.pos
>          self.assertRaises(RuntimeError, get_pos_prop, handle_to_non_existing_particle)
>  
> +    def test_parallel_property_setters(self):
> +        s= self.system
> +        s.part.clear()
> +        s.part.add(pos=s.box_l*np.random.random((100,3)))
> +
> +        # Copy individual properties of particle 0
> +        print("If this test hangs, there is an mpi deadlock in a particle property setter." )
> +        for p in espressomd.particle_data.particle_attributes:
> +            if p in ["director","image_box","node","vs_relative"  ]: continue
> +            # Uncomment to identify guilty property
> +            #print( p)
> +            setattr(s.part[:],p,getattr(s.part[0],p))
> 
> This does not work because there are read-only properties.
Hence the exclusions.

fweik:
Yes, but you are at lease missing mass. See https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/17963/raw Also this approach does not scale, it will break every time there is a new read-only property.

RudolfWeeber:
Properties in cdef classes are implemented differently from Python properties, apparently.
So, usual ways to figure out if they have a setter don't apply.
The test now passes over AttributeErrors on setting, which is thorwn for read-only properties.
I changed those read-only properties with setters present but throwing different exception types to AttributeError.

fweik (approved):

PR #1964, opened 2018-05-03
extended widom insertion to allow the user to determine multiple exce…

jonaslandsgesell:
…ss chemical potentials at once

### DIFF
--- a/src/core/reaction_ensemble.cpp
+++ b/src/core/reaction_ensemble.cpp
@@ -1793,8 +1793,8 @@ double WidomInsertion::measure_excess_chemical_potential(int reaction_id) {
   restore_properties(changed_particles_properties, number_of_saved_properties);
 
   double const exponential = exp(-1.0 / temperature * (E_pot_new - E_pot_old));
-  summed_exponentials.at(reaction_id) += exponential;
-  number_of_insertions.at(reaction_id) += 1;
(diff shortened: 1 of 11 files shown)


src/core/reaction_ensemble.cpp
```diff
@@ -1765,6 +1765,15 @@ double ConstantpHEnsemble::calculate_acceptance_probability(
   return bf;
 }
 
+void WidomInsertion::add_reaction (double Gamma, const std::vector<int> &_reactant_types,
```
fweik: Please stick to the naming convention.

fweik (changes requested):

src/core/reaction_ensemble.cpp
jonaslandsgesell: Which one? https://github.com/espressomd/espresso/wiki/Naming-conventions-and-folder-structure I do not see which convention listed there would apply.

jonaslandsgesell: Seems like you refer to https://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier

fweik:
You also have to rename `Gamma` to `gamma`.

fweik (approved):

jonaslandsgesell:
Also fixes #1981.

PR #1966, opened 2018-05-03
core: reaction_ensemble: Fixed performance bug and cleanup.

fweik:
Avoid some copies in reaction ensemble.

@jonaslandsgesell Why do you do floating point comparison in `is_in_list`? As far as I can tell it's only used with ints. Also both of these can be expressed more easily with standard algorithms.


jonaslandsgesell (approved):

jonaslandsgesell:
is_in_list() used to be used in an other context i think. you are right one could use find or find_if here.

PR #1972, opened 2018-05-04
core: dd: Obey min num cells.

### LINKED ISSUE #1897: Setting CellSystem.min_num_cells has no effect
For the DD, this eventually triggers`dd_topology_init`, which resets `min_num_cells` to its default value.
Also, the topology reinit should only be run if if current number of cells is smaller than the new `min_num_cells`.

### LINKED ISSUE #1971: Domain decomposition does not obey min_num_cells
```python
import espressomd

print(system.cell_system.get_state())
```
shows `'min_num_cells': 8`, but `'cell_grid': array([1, 1, 1])`, which are incompatible.
Also it chooses the largest possible cell size, which results in poor performance, because
it maximizes the size of the the ghost layer to the whole box (in every direction).

fweik:
Fixes #1971.Fixes #1897.


RudolfWeeber:
At first glance, it looks like you are removing the cells in
LEES_EDWARDS y direction =1, when min_num_cells >1.
Is that intended?

Otherwise looks good.

fweik:
I don't understand your comment.

RudolfWeeber:
Took a closer look. This pr can be merged.

KaiSzuttor:
NPT test seems not to converge independent of architecture see failing Mac test now.

fweik:
I'll have a look, apparently there is still something  wrong.

RudolfWeeber:
Does this go into 4.0?

fweik:
If we can somehow manage this would be good as it contains two bug fixes.

fweik:
@RudolfWeeber @KaiSzuttor I'm not really sure what to do about the failing npt tests. I don't feel qualified to judge if there is a problem. Also this is not really related to the changes in the PR, which only influence how the cell grid is chosen (which might of course influence NPT with it's constant box changes).

KaiSzuttor:
is this related to #2057 ?

fweik:
No, #2057 was because of the external field/constraints.

PR #1974, opened 2018-05-04
Checkpointing of system.actors

### LINKED ISSUE #1968: actors not checkpointed

### LINKED ISSUE #1969: Massive performance loss after restoring checkpoint
```python
import numpy as np
import espressomd
import time

checkpoint = espressomd.checkpointing.Checkpointing(checkpoint_id="checkpoint")

start_sample = 0
checkpoint.register("start_sample")

system.periodicity = [1,1,1]
system.cell_system.skin = 0.1
system.time_step = 0.01
checkpoint.register("system")

if not checkpoint.has_checkpoints():
    for i in range(10):
        system.part.add(pos=system.box_l * np.random.random(3))
    checkpoint.register("system.part")

if not checkpoint.has_checkpoints():

restarted = False
if checkpoint.has_checkpoints():
    print("Restoring Equilibration Checkpoint")
    print("Checkpoint loaded")
    restarted = True

for k in range(start_sample,200):
    start = time.time()
    system.integrator.run(steps=5000, reuse_forces=restarted)
    print("Step: {}, Time: {}".format(k,time.time() - start))
    restarted = False
    start_sample = k + 1
```
After restoring the checkpoint, integration is a **factor of 5 to 6 slower!**

This is reproducible in the latest [`espressomd/espresso-ubuntu`](https://hub.docker.com/r/espressomd/espresso-ubuntu/) container with `maxset.hpp`.
```
root@9d314b7fbda0:/tmp/espresso/build# git show
commit bbf1a43c6db393fa473a5824385aa87b3563a2a1
Merge: e9b7a5c 069db89
Author: Florian Weik <florianweik@gmail.com>
Date:   Thu May 3 21:38:12 2018 +0200

    Merge pull request #1967 from RudolfWeeber/id
    
    py: Make ParticleHandle.id read-only

root@9d314b7fbda0:/tmp/espresso/build# ./pypresso test.py 
Step: 0, Time: 0.0387508869171
Step: 1, Time: 0.0388331413269
Step: 2, Time: 0.0390479564667
Step: 3, Time: 0.0391790866852
Step: 4, Time: 0.0393280982971
Step: 5, Time: 0.0391201972961
Step: 6, Time: 0.0390129089355
Step: 7, Time: 0.0390889644623
^CTraceback (most recent call last):
  File "test.py", line 35, in <module>
    system.integrator.run(steps=5000, reuse_forces=restarted)
KeyboardInterrupt
root@9d314b7fbda0:/tmp/espresso/build# ./pypresso test.py 
Restoring Equilibration Checkpoint
Checkpoint loaded
Step: 8, Time: 0.229771852493
Step: 9, Time: 0.269323825836
Step: 10, Time: 0.227708816528
Step: 11, Time: 0.228987932205
Step: 12, Time: 0.232725858688
Step: 13, Time: 0.226660966873
Step: 14, Time: 0.22985291481
Step: 15, Time: 0.229024887085
Step: 16, Time: 0.229374885559
^CTraceback (most recent call last):
  File "test.py", line 35, in <module>
    system.integrator.run(steps=5000, reuse_forces=restarted)
KeyboardInterrupt
root@9d314b7fbda0:/tmp/espresso/build# 
```

KaiSzuttor:
Fixes #1968. Fixes #1969

Description of changes:
 - removed system instance from actors, since this leads to circular dependency
 - moved neutrality check to core
 - added setstate and gestate methods for actors
 - added p3m actor test to checkpoint unittest

PR Checklist
------------
 - [x] Tests?
   - [x] Interface
   - [ ] Core 
 - [ ] Docs?


src/core/utils/checks/charge_neutrality.hpp
```diff
@@ -0,0 +1,33 @@
+#include <cmath>
+
+#include "PartCfg.hpp"
```
fweik: No.

src/core/utils/checks/charge_neutrality.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Utils {
+
+inline void check_charge_neutrality(PartCfg &partCfg) {
```
fweik: Make the particle range a template parameter so that the range can be used on any range type.

src/core/utils/checks/charge_neutrality.hpp
```diff
@@ -0,0 +1,33 @@
...
+  double total_charge = 0.0;
+  double min_charge = 0.0;
+  for (auto const &particle : partCfg) {
```
fweik: Reimplement this with `std::accumulate` and `std::min`. Also probably there should be no magic number.

fweik (changes requested):

src/core/utils/checks/charge_neutrality.hpp
KaiSzuttor: ?

src/core/utils/checks/charge_neutrality.hpp
KaiSzuttor: Why? Which magic number?

fweik: Because it is hard to read and overly complex. And `1e-10`

KaiSzuttor: that magic number was already there before

fweik: Isn't that sentence already longer than putting `numeric_limits` instead of the number?

src/core/utils/checks/charge_neutrality.hpp
fweik: As we have iterated multiple times, please do not put dependencies on the core into utils.

KaiSzuttor:
the check fails for some tests due to precision. For elc_vs_mmm2d_neutral.py the ratio of total charge and smallest non zero charge is: 3.3306690738754696e-16 while ```std::numeric_limits<double>::epsilon()``` equals 2.2204460492503131e-16

src/core/utils/checks/charge_neutrality.hpp
```diff
@@ -1,33 +1,33 @@
...
-  double min_charge = 0.0;
-  for (auto const &particle : partCfg) {
-    total_charge += particle.p.q;
-    double abs_charge = std::abs(particle.p.q);
-    if (min_charge == 0.0 and abs_charge != 0.0)
-      min_charge = abs_charge;
-    if (abs_charge < min_charge and abs_charge != 0.0)
-      min_charge = particle.p.q;
+template<typename ParticleRange, typename Particle>
+inline void check_charge_neutrality(ParticleRange &prange) {
```
hmenke: Better use
```cpp
template<typename ParticleRange>
inline void check_charge_neutrality(ParticleRange &prange) {
    using Particle = typename ParticleRange::value_type;
```
then you don't have to specify explicit template arguments.  Even better would be
```cpp
template<int&... ExplicitArgumentBarrier, typename ParticleRange>
inline void check_charge_neutrality(ParticleRange &prange) {
    using Particle = typename ParticleRange::value_type;
```
to completely inhibit giving explicit template arguments but that probably  wouldn't work with Cython.

src/core/utils/checks/charge_neutrality.hpp
```diff
@@ -1,33 +1,33 @@
...
+  std::vector<double> charges;
```
hmenke:
```cpp
charges.reserve(prange.size());
```

src/core/utils/checks/charge_neutrality.hpp
```diff
@@ -1,33 +1,33 @@
...
+  std::transform(prange.begin(), prange.end(), std::back_inserter(charges),
+          [](Particle const &p){return p.p.q;});
+  double total_charge = std::accumulate(charges.begin(), charges.end(), 0.0);
+  std::sort(charges.begin(), charges.end(),
+          [](double a, double b) {return std::abs(a) < std::abs(b);});
+  //Find first non-zero element.
+  auto result = std::find_if(charges.begin(), charges.end(), [](double a){return std::abs(a) > std::numeric_limits<double>::epsilon();});
```
hmenke:
```cpp
return std::abs(a) > 0.0;
```
You're looking at values here, not at differences.  What if the smallest charge is `1e-30`?

src/core/utils/checks/charge_neutrality.hpp
```diff
@@ -0,0 +1,33 @@
+#include <cmath>
+#include <limits>
+
+namespace Utils {
+
+template<typename ParticleRange, typename Particle>
+inline void check_charge_neutrality(ParticleRange &prange) {
```
fweik: I think you can get that easier and with a single pass if you'd use accumulate:
```c++
template<typename ParticleRange>
void check_charge_neutrality(ParticleRange prange) {
    using Particle = typename ParticleRange::value_type;

  auto const sum_and_min = std::accumulate(prange.begin(), prange.end(), {0.0, 0.0}, 
      [](std::pair<double, double> s, const Particle &p) {
           s.first += p.p.q;
           s.second = std::min(s.second, std::abs(p.p.q));

          return s;
      });
   ...
}
```

src/core/utils/checks/charge_neutrality.hpp
fweik: Just out of curiosity, why would you put in the stopper?

src/core/utils/checks/charge_neutrality.hpp
hmenke: You could also use a `std::tuple`, then you could take the result apart into two variables using `std::tie`.  Also I think `prange` should be taken by reference to const.
```cpp
template <typename ParticleRange>
void check_charge_neutrality(ParticleRange const &prange)
{
    using Particle = typename ParticleRange::value_type;

    double sum, min;
    std::tie(sum, min) =
        std::accumulate(prange.begin(), prange.end(), std::make_tuple(0.0, 0.0),
                        [](std::tuple<double, double> s, const Particle &p) {
                            std::get<0>(s) += p.p.q;
                            std::get<1>(s) =
                                std::min(std::get<1>(s), std::abs(p.p.q));
                            return s;
                        });

    ...
}
```

fweik: I think it's more important to make it const. Time we get C++17.

KaiSzuttor:
thanks you for your comments, but none of these refere to the problem with the precision. I wont put any further time into that, since the principal algorithm does not seem to work.

src/core/utils/checks/charge_neutrality.hpp
KaiSzuttor: then you have other issues anyways

src/core/utils/checks/charge_neutrality.hpp
hmenke: To prevent people from trying to interfere with template argument deduction.  Way too often you see things like `std::make_pair<int,std::string>(1,"Hello");`.

jonaslandsgesell:
As a comment on the checking algorithm: the check works - in a physical sense! You need to make sure that your system is electroneutral ---- compared on a scale which is dictated by the smallest absolute nonzero charge. That is done via dividing Q_tot with the minimum absolute nonzero charge. If you do that you do not need to perform a comparison with numeric limits but only a physically reasonable small number: an experimental physicist would consider a system to be charge neutral if the total charge in units of *the minimum absolute nonzero charge* is less than 1e-10 (as it was checked before). This small number is related to the max. pairwise error and one can probably derive how it should relate to it.

KaiSzuttor:
Okay but where does the magic number come from? We should came up with a physical reasoning if - as you said - the algorithm works in a physical sense.

fweik:
Since we're comparing the total excess charge to a inherent charge scale of the system (namely the smallest charge in the system), its sufficient to check that this ration is much smaller than one. Sorry I misunderstood that at first, I thought the excess total charge was compared to some random charge. I think it's reasonable to pick a small number which is much smaller than one but much larger than the numerical precision to avoid problems from summing. Therefor sth. like 1e-10 should be fine, this will catch most of the cases. Also no result is affected by this. I think the only outcome is a false positive for the check, which is not the end of the world as it can be disabled.

fweik:
@KaiSzuttor I think you have suffered enough from this, I'll take it from here. Also I'm sorry that I started this bikeshedding exercise.

fweik (approved):

src/core/utils/checks/charge_neutrality.hpp
```diff
@@ -0,0 +1,36 @@
...
+#include <boost/accumulators/accumulators.hpp>
+#include <boost/accumulators/statistics/sum_kahan.hpp>
+
+#include <cmath>
+#include <limits>
+#include <stdexcept>
+
+namespace Utils {
+template <typename ParticleRange>
+bool check_charge_neutrality(
```
jonaslandsgesell, line 13: This would better be named is_charge_neutral(...)

Issue #1960, opened 2018-04-27
Restoring checkpoint hangs with multiple threads


hmenke:
Just when you thought it was over...

Here is my (not so) minimal script:
```python
from __future__ import print_function
import numpy as np
import espressomd
import espressomd.checkpointing

checkpoint = espressomd.checkpointing.Checkpointing(checkpoint_id="checkpoint")

start_sample = 0
checkpoint.register("start_sample")

system = espressomd.System(box_l = [10,10,10])
system.periodicity = [1,1,1]
system.cell_system.skin = 0.1
system.time_step = 0.01
checkpoint.register("system")

if not checkpoint.has_checkpoints():
    for i in range(10):
        system.part.add(pos=system.box_l * np.random.random(3))
    checkpoint.register("system.part")

if not checkpoint.has_checkpoints():
    checkpoint.save(checkpoint_index=0)

restarted = False
if checkpoint.has_checkpoints():
    print("Restoring Equilibration Checkpoint")
    checkpoint.load(checkpoint_index=0)
    print("Checkpoint loaded")
    restarted = True

for k in range(start_sample,200):
    print(k)
    system.integrator.run(steps=5000, reuse_forces=restarted)
    restarted = False
    start_sample = k + 1
    checkpoint.save(checkpoint_index=0)
```
Reproducible in the Docker container with `maxset.hpp`.
```
root@009346a1bf7b:/tmp/espresso/build# git show
commit 39846e4328e25aa8316bd1d2648a3c9dae8692a0
Merge: 0aad12e e4d7bb4
Author: Florian Weik <florianweik@gmail.com>
Date:   Thu Apr 26 15:00:31 2018 +0200

    Merge pull request #1958 from KaiSzuttor/fix_integrator_checkpoint
    
    integrator: setstate and getstate methods.

root@009346a1bf7b:/tmp/espresso/build# mpirun --allow-run-as-root -n 2 ./pypresso test.py 
0
1
2
3
4
5
6
^Croot@009346a1bf7b:/tmp/espresso/build# mpirun --allow-run-as-root -n 2 ./pypresso test.py 
Restoring Equilibration Checkpoint
```
---

It runs fine in a single thread but I noticed that after restoring the checkpoint the integration seems to be much slower than before.
![test](https://user-images.githubusercontent.com/1903556/39339135-2ad5eb7e-4a1b-11e8-8c7b-9c2ad9d35415.gif)

RudolfWeeber:
[...]

Issue #1938, opened 2018-04-02
Checkpointing is broken


hmenke:
Restoring a checkpoint results in things being discarded.  Unfortunately I don't know why.

Here is a “minimal” sample script
```python
import numpy as np

import espressomd
import espressomd.checkpointing

checkpoint = espressomd.checkpointing.Checkpointing(checkpoint_id="checkpoint")

restore_point = 0
restarted = False
if checkpoint.has_checkpoints():
    restore_point = 1

# We restart the simulation loops at start_sample
start_sample = 0
checkpoint.register("start_sample")

system = espressomd.System(box_l = [10,10,10])
system.time_step = 0.01
system.cell_system.skin = 0.1

checkpoint.register("system")
checkpoint.register("system.analysis")
checkpoint.register("system.integrator") # This is suddenly necessary

if restore_point == 0:
    for i in range(100):
        system.part.add(pos=system.box_l*np.random.random(3),type=0)

checkpoint.register("system.part")

system.non_bonded_inter[0,0].lennard_jones.set_params(
        epsilon=1.0, sigma=1.0, cutoff=1.0, shift="auto")

checkpoint.register("system.non_bonded_inter")

# Warmup
if restore_point == 0:
    system.minimize_energy.init(f_max=1.0, gamma=0.01, max_steps=1000, max_displacement=0.001)
    system.minimize_energy.minimize()
    checkpoint.save(checkpoint_index=0)

system.thermostat.set_langevin(kT=1.0, gamma=6.0*np.pi)
checkpoint.register("system.thermostat")

# Equilibration
start_sample = 0
if restore_point < 2:
    if restore_point == 1:
        checkpoint.load(checkpoint_index=0)
        restarted = True

    for k in range(start_sample,1000):
        print("Equilibration: {} of {}".format(k,1000))
        print("Total = {e[total]}\n  Kinetic = {e[kinetic]}\n  Non-bonded = {e[non_bonded]}".format(e=system.analysis.energy()))
        if restarted:
            system.integrator.run(steps=1000, reuse_forces=True)
            restarted = False
        else:
            system.integrator.run(steps=1000)

        start_sample = k + 1
        checkpoint.save(checkpoint_index=0)
```
When I run this script and exit using Ctrl+C somewhere during equilibration and then restart the script, all the dynamics get lost.  As you can see, after restoring none of the energies change anymore (which shouldn't happen because there is a thermostat present).
```
Equilibration: 30 of 1000
Total = 156.155161993
[...]

PR #1979, opened 2018-05-09
fix checkpointing of non bonded interactions

KaiSzuttor:
… having a system instance.


KaiSzuttor:
~stil~l non_bonded_inter can~not~ be recovered correctly:

``` python
import numpy as np
import espressomd

checkpoint = espressomd.checkpointing.Checkpointing(checkpoint_id="checkpoint")


if not checkpoint.has_checkpoints():
    print("Starting from scratch")
    checkpoint.register("system")
    system.cell_system.skin = 0.1
    system.time_step = 0.01
    for i in range(10):
        system.part.add(pos=system.box_l * np.random.random(3))
    system.non_bonded_inter[0, 1].lennard_jones.set_params(sigma=1.0, epsilon=1.0, shift='auto', cutoff=2.3)
    print(system.non_bonded_inter[0, 1].lennard_jones)

restarted = False
if checkpoint.has_checkpoints():
    print("Restoring Equilibration Checkpoint")
    checkpoint.load()
    print(system.non_bonded_inter[0, 1].lennard_jones)
    print("Checkpoint loaded")
    restarted = True
```

src/core/interaction_data.cpp
```diff
@@ -162,6 +163,7 @@ void ia_params_set_state(std::string const &state) {
   boost::archive::binary_iarchive ia(ss);
   ia_params.clear();
   ia >> ia_params;
+  ia >> max_seen_particle_type;
```
fweik, line 166: I think you have to set max_seen_particle_type to the maximum of its current value and the value from the archive, as it's current value could be higher than the stored one.

fweik (changes requested):

src/core/interaction_data.cpp
KaiSzuttor: but then I'm also not allowed to clear ```ia_params```, somebody could have created interactions before loading the checkpoint

KaiSzuttor: in my opinion the user cannot expect that espresso is able to merge current state with saved state

src/core/interaction_data.cpp
```diff
@@ -162,6 +163,9 @@ void ia_params_set_state(std::string const &state) {
...
+  mpi_bcast_max_seen_particle_type(ia_params.size());
```
fweik: I don't think this is correct, ia_params contains all type pairs.

fweik (changes requested):

KaiSzuttor:
I just noticed that it's still not possible to correctly load a system instance without having created a system instance before. The checkpoint test now fails for the particle position. I don't know why yet.

KaiSzuttor:
The following problem arose during this PR:
We have a circular dependency that cannot be overcome without changing the core:
 - CellSystem instance is constructed in Systems constructor
 - CellSystem inherently depends on the box geometry
 - The box geometry is set by the System constructor
 - System instance depends on CellSystem instance

KaiSzuttor:
Similar problem arises for the dependency of ParticleList on the global variable ```time_step```. See failing testcase.

fweik:
So there is no solution for this on the python level?

KaiSzuttor:
if we introduce an object instance that sets all globals and is initialized by pickle first, I think that should work

hmenke:
I haven't been following the changes in the repo too closely in the recent past.  What has changed that makes checkpointing such a big issue suddenly?  I remember that it used to work just fine.

KaiSzuttor:
its not a big issue, I'm trying to make it more comfortable by making it possible to checkpoint "system" instead of single components.

hmenke:
I admit, my comment is not well-placed in this thread.  Actually I was more referring to #1938, #1960, and #1969.

KaiSzuttor:
@fweik introducing an object that sets global variables would lead to an interface change, e.g.
``` python
system.globals.box_l = [10.0, 10.0, 10.0]
```
do we want that just because of checkpointing issues?

RudolfWeeber:
On Fri, May 25, 2018 at 03:00:52AM -0700, Kai Szuttor wrote:
> @fweik introducing an object that sets global variables would lead to an interface change, e.g.
> ``` python
> system.globals.box_l = [10.0, 10.0, 10.0]
> ```
> do we want that just because of checkpointing issues?
I'd say no. It's cumbersome and what is or isn't a 'global' in the core shouldn't be of concern to the user.

fweik:
Can't you get that with a property without an interface change?

fweik (approved):

PR #1980, opened 2018-05-14
Refactor class structure in core and scriptinterface (not yet working)

KaiSzuttor:

### DIFF
--- a/src/core/observables/CylindricalLBProfileObservable.hpp
+++ b/src/core/observables/CylindricalLBProfileObservable.hpp
@@ -4,42 +4,12 @@
 #include <cmath>
 
 #include "CylindricalProfileObservable.hpp"
+#include "LBObservable.hpp"
 #include "Vector.hpp"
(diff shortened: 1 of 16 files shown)


fweik:
I think this is so messy because the autoparameters do not play well with inheritance. I also have vague idea on how to this better using free functions, but have not yet implemented it. In general I think it is better idea do implement the autoparameters stuff outside of the class hierarchy e.g. as template specializations of a free function that simply returns the autoparameters which could then be easily used in a derived class by simply calling the respective functions for the (potentially) multiple bases one after another and adding the returned parameters. I'm currently experimenting with such a construction in a different PR which I'm hoping to submit today. After that is finished I'll look into whether this can be reused here too.

fweik:
@KaiSzuttor can this be closed? I think a lot has changed since then and I don't remember what this was about...

KaiSzuttor:
can be closed. it was about multiple inheritance in the scriptinterface...

PR #1984, opened 2018-05-22
Remove scaling

KaiSzuttor:
Description of changes:
 - removed scaling of velocities and forces in the core


fweik:
Why did you change the tests? Shouldn't this change have no influence on the results except for rounding?

KaiSzuttor:
the additional multiplications seemed to have a small influence, therefore the accuracy slightly dropped

KaiSzuttor:
I have no glue why langevin test fails for SHANCHEN config.

RudolfWeeber:
What is the change
@@ -163,9 +162,6 @@ void ia_params_set_state(std::string const &state) {
   boost::archive::binary_iarchive ia(ss);
   ia_params.clear();
   ia >> ia_params;
-  ia >> max_seen_particle_type;
-  mpi_bcast_max_seen_particle_type(max_seen_particle_type);
-  boost::mpi::broadcast(comm_cart, ia_params, 0);
 }
about?

KaiSzuttor:
its related to checkpointing of non-bonded interactions. The interaction matrix has to be communicated.

RudolfWeeber:
Once CI is happy, I think we can merge this.

KaiSzuttor:
Actually the change you mentioned is not part of this PR, maybe your diff is not up to date.

src/core/integrate.cpp
```diff
@@ -481,7 +476,6 @@ if (integ_switch != INTEG_METHOD_STEEPEST_DESCENT) {
   /* Steepest descent operatates on unscaled forces,
      so we have to scale them back now. */
   if(integ_switch == INTEG_METHOD_STEEPEST_DESCENT) {
```
fweik: This `if` can be completely removed.

src/core/integrate.cpp
```diff
@@ -576,12 +543,12 @@ void rescale_forces_propagate_vel() {
 #ifdef NPT
         if (integ_switch == INTEG_METHOD_NPT_ISO &&
             (nptiso.geometry & nptiso.nptgeom_dir[j])) {
-          nptiso.p_vel[j] += Utils::sqr(p.m.v[j]) * p.p.mass;
-            p.m.v[j] += p.f.f[j] + friction_therm0_nptiso(p.m.v[j]) / p.p.mass;
+            nptiso.p_vel[j] += Utils::sqr(p.m.v[j] * time_step) * p.p.mass;
```
fweik, line 540: This looks strange: `[p_vel] = L^2 * M`? Is this compensated by some coefficient?

fweik (changes requested):

src/core/integrate.cpp
KaiSzuttor: ```p_vel``` gets divided by ```time_step * time_step``` in https://github.com/KaiSzuttor/espresso/blob/3342bd04b01a783bd053b74a922e87f039001b0a/src/core/integrate.cpp#L576

fweik: Ok

fweik (approved):

fweik:
@RudolfWeeber are you happy with this?

KaiSzuttor:
@fweik dpd does not reach desired accuracy for odd no of cores, can you guess why that is?

fweik:
The test fails sometimes because it's not deterministic. Maybe you can just fix some of the seeds...

KaiSzuttor:
I will add a viscous coupling test for LB, then this can be merged.

fweik:
We should get @RudolfWeeber to sign off on this, I'm not sure he commented on the right PR.

KaiSzuttor:
@fweik please clearify

fweik:
I'm not sure if Rudolf has approved the correct PR, as he was mentioning a change from a different PR. He uses the github cli where you refer to the PRs by number, so mistakes can happen.

RudolfWeeber:
On Wed, May 30, 2018 at 05:37:37AM -0700, Florian Weik wrote:
I was doing the diff against an outdated python branch, I think,
We can merge this, once the LB viscous couplign test (one which depends on the correct "fric") is in.

fweik:
Alright, just wanted do be sure.

KaiSzuttor:
:runner: seems like viscous force depends on ```agrid```

KaiSzuttor:
Actually it seems it does not have to do with the scaling, the current python branch also fails for the following simple check (for agrid=1 it works):

``` python
import numpy as np

import espressomd
import espressomd.lb


system = espressomd.System(box_l=[10.0]*3)
system.time_step = 0.01
system.cell_system.skin = 0.2
lbf = espressomd.lb.LBFluid(visc=1.0, dens=1.0, agrid=.5, tau=0.01, fric=3.5)
system.actors.add(lbf)
lbf[0, 0, 0].velocity = [1.0, 0.6, 0.2]
system.part.add(pos=[0.25]*3, v=[2.3, 0.0, 0.0], fix=[1, 1, 1])

system.integrator.run(1)
print("expected: {}\ngot {}".format(-3.5* (np.array([2.3, 0.0,  0.0]) - np.array([1.0, 0.6, 0.2])), system.part[0].f))
```

fweik:
The velocity is interpolated, so if the velocity field is not constant, the result will depend on `agrid`.

KaiSzuttor:
not if i place the particle onto the node

fweik:
Ah it's shifted, sorry.

KaiSzuttor:
ill add a (failing) test in a separate PR, and I think we can merge this for now

PR #1988, opened 2018-05-23
core: Use folded coords for constraint energy calc.

### LINKED ISSUE #1986: Constraint energy calculated from unfolded position
Forces are calculated using the folded one. Is this difference intended?

fweik:
Fixes #1986.


RudolfWeeber:
Merged Py branch. Once tests go through, it can be merged.

PR #1989, opened 2018-05-23
vectorial properties in ParticlePosition, remove outdated external_potential

RudolfWeeber:
As a step of the particle data structure refactor, this changes the 
double x[n] 
members in ParticlePosition to Vector<n,double> x.
I did not attempt to get rid of the c-stype arrays everywhere, I just did enough for the code to build again. I.e., in most caes, at some point a Vecotr<n,T>.data() is passed to existing code based on c-stype arrays.
I replaced some type-specific helpers in grid.hpp and utils.hpp such as distance2() by templates which can cope with c-stype arrays and Vector3d.


fweik:
Argh will fix tomorrow.

PR #1990, opened 2018-05-28
Visualizer shows system info

KonradBreitsprecher:
Doubleclick in emtpy region show a list of
- Bonded interactions
- Actors
- Thermostat
- Active Non-bonded interactions and the corresponding type pair
- Constraints

visualization_constraints_0005

with all interface parameters.
Also, snapshots are now named by the python script name.


fweik:
Do autopep, than this is good to merge.

fweik (approved):

PR #1991, opened 2018-05-28
added coulomb stress tensor documentation

### LINKED ISSUE #1834: Documentation of formula for pressure and stress tensor is wrong
http://espressomd.org/html/doc/analysis.html?highlight=pressure%20tensor#stress-tensor
The actual formula which is used, is not only the common virial part but there is some short range part which is calculated via the common virial expression and there is some long ranged part which is calculated via the Maxwell stress tensor formula. This needs to be properly documented.

Also it should be stated that these _instantaneous_ values for the pressure/stress need to be ensemble averaged to obtain the actual pressure/stress.

Also what the different contributions mean needs to be explained. Those are the contributions that arise from force splitting f=f_coulomb+f_bond+f_nonbonded... in the virial stress tensor formula. Later when the user calculates the ensemble average via \langle P \rangle \approx 1/N sum_{i=1}^N however the ensemble average with all interactions present is performed. That means the contributions are not easy to interpret. Those are the contributions to the stress/pressure in a system where all interactions are present and therefore coupled.

Also those formulas are only true if there is no average momentum flux across the periodic boundaries (http://aip.scitation.org/doi/pdf/10.1063/1.437577)

jonaslandsgesell:
Peter and I decided in private communication to allow e.g. pH <0 (together with displaying a warning) in the constant pH method. Fixes #1834 .


fweik:
Please separate the pressure and reaction ensemble stuff into two PRs. Those are not related.

doc/sphinx/advanced_methods.rst
```diff
@@ -1551,3 +1551,262 @@ bonded particles.
...
+chemical potential of each product alone.
+
+.. Since the Reaction Ensemble acceptance transition probability can be
+.. derived from the grand canonical acceptance transition probability we
+.. can use the reaction ensemble to implement grand canonical simulation
+.. moves. This is done via adding reactions that only have reactants (for the
+.. deletion of particles) or only have products (for the creation of
+.. particles). There exists a one to one mapping of the expressions in the
+.. grand canonical transition probabilities and the expressions in the
+.. reaction ensemble transition probabilities.
```
KonradBreitsprecher: Does this belong to the commented section below?

src/python/espressomd/reaction_ensemble.pyx
```diff
@@ -357,6 +357,11 @@ cdef class ConstantpHEnsemble(ReactionAlgorithm):
     def __dealloc__(self):
         del(self.constpHptr)
 
+    def add_reaction(self, *args, **kwargs):
+        if(len(kwargs["product_types"])!=2 or len(kwargs["reactant_types"])!=1):
+            raise ValueError("The constant pH method is not implemented for this kind of reactions.")
```
KonradBreitsprecher: Maybe better state that constant pH requires "product_types" == 2 and "reactant_types" == 1.

src/python/espressomd/reaction_ensemble.pyx
```diff
@@ -368,8 +373,8 @@ cdef class ConstantpHEnsemble(ReactionAlgorithm):
             Sets the pH that the method assumes for the implicit pH bath.
 
             """
-            if(pH<=0):
-                raise ValueError("pH must be strictly positive")
+            if(pH<4 or pH >10):
+                print("Be extremely careful in this extreme pH range when using the constant pH method. The ions which make up the pH (H+ or OH-) are not included in the method. Therefore the method might get screening effects wrong. See paper: Simulation of weak polyelectrolytes: a comparison between the constant pH and the reaction ensemble method by Landsgesell et al.")
```
KonradBreitsprecher: This is pretty vague..

KonradBreitsprecher (commented):
I think the documentation covers the the theory pretty well but lacks python snippets and actual espresso usage. From the doc alone I would have a hard time using it. The module descriptions are good, but the bigger picture of how to use it only can be drawn from the examples right now. So link the sample scripts somewhere (wang_landau_reaction_ensemble.py, widom_insertion.py, reaction_ensemble.py are all I think) or even better put in some general recipe like
- Define the ensemble
- Add an reaction
- Map types
- Call reaction()

src/python/espressomd/reaction_ensemble.pyx
jonaslandsgesell: it shall raise an error if(len(kwargs["product_types"])!=2 or len(kwargs["reactant_types"])!=1). 
you could rewrite as:
if(len(kwargs["product_types"])==2 or len(kwargs["reactant_types"])==1):
   pass
else:
   raise...

I do not see why doing it though.

src/python/espressomd/reaction_ensemble.pyx
jonaslandsgesell: True. When exactly the problems with constant pH start will depend on the exact system that you are looking at and also at the salt concentration that you set there.

doc/sphinx/advanced_methods.rst
jonaslandsgesell: it is not published in a paper. but I think we can uncomment it.

src/python/espressomd/reaction_ensemble.pyx
fweik: I think this belongs into the documentation and not into a warning. If the method does not work as advertised with those parameters, it should not run, otherwise it should do as it is told. We should consider to make it a policy not to do warnings. Either something is impossible or very likely unintended, then it should throw and make the user opt-in explicitly, or it should just work.

KaiSzuttor: Im also against warnings!

jonaslandsgesell: Well the algorithm works for any input. It's just you will very likely produce very wrong physics. I am in favour of this warning. People made this wrong for over 20 years. I see a need to warn them here.

KaiSzuttor:
If it produces wrong physics, it does not work, right?

jonaslandsgesell:
Well, no. I would differentiate. The algorithmic part works. Just the result *may be* wrong. Whether it is badly wrong or just a little wrong depends on your system parameters (mainly the salt concentration, that you decide to be present).

KaiSzuttor:
That may be, but IIRC we decided not to warn users for potentially wrong physics. If we'd decide to do so, we'd also have to warn about many other simulation parameters (e.g. LB parameters).

src/python/espressomd/reaction_ensemble.pyx
KaiSzuttor: In addition you now allow negative pH values (it does not raise an error anymore)

jonaslandsgesell:
I guess the warning policy is related to this sentence from the readme.md:
"ESPResSo is intended to be used by people that have proper knowledge of simulation techniques and know how to use them. We do not take responsibility if you use ESPResSo to create and publish bogus results. You have been warned!"

KaiSzuttor:
It's ESPResSo CE :smile:

PR #1993, opened 2018-05-28
Converts remaining vectorial particle props to Vector3d

RudolfWeeber:


src/core/particle_data.hpp
```diff
@@ -104,19 +104,19 @@ struct ParticleProperties {
...
+  Vector3d rinertia = {1., 1., 1.};
 #else
-  static constexpr const double rinertia[3] = {1., 1., 1.};
+  static const constexpr double rinertia[3] = {1., 1., 1.};
 #endif
 
 #ifdef AFFINITY
   /** parameters for affinity mechanisms */
-  double bond_site[3] = {-1., -1., -1.};
+  Vector3d bond_site[3] = {-1., -1., -1.};
```
fweik: This should not be an array of Vectors.

fweik (changes requested):

fweik (approved):

RudolfWeeber:
AFFINITY was not part of maxset, so a wider regression was not noticed.

I had to re-add the Affinity Interactoin Python class in interactions.pyx from an older version.
I do not know, when it went missing.
@icimrak, would you be willing to contribute a testcase for AFFINITY and for MEMBRANE_COLLISOIN?
It compiles now, but I have no way of telling, whether it still does what it should.

PR #1996, opened 2018-06-01
LB viscous coupling test

KaiSzuttor:
Description of changes:
 - tests the viscous force on a particle for CPU and GPU LB


KaiSzuttor:
It seems like the interpolated velocity is lacking a factor of ```agrid*agrid```. If I divide the interpolated velocity by that factor, the results are correct, but I have no clue where it comes from.

RudolfWeeber:
So, single node and interpolated velocities agree
lbf[0,0,0],lbf.get_interpoalted_velocity([agrid/2]*3)
But the setters and getters 
        self.lbf[0, 0, 0].velocity = v_fluid
        np.testing.assert_allclose(v_fluid,self.lbf.get_interpolated_velocity(self.system.part[0].pos),atol=1E-5)
        np.testing.assert_allclose(v_fluid,self.lbf[0,0,0].velocity,atol=1E-5)
In the code lb_node_set_u and ..._get_u
Getter: lb -> md
p_u[0] = j[0] / rho * lbpar.agrid / lbpar.tau;

Setter: md -> lb
j[0] = rho * u[0] * lbpar.tau * lbpar.agrid;

Personally, I believe the getter more, but that may not mean much./

fweik:
The getter is not correct, it assumes that the `rho` returned from `mpi_recv_fluid` does contain the density, but the getter for `rho`, and the definition of the eq populations in the core, suggests that it actually is the mass of the cell, not the density.

KaiSzuttor:
> So, single node and interpolated velocities agree

You mean disagree?

KaiSzuttor:
So it turned out that the interpolated velocity was correct :dizzy_face:, but the setter for the node velocity was wrong. I added a test for the setting/getting of node velocity.

RudolfWeeber:
On Mon, Jun 04, 2018 at 10:54:32PM -0700, Kai Szuttor wrote:
> > So, single node and interpolated velocities agree
> You mean disagree?
Agree. They also use the same getter.

The inconsistency between setter and getter is an agrid^2.

maintainer/configs/maxset.hpp
```diff
@@ -24,7 +24,7 @@
 #define METADYNAMICS
 
 #define LB
-#define LB_BOUNDARIES
+//#define LB_BOUNDARIES
```
hmenke: Are you sure about that?

hmenke (changes requested):

maintainer/configs/maxset.hpp
KaiSzuttor: No its a usual ```git commit -am``` relict

hmenke:
I just git blamed `lb_lbnode_set_u`.  Turns out this error is 7 years old!  Here is the commit: https://github.com/espressomd/espresso/commit/c97f7319eba76fecd0b4471ea575d3f142eb135c

fweik:
LGTM when Henri's comment is addressed.

PR #1997, opened 2018-06-01
Summed normal force on constraints for pressure calculation

jonaslandsgesell:
Fixes # constraint.total_force() was useless to calculate the pressure. introducing constraint.total_normal_force() which can be used to calculate the pressure acting on a constraint.
Contains tests for wall and cylinder.

### DIFF
--- a/src/core/constraints/ShapeBasedConstraint.cpp
+++ b/src/core/constraints/ShapeBasedConstraint.cpp
@@ -16,12 +16,12 @@ Vector3d ShapeBasedConstraint::total_force() const {
   return total_force;
 }
 
-double ShapeBasedConstraint::total_summed_outer_normal_force() const {
-  double total_summed_scalar_product_outer_normal_force;
-  boost::mpi::all_reduce(comm_cart, m_summed_scalar_product_outer_normal_force, total_summed_scalar_product_outer_normal_force,
+double ShapeBasedConstraint::total_normal_force() const {
+  double total_normal_force;
+  boost::mpi::all_reduce(comm_cart, m_outer_normal_force, total_normal_force,
                          std::plus<double>());
 
-  return total_summed_scalar_product_outer_normal_force;
+  return total_normal_force;
 }
 
 double ShapeBasedConstraint::min_dist() {
@@ -140,7 +140,7 @@ void ShapeBasedConstraint::add_force(Particle *p, Vector3d& folded_pos) {
   }
   force=-force; //Newtons third law. Force acting on constraint due to particle is minus the force acting on the particle
   m_local_force += force;
-  m_summed_scalar_product_outer_normal_force+=outer_normal_vec.dot(force);  
+  m_outer_normal_force+=outer_normal_vec.dot(force);  
 }
 
 void ShapeBasedConstraint::add_energy(Particle *p, Vector3d& folded_pos,
--- a/src/core/constraints/ShapeBasedConstraint.cpp
+++ b/src/core/constraints/ShapeBasedConstraint.cpp
@@ -107,7 +107,7 @@ void ShapeBasedConstraint::add_force(Particle *p, Vector3d& folded_pos) {
                                  force.data(), torque1.data(), torque2.data());
 #ifdef DPD
       if (thermo_switch & THERMO_DPD) {
-        add_dpd_pair_force(p, &part_rep, ia_params, dist_vec, dist, dist2);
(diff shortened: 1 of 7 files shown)


src/core/shapes/Wall.hpp
```diff
@@ -46,7 +46,7 @@ class Wall : public Shape {
   double &d() { return m_d; }
 
 private:
-  /** normal vector on the plane. */
+  /** normal vector on the plane, facing inwards */
```
fweik: What does `inwards` mean?

fweik: What does `inwards` mean?

src/python/espressomd/constraints.py
```diff
@@ -137,7 +137,7 @@ def min_dist(self):
 
     def total_force(self):
         """
-        Get total force acting on this constraint.
+        Get total force acting on this constraint. (This cannot be used to obtain the pressure acting on the constraint).
```
fweik: To me the added comment is just confusing. It can also not be used to make tea.

src/python/espressomd/constraints.py
```diff
@@ -166,7 +166,14 @@ def total_force(self):
 
         """
         return self.call_method("total_force", constraint=self)
-
+        
+    def total_normal_force(self):
+        """
+        Get the total summed normal force acting on this constraint. Via dividing by the area of the constraint one can obtain the Pressure acting on the constraint. A positive value means that the constraint
```
fweik: The first sentence is a perfect description, please remove the rest.

src/script_interface/constraints/Constraints.hpp
```diff
@@ -27,13 +27,15 @@
...
 #include "script_interface/ScriptInterface.hpp"
+#include "core/initialize.hpp"
 
 namespace ScriptInterface {
 namespace Constraints {
 
 class Constraints : public ScriptObjectRegistry<Constraint> {
   virtual void add_in_core(std::shared_ptr<Constraint> obj_ptr) override {
     ::Constraints::constraints.add(obj_ptr->constraint());
+    ::on_constraint_change();
```
fweik, line 38: If this is needed you should probably also call it when a constraint is removed?

jonaslandsgesell:
This is my understanding:

Inwards means inwards in the mathematical sense. There is an
outward-pointing normal and an inward-pointing normal for the boundary of a
system.
http://mathworld.wolfram.com/NormalVector.html

Inward here means that your system is enclosed by the boundary. if you
follow the direction of the inward-pointing normal you will get "deeper"
into your system (going away from the boundary).

Florian Weik <notifications@github.com> schrieb am Mo., 4. Juni 2018, 19:58:

src/core/constraints/ShapeBasedConstraint.cpp
```diff
@@ -16,6 +16,14 @@ Vector3d ShapeBasedConstraint::total_force() const {
   return total_force;
 }
 
+double ShapeBasedConstraint::total_normal_force() const {
+  double total_normal_force;
```
fweik: There is an overload that returns the value, so you can just do 
```c++
return all_reduce(comm_cart, m_outer_normal_force, std::plus<>());
```

fweik (changes requested):

fweik:
This does not make sense for all of the shapes. Where direction is inward for a plain?

jonaslandsgesell:
Well a "normal vector" is ambiguous too. We need to define then which one we want to use. There is always two and you need to use the right one. Which one is the right normal vector to use should be encoded in it s name.

fweik:
I can't see how calling it `inward` adds anything to the understanding. Again, how would you tell which side is inward for a `Wall`? Walls are the most common `Constraints`.

jonaslandsgesell:
If there are *two* parallel walls which are a boundary to the system it is easy to define: the normal which points inwards shows the direction in which to go deeper into the system. If the first wall is at x=0 an the second at x'=2. then the normal for the first wall is (1,0,0) and the second one is (-1,0,0). Why would you otherwise choose (-1,0,0) as the normal for the wall for the second boundary?

src/script_interface/constraints/Constraints.hpp
jonaslandsgesell: True! Otherwise forces are not recalculated.

fweik (approved):

PR #2001, opened 2018-06-04
units: added header file for usage of boost units.

KaiSzuttor:


hmenke (commented):
What does this do?  It doesn't seem to be used anywhere.  Is there more to come in this PR?

KaiSzuttor:
It's not yet clear

fweik:
This is only for discussion purposes. Maybe we should keep in a feature branch instead of a PR? This *could* be the start of a longer project after all.

KaiSzuttor:
I'll push it to a branch

PR #2008, opened 2018-06-06
maintainer: Removed dysfunct endangledist feature.

fweik:
Not working and also not clear what it does.


fweik:
was discussed and approved offline

PR #2010, opened 2018-06-06
Improved interface for constraints + Removed reflection code

fweik:
- Constraints return forces instead of adding
- Better interface
- const
- Removed buggy reflection code


src/core/constraints/ShapeBasedConstraint.hpp
```diff
@@ -43,39 +43,27 @@ class ShapeBasedConstraint : public Constraint {
 
   Shapes::Shape const &shape() const { return *m_shape; }
 
-  ReflectionType const &reflection_type() const;
-
   void reset_force() override { m_local_force = Vector3d{0, 0, 0}; m_outer_normal_force=0.0;}
+
   int &only_positive() { return m_only_positive; }
```
jonaslandsgesell: this is better a bool, also penetrable

src/core/constraints/ShapeBasedConstraint.cpp
```diff
@@ -109,33 +71,29 @@ void ShapeBasedConstraint::add_force(Particle *p, Vector3d& folded_pos) {
...
         runtimeErrorMsg() << "Constraint"
                           << " violated by particle " << p->p.identity
                           << " dist " << dist;
-      }
     }
   }
-  for (int j = 0; j < 3; j++) {
-    p->f.f[j] += force[j];
+  
+    m_local_force -= force;
```
KaiSzuttor, line 85: Why are those two lines added?

KaiSzuttor (approved):

src/core/constraints/ShapeBasedConstraint.cpp
fweik: They are just moved up from the end of the function because of the return.

jonaslandsgesell:
Did you change int ->bool also in the tutorials and examples?

jonaslandsgesell (approved):

PR #2016, opened 2018-06-08
pickle particles before actors

davidsean:
Fixes
Loading a checkpoint with P3M fails when it tunes since particles are not yet present.

Description of changes:
 - pickling order changed: particles before actors 

PR Checklist
------------
 - [ ] Tests?
   - [x] Interface
   - [ ] Core 
 - [ ] Docs?

### DIFF
--- a/samples/load_checkpoint.py
+++ b/samples/load_checkpoint.py
@@ -5,22 +5,21 @@ from espressomd import checkpointing
 checkpoint = checkpointing.Checkpointing(checkpoint_id="mycheckpoint")
 checkpoint.load()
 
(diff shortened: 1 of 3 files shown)


davidsean:
@KaiSzuttor

KaiSzuttor:
LGTM, @davidsean could you please update the sample scripts (no need of creating a system before loading the checkpoint and no need to register anything else besides ```system``` (and accumulators)?

davidsean:
Ok, I kept the p3m reference since it is annoying to get from the actor list (unless you know the index). Perhaps one day it could be accessed with a key, but it's not there now.

KaiSzuttor (approved):

PR #2022, opened 2018-06-09
min_global_cut check for rigid bodies

davidsean:
Fixes error needlessly thrown when using a single proc. If @RudolfWeeber is right, the error is only a concern for multiple processes.

Description of changes:
 - check n_nodes>1 before throwing error
 - added minimal rigid body script in samples

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [x] Core 
 - [ ] Docs?


samples/rigid_body.py
```diff
@@ -0,0 +1,121 @@
...
+system.box_l = [box_l, box_l, box_l]
+system.thermostat.set_langevin(kT=1.0, gamma=1.0)
+
+### Particle types
+type_centre = 0
+type_A = 1
+
+
+for i in xrange(2):
+    for j in xrange(2):
```
hmenke:
```python
from itertools import product
for i,j in product(range(2),range(2)):
```
Also `xrange` does not work in Python 3.

samples/rigid_body.py
```diff
@@ -0,0 +1,121 @@
...
+    if dist>max_dist: max_dist=dist
+max_dist=np.sqrt(max_dist)
+
+# only change this for multiple GPUS
+#system.min_global_cut = max_dist
+print("max dist is {}".format(max_dist))
+print("moment of intertia is", momI)
+
+#place center bead
+system.part.add(id=0, pos=com, mass=branch_len*6, rinertia=momI, rotation=[1,1,1], type=type_centre)
```
hmenke: Use
```python
p_center = system.part.add(pos=com, mass=branch_len*6, rinertia=momI, rotation=[1,1,1], type=type_centre)
```
and
```python
system.part[i].vs_auto_relate_to(p_center.id)
```
and remove all the manual id counting.

hmenke (changes requested):

samples/rigid_body.py
davidsean: come to think of it, this example does not even need a LJ interaction

hmenke (approved):

samples/rigid_body.py
```diff
@@ -0,0 +1,107 @@
...
+# rigid stuff
+#############################################################
+system.virtual_sites= VirtualSitesRelative(have_velocity=True)
+
+#here we calculate the center of mass position (com) and the moment of inertia (momI) of the object
+com=np.zeros(3)
+momI=0
+for p in system.part:
+    com+=p.pos
+com/=(branch_len*6.)
```
KaiSzuttor: can’t you advertise the [core analysis](http://espressomd.org/html/doc/espressomd.html#espressomd.analyze.Analysis.center_of_mass) here?

KaiSzuttor (changes requested):

samples/rigid_body.py
davidsean: good point, I also did it for the moment of inertia and couldn't get matching results. Turns out I copy/pasted wrong physics from the tutorial. Hopefully this now fixes everything. Perhaps whoever setup the raspberry tutorial can have a look.
@jonaslandsgesell

samples/rigid_body.py
```diff
@@ -0,0 +1,106 @@
...
+        p.vs_auto_relate_to(p_center.id)
+
+############################################################
+
+
+h5_file = h5md.H5md(filename=h5md_file_name, write_pos=True, write_vel=False,
+                     write_force=False, write_species=True, write_mass=False,
+                     write_charge=True, write_ordered=True)
+
+h5_file.write()
```
KaiSzuttor: I'd keep the rigid body example as short as possible and remove any possible distructions for new users. Therefore, you probably want to remove file I/O in the sample.

samples/rigid_body.py
```diff
@@ -0,0 +1,106 @@
...
+from espressomd.virtual_sites import VirtualSitesRelative
+from espressomd.io.writer import h5md
+
+import numpy as np
+
+system = espressomd.System(box_l=[1.0, 1.0, 1.0])
+system.set_random_state_PRNG()
+system.seed = system.cell_system.get_state()['n_nodes'] * [1234]
+
+sim_ID="rigid"
```
KaiSzuttor: ```sim_ID``` is not used anywhere else in the script

samples/rigid_body.py
```diff
@@ -0,0 +1,106 @@
...
+#############################################################
+# rigid stuff
+#############################################################
+system.virtual_sites= VirtualSitesRelative(have_velocity=True)
+
+#here we calculate the center of mass position (com) and the moment of inertia (momI) of the object
+com = system.analysis.center_of_mass(part_type=type_A)
+print("center of mass is:", com)
+
+# only change this for multiple GPUS
```
KaiSzuttor: you can just leave the statement in and check for the number of nodes (not GPUs)

samples/rigid_body.py
davidsean: agree with minimalist simulation, but the visual inspection of the floating rigid body is quite fun to see....

RudolfWeeber:
I have not looked at the code. If that's needed, I can do it when I'm back around the 28th.

So these may be adressed already.
* The virual_sites_relative testcase probably needs to be adapted to run without min_global_cut on a single node and, if possible, also with a short-range cutoff 
shorter than the vs-center distance. Otherwise it is not really testing that situation.
* The min_global_cut is probably mentioned in the docs. There, an adaptation would be needed.

KaiSzuttor (approved):

RudolfWeeber:
diff --git a/doc/tutorials/05-raspberry_electrophoresis/scripts/simulate_raspberry_electrophoresis.py b/doc/tutorials/05-raspberry_electrophoresis/scripts/simulate_raspberry_electrophoresis.py
index 9b77b63..93fda5b 100644
--- a/doc/tutorials/05-raspberry_electrophoresis/scripts/simulate_raspberry_electrophoresis.py
+++ b/doc/tutorials/05-raspberry_electrophoresis/scripts/simulate_raspberry_electrophoresis.py
@@ -109,16 +109,12 @@ system.min_global_cut = radius_col+1
 
 #here we calculate the center of mass position (com) and the moment of inertia (momI) of the colloid
 com=np.zeros(3)
-momI=0
 for i in range(n_col_part):
     com+=system.part[i].pos
-    momI+=np.power(np.linalg.norm(com-system.part[i].pos),2)
 com/=n_col_part
 
-for i in range(n_col_part):
-    momI+=np.power(np.linalg.norm(com-system.part[i].pos),2)
 #Note the 2/3 factor is for a spherical shell
-momI*=(2./3.)
+momI=(2./3.)*np.ones(3)*radius_col
Unrelated change. ARe you sure?

diff --git a/samples/rigid_body.py b/samples/rigid_body.py
new file mode 100644
index 0000000..251eecc
--- /dev/null
+++ b/samples/rigid_body.py
@@ -0,0 +1,90 @@
[...]

+
+import espressomd
+from espressomd import thermostat
+from espressomd import integrate
+from espressomd.virtual_sites import VirtualSitesRelative
+
+import numpy as np
+
+system = espressomd.System(box_l=[1.0, 1.0, 1.0])
+system.set_random_state_PRNG()
+system.seed = system.cell_system.get_state()['n_nodes'] * [1234]
+
+system.time_step = 0.01
+skin=10.0
+system.cell_system.skin = skin
+box_l=100
+system.box_l = [box_l, box_l, box_l]
Please set the correct box length when instanceing the System.
Skin=10 seems odd. Maybe use tune_skin()?
[...]
center=system.box_l/2
for direction in (1,0,0),(-1,0,0),(0,1,0),...:
    for n in range(branch_len):
        system.part.add(pos=center+(n+1)*direction, type=type_A, virtual=1)
Avoids code duplication.

[...]
+#place center bead
+p_center = system.part.add(pos=com, mass=branch_len*6, rinertia=momI, rotation=[1,1,1], type=type_centre)
Center of mass contribution to mass?


Also, the doc change mentioned in the previous comment is needed.

fweik:
@RudolfWeeber merge if you are happy.

RudolfWeeber:
Pushed two minor fixes. Once CI is through, this can be merged.

fweik (approved):

PR #2026, opened 2018-06-12
Eigenvectors from gyration_tensor()

davidsean:
Fixes wrong output from the principal component analysis of analysis.gyration_tensor(). The eigenvalues were correct, but the eigenvectors were wrong. Instead of finding what was wrong with the code I just replaced it with GSL's solver.

Description of changes:
 - removed calc_eigenvalues_3x3() and calc_eigenvector_3x3() from utils.cpp (only used by the gyration_tensor())
 - replaced by GSL's linear algebra solver
 - added a unittest that compares results to numpy.linalg's solver

PR Checklist
------------
 - [x] Tests?
   - [ ] Interface
   - [x] Core 
 - [ ] Docs?


davidsean:
hrm.. looks like using GSL may not have been the best solution, as it is not a REQUIRED package but an OPTIONAL one.

KaiSzuttor:
I think we don't want a gsl dependency, I suggest to remove the wrong code in the core and use numpy for the calculation (we don't have to provide that in the interface) @fweik whats your opinion on that?

hmenke:
Or you could use LAPACK which has to be installed anyway, because it is a dependency of NumPy.

RudolfWeeber:
On Wed, Jun 13, 2018 at 12:46:29AM -0700, Kai Szuttor wrote:
> I think we don't want a gsl dependency, I suggest to remove the wrong code in the core and use numpy for the calculation (we don't have to provide that in the interface)
I don't think, calling np.eig in the interface would be too much of a sin.
Please do not remove the (optional) gsl stuff from the build system, as some cluster analysis features require it.

PR #2030, opened 2018-06-15
refactor: mostly variable scope narrowing.

KaiSzuttor:


fweik:
LGTM except for `src/core/lb.cpp:2047`. For the future if you touch such things by hand you could think about making the variable const if possible.

fweik (approved):

PR #2031, opened 2018-06-18
Throw on box change

fweik:
Preliminary throw an error if the box size is changed with active constraints,
this does not work properly with all shapes and can have some non-intuiitive
side effects. Also removed CONSTRAINT feature and moved constraints out
of the short range loop.


jonaslandsgesell:
Some test cases might still contain a check for the constraint feature. E.g. constraint_shape_based.py

fweik:
Ready for review.

src/core/constraints/Constraints.hpp
```diff
@@ -0,0 +1,56 @@
...
+      for (auto const &c : *this) {
+        c->add_energy(p, pos, energy);
+      }
+    }
+  }
+
+  void on_boxl_change() const {
+    if (not this->empty()) {
+      throw std::runtime_error("The box size can not be changed because there "
+                               "are active constraints. " +
```
KaiSzuttor: Can you please change the order of the strings, we get a nicer error message then.

KaiSzuttor (approved):
LGTM besides comment

PR #2035, opened 2018-06-18
gyration_tensor() with numpy

davidsean:
analysis.gyration_tensor() now solved with numpy. The eigenvectors from the core code was wrong in cases when two vectors were the same (or similar).

Description of changes:
 - removed core code
 - added test case for a simple geometry (cube and cube with decorations)
 - added cube geometry test also looks at moment_of_inertia_matrix()
 - changed centre_of_mass() kwarg from part_type to p_type 

PR Checklist
------------
 - [x] Tests?
   - [x] Interface
   - [x] Core 
 - [ ] Docs?


src/python/espressomd/analyze.pyx
```diff
@@ -192,13 +192,13 @@ class Analysis(object):
     # Analyze center of mass
     #
 
-    def center_of_mass(self, part_type=None):
+    def center_of_mass(self, p_type=None):
```
KaiSzuttor, line 195: Why not make it a plain argument, then nobody has to remember the name of the keyword and we don't have to check if it was provided or not.

fweik (approved):

src/python/espressomd/analyze.pyx
davidsean: There definitely needs to be some kind of cleanup regarding this. Sometimes it is "p_type", "part_type" or just "type". Sometimes it has to be a single int, othertimes a list of ints and sometimes "all". It would be great if the all the analysis routines could just take a ParticleSlice for exampe from particle_data.ParticleHandle.select()

fweik:
So @KaiSzuttor are there any concrete changes you want? I'm eager to merge this as it removes the single most awful code from espresso.

KaiSzuttor:
probably not in the scope of this PR

PR #2037, opened 2018-06-21
Reaction ensemble vs constant pH tutorial written by Andrea 

### LINKED ISSUE #1551: Tutorial which nicely demonstrates the differences between the constant pH method and the reaction ensemble
The differences between both methods (https://arxiv.org/abs/1702.04911) can be nicely shown.
In the tutorial especially the number of H+ ions can be observed as a function of different box volumina.
*For the constant pH method one observes that the box volume does not affect the number of H+ ions
*For the reaction ensemble one observes a change in the H+ number.

*The different electrostatic screening between the reaction ensemble (in the K sweeping mode and the real titration mode) and the constant pH method at the same "pH" can be shown.

jonaslandsgesell:
Fixes #1551 
Not in ipython yet. 
Maybe Peter @kosovan maybe you want to have a look at it too. I went over it once and corrected what was easy to spot.


davidsean:
tutorial.pdf and 10-reaction_ensemble.pdf are the same.
I have written comments on a printout, feel free to pick them up.

fweik:
Can we go forward with this? @jonaslandsgesell did you look at Davids suggestions?

jonaslandsgesell:
Andrea is dealing with it. Let me see what s the status on it :)

PR #2039, opened 2018-06-22
Widom insertion for multiple reactions

jonaslandsgesell:
Make use of accumulator,
Make use of the sample variance in the accumulator

Description of changes:
 - 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


src/python/espressomd/reaction_ensemble.pxd
```diff
@@ -40,6 +40,8 @@ cdef extern from "reaction_ensemble.hpp" namespace "ReactionEnsemble":
         double slab_start_z
         double slab_end_z
         int non_interacting_type
+        vector[int] number_of_insertions
```
jonaslandsgesell: Needs to be removed

src/core/reaction_ensemble.cpp
```diff
@@ -188,6 +188,14 @@ void ReactionAlgorithm::check_reaction_ensemble() {
 #endif
 }
 
+std::vector<double> ReactionAlgorithm::get_excess_chemical_potential_change_during_reaction(int reaction_id){
```
fweik: Please don't use heap vectors if you know the length in advance. Also this should probably be a member function on `SingleReaction`, because it does not use any other data.

src/core/reaction_ensemble.cpp
```diff
@@ -473,6 +481,9 @@ bool ReactionAlgorithm::generic_oneway_reaction(int reaction_id) {
       current_reaction, E_pot_old, E_pot_new, old_particle_numbers,
       old_state_index, new_state_index, only_make_configuration_changing_move);
 
+  std::vector<double> exponential = {exp(-1.0 / temperature * (E_pot_new - E_pot_old))};
```
fweik, line 482: This is very wasteful.

src/core/reaction_ensemble.cpp
```diff
@@ -1810,11 +1822,13 @@ double WidomInsertion::measure_excess_chemical_potential(int reaction_id) {
   // 2)restore previously changed reactant particles
   restore_properties(changed_particles_properties, number_of_saved_properties);
 
-  double const exponential = exp(-1.0 / temperature * (E_pot_new - E_pot_old));
-  summed_exponentials[reaction_id] += exponential;
-  number_of_insertions[reaction_id] += 1;
-  double const average_exponential = summed_exponentials[reaction_id] / number_of_insertions[reaction_id];
-  return -temperature * log(average_exponential);
+  std::vector<double> exponential = {exp(-1.0 / temperature * (E_pot_new - E_pot_old))};
```
fweik: This is very wasteful

src/core/reaction_ensemble.cpp
```diff
@@ -1783,7 +1795,7 @@ double ConstantpHEnsemble::calculate_acceptance_probability(
   return bf;
 }
 
-double WidomInsertion::measure_excess_chemical_potential(int reaction_id) {
+std::vector<double> WidomInsertion::measure_excess_chemical_potential(int reaction_id) {
```
fweik: Please don't use heap vectors if you know the size in advance.

src/core/utils/Accumulator.hpp
```diff
@@ -71,14 +71,32 @@ inline std::vector<double> Accumulator::get_mean() const {
...
+    res=std::vector<double>(m_acc_data.size(),std::numeric_limits<double>::max());
+  }
   return res;
 }
+
+/**
+returns the standard error of the mean of uncorrelated data. if data are correlated the correlation time needs to be known...
+*/
+inline std::vector<double> Accumulator::get_std_error() const {
+    std::vector<double> variance=get_variance();
```
fweik: `auto const`

fweik (changes requested):

src/core/reaction_ensemble.cpp
jonaslandsgesell: How would i expose a std::array<double,2> in cython?

fweik:
You could use Vector which has puthon bindings

fweik (approved):

PR #2046, opened 2018-06-28
Adapt lj-demo to recent API changes

mkuron:
This isn't quite working yet. `system.analyze.pressure` returns unreasonable values and `system.analyze.energy['kinetic'] * 2/(3*len(system.part))` does not match the temperature set on the thermostat. According to git-bisect, that was broken by #1418.

### DIFF
--- a/samples/lj-demo.py
+++ b/samples/lj-demo.py
@@ -28,6 +28,7 @@ from threading import Thread
 from traits.api import HasTraits, Button, Any, Range, List, Enum, Float
 from traitsui.api import View, Group, Item, CheckListEditor, RangeEditor, EnumEditor
 import sys
+import time
 
 use_opengl = "opengl" in sys.argv
 use_mayavi = "mayavi" in sys.argv
@@ -541,6 +542,8 @@ def midi_thread():
 
         except Exception as e:
             print(e)
+
+        time.sleep(0.01)
 
 
 last_plotted = 0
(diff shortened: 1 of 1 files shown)


KaiSzuttor:
I don't get why the tests are still parsing despite you saying that energy and pressure calculation is broken.

RudolfWeeber:
On Thu, Jul 05, 2018 at 09:45:39AM -0700, Kai Szuttor wrote:
> I don't get why the tests are still parsing despite you saying that energy and pressure calculation is broken.
If I understand correctly, force capping is active in the demo.
The old capping code ignored thermostat forces, while the new one
probably doesn't.

fweik:
For the demo script this can probably be fixed by just increasing the cap to at least the maximal thermostat noise amplitude.

mkuron:
> For the demo script this can probably be fixed by just increasing the cap to at least the maximal thermostat noise amplitude.

I have tried that and don't see a difference. Either I am underestimating the amplitude by orders of magnitude or something else is broken.

mkuron:
Please merge. I now switch on the thermostat only when parameters have changed. Pressue and temperature measured now looks reasonable. It is not possible anymore to crash the simulation by drastic temperature, particle number or volume changes. Drastic pressure changes can occasionally still cause crashes (negative volumes encountered), but that's not a new problem and may be related to what kind of thermostat is used (I find the "piston mass" choice quite arbitrary, by the way).

fweik (approved):

PR #2054, opened 2018-07-10
LB bulk stress

bindgens1:
Fixes #

Description of changes:
Added a observable for the bulk stress of the LB fluid. 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


testsuite/lb_bulk_stress.py
```diff
@@ -0,0 +1,102 @@
...
+  def test(self):
+
+    lb2 = md.lb.LBFluidGPU(agrid=1.0, dens=1.0, visc=1.0, fric=1.0, tau=0.01)
+    sys.actors.add(lb2)
+    sys.integrator.run(100)
+    
+    bulk_stress_observable = np.copy(lb2.pi_bulk)
+    bulk_stress_manual = 0.0
+    
+    for i in range(box_l):
```
KaiSzuttor: there is a lot code duplication here, use functions and a single testclass here

src/core/lb.cpp
```diff
@@ -1382,6 +1382,65 @@ int lb_lbnode_get_pi_neq(int *ind, double *p_pi) {
...
 
+int lb_bulk_get_pi(double *p_pi) {
+
+  for (int l = 0; l < 6; l++) {
+    p_pi[l] = 0.0;
+  }
+
+  if (lattice_switch & LATTICE_LB_GPU) {
+#ifdef LB_GPU
+    lbpar_gpu.dim_x = (unsigned int)rint(box_l[0]/lbpar_gpu.agrid);
```
KaiSzuttor: Why is this needed? It's set in ```lb_lbfluid_set_agrid``` in lb.cpp?

src/core/lb.cpp
```diff
@@ -1382,6 +1382,65 @@ int lb_lbnode_get_pi_neq(int *ind, double *p_pi) {
...
+    lbpar_gpu.dim_y = (unsigned int)rint(box_l[1]/lbpar_gpu.agrid);
+    lbpar_gpu.dim_z = (unsigned int)rint(box_l[2]/lbpar_gpu.agrid);
+    int number_of_nodes_GPU = lbpar_gpu.dim_x * lbpar_gpu.dim_y * lbpar_gpu.dim_z;
```
KaiSzuttor: ```lbpar_gpu.number_of_nodes```

KaiSzuttor (changes requested):

src/core/lb.cpp
```diff
@@ -1382,6 +1382,65 @@ int lb_lbnode_get_pi_neq(int *ind, double *p_pi) {
   return 0;
 }
 
+int lb_bulk_get_pi(double *p_pi) {
```
fweik: Looking at the other methods this should probably be called `lb_lbfluid`...

src/core/observables/LBBulkStress.hpp
```diff
@@ -0,0 +1,29 @@
...
+#include "Observable.hpp"
+#include "lb.hpp"
+
+#include <vector>
+
+namespace Observables {
+
+class LBBulkStress : public Observable {
+public:
+  virtual int n_values() const override {
```
fweik: Remove `virtual`.

src/core/observables/LBBulkStress.hpp
```diff
@@ -0,0 +1,29 @@
...
+    return 6;
+  }
+  virtual std::vector<double> operator()(PartCfg &) const override {
```
fweik: Remove `virtual`.

src/python/espressomd/lb.pyx
```diff
@@ -342,7 +352,7 @@ IF LB or LB_GPU:
 
             def __set__(self, value):
                 raise NotImplementedError
-
+
```
fweik: Remove whitespace

src/python/espressomd/observables.py
```diff
@@ -172,6 +172,9 @@ class ForceDensityProfile(Observable):
 class LBVelocityProfile(Observable):
     _so_name = "Observables::LBVelocityProfile"
 
+@script_interface_register
+class LBBulkStress(Observable):
```
fweik: Please add a docstring describing what the observable calculates.

fweik (changes requested):

testsuite/lb_bulk_stress.py
```diff
@@ -0,0 +1,102 @@
...
+
+box_l = 9
+sys = md.System(box_l=[box_l, box_l, box_l])
+sys.time_step = 0.01
+sys.cell_system.skin = 0.4
+
+@ut.skipIf(not md.has_features(["LB"]),
+          "Features not available, skipping test!")
+class LBTest(ut.TestCase):
+
```
KaiSzuttor: Please remove those blank lines and add a short docstring which describes what the test is doing.

RudolfWeeber:
What is the state, here?
At first glance, the following seems to be open:
* All the test stuff should probably go into the class and the code for gpu and cpu can probably be shared now.
* It should be run on more than one node for lb cpu 
* Ideally, it would be integrated into lb.py
* All the 6-vectors and 3-vectors should be Vector<6,double> and Vector3d, which would remove the akward loops over the elements.
* It is not clear to me, whether an mpi reduce over the nodes is needed for the cpu version.  The test should include some gradient across the system to be sure that is included

bindgens1:
- First of all: Is this feature necessary/useful for anybody else than me? 

- The idea is to test this later together with the Lees Edwards implementation (at least for the GPU version) since we have a known flow field there and can determine the stress from the fluid properties. 

- Can I really use the same test for GPU and CPU. They are different features and I'm unsure if GPU lb is available on every test that runs the CPU lb. If this is the case I'll combine the tests and use functions. Else I could check for a certain feature withing the test. 

- For the mpi reduce multicore stuff I'm unsure how to proceed.

src/python/espressomd/lb.pyx
```diff
@@ -263,6 +263,16 @@ IF LB_GPU or LB:
             if lb_set_lattice_switch(0):
                 raise Exception("lb_set_lattice_switch error")
 
+        property pi_fluid:
```
fweik: Please find a better name than `pi`.

src/python/espressomd/lb.pxd
```diff
@@ -105,7 +105,8 @@ IF LB_GPU or LB:
         int lb_lbnode_get_boundary(const Vector3i & ind, int * int_return)
         int lb_lbfluid_set_couple_flag(int c_couple_flag)
         int lb_lbfluid_get_couple_flag(int * c_couple_flag)
-        int lb_lbfluid_get_interpolated_velocity_global(Vector3d & p, double * v)
+        int lb_lbfluid_get_interpolated_velocity_global(Vector3d &p, double *v)
+        int lb_lbfluid_get_pi(double *pi)
```
fweik: Please find a better name than `pi`.

src/core/grid_based_algorithms/lb.hpp
```diff
@@ -542,6 +542,7 @@ int lb_lbnode_set_pop(const Vector3i &ind, double *pop);
  *  can be called without the position needing to be on the local processor.
  */
 int lb_lbfluid_get_interpolated_velocity_global(Vector3d &p, double *v);
+int lb_lbfluid_get_pi(double *pi);
```
fweik: Please find a better name than `pi`.

RudolfWeeber:
I merged the dev branch into this.
Testining for gpu and cpu ios now in testsuite/python/lb.py in the test_stress_tensor() function.
The observable works for both lb implementations and with more than one core.
MPI_reduce() is not needed, since the observable runs on the head node and fires off mpi callbacks for each lb cell.
That's not efficient, but on a single cpu (1-8 cores) the drag is not huge. Takes 15ms on one core and 30 ms on eight.
The GPU version is really slow, as a Cuda kernel call is issued per LB cell.
Takes 150ms vs. 300ms for the Python implementation.
I think that can only be improved by implementing a kernel that fetches per-node properties for several nodes at once.
I left the timings in the test case for now.

Currently the observbale returns the 6 independent values for the stress tensor. The advantage is that there will be less work to do for the correlator, the disadvantage is that the user will have to know the conversion rule.
See the test case.

src/core/grid_based_algorithms/lb.cpp
```diff
@@ -1414,6 +1414,64 @@ int lb_lbnode_get_pi_neq(const Vector3i &ind, double *p_pi) {
...
+    }
+
+    for (int l = 0; l < 6; l++) {
+      p_pi[l] /= lbpar_gpu.number_of_nodes;
+    }
+#endif
+  } else {
+#ifdef LB
+    int gridsize[3];
+    gridsize[0] = box_l[0] / lblattice.agrid[0];
```
fweik: This is available from `lblattice.global_grid`

RudolfWeeber:
Timings are for an 18^3 lattice.

fweik (changes requested):

bindgens1:
Thanks you very much!
A factor of 2 for the GPU implementation still saves at least some time.

src/python/espressomd/lb.pyx
bindgens1: Would you agree on "stress"?

fweik:
There is a function for the GPU LB that evaluates the velocity at multiple points at once, this can be used to speed this up.

RudolfWeeber:
I pushed a new gpu implementation using code from the VTK writers.
The amount of memory transfer is as wasteful as before, but it still is 100 times faster.
For 18^3 it's below 1ms, for 50^3 it's 12ms.
This could still be improved by doing the averaging in a custom kernel on the gpu, but I think this one is fast enough for current applications.

testsuite/python/lb.py
```diff
@@ -159,6 +161,45 @@ def test_mass_momentum_thermostat(self):
...
+                for k in range(int(system.box_l[2] / agrid)):
+                    stress += self.lbf[i, j, k].pi
+
+        stress /= system.volume() / agrid**3
+        print(time() - tick)
+
+        obs = LBFluidStress()
+        tick = time()
+        obs_stress = obs.calculate()
+        print(time() - tick)
```
KaiSzuttor: please remove the timing and print

src/core/grid_based_algorithms/lb.cpp
```diff
@@ -1414,6 +1420,63 @@ int lb_lbnode_get_pi_neq(const Vector3i &ind, double *p_pi) {
...
+                    for (int i = 0; i < 6; i++)
+                      p_pi[i] += v.pi[i];
+                  });
+
+    // Normalize
+    for (int i = 0; i < 6; i++)
+      p_pi[i] /= lbpar_gpu.number_of_nodes;
+
+    // Add equilibrium stress;
+    const double p0 = lbgpu_eq_stress();
```
KaiSzuttor: auto

src/core/grid_based_algorithms/lb.cpp
```diff
@@ -1414,6 +1420,63 @@ int lb_lbnode_get_pi_neq(const Vector3i &ind, double *p_pi) {
...
+  } else {
+#ifdef LB
+    int number_of_nodes = lblattice.global_grid[0] * lblattice.global_grid[1] *
+                          lblattice.global_grid[2];
+
+    for (int i = 0; i < lblattice.global_grid[0]; i++) {
+      for (int j = 0; j < lblattice.global_grid[1]; j++) {
+        for (int k = 0; k < lblattice.global_grid[2]; k++) {
+          const Vector3i node{{i, j, k}};
+          double pi_bulk[6] = {0.0, 0.0, 0.0, 0.0, 0.0, 0.0};
```
KaiSzuttor: ``Vector<6, double>``

src/core/grid_based_algorithms/lb.cpp
```diff
@@ -1344,6 +1344,15 @@ int lb_lbfluid_get_interpolated_velocity_global(Vector3d &p, double *v) {
 
   return 0;
 }
+#ifdef LB_GPU
+double lbgpu_eq_stress() {
+  double p0 = 0.;
+  for (int ii = 0; ii < LB_COMPONENTS; ii++) {
+    p0 += lbpar_gpu.rho[ii] * lbpar_gpu.agrid * lbpar_gpu.agrid /
```
KaiSzuttor: is ther any literature for this formula?

KaiSzuttor (changes requested):

RudolfWeeber:
> +double lbgpu_eq_stress() {
> +  double p0 = 0.;
> +  for (int ii = 0; ii < LB_COMPONENTS; ii++) {
> +    p0 += lbpar_gpu.rho[ii] * lbpar_gpu.agrid * lbpar_gpu.agrid /
> 
> is ther any literature for this formula?
Didn't check. I extracted it from the per-node getter.
I think it's rho c_s^2 with c_s=speed of sound.

src/core/grid_based_algorithms/lb.cpp
bindgens1: I think the formula is correct [Ulf Schiller's lecture](http://espressomd.org/wordpress/wp-content/uploads/2017/10/summer_school_2017_lattice_boltzmann.pdf) on slide 38+. However, I ran through the slides and on slide 43 there is a formula for the stress observable as 0.5(f_i + f*_i).
The questions is if we want to supply the user with the average or with two observables and he has to do the post calculation on his own.

src/core/grid_based_algorithms/lb.cpp
bindgens1: I tried something along:
`Vector<6, double> pi_bulk{{0.0, 0.0, 0.0, 0.0, 0.0, 0.0}};`
which didn't quite work. Will try again next week.

RudolfWeeber:
Fixes #2611

Double unit conversion for lb cpu stress removed

Now that unit conversion happens in the Py interface, the equilibrium pressure has to be added in LB units in the core
(rho *c_sound^2)

The unit conversion for the LBFluidStress observable is now in the observable itself.

The LB Shear test now checks the stress tensor for some individual lb nodes.
Combinations of Cartesian shear planes and directions are tried.

The LB Test checks agreement betwen stress summed up over all nodes (in Python) and the LBFluidStress observable.

src/python/espressomd/lb.pxd
```diff
@@ -301,16 +302,3 @@ IF LB_GPU or LB:
         cdef double c_density = lb_lbnode_get_density(node)
         return c_density / agrid / agrid / agrid
 
-    cdef inline Vector6d python_lbnode_get_pi(Vector3i node):
```
KaiSzuttor: I'd prefer to have the unit conversion here and not in the core

src/core/EspressoSystemInterface_cuda.cu
```diff
@@ -84,13 +84,21 @@ __global__ void split_kernel_v(CUDA_particle_data *particles, float *v, int n) {
   if (idx >= n)
     return;
 
+#ifdef LB_GPU
```
fweik: This is redundant, the whole function is `#ifdef LB_GPU`, the `elif` branch is never taken.

fweik (changes requested):

RudolfWeeber:
Sorry, I did not realize that HEAD was detached. Pushed old code.

KaiSzuttor:
closing due to #2633

PR #2059, opened 2018-07-13
python: Fixed parameter name to allow removal of LB boundaries.

### LINKED ISSUE #1935: LBBoundary can not be removed
``` python
import espressomd                                                       
import espressomd.lb                                                    
import espressomd.shapes                                                
import espressomd.lbboundaries                                          
                                                                        
system = espressomd.System(box_l=[10.0]*3, time_step = 0.01)            
system.cell_system.skin = 0.4                                           
                                                                        
lbf = espressomd.lb.LBFluidGPU(                                  
    visc=1.0,                                                           
    dens=1.0,                                                           
    agrid=1.0,                                                          
    tau=0.01,                                                           
    fric=1.0,                                                           
    ext_force = [0.0, 0.0, 0.1])                                        
system.actors.add(lbf)                                                                                      
wall_shape1 = espressomd.shapes.Wall(normal=[1., 0., 0.], dist=1.0)     
wall_shape2 = espressomd.shapes.Wall(normal=[-1., 0., 0.], dist=-9.0)   
boundary1 = espressomd.lbboundaries.LBBoundary(shape=wall_shape1)       
boundary2 = espressomd.lbboundaries.LBBoundary(shape=wall_shape2)       
system.lbboundaries.add(boundary1)                                      
system.lbboundaries.add(boundary2)                                      
for b in system.lbboundaries:                                                                                
    system.lbboundaries.remove(b)                                       
```

fweik:
Fixes #1935.

### DIFF
--- a/testsuite/lb_boundary.py
+++ b/testsuite/lb_boundary.py
@@ -7,7 +7,7 @@ from espressomd.shapes import Wall
 import espressomd.lbboundaries
 from itertools import product
 
-class LBBoundariesBase():
+class LBBoundariesBase(object):
     system = espressomd.System(box_l=[10.0, 10.0, 10.0])
 
     wall_shape1 = Wall(normal=[1., 0., 0.], dist=2.5)
--- a/testsuite/lb_boundary.py
+++ b/testsuite/lb_boundary.py
@@ -66,10 +66,10 @@ class LBBoundariesBase(object):
         rng = range(20)
         lbf = self.lbf
 
-        for i in product(range(5), rng, rng):
+        for i in product(range(0, 5), rng, rng):
             self.assertEqual(lbf[i].boundary, 1)
 
-        for i in product(range(6, 15), rng, rng):
+        for i in product(range(5, 15), rng, rng):
             self.assertEqual(lbf[i].boundary, 0)
 
         for i in product(range(15, 20), rng, rng):
@@ -82,14 +82,17 @@ class LBBoundariesBase(object):
 @ut.skipIf(not espressomd.has_features(["LB_BOUNDARIES"]),
            "Features not available, skipping test!")
 class LBBoundariesCPU(ut.TestCase, LBBoundariesBase):
-    lbf = espressomd.lb.LBFluid(
-        visc=1.0,
-        dens=1.0,
-        agrid=0.5,
-        tau=1.0,
-        fric=1.0)
+    lbf = None
 
     def setUp(self):
+        if not self.lbf:
+            self.lbf = espressomd.lb.LBFluid(
+                visc=1.0,
+                dens=1.0,
+                agrid=0.5,
+                tau=1.0,
+                fric=1.0)
+
         self.system.actors.add(self.lbf)
 
     def tearDown(self):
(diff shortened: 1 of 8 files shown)


RudolfWeeber:
Did you check that removing a boundary actually updates the boundary flags of the lb nodes?

fweik:
It does not I think. Good point.

KaiSzuttor (approved):

fweik:
Please don't merge, we still need to make sure the boundary field is updated is boundaries are added or removed.

KaiSzuttor:
i checked the code and it seems not to be trivial to accomplish this.

fweik:
Why can't we just call `lb_boundary_init` after every change?

KaiSzuttor:
good point...

fweik:
@KaiSzuttor Now of course the tests fail because the shear plan is shifted....

fweik:
@KaiSzuttor ready for review. The refactoring was supposed to go to a different branch

fweik:
Added basic test, apparently this is broken for the GPU version. Unfortunately the code is completely unreadable.

KaiSzuttor:
feature guards missing in the script interface? feature guard in the testcase

testsuite/lb_boundary.py
```diff
@@ -0,0 +1,117 @@
...
+    def setUp(self):
+        self.system.actors.add(self.lbf)
+
+    def tearDown(self):
+        self.system.lbboundaries.clear()
+        self.system.actors.remove(self.lbf)
+
+@ut.skipIf(not espressomd.has_features(["LB_BOUNDARIES_GPU"]),
+           "Features not available, skipping test!")
+class LBBoundariesGPU(ut.TestCase, LBBoundariesBase):
```
KaiSzuttor, line 104: :+1:

testsuite/lb_boundary.py
```diff
@@ -0,0 +1,123 @@
...
+        lbf = self.lbf
+
+        for i in product(range(0, 5), rng, rng):
+            self.assertEqual(lbf[i].boundary, 1)
+
+        for i in product(range(5, 15), rng, rng):
+            self.assertEqual(lbf[i].boundary, 0)
+
+        for i in product(range(15, 20), rng, rng):
+            self.assertEqual(lbf[i].boundary, 2)
```
KaiSzuttor, line 76: How can this be two if its of type boolean in the core?

testsuite/lb_boundary.py
```diff
@@ -0,0 +1,123 @@
...
+
+        lbb.clear()
+        for i in product(rng, rng, rng):
+            self.assertEqual(lbf[i].boundary, 0)
+
+@ut.skipIf(not espressomd.has_features(["LB_BOUNDARIES"]),
+           "Features not available, skipping test!")
+class LBBoundariesCPU(ut.TestCase, LBBoundariesBase):
+    lbf = None
```
KaiSzuttor, line 85: this variable declaration plus the teardown method can go into the base

testsuite/lb_boundary.py
fweik: Its not, it's an int (1 + the boundary id).

testsuite/lb_boundary.py
fweik: I don't think it makes sense to further refactor the test while the core functionality is not working.

KaiSzuttor: this is orthogonal to the core functionality

fweik: Yes, but the test may change.

testsuite/lb_boundary.py
KaiSzuttor: yes but in #2073 it's a bool... maybe we should merge those two PR's

fweik: No, this bugfix should go into 4.0, I don't think we have time to finish and test the other LB stuff until next week. I have still other things to do.

PR #2063, opened 2018-07-20
LB: some minor code improvements.

KaiSzuttor:


KaiSzuttor:
thanks to @mkuron

src/core/lb.hpp
```diff
@@ -82,27 +82,26 @@
 /** Description of the LB Model in terms of the unit vectors of the
  *  velocity sub-lattice and the corresponding coefficients
  *  of the pseudo-equilibrium distribution */
-typedef struct {
-
+template<size_t N_vel=19>
```
mkuron, line 85: I don't see the point of making this a template parameter. Espresso's LB is D3Q19 everywhere. Nobody would ever dare to change that.

mkuron (approved):

PR #2076, opened 2018-07-24
Test speedup

RudolfWeeber:
random pairs and wang landau reaction ensemble

### DIFF
--- a/testsuite/random_pairs.py
+++ b/testsuite/random_pairs.py
@@ -28,7 +28,6 @@ class RandomPairTest(ut.TestCase):
 
         s.part.add(pos=s.box_l * np.random.random((n_part, 3)))
         self.all_pairs=[]
-        parts = self.system.part
 
         dist_func=self.system.distance
         for pair in self.system.part.pairs():
--- a/testsuite/random_pairs.py
+++ b/testsuite/random_pairs.py
@@ -27,20 +27,32 @@ class RandomPairTest(ut.TestCase):
         np.random.seed(2)
 
         s.part.add(pos=s.box_l * np.random.random((n_part, 3)))
+        self.all_pairs=[]
+        parts = self.system.part
+
+        dist_func=self.system.distance
+        for pair in self.system.part.pairs():
+                if dist_func(pair[0], pair[1]) < 1.5:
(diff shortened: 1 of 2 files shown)


testsuite/random_pairs.py
```diff
@@ -27,20 +27,32 @@ def setUp(self):
         np.random.seed(2)
 
         s.part.add(pos=s.box_l * np.random.random((n_part, 3)))
+        self.all_pairs=[]
+        parts = self.system.part
```
KaiSzuttor: this variable is not used, please remove

testsuite/random_pairs.py
```diff
@@ -27,20 +27,32 @@ def setUp(self):
...
-        parts = self.system.part
-
-        pairs = []
-        for i in range(len(parts)):
-            for j in range(i + 1, len(parts)):
-                if self.system.distance(parts[i], parts[j]) < dist:
-                    pairs.append((i, j))
-
-        self.assertTrue(len(pairs))
+        # Go through list of all possible pairs for full periodicy
```
KaiSzuttor, line 46: Could you please turn that into a docstring for the method?

KaiSzuttor (approved):
LGTM besides comment

KaiSzuttor:
the macs seem to have lower precision? is this compiler related?

fweik:
@KaiSzuttor We've observed this before. One reason could be that the standard library implementation differs (clang vs gcc), and there is some implementation freedom there for the rng distributions (not the rng itself, that has to be identical). As there are other more technical differences (e.g. the handling of memory pages is slightly different on macOS than on Linux), it could also be some subtle bug. Unfortunately this is very hard to investigate, especially without a mac.

KaiSzuttor:
I see, but we also have a clang build on linux...

mkuron:
Clang on Linux still uses libstdc++, so things are expected to match. On Mac, it uses libc++. You could actually try libc++ on Linux, but you would need to recompile Boost.

KaiSzuttor:
what about a gcc build on mac?

RudolfWeeber:
So, I'm not sure the accuracy on Mac is (on average) worse. When it is better, we probably don't notice.

PR #2079, opened 2018-07-26
codecov: ignore travis-ci and split coverage report for test and rest.

KaiSzuttor:


mkuron (approved):
I think it's fine. You don't need to add `testsuite/` to the list of paths though because it doesn't contain any cpp files.

PR #2082, opened 2018-07-26
Gitlab-CI: remove git_clone step

mkuron:
@KaiSzuttor


KaiSzuttor:
IIRC we had the clone step because github had some latency issues and we had to clone the h5xx submodule for every test. @mkuron do you think this is not necessary anymore?

mkuron:
I think it's fine. h5xx is tiny. If that issue does come back, we can think of a new solution that doesn't cause Gitlab-CI to produce a warning, instead of putting the git directory into an artifact .

PR #2085, opened 2018-07-26
Missing tests

fweik:
Description of changes:
 Added some tests to CMake that were missing, added a simple script to check for that situation.
Didn't add the `virtual_sites_tracers_const_force.py` test as it does not actually test something.


RudolfWeeber:
Go

fweik:
In the python3 tests the oif module is not found. Any ideas? @KaiSzuttor?

mkuron:
You need relative imports in the OIF module. Fix in https://github.com/fweik/espresso/pull/7

PR #2087, opened 2018-07-26
testsuite: added poiseuille test for LB.

KaiSzuttor:


testsuite/lb_poiseuille.py
```diff
@@ -0,0 +1,80 @@
+import matplotlib.pyplot as plt
```
fweik: Not needed

testsuite/lb_poiseuille.py
```diff
@@ -0,0 +1,80 @@
...
+                for z in range(int(self.system.box_l[2] + 1)):
+                    v_tmp.append(self.lbf[x, y, z].velocity[2])
+            velocities[x, 1] = np.mean(np.array(v_tmp))
+            velocities[x, 0] = x * AGRID
+
+        v_measured = velocities[1:-1, 1]
+        v_expected = poiseuille_flow(velocities[1:-1, 0] + 0.5 * AGRID - 0.5 *
+                                     self.system.box_l[0], self.system.box_l[0] - 2.0 * AGRID, EXT_FORCE, VISC * DENS)
+        rmsd = np.sqrt(np.sum(np.square(v_expected - v_measured)))
+        self.assertLess(rmsd, 0.1*AGRID/TIME_STEP)
```
fweik: The tolerance seems very high to me, is this the smallest you can get?

fweik (changes requested):

testsuite/lb_poiseuille.py
```diff
@@ -0,0 +1,79 @@
+import numpy as np
+import scipy.optimize
```
fweik: Also not needed, also absent in docker.

fweik (approved):

PR #2110, opened 2018-07-30
Use new Docker images and support current Linux distributions

### LINKED ISSUE #2068: Update CI Docker images
Before the release, we need to add Docker images and CI testing for some recent Linux distributions to ensure that Espresso 4.0 runs properly on all our users' machines:
- debian:stretch
- opensuse/leap:15.0
- ubuntu:bionic
- nvidia/cuda:9.0-devel-ubuntu16.04

The ubuntu-clang-cuda image should be upgraded to Clang 6.0, which is available in the xenial-updates repository and has better static analysis. The ubuntu-python3 image should be rebased on ubuntu:bionic which has Python 3.6. The ubuntu-intel image should be duplicated to also test Intel 2018.

We can drop debian:jessie as that has recently moved to long-term support and is pretty similar to Ubuntu 14.04 which we test CUDA 8.0 on. All the other distributions that we currently have are still supported and widely used, so we'll end up with a few more CI jobs.

#883 requested an image with address sanitizer enabled, which requires rebuilding Python. Maybe I can also take care of that one while I'm at it.

Please assign this ticket to me for the QA day.

mkuron:
Fixes #2068


CMakeLists.txt
```diff
@@ -359,7 +359,7 @@ if(WARNINGS_ARE_ERRORS)
     set(CUDA_NVCC_FLAGS "${CUDA_NVCC_FLAGS} -Xcompiler -Werror")
   endif()
 endif(WARNINGS_ARE_ERRORS)
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wno-sign-compare -Wno-unused-function -Wno-unused-private-field -Wno-unused-variable")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wno-sign-compare -Wno-unused-function -Wno-unused-private-field -Wno-unused-variable -Wno-terminate")
```
mkuron: This change will go away once https://github.com/h5md/h5xx/pull/1 is merged.

src/script_interface/CMakeLists.txt
```diff
@@ -41,7 +41,7 @@ include_directories(${CMAKE_SOURCE_DIR}/src)
 include_directories(${CMAKE_SOURCE_DIR}/src/core)
 include_directories(${CMAKE_BINARY_DIR}/src/core)
 if (HDF5_FOUND)
-    include_directories(${CMAKE_SOURCE_DIR}/libs/h5xx)
+    include_directories(${CMAKE_SOURCE_DIR}/libs/h5xx SYSTEM)
```
mkuron, line 44: This will be changed back once https://github.com/h5md/h5xx/pull/1 and https://github.com/h5md/h5xx/pull/2 are merged.

src/core/unit_tests/Ellipsoid_test.cpp
```diff
@@ -35,7 +35,7 @@ bool check_distance_function(const Shapes::Shape &s) {
   for (int i = 0; i < N; i++) {
     for (int j = 0; j < N; j++) {
       double theta = 2. * i / N * M_PI;
-      double v = j / (N - 1);
+      double v = j / (N - 1.);
```
mkuron, line 38: This changes the behavior, but the old one was obviously wrong (`v` would always be zero).

src/python/espressomd/scafacos.pyx
```diff
@@ -84,9 +84,13 @@ IF SCAFACOS == 1:
...
             if self.dipolar: 
-                res["prefactor"] =magnetostatics.coulomb.Dprefactor
+                IF DIPOLES == 1:
+                    res["prefactor"] = magnetostatics.coulomb.Dprefactor
+                pass
             else:
-                res["prefactor"] =electrostatics.coulomb.prefactor
+                IF ELECTROSTATICS == 1:
+                    res["prefactor"] = electrostatics.coulomb.prefactor
+                pass
```
mkuron, line 93: The default myconfig didn't build on newer Cython versions without this change.

src/core/p3m.cpp
```diff
@@ -361,7 +361,7 @@ void p3m_init() {
 
     /* position offset for calc. of first meshpoint */
     p3m.pos_shift =
-        (double)((p3m.params.cao - 1) / 2) - (p3m.params.cao % 2) / 2.0;
+        std::floor((p3m.params.cao - 1) / 2) - (p3m.params.cao % 2) / 2.0;
```
mkuron: Not sure if this change (and the same one in p3m-dipolar.cpp) matches what was intended. I don't understand the surrounding code. It silences the warning while preserving old behavior, but looks odd.

mkuron:
@fweik, can you have a look and then merge? Most changes are trivial and the ones that aren't I've commented on above.

src/core/p3m.cpp
fweik: The floor does not do anything, as this is integer division. In p3m-dipolar you made the `2` a float. If this works it's not clear to me what raised the warning in the first place. (I *think* btw that the integer division is intended and the old code was correct).

mkuron: The warning was about casting an integer division to float. There's nothing wrong per se, but most of the time such an operation is not intended. The floor keeps it consistently as float, but you are right, I forgot to make the 2 a 2.0 here.

fweik (approved):

PR #1605, opened 2017-11-09
Tutorial4 lb ipy

iammmiru:
Fixes #

Description of changes:
ipython notebook for LB tutorial

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?

PR #2118, opened 2018-07-30
Fixed LB Tutorial scripts: updated syntax, fixed imports

KunTares:


RudolfWeeber:
Isn't there an IPy Notebook for that one? If so, changes need to be ported

KaiSzuttor:
there is an open PR from @iammmiru #1605

fweik:
Ah sorry, 8 dod not think of that. Yes this needs to be ported.

jonaslandsgesell:
Ipy notebooks should be running, irrespectively of the changes here. Please close this PR.

PR #2121, opened 2018-07-30
Add ICP gitlab domain to codecov config

KaiSzuttor:
Maybe helps to avoid the PR comment update for each coverage upload


mkuron (changes requested):
No, it doesn't help. This pull request was edited five times.

fweik:
Does not work.

PR #2123, opened 2018-07-31
Mac install guide updates

mkuron:
Again as suggested by Christian Holm


fweik:
```
Warning, treated as error:
/builds/espressomd/espresso/build/doc/sphinx/installation.rst:915:Footnote [6] is not referenced.
```

PR #2124, opened 2018-07-31
FIx and test of polymer creation

### LINKED ISSUE #2104: Polymer command should not default to start pos (0,0,0)
for the 1st polymer it places.

-> Random or explicitly passsed

RudolfWeeber:
first particle position explicit,
more tolerant type checkign for start_pos
pass nullptr for pos2 when not used
basic test case

Fixes #2104


KaiSzuttor:
@mkuron seems like we again have the h5xx issue

fweik (approved):

PR #2113, opened 2018-07-30
checked in C.Holms changes to active matter tutorial.

KaiSzuttor:

PR #2129, opened 2018-08-01
WIP: Tutorial 6 updates

hmenke:
I was trying to switch this tutorial to IPython but it seems that this is not the right format for this tutorial because the scripts are running quite long.  Also some of the exercises are very different for which we require different Espresso instances because there is no way to reset Espresso.

N.B. Use “hide whitespace changes” when viewing the diff.


mkuron:
> this is not the right format for this tutorial because the scripts are running quite long

The reason we wanted to switch to IPython is so that we could automatically test the tutorials. Right now they are constantly broken by API changes. For this, it's sufficient if the notebook only runs for a short time (e.g. have a variable that the user needs to change before it does the long run).

> because there is no way to reset Espresso

Should probably be fixed on the Espresso level. For now, you could use separate notebooks for each instance.

hmenke:
> The reason we wanted to switch to IPython is so that we could automatically test the tutorials.

Automatic testing is also possible in the current form.  At least tutorial 6 has a set of solution scripts which should run without any user interaction.  To test them you'd only need some reference data to compare to.  If this is not even required but you only want to catch API changes, no further adjustments except runtime are necessary.  (Same holds for tutorial 9).

If we wanted to switch these tutorials to the notebook architecture they basically have to be redone completely.  The current scheme of completing exercise scripts can not really be adapted in a notebook.  On top of that you have the multiple instance issue.

Another non-negligible problem is that we make extensive use of LaTeX features for bibliography and cross-references which are just not there in IPython (some can be achieved by notebook extensions but that unnecessarily complicates the setup for the enduser).

mkuron:
> Automatic testing is also possible in the current form.

Not really. The PDFs and the scripts tend to diverge over time.

RudolfWeeber:
What is the status here?
@hmenke, for the release, we need a working version of the tutorial quickly, including #2113. If IPython notebooks take longer, these can be added later.

hmenke:
I have integrated the changes from #2113 into this PR.  I haven't had the time to test all the scripts yet (`enhanced_diffusion.py` works definitely).  What is the targeted release date for 4.0?  IPython notebooks will be available for the Summer School.

RudolfWeeber:
@hmenke, release within approximately one week, I'd say.

fweik:
LGTM

mkuron:
Not ready to merge. The hollow cone is missing inside the rectifying geometry.

jngrad:
It seems the `HollowCone::calculate_dist()` method only returns absolute values. It should return negative values when a node or particle is inside the shape. The method is not self-explanatory; I've tried capturing the sign of various variables in the code to flip the sign of the return value, without success. In older commits the method returns negative values, but not in a consistent manner (the sign flips randomly when the particle or node is too close to the surface, either inside or outside).

RudolfWeeber:
@jngrad, could you please check whether the following change in 
src/core/shapes/HollowCone.cpp 
fixes the issue (at the very end of the file)

-  *dist = std::copysign(distance, m_direction);
+  *dist = distance * std::copysign(1.0, m_direction);

The changed version maintains the sign and passes your hollowcone test

RudolfWeeber:
@hmenke, we'll stay with the Tex/PY for the 4.0 release, since that has been checked now, and I do not want to delay the release further.
Your work on IPython will go to 4.0.1.

hmenke:
Closing in favour of #2202

PR #2133, opened 2018-08-02
Lb tests

KaiSzuttor:
Description of changes:
 - minor code improvements for LB
 - added streaming test for LB


KaiSzuttor:
osx vms do currently not work @mkuron

mkuron:
@KaiSzuttor, OS X VMs fixed this morning: https://github.com/espressomd/espresso/pull/2132#issuecomment-409852827. Your current broken Docker builds are fixed by https://github.com/espressomd/espresso/pull/2134.

maintainer/gh_post_status.sh
```diff
@@ -8,4 +8,4 @@ STATUS="$1"
 curl "https://api.github.com/repos/espressomd/espresso/statuses/$GIT_COMMIT?access_token=$GITHUB_TOKEN" \
      -H "Content-Type: application/json" \
      -X POST \
-     -d "{\"state\": \"$STATUS\", \"context\": \"ICP GitLab CI\", \"target_url\": \"$URL\"}"
+     -d "{\"state\": \"$STATUS\", \"context\": \"ICP GitLab CI\", \"target_url\": \"$URL\", \"description\": \"GitLab CI is performing tests.\"}"
```
mkuron: Depending on whether `$STATUS` is `pending`, `success` or `failure`, the description should be different.

mkuron (changes requested):

src/core/communication.cpp
```diff
@@ -1864,7 +1864,7 @@ void mpi_send_exclusion_slave(int part1, int part2) {
 }
 
 /************** REQ_SET_FLUID **************/
-void mpi_send_fluid(int node, int index, double rho, double *j, double *pi) {
+void mpi_send_fluid(int node, int index, double rho, const std::array<double, 3> j, const std::array<double, 6> pi) {
```
fweik: Is there a specific reason to take copies of j and pi here?

KaiSzuttor: no, fixed

maintainer/gh_post_status.sh
KaiSzuttor: was a test, removed it

fweik:
Can you please check if the CPU test works with #2136?

KaiSzuttor:
works...

KaiSzuttor:
@fweik i think it's ready for review

fweik:
Did you look at the latest changes in #2136 ? I think that one has to go first.

PR #2136, opened 2018-08-03
Lb fluid refactor

fweik:
Description of changes:
 - Replaced lbfluid by multi_grid

This is just an intermediate solution, because I didn't wanted to
change the indexing before the release.


fweik:
Argh this has a 20% performance regression, because multi_array does not support move semantics and the swap will make copies. Will fix.

src/core/virtual_sites/lb_inertialess_tracers.cpp
```diff
@@ -5,22 +5,23 @@
```
KaiSzuttor, line 5: did you change anything in that file besides format?

KaiSzuttor (approved):
LGTM

src/core/virtual_sites/lb_inertialess_tracers.cpp
fweik: No I don't think so. It's formated because there where intermediate changes to it.

fweik:
@KaiSzuttor ready for review
Performance is now unchanged, according to this benchmark:
```python

import time
import numpy as np

import espressomd
import espressomd.lb

cells_per_side=16
n_cells = cells_per_side**3
agrid = 1.
box_l = cells_per_side*agrid

system = espressomd.System(box_l=[box_l]*3)

system.time_step = 0.01
system.cell_system.skin = 0.1
lb_params = {'agrid': agrid, 'fric': 1, 'dens': 1, 'visc': 1, 'tau': 0.01,
             'ext_force_density': [0, 0, -1.0 / (box_l**3)]}
lbf = espressomd.lb.LBFluid(**lb_params)
system.actors.add(lbf)

steps=1000
n_nodes = system.cell_system.get_state()['n_nodes']

while True:
    tick = time.time()
    system.integrator.run(steps=steps)
    tock = time.time()
    dt = tock - tick
    l = steps*n_cells / dt / 1000.

    print("{} kLU/s, {} kLU/s/node".format(l, l/n_nodes))
```

KaiSzuttor:
```/home/espresso/travis/src/core/lb.cpp:2550:44: error: invalid initialization of non-const reference of type 'boost::multi_array<double, 2ul>&' from an rvalue of type 'boost::multi_array<double, 2ul>'```

KaiSzuttor:
I just measured with your sample script...
single core: approx. 1.3 times faster
four cores: approx. 2.6 times faster
compared to current python branch.

LGTM

KaiSzuttor (approved):

fweik:
It's just faster because the particle coupling is now skipped when there are no particles, which steps over the halo comm. So the speedup isn't really relevant in practice.

KaiSzuttor:
just checked with added particle, then they are very similar

PR #2139, opened 2018-08-05
External fields

### LINKED ISSUE #470: Python support for location-dependent external fields

fweik:
Fixes #470. (eventually).

Description of changes:
 - Implementation of some external fields from
   constants and interpolation from table data.

Will push the docs tomorrow.

PR Checklist
------------
 - [x] Tests?
   - [x] Interface
   - [x] Core 
 - [ ] Docs?


src/core/Vector.hpp
```diff
@@ -50,12 +50,18 @@ template <size_t n, typename Scalar> class Vector {
   Vector(Vector const &) = default;
   Vector &operator=(Vector const &) = default;
 
+  void swap(Vector &rhs) {
```
hmenke:
```cpp
void swap(Vector &rhs) noexcept(noexcept(
    std::swap(std::declval<Scalar &>(), std::declval<Scalar &>()))) {
    std::swap(d, rhs.d);
}
```
or, if you don't care about the exception specification, just
```cpp
void swap(Vector &rhs) {
    std::swap(d, rhs.d);
}
```

src/core/constraints/ExternalField.hpp
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * @brief Constraint interface for ExternalField::ForceField.
+ */
+template <typename Coupling, typename Field>
+class ExternalField : public Constraint {
+  FieldCoupling::ForceField<Coupling, Field> impl;
+
+public:
+  template <typename... Args>
+  ExternalField(Args... args)
```
hmenke: `Args&&... args` to avoid copy.

src/core/field_coupling/couplings/Scaled.hpp
```diff
@@ -0,0 +1,42 @@
...
+  double const &default_scale() const { return m_default; }
+  std::unordered_map<int, double> &particle_scales() { return m_scales; }
+  std::unordered_map<int, double> const &particle_scales() const { return m_scales; }
+
+private:
+  template <typename Particle> double scale(Particle const &p) const {
+    auto const &val = m_scales.find(p.identity());
+    if (val != m_scales.end())
+      return val->second;
+    else
```
hmenke, line 29: You don't need `else` after `return`.

src/core/field_coupling/fields/Interpolated.hpp
```diff
@@ -0,0 +1,145 @@
...
+* @brief A vector field interpolated from a regular grid.
+*
+* This is an interpolation wrapper around a boost::multi_array,
+* which can be evaluated on any point in space by spline interpolation.
+*/
+template <typename T, size_t codim> class Interpolated {
+public:
+  using value_type = typename decay_to_scalar<Vector<codim, T>>::type;
+  using gradient_type = detail::gradient_type<T, codim>;
+  using storage_type = boost::multi_array<value_type, 3>;
```
hmenke, line 45: Some time ago I've written a `constexpr`-friendly statically sized multidimensional array.  In other words, an N-dimensional `std::array`.  It looks like here everything is 3D, so `boost::multi_array` might incur an additional overhead for runtime allocation.
My code is also usable in CUDA.  License is CC0:
https://github.com/hmenke/cpp-patterns/blob/master/multi_array/multi_array.hpp
This code will find its way into Espresso sooner or later because I make extensive use of it in my SD implementation.

src/core/unit_tests/AutoParameter_test.cpp
```diff
@@ -58,6 +58,20 @@ BOOST_AUTO_TEST_CASE(user_provided) {
...
 
+BOOST_AUTO_TEST_CASE(user_provided_read_only) {
+  using namespace ScriptInterface;
+  int i{12};
+
+  auto getter = [&i]() { return i; };
+
+  auto p = AutoParameter("i", AutoParameter::ReadOnly{}, getter);
+
+  BOOST_CHECK(p.type == VariantType::INT);
```
hmenke, line 69: `BOOST_CHECK_EQUAL(p.type, VariantType::INT);` gives a more useful error message.

src/core/utils/math/bspline.hpp
```diff
@@ -0,0 +1,273 @@
+#ifndef UTILS_MATH_BSPLINE_HPP
+#define UTILS_MATH_BSPLINE_HPP
+
+#include "sqr.hpp"
+
+#include <cassert>
+#include <stdexcept>
+
+namespace Utils {
+template <int order, typename T = double> inline T bspline(int i, T x) {
```
hmenke, line 10: If you use template specializations instead of switches, the compiler can inline everything and most of the specializations can be marked `constexpr`.

hmenke (changes requested):
This is a great feature.  Once this is merged I will also extend this to external fields from mathematical expressions in #1644.

RudolfWeeber:
+++ b/src/core/unit_tests/field_coupling_couplings_test.cpp:

+BOOST_AUTO_TEST_CASE(mass) {
+  static_assert(Charge::is_linear, "");


+BOOST_AUTO_TEST_CASE(viscous) {
+  static_assert(Scaled::is_linear, "");

src/core/Vector.hpp
fweik: Calling `std::swap` will cause a stack overflow :-)

src/core/field_coupling/fields/Interpolated.hpp
fweik: Actually the dim is static in `multi_array`, so there is not cost. It's terrible for a different reason: it has strange copy semantics and is not movable. So we'll replace this sooner or later.

fweik:
@RudolfWeeber there is no comment?

src/core/Vector.hpp
hmenke: Why should it?  This is calling `std::swap` on the `std::array` data member.

fweik: Ah yeah, it's too hot here...

src/core/utils/math/bspline.hpp
fweik: Codegen for this is fine. I'm currently working on refactoring this anyway, so I'll leave this unchanged here....

fweik:
@RudolfWeeber ?

RudolfWeeber:
Generally, looks good.

I found the following:

src/core/constraints/Constraint.hpp:
* Please document what fits_in_box() does

/src/core/field_coupling/fields/Interpolated.hpp
* What is the "F" templated parameter in tthe interpolation function?

/src/core/field_coupling/fields/gradient_type.hpp
* Please document graidnet_type

/src/core/unit_tests/field_coupling_couplings_test.cpp
+BOOST_AUTO_TEST_CASE(mass) {
+  static_assert(Charge::is_linear, "");
Test title and assert mismatch. Several.

/src/core/utils/interpolation/bspline_3d.hpp
* Please mention the requirements on / use of "kernel" in the docstrings

/src/core/utils/interpolation/detail/ll_and_dist.hpp
Please put a docstring explaining what this does. Is that the lower left of the grid points relevant for interpolation?

/src/core/utils/math/bspline.hpp:
* In bspline(), orders >7 don't seem to be caught. "default" in swithc?
* Same in bspline_d()

fweik:
> Please document what fits_in_box() does

Added docstring.

> What is the "F" templated parameter in tthe interpolation function?

Template parameter has been removed.

> Please document graidnet_type

Added docstring.

> Test title and assert mismatch. Several.

Fixed

> Please mention the requirements on / use of "kernel" in the docstrings

Added docstring

> Please put a docstring explaining what this does. Is that the lower left of the grid points relevant for interpolation?

Added docstrings and made intent clearer.

> In bspline(), orders >7 don't seem to be caught. "default" in swithc?
> Same in bspline_d()

Actually  it falls thru all to the end of the function, where an exception is thrown.

RudolfWeeber:
In commit 4185e, I think, you change the return value of pos_shift.
e.g. -0.5 vs 0.5 for order 1.
Was that intentional?

RudolfWeeber:
/src/core/field_coupling/fields/Interpolated.hpp:
why is the interpolation order hard-coded in the call operator?

fweik:
> In commit 4185e, I think, you change the return value of pos_shift.
> e.g. -0.5 vs 0.5 for order 1.
> Was that intentional?

Yes.

> /src/core/field_coupling/fields/Interpolated.hpp:
> why is the interpolation order hard-coded in the call operator?

Because only linear interpolation is supported. For higher order, the coefficients
have to be determined by convolution or by solving some linear system, which
is not implemented. This is not worth the effort until somebody needs this.

RudolfWeeber:
On Mon, Aug 20, 2018 at 05:57:12AM -0700, Florian Weik wrote:
> > In commit 4185e, I think, you change the return value of pos_shift.
> > e.g. -0.5 vs 0.5 for order 1.
> > Was that intentional?
> 
> Yes.
> 
> > /src/core/field_coupling/fields/Interpolated.hpp:
> > why is the interpolation order hard-coded in the call operator?
> 
> Because only linear interpolation is supported. For higher order, the coefficients
> have to be determined by convolution or by solving some linear system, which
> is not implemented. This is not worth the effort until somebody needs this.
OK. Then this pr can be mered.

PR #2141, opened 2018-08-08
core structuring

### LINKED ISSUE #764: Structure the code
It would be very helpfull for new developers if the code would be more structured, i.e. Have subdirs in src/core and src/python.
E.g. I've done it in the h5mdcore PR.

KaiSzuttor:
Fixes #764


src/core/io/reader/CMakeLists.txt
```diff
@@ -0,0 +1 @@
+add_library(pdpreader SHARED readpdb.cpp)
```
KaiSzuttor: typo

KaiSzuttor:
@fweik

RudolfWeeber:
Generally, I like it. One issue is the magnetic stuff in the electrostatics folder. 
Maybe "long_range_interactions"?

RudolfWeeber:
Some stuff is built before the myconfig.hpp checking (feature validity) is executed. That might lead to confusing errors

RudolfWeeber:
I think, bonded interactions and friends need to depend on EspressoConfig in CMakeLists.txt
Builds with ninja fail, because espresso-features.hpp is not available in time.

fweik:
This is a clear improvement. I haven't checked all the cmake yet, but as @RudolfWeeber pointed out there is still some dependency issue. 
As for the naming: long_range_ia is also not good, as it contains short-range parts in all of the Ewald methods, the Dip-Dip IA is technically not long-range, and not all long-range methods are contained. Also it's not per se magnetic, but solves Poisson's equation with point dipole sources. These dipoles could very well be electrostatic. Naming is hard.

We could go with physical names for the subsystems, like Coulomb, but I think then they should be closed. This would probably come with additional cost in either performance, e.g. running an additional
short-range loop for electrostatics if needed, or compile time and complexity, e.g. make the whole force calculation into on large (expression) template, and collecting everything at instantiation into one efficient version. The last version would be nice, but I don't think we have the man power to either write or maintain that. The performance impact of the first version is not clear to me and should be measured. It has also performance benefits as the independent subsystems open up the way to shared-memory task parallelism, which might yield better node-level performance.

KaiSzuttor:
I could not reproduce the issue with ninja, but I will add the correct dependencies. I think for now even a  name for the subdir that does not fit perfectly is an improvement compared to having no structure at all.

fweik:
Where do we stand on this? If this should go into the release, it should be merged now. If not we can e.g. factor out the utils so simplify the dependency graph...

RudolfWeeber:
I'd still prefer the magnetic stuff not to hide in electrostatics.
Other naming suggestions:
*charges_dipoles
*electrostatics_magnetostatics

The section on code structure in the developer's guide may have to be updated for this pr.

With regards to the release:
Since the current build system works, I think the options are
merging this pr or doing nothing. I'd not invest in partial solutions.
I'd tend to merge now.

fweik:
Also some things are still misplaced: nemd is in the electrostatics folder, but unrelated. Some magnetic methods on the other hand didn't make it....

KaiSzuttor:
i fixed it and im in the process of polishing

KaiSzuttor:
@fweik

RudolfWeeber:
I'd like to merge this after the release, to avoid further delays.

fweik:
@KaiSzuttor we should merge this with priority because it will diverge very quickly.
@RudolfWeeber can you check if you still have problem with ninja?

KaiSzuttor:
Besides (the wrong) red cross from codefactor this PR is green, so merge if you are happy @RudolfWeeber

KaiSzuttor:
offline approved by @RudolfWeeber

PR #2147, opened 2018-08-10
Speedup trivially speedupable tests

KaiSzuttor:


RudolfWeeber:
Please leave integration steps in virtual sites test unchanged. Errors in the handling of vs short range interactions might not occur at once.
That aside, go.

KaiSzuttor:
I think NPT is broken for an odd number of processes. It seems to converge, but a lot slower than for even number of processes.

fweik:
I think it just has bad statistics. Conceptually it should not depend on the number of nodes except for the noise, it just does an reduction of the pressure over the nodes. Also turning the noise off yields the correct pressure and very fast convergence independent of the number of nodes.

PR #2148, opened 2018-08-12
Rotational diffusion tests for anisotropic particle 

### LINKED ISSUE #2029: Anisotropic particle rotational diffusion test for both LD and BD

bogdan-tanygin:
Fixes #2029

Description of changes:
 - The anisotropic particle rotational diffusion tests based on the reference work [Perrin, F. (1936) Journal de Physique et Le Radium, 7(1), 1-11. [https://hal.archives-ouvertes.fr/jpa-00233379/](https://hal.archives-ouvertes.fr/jpa-00233379/)]

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


testsuite/rotational-diffusion-aniso.py
```diff
@@ -0,0 +1,326 @@
+from __future__ import print_function
+import unittest as ut
```
KaiSzuttor: could you please sort the imports, first external - last espressomd related

testsuite/rotational-diffusion-aniso.py
```diff
@@ -0,0 +1,326 @@
...
+    @classmethod
+    def setUpClass(cls):
+        np.random.seed(4)
+
+    def setUp(self):
+        self.es.time = 0.0
+        self.es.part.clear()
+        print("\n")
+
+    def generate_scalar_ranged_rnd(self, min_par, max_par):
```
KaiSzuttor: you don't need those two functions if you use https://docs.scipy.org/doc/numpy/reference/generated/numpy.random.uniform.html
and if this does not serve your needs (but I think it does), you can use the shape as an argument

testsuite/rotational-diffusion-aniso.py
```diff
@@ -0,0 +1,326 @@
...
+
+@ut.skipIf(not espressomd.has_features(["ROTATION",
+                                        "PARTICLE_ANISOTROPY",
+                                        "ROTATIONAL_INERTIA",
+                                        "DIPOLES"]),
+           "Features not available, skipping test!")
+class RotDiffAniso(ut.TestCase):
+    longMessage = True
+    # Handle for espresso system
+    es = espressomd.System(box_l=[1.0,1.0,1.0])
```
KaiSzuttor: could you please use the standard naming convention in the testsuite and samples: ```system = espressomd.System(...)```

testsuite/rotational-diffusion-aniso.py
```diff
@@ -0,0 +1,326 @@
...
+        n : :obj:`int`
+            Number of particles of the each type. There are 2 types.
+
+        """
+        # Global diffusivity tensor in the body frame:
+        D = self.kT / self.gamma_global
+        dt0 = self.J / self.gamma_global
+
+        # Thermalizing...
+        therm_steps = int(100)
```
KaiSzuttor: integer cast not necessary

testsuite/rotational-diffusion-aniso.py
```diff
@@ -0,0 +1,326 @@
+from __future__ import print_function
+import unittest as ut
+import numpy as np
+from numpy.random import random, seed
+import espressomd
+import math
+import tests_common as tc
```
KaiSzuttor: I think the test is easier to read if you don't use custom abbreviations like ```tc```

testsuite/rotational-diffusion-aniso.py
```diff
@@ -0,0 +1,326 @@
...
+    J = 0.0,0.0,0.0
+    
+    @classmethod
+    def setUpClass(cls):
+        np.random.seed(4)
+
+    def setUp(self):
+        self.es.time = 0.0
+        self.es.part.clear()
+        print("\n")
```
KaiSzuttor: Please remove print statements

testsuite/rotational-diffusion-aniso.py
```diff
@@ -0,0 +1,326 @@
...
+
+    def test_case_00(self):
+        n = 400
+        self.rot_diffusion_param_setup(n)
+        self.es.thermostat.set_langevin(kT=self.kT, gamma=self.gamma_global)
+        # Actual integration and validation run
+        self.check_rot_diffusion(n)
+
+if __name__ == '__main__':
+    print("Features: ", espressomd.features())
```
KaiSzuttor: please remove print statements

KaiSzuttor (dismissed):
Thanks a lot for your contribution, I commented on some minor thinks that should be changed. Please also add the new testcase to the CMakeLists.txt in the testsuite directory. Could you please make clear what is additionally tested in your testcase compared to the langevin testcase?

bogdan-tanygin:
@KaiSzuttor thanks for the review. I've introduce the requested changes. Regarding the question:

> Could you please make clear what is additionally tested in your testcase compared to the langevin testcase?

Sorry for the long following fragment, probably, the 1st item is enough :-)

1. The suggesting here test `rotational-diffusion-aniso` validates the case of both an anisotropic momentum of inertia and an anisotropic rotational friction of the particle. While the langevin testcase implies the isotropic .rinertia in all the assignments. And there is a sensible reason for this (nothing wrong is there): the Euler's rotational equations should be linear for the known relations used there.

2. The langevin testcase validates the (angular) velocity based known expressions like Green-Kubo relations and Maxwell distribution. This is correct that they are fundamental and theoretically sufficient. However, probably, it makes sense to validate _additionally_ the "derived" from them explicitly position / quaternion diffusion properties (done in the `rotational-diffusion-aniso`) due to the following two reasons:

2.1. For instance, the Velocity Verlet (VV) method has specific relations between [local errors, global errors] x [position, velocity]. In order to test that the VV integration fits well the Langevin thermostat, it is better to test positions directly, especially if does not require much computational resources (`33/111 Test  #33: rotational-diffusion-aniso ..............   Passed    3.03 sec`).

2.2. The quaternion based VV integration used in ESPResSo [http://dx.doi.org/10.1103/PhysRevE.59.3733](http://dx.doi.org/10.1103/PhysRevE.59.3733) is quite known, conventional and solid. However, the approach to quaternion changes and the angular velocity changes there are different and complex compare to the simple and direct VV. E.g. we have the constraint lambda-force or the iterative VV STEP-4 for the angular velocity. And this is OK, it relates to the classical issues with the rotational motion simulation. It seems that this additional test is important in order to see how well the quaternion dynamics fits to the angular velocity based diffusion properties already tested in the `langevin_thermostat`. At least because that many of fundamental works, which validate the methods being discussed here, use an absence of the energy drift as a symplectic integration method correctness criterion. However, the positional diffusion is usually out of explicit testing.

All this will be also important when any change will be introduced to the integration loop (especially in the quaternion part), when the Brownian thermostat will be used, etc.

bogdan-tanygin:
The checks have been passed. The requested changes have been introduced.

bogdan-tanygin:
The conflicts have been resolved.

RudolfWeeber:
Subject to CI, this can be merged from my side.

fweik (approved):

fweik:
Thank you for your contribution. Sorry that it took so long to get through review.

bogdan-tanygin:
no problem, thanks for a professional review

PR #2152, opened 2018-08-14
Force exact resorting without Verlet lists

hirschsn:
9c9ef3b removed the forced particle resort in `cells_update_ghosts` if Verlet lists are not used. However, this resort is necessary for correctness. This PR reintroduces it.

Description of changes:
 - Force `resort_particles` to at least `Cells::RESORT_LOCAL` if Verlet lists are not used.

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?

### DIFF
--- a/src/core/cells.cpp
+++ b/src/core/cells.cpp
@@ -385,6 +385,10 @@ void check_resort_particles() {
 
 /*************************************************/
 void cells_update_ghosts() {
+  // If no Verlet lists are used we need exact sorting of the particles
+  if (!cell_structure.use_verlet_list)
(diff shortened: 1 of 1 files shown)


fweik:
Can you please explain why this is needed?

hirschsn:
Argh, you're right. 9c9ef3b is *correct* to remove this, as this case is handled by `check_resort_particles`.

I was debugging some code and had a problem with the particles in cells across ESPResSo subsystems (not MD only). I was obviously running the wrong version of a script which caused `check_resort_particles` to not do what it should.

Sorry for the noise, tI will close this.

PR #2156, opened 2018-08-18
Fixed neighbor search

### LINKED ISSUE #2102: Parallel collision detection causes segfaults
this was hidden due to a wrong box size in the test (10 rather than 1) which skipped the cross-node scenario.

Three particle binding at 8 nodes
Glue to surface node with 16 nodes

fweik:
Fixes #2102.

Description of changes:
 - Store complete neighbor information on cells
 - Use this to do collision detection independent of cell system.
 - Added range assertions on `get_linear_index`. (This would have catched
    the bug right away).

I think this is finally correct now. Maybe we can now even release with properly working collision detection.

### DIFF
--- a/src/core/Cell.hpp
+++ b/src/core/Cell.hpp
@@ -4,36 +4,81 @@
 #include <functional>
 #include <vector>
 
-/* Needed for transform_iterator to work with
-   lambdas on older compilers. */
-#define BOOST_RESULT_OF_USE_DECLTYPE
-
-#include <boost/iterator/transform_iterator.hpp>
-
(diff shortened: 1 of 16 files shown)


RudolfWeeber:
This does not resolve (all) issues with the collision detection. 
So the block against parallel collision detection has to stay. The rest of the pr can be merged.

Tested with testsuite/collision_detection.py from my coldet_par_fix branch.

fweik:
@RudolfWeeber I readded the check for the number of nodes.

RudolfWeeber:
Can be merged.

PR #2163, opened 2018-08-20
Lb thermo virtual

### LINKED ISSUE #2157: LB Thermostat: acton_on_virtual should default to True

fweik:
Fixes #2157.

Description of changes:
 - Default `thermo_virtual` to `true`.

### DIFF
--- a/src/core/thermostat.cpp
+++ b/src/core/thermostat.cpp
@@ -39,7 +39,7 @@ int thermo_switch = THERMO_OFF;
 double temperature = 0.0;
 
 /** True if the thermostat should acton on virtual particles. */
-bool thermo_virtual = false;
+bool thermo_virtual = true;
 
 using Thermostat::GammaType;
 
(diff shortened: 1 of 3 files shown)


RudolfWeeber:
thermostat.turn_off() needs to restore the core value to true, otherwise
using lb after lgv leads to different result than using lb directly.

RudolfWeeber:
Can be merged.

fweik:
Resetting the flag in `turn_off()` breaks the langevin test. This is confusing.

fweik:
Maybe this is to brittle after all....

RudolfWeeber:
On Mon, Aug 20, 2018 at 01:52:40PM +0000, Florian Weik wrote:
> Maybe this is to brittle after all....
The Langevin test also fails in the Python branch on 3 cores for me with the same deviation.
Probably, the accuracy target on the checks for the Maxwell-Boltzmann distribution need to be lowerd.

fweik:
Why is this so sensitive to the number of procs?

PR #2165, opened 2018-08-20
Fix rintertia build

fweik:
Fixes missing symbol if ROTATIONAL_INTERTIA is not defined,
and added missing return value if !DIPOLES.


RudolfWeeber:
Can be merged

PR #2168, opened 2018-08-20
python: Check runtime errors after integration independent of integra…

### LINKED ISSUE #2166: py: integrator.run() needs to call handle_errors() independently of return value from core

fweik:
…tion method and return value.

Fixes #2166.


RudolfWeeber:
Can be mergedc

PR #2169, opened 2018-08-20
remove the fake-open from the samples

### LINKED ISSUE #1932: The fake-open() which opens temp files has to be removed from the samples
A user expects the file that's being written into it. 
Fro some samples, it is also necessary (store_properties)
When a tempfile is intended, that needs to be visible

pdebuyl:
Fixes #1932 

Description of changes:
 - remove the file `samples_common.py' and the related imports

I am aware that it is not a big priority but (1) it is rather straightforward and (2) all the output files when running the concerned examples sum up to less than 1 MB.

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


RudolfWeeber:
Thank you. I'll milestone this for 4.0.1, because the samples have already been checked manually for 4.0.

RudolfWeeber:
Can be merged once CI is completed

Issue #2176, opened 2018-08-21
PARTICLE_ANISOTROPY should still accept single numbers for parameters


RudolfWeeber:
system.part[0].gamma =1. So that scripts not written for particle anisotropy still run.

PR #2177, opened 2018-08-21
Update tutorial9

RudolfWeeber:
@fweik


fweik:
LGTM but needs #2176 to be fixed.

PR #2178, opened 2018-08-21
Add new shape tests

### LINKED ISSUE #2098: Add test for some shapes
* spherocylinder
  * slit pore @KonradBreitsprecher 
  * rhmoboid
  * hollow cone

jngrad:
Fixes #2098

Description of changes:
 - added tests for shapes spherocylinder, slit pore, rhomboid, hollow cone
 - fixed an error when the wall shape test is ran first


RudolfWeeber:
Generally, looks good.

Can you please mention the origin of the magic number in the hollow cone test?

There is no need to store the output of system.constraints.add() as in 
+        hollowcone_wall = system.constraints.add(hollowcone_constraint)

Please see the report from the gitlab CI on github. Your scripts fails for three nodes, because the box is too small for the interaction range of one of the lj forces. The domain of each processor needs to be as big as the longest interacito cutoff + 2 skin.

fweik:
There is no apparent benefit of running these tests with more than 2 cores, so you can just limit the number of procs in the build system (`testsuite/CMakeLists.txt`:62).

jngrad:
Correct me if I'm wrong, but the python unittest module does not support multithreading, so a series of tests in a single file will run on 1 core, no matter the number of cores available on the machine. Even when using the `python -m unittest testFile.testClass.testMethod` syntax to run multiple tests in parallel, each test runs in a dedicated Python session and cannot interact with the other tests. So if parallelism is the issue, it comes from the parallel implementation of espresso, not from the test script nor from the shapes.
To me, it looks like a single test failed, causing the next 3 tests in the queue to raise an error. I can reproduce this behavior by inserting a `raise ValueError('')` in any method of the test class; the method will fail and the next ones will either fail or raise an error. The reason is, `espressomd.System` can only be instanciated once, so it needs to be a global variable and its internal state (number of particles, LJ potentials) must be reset at the end of each test. When one test fails, the internal state is not reset and all following tests fail.

fweik:
@jngrad You are right, the whole test is run with mpi, there is no threading involved. The maximal interaction range that can be supported by the cell system depends on how the box is distributed among the nodes, which depends on the number of nodes. So not every parameter set necessarily works with every number of nodes. It's also not sufficient to just scale up the box size with the number of nodes, as
the node grid depends on the divisors of the node number: If the number of nodes is prime, the only possible distribution is 1 x 1 x p (up to permutation). This is the worst case, where the smallest local box length is box_l / p. This is a recurring problem with tests, to avoid this problem the number of nodes used for every test can be limit via the build system, in the file I mentioned.

jngrad:
@fweik Thanks for the details, it makes sense now. I've updated line 62.

@RudolfWeeber Thanks for the feedback. I dropped variables named `*_wall`, did not notice they were redundant. The magic number in the hollow cone was obtained by placing the particle near the cone and measuring the minimal distance to the shape. Because the cone has a thickness and a bevel, it is not trivial to find a position for the particle that will be a round number (let's say 1.0 unit) away from the shape (rhomboids and cylinders are more straightforward in that regard). It is now briefly explained in the comments.

Concerning the error at line 227 in `test_spherocylinder()`, the code is a verbatim copy of the code in `test_cylinder()`. The spherocylinder is long enough for the spherical caps to lie outside the box, such that the spherocylinder behaves like a cylinder. It should not throw an exception.

Concerning the error at line 332 in `test_wall_forces()`, the only change to the code is the addition of `system.time_step = 0.01` to allow the method to run successfully if executed first (unittest can run tests out-of-order). It should not throw an exception.

PR #2190, opened 2018-08-22
Clang format

fweik:


mkuron:
A check should be added to CI to enforce the format henceforth.

RudolfWeeber:
@mkuron, discussion on the web suggest this is non-trivial, because clang-format output is version-dependent.

mkuron:
Something like this should do: https://github.com/ros-planning/moveit_ci/blob/master/check_clang_format.sh . For now we can just stick to whatever version @fweik used. I'll give it a try later.

mkuron:
This did some pretty bad things about comment length: For example, it converted
```
__threadfence();       // make sure data are visible before setting mass
```
to
```
__threadfence();       // make sure data are visible before setting
//mass
```
which gets converted to
```
__threadfence();       // make sure data are visible before setting
// mass
```
when you run clang-format again. The correct thing would have been
```
__threadfence();       // make sure data are visible before setting
                       // mass
```
@fweik, should we revert #2190 and run clang-format again without the length limit? I am sure there were other issues too.

PR #2191, opened 2018-08-23
Autopep8 .py and .pyx files

RudolfWeeber:


RudolfWeeber:
@mkuron, can I merge this?

mkuron:
I haven't looked at every single of of the 5279 changed lines, but since it's only formatting, I see no reason not to. The only thing I found in the diff was that lines that contained a bunch of `#` to act as a visible separator now contain a single `#`. There are several hundred instances of this. Either the line of `#`s should be restored or the empty comments should be removed completely.

mkuron:
I also fixed the `########################...` problem in #2197. So you can close this one.

PR #2192, opened 2018-08-23
Small fix for msd plot in tut 01

KonradBreitsprecher:
Updated the LJ ipython tut, msd data was plotted wrong.


RudolfWeeber:
On Thu, Aug 23, 2018 at 11:28:13AM +0000, Konrad Breitsprecher wrote:
> Updated the LJ ipython tut, msd data was plotted wrong.
Are you sure, about summing vs. averaging the components?
Otherwise, looks good.

KonradBreitsprecher:
Summing the components is fine, as the square is a scalar product of the displacement dr in vectorial notation.

PR #2194, opened 2018-08-23
clang-format and autopep8 in CI

mkuron:
Inspired by #2190 and #2191.

If necessary, we can switch to [git-clang-format](https://llvm.org/svn/llvm-project/cfe/trunk/tools/clang-format/git-clang-format) so that only changed lines are checked for style.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -248,14 +248,19 @@ style:
   stage: additional_checks
   image: gitlab.icp.uni-stuttgart.de:4567/espressomd/docker/clang:6.0
   script:
-    - find . -name '*.h' -o -name '*.hpp' -o -name '*.cpp' -o -name '*.cu' | xargs clang-format-6.0 -i -style=file
-    - find . -name '*.py' -o -name '*.pyx' -o -name '*.pxd' | xargs autopep8 --in-place --aggressive --aggressive
-    - git --no-pager diff
-    - git diff-index --quiet HEAD -- || echo "Failed style check" && exit 1
+    - find . -name '*.hpp' -o -name '*.cpp' -o -name '*.cu' | xargs -P 8 clang-format-6.0 -i -style=file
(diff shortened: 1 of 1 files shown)


mkuron:
@fweik @RudolfWeeber, what arguments did you use when you ran clang-format and autopep8? The defaults recommend quite a few changes that look legit but weren't included in your formatting pull requests.

fweik:
There is a .clang-format cheked in, that one was used. Iirc clang-format
looks for a config file.

mkuron:
I used that. What clang-format version did you run? I tried 3.8, 4.0, 5.0 and 6.0 but it always made a large number of changes.

fweik:
Whatever is default in ubuntu xenial. Sorry can't check right now. It's
unfortunate that this depends so much on the version.

RudolfWeeber:
On Thu, Aug 23, 2018 at 06:42:07AM -0700, Florian Weik wrote:
> Whatever is default in ubuntu xenial. Sorry can't check right now. It's
> unfortunate that this depends so much on the version.
I also just used autopep8 -i from the ICP installation. 

For pep8, we can probably stick with what's checked now (i.e.,
indentation mainly). The scripts in the CI select explicit rules to
check, iirc.

mkuron:
I tried the ICP versions and get different results. There's some whitespace stuff that should have been caught in #2190, for example (trailing spaces, too many empty lines, wrong indentation, ...).

RudolfWeeber:
So, it seems that both, chromium and MongoDB ship (or download) pre-compiled versions of clang-format for different platforms with their source tree to resolve the inconsistencies. 
http://llvm.org/releases/download.html

mkuron:
I guess we can just decide on one version to use in the CI. Most people aren't going to run it locally on their own computer anyway. Either we enforce style in the CI, or we will quickly diverge from the formatting that #2190 introduced.

But that's not the main problem anyway: #2190 doesn't seem like it's reproducible. If I run the same version that @fweik ran there, I get more changes.

mkuron:
Ready to merge once #2197 is merged.

PR #2197, opened 2018-08-24
Reformat with clang-format 6.0

mkuron:
This reverts #2190 and then applies clang-format 6.0 again. Some manual fixes were required after that to make clang-format converge.

Previously clang-format 3.8 was used by @fweik, which didn't do a very good job. 6.0 is much better.

I compared 5.0 and 6.0 and noticed that 6.0 gave the same result except for some places where the 5.0 formatting was obviously wrong. This suggests that clang-format's format is now stable and mainly receives bugfixes from version to version.

Fixes #2191 as I also ran autopep8. I used version 0.9.1 which we will also have in CI with #2194. For comparison, I also tested 1.3.5 which is slightly different: some changes for the better, some for the worse, so autopep8 definitely has a more stable output than clang-format.


mkuron:
@RudolfWeeber

PR #2198, opened 2018-08-24
Copyright headers

RudolfWeeber:
This changes the line containing the copyright of the Espresso project to
"Copyright (C) 2010-2018 The ESPResSo Project"
and adds maintainer/header_template.txt to all .[c|h]pp and .py files.

This PR also contains the scripts that allpy the chainges

Based on https://www.gnu.org/prep/maintain/html_node/Copyright-Notices.html, my understanding is that one can use a common header for the entire project and does not have to track the times when individual files were touched.
Also, since we continously 'publish' on Github, it is my understanding that all years from 2010 to 2018 are valid years.

Description of changes:
 - 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


RudolfWeeber:
@mkuron?

mkuron:
Remove the following as I have no idea where that came from:
```
# For short files, use:
# 
# Copying and distribution of this file, with or without modification,
# are permitted in any medium without royalty provided the copyright
# notice and this notice are preserved.  This file is offered as-is,
# without any warranty.
```

KaiSzuttor:
>You can use a range (‘2008-2010’) instead of listing individual years (‘2008, 2009, 2010’) if and only if: 1) every year in the range, inclusive, really is a “copyrightable” year that would be listed individually; and 2) you make an explicit statement in a README file about this usage.

RudolfWeeber:
On Fri, Aug 24, 2018 at 02:25:44PM +0000, Kai Szuttor wrote:
> >You can use a range (‘2008-2010’) instead >of listing individual years (‘2008, 2009, >2010’) if and only if: 1) every year in the >range, inclusive, really is a “copyrightable” >year that would be listed individually; and >2) you make an explicit statement in a >README file about this usage.
What do I have to specify explicitly in the readme?

KaiSzuttor:
i guess the usage of the range instead of explitely each year

mkuron:
The AUTHORS file also needs to be updated

mkuron:
And there is a bunch of unnecessary stuff at the end of the readme.

RudolfWeeber:
@mkuron, can I merge this?

mkuron (approved):
I'm not going to read through the entire patch, but having checked a representative sample, I approve of this pull request.

PR #2199, opened 2018-08-25
Fix clang-format CI job

mkuron:
Seems like #2194 broke the build: 
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/34042


mkuron:
@RudolfWeeber

PR #2206, opened 2018-08-28
WIP: Ci speedup

RudolfWeeber:
* disable compile-time optimizations for builds that don't run the testsuite (gives a factor of 2 on ICP machines with -j2)

* Don't run the full test suite for intel compiler buidls, but keep the mpirun of testsuite/particle.py as it is.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -77,7 +77,7 @@ debian:9:
   script:
     - export with_cuda=false
     - export myconfig=maxset make_check=false
-    - export cmake_params=-DCMAKE_BUILD_TYPE=Debug
+    - export cmake_params="-DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_FLAGS_DEBUG='-g -DNDEBUG'"
     - bash maintainer/CI/build_cmake.sh
   tags:
     - docker
@@ -88,7 +88,7 @@ opensuse:42.3:
   image: gitlab.icp.uni-stuttgart.de:4567/espressomd/docker/$CI_JOB_NAME
   script:
     - export with_cuda=false myconfig=maxset make_check=false
-    - export cmake_params=-DCMAKE_BUILD_TYPE=Debug
+    - export cmake_params="-DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_FLAGS_DEBUG='-g -DNDEBUG'"
     - bash maintainer/CI/build_cmake.sh
   tags:
     - docker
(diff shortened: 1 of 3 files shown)


mkuron:
Please leave the Intel MPI tests there. They actually helped us find bugs before, e.g. when we depended on some undocumented OpenMPI-specific behavior.

mkuron (changes requested):
Please leave the Intel tests there. They actually helped us find bugs before, e.g. when we depended on some undocumented OpenMPI-specific behavior or some C++ corner case.

src/core/utils/Span.hpp
```diff
@@ -78,8 +78,7 @@ template <class T> class Span {
   constexpr reverse_iterator rend() const { return reverse_iterator(begin()); }
 
   constexpr reference operator[](size_type i) const {
-    assert(i < size());
-    return m_ptr[i];
+    return assert(i < size()), m_ptr[i];
```
hmenke: Because the form of `assert` is implementation-defined, it is a good idea to disable a possibly overloaded comma operator using
```cpp
return static_cast<void>(assert(i < size())), m_ptr[i];
```

hmenke: I had a look at the C99 standard and “7.2.1.1 The assert macro” states

> **Returns**
> The `assert` macro returns no value

so you can forget about the above comment.

RudolfWeeber:
@mkuron

CMakeLists.txt
```diff
@@ -369,7 +369,11 @@ if(WARNINGS_ARE_ERRORS)
     set(CUDA_NVCC_FLAGS "${CUDA_NVCC_FLAGS} -Xcompiler -Werror")
   endif()
 endif(WARNINGS_ARE_ERRORS)
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wno-sign-compare -Wno-unused-function -Wno-unused-private-field -Wno-unused-variable")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wno-sign-compare -Wno-unused-function -Wno-unused-variable")
+if (NOT CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
```
mkuron: Please add `AND NOT CMAKE_CXX_COMPILER_ID STREQUAL "Intel"`.

mkuron (changes requested):

.gitlab-ci.yml
```diff
@@ -77,6 +77,7 @@ debian:9:
   script:
     - export with_cuda=false
     - export myconfig=maxset make_check=false
+    - export cmake_params="-DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_FLAGS_DEBUG='-g -DNDEBUG'"
```
fweik, line 80: Why do you need `-DNDEBUG` if the tests are not run anyway?

fweik:
I like #2208 better. :-)

PR #2208, opened 2018-08-29
Ndebug

fweik:
Description of changes:
 - Enable assertions for the CI testing
 - Switch builds that are not run to '-O0' to speed them up.

Fixes #2207. Closes #2206.

PR #2209, opened 2018-08-29
Test speed

RudolfWeeber:
Those dominating the clang sanitizer test times


fweik (changes requested):
Please revert the gitlab-ci changes, this is handled better in #2208. And I agree with @mkuron that we should keep the intel tests.

RudolfWeeber:
Wrong base branch. Should be fixed now

fweik (approved):

PR #2216, opened 2018-09-01
Clean-up documentation

### LINKED ISSUE #2065: Run Grammar check on all documentaion
E.g. diction (http://manpages.ubuntu.com/manpages/bionic/man1/diction.1.html ) may be used. This task does not require knowing the code too much and would be suited for a student assistant.

jngrad:
Used `pyenchant` for typos in the first two commits. These changes ignore british/american spelling differences and limit themselves to the documentation (`doc/*`, commit 4c4c40f) and docstrings/comments (`src/*`, commit effbd9f). The grammar commit 5ecfa96 partially fixes #2065. Commands `git diff --word-diff -U1` and `git diff --word-diff-regex=. -U1` produce more readable diffs for individual commits in this PR.

Notes on the Sphinx documentation:
- Multiple `doc/sphinx/*.rst` files are missing words. I deduced some of them based on context, using  command `grep -P '[^\ \#]\ +[,\.](?!\.)' doc/sphinx/*.rst` to spot missing words based on punctuation marks, however they cannot always be detected this way (e.g. in `doc/sphinx/constraints.rst` line 576, the sentence "the spring is at and its force constant is ." is missing a word after "at" and another one before the final dot).
- Currently, Sphinx generates a single reSt file for the complete python module `espressomd`, which makes navigation a little clumsy. We could consider splitting the large `espressomd.html` page into multiple pages, one per Python file.
- Python code blocks have two syntaxes:
  
      for x in listing:
          print x
  
  and
  
      >>> for x in listing:
      ...     print x
      [1.0, 5.0, -3.0]
      [4.0, 2.0, -1.0]
  
  The second syntax shows code output, which might be important to the reader depending on the situation, but such code blocks cannot be re-used easily. There is however a [piece of JavaScript code](https://hg.python.org/cpython/rev/18bbfed9aafa) from the CPython project that creates buttons on such code blocks. When clicked, the leading '>>>' symbols and any code output are erased, allowing direct copy-pasting ([example](https://docs.python.org/3.7/library/sys.html#sys.float_info)).


src/core/fft-common.hpp
```diff
@@ -291,7 +291,8 @@ void fft_unpack_block(double *in, double *out, int start[3], int size[3],
                       int dim[3], int element);
 
 /** Debug function to print global fft mesh.
-    Print a globally distributed mesh contained in data. Element size is element.
+    Print a globally distributed mesh contained in data. Element size is
+ element.
```
mkuron: The automatic indentation was wrong here. Needs three extra spaces and stars at the beginning of the lines.

src/core/reaction_ensemble.cpp
```diff
@@ -1308,8 +1308,9 @@ int WangLandauReactionEnsemble::do_reaction(int reaction_steps) {
     double minimum_wang_landau_potential = find_minimum_non_negative_value(
         wang_landau_potential.data(), wang_landau_potential.size());
     for (int i = 0; i < wang_landau_potential.size(); i++) {
-      if (wang_landau_potential[i] >= 0) // check for whether we are in the valid
-                                         // range of the collective variable
+      if (wang_landau_potential[i] >=
+          0) // check for whether we are in the valid
+             // range of the collective variable
```
mkuron: Unusal indentation here too. Put the `0)` on the `if`'s line and make the two-line comment a three-line comment.

src/core/fft-common.hpp
```diff
@@ -291,7 +291,8 @@ void fft_unpack_block(double *in, double *out, int start[3], int size[3],
                       int dim[3], int element);
 
 /** Debug function to print global fft mesh.
-    Print a globaly distributed mesh contained in data. Element size is element.
+    Print a globally distributed mesh contained in data. Element size is
+ element.
```
mkuron: Extra spaces are also needed here and in some other places.

src/core/fft-common.hpp
jngrad: Do you mean the "Print a globally..." paragraph should be indented by 3 extra whitespaces, i.e. start at column 8?

src/core/fft-common.hpp
jngrad: Is there a .cpp or .hpp file I could use as reference to correctly format Doxygen comments in the future?

src/core/fft-common.hpp
mkuron: `element` should be aligned with `Print`. Also there should be `*`s in the same column as the first `*` in line 293.

src/core/fft-common.hpp
mkuron: Not really. We were pretty negligient about Doxygen comments. The official documentation is pretty good though: https://www.stack.nl/~dimitri/doxygen/manual/docblocks.html.

doc/sphinx/dg.rst
```diff
@@ -68,11 +73,11 @@ The build system of |es| is based on CMake.
 
 The central source files of the build system are the following:
 
--  ``CMakeList.txt``
+-  :file:`CMakeList.txt`
```
KaiSzuttor: ```CMakeLists.txt```

doc/sphinx/dg.rst
```diff
@@ -68,11 +73,11 @@ The build system of |es| is based on CMake.
...
 
--  Contents of the ``cmake`` directory
+-  Contents of the :file:`cmake` directory
 
--  The CMakeList.txt files in the ``src/``, ``doc/``, and ``testsuite/`` directories and their sub-directories
+-  The CMakeList.txt files in the :file:`src/`, :file:`doc/`, and :file:`testsuite/` directories and their sub-directories
 
 The most common reasons for editing these files are:
 -  Adding new source files
```
KaiSzuttor: Could you please add a blank line before list

doc/sphinx/electrostatics.rst
```diff
@@ -56,12 +56,12 @@ Coulomb P3M
...
 Required parameters:
-    * prefactor
-    * accuracy
+    * ``prefactor``
+    * ``accuracy``
 
 For this feature to work, you need to have the ``fftw3`` library
-installed on your system. In , you can check if it is compiled in by
-checking for the feature ``FFTW`` with ``espressomd.features()``
+installed on your system. In python you can check if it is compiled in by
```
KaiSzuttor: I think here should a ```|es|``` inserted.

doc/sphinx/installation.rst
```diff
@@ -490,16 +498,16 @@ Interaction features
...
 section :ref:`Isotropic non-bonded interactions`):
 
--  ``TABULATED`` Enable support for user–defined interactions.
+-  ``TABULATED`` Enable support for user-defined interactions.
 
--  ``LENNARD_JONES`` Enable the Lennard–Jones potential.
+-  ``LENNARD_JONES`` Enable the Lennard--Jones potential.
 
--  ``LENNARD_JONES_GENERIC`` Enable the generic Lennard–Jones potential with configurable
+-  ``LENNARD_JONES_GENERIC`` Enable the generic Lennard--Jones potential with configurable
```
KaiSzuttor: Why use double dashes?

doc/sphinx/inter_bonded.rst
```diff
@@ -135,26 +135,26 @@ harmonic bond.
 
 .. _Bonded coulomb:
```
KaiSzuttor: Please also change the label if you change the heading

src/core/p3m-dipolar.cpp
```diff
@@ -1227,7 +1227,7 @@ double calc_surface_term(int force_flag, int energy_flag) {
     }
 
     // for (i = 0; i < n_part; i++){
-    //    fprintf(stderr,"part %d, correccions torque  x:%le, y:%le,
```
KaiSzuttor: i opt for removing commented out printf's.

KaiSzuttor (changes requested):
i skimmed over the changes, lookes good besides the comments.

fweik (approved):

PR #2221, opened 2018-09-03
Speedup Gpu tests

RudolfWeeber:
Those that stood out in CI


KaiSzuttor:
ive refactored the ek test in the cleanup pr...

RudolfWeeber:
On Mon, Sep 03, 2018 at 04:21:51PM +0000, Kai Szuttor wrote:
> ive refactored the ek test in the cleanup pr...
The commit only reduces integration_length, so it should be easily mergable.

fweik (approved):

PR #2224, opened 2018-09-04
Disable short range if max_cut <=0

RudolfWeeber:


fweik (approved):

RudolfWeeber:
@fweik

fweik:
LGTM, except for the Bond stuff. I don't feel comfortable judging if the changes there are ok. Can you quickly explain what you changed?

RudolfWeeber:
On Tue, Oct 16, 2018 at 04:48:57PM +0000, Florian Weik wrote:
> LGTM, except for the Bond stuff. I don't feel comfortable judging if the changes there. Can you quickly explain what you changed?
The static analyser complained, because the old code could call a
three particle bond without fetching the 3rd particle.
In practice that didn't happen, as long as the number of partners in
the bonds (bonded_ia_params.num) as set correctly.
The new code, has independent switch() blocks for the different
numbers of bond partners, so that 3 paritcle bonds are only
considreed, whenthe 'num' parameter matches. I.e., the saem criteria
are applied for fetching th 3rd particle and for considering
three-particle bonds. Same for 4-particle bonds.

I also narrowd the scope of the force...-result variables to be sure
they start out as zero for all bonds. This mattered for the oif
global forces.

fweik:
Ok thank you. I think I get it now. I got confused by the moved-around stuff. It's not great obviously, but I think it's slightly better than before.

fweik (approved):

RudolfWeeber:
On Tue, Oct 16, 2018 at 11:23:35AM -0700, Florian Weik wrote:
> Ok thank you. I think I get it now. I got confused by the moved-around stuff. It's not great obviously, but I think it's slightly better than before.
Making bonds great again is a task for a different day.

PR #2225, opened 2018-09-04
Particle type specific verlet criterion for coldet mode glue to surface

RudolfWeeber:
The method acts only on pairs of particles with specific types, so the number of pairs included in the Verlet list can be much smaller.
This is the case particularly, when the collision detection cutoff is large compared to the LJ sigma of other particles in the system.


fweik:
@RudolfWeeber you have to account for the fact that the parameter struct does not always exist.

src/core/collision.hpp
```diff
@@ -162,4 +162,19 @@ inline double collision_detection_cutoff() {
   return 0.;
 }
 
+/** @brief Whether or not the verlet criterion mathes particle types */
+inline bool collision_detection_match_part_type() {
+  return collision_params.mode == COLLISION_MODE_GLUE_TO_SURF;
+}
+
+/** @brief first particle type for verlet criterion */
+inline int collision_detection_part_type1() {
```
KaiSzuttor, line 171: why those one line functions and why do the have a name that doesn't match the variable they are returning?

KaiSzuttor:
@RudolfWeeber can this be closed or are you still working on that?

fweik:
If this is still desired it should be integrated in #3504, otherwise closing due to inactivity.

PR #2227, opened 2018-09-05
Fix build from 4.0.0 tarball

### LINKED ISSUE #1032: New release
Espresso-3.3.1 fail on Fedora with:
```
make[4]: Entering directory '/home/junghans/rpmbuild/BUILD/espresso-3.3.1/no_mpi/src/core'
/bin/sh ../../libtool  --tag=CXX   --mode=compile g++ -DHAVE_CONFIG_H -I. -I../../../src/core -I../../src  -I../../../src/core/mpifake   -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector-strong --param=ssp-buffer-size=4 -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -m64 -mtune=generic -c -o config-features.lo ../../../src/core/config-features.cpp
libtool: compile:  g++ -DHAVE_CONFIG_H -I. -I../../../src/core -I../../src -I../../../src/core/mpifake -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector-strong --param=ssp-buffer-size=4 -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -m64 -mtune=generic -c ../../../src/core/config-features.cpp  -fPIC -DPIC -o .libs/config-features.o
/bin/sh ../../libtool  --tag=CXX   --mode=compile g++ -DHAVE_CONFIG_H -I. -I../../../src/core -I../../src  -I../../../src/core/mpifake   -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector-strong --param=ssp-buffer-size=4 -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -m64 -mtune=generic -c -o cells.lo ../../../src/core/cells.cpp
libtool: compile:  g++ -DHAVE_CONFIG_H -I. -I../../../src/core -I../../src -I../../../src/core/mpifake -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector-strong --param=ssp-buffer-size=4 -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -m64 -mtune=generic -c ../../../src/core/cells.cpp  -fPIC -DPIC -o .libs/cells.o
In file included from ../../../src/core/cells.cpp:30:0:
../../../src/core/utils.hpp:150:16: error: 'std::pmalloc' has not been declared
 #define malloc pmalloc
                ^
../../../src/core/utils.hpp:147:17: error: 'std::prealloc' has not been declared
 #define realloc prealloc
                 ^
Makefile:889: recipe for target 'cells.lo' failed
make[4]: *** [cells.lo] Error 1
```
This is fixed in the master version.  Could you guys tag a new release please?

mkuron:
- Version number should be 4.0.0, not 4.0-dev
- CMake version detection throws warnings if run outside of git
- h5xx needs to be downloaded. If that fails, HDF5 needs to be disabled.

Fixes #1032

Please re-tag 4.0.0 after merging this, @KaiSzuttor @fweik.

Also, please merge to the python branch afterwards and increment the version to _4.0.1-dev_.


KaiSzuttor:
could you please also take care of the following cases:
 - we are in a git repo but failed to find a git command
 - we are in a git repo found a git command but failed to clone the h5xx submodule

KaiSzuttor:
@fweik should we update the release?

mkuron:
The `version` module internally checks for the git command and does nothing if it's not found, so that doesn't need any change.

For h5xx, I now check whether the header file is there after fetching the submodule. If it isn't, I download the tarball.

KaiSzuttor:
So, now we rely on that whenever someone changes the version of h5xx he will remember to change that line in the CMakeLists.txt?

mkuron:
> should we update the release?

We should either update the release or publish a 4.0.0.1. @junghans wants to build binaries for various Linux distributions, which he can only do for a release.

> now we rely on that whenever someone changes the version of h5xx he will remember to change that line in the CMakeLists.txt?

Yes, and there is no way around that. The .gitmodules file does not contain that information. We might be able to get it from the GitHub API, but since the tarball doesn't contain information about from what tag is was created, we'd then have to hard-code the tag name instead of the h5xx version.

KaiSzuttor:
>Yes, and there is no way around that. The .gitmodules file does not contain that information. We might be able to get it from the GitHub API, but since the tarball doesn't contain information about from what tag is was created, we'd then have to hard-code the tag name instead of the h5xx version.

There is a way around that: we only publish to espressomd/releases manually and mark the automatically generated tar on github as not usable. I'm against hardcoding the version of h5xx, only because github can't handle submodules. I think the proper way would be to handle the dependency on h5xx differently (e.g. conan or similar).

KaiSzuttor (approved):

junghans:
I am ok with re-tagging as the tarball from https://github.com/espressomd/release was unusable (due to missing path prefix) and hence I haven't started packaging it up yet!

And yeah, GitHub and submodules is its.own issue! We usually work around that by adding a custom tarball, which includes the source of the submodules, to the release tag!

pdebuyl:
+1 on adding the source of h5xx. This makes for two types of distributions: git for the current development version (developers, advanced users) and the tarball for stable releases.

mkuron:
> I think the proper way would be to handle the dependency on h5xx differently (e.g. conan or similar).

h5xx is the only problem right now, but it won't remain that way. #1644 introduces another submodule and one day we'll have https://github.com/lssfau/walberla as a submodule. I could imagine https://github.com/scafacos/scafacos becoming one too as it's a bit of a mess to build manually.
As @junghans suggested, we should upload a custom tarball with all submodules included (just like in the Tcl days we used to provide one which had already been bootstrapped, i.e. autoconf and automake already run).
The fix in this pull request is still necessary so that the automatic tarballs are not completely useless, but we should be directing people toward the custom one.

KaiSzuttor:
We have this custom tarball in the release project.

KaiSzuttor:
>h5xx is the only problem right now, but it won't remain that way. #1644 introduces another submodule and one day we'll have https://github.com/lssfau/walberla as a submodule. I could imagine https://github.com/scafacos/scafacos becoming one too as it's a bit of a mess to build manually.

Right, therefore we don't want to have a forest of  submodules at one day but rather use a c++ package manager.

KaiSzuttor:
>I am ok with re-tagging as the tarball from https://github.com/espressomd/release was unusable (due to missing path prefix) and hence I haven't started packaging it up yet!

@junghans what do you mean with missing path prefix?

junghans:
> @junghans what do you mean with missing path prefix?

When you unpack `espresso-4.0.0.tar.gz`all files should be in directory called `espresso-4.0.0`, just have a look at the older tarballs, but currently in `espresso-4.0.0.tar.gz` there is no subdirectory.

mkuron:
> We have this custom tarball in the release project.

Can we please get rid of https://github.com/espressomd/release? Each of these tarballs should be uploaded as a binary for the respective release at https://github.com/espressomd/espresso/releases. It's just wrong to have tarballs in a Git repository. It's not immediately obvious that these are related to the espresso repository. And we lose functionality like GitHub's "14 commits to master since this release" right next to the download button.

junghans:
Normally you would use `git archive --prefix=<bla>`, but for submodules support you need [git-archive-all](https://github.com/Kentzo/git-archive-all).

junghans:
The other thing you should test with the tarball is:
```
make install DESTDIR=/tmp
```
and that
```
make install
```
actually installs all libraries ;-) (see #772, #1056 and #1830)

PR #2235, opened 2018-09-06
CI: do not perform checks for shanchen.

KaiSzuttor:
Since it only affects LB GPU.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -132,7 +132,7 @@ cuda-shanchen:
   stage: build
   image: gitlab.icp.uni-stuttgart.de:4567/espressomd/docker/cuda:9.0
   script:
-    - export myconfig=shanchen with_coverage=true python_version=3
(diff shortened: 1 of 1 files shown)


mkuron:
Please open an issue about missing test cases for SHANCHEN. Once these are added, your pull request needs to be reverted. Right now we don't even run the standard LB tests if SHANCHEN is enabled, so we don't even know whether it breaks the LB.

KaiSzuttor:
We don't run the standard LB test, since we can't at the moment.

RudolfWeeber:
On Thu, Sep 06, 2018 at 12:31:30PM +0000, Michael Kuron wrote:
> Please open an issue about missing test cases for SHANCHEN. Once these are added, your pull request needs to be reverted.
The long term solution is that SHANCHEN no longer is switchable via compile-time define.

fweik (approved):

PR #2240, opened 2018-09-07
Mmm2d cleanup

fweik:
Removed some code duplication from mmm2d to prepare a change I want do make.
In general I think the performance of the method could be considerably increase by
changing the loop structure, but I'm not sure if anybody uses this anyway.

### DIFF
--- a/src/core/electrostatics_magnetostatics/mmm2d.cpp
+++ b/src/core/electrostatics_magnetostatics/mmm2d.cpp
@@ -203,10 +203,12 @@ typedef struct {
 } SCCache;
 
 /** sin/cos caching */
-static SCCache *scxcache = nullptr;
+static std::vector<SCCache> scxcache;
+/* _Not_ the size of scxcache */
 static int n_scxcache;
 /** sin/cos caching */
-static SCCache *scycache = nullptr;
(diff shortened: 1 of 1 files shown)


src/core/electrostatics_magnetostatics/mmm2d.cpp
```diff
@@ -198,15 +198,17 @@ static double *gblcblk = nullptr;
...
 
 /** sin/cos caching */
-static SCCache *scxcache = nullptr;
+static std::vector<SCCache> scxcache;
+/* _Not_ the size of scxcache */
 static int n_scxcache;
 /** sin/cos caching */
-static SCCache *scycache = nullptr;
+static std::vector<SCCache> scycache;
+/* _Not_ the size of scycache */
```
KaiSzuttor, line 211: If it's *not* the size of ```scycache``` could you please add a comment describing what it is?

KaiSzuttor (approved):

src/core/electrostatics_magnetostatics/mmm2d.cpp
fweik: If only I would know. It's something like the logical size.

PR #2242, opened 2018-09-08
Remove some superfluous executable bits

junghans:
`chmod -x AUTHORS ChangeLog COPYING INSTALL NEWS README`

There are a couple more in the source, these are the ones `rpmlint` is complaining about.


hmenke:
> There are a couple more in the source

I think its okay to just remove all of them (except the ones on actual executables of course).

hmenke:
I have now removed the executable bit from all files which do not start with a shebang `#!` and added a shebang in two places where it seems appropriate that the file remains executable.

junghans (commented):
LGTM, thanks!

KaiSzuttor (approved):

PR #2243, opened 2018-09-08
cmake: install everything in python sitedir

junghans:
Without tcl is makes not much sense to install libraries in libdir
anymore.


KaiSzuttor:
#2245

KaiSzuttor:
```
CMake Error at src/core/io/writer/CMakeLists.txt:10 (install):
  install TARGETS given no LIBRARY DESTINATION for shared library target
  "H5mdCore".
```

fweik:
@KaiSzuttor ?

KaiSzuttor (approved):

PR #2244, opened 2018-09-09
cmake: workaround for cmake bug#18349

junghans:
https://gitlab.kitware.com/cmake/cmake/issues/18349

Usually I am against adding random workarounds to any code base, but this is a bug in such wide range of CMake versions that you guys will have users with this issue for years....

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -286,7 +286,10 @@ endif(WITH_VALGRIND_INSTRUMENTATION)
 #######################################################################
 
 find_package(MPI REQUIRED)
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${MPI_COMPILE_FLAGS}")
+# Workaround for https://gitlab.kitware.com/cmake/cmake/issues/18349
+foreach(_MPI_FLAG ${MPI_COMPILE_FLAGS})
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${_MPI_FLAG}")
+endforeach()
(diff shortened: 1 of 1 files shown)


mkuron (approved):

hmenke (approved):

KaiSzuttor:
@junghans have you checked if doing the following would fix the issue?:
```
set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${MPI_CXX_COMPILE_FLAGS}")
```
The [documentation](https://cmake.org/cmake/help/v3.0/module/FindMPI.html) says:
>In new projects, please use the MPI_<lang>_XXX equivalents

junghans:
@KaiSzuttor the problem persists. If you look at [my fix](https://gitlab.kitware.com/cmake/cmake/merge_requests/2361), the problem is actually in `MPI_COMPILE_OPTIONS`, which then becomes `MPI_${LANG}_COMPILE_OPTIONS`, which is then part of `MPI_CXX_COMPILE_FLAGS`.

junghans:
@KaiSzuttor: You guys required cmake-3.0 which came with https://gitlab.kitware.com/cmake/cmake/blob/v3.0.0/Modules/FindMPI.cmake, so this might be a good time to switch to `MPI_CXX_COMPILE_FLAGS` anyhow.

KaiSzuttor:
thanks, @junghans Could you please switch to the new syntax since you are touching this?

junghans:
@KaiSzuttor  Done

KaiSzuttor:
thanks @junghans

KaiSzuttor (approved):

junghans:
Fixed in CMake: https://gitlab.kitware.com/cmake/cmake/merge_requests/2368

PR #2256, opened 2018-09-11
Use ccache in the CI

KaiSzuttor:
I investigated the percentages of the build times in the CI jobs on gitlab and they were between 45 and 85 % of the total job time (neglecting builds with make_check=false). I added a mount of a local cache directory on the CI machines. We have to wait until the docker images are deployed since we need ccache to be installed for this PR to work.

* centos7: 64 %, tot: 5 min, with ccache: 3 min
* clang6: 64 %, tot: 31 min
* cuda8: 51 %, tot: 20 min 
* cuda: 45 %, tot: 21 min
* default: 70 %, tot: 15 min, with ccache: 10 min
* empty: 85 %, tot: 12 min, with ccache: 6 min
* fedora: 72 %, tot: 14 min, with ccache: 6 min
* intel15: 48 %, tot: 19 min
* intel17: 45 %, tot: 21 min, with ccache: 15 min
* maxset: 58 %, tot: 19 min , with ccache: 12 min


mkuron:
You used /cache on the host machines. That is only a 30 GB partition which will fill up quickly. Please use /scratch/cache instead.

.gitlab-ci.yml
```diff
@@ -10,6 +10,8 @@ stages:
 
 variables:
   GIT_SUBMODULE_STRATEGY: recursive
+  CCACHE_DIR: /cache
+  with_ccache: "true"
```
mkuron: Please add `CCACHE_MAXSIZE=100G` after moving the cache volume to the bigger partition.

mkuron (changes requested):

KaiSzuttor:
@mkuron do you have an idea why clang6 now fails?

mkuron:
No idea. Does it go away if you disable ccache or is it due to an updated package in the new Docker image?

CMakeLists.txt
```diff
@@ -98,6 +98,10 @@ if(WITH_CCACHE)
     else()
       set(CMAKE_CXX_COMPILER_LAUNCHER ${CCACHE} CACHE STRING "ccache executable")
     endif()
+    # See http://peter.eisentraut.org/blog/2014/12/01/ccache-and-clang-part-3/ why that is needed.
+    if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
```
fweik: Probably we should just set this to true (as is default starting from [ccache 3.3](https://ccache.samba.org/releasenotes.html)) to avoid problems in the future.

KaiSzuttor:
problem only occurs with ccache

CMakeLists.txt
mkuron: Agreed. This problem can appear on GCC at any time too.

mkuron:
You might need to add `CCACHE_CPP2` to the environment variables in .gitlab-ci.yml. I'm not sure if setting environment variables from CMake even works as expected. If you google for it, you find ugly solutions like
```
set(CCACHE "CCACHE_CPP2=yes ${CCACHE}")
```

CMakeLists.txt
```diff
@@ -86,6 +86,23 @@ if( CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND NOT APPLE )
...
 option(WARNINGS_ARE_ERRORS "Treat warnings as errors during compilation" OFF)
+option(WITH_CCACHE "Use ccache compiler invocation." OFF)
+
+if(WITH_CCACHE)
+  find_program(CCACHE ccache)
+  if(CCACHE)
+    message(STATUS "Found ccache ${CCACHE}.")
+    if(CMAKE_VERSION VERSION_LESS 3.4)
+      set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ccache)
+      set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK ccache)
```
mkuron: These should use `${CCACHE}` instead of `ccache`.

mkuron (changes requested):

.gitlab-ci.yml
```diff
@@ -12,6 +12,7 @@ variables:
   GIT_SUBMODULE_STRATEGY: recursive
   CCACHE_DIR: /cache
   CCACHE_MAXSIZE: 100G
+  CCACHE_CPP2: "yes"
```
mkuron: Do we even need this anymore with https://github.com/espressomd/docker/pull/35?

KaiSzuttor: no

mkuron (approved):

PR #2262, opened 2018-09-13
CI: make errors in the build script visible.

KaiSzuttor:
man bash:
```
Exit immediately if a simple command (see SHELL GRAMMAR above) exits with a non-zero status.  The shell does not exit if the command that fails is part of the command  list  immediately  following a while or until keyword, part of the test in an if statement, part of a && or || list, or if the command's return value is being inverted via !.  A trap on ERR, if set, is executed before the shell exits.
```

### DIFF
--- a/maintainer/CI/build_cmake.sh
+++ b/maintainer/CI/build_cmake.sh
@@ -74,7 +74,6 @@ if [ -z "$cxx_flags" ]; then
     else
         if $make_check; then
             cxx_flags="-O3"
-            fdsasdf
         else
             cxx_flags="-O0"
         fi
--- a/maintainer/CI/build_cmake.sh
+++ b/maintainer/CI/build_cmake.sh
@@ -7,6 +7,14 @@
 # notice and this notice are preserved.  This file is offered as-is,
 # without any warranty.
 
+abort()
+{
+    echo "An error occurred. Exiting..." >&2
+    echo "Command that failed: $BASH_COMMAND" >&2
+    exit 1
+}
+
+trap 'abort' 0
 set -e
 
 # HELPER FUNCTIONS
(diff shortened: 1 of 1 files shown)


mkuron:
This won't work. Many of the lines of the script are allowed to fail. E.g. `[ -z "$insource" ] && insource="false"` evaluates to false if `$insource` is set already. You'll need to add `|| true` in many places or use proper `if`s.

KaiSzuttor:
right, I will look into that. But I think it's worth the effort.

KaiSzuttor:
~~~@mkuron works for me~~~ it does not fail although it should

mkuron:
I find that hard to believe. If I try out lines like the `[ -z "$insource" ] && insource="false"` or `grep -q suse /etc/os-release` in a terminal, I see that they return nonzero exit codes. It almost seems like your `set -e` does not have the intended effect.

KaiSzuttor:
it does not help since it only captures simple statements and not the ones that are piped or are in an if statement... argh

mkuron:
Should be possible though. You just need to add a bunch of `|| true` in the right places. Or rewrite this script in Python and do the termination logic yourself.

KaiSzuttor:
this should now fail...

mkuron:
Odd, it fails only on Travis but not on GitLab.

KaiSzuttor:
Seems to do the job! I really ~~~like~~~ love bash! @mkuron happy with this solution?

KaiSzuttor:
it fails only if ```make_check``` because I put a non-existing command behind the if

mkuron:
I guess that does it. Can you explain `trap abort 0` to me? What does `abort` include and what is `0`?

KaiSzuttor:
https://mywiki.wooledge.org/SignalTrap

mkuron (approved):

KaiSzuttor:
hm, doesn't work yet

PR #2263, opened 2018-09-13
testsuite: Check planar shear profile of LB

fweik:


KaiSzuttor:
Sooo... lets rewrite LBGPU? 😒

fweik:
Well I don't know. Disable the test for GPU an tell nobody about it?
Actually I think only the conversion of the boundary velocities from MD to LB units is missing.

fweik:
@KaiSzuttor

testsuite/lb_shear.py
```diff
@@ -0,0 +1,147 @@
...
+
+def shear_flow(x, t, nu, v, h, k_max):
+    """
+    Analytical solution for driven shear flow between two plates.
+
+    Parameters
+    ----------
+    x : :obj:`float`
+        Position from the left plane.
+    nu : :obj:`float`
```
KaiSzuttor: please add doc for ```t```

testsuite/lb_shear.py
```diff
@@ -0,0 +1,147 @@
...
+        Kinematic viscosity.
+    v : :obj:`float`
+        Shear rate.
+    h : :obj:`float`
+        Distance between the plates.
+    k_max: :obj:`int`
+        Maximum considered wave number.
```
KaiSzuttor: please add doc for returned value

testsuite/lb_shear.py
```diff
@@ -0,0 +1,147 @@
...
+    ----------
+    x : :obj:`float`
+        Position from the left plane.
+    nu : :obj:`float`
+        Kinematic viscosity.
+    v : :obj:`float`
+        Shear rate.
+    h : :obj:`float`
+        Distance between the plates.
+    k_max: :obj:`int`
```
KaiSzuttor: missing space after ```k_max```

KaiSzuttor (changes requested):

testsuite/lb_shear.py
fweik: What do you mean by that?

KaiSzuttor: Sth. like
```
Returns
———-
:obj:`double` : Analytical velocity

```

fweik: Ah ok. But there is not return type as this is python. Depends on what you put in. That's also true for the other ones. In the test it is used with numpy arrays, but it would also work with floats.

KaiSzuttor: i have 5. of https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard in mind

KaiSzuttor (approved):

PR #2266, opened 2018-09-14
Always define q

### LINKED ISSUE #2248: MDA_ESP does not work without ELECTROSTATICS
Also I think there is no test.

pdebuyl:
Fixes #2248 (tentative)

Description of changes:
 - remove preprocessor barriers to the definition of `q` as a particle propertpy
 - goal: avoid failure when a feature "blindly" accesses q (here, for the MDA sample)

I file the PR now to get an idea of the CI result, as I could only test a limited number of configurations.

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?

### DIFF
--- a/src/core/particle_data.cpp
+++ b/src/core/particle_data.cpp
@@ -602,7 +602,6 @@ int set_particle_q(int part, double q) {
 constexpr double ParticleProperties::q;
 #endif
 
-
 #ifdef LB_ELECTROHYDRODYNAMICS
 int set_particle_mu_E(int part, double mu_E[3]) {
   auto const pnode = get_particle_node(part);
--- a/src/core/particle_data.cpp
+++ b/src/core/particle_data.cpp
@@ -598,6 +598,10 @@ int set_particle_q(int part, double q) {
   mpi_send_q(pnode, part, q);
   return ES_OK;
 }
+#ifndef ELECTROSTATICS
+const constexpr double ParticleProperties::q;
+#endif
+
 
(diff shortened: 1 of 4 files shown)


fweik:
This probably means that a lot of `ifdef`'s can be removed.

src/core/particle_data.cpp
```diff
@@ -598,6 +598,10 @@ int set_particle_q(int part, double q) {
   mpi_send_q(pnode, part, q);
   return ES_OK;
 }
+#ifndef ELECTROSTATICS
+const constexpr double ParticleProperties::q;
```
KaiSzuttor: you can drop the const here

pdebuyl:
Should I hunt down the `#ifdef ELECTROSTATICS` that protect access to q?

src/core/particle_data.cpp
fweik: No, `const` is needed in C++11, `constexpr` only implies `const` in C++14 an newer.

KaiSzuttor: are you sure?
https://stackoverflow.com/questions/28845058/constexpr-const-vs-constexpr-variables
https://stackoverflow.com/questions/14116003/difference-between-constexpr-and-const

fweik:
@pdebuyl that would be helpful. There is at least one for which the performance impact needs to be checked (the one in the pair force calculation).

pdebuyl:
next week :-)

fweik (approved):

Issue #2245, opened 2018-09-10
Add test for installation to CI


KaiSzuttor:
Currently we don't test the install target but we provide it.

junghans:
You should also test `make install DESTDIR=/some/dir`

RudolfWeeber:
@jngrad, this is done, isn't it?

jngrad:
Yes, you should close it.

PR #2268, opened 2018-09-17
Unit testing cmake

jngrad:
Unit tests for CMake output files: espresso binaries, Python bindings, Doxygen documentation and Sphinx documentation. Part of the `make check` workflow. If we ever need a shell-independent solution, there is [kward/shunit2](https://github.com/kward/shunit2).

Regarding #2245, the install test checks `make install DESTDIR=/some/dir`, yet for `make install` we need write rights for `/usr/local/lib`, which the end user doesn't have without `sudo`. We could either:

1. make the CMake tests exclusive to the CI workflow, testing `make install` becomes trivial
2. keep the `make check` workflow as-is and add a CI-specific target `check_cmake_CI` to test `make install`

Solution 2 allows external developers to run CMake tests with  `make check`, while solution 1 requires them to find the CI-specific CMake target and adapt it to run without sudo rights.

Note: the CMake variables `PYTHON_FOUND` and other `*_FOUND` variables defined in `/CMakeLists.txt` seem out of scope in `/testsuite/cmake/CMakeLists.txt`. Is there any way to import them? At the moment I re-declared them in `/testsuite/cmake/CMakeLists.txt` (can break if we update version numbers in `/CMakeLists.txt`):

```cmake
find_package(PythonInterp REQUIRED)
find_package(Doxygen)
find_package(Sphinx 1.6.6)
```


mkuron:
> for make install we need write rights for /usr/local/lib

You could change the destination to some other place that‘s writable, like /tmp. There is a CMake variable for that.

hmenke:
```
cmake -DCMAKE_INSTALL_PREFIX=/tmp/espresso ..
make
make install
```

testsuite/CMakeLists.txt
```diff
@@ -1,193 +1,2 @@
...
-add_dependencies(check_python_parallel pypresso python_test_data)
-
-add_custom_target(check_python_parallel_odd COMMAND ${CMAKE_CTEST_COMMAND} $(ARGS) -C parallel_odd --output-on-failure)
-add_dependencies(check_python_parallel_odd pypresso python_test_data)
-
-add_custom_target(check_python)
-add_dependencies(check_python pypresso python_test_data check_python_serial)
-
-add_dependencies(check check_python)
+add_subdirectory(python)
```
KaiSzuttor, line 1: we lost git history here... better: ```git mv CMakeLists.txt python```

jngrad: I see no difference between `testsuite/CMakeLists.txt` in upstream and `testsuite/python/CMakeLists.txt` in this PR. Commit c273d64 shows everything was renamed with `git mv`. Commit [8e80916](https://github.com/espressomd/espresso/pull/2268/commits/8e80916691bf59fa41c566a57180d339a3a86a1c) applied the changes by junghans to preserve git history.

KaiSzuttor:
the ```*FOUND``` variables are available in the subdirectories, for ```PythonInterp``` it's ```PYTHONINTERP_FOUND```

testsuite/CMakeLists.txt
KaiSzuttor: ah git is not good in tracking moves...

jngrad:
@KaiSzuttor: thanks, I see the issue with `*_FOUND` variables now
@mkuron, @hmenke: good idea! CI now uses /tmp to test the standard installation procedure. This test is disabled for the user, as it would collide with the normal installation workflow (CMake environment variables set by the user would be erased). Is it necessary to add a cleanup function, or is /tmp automatically cleaned after CI?

KaiSzuttor:
We use docker containers / VM, so you dont have to worry about /tmp

jngrad:
Good to know:
* depending on the linux distribution, Python modules are installed in `pythonX.Y/dist-packages` or `pythonX.Y/site-packages`
* mpiexec and its siblings from the Open MPI suite can stall indefinitely when running a bash script containing two subprocesses with the syntax `(set -e; python -m ...)`. Ctrl+C does not terminate the process, Ctrl+Z hangs the process but does not terminate it, such that you cannot cleanup files created by the unit tests until you SIGTERM mpiexec in `htop`. I've disabled MPI for CMake tests.

jngrad:
The CMake tests are now included in `/maintainer/CI/build_cmake.sh` and are running in CI, however Sphinx needs some troubleshooting. Depending on the container, either Sphinx extensions are missing:
```Extension error:
Could not import extension sphinxcontrib.bibtex (exception: No module named bibtex)
make[7]: *** [doc/sphinx/CMakeFiles/sphinx] Error 1
```
or it cannot load a Python module:
```
Warning, treated as error:
autodoc: failed to import module 'espressomd.visualization_mayavi'; the following exception was raised:
No module named 'particle_data'
make[7]: *** [doc/sphinx/CMakeFiles/sphinx] Error 2
doc/sphinx/CMakeFiles/sphinx.dir/build.make:57: recipe for target 'doc/sphinx/CMakeFiles/sphinx' failed
```

KaiSzuttor:
The latter is an issue of install as sphinx does not seem to find necessary files. I’m also not so sure if we want to test sphinx and doxygen for the install target since we deploy both with each merge via the CI.

jngrad:
Agreed, let's just test installation. Now most containers pass CI, however `maxset-python3`, `clang:6.0` and `intel:15` seem unable to copy python files into the build directory, despite being configured with `$with_python_interface` set to true.

jngrad:
Alright, it was just a race condition in the Makefile for two containers. I also found an interaction with Clang's UBSan that prevents python from importing the espressomd module after installation in `/tmp`:
```
File "/tmp/espresso-unit-tests/lib/python2.7/dist-packages/espressomd/__init__.py", line 23, in <module>
    import espressomd._init
ImportError: /builds/espressomd/espresso/build/src/core/grid_based_algorithms/liblbgpu_cuda.so: undefined symbol: __ubsan_vptr_type_cache
```
Not really sure why a shared object in the build directory would be called at that point (linking error?), nor why the error doesn't show up before installation. I get a different pointer error when compiling espresso with UBSan on my Ubuntu machine. UBSan is currently enabled only in container `clang-6.0`; I've changed the CMake test to skip the import test when UBSan is enabled.

Then, code coverage stopped working for container `cuda-maxset` (yet unit tests pass).

KaiSzuttor:
maybe we should try to get the test pass also with UBSan. @mkuron can you please have a look?

mkuron:
You need the `LD_PRELOAD` from the pypresso script in order to inject ubsan _into python itself_.

jngrad:
Using `pypresso` instead of `python2` exits the interpreter on my machine with `UndefinedBehaviorSanitizer: failed to parse suppressions` instead of raising warnings. I tried removing lines in `/tools/ubsan-suppressions.txt` to check if one of them caused the parsing error, but the message did not change (except when the file is completely empty, causing a read error).

KaiSzuttor:
@jngrad could you please have a look at my changes?

jngrad:
Looks good to me.

testsuite/cmake/test_doxygen.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
```
KaiSzuttor: since we decided to not test sphinx nor doxygen, we can remove the testscripts.

testsuite/cmake/test_sphinx.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
```
KaiSzuttor: this file can also be removed

KaiSzuttor (changes requested):

testsuite/cmake/test_doxygen.sh
jngrad: yes

testsuite/cmake/test_sphinx.sh
jngrad: yes

KaiSzuttor (approved):

KaiSzuttor:
@jngrad i don't understand the interplay between your cmake tests and the failing coverage commands, any idea?

jngrad:
I've got the same coverage command issue in #2280 despite only changing comment lines in that PR.

jngrad:
Our GitLab instance crashed on Saturday. It automatically restarted afterwards, then a few hours later I got the coverage command issue here (57d8d12) and on the other PR (8fedeb9). The issue seems random, as your commit 31b76f8 passed while my next commit 71fa626 did not (although I just re-started the failed test with container `cuda-shanchen` and it now passes).

jngrad:
Maybe [this](https://github.com/travis-ci/travis-ci/issues/7830) is relevant: I used `git commit --amend --no-edit; git push --force` in this PR and a similar rebase operation the other PR, while CI was still running in both cases. I'll halt any pending CI before `push --force` in the future.

fweik:
@KaiSzuttor @jngrad What is the status on this? You should merge this soon to avoid merge conflicts.

KaiSzuttor (approved):

PR #2271, opened 2018-09-18
Fix MPI on Ubuntu 18.04 with CUDA

mkuron:
OpenMPI 2.1.1 has a broken _vader_ BTL (byte-transport layer), requiring us to disable its single-copy mode. The brokenness manifests itself in messages like
```
Read -1, expected 4000000, errno = 14
```
and leads to broken MPI communication between multiple ranks on the same machine. This seems to somehow be CUDA-related as Espresso compiled without CUDA support does not have the issue. As Ubuntu hasn't backported the fix from 2.1.3 (https://www.mail-archive.com/users@lists.open-mpi.org/msg32357.html), we have to resort to disabling vader's single-copy mode. This makes _vader_ behave similar to the old _sm_ BTL, which was the default before OpenMPI 2.

This patch has the same effect as setting the environment variable `OMPI_MCA_btl_vader_single_copy_mechanism=none` on OpenMPI 2.0-2.1.2 and 3.0.0. ~~However, we cannot set that from inside Espresso, so we use the MPI_T interface instead.~~

Reported by @pkreissl.

### DIFF
--- a/src/core/communication.cpp
+++ b/src/core/communication.cpp
@@ -227,7 +227,7 @@ int mpi_check_runtime_errors(void);
  * procedures
  **********************************************/
 
-#if defined(OPEN_MPI) && defined(MPI_T_ERR_NOT_INITIALIZED)
+#if defined(OPEN_MPI)
 /*! Workaround for "Read -1, expected XXXXXXX, errno = 14" that sometimes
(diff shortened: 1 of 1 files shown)


pkreissl (approved):
Commit 43b49c3 fixes the issue.

fweik (approved):

mkuron:
Please milestone for Espresso 4.0.1

PR #2272, opened 2018-09-20
Test observable_cylindricalLB: Split CPU and GPU Parts and Correct skipIfs

pkreissl:
The test `observable_cylindricalLB.py` has incorrect `skipIf`s. It is skipped if `LB_GPU` is not compiled in (or `SHANCHEN` is activated), however it also relies on the `LB` feature.
The test wrongly fails, when  `LB_GPU` is compiled **and** `LB` is not compiled in.

Is fixed by this PR, which splits cpu and gpu parts of the test functions and skips correctly.

### DIFF
--- a/testsuite/observable_cylindricalLB.py
+++ b/testsuite/observable_cylindricalLB.py
@@ -1,19 +1,19 @@
 # Copyright (C) 2010-2018 The ESPResSo project
-#  
+#
 # This file is part of ESPResSo.
-#  
+#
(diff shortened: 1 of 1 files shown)


KaiSzuttor:
the whole test is poorly written, I think it's a quickfix but we need to improve that testcase and remove code duplication by using inheritance like in the other tests that test cpu and gpu LB.

fweik (approved):

PR #2273, opened 2018-09-21
Icc stuff

fweik:
Description of changes:
 - Made icc initialization faster
 - Fixed ghost charge synchronization issue
 - Cleanup
 - Allow elc and icc to be used together

### DIFF
--- a/src/core/electrostatics_magnetostatics/icc.cpp
+++ b/src/core/electrostatics_magnetostatics/icc.cpp
@@ -126,9 +126,7 @@ void iccp3m_alloc_lists() {
 
   iccp3m_cfg.areas.resize(n_ic);
   iccp3m_cfg.ein.resize(n_ic);
-  iccp3m_cfg.nvectorx.resize(n_ic);
-  iccp3m_cfg.nvectory.resize(n_ic);
-  iccp3m_cfg.nvectorz.resize(n_ic);
+  iccp3m_cfg.normals.resize(n_ic);
   iccp3m_cfg.sigma.resize(n_ic);
 }
 
@@ -166,29 +164,21 @@ int iccp3m_iteration() {
         auto const del_eps = (iccp3m_cfg.ein[id] - iccp3m_cfg.eout) /
                              (iccp3m_cfg.ein[id] + iccp3m_cfg.eout);
         /* calculate the electric field at the certain position */
-        auto const ex = p.f.f[0] / p.p.q + iccp3m_cfg.extx;
-        auto const ey = p.f.f[1] / p.p.q + iccp3m_cfg.exty;
-        auto const ez = p.f.f[2] / p.p.q + iccp3m_cfg.exty;
-        /* let's add the contribution coming from the external field */
+        auto const E = p.f.f / p.p.q + iccp3m_cfg.ext_field;
 
-        if (ex == 0 && ey == 0 && ez == 0) {
+        if (E[0] == 0 && E[1] == 0 && E[2] == 0) {
           runtimeErrorMsg()
               << "ICCP3M found zero electric field on a charge. This must "
                  "never happen";
         }
 
-        /* the dot product   */
-        auto const fdot = ex * iccp3m_cfg.nvectorx[id] +
-                          ey * iccp3m_cfg.nvectory[id] +
-                          ez * iccp3m_cfg.nvectorz[id];
         /* recalculate the old charge density */
         auto const hold = p.p.q / iccp3m_cfg.areas[id];
         /* determine if it is higher than the previously highest charge
          * density */
-
         hmax = std::max(hmax, std::abs(hold));
 
-        auto const f1 = del_eps * fdot * pref;
+        auto const f1 = del_eps * pref * (E * iccp3m_cfg.normals[id]);
         auto const f2 = (not iccp3m_cfg.sigma.empty())
                             ? (2 * iccp3m_cfg.eout) /
                                   (iccp3m_cfg.eout + iccp3m_cfg.ein[id]) *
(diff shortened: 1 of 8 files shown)


src/core/electrostatics_magnetostatics/icc.cpp
```diff
@@ -126,194 +121,118 @@ inline void add_non_bonded_pair_force_iccp3m(Particle *p1, Particle *p2,
...
+    return 0;
 
   iccp3m_sanity_check();
 
   if ((iccp3m_cfg.eout <= 0)) {
-    ostringstream msg;
-    msg << "ICCP3M: nonpositive dielectric constant is not allowed. Put a "
+    runtimeErrorMsg()
+        << "ICCP3M: nonpositive dielectric constant is not allowed. Put a "
            "decent exception here\n";
```
KonradBreitsprecher: Decent enough I guess

fweik:
Gitlab passed (https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/pipelines/3518), not sure why it is now showing.

mkuron:
@fweik, we had a Gitlab outage this morning, so half an hour's worth of pull requests was not mirrored. You have to add another commit if you want to rebuild this pull request. The link you posted is unrelated and for #1924.

src/core/electrostatics_magnetostatics/icc.cpp
```diff
@@ -126,194 +121,118 @@ inline void add_non_bonded_pair_force_iccp3m(Particle *p1, Particle *p2,
...
-      } /* cell particles */
-    }   /* local cells */
+      }
+    } /* cell particles */
+    /* Update charges on ghosts. */
+    ghost_communicator(&cell_structure.exchange_ghosts_comm);
+
     iccp3m_cfg.citeration++;
+
     MPI_Allreduce(&diff, &globalmax, 1, MPI_DOUBLE, MPI_MAX, comm_cart);
```
KonradBreitsprecher, line 219: ICC_DEBUG is unused, I suggest to add sth like this here:
```
#ifdef ICC_DEBUG
    fprintf(stderr, "Iteration %d globalmax %f\n", iccp3m_cfg.citeration, globalmax);
#endif
```

src/core/electrostatics_magnetostatics/icc.cpp
```diff
@@ -126,194 +121,118 @@ inline void add_non_bonded_pair_force_iccp3m(Particle *p1, Particle *p2,
...
+            (1. - iccp3m_cfg.relax) * hold + (iccp3m_cfg.relax) * (f1 + f2);
+
+        /* Take the largest error to check for convergence */
+        auto const relative_difference =
+            std::abs(1 * (hnew - hold) / (hmax + std::abs(hnew + hold)));
+
+        diff = std::max(diff, relative_difference);
+
+        p.p.q = hnew * iccp3m_cfg.areas[id];
+
```
KonradBreitsprecher, line 197: ICC_DEBUG is unused, I suggest to add sth like this here:

```
#ifdef ICC_DEBUG
   fprintf(stderr, "Particle %d q_old %10.5f q_new %10.5f F_n %10.5f E %10.5f %10.5f %10.5f\n", p.p.identity, hold * iccp3m_cfg.areas[id], p.p.q, f1, ex,ey,ez);
#endif
```

KonradBreitsprecher (changes requested):
Much cleaner. About deactivating ICC, I guess it's ok to leave the ICC charges on their last value, resetting to zero can be easily done on python level.
I'd like to have the ICC_DEBUG stuff in, otherwise it's fine. ICC+ELC needs testing however, see issue #2278.

src/core/electrostatics_magnetostatics/icc.cpp
```diff
@@ -126,194 +121,118 @@ inline void add_non_bonded_pair_force_iccp3m(Particle *p1, Particle *p2,
...
+      if (p.p.identity < iccp3m_cfg.n_ic + iccp3m_cfg.first_id &&
+          p.p.identity >= iccp3m_cfg.first_id) {
+        auto const id = p.p.identity - iccp3m_cfg.first_id;
+        /* the dielectric-related prefactor: */
+        auto const del_eps = (iccp3m_cfg.ein[id] - iccp3m_cfg.eout) /
+                             (iccp3m_cfg.ein[id] + iccp3m_cfg.eout);
+        /* calculate the electric field at the certain position */
+        auto const ex = p.f.f[0] / p.p.q + iccp3m_cfg.extx;
+        auto const ey = p.f.f[1] / p.p.q + iccp3m_cfg.exty;
+        auto const ez = p.f.f[2] / p.p.q + iccp3m_cfg.exty;
```
KonradBreitsprecher: Should be ext**z** here.

fweik:
Good catch.

fweik:
Do you see any benefit for supporting MMM2D with ICC when you could use ELC instead? It seems to me that ELC supports more boundary conditions and is faster...

KonradBreitsprecher:
Not really, but if we figured out the validity of 2D methods + ICC, why would you take it out?

fweik:
Wouldn't it be better to collect proper statistics (maybe on demand) than the debug statements? How do you use this?

KonradBreitsprecher:
I collect the number of iterations to adjust relaxation parameter and convergence criterion to minimize ICC iterations. This is actually quite essential and should be accessible. The detailed debug statement for all ICC particles with old and new charges etc. is probably not worth interfacing.

fweik:
Ok I've exposed the number of iterations needed in the last round. Is that sufficient?

KonradBreitsprecher:
That should do it.

KonradBreitsprecher (approved):

PR #2274, opened 2018-09-21
testsuite: Added momentum convervation test for inertialess tracers.

fweik:
Is apparently broken. The z-component behaves notably different. The tests that existed so fare were mostly pointless because they did not check dynamics. There is more stuff broken: the tracers a lagging behind one time step with regard to the non-virtual particles, which causes forces between co-moving particles.


fweik:
@RudolfWeeber can you please look at the test and check that I'm no being stupid.

PR #2285, opened 2018-09-28
Update PDF of tutorial 6 with bibliography

hmenke:
Last time the PDF was updated, BibTeX was not run and therefore the bibliography and all the cross-references were missing.  This is what the first paragraph currently looks like:

test

Unfortunately it was shipped like this with ESPResSo 4.0


mkuron (approved):
Please tag for 4.0.1

PR #2292, opened 2018-10-02
Weeks-Chandler-Anderson Potential

fweik:
Fixes me typing `2**(1./6.)` a lot. Is also a little bit faster than `LENNARD_JONES`.

Description of changes:
 - Added an explicit WCA pair potential.

PR Checklist
------------
 - [x] Tests?
   - [x] Interface
   - [x] Core 
 - [x] Docs?


hmenke (approved):
Looks good to me.  Just one design question: `LennardJonesInteraction` is spelled out, but `WCAInteraction` is abbreviated.  Should it be more consistent?

fweik:
~~Ah you are probably right. I'll rename it in the interface. In the the stuff is also called lj.~~ 
Nah, `non_bonded_inter[0,0].weeks_chandler_anderson.set_params` is too long. @KaiSzuttor what do you think?

KaiSzuttor:
In the context of interactions i think WCA is a well known abbreviation whereas I would say LJ is not too common. In papers we always talk about WCA interaction but would not write LJ interaction. Thus, I think it's fine as it is.

doc/sphinx/inter_non-bonded.rst
```diff
@@ -216,6 +216,40 @@ interaction, while :math:`\delta` varies how smoothly the potential goes to zero
 alchemical transformations, where a group of atoms can be slowly turned
 on/off during a simulation.
 
+.. _Weeks-Chandler-Anderson interaction:
+
+Weeks-Chandler-Anderson interaction
```
KaiSzuttor: it's Andersen not Anderson

KaiSzuttor (changes requested):
please change doc for the typo

doc/sphinx/inter_non-bonded.rst
```diff
@@ -216,6 +216,40 @@ interaction, while :math:`\delta` varies how smoothly the potential goes to zero
...
+
+This command defines a Weeks-Chandler-Anderson interaction between particles of the
+types ``type1`` and ``type2``. The potential is defined by
+
+.. math::
+
+   \label{eq:wca}
+     V_\mathrm{WCA}(r) =
+       \begin{cases}
+         4 \epsilon \left[ \left(\frac{\sigma}{r}\right)^{12}
```
KaiSzuttor, line 242: It may be handy to allow an ```offset``` similar to the offset in the Lennard-Jones interaction.

fweik: My point was to rather keep it simple, as there is already a Lennard-Jones that has all the bells and whistles.

KaiSzuttor: It's open for extension, so I'm okay

KaiSzuttor (approved):

PR #2299, opened 2018-10-11
Document CUDA workaround on Ubuntu 18.04

mkuron:
Requested by @jdegraaf 

Please tag for 4.0.1

### DIFF
--- a/doc/sphinx/installation.rst
+++ b/doc/sphinx/installation.rst
@@ -60,10 +60,10 @@ Python
 Cython
     Cython is used for connecting the C++ core to Python
 
-.. _Installing Requirements on Ubuntu 16.04 LTS:
+.. _Installing Requirements on Ubuntu Linux:
 
-Installing Requirements on Ubuntu 16.04 LTS
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Installing Requirements on Ubuntu Linux
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 To make ESPResSo run on Ubuntu 16.04 LTS or 18.04 LTS, its dependencies can be
 installed with:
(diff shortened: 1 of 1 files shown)


fweik (approved):

fweik:
Maybe we should just recommend not using nvidia's compiler ^^

jdegraaf:
Hi Michael,

I tried the sed stuff. It seems to have modified the files correctly.
However, you might have to indicate that this also necessitates running

cmake -D CMAKE_C_COMPILER=gcc-6 -D CMAKE_CXX_COMPILER=g++-6 ..

in the build directory, since otherwise cmake defaults to the highest level
available and this won't get the stuff up and running. Or did you have
another route in mind that is better/less convoluted?

Best Wishes,

Joost

mkuron:
Once you apply this workaround, gcc-7 should work just fine. It does for me at least. gcc-6 is supported even without the workaround. Can you clear your build directory and recompile? If any error messages appear, please send them to me so we can figure out what is going wrong.

jdegraaf:
Hi Michael,

I did. But I'll try this again. When I do this, I still get:

In file included from /usr/include/host_config.h:50:0,
                 from /usr/include/cuda_runtime.h:78,
                 from <command-line>:0:
/usr/include/crt/host_config.h:121:2: error: #error -- unsupported GNU
version! gcc versions later than 6 are not supported!
 #error -- unsupported GNU version! gcc versions later than 6 are not
supported!
  ^~~~~
CMake Error at
EspressoCuda_generated_EspressoSystemInterface_cuda.cu.o.Release.cmake:215
(message):
  Error generating

/home/joost/FILES/PYPRESSO/build/src/core/CMakeFiles/EspressoCuda.dir//./EspressoCuda_generated_EspressoSystemInterface_cuda.cu.o


make[2]: ***
[src/core/CMakeFiles/EspressoCuda.dir/EspressoCuda_generated_EspressoSystemInterface_cuda.cu.o]
Error 1
make[1]: *** [src/core/CMakeFiles/EspressoCuda.dir/all] Error 2
make: *** [all] Error 2

I guess indicating that the workaround did not get triggered. Please find
attached my host_config.h file, so that you can see if the problem is in
there.

Best Wishes,

Joost

PR #2302, opened 2018-10-12
AMD GPU support

mkuron:
When built on a machine with AMD's HIP framework installed, we can now use that wherever CUDA is used in Espresso. It is still work in progress and not all tests pass on AMD yet, but considering that I only worked on this occasionally during the summer school, it was actually quite easy. It was mainly a find-and-replace job and working around some limitations (constant memory can only be used for array types; `cudaGetSymbolAddress` doesn't exist; different syntax for kernel launch, shared memory and symbol access).

I like to call this one the 300k€ patch as that is the amount of money you can save on a 50-node compute cluster if you get AMD Vega 64 cards instead of the Nvidia Tesla V100.


mkuron:
@fweik, can we get this merged already? Since this pull request touches quite many lines, it is very prone to merge conflicts. On Nvidia platforms, it only makes syntactic changes, so it is safe to merge.

On AMD platforms, it mainly needs some more testing. The testsuite passes except for some accuracy issues (EK, P3M) and some things are rather slow (Barnes Hut). I'll submit further patches within in the next few weeks once I understand the problems.

We don't have an AMD build machine yet, so we can only test whether it compiles.

hmenke (approved):

src/core/cuda_wrapper.hpp
```diff
@@ -0,0 +1,65 @@
+#pragma once
```
KaiSzuttor: prefer include guards since we are using it all over and it's not part of the standard?

KaiSzuttor (commented):
Do we want to add the hip project as a submodule?

mkuron:
The only thing we need from HIP is the FindHIP CMake module. Pulling in their entire Git repo takes unnecessarily long because it is pretty big. I'd actually like to remove FindHIP once the next version of AMD ROCm is released and just use the one installed in /opt/rocm/hip/cmake. The one that is currently shipping in 1.9.1 has a bug, so I had to use their current master version instead.

I have changed the header guards, so you can merge now, @KaiSzuttor.

KaiSzuttor (approved):

PR #2303, opened 2018-10-14
Utils

fweik:
Split up most of utils and additional check. Removed moved of the legacy vector ops. Should not cause functional changes.


jngrad:
Gitlab CI did not trigger on the last commit.

Benchmark shows a 13% increase in computational time for LJ liquid simulations from 8962c98 onwards compared to the current python branch (c1d850ef). Not sure which commit causes it, those before 8962c98 did not compile.

jngrad:
Nevermind, the latest python commit (4aa99a0) has a 36% slow-down, which goes down to 11% after merging this PR.

jngrad:
Something is wrong with the osx-python3 container. The 1h timeout was not triggered.

fweik:
This sometimes happens, I think mostly because the machine the vm runs on is rebooted for some reason. In this case the jobs need to be restarted in gitlab.

jngrad (approved):
looks good

PR #2314, opened 2018-10-18
Fully connected

bindgens1:
Fixes #

Description of changes:
Added fully_connected property + test in the domain decomposition cell system. This is required to support Lees Edwards boundary conditions.

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?

### DIFF
--- a/src/core/domain_decomposition.cpp
+++ b/src/core/domain_decomposition.cpp
@@ -498,9 +498,10 @@ void dd_update_communicators_w_boxl() {
 void dd_init_cell_interactions() {
   int m, n, o, p, q, r, ind1, ind2;
 
-for(int i=0; i<3; i++) {
-  if (dd.fully_connected[i] == true and node_grid[i] != 1) {
-    runtimeErrorMsg() << "Node grid not compatible with fully_connected property"; 
+  for (int i = 0; i < 3; i++) {
+    if (dd.fully_connected[i] == true and node_grid[i] != 1) {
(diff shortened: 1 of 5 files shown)


RudolfWeeber:
Is my impression correct that 
if fully_connected[i]
then all cells in the boundary layer perpendicular to the
i-direciotn are connected with all cells of the layer across the
boundary?
If so, that is more than needed for Lees-Edwards, I think.
It should be enough to connect with the same column and its two
immediate neighbors parallel to the shear direction.
The fully_connected property therefore needs to encode the shear
direciotn rather than just true/false.

Or am I missing something?

fweik:
@RudolfWeeber can you please elaborate? Fully connected means that all cells in this direction are neighbors, not only in the boundary layer. It is correct that for LE it would be sufficient to implement this
only in the boundary layer, but this was easier to implement. It should not be hard to make this more
efficient (`if(fully_connected[i] && is_boundary_cell[i])`...). This would actually make the overhead an surface effect...

bindgens1:
@RudolfWeeber Shouldn't the user set the fully connected property based on the protocol / shearing he wants to apply? Also: If ESPResSo sets the fully connected property based on the shearing protocol, who prevents the user to manually choose another cell system or fully connected property. I think this should be rather solved with a proper manual in the user guide.

fweik (approved):

PR #2318, opened 2018-10-23
Lb index

### LINKED ISSUE #2300: py; lbgpu doesn't complain when accessing non-existing nodes
lbf[x,y,z].velocity
when  on of x,y,z > grid dim

KaiSzuttor:
Fixes #2300

Description of changes:
 - added a function to check validity of node index for LB
 - added testcase on interface level

PR Checklist
------------
 - [x] Tests?
   - [x] Interface
   - [ ] Core

### DIFF
--- a/src/core/Vector.hpp
+++ b/src/core/Vector.hpp
@@ -164,6 +164,7 @@ private:
 // Useful typedefs
 
 typedef Vector<3, double> Vector3d;
+typedef Vector<3, int> Vector3i;
 typedef Vector<2, double> Vector2d;
 
 namespace detail {
--- a/src/core/grid_based_algorithms/lb.cpp
+++ b/src/core/grid_based_algorithms/lb.cpp
@@ -28,6 +28,7 @@
  */
 
 #include "grid_based_algorithms/lb.hpp"
+#include "grid_based_algorithms/lbgpu.hpp"
 #include "nonbonded_interactions/nonbonded_interaction_data.hpp"
 
 #ifdef LB
@@ -36,7 +37,6 @@
 #include "communication.hpp"
 #include "global.hpp"
 #include "grid.hpp"
-#include "grid_based_algorithms/lb.hpp"
 #include "grid_based_algorithms/lbboundaries.hpp"
 #include "halo.hpp"
 #include "lb-d3q19.hpp"
@@ -678,7 +678,7 @@ int lb_lbfluid_print_vtk_boundary(char *filename) {
 #endif // LB_GPU
   } else {
 #ifdef LB
-    int pos[3];
+    Vector3i pos;
     int boundary;
     int gridsize[3];
 
@@ -792,7 +792,7 @@ int lb_lbfluid_print_vtk_velocity(char *filename, std::vector<int> bb1,
     bb_high.push_back(std::max(*val1, *val2));
   }
 
-  int pos[3];
+  Vector3i pos;
   if (lattice_switch & LATTICE_LB_GPU) {
 #ifdef LB_GPU
     size_t size_of_values = lbpar_gpu.number_of_nodes * sizeof(LB_rho_v_pi_gpu);
@@ -881,7 +881,7 @@ int lb_lbfluid_print_boundary(char *filename) {
 #endif // LB_GPU
   } else {
 #ifdef LB
-    int pos[3];
+    Vector3i pos;
     int boundary;
     int gridsize[3];
 
@@ -941,7 +941,7 @@ int lb_lbfluid_print_velocity(char *filename) {
 #endif // LB_GPU
   } else {
 #ifdef LB
-    int pos[3];
+    Vector3i pos;
     double u[3];
     int gridsize[3];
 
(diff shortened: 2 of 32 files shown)


src/core/Vector.hpp
```diff
@@ -164,6 +164,7 @@ template <size_t n, typename Scalar> class Vector {
 // Useful typedefs
 
 typedef Vector<3, double> Vector3d;
+typedef Vector<3, int> Vector3dInt;
```
fweik: That should be called `Vector3i`. Please also use `using` over `typedef`.

RudolfWeeber:
I thought the d in vector3d stood for double? then it would be vector3i.

Also, for the bugfix release, this pr should be split into a part that fixes the bug and will go into 4.0.1, and other changes, that won't.

src/core/grid_based_algorithms/lb.cpp
```diff
@@ -1556,7 +1576,9 @@ int lb_lbnode_set_pop(int *ind, double *p_pop) {
   return 0;
 }
 
-int lb_lbnode_set_extforce_density(int *ind, double *f) { return -100; }
+int lb_lbnode_set_extforce_density(const Vector3dInt &ind, double *f) {
+  return -100;
```
fweik: What does `-100` mean?

src/python/espressomd/utils.pxd
```diff
@@ -92,6 +92,10 @@ cdef extern from "<limits>" namespace "std" nogil:
         T max()
 
 cdef extern from "Vector.hpp":
+    cppclass Vector3dInt:
```
fweik: Cython has in principal template support, (with the `[]` syntax, like `vector[double]`). Can you please try if you can get this to work without the code duplication with the `Vector3d`?

src/core/grid_based_algorithms/lb.cpp
```diff
@@ -1167,7 +1167,25 @@ int lb_lbfluid_load_checkpoint(char *filename, int binary) {
   return ES_OK;
 }
 
-int lb_lbnode_get_rho(int *ind, double *p_rho) {
+bool lb_lbnode_is_index_valid(const Vector3dInt &ind) {
+  auto within_bounds = [](const Vector3dInt &ind, const Vector3dInt &limits) {
+    return ind < limits;
```
fweik: I think this should be `ind >= {} && ind < limits`, as the indices are signed.

fweik (changes requested):

src/core/Vector.hpp
KaiSzuttor: ```using``` and ```typedef``` are equivalent and i adopted to the existing usage.

src/core/grid_based_algorithms/lb.cpp
KaiSzuttor: Who knows?

src/python/espressomd/utils.pxd
KaiSzuttor: It's not possible, cython does not support non-type template parameters.

src/core/grid_based_algorithms/lb.cpp
fweik: I'm not an expert in these matters, but this does not look like a sensible function to me.

KaiSzuttor:
@RudolfWeeber  what part should not go into bugfix release?

RudolfWeeber:
On Tue, Oct 23, 2018 at 06:37:27AM -0700, Kai Szuttor wrote:
> @RudolfWeeber  what part should not go into bugfix release?
Everything which is not directly related to checkint ghe indices in LB.

To make sure that no new issues are introduced and to still be able to cherry-pick bugfixes with minimal edits, those fixes have to be as small as possible.
You can simply rebase your changes into two separate commits, one which introduces Vector3i and sorts out the lb index checking, and another with everything else.

src/python/espressomd/utils.pxd
fweik:
```c++
template<typename T>
using Vector3 = Vector<3, T>;
```?

fweik (approved):

src/python/espressomd/utils.pxd
KaiSzuttor: Oh, I misunderstood... that should be possible

KaiSzuttor:
@RudolfWeeber I'm not sure how to go about this

fweik:
@KaiSzuttor you can only import `from .lb cimport lb_lbnode_is_index_valid` if `LB`.

RudolfWeeber:
On Tue, Oct 23, 2018 at 07:21:50AM -0700, Kai Szuttor wrote:
> @RudolfWeeber I'm not sure how to go about this
One way is to save the diff for the entire pr, apply it to a fresh
branch forked from upstream.
Then you can add use 
git add
to add only the part related to lb (Vecotr.hpp, lb*.?pp, lb*.cu,
src/python/espressomd/lb.p?)
and commit that.
This can then be cherry-picked into the bugfix branch.
In a 2nd commit, you can add the use of the Vector3i in all the
other places.

KaiSzuttor:
@RudolfWeeber I changed the commit history accordingly (git rebase espressomd/python; git reset upstream/python, then recommit).

PR #2324, opened 2018-10-25
CI: perform style_check first.

KaiSzuttor:


mkuron (commented):
Looks good, but please put it into the existing `prepare` stage instead of adding a new one. That way, the job can already run while the other preparation task is performed.

mkuron (approved):

PR #2341, opened 2018-10-28
Core: VS tracers: don't complain about missing lb for non-virtual par…

### LINKED ISSUE #2204: Inertialess tracers complains when non-virtual particle is integrated before initializing LB
```
from espressomd import lb, System
from espressomd.virtual_sites import VirtualSitesInertialessTracers

system = System(box_l=[10,10,10])
system.time_step = 0.01
system.cell_system.skin = 0.4
system.set_random_state_PRNG()

system.virtual_sites = VirtualSitesInertialessTracers()
system.virtual_sites.have_velocity = True

system.part.add(id=0,pos=[5,5,5])
system.integrator.run(10)

lbf = lb.LBFluidGPU(agrid=1, dens=1, visc=1, fric=1, tau=0.01)
system.part[:].virtual = 1
system.actors.add(lbf)
system.integrator.run(10)
```

Running the above script throws the following error
```
ERROR: Inertialess Tracers: No LB method was active.
  File "test_tracers.py", line 13, in <module>
    system.integrator.run(10)
```

The error occurs when the non-virtual particle is integrated before LB is initialized. I believe that `VirtualSitesInertialessTracers()` should not have an effect on integrating non-virtual particles whether or not LB is activated.

RudolfWeeber:
Fixes #2204

### DIFF
--- a/src/core/virtual_sites/VirtualSitesInertialessTracers.cpp
+++ b/src/core/virtual_sites/VirtualSitesInertialessTracers.cpp
@@ -39,7 +39,13 @@ void VirtualSitesInertialessTracers::after_force_calc() {
     return;
   }
 #endif
-  runtimeErrorMsg() << "Inertialess Tracers: No LB method was active.";
+  for (auto &p : local_cells.particles()) {
+    if (p.p.is_virtual) {
+      runtimeErrorMsg() << "Inertialess Tracers: No LB method was active but "
+                           "virtual sites present.";
+      return;
+    }
+  }
 }
 
 void VirtualSitesInertialessTracers::after_lb_propagation() {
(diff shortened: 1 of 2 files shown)


src/core/virtual_sites/VirtualSitesInertialessTracers.cpp
```diff
@@ -39,7 +39,13 @@ void VirtualSitesInertialessTracers::after_force_calc() {
     return;
   }
 #endif
-  runtimeErrorMsg() << "Inertialess Tracers: No LB method was active.";
+  for (auto &p : local_cells.particles()) {
```
fweik:
```c++
#include <boost/algorithm/cxx11/any_of.hpp>

[...]

if(boost::any_of(local_cells.particles(), [](Particle const& p) {
  return p.p.is_virtual;
}) {
      runtimeErrorMsg() << "Inertialess Tracers: No LB method was active but "
                           "virtual sites present.";
}
```
is almost a correct English sentence and does short circuit.

fweik (changes requested):

fweik (approved):

PR #2348, opened 2018-10-29
LB philox RNG

### LINKED ISSUE #2321: Replace LB GPU RNG
The LB GPU fluid thermalization uses a custom implementation of ran01, which is untested. This should be replaced by library code.

KaiSzuttor:
Fixes #2321

Description of changes:
 - use philox rng from curand library


src/core/grid_based_algorithms/lbgpu.hpp
```diff
@@ -30,7 +30,7 @@
 #include "config.hpp"
 #include "utils.hpp"
 #ifdef LB_GPU
-
+#include <curand_kernel.h>
```
mkuron: This is a CUDA header file and should thus not be included in lbgpu.hpp.

src/core/grid_based_algorithms/lbgpu.hpp
```diff
@@ -180,6 +180,7 @@ typedef struct {
   unsigned int *seed;
   /** flag indicating whether this site belongs to a boundary */
   unsigned int *boundary;
+  curandStatePhilox4_32_10_t *philox_state;
```
mkuron: Use a void pointer so you don't need the header file.

src/core/grid_based_algorithms/lbgpu_cuda.cu
```diff
@@ -41,6 +41,7 @@
 #include "grid_based_algorithms/electrokinetics_pdb_parse.hpp"
 #include "grid_based_algorithms/lbgpu.hpp"
 
+#include <curand_kernel.h>
```
mkuron: This needs a wrapper like [cufft_wrapper.hpp](https://github.com/espressomd/espresso/blob/python/src/core/cufft_wrapper.hpp), otherwise your build will fail on AMD GPUs.

src/core/grid_based_algorithms/lbgpu_cuda.cu
```diff
@@ -217,6 +218,11 @@ __device__ void gaussian_random(LB_randomnr_gpu *rn) {
   rn->randomnr[1] = x1 * fac;
 }
 /* wrapper */
+__device__ float random_wrapper_philox(curandStatePhilox4_32_10_t *state) {
+  constexpr float sqrt12 = 3.46410161514f;
+  return (curand_uniform(state) - 0.5f) * sqrt12;
+}
+
```
mkuron, line 235: Please directly call `curand_Philox4x32_10`, then you don't need the state object. It just takes four integers as counter and four integers as key.

src/core/grid_based_algorithms/lbgpu_cuda.cu
```diff
@@ -818,26 +824,24 @@ __device__ void thermalize_modes(float *mode, unsigned int index,
...
        para->rho[0] * para->agrid * para->agrid * para->agrid) /
       Rho_tot;
-  random_wrapper(rn);
   for (int ii = 0; ii < LB_COMPONENTS; ++ii) {
     mode[1 + ii * LBQ] +=
         sqrtf(c * (1 - c) * Rho_tot *
               (para->mu[ii] * (2.0f / 3.0f) *
                (1.0f - (para->gamma_mobility[0] * para->gamma_mobility[0])))) *
-        (2 * ii - 1) * rn->randomnr[0];
+        (2 * ii - 1) * random_wrapper_philox(&n_a.philox_state[index]);
```
mkuron: You don't need to store a per-cell state object. That's just a waste of memory. If you do as I suggested above, you can use the cell coordinate and the mode number as the counter and the seed and the time step as the key.

src/core/grid_based_algorithms/lbgpu_cuda.cu
```diff
@@ -2964,6 +2960,7 @@ __global__ void calc_n_from_rho_j_pi(LB_nodes_gpu n_a, LB_rho_v_gpu *d_v,
 
       /**set different seed for randomgen on every node */
       n_a.seed[index] = para->your_seed + index;
+      curand_init(para->your_seed + index, 0, 0, &n_a.philox_state[index]);
```
mkuron: See above. Put the seed and the time step into the key.

src/core/grid_based_algorithms/lbgpu_cuda.cu
```diff
@@ -3936,11 +3934,16 @@ void lb_init_GPU(LB_parameters_gpu *lbpar_gpu) {
...
+  free_realloc_and_clear(nodes_a.philox_state,
+                         lbpar_gpu->number_of_nodes *
+                             sizeof(curandStatePhilox4_32_10_t));
   free_realloc_and_clear(nodes_a.seed,
                          lbpar_gpu->number_of_nodes * sizeof(unsigned int));
   free_realloc_and_clear(nodes_a.boundary,
                          lbpar_gpu->number_of_nodes * sizeof(unsigned int));
+  free_realloc_and_clear(nodes_b.philox_state,
+                         lbpar_gpu->number_of_nodes *
+                             sizeof(curandStatePhilox4_32_10_t));
```
mkuron: not needed if you make it stateless as I suggested

mkuron (changes requested):

KaiSzuttor:
@mkuron could you pinpoint the API documentation for that stateless call?

mkuron:
Unfortunately it's not documented, but it is a public function because its name begins with `curand_`. Just `#include <curand_kernel.h>`. If you look into curand_philox4x32_x.h, which is included by the other header file, you can see that the signature is
```
uint4 curand_Philox4x32_10( uint4 c, uint2 k);
```
So you get four random unsigned integers at a time. `uint4` is a struct with `unsigned int` members `w`, `x`, `y` and `z`, while `uint2` contains `x` and `y`. You may find these helper functions useful:
```
uint4 make_uint4(unsigned int x, unsigned int y, unsigned int z, unsigned int w);
uint2 make_uint2(unsigned int x, unsigned int y);
```

src/core/grid_based_algorithms/lbgpu.hpp
```diff
@@ -180,6 +179,7 @@ typedef struct {
   unsigned int *seed;
   /** flag indicating whether this site belongs to a boundary */
   unsigned int *boundary;
+  void *philox_state;
```
fweik: Please find a type safe solution for this. I think you can just move the struct out of the header. Also I'd prefer a more ideomatic name

mkuron: It's going away anyway with the switch to the stateless variant discussed above.

src/core/grid_based_algorithms/lbgpu_cuda.cu
```diff
@@ -216,7 +218,26 @@ __device__ void gaussian_random(LB_randomnr_gpu *rn) {
   rn->randomnr[0] = x2 * fac;
   rn->randomnr[1] = x1 * fac;
 }
-/* wrapper */
+
+
+__device__ static volatile unsigned int philox_counter = 0;
```
fweik: Why `volatile`?

fweik:
I don't think that his counter construction is a good choice. If `volatile` does what you want (which I don't think) this would mean an ordering of the threads, which ruins performance. I think you should construct the state from a counter that is updated once for every force calculation and the index of the node.

KaiSzuttor:
Yea its WIP

KaiSzuttor:
~~The problem that I see is that the state is not unique for SHANCHEN, where you have two fluids.~~

KaiSzuttor:
I'm considering creating an alias in my ```.zshrc```:
```bash
alias gcf=`git commit -am "Formatting."`
```

KaiSzuttor:
@fweik is that what you had in mind?

KaiSzuttor:
@mkuron it still can't find the header 
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/48089

src/core/grid_based_algorithms/lbgpu_cuda.cu
```diff
@@ -4573,6 +4590,7 @@ void lb_integrate_GPU() {
                lb_boundary_velocity, lb_boundary_force);
   }
 #endif
+  philox_counter += 1;
```
mkuron, line 4588: Just use the current timestep number here instead of adding a global static variable...

src/core/grid_based_algorithms/lbgpu_cuda.cu
```diff
@@ -806,8 +827,10 @@ __device__ void relax_modes(float *mode, unsigned int index,
  * @param *rn     Pointer to random number array of the local node
  */
 __device__ void thermalize_modes(float *mode, unsigned int index,
-                                 LB_randomnr_gpu *rn) {
+                                 LB_randomnr_gpu *rn, LB_nodes_gpu n_a,
```
mkuron: I don't think you use `n_a` anywhere.

mkuron (changes requested):

src/core/curand_wrapper.hpp
```diff
@@ -0,0 +1,13 @@
+#ifndef CURAND_WRAPPER_HPP
+#define CURAND_WRAPPER_HPP
+
+#if defined(__HIPCC__) // AMD or Nvidia via HIP
+
+#include <hiprand_kernel.h>
```
mkuron: I think you need _hiprand/hiprand_kernel.h_. It still won't work as the function's name is different, but I can take care of that.

mkuron (changes requested):

src/core/grid_based_algorithms/lbgpu_cuda.cu
KaiSzuttor: that increases coupling

mkuron: Yours increases the number of globals. Both are bad.

The advantage of using the time step is that you can use `integrate(0)` and it will produce the same LB forces every time. Also, you can then calculate the missing LB forces in the first time step (the ones that we currently print a warning about). Same after loading a checkpoint, if you checkpoint the time step number.

src/core/grid_based_algorithms/lbgpu_cuda.cu
```diff
@@ -216,7 +218,26 @@ __device__ void gaussian_random(LB_randomnr_gpu *rn) {
   rn->randomnr[0] = x2 * fac;
   rn->randomnr[1] = x1 * fac;
 }
-/* wrapper */
+
+static unsigned int philox_counter = 0;
+__device__ float4 random_wrapper_philox(unsigned int index, unsigned int mode,
+                                        unsigned int philox_counter) {
+  unsigned int xyz[3];
+  index_to_xyz(index, xyz);
```
mkuron: Just use the index directly and set the other two parameters to zero. This just wastes expensive division operations.

mkuron (changes requested):

src/core/grid_based_algorithms/lbgpu_cuda.cu
KaiSzuttor: Since there is a plan to refactor the LB we want to have it as isolated as possible. During refactoring the globals will go, the coupling to the MD core won't.

KaiSzuttor: Also it's not hard to checkpoint a single number...

mkuron: How do you want to deal with `integrate(0)`? That shouldn‘t advance the counter.

KaiSzuttor: Sure, there is an interface missing that just calculatesforces from the fluid without propagating it. Whats your point?

mkuron: You wouldn‘t need a separate function if you use the time step number directly. You have introduced a variable here that will always need to be equal to the time step number. That kind of design is just as bad as the increased coupling: If you don‘t directly use the time step number, you always risk the two numbers diverging, e.g. if you forget to take care of a special code path that doesn‘t increase the time step number.

mkuron: The place that calls `lb_integrate_GPU()` has access to the `sim_time` variable. Just pass it in from there, then you don‘t need to increase coupling.

KaiSzuttor:
@fweik

mkuron:
Perhaps @fweik can also comment on https://github.com/espressomd/espresso/pull/2348#discussion_r229956985. I'd still prefer using the time step number over keeping a separate counter because then we don't need to worry about the two diverging if we forget some code path.

fweik:
Please merge the branch with the velocity distribution test.

src/core/grid_based_algorithms/lbgpu.cuh
```diff
@@ -0,0 +1,25 @@
+#ifndef LBGPU_CUH
```
fweik, line 1: Why do you need a header for this? Just declare it in the cuda file, it's not used anywhere else.

fweik (changes requested):

src/core/grid_based_algorithms/lbgpu.cuh
KaiSzuttor: it's used in electrokinetics_cuda.cu

fweik: Ah ok I did not see that.

fweik:
@mkuron The time can be set by the user, so using this as RNG state will have surprising side-effects. This is why we settled with a separate counter. In general it would probably good to have a global `generation` counter that just enumerates all the subsequent states, and can only ever be increased;
but that's for the future I guess.

fweik (approved):

PR #2352, opened 2018-10-30
Properly detect CUDA libraries if multiple are available

mkuron:
The first two changes are needed when using Clang to compile CUDA code and the CUDA versions installed by the package manager in /usr and a manually-installed one in /usr/local/cuda are of different versions. Otherwise headers and libraries may be incompatible and you will get crashes.

The third change is needed when using regular NVCC when you specify CUDA_NVCC_EXECUTABLE or the NVCC environment variable to be somewhere other than the automatically detected toolkit path. If there is a mismatch, you can get the same problems as above.

Please tag for 4.0.1.


KaiSzuttor (approved):

RudolfWeeber:
@mkuron, I did not manage to cherry-pick this for 4.0.1, because a lot of hip-related stuff ends up in a merge conflict.
Could you please open a pull request against the 401test branch on my github repo with the changes that should go into 4.0.1?

Thanks

PR #2359, opened 2018-11-02
Lb density setters are broken

RudolfWeeber:
This PR adds Python support for setting lb node densities.
It works, when first setting density, then veloicyt but not the other way round.
For lbcpu, the velocity is rescaled by setting the density on the node, on lbgpu, the velocity is set to 0.
I think both setters in the core are likely incorrect.


fweik (approved):

fweik:
@RudolfWeeber fix or close please.

KaiSzuttor:
Isn't that behavior expected? Density and velocity are not independent. Density is sum_i n_i and velocity is
sum_i c_i n_i with n_i being the i-th population and c_i the velocity in direction i.

fweik:
This will be fixed by #2392.

PR #2363, opened 2018-11-02
Lb philox rng particle

KaiSzuttor:
Follow up on #2348 

Description of changes:
 - remove old self-written RNG for particle thermalization in LB GPU


fweik:
Sorry I did not catch this before. But 32 bit for the state seems a little bit short for me (this is also the period obviously), if you could easily use 64 bit please do so.

Also since you removed the checkpointing code for seed, how is the state checkpointed?

src/core/grid_based_algorithms/lbgpu_cuda.cu
```diff
@@ -2646,36 +2521,14 @@ __device__ void calc_viscous_force(
...
-        para->lb_coupl_pref[ii] * (rn_part->randomnr[0] - 0.5f);
-    viscforce_density[1 + ii * 3] +=
-        para->lb_coupl_pref[ii] * (rn_part->randomnr[1] - 0.5f);
-    random_01(rn_part);
-    viscforce_density[2 + ii * 3] +=
-        para->lb_coupl_pref[ii] * (rn_part->randomnr[0] - 0.5f);
-#elif defined(GAUSSRANDOMCUT)
-    gaussian_random_cut(rn_part);
+    float4 random_floats =
+        random_wrapper_philox(part_index, ii, philox_counter);
```
mkuron: Right now this generates the same random numbers as for some fluid cells. You should replace `ii` with `ii + 1024` or something.

mkuron (changes requested):

mkuron:
> But 32 bit for the state seems a little bit short for me (this is also the period obviously)

It's the period per particle, so it is sufficient. Each particle gets a completely different series of 2^32 numbers.

> Also since you removed the checkpointing code for seed, how is the state checkpointed?

That's why I said it should just use the time step number. Though I actually don't know if that is even checkpointed.

fweik:
2^32 isn't so big, and I can see no reason to leave just half of the state unused.

mkuron:
Of course we could use more bits as state, but it's probably not needed very often. Right now we can have 4 billion time steps times 4 billion seeds before anything repeats. But we imply that you have less than 4 billion particles.

fweik:
From my understanding it's not `times`, after the overflow everything will repeat, or do I misunderstand something?

fweik:
It's used like
```c++
    float4 random_floats =
random_wrapper_philox(part_index, ii, philox_counter);
```
`part_index` and `ii` are fixed, so if `philox_counter` overflows, everything repeats.

fweik:
More correctly, the distribution of `part_index` is unknown, it's maybe even not such a great choice.

mkuron:
`part_index` is the ID of the particle, so you get independent series of random numbers for each particle (as long as you have less than 4 billion of them). `philox_counter` overflows once every 4 billion time steps, so the series for each particle repeats after 4 billion time steps. `ii` is indeed fixed. The distribution of `part_index` does not matter because a hash function produces a random distribution from anything.

fweik:
`part_index` is not the id.

mkuron:
What else is it? It looks like it is the particle ID, and @KaiSzuttor probably used it because he thought that too.

fweik:
It's not, the id is not known on the device. Its some random index depending on the order of the particles in the cells and nodes.

KaiSzuttor:
urgh maybe we think about changing the particle thermalization when coupling is done on the cpu. its currently also not clear to me how we would achieve normal distribution with the “stateless” philox

KaiSzuttor:
yes, checkpointing is broken and we don’t test

mkuron:
> Its some random index depending on the order of the particles in the cells and nodes

As long as the tuple (`part_index`, `philox_counter`) is unique, we are good. And for that the order of the particles doesn’t matter as long as `part_index` is unique within a single time step.

> normal distribution with the “stateless” philox

Box-Muller maybe? Converts pairs of uniformly-distributed numbers to pairs of normal-distributed ones. That is no different from the stateful one, except that you may end up discarding a few percent of your random numbers (which is still going to be cheaper than fetching a state from memory).

fweik:
I don't like the `particle_index`, we can just copy the id to the device.

mkuron:
You are probably right. Even though it is unique, if it depends on some sort order, it's probably not deterministic. So rerunning with the same seed may produce different results.

fweik (changes requested):
The counter is still 32 bit.

KaiSzuttor:
@fweik, I'm not sure what you mean. The curand function expects uint.

mkuron (approved):

mkuron:
I think @fweik means that `philox_counter` rolls over after 4 billion time steps. This most likely isn't going to be a common thing to happen: if your simulation is running at 1000 time steps per second, it would have to run for 7 weeks before the counter rolls over.

The curand function takes 4x32 bits as counter and 2x32 bits as key. Right now we use the cell index or particle ID, two zeros, and the mode number or 32 as _counter_ and `philox_counter` and the seed as _key_. To roll over beyond 4 billion time steps, we could use a 64-bit `philox_counter` as _key_ (just reinterpret it as a `uint2`) and replace one of the zeros in _counter_ with the seed.

KaiSzuttor:
@mkuron, I was faster :smiley:

KaiSzuttor:
@fweik

fweik (approved):

PR #2364, opened 2018-11-03
Enable AMD GPU testing

mkuron:
Follow-up to #2302.

This depends on https://github.com/ROCmSoftwarePlatform/rocFFT/pull/141 and 
https://github.com/espressomd/docker/pull/45.

Right now all the important stuff (LB, EK) works. Barnes-Hut (@psci2195) and P3M GPU (@fweik) don't work; perhaps the original authors can have a look. Until that happens, I have disabled just these tests on AMD GPUs inside CI. They probably depended on some Nvidia implementation detail. The former method gets caught in an infinite loop in `sortKernel`, while the latter simply gives wrong results.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -211,11 +211,10 @@ rocm-maxset:
   stage: build
   image: gitlab.icp.uni-stuttgart.de:4567/espressomd/docker/rocm:latest
   script:
-    - export myconfig=maxset make_check=false HIP_PLATFORM=hcc HCC_AMDGPU_TARGET=gfx900
+    - export myconfig=maxset
     - bash maintainer/CI/build_cmake.sh
   tags:
-    - docker
-    - linux
+    - amdgpu
 
 ### Builds with OS X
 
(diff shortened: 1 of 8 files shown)


bogdan-tanygin:
@mkuron could you please clarify which physical AMD GPU is embedded in the machine where the Gitlab build rocm-maxset is running now? There is no command similar to nividia-smi in the build log, correct?

mkuron:
@psci2195, it's a Radeon RX Vega 64. Unfortunately AMD's `nvidia-smi`-equivalent does not run without root privileges, so we can't print the GPU model in the CI job log.

I expect you will be able to reproduce the issue on any ROCm-supported GPU, e.g. Radeon R9 Fury/Nano/Fury X, Radeon RX 460 and higher, Radoen RX 540 and higher, Radeon RX Vega. For installation instructions, see https://rocm.github.io/ROCmInstall.html#installing-from-amd-rocm-repositories.

KaiSzuttor (approved):

PR #2366, opened 2018-11-06
LB stokes sphere test case update

### LINKED ISSUE #2097: Lb Stokes sphere test on CPU

jainka:
Fixes #2097

Description of changes:
 - 

PR Checklist
------------
 - [x ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


KaiSzuttor (changes requested):
we should perform this test for CPU LB aswell

testsuite/python/lb_stokes_sphere.py
```diff
@@ -32,39 +32,43 @@
...
+
+class Stokes(object):
+    lbf = None
+    system = espressomd.System(box_l=[real_width, real_width, box_length])
+    system.box_l = [real_width, real_width, box_length]
+    system.time_step = TIME_STEP
+    system.cell_system.skin = 0.4
+
+    # The temperature is zero.
+    system.thermostat.set_lb(kT=0)
```
KaiSzuttor: this has to be done in the test method because the guard for the feature does not catch this statement which is not available when neither ```LB``` nor ```LB_GPU``` is compiled in.

KaiSzuttor (changes requested):

testsuite/python/lb_stokes_sphere.py
jainka: are you sure about that? I see that the thermostat should be set in the test method but skin is set within the main class in other test cases also (lb.py, lb_poiseuille.py)!

KaiSzuttor: yes I only meant the thermostat

KaiSzuttor (approved):

PR #2369, opened 2018-11-08
Document AMD GPU installation

mkuron:
Follow-up to #2302


hmenke (approved):

src/core/cuda_wrapper.hpp
```diff
@@ -25,7 +25,6 @@
 #define cudaFreeHost hipHostFree
 #define cudaHostAlloc hipHostMalloc
 #define cudaHostAllocWriteCombined hipHostMallocWriteCombined
-#define cudaLaunchKernelGGL hipLaunchKernelGGL
```
fweik, line 28: Is that intentional?

hmenke: Since the CI passes (apart from an unrelated GitLab CI Runner error), this function is probably unused an can be safely removed.

PR #2370, opened 2018-11-08
Normr

fweik:
Description of changes:
 - Minor modernization


src/core/bonded_interactions/subt_lj.hpp
```diff
@@ -48,25 +48,24 @@ int subt_lj_set_params(int bond_type);
     @return true if bond is broken
 */
 inline int calc_subt_lj_pair_force(Particle *p1, Particle *p2,
-                                   Bonded_ia_parameters *iaparams, double dx[3],
+                                   Bonded_ia_parameters *, double dx_[3],
```
KaiSzuttor, line 51: can't you just drop ```Bonded_ia_parameters```?

fweik: The interactions all have the same signature, I did not want to make this one special.

fweik:
@KaiSzuttor

RudolfWeeber:
Was the removal of the thermalized bon energy intentional (does it have an energy contrib?)

Apart from that, it can be merged.

fweik:
No, it does not have an energy contribution as it is not Hamiltonian.

PR #2377, opened 2018-11-15
use correct Maxwell-Boltzmann velocity distribution in reaction ensemble

### LINKED ISSUE #1770: Draw Maxwell distributed velocities in ReactionAlgorithm::do_global_mc_move_for_particles_of_type
In order to get kinetic observables right (e.g. the kinetic part of the stress tensor). Right now the velocities of particles are not changed. That means if particles are created without velocity they stay having no velocity. Static observables are not affected here.

thepith:
Fixes #1770 and #2377 

The velocities of the particles inserted by the reaction ensemble did not follow the Maxwell-Boltzmann velocity distribution [1]. Even worse, with increasing temperature the mean squared velocity was decreasing. The effect of this PR on the velocity distribution function are shown in the following figure:

maxwell_plot

As you can see, now the measured velocity distribution from the simulation (marked as _Sim_ in the plots).becomes wider with increasing temperature and follows the Maxwell-Boltzmann distribution (marked as _Theo_ in the plots).

[1]: Atkins, P. W., and J. de Paula. Atkins’ Physical Chemistry. Oxford, UK: Oxford University Press, 2006.

### DIFF
--- a/src/core/reaction_ensemble.cpp
+++ b/src/core/reaction_ensemble.cpp
@@ -717,9 +717,9 @@ int ReactionAlgorithm::create_particle(int desired_type) {
   // for components
   double vel[3];
   // we use mass=1 for all particles, think about adapting this
-  vel[0] = std::pow(2 * PI * temperature, -3.0 / 2.0) * gaussian_random();
-  vel[1] = std::pow(2 * PI * temperature, -3.0 / 2.0) * gaussian_random();
-  vel[2] = std::pow(2 * PI * temperature, -3.0 / 2.0) * gaussian_random();
+  vel[0] = std::sqrt(temperature) * gaussian_random();
+  vel[1] = std::sqrt(temperature) * gaussian_random();
+  vel[2] = std::sqrt(temperature) * gaussian_random();
 #ifdef ELECTROSTATICS
   double charge = charges_of_types[desired_type];
 #endif
--- a/src/core/reaction_ensemble.cpp
+++ b/src/core/reaction_ensemble.cpp
@@ -831,6 +831,12 @@ bool ReactionAlgorithm::do_global_mc_move_for_particles_of_type(
     p_id = p_id_s_changed_particles[i];
     // change particle position
     new_pos = get_random_position_in_box();
+    double vel[3];
+    // we use mass=1 for all particles, think about adapting this
+    vel[0] = std::sqrt(temperature) * gaussian_random();
+    vel[1] = std::sqrt(temperature) * gaussian_random();
+    vel[2] = std::sqrt(temperature) * gaussian_random();
+    set_particle_v(p_id, vel);
     // new_pos=get_random_position_in_box_enhanced_proposal_of_small_radii();
     // //enhanced proposal of small radii
     place_particle(p_id, new_pos.data());
--- a/src/core/reaction_ensemble.cpp
+++ b/src/core/reaction_ensemble.cpp
@@ -832,10 +832,10 @@ bool ReactionAlgorithm::do_global_mc_move_for_particles_of_type(
     // change particle position
     new_pos = get_random_position_in_box();
     double vel[3];
-    // we use mass=1 for all particles, think about adapting this
-    vel[0] = std::sqrt(temperature) * gaussian_random();
-    vel[1] = std::sqrt(temperature) * gaussian_random();
-    vel[2] = std::sqrt(temperature) * gaussian_random();
+    auto const &p = get_particle_data(p_id);
+    vel[0] = std::sqrt(temperature / p.p.mass) * gaussian_random();
+    vel[1] = std::sqrt(temperature / p.p.mass) * gaussian_random();
+    vel[2] = std::sqrt(temperature / p.p.mass) * gaussian_random();
     set_particle_v(p_id, vel);
     // new_pos=get_random_position_in_box_enhanced_proposal_of_small_radii();
     // //enhanced proposal of small radii
--- a/src/python/espressomd/reaction_ensemble.pyx
+++ b/src/python/espressomd/reaction_ensemble.pyx
@@ -16,9 +16,9 @@ cdef class ReactionAlgorithm(object):
     reaction algorithm by setting the standard pressure, temperature, and the
     exclusion radius.
 
-    Note: When creating particles the velocities the new particles are set
-    according the Maxwell distribution. In this step the mass of the new particle
-    is assumed to equal 1.
+    Note: When creating particles the velocities of the new particles are set
+    according the Maxwell-Boltzmann distribution. In this step the mass of the
+    new particle is assumed to equal 1.
 
 
     Parameters
--- a/src/python/espressomd/reaction_ensemble.pyx
+++ b/src/python/espressomd/reaction_ensemble.pyx
@@ -16,6 +16,9 @@ cdef class ReactionAlgorithm(object):
     reaction algorithm by setting the standard pressure, temperature, and the
     exclusion radius.
 
+    Note: When setting the velocities of the particles according the the
+    Maxwell distribution the mass of all particles is assumed to equal 1.
+
 
     Parameters
     ----------
--- a/src/python/espressomd/reaction_ensemble.pyx
+++ b/src/python/espressomd/reaction_ensemble.pyx
@@ -16,8 +16,9 @@ cdef class ReactionAlgorithm(object):
     reaction algorithm by setting the standard pressure, temperature, and the
     exclusion radius.
 
-    Note: When setting the velocities of the particles according the the
-    Maxwell distribution the mass of all particles is assumed to equal 1.
+    Note: When creating particles the velocities the new particles are set
+    according the Maxwell distribution. In this step the mass of the new particle
+    is assumed to equal 1.
 
 
     Parameters


fweik:
Thank you for your contribution. This looks more correct than before to me, but finally @jonaslandsgesell has to decide.

src/core/reaction_ensemble.cpp
```diff
@@ -717,9 +717,9 @@ int ReactionAlgorithm::create_particle(int desired_type) {
   // for components
   double vel[3];
   // we use mass=1 for all particles, think about adapting this
-  vel[0] = std::pow(2 * PI * temperature, -3.0 / 2.0) * gaussian_random();
-  vel[1] = std::pow(2 * PI * temperature, -3.0 / 2.0) * gaussian_random();
-  vel[2] = std::pow(2 * PI * temperature, -3.0 / 2.0) * gaussian_random();
+  vel[0] = std::pow(temperature, 1.0 / 2.0) * gaussian_random();
+  vel[1] = std::pow(temperature, 1.0 / 2.0) * gaussian_random();
+  vel[2] = std::pow(temperature, 1.0 / 2.0) * gaussian_random();
```
hmenke: `std::pow(temperature, 1.0 / 2.0)` is also known as `std::sqrt(temperature)` .

hmenke (changes requested):

src/core/reaction_ensemble.cpp
thepith: changed in 314115186f79749bda8bfc3280cb37dfa03fea48

src/core/reaction_ensemble.cpp
```diff
@@ -717,9 +717,9 @@ int ReactionAlgorithm::create_particle(int desired_type) {
   // for components
   double vel[3];
   // we use mass=1 for all particles, think about adapting this
-  vel[0] = std::pow(2 * PI * temperature, -3.0 / 2.0) * gaussian_random();
-  vel[1] = std::pow(2 * PI * temperature, -3.0 / 2.0) * gaussian_random();
-  vel[2] = std::pow(2 * PI * temperature, -3.0 / 2.0) * gaussian_random();
+  vel[0] = std::sqrt(temperature) * gaussian_random();
```
jonaslandsgesell, line 720: Thank you for your changes.
This should read  vel[0] = std::sqrt(temperature/particle_mass) * gaussian_random(), etc since the Maxwell-Boltzmann probability density is mass dependent. However particle_mass would be needed to be set in the reaction ensemble interface (in the same way the charge is set). Maybe we should also rename tempertature to kbT.

We could also fix issue #1770 via assigning a new Maxwell-Boltzmann distributed velocity after line https://github.com/espressomd/espresso/blob/314115186f79749bda8bfc3280cb37dfa03fea48/src/core/reaction_ensemble.cpp#L836 here.

jonaslandsgesell (changes requested):

src/core/reaction_ensemble.cpp
thepith: I would leave it as the temperature, as written in the manual:

> A special note is due regarding the temperature, which is coupled to the energy scale by Boltzmann’s constant. However, since ESPResSo does not enforce a particular unit system, we also don’t know the numerical value of the Boltzmann constant in the current unit system. Therefore, when specifying the temperature of a thermostat, you actually do not define the temperature, but the value of the thermal energy kBT in the current unit system.

Therefore it is kind of implied that the temperature equals kbT.

thepith: I added a comment on the missing mass in the documentation. I think that is enough until somebody wants to implement the correct mass.

src/python/espressomd/reaction_ensemble.pyx
```diff
@@ -16,6 +16,9 @@ cdef class ReactionAlgorithm(object):
     reaction algorithm by setting the standard pressure, temperature, and the
     exclusion radius.
 
+    Note: When setting the velocities of the particles according the the
+    Maxwell distribution the mass of all particles is assumed to equal 1.
```
jonaslandsgesell: Yes, I would note that this is not a problem as long as a thermostat is acting which thermalizes the particle velocities correctly before sampling an observable which is depending on the velocities (like e.g. the kinetic parts of the virial pressure).

src/core/reaction_ensemble.cpp
thepith: If we assign a new velocity to a particle in the global_mc move, do we have to take the change in the kinetic energy into account in the acceptance probability?

src/python/espressomd/reaction_ensemble.pyx
thepith: But it can still break you integration, if you have a particle with high mass which now has a much to high momentum.

I would let the user decide whether this is poses a problem or not.

src/core/reaction_ensemble.cpp
jonaslandsgesell: To my understanding we do not need to take the kinetic energy into account in the acceptance probability. This would be in accordance with the algorithm described in https://aip.scitation.org/doi/pdf/10.1063/1.2359441?class=pdf (_Mesoscale simulation of polymer reaction equilibrium: Combining dissipative particle dynamics with reaction ensemble Monte Carlo. I. Polydispersed polymer systems_) [search for Maxwell-Boltzmann in this paper and also recognize that the kinetic energy does not occur in the acceptance probabilities]. The reasoning behind it is that drawing a Maxwell-Boltzmann distributed velocity from time to time resembles an Andersen Thermostat. The frequency with which a new velocity needs to be assigned is arbitrary (_Allen, Tildesley Computer Simulation of Liquids, 1991, p. 228_). At least in the MC move however the particle masses are important if no other thermostat is used to thermalize the system. Otherwise e.g. the kinetic virial pressure is wrong.

thepith: I think you are right, as long as we sample from the correct distribution of the kinetic energy, then we do not need to take the kinetic energy into account. (This follows quite easily from the Hastings formalism...)

And apparently the Maxwell distribution is valid generally (see https://doi.org/10.4236/jamp.2018.63052 and https://chemistry.stackexchange.com/a/94294). I am not sure yet if this applies to more "exotic" systems (relativistic systems and quantized systems), but I guess these are not represented in espressomd anyways.

Still, I would propose, to address this in a future pull request. At least, I don't feel confident enough yet in this section of the code to claim that I know what I am doing.

jonaslandsgesell:
I think if you would add the lines


double vel[3];

vel[0] = std::sqrt(temperature) * gaussian_random();

vel[1] = std::sqrt(temperature) * gaussian_random();

vel[2] = std::sqrt(temperature) * gaussian_random();

set_particle_v(p_id, vel);

below https://github.com/espressomd/espresso/blob/314115186f79749bda8bfc3280cb37dfa03fea48/src/core/reaction_ensemble.cpp#L836


this would fix the problem #1770 (at least for the case that we have
particles with mass unity). We could then postpone the case for
particles with different masses than 1 (and no other thermostat which
does the correct thermalization).




Am Mo., 26. Nov. 2018 um 17:47 Uhr schrieb Pascal Hebbeker <
notifications@github.com>:

src/core/reaction_ensemble.cpp
jonaslandsgesell: I think if you would add the lines 

```
double vel[3];
vel[0] = std::sqrt(temperature) * gaussian_random();
vel[1] = std::sqrt(temperature) * gaussian_random();
vel[2] = std::sqrt(temperature) * gaussian_random();
set_particle_v(p_id, vel);
```
below https://github.com/espressomd/espresso/blob/314115186f79749bda8bfc3280cb37dfa03fea48/src/core/reaction_ensemble.cpp#L836

this would fix the problem #1770 (at least for the case that we have particles with mass unity). We could then postpone the case for particles with different masses than 1 (and no other thermostat which does the correct thermalization).

jonaslandsgesell (approved):

jonaslandsgesell:
This PR fixes #1770 and #2377 for particles of mass 1. For particles with different masses the Maxwell-Boltzmann distribution needs to take into account the particle mass which has to be provided to the reaction ensemble algorithm as input then. This is not yet fixed with this PR and remains an issue. I opened issue #2388 to document this.

fweik:
I don't get it, `ReactionAlgorithm::create_particle` only creates particles of mass 1, no?

jonaslandsgesell:
Yes but this should be changed if one wants to be as correct as possible.
One example would be if one wants to sample the (mass dependent) velocity
distribution in a reacting system. Then the masses of all types need to be
correct.

One can also construct exotic cases where masses become a problem. This is
obvious when you think about only the particle moving Monte Carlo moves
(which are also implemented in Espresso in the reaction ensemble module).
Consider a case where a user created particles with mass 0.5 but has not
assigned velocities. If there is no other thermostat running the velocities
will stay 0 if we only use MC moves to sample the configuration space. The
configuration space is sampled correctly. However the kinetic part of the
virial pressure will be 0 which is not how a termalized system would
behave. In order to solve this problem the correct velocity distribution
(which is mass dependent) needs to be assigned to the particles... All in
all this is not a problem if the system becomes correctly thermalized using
e.g. a Langevin thermostat before taking a sample.

Am Mo., 26. Nov. 2018 um 18:53 Uhr schrieb Florian Weik <
notifications@github.com>:

fweik:
I mean you could just draw the velocity from the correct MBD for the particle, and this would be correct for ever right? Why can't you just put in the mass?

jonaslandsgesell:
Yes exactly. This will be the way to go. Pascal suggested to postpone this
though since it requires the Reaction Ensemble to be aware of masses (User
input).

Am Mo., 26. Nov. 2018, 19:43 hat Florian Weik <notifications@github.com>
geschrieben:

thepith:
Or perhaps to rephrase myself: I currently do not have the capacity to do it. If @jonaslandsgesell wants to do it, we can incorporate it in this pull request.

Maintainers are allowed to edit, so I believe Jonas can push to my branch...

src/core/reaction_ensemble.cpp
```diff
@@ -831,6 +831,12 @@ bool ReactionAlgorithm::do_global_mc_move_for_particles_of_type(
     p_id = p_id_s_changed_particles[i];
     // change particle position
     new_pos = get_random_position_in_box();
+    double vel[3];
```
jonaslandsgesell, line 834: Can you please add:
auto const &p = get_particle_data(p_id);
vel[0] = std::sqrt(temperature/p.p.mass) * gaussian_random();
vel[1] = std::sqrt(temperature/p.p.mass) * gaussian_random();
vel[2] = std::sqrt(temperature/p.p.mass) * gaussian_random();
etc in this place? This way the particle moving MC moves give correctly thermalized particle velocities.

Unfortunately I cannot edit this on my own.

thepith: Can we use the same to get the charges?

Can we use the same at the position where we set the particle velocities when we create a particle?

jonaslandsgesell: Here we can ask for the particle mass, since it was already created by the user beforehand (with the correct mass). In the other location we cannot ask for the particle mass or charge, since the particle is about to be created with a certain user provided mass or charge. Espresso cannot know what charge or mass should be assigned if a new particle is to be created (since this is physical input).

thepith: So I incorporated the changes. Is there more to do, or is this pull request ready to be merged?

src/python/espressomd/reaction_ensemble.pyx
```diff
@@ -16,6 +16,10 @@ cdef class ReactionAlgorithm(object):
     reaction algorithm by setting the standard pressure, temperature, and the
     exclusion radius.
 
+    Note: When creating particles the velocities the new particles are set
```
KaiSzuttor: *of* the new particles[...]

src/python/espressomd/reaction_ensemble.pyx
```diff
@@ -16,6 +16,10 @@ cdef class ReactionAlgorithm(object):
...
+    according the Maxwell distribution. In this step the mass of the new particle
```
KaiSzuttor: Maxwell-Boltzmann

KaiSzuttor (changes requested):

PR #2379, opened 2018-11-20
Fix Simple Pore

lahnerml:
Fixes simple pore shape, in particular distance calculation and if a point is inside or outside the shape.


RudolfWeeber:
@fweik?

fweik:
This needs a test.

fweik (approved):

PR #2296, opened 2018-10-07
Refactor and benchmark Particle

jngrad:
This PR is not meant to be merged (yet).

Following our exchange on #2239, I've developed a benchmarking suite to measure the integration time of pre-equilibrated particle simulations. The suite currently provides scripts for a LJ gas, a LJ liquid, a P3M saline solution and a P3M ionic gas, and runs them using 1/2/4/8/16 cores, 1k/10k particles per core, and 3 versions of `myconfig.hpp`.

The benchmarks are activated in CMake with `-DWITH_BENCHMARKS=ON`. To add your own benchmarks, use `/src/maintainer/benchmarks/lj.py` as a template and add a new line in `/src/maintainer/benchmarks/CMakeLists.txt`. Pass option `--visualize` to `lj.py` to start OpenGL and visualize the simulation.

Execute `/maintainer/benchmarks/suite.sh` to run the benchmarks. The script loops over various `myconfig.hpp` files and measures the impact of including more features in the `struct Particle`. The script also loops over commits in the git history to measure the performance of various implementations of `struct Particle`. The results are stored in `benchmarks.csv`, where the relevant fields are:

* `commit`: git commit (currently testing the original `struct Particle` implementation, a raw pointer implementation and a unique_ptr implementation)
* `config`: version of `myconfig.hpp` (currently testing minimal, default and maxset)
* `script`: benchmark file (currently running `lj.py` and `p3m.py`)
* `arguments`: space-separated list of arguments passed to the benchmark
* `cores`: MPI cores (1 if `mpiexec` was not used)
* `MPI`: `True` if `mpiexec` was used, `False` otherwise
* `mean`: mean execution time of a single integration step (seconds)
* `ci`: 95% confidence interval for the mean

Debug fields:

* `steps_per_tick`: sample size of the mean
* `duration`: duration of the simulation (seconds), should be 1-2 min long
* `E1`, `E2`, `E3`: energy values from the final state of the simulations (to check the benchmark is reproducible)

The benchmark takes 20 hours on my workstation, you can get the raw results [here](https://github.com/jngrad/espresso/blob/4b8ac7d773cf8dc9387e3dc467b35d24fee9a348/data/benchmarks-workstation.csv) (I'll run it on bee next week). The pointer implementation of `struct Particle` runs 23% slower than the original implementation, for both raw and smart pointers. Using the original implementation and default features as a baseline, the increase in integration time can be broken down as follows:

| implementation  | minimal | default | maxset |
|-----------------|---------|---------|--------|
[...]

Issue #2239, opened 2018-09-07
Refactor Particle struct


jngrad:
Objective: reduce memory footprint of particle lists during `short_range_loop()` execution
- [ ] **re-shuffle Particle struct into dedicated structs or classes**
  + group members relevant to `short_range_loop()` in the same struct
  + refactor all functions processing Particle structs (member names should already be unique)
- [x] **write benchmarks for different projects**: magnetic fluids, ...
- [x] harmonize Python and C++ interface
  +  use same names for particle properties in C++ and Python?
  + be able to change the particle properties directly in Python, if possible
- [x] make it easy to add new particle properties in the relevant structs

Possible implementation:
```c++
struct Particle {
  ParticleProperties* p; /* data that has to be known for all ghosts, doesn't change */
  ParticleLocal* l; /* data needed only on the node the particle belongs to */
  ParticlePosition* r; /* positional data */
  ParticleMomentum* m;
  ParticleForce* f;
};
```
Rudolf's proposition to move members to dedicated structs:
- ParticleProperties:
  * Shan-Chen -> ParticleLocal?
  * affinity: probably part of object in fluid. Can it go to ParticleLocal?
  * membrane collision -> ParticleLocal?
  * rotation -> ParticleLocal?
  * electrohydrodynamic mu_e -> ParticleLocal
  * ext_force -> ParticleLocal
- ParticlePosition:
  * remove quatu (director)
  * remove dip
  * bond constraint: p_old -> ParticleLocal
  * Shan-Chen fluid composition -> ParticleLocal
- Particle:
  * can/should the bond and exclusion lists go to ParticleProperties?
  * swimming properties -> ParticleProperties or ParticleLocal

hmenke:
I think that's a bad idea because older parts of Espresso might rely on the exact order of the data members (for strided `memcpy` an such).

fweik:
@hmenke I don't think so. Do you have something specific in mind? Also such code should be fixed/removed anyhow...

pdebuyl:
The only occurence of memcpy in the current python branch is to actually error out on the use of memcpy for the particles.

https://github.com/espressomd/espresso/blob/python/src/core/particle_data.hpp#L412

hmenke:
@fweik That is the case for swimmers in LB GPU.  It's also the most efficient way to perform strided `memcpy`.

[...]

PR #2381, opened 2018-11-22
Particle benchmarks

jngrad:
LJ and P3M benchmarks discussed in #2239 and extracted from #2296.


RudolfWeeber:
In the lj script, you're setting a force cap (20) and never remove it.
You don't need a force cap for steepest descent-based warmups.
Please remove the capping.

Please switch the command line handling to argparse in both test scripts.
I think the simplest solution would be to have
--volume_fraction and --particles_per_core 
and for p3m --bjerrum_length


Please also simplify the p3m script:
* please make use of reduced units 
  q=+1 and q=-1 on half of the particles each
  kT=gamma=1
  p3m accuracy=1E-4
  prefactor to give a Bjerrum length provided by the cmd line

Typical Bjerrum lengths would be on the order of ~4 sigma


The runner script should be written to time the currently checked out version.
Comparisson should than be handled by a separate script which uses the timing script twice.

jngrad:
Good idea. `suite.sh` is now responsible for checking out older git revisions and calling `runner.sh`, which runs the benchmarks on the currently checked out version. I've also rewritten the benchmarking CMake file with CTest to avoid breaking the workflow when a benchmark fails. The LJ and P3M scripts were simplified.

Could you please have a look at the output of the LJ script? For volume fractions below 0.5 without force cap, the script sometimes crashes with this error message:
```
ERROR: particle coordinate out of range, pos = 2.22303e+13, image box = -2147483648
ERROR: particle coordinate out of range, pos = -2.19555e+13, image box = -2147483646
  File "lj.py", line 138, in <module>
    system.integrator.run(min(30 * measurement_steps, 60000))
  File "integrate.pyx", line 91, in espressomd.integrate.Integrator.run
  File "utils.pyx", line 269, in espressomd.utils.handle_errors
Exception: Encoutered errors during integrate: ERROR: particle coordinate out of range, pos = 2.22303e+13, image box = -2147483648
```
You may have to test different seed values at line 88 in lj.py until the error appears (cmd: `pypresso lj.py --particles_per_core 10000 --volume_fraction 0.02`). This also happens with your original script.

RudolfWeeber:
Please add calls to tune_skin before and after p3m setup. The optimal skin is density dependent.
Also, please use particle type 0 for both positively and negatively charged particles, since having more particle types has a performance impact.

fweik:
@RudolfWeeber

RudolfWeeber:
Can be merged. 
Can we somehow schedule a CI job for this once a week at night?
It would have to occupy an entire build machine to be consistent.

jngrad:
Are you sure? The LJ benchmark still fails on my machine depending on the parameters.

KaiSzuttor:
i think we planned to buy CI machines. we need a dedicated one for such benchmarks.

mkuron:
> Can we somehow schedule a CI job for this once a week at night?

Sure.

> It would have to occupy an entire build machine to be consistent.

That's difficult. Our runners are configured for multiple concurrent jobs. We would need to do something hacky that reconfigures the runner for a single job, via a cron job or something.

> I think we planned to buy CI machines.

That depends on whether the university administration gives us the money for it.

> we need a dedicated one for such benchmarks.

Not necessarily, we just need one that doesn't do other stuff at the same time.

RudolfWeeber:
> Are you sure? The LJ benchmark still fails on my machine depending on the parameters.
What parameters precisely. I ran both of them, but did not vary parameters extensively.

jngrad:
Try `mpiexec -n 8 pypresso maintainer/benchmarks/lj.py --particles_per_core=10000 --volume_fraction=0.02`

RudolfWeeber:
I can't reproduce this. Can you please post the error msg?

jngrad:
```
ERROR: particle coordinate out of range, pos = 2.22303e+13, image box = -2147483648
ERROR: particle coordinate out of range, pos = -2.19555e+13, image box = -2147483646
  File "lj.py", line 138, in <module>
    system.integrator.run(min(30 * measurement_steps, 60000))
  File "integrate.pyx", line 91, in espressomd.integrate.Integrator.run
  File "utils.pyx", line 269, in espressomd.utils.handle_errors
Exception: Encoutered errors during integrate: ERROR: particle coordinate out of range, pos = 2.22303e+13, image box = -2147483648
```

RudolfWeeber:
I think I fixed the remaining issues. I'll merge this for now, so we can make use of it.

Issue #2386, opened 2018-11-26
CI failures due to LB and timeout


jngrad:
Multiple PR are currently failing CI due to an LB thermostat error unrelated to the committed changes:
   * #2385, [CI](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/pipelines/3963): added a Markdown file
   * #2381, [CI](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/pipelines/3940): added an optional test, which is disabled by default
   * #2379, [CI](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/pipelines/3913): fixed a Shape test
   * #2378, [CI](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/pipelines/3911): simplified a function in reaction ensemble

Likely related to #2247 and #2226.

There is also an issue with the new rocm-maxset container whose 20 min timeout is too short, see for example #2381, [CI](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/pipelines/3940) and #2379, [CI](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/pipelines/3913). Other containers have a 1 hour timeout.

mkuron:
The timeout on rocm is short on purpose because that's my desktop computer, not a dedicated machine. Normally the jobs take about 8 minutes, so the 20-minute limit should be plenty. If something runs into the timeout, it's usually a frozen test.

PR #2390, opened 2018-11-27
Only copy_v_cs_from_GPU if there are swimming particles

### LINKED ISSUE #2389: ENGINE: Un-needed data transfers
@hmenke @jdgraaf
In its current form, the engine feature copies v_cs data 
(velocities at particle center and 2nd point, I assume) 
from the gpu and distributes it via mpi, independently of whether swimming particles are there.
This takes some 20% of the execution time in an unrelated lbgpu+particle simulation.
19.97%  python         EspressoCore.so.4               [.] _ZN12_GLOBAL__N_18set_v_csEN5Utils5RangeI16ParticleIterator

we are currently preparing to extend espresso's default myconfig to include nearly all features, so we can ship pre-built packages for common Linux distros.

Unless above issue is resolved, we cannot include the ENGINE feature in this default config.
That would be a pity, because there is a tutorial for this feature.

hmenke:
Fixes #2389

Description of changes:
 - Add a global variable that indicates that there are swimming particles

### DIFF
--- a/src/core/particle_data.cpp
+++ b/src/core/particle_data.cpp
@@ -76,6 +76,7 @@ void add_id_to_type_map(int part_id, int type);
 
 int max_seen_particle = -1;
 int n_part = 0;
+bool swimming_particles_exist = false;
 /**
  * @brief id -> rank
  */
--- a/src/core/particle_data.hpp
+++ b/src/core/particle_data.hpp
@@ -479,6 +479,8 @@ struct ParticleList {
 extern int max_seen_particle;
 /** total number of particles on all nodes. */
 extern int n_part;
+/** flag that active swimming particles exist */
+extern bool swimming_particles_exist;
 
 /** id->particle mapping on all nodes. This is used to find partners
     of bonded interactions. */
--- a/src/core/rotation.cpp
+++ b/src/core/rotation.cpp
@@ -294,7 +294,7 @@ void convert_torques_propagate_omega() {
       fprintf(stderr, "%d: convert_torques_propagate_omega:\n", this_node));
 
 #if defined(LB_GPU) && defined(ENGINE)
-  if (lattice_switch & LATTICE_LB_GPU) {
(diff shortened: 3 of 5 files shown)


hmenke:
GitLab CI failed for `maxset` and `osx` in the `lb_thermostat` test (unsure whether this is related, it shouldn't be)
```
FAIL: test_velocity_distribution (__main__.LBCPUThermostat)
  File "/Users/testaccounticp/builds/espressomd/espresso/build/testsuite/python/lb_thermostat.py", line 77, in test_velocity_distribution
    self.assertLessEqual(abs(found - expected), error_tol)
AssertionError: 0.010366421182996383 not less than or equal to 0.01
```
The `rocm-maxset` job didn't start and failed with “No such container”.

mkuron (approved):

mkuron:
Thermostat: unrelated, see #2386. rocm: job is being retried right now.

src/core/particle_data.cpp
```diff
@@ -76,6 +76,7 @@ void add_id_to_type_map(int part_id, int type);
 
 int max_seen_particle = -1;
 int n_part = 0;
+bool has_swimming_particles = false;
```
KaiSzuttor: What "has" swimming particles? Maybe I'm pedantic, but a name like ```swimming_particles_exist``` would make more sense in my opinion.

KaiSzuttor (approved):

KaiSzuttor:
@RudolfWeeber

RudolfWeeber:
Milestoning this for 4.1, since it only affects performance in custom myconfigs including  but not using ENGINE in 4.0.

RudolfWeeber:
Please make activating the switch conditional on parameters that actually lead to swimming.
The current implementation will probably trigger even when a checkpoint is restored which does not contain swimming particles.

Once that's done, the pr can be merged.

PR #2393, opened 2018-11-30
core: dd: Slight improvement of communication pattern. 

fweik:
Description of changes:
 - Slight improvement to the communication pattern

Due to the implementation of the async operations for non-mpi types in boost.mpi,
this is slightly better. It also looks nicer because it is symmetric.


RudolfWeeber:
Can be merged.

PR #2395, opened 2018-12-03
Improve constrainst visualization sample; directly set box_l where appropriate

pkreissl:
In `visualization_constraints.py`:
  - Variable `box_l` was defined but not used.
  - Commented out code was formatted as if descriptive comment. As a result you could **not** just comment in and run -- which is the desired behaviour here.

In the other samples, `System` was set up with some randomly chosen `box_l`, which a few lines later was just overwritten by a different `box_l`. This didn't really make sense.

### DIFF
--- a/samples/minimal-polymer.py
+++ b/samples/minimal-polymer.py
@@ -30,14 +30,13 @@ import numpy as np
 # System parameters
 #############################################################
 
-system = espressomd.System(box_l=[1.0, 1.0, 1.0])
+system = espressomd.System(box_l=[100, 100, 100])
 system.set_random_state_PRNG()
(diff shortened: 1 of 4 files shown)


jngrad:
The case of `visualization_constraints.py` was handled differently in #2452, where the sample is run with `pypresso visualization_constraints.py --<shape_name>` to select which shape to display without having to inspect commented code ([3f10a29/samples/visualization_constraints.py](https://github.com/espressomd/espresso/blob/3f10a29453534fd6098c4d20d1fa4f6d6dac7352/samples/visualization_constraints.py)). This also helps with CI, as commented code cannot be tested. Maybe this PR can be merged into #2452.

jngrad:
Merged into #2452.

PR #2397, opened 2018-12-03
Scriptinterface: Avoid cast to float

### LINKED ISSUE #2396: Loss of precision in ScriptInterface
@pdebuyl

During the introduction of the Lees Edwards boundary conditions, we discovered a serious issue with the ScriptInterface. This issue is of critical importance as it leads to ballooning errors and has widespread effects.

The "double" floating point parameters set via "set_params" are rounded to "float" with a relative accuracy of about 10e-8 / 10e-9. This is not only a problem with regard to Lees Edwards boundary conditions, but probably affects all parts of the code using the ScriptInterface. This includes basic components of the code like the MD timestep and system.time or the collision detection, as we show below (tested with current python branch).

```py
import espressomd as md
from espressomd.interactions import HarmonicBond

system = md.System(box_l=[5.0, 5.0, 5.0])
system.time_step = 0.3
system.cell_system.skin = 0.4
bond_centers = HarmonicBond(k=1000, r_0=0.0)
system.bonded_inter.add(bond_centers)
system.min_global_cut = 1.0

system.collision_detection.set_params(mode='bind_centers', distance=0.6, bond_centers=bond_centers)
print("Col Det:", system.collision_detection.distance)
```
returns:` 0.600000023842`

```py
system.integrator.run(steps=1000000)
print(system.time)
```

returns: `299999.999994`

The result is the same as explicitly casting a double to real using NumPy. For illustration:
```py
print("%20.15f" % np.array(0.6, dtype=np.float32))
```
returns: `0.600000023841858`

Exactly representable values (1, 2, 1/2, 1/4, etc) do not lose accuracy, which supports our hypothesis. Of course, a relative error of 10e-8 in the code does not impact the final outcome of a single point very much. However, this error causes rapidly increasing errors with each subsequent call, preventing consistent testing of the interface.

We understand the need to switch from globals to the ScriptInterface and were disappointed that this feature is not working as expected. The code introduced by @pdebuyl and me worked fine with global variables. We followed a very constructive and ESPResSo-conforming path for our work and updated our code to use the now-recommended ScriptInterface. However, this change resulted in a significant delay and wasted effort as we debugged our code assuming that the current ScriptInterface implementation worked properly. Our implementation of Lees-Edwards is almost complete except for this bug. 

Due to the complexity of the Cython/Boost/C++ binding, we were unable to locate the root of the problem. However, there is no obvious reason why the ScriptInterface should behave in such a way. Thus, we have no immediate view on how to solve this issue, but it is clearly of general interest for ESPResSo users and developers. This issue should be resolved as soon as possible, perhaps as early as 4.0.1, as it is not possible to rely on ScriptInterface in the current state of the code, which will hamper contributions from the userbase and lead to frustrations.

RudolfWeeber:
Sorry for the late action. Fixes #2396 
It is not clear to me, why that didn't get caught by the clang checks.

Description of changes:
 - 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?

### DIFF
--- a/src/python/espressomd/script_interface.pyx
+++ b/src/python/espressomd/script_interface.pyx
@@ -161,7 +161,7 @@ cdef class PScriptInterface(object):
         elif np.issubdtype(np.dtype(type(value)), np.signedinteger):
             return make_variant[int](value)
         elif np.issubdtype(np.dtype(type(value)), np.floating):
-            return make_variant[float](value)
+            return make_variant[double](value)
         else:
             raise TypeError("Unkown type for conversion to Variant")
 
--- a/src/script_interface/VariantTester.hpp
+++ b/src/script_interface/VariantTester.hpp
@@ -93,6 +93,10 @@ public:
         return is_vector(par.at("value"));
       }
     }
+    
+    if (method == "mirror") {
+      return par.at("value");
+    }
 
     throw std::runtime_error("Unknown method");
   }
--- a/src/script_interface/VariantTester.hpp
+++ b/src/script_interface/VariantTester.hpp
@@ -93,7 +93,7 @@ public:
         return is_vector(par.at("value"));
       }
     }
-    
+
     if (method == "mirror") {
       return par.at("value");
     }
--- a/testsuite/python/collision_detection.py
+++ b/testsuite/python/collision_detection.py
@@ -733,7 +733,7 @@ class CollisionDetection(ut.TestCase):
         res = reduce[0](reduce[1][0])
         self.assertEqual(res.__class__.__name__, "CollisionDetection")
         self.assertEqual(res.mode, "bind_centers")
-        self.assertAlmostEqual(res.distance, 0.11, delta=1E-9)
+        self.assertAlmostEqual(res.distance, 0.11, delta=1E-12)
         self.assertEqual(res.bond_centers, self.H)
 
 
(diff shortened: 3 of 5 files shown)


mkuron:
It didn't get caught by the clang check because we disabled all conversion warnings. Espresso is full of int-to-double and double-to-int conversions.

RudolfWeeber:
Why did 
        self.check_type_and_value(float, 3.14159, ret[2])
in the variant conversion test not trigger?

fweik:
Can you please find out why this did not trigger `testsuite/python/variant_conversion.py:59`, and adapt it accordingly?

RudolfWeeber:
Because only passing doubles from core to Python was tested, which was not affected by the bug.
I pushed a parameter mirror test, which fails when the narrowing is re-introduced

fweik (approved):

pdebuyl:
Hi @RudolfWeeber thank you for fixing this. I had looked at the C++ script interface code and couldn't find the issue.

Will this go in 4.0.1?

@mkuron any reason why ESPResSo uses int-to-double and double-to-int conversions?

mkuron:
There are lots of places where integer constants are given as integers and stuff like that. You would need to add a cast around every single one of them.

RudolfWeeber:
> Hi @RudolfWeeber thank you for fixing this. I had looked at the C++ script interface code and couldn't find the issue.
It's a bit unfortunate that in Cython `float` can mean different things depending on the context.

> 
> Will this go in 4.0.1?
Yes.

RudolfWeeber:
> There are lots of places where integer constants are given as integers and stuff like that. You would need to add a cast around every single one of them.
At some point, we should look into  cppcoreguidelines-narrowing-conversions in clang-tidy.

PR #2400, opened 2018-12-04
WIP: Local particles

fweik:
Description of changes:
 - Replace the sparse array that was used for the local particle
   index by a hash map. 
-  Adding particles when initializing a new cell system is now handled
   by the regular particle sorting code instead of duplicated code.

This finally makes Espresso linear in memory consumption.
If the performance with the `std::unordered_map` is too bad
we could replace it by a better one e.g. from abseil, but for
the reduced complexity in bookkeeping I really would like to
get this in.


RudolfWeeber:
+  for (auto c : local_cells) {
+    for (int i = 0; i < c->n; i++) {
+      auto const &p = c->part[i];
+      check_index_ghost_part(p);
     }
   }
 }
Should that loop go over ghost cells?

RudolfWeeber:
Performance decreases by some 5% for a 1000 particle lj fluid at 0.1 volume fraction.
For 200 polymers of length 5 at 0.1 volume fraction, the slow down is some 15%.

PR #2403, opened 2018-12-05
CMake Refactor

fweik:
Description of changes:
 - Made EspressoConfig into a proper cmake target
 - Some cmake cleanup, mostly putting everything that is
    interdependent into one target
 - Needs cmake 3.4 for support of imported targets in FindCUDA and FindMPI


src/core/CMakeLists.txt
```diff
@@ -142,7 +142,36 @@ set(EspressoCore_SRC
...
+        grid_based_algorithms/lbgpu_cuda.cu
+        grid_based_algorithms/fd-electrostatics_cuda.cu
+        virtual_sites/lb_inertialess_tracers_cuda.cu
+        grid_based_algorithms/electrokinetics.cpp
+        grid_based_algorithms/electrokinetics_pdb_parse.cpp
+        grid_based_algorithms/lbgpu.cpp
+    )
+
+  set(CUDA_LINK_LIBRARIES_KEYWORD PUBLIC)
+  cuda_add_library(EspressoCore SHARED ${EspressoCore_SRC} ${EspressoCuda_SRC})
```
mkuron: Not sure that's a good idea. If we ever want to get Espresso into the Ubuntu repo, it should only depend on CUDA at runtime, meaning that any libcuda-dependent libraries should be conditionally loaded at runtime if they are installed. I.e. the Espresso package should only optionally depend on CUDA.

fweik: The files in `EspressoCore_SRC` and `EspressoCuda_SRC` are mutually dependent, this is the only way to write this down in CMake without lying to the build system. If this really becomes a problem it has to be solved by actually disentangling the code.

CMakeLists.txt
```diff
@@ -38,27 +38,12 @@ endif (NOT ${SOVERSION} MATCHES "^[1-9]+$")
 # Cmake modules/macros are in a subdirectory to keep this file cleaner
 set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 
-# If the version is smaller than 3.4, use custom find HDF5 module
```
KaiSzuttor, line 41: the comment says you can remove the custom hdf5 cmake module

src/CMakeLists.txt
```diff
@@ -17,30 +17,71 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.4)
```
KaiSzuttor: Does it make sense to have a cmake version requirement besides the one in the root CMakeLists.txt?

fweik:
@KaiSzuttor

fweik:
Remaining errors are because there are gpu's in the CI pool that only support compute capability 2.0, which is not build. It's not clear to me why this worked before. (It did not, other branches are failing for the same reason -> https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/63950).

mkuron:
Frank just added coyote7 and coyote8 this week. I have removed the `cuda` tag from these runners for now.

fweik:
Thank you, that fixed one.

CMakeLists.txt
fweik: Has been removed

src/CMakeLists.txt
KaiSzuttor: ?

fweik: Dunno, was there before. Any preference?

KaiSzuttor: I think we have the most uptodate requirement at the top level and we probably don't really want to figure out and keep up to date which cmake version is required for each subdirectory. So, I would remove it.

mkuron: Agreed. We haven't supported 2.8 even in the src directory for quite a long time.

CMakeLists.txt
```diff
@@ -38,27 +38,12 @@ endif (NOT ${SOVERSION} MATCHES "^[1-9]+$")
...
-# else, use prefer parallel feature of default module.
-if(CMAKE_VERSION VERSION_LESS 3.4)
-  set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH}
-                        ${CMAKE_CURRENT_SOURCE_DIR}/cmake/hdf5)
-else()
-  set(HDF5_PREFER_PARALLEL 1)
-endif()
+set(HDF5_PREFER_PARALLEL 1)
```
KaiSzuttor: you can tidy up by moving this line just before the line with ``find_package(HDF5 "1.8" COMPONENTS C)``

src/python/espressomd/shapes.py
```diff
@@ -157,6 +135,7 @@ class Slitpore(ScriptInterfaceHelper):
     pore_mouth : :obj:`float`
     pore_width : :obj:`float`
     upper_smoothing_radius : :obj:`float`
+    dividing_plane : :obj:`float`
```
KaiSzuttor, line 138: have you added documentation for the new parameter?

KaiSzuttor (approved):

src/python/espressomd/shapes.py
fweik: I have now :-)

mkuron:
Try removing the `-Wl,--no-as-needed` flag, which might not be necessary any more and was an ugly hack.

PR #2407, opened 2018-12-07
Declarative style for particle update comm

fweik:
and reduce duplication.


fweik:
@KaiSzuttor the AMD issue is unrelated.

mkuron:
AMD fixed in #2412.

RudolfWeeber:
Looks good. A few doc comments would be helpful, though.

```diff
diff --git a/src/core/particle_data.cpp b/src/core/particle_data.cpp
+template <typename S, S Particle::*s, typename T, T S::*m>
+struct UpdateParticle {
+  int id;
+  T value;
+
+  void operator()(Particle &p) const { (p.*s).*m = value; }
+
+  template <class Archive> void serialize(Archive &ar, long int) {
+    ar &id;
+    ar &value;
+  }
+};

+
+template <typename T, T ParticleProperties::*m>
+using UpdateProperty = UpdateParticle<ParticleProperties, &Particle::p, T, m>;
+template <typename T, T ParticlePosition ::*m>
+using UpdatePosition = UpdateParticle<ParticlePosition, &Particle::r, T, m>;
+template <typename T, T ParticleMomentum ::*m>
+using UpdateMomentum = UpdateParticle<ParticleMomentum, &Particle::m, T, m>;
+template <typename T, T ParticleForce ::*m>
+using UpdateForce = UpdateParticle<ParticleForce, &Particle::f, T, m>;
```
Please insert some docs, also with regards to the template params
```diff
+template <typename S, S Particle::*s, typename T, T S::*m> struct message_type;
```
Please also insert some comments here about how the message type distinciotn works.
```diff
+

+template <typename S, S Particle::*s, typename T, T S::*m>
+using message_type_t = typename message_type<S, s, T, m>::type;
+
+struct UpdateParticleVisitor : public boost::static_visitor<void> {
+  template <typename Message> void operator()(const Message &msg) const {
+    assert(local_particles[msg.id]);
+    msg(*local_particles[msg.id]);
+  }
+};
```
Please mention here via what mechanism the actual update gets called since not all readers of this might be familiar with the pattern.

KaiSzuttor:
collision detection seems to have issues

fweik:
Yeah, I called the master instead of the local interface...

KaiSzuttor:
please merge the current python branch

fweik:
@KaiSzuttor

src/python/espressomd/particle_data.pyx
```diff
@@ -368,8 +366,7 @@ cdef class ParticleHandle(object):
             IF MASS == 1:
                 check_type_or_throw_except(
                     _mass, 1, float, "Mass has to be 1 floats")
-                if set_particle_mass(self._id, _mass) == 1:
-                    raise Exception("Set particle position first.")
+                set_particle_mass(self._id, _mass)
```
KaiSzuttor, line 365: why did you remove the exception here?

fweik: Because I removed the return value from the setter, and forgot it for all the other particles.
I *think* this can only fail if the particle does not actually exist (module some internal errors
that should be handled by asserts), in which case I would throw from the core. The error
message is also misleading.

fweik: So I guess I'm going to remove the error codes from all the functions? Almost all of them have a hard-coded return value anyhow.

fweik:
@KaiSzuttor I removed the unused return values and cleaned up a bit.

KaiSzuttor (approved):

PR #2416, opened 2018-12-13
Static analysis cleanup

jngrad:
Using maxset, before/after:

 * Clang scan-build: from 99 warnings to 54 (edge cases, MPI-related, Cython-related)
 * gcc: from 65 warnings to 0 (`-Wall -Wextra -Wno-unused-parameter -Wno-implicit-fallthrough`)
 * gcc: from 2485 warnings to 2275 (`-Wall -Wextra`)
 * Doxygen: from 118 warnings to 25

The remaining 54 scan-build warnings are either edge cases (dereference of a null pointer carried through 20 or more if/else branches) or related to Boost::MPI and Cython. The majority of the remaining 2275 gcc warnings are `-Wunused-parameter` warnings from a small group of functions related to energies and forces in `/src/core/nonbonded_interactions/` and `/src/core/bonded_interactions/`. They share a similar signature `inline int function(Particle *p1, Particle *p2, IA_parameters *ia_params, double d[3], double dist, double force[3])`, probably to help copy-pasting, although most of the parameters are not used. Changing all signatures to their respective bare minima is doable but probably not necessary.

Method used:

```sh
cd "$(git rev-parse --show-toplevel)"
mkdir -p build-scan
cd build-scan
# enable most features
cp ../maintainer/configs/nocheck-maxset.hpp myconfig.hpp
cat >> myconfig.hpp <<EOF
#define ENGINE
#define MEMBRANE_COLLISION
#define COLLISION_DETECTION
EOF
# scan-build
rm -rf src/ testsuite/ CMake* Makefile cmake_install.cmake
scan-build cmake -DCMAKE_BUILD_TYPE=Debug -DWITH_CUDA=OFF ..
scan-build -k -o static-analysis/ make -j$(nproc)
scan-build -k -o static-analysis/ make -j$(nproc) check_unit_tests
# gcc
rm -rf src/ testsuite/ CMake* Makefile cmake_install.cmake
cmake -DCMAKE_BUILD_TYPE=Debug -DWITH_CUDA=ON \
  -DCMAKE_CXX_FLAGS="-Wextra -Winline -Wpedantic" ..
cp ../maintainer/configs/nocheck-maxset.hpp myconfig.hpp
make -j$(nproc) 2>&1 | tee compilation-1-original-long.log
make -j$(nproc) check_unit_tests 2>&1 | tee -a compilation-1-original-long.log
# gcc
rm -rf src/ testsuite/ CMake* Makefile cmake_install.cmake
cmake -DCMAKE_BUILD_TYPE=Debug -DWITH_CUDA=ON \
  -DCMAKE_CXX_FLAGS="-Wextra -Wpedantic -Wno-unused-parameter -Wno-implicit-fallthrough" ..
cp ../maintainer/configs/nocheck-maxset.hpp myconfig.hpp
make -j$(nproc) 2>&1 | tee compilation-1-original-short.log
make -j$(nproc) check_unit_tests 2>&1 | tee -a compilation-1-original-short.log
# Doxygen
rm -rf doc/
cmake -DCMAKE_BUILD_TYPE=Debug ..
make doxygen 2>&1 | tee doxygen-1-original.log
```

--------------------------------------------------------------------------------

While cleaning the code I found some unclear patterns that may need fixing:

+ In `/src/core/grid_based_algorithms/lb.cpp`, function `lb_lbfluid_save_checkpoint()` declares a `host_checkpoint_philox_counter` variable and passes by pointer to `void lb_save_checkpoint_GPU()` from `/src/core/grid_based_algorithms/lbgpu_cuda.cu`, which does not change its value. The variable is then used un-initialized in the rest of the program.
+ In `/src/core/electrostatics_magnetostatics/maggs.cpp`, the body of the function `void maggs_calc_interpolated_self_force()` was commented out. Its call in `void maggs_calc_forces()` was also commented out.
+ What is the status of `/src/core/observables/not_yet_implemented/`? Most files were deleted during #1384


mkuron:
Please enable those warnings in CMakeLists.txt so that we don‘t accidentally reintroduce them in the future.

Also, merging this needs to wait until 4.0.1 is released because it will make cherry-picking later commits difficult.

jngrad:
It is not meant for 4.0.1, I should have mentioned it. Reviewing this PR will take some time considering the diversity of warnings that were addressed.

The `CMakeLists.txt` now has the -Wextra flag and its suppressions.

KaiSzuttor:
What @mkuron meant was that if we merge your PR before the 4.0.1 release and want to also fix sth. after that it will be hard to cherry pick from the python branch

fweik:
@KaiSzuttor could you please have a look at the cmake changes? I think there is possibly a more modern way to manage the flags.

fweik:
I looked at the non-comment changes and they look good to me. Most just removed the name from an unused parameters or similar.

KaiSzuttor:
@fweik one could use ``set_target_properties``, but I think it would be a mess with the current cmake style all over the core. That may be possible after merging #2403

fweik:
I agree, let's remove #2403 first, and then cleanup the defines. That PR greatly reduces the number of targets which makes changes easier.

PR #2419, opened 2018-12-17
python: set exclusions after all particles are set.

### LINKED ISSUE #2418: Particle exclusions cannot be checkpointed
see mailing list:

> Hi everybody!
> 
> I wanted to start  using the checkpointing function in Espresso 4.0 and followed closely the documentation.
> I can checkpoint everything without an error, registering quantities with 
> 
> checkpoint.register("system.non_bonded_inter")
> checkpoint.register("system.bonded_inter")
> checkpoint.register("system.thermostat")
> checkpoint.register("system.part")
> 
> commands and saving everything in my integration loop with the checkpoint.save( ) command works fine.
> 
> But upon calling the checkpoint-file with checkpoint.load( ), I get the following error:
> 
> Traceback (most recent call last):
>   File "newcubes.py", line 135, in <module>
>     checkpoint.load()
>   File "/home/lv70488/a1306705/espresso-4-0/build/src/python/espressomd/checkpointing.py", line 232, in load
>     checkpoint_data = pickle.load(f)
>   File "particle_data.pyx", line 1819, in espressomd.particle_data.ParticleList.__setstate__
>   File "particle_data.pyx", line 1918, in espressomd.particle_data.ParticleList._place_new_particle
>   File "particle_data.pyx", line 1593, in espressomd.particle_data.ParticleHandle.update
>   File "particle_data.pyx", line 1149, in espressomd.particle_data.ParticleHandle.exclusions.__set__
> ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
> 
> Now as far as I know, this is an error-message produced by numpy if one wants to check  the truth value of an array.
> 
> Is this a bug in the code or did I not pass an argument correctly?
> 
> Thanks for the help!
> Martin

KaiSzuttor:
Fixes #2418

Description of changes:
 - the particle exclusions are stored on all participating particles, which leads to an error if one of the
    exclusion partners does not exist while setting the exclusion for a particle. Therefore the exclusion
    partners have to be set after setting up all particles

### DIFF
--- a/testsuite/python/tests_common.py
+++ b/testsuite/python/tests_common.py
@@ -595,5 +595,6 @@ def single_component_maxwell(x1, x2, kT):
     x = np.linspace(x1, x2, 1000)
     return np.trapz(np.exp(-x**2 / (2. * kT)), x) / np.sqrt(2. * np.pi * kT)
 
+
 def lists_contain_same_elements(list1, list2):
     return len(list1) == len(list2) and sorted(list1) == sorted(list2)
--- a/testsuite/python/tests_common.py
+++ b/testsuite/python/tests_common.py
@@ -139,8 +139,8 @@ def lj_force_vector(v_d, d, lj_params):
 
 
 def verify_lj_forces(system, tolerance, ids_to_skip=[]):
-    """Goes over all pairs of paritcles in system and compares the forces on them
-       to what would be expected based on the systems lj parametes.
+    """Goes over all pairs of particles in system and compares the forces on them
+       to what would be expected based on the systems LJ parametes.
        Particle ids listed in ids_to_skip are not checked
        Do not run this with a thermostat enabled."""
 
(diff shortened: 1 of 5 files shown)


testsuite/python/tests_common.py
```diff
@@ -594,3 +594,7 @@ def single_component_maxwell(x1, x2, kT):
     """Integrate the probability density from x1 to x2 using the trapezoidal rule"""
     x = np.linspace(x1, x2, 1000)
     return np.trapz(np.exp(-x**2 / (2. * kT)), x) / np.sqrt(2. * np.pi * kT)
+
+
+def lists_contain_same_elements(list1, list2):
```
fweik, line 599: Isn't that just `set(list1) == set(list2)` with extra steps?

KaiSzuttor: no, because then the order does matter which is not what the function name advertises

fweik:
```python
assert set([1,2,3]) == set([3,2,1])
```

KaiSzuttor: this requires the elements to be hashable...

fweik (approved):

PR #2420, opened 2018-12-17
Added "11-ferrofluid_monolayer" to /doc/tutorials and edited CMakeLists.txt accordingly

mbschumacher:
Description of changes:
 - added a three part ferrofluid tutorial in /doc/tutorials/11-ferrofluid_monolayer/
   - part 1: introduction and structural analysis
   - part 2: influence of an external magnetic field and computing a magnetization curve
   - part 3: computing an initial susceptibility using zero field fluctuations
 - edited CMakeLists.txt in /doc/tutorials


fweik:
You did not actually check in `doc/tutorials/11-ferrofluid_monolayer/`

RudolfWeeber:
Thanks, Moritz. It reads well.

Can you please reduce the equilibration in part 1 and 2 to 30 x 1000 steps.
That should be enough for a tutorial. Please also communicate the progress in the prints via 1/30 etc.

There is a unit problem in part 2.
The Langevin parameter alpha represents the ratio of Zemann energy / thermal energy.
You are using alph=40 for the video, but the energy for 1k particles is -80k + something from the dipole dipoel interaction.
At the very least, you need to devide by the magnitude of the dipole moments when calculating the H-field you pass to the constraint.

RudolfWeeber:
Part 3:
I think, the variable dipm is never used again in this part, so it can be removed
Like in part 2, please insert an explanation with regards to the particles' dipole moment and the prefactor in the dipolar p3m.

For the fluctuation based susceptibility, please use
* lambda=2 rather than 3. 2nd order mean field theory does not hold at lambda=3 *because the partilces form chains, which are not included in the mean field picture)
  Or something comparable for which a tabulated values is published
* 10x 10k steps for the equilibration
* loops=2000 and 20 integration steps between measuring the total dipole moment

Please mention the reference to the eqn (including eqn number) at the point where you calculate chi from the measurements


Please use system.volume() rather than box_size**3

When discussing the result, please mention the limitations of mean field theory and that cluster formation has to be considered for higher densities and more strongly interacting dipoles

For the magnetizaiotn:
Please use alphas=(0.25,0.5, 1, 2, 4, 8) or something comparable to look more closely at the initial part of the magnetization curve
You can use the result from the fluctuation based meausrements for alpha=0


When setting up the magnetic field, please mention again, that the particle dipole moment was set to 1.

Please use 10x 1k steps for warmup and 500x20 steps for sampling.

'Better mode" -> second order mean field model

Here, please also mention the limitations of the mean field model.

RudolfWeeber (approved):

PR #2421, opened 2018-12-17
core: Work around another boost.mpi bug

fweik:
Older versions of boost.mpi had a broken implementation of communicator::isend for non-mpi data types. This works around that.

### DIFF
--- a/src/core/unit_tests/sendrecv_test.cpp
+++ b/src/core/unit_tests/sendrecv_test.cpp
@@ -52,7 +52,7 @@ BOOST_AUTO_TEST_CASE(blocking) {
 
   /* Non-MPI datatype */
   {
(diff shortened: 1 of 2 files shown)


mkuron:
I think we might be able to put a fixed version of `mpi::communicator::isend` into a cpp file, build that to a shared library with `-Wl,-z -Wl,interpose` and link it into Espresso. Of course, we'd only do that on Linux platforms and only if we are below Boost 1.68.0.

mkuron:
For reference, this is the patch: https://github.com/boostorg/mpi/commit/e3b0d08c7e1cf1d4df2ce348f65dd2fa59d97cb2. It was first included in Boost 1.67.0.

PR #2422, opened 2018-12-18
Fold

RudolfWeeber:
@fweik @mkuron

### DIFF
--- a/src/core/grid.hpp
+++ b/src/core/grid.hpp
@@ -271,14 +271,9 @@ template <typename T1, typename T2> void fold_position(T1 &pos, T2 &image_box) {
  */
 inline Vector3d folded_position(Particle const &p) {
   Vector3d pos{p.r.p};
-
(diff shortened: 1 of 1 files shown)


fweik:
Can you please remove the scafacos/cmake stuff? That does not work yet as expected an is unrelated.

src/core/grid.hpp
```diff
@@ -271,14 +277,9 @@ template <typename T1, typename T2> void fold_position(T1 &pos, T2 &image_box) {
  */
 inline Vector3d folded_position(Particle const &p) {
   Vector3d pos{p.r.p};
-
+  int tmp[3];
```
fweik: You need to initialize this as it is compared with `INT_MAX`, this is also what clang-tidy does not like.

fweik (changes requested):

fweik (approved):

PR #2427, opened 2019-01-03
Thermo test performance significant improvement

bogdan-tanygin:
Description of changes:
 - by means of tests' parameters, the test mass-and-rinertia_per_particle performance is improved significantly (1.45 s now compare to 24.33 s before). Shortly, main performance had been consuming before not by an actual simulation but rather by multiple integration invocations. Simply, more integration steps by the single integrator run is required for better performance;
- also, the thermalization steps before a real summation were underestimated before;
- the number of stat loops is turned out to be less important than the thermalization;
- enough steps within the loops is needed though in order to have more "Markovian" loops;
- mpi run is also enabled in mass-and-rinertia_per_particle. Minor effect here; one is probably more apparent for larger number of particles (reduced now) and more complex tests.

Not sure how was I able to miss this before :-)

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


RudolfWeeber:
This seems to have too weak statistics. The results are outside the tolerance limits on my system for 1 and 3 cores. Please make sure, the test runs at least 200x without failure, when seeding it randomly (both numpy and espresso)

bogdan-tanygin:
> This seems to have too weak statistics. The results are outside the tolerance limits on my system for 1 and 3 cores. Please make sure, the test runs at least 200x without failure, when seeding it randomly (both numpy and espresso)

@RudolfWeeber thx, I agree with this advice. Strong statistics is, indeed, important even not only out of a random testing approach but for a correct physical validation of exotic issues.

Now the test is validated by 200+ runs with OS-originated random seeds: I've left this in comments for future. It is done for 1, 2, 3, and 4 cores. Now the performance is 8.35 sec.

testsuite/python/mass-and-rinertia_per_particle.py
```diff
@@ -63,44 +64,21 @@ class ThermoTest(ut.TestCase):
 
     @classmethod
     def setUpClass(cls):
-        np.random.seed(15)
+        # Handle a random generator seeding
```
KaiSzuttor, line 67: please remove commented out code

fweik (approved):

PR #2429, opened 2019-01-06
cmake: fix GSL linking.

KaiSzuttor:
fixes build with gsl.


fweik (approved):

KaiSzuttor:
@RudolfWeeber  a similar fix should be applied to the 4.0.1 release

PR #2432, opened 2019-01-09
Clang CUDA version check

mkuron:
@KaiSzuttor


KaiSzuttor:
thanks, but I already included it in the LB PR

mkuron:
I've added two comments at https://github.com/espressomd/espresso/pull/2392#pullrequestreview-190721424

PR #2433, opened 2019-01-09
Fix tutorials and samples

jngrad:
Fix broken samples and tutorials for the 4.0.1 release
@RudolfWeeber, @KaiSzuttor

Two samples don't run in my environment:

- `/samples/lj-demo.py` crashes when using Mayavi, but works fine with OpenGL
- `/samples/wang_landau_reaction_ensemble.py` does not seem to work (output: "Histogram is flat", "Refining: Wang-Landau parameter is now 1/t.")

@KonradBreitsprecher: could you please check that `/samples/espresso_logo.py` is now correctly using your new implementation of material types?

### DIFF
--- a/doc/tutorials/02-charged_system/scripts/nacl.py
+++ b/doc/tutorials/02-charged_system/scripts/nacl.py
@@ -61,11 +61,11 @@ system.thermostat.set_langevin(kT=temp, gamma=gamma)
 
 # Place particles
 for i in range(int(n_ionpairs)):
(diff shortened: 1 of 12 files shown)


KaiSzuttor:
according to @jonaslandsgesell the wang landau sample is expected to throw at that point

KaiSzuttor:
the lj-demo works for both visualizers in my environment.

RudolfWeeber:
Can be merged

PR #2401, opened 2018-12-05
CI builds for non-amd64 architectures

mkuron:
fixes #2258 

These tests will only run once a week and for tagged releases.

Should be tagged for Espresso 4.0.1 as it is part of the release checklist. We only need the .gitlab-ci.yml changes there though as I think everything else was caused by changes since the 4.0 release.

PR #2438, opened 2019-01-14
backport support for non-amd64 architectures to 4.0.1

mkuron:
Backported from #2401

@RudolfWeeber

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -223,14 +223,12 @@ ubuntu:wo-dependencies:
   image: gitlab.icp.uni-stuttgart.de:4567/espressomd/docker/$CI_JOB_NAME
   script:
     - export with_cuda=false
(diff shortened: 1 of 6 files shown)


mkuron:
Test result:
- #2410 appears to have been incompletely cherry-picked into 4.0.1, thus leaving particle removal broken on i386. @fweik
- coordinate folding is still broken on i386 as in #2401. @RudolfWeeber @fweik
- _lb_stokes_sphere_ produces correct result but then hangs forever on arm64, ppc64le, s390x. Works fine on current master. @mkuron

mkuron:
@RudolfWeeber, please merge this pull request. Regarding the failed tests listed above: #2441 fixes the particle removal. The _lb_stokes_sphere_ test is just extremely slow, it does not actually hang. So the only thing left is the coordinate folding, which per #2440 now affects amd64 too.

PR #2440, opened 2019-01-15
Testsuite: particle position folding corner cases

RudolfWeeber:
This tests folding of 0, box_l +- epsilon with a cell grid that is not a multiple of 2.
This test passes on 64bit and crashes on 32bit on 2,3,4 cores.
The failure arises due to a difference of opinion between fold_coordinate() and dd_save_position_to_cell().

This can probably made to go away if cells accept particles sitting on their right boundary. Then, the position to cell relationship is non-unique, however.


mkuron:
Fails on 64-bit too, so now we finally have a way to reproduce this on our primary platform.

RudolfWeeber:
Has found its way into the dev branch via a different route

PR #2446, opened 2019-01-16
Speedup back assignment of forces/torques from gpu methods

RudolfWeeber:
By not using cuda pinned memory. 
In lj+lb fluids with n=1000, phi=0.1, agrid=sigma, this increases performance by a third.

### DIFF
--- a/src/core/cuda_common_cuda.cu
+++ b/src/core/cuda_common_cuda.cu
@@ -210,7 +210,7 @@ void gpu_change_number_of_part_to_comm() {
 
     // if the arrays exists free them to prevent memory leaks
     if (particle_forces_host) {
-      cuda_safe_mem(cudaFreeHost(particle_forces_host));
+      free(particle_forces_host);
       particle_forces_host = nullptr;
     }
     if (particle_data_host) {
@@ -231,13 +231,13 @@ void gpu_change_number_of_part_to_comm() {
     }
 #ifdef ENGINE
     if (host_v_cs) {
-      cudaFreeHost(host_v_cs);
+      free(host_v_cs);
       host_v_cs = nullptr;
     }
 #endif
 #if (defined DIPOLES || defined ROTATION)
     if (particle_torques_host) {
-      cudaFreeHost(particle_torques_host);
+      free(particle_torques_host);
       particle_torques_host = nullptr;
     }
 #endif
@@ -262,26 +262,21 @@ void gpu_change_number_of_part_to_comm() {
     if (global_part_vars_host.number_of_particles) {
 
       /**pinned memory mode - use special function to get OS-pinned memory*/
-      cuda_safe_mem(cudaHostAlloc((void **)&particle_data_host,
-                                  global_part_vars_host.number_of_particles *
-                                      sizeof(CUDA_particle_data),
-                                  cudaHostAllocWriteCombined));
-      cuda_safe_mem(cudaHostAlloc(
-          (void **)&particle_forces_host,
-          3 * global_part_vars_host.number_of_particles * sizeof(float),
-          cudaHostAllocWriteCombined));
+      cuda_safe_mem(
+          cudaHostAlloc((void **)&particle_data_host,
+                        global_part_vars_host.number_of_particles *
+                            sizeof(CUDA_particle_data),
+                        cudaHostAllocWriteCombined | cudaHostAllocMapped));
+      particle_forces_host = (float *)malloc(
+          3 * global_part_vars_host.number_of_particles * sizeof(float));
 
 #ifdef ENGINE
-      cuda_safe_mem(cudaHostAlloc((void **)&host_v_cs,
-                                  global_part_vars_host.number_of_particles *
-                                      sizeof(CUDA_v_cs),
-                                  cudaHostAllocWriteCombined));
+      host_v_cs = (CUDA_v_cs *)malloc(
+          global_part_vars_host.number_of_particles * sizeof(CUDA_v_cs));
 #endif
 #if (defined DIPOLES || defined ROTATION)
-      cudaHostAlloc((void **)&particle_torques_host,
-                    global_part_vars_host.number_of_particles * 3 *
-                        sizeof(float),
-                    cudaHostAllocWriteCombined);
+      particle_torques_host = (float *)malloc(
+          3 * global_part_vars_host.number_of_particles * sizeof(float));
 #endif
 
 #ifdef SHANCHEN
--- a/src/core/cuda_common_cuda.cu
+++ b/src/core/cuda_common_cuda.cu
@@ -51,13 +51,13 @@ static CUDA_fluid_composition *fluid_composition_device = nullptr;
 static CUDA_energy *energy_device = nullptr;
 
 CUDA_particle_data *particle_data_host = nullptr;
-float *particle_forces_host = nullptr;
+std::vector<float> particle_forces_host;
 CUDA_energy energy_host;
-float *particle_torques_host = nullptr;
 
+std::vector<float> particle_torques_host;
 CUDA_fluid_composition *fluid_composition_host = nullptr;
 #ifdef ENGINE
-CUDA_v_cs *host_v_cs = nullptr;
+std::vector<CUDA_v_cs> host_v_cs;
 #endif
 
 /**cuda streams for parallel computing on cpu and gpu */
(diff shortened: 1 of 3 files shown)


src/core/cuda_common_cuda.cu
```diff
@@ -266,22 +266,16 @@ void gpu_change_number_of_part_to_comm() {
                                   global_part_vars_host.number_of_particles *
                                       sizeof(CUDA_particle_data),
                                   cudaHostAllocWriteCombined));
-      cuda_safe_mem(cudaHostAlloc(
-          (void **)&particle_forces_host,
-          3 * global_part_vars_host.number_of_particles * sizeof(float),
-          cudaHostAllocWriteCombined));
+      particle_forces_host = (float *)malloc(
```
fweik: Please don't use c-style casts. Don't use raw pointers. Don't use malloc.

src/core/cuda_common_cuda.cu
```diff
@@ -266,22 +266,16 @@ void gpu_change_number_of_part_to_comm() {
...
+          3 * global_part_vars_host.number_of_particles * sizeof(float));
 
 #ifdef ENGINE
-      cuda_safe_mem(cudaHostAlloc((void **)&host_v_cs,
-                                  global_part_vars_host.number_of_particles *
-                                      sizeof(CUDA_v_cs),
-                                  cudaHostAllocWriteCombined));
+      host_v_cs = (float *)malloc(global_part_vars_host.number_of_particles *
```
fweik: Please don't use c-style casts. Don't use raw pointers. Don't use malloc. Using a vector would also have avoided the wrong type here that makes the build fail.

src/core/cuda_common_cuda.cu
```diff
@@ -266,22 +266,16 @@ void gpu_change_number_of_part_to_comm() {
...
+                                  sizeof(CUDA_v_cs));
 #endif
 #if (defined DIPOLES || defined ROTATION)
-      cudaHostAlloc((void **)&particle_torques_host,
-                    global_part_vars_host.number_of_particles * 3 *
-                        sizeof(float),
-                    cudaHostAllocWriteCombined);
+      particle_torques_host = (float *)malloc(
```
fweik: Please don't use c-style casts. Don't use raw pointers. Don't use malloc.

fweik (changes requested):

mkuron (changes requested):
Please use `free` with these variables (around line 240) instead of the previous `cudaFreeHost`.

mkuron:
Do you why these special allocations were previously used? If `cudaHostAllocDefault` or `cudaHostAllocMapped` were used, it could be used with `cudaMemcpyAsync`, but that's not the case here. The documentation says
> `cudaHostAllocWriteCombined`: Allocates the memory as write-combined (WC). WC memory can be transferred across the PCI Express bus more quickly on some system configurations, but cannot be read efficiently by most CPUs. WC memory is a good option for buffers that will be written by the CPU and read by the device via mapped pinned memory or host->device transfers.

So since these variables are used for copying data from the GPU to the CPU and then reading that data on the CPU, that option is clearly inappropriate.

mkuron:
There are two more instances of `cudaHostAlloc` in this file. The one for `fluid_composition_host` looks like it's the same situation as the one you fixed for the other three variables, while the one for `particle_data_host` should probably use `cudaHostAllocWriteCombined|cudaHostAllocMapped` (right now the memory isn't pinned and the `cudaMemcpyAsync` will be synchronous).

RudolfWeeber:
On Wed, Jan 16, 2019 at 11:20:12AM -0800, Michael Kuron wrote:
> There are two more instances of `cudaHostAlloc` in this file. The one for `fluid_composition_host` looks like it's the same situation as the one you fixed for the other three variables, while the one for `particle_data_host` should probably use `cudaHostAllocWriteCombined|cudaHostAllocMapped` (right now the memory isn't pinned and the `cudaMemcpyAsync` will be synchronous).
I'm not going to touch the ShanChen stuff because it has not tests
the device, so it probably should stay the way it is.

I matched the frees and changed the malloc type. No difference in
timing (within 5%)

For 10k particles, the difference between non-lb an lb is 4 vs 4.7ms
/ time step.
Of the 0.7ms the leading (host-blocking) gpu operation is 0.2 ms for
the blocking transfer of the forces from the gpu to the host.

fweik:
If think what @mkuron wante to say that you have to use `free` with `malloc` and `cudaFree` with `cudaMalloc`. Otherwise you are assuming cuda uses malloc to allocate memory.

RudolfWeeber:
On Wed, Jan 16, 2019 at 11:56:27PM -0800, Florian Weik wrote:
> If think what @mkuron wante to say that you have to use `free` with `malloc` and `cudaFree` with `cudaMalloc`. Otherwise you are assuming cuda uses malloc to allocate memory.
I think I did that.

fweik:
As sorry I misread. Why can't use just use a vector? I not going to merge it like this.

src/core/cuda_common_cuda.cu
```diff
@@ -262,26 +262,21 @@ void gpu_change_number_of_part_to_comm() {
...
-                                  cudaHostAllocWriteCombined));
-      cuda_safe_mem(cudaHostAlloc(
-          (void **)&particle_forces_host,
-          3 * global_part_vars_host.number_of_particles * sizeof(float),
-          cudaHostAllocWriteCombined));
+      cuda_safe_mem(
+          cudaHostAlloc((void **)&particle_data_host,
+                        global_part_vars_host.number_of_particles *
+                            sizeof(CUDA_particle_data),
+                        cudaHostAllocWriteCombined ));
```
mkuron: This should use `cudaHostAllocWriteCombined|cudaHostAllocMapped`. Otherwise the memory copy isn't actually going to be asynchronous because the memory isn't pinned. Right now the comment above that claims it is pinned is simply wrong.

mkuron (changes requested):

RudolfWeeber:
On Thu, Jan 17, 2019 at 12:21:51AM -0800, Michael Kuron wrote:
> mkuron requested changes on this pull request.
> 
> 
> 
> > @@ -262,26 +262,21 @@ void gpu_change_number_of_part_to_comm() {
>      if (global_part_vars_host.number_of_particles) {
>  
>        /**pinned memory mode - use special function to get OS-pinned memory*/
> -      cuda_safe_mem(cudaHostAlloc((void **)&particle_data_host,
> -                                  global_part_vars_host.number_of_particles *
> -                                      sizeof(CUDA_particle_data),
> -                                  cudaHostAllocWriteCombined));
> -      cuda_safe_mem(cudaHostAlloc(
> -          (void **)&particle_forces_host,
> -          3 * global_part_vars_host.number_of_particles * sizeof(float),
> -          cudaHostAllocWriteCombined));
> +      cuda_safe_mem(
> +          cudaHostAlloc((void **)&particle_data_host,
> +                        global_part_vars_host.number_of_particles *
> +                            sizeof(CUDA_particle_data),
> +                        cudaHostAllocWriteCombined ));
> 
> This should use `cudaHostAllocWriteCombined|cudaHostAllocMapped`. Otherwise the memory copy isn't actually going to be asynchronous.
I read the docs differently. Mapped memory can be accessed directly
from the gpu without an explicit copy. It is however not supported
on HIP. Adding the flag also does not impact performance.
I also tried adding cudaHostAllocPortable which also explicitly
mentions "pinned". Also no impact.
I think, the docs are unclear.

Furthermore, the profile (from nvprof) suggests that the transfer is
quite efficient (if not actually asynchroneous):
      API calls:   
33.26%  258.92ms      5006  51.722us  cudaMemcpy (back transfer device->host)
30.51%  237.48ms     15010  15.821us  cudaLaunch
13.38%  104.15ms         1  104.15ms  cudaStreamCreate
12.78%  99.460ms      5002  19.883us  cudaMemcpyAsync (particle data
host->gpu)

So, a third of the time for 3x the amount of data (3pos, 3v, 1id vs
3force).

Further performance work has to go to the back transfer of the
forces, I think.
For example, the transfer of torques (for builds with rotation) can
be ommitted if no gpu method actually uses them.

RudolfWeeber:
On Thu, Jan 17, 2019 at 12:16:14AM -0800, Florian Weik wrote:
> As sorry I misread. Why can't use just use a vector? I not going to merge it like this.
It's part of the pr now. I still find that an unrelated task,
though.

src/core/cuda_interface.hpp
```diff
@@ -180,7 +181,8 @@ void cuda_mpi_send_composition(ParticleRange, CUDA_fluid_composition *);
 
 #ifdef ENGINE
 void copy_v_cs_from_GPU(ParticleRange particles);
-void cuda_mpi_send_v_cs(ParticleRange particles, CUDA_v_cs *host_v_cs);
+void cuda_mpi_send_v_cs(ParticleRange particles,
+                        std::vector<CUDA_v_cs> host_v_cs);
```
fweik, line 185: This should probably be a reference

fweik (approved):

PR #2447, opened 2019-01-17
Particle fold corner

fweik:


mkuron:
On i386 (#2448), the _particle_ test now passes, but _collision_detection_ is still broken.

RudolfWeeber (approved):

PR #2452, opened 2019-01-19
Unit testing tutorials and samples

### LINKED ISSUE #1377: Add scripts from tutorials to testing
where appropriate and fast enough.

jngrad:
Fixes #1377

Description of changes:
- **tests for tutorials and samples**
- **fixed broken samples and tutorial scripts**
- cleaned up tutorials and samples
- converted tutorial 8 to IPython

PR Checklist
------------
 - [x] Write tests (`/testsuite/scripts`)
 - [x] Document tests (`/doc/sphinx/testing.rst`)
 - [x] Integrate tests to CI (#1377):
   - update containers with missing Python packages:
     - nbconvert, matplotlib, scipy, mock (only Python2)
     - see error log at https://travis-ci.org/jngrad/espresso/builds/481848614
   - check containers have CMake version 3.7.0 or above (for test fixtures)
     - for example in Travis-CI job logs, system_info states CMake 3.9.2 when in reality it's 3.5.1
   - maybe schedule these tests differently (12 min total runtime)


RudolfWeeber:
@jngrad, it seems that the build system does not notice, when the original python files being checked have changed. The test still fails, even if the error in the underlying file has been fixed.

jngrad:
@RudolfWeeber I couldn't reproduce your observation. When I fix a sample, the build system immediately detects the change. It has to, because the first step of `check_samples` is to erase the local copy of the sample and start with a fresh one. This is done by the `local_samples` dependency, which is guaranteed to always be out-of-date thanks to `add_custom_target()`. It's wasteful to erase and copy the `samples` folder even if its content didn't change, but at least we avoid the situation you described. `check_samples` also depends on `pypresso`, so any change in the core will trigger the build system again. Didn't try on a tutorial, but the same CMake logic is used there as well. Could you please give me more details on what you did?

jngrad:
@RudolfWeeber the diamond polymer issue is fixed. I'm currently writing down the gitlab-ci.yaml section for the upcoming `cuda-tutorials` container. We probably don't want coverage enabled, right? The tutorials call many functions returning numerical results, but most of these results aren't checked. Also, these tutorials run on a different schedule.

jngrad:
The `cuda:tutorials` container works as expected. The next steps are:
1. reviewing this PR
2. edit: adding `schedules` to the `tutorials-samples` job in `.gitlab-ci.yml`
3. merging this PR
4. scheduling the tutorials tests

.gitlab-ci.yml
```diff
@@ -168,6 +168,17 @@ cuda-maxset:
     - linux
     - cuda
 
+cuda-tutorials:
```
KaiSzuttor: why is the job named cuda-tutorials? we are not only checking cuda-tutorials but tutorials plus samples, right? Maybe just name it ```tutorials-samples```?

KaiSzuttor:
Why do you want to remove the container? You probably mean the job, but you can just add
``` yaml
  only:
    - schedules
```

to the job

jngrad:
Nevermind, didn't realize scheduled jobs were defined in the same file.

.gitlab-ci.yml
jngrad: Good idea.

testsuite/scripts/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+  include(ProcessorCount)
+  ProcessorCount(NP)
+  math(EXPR TEST_NP "${NP}/2 + 1")
+endif()
+
+set(TEST_FILE_CONFIGURED_IMPORTLIB_WRAPPER ${CMAKE_CURRENT_BINARY_DIR}/test_importlib_wrapper.py)
+configure_file(importlib_wrapper.py ${CMAKE_CURRENT_BINARY_DIR}/importlib_wrapper.py)
+configure_file(test_importlib_wrapper.py ${TEST_FILE_CONFIGURED_IMPORTLIB_WRAPPER})
+
+macro(PYTHON_SCRIPTS_TEST)
```
KaiSzuttor, line 11: we should add a ```sample``` / ```tutorial``` prefix for these tests to distinguish them from the rest of the python tests

jngrad: like so?
```
Test project /work/jgrad/espresso/build/testsuite/scripts/samples
      Start  1: importlib_wrapper
 1/59 Test  #1: importlib_wrapper ............................   Passed    0.60 sec
      Start  2: sample_save_checkpoint
 2/59 Test  #2: sample_save_checkpoint .......................   Passed    4.35 sec
      Start  3: sample_store_properties
 3/59 Test  #3: sample_store_properties ......................   Passed   10.78 sec
```
`importlib_wrapper` is neither a tutorial nor a sample, and won't appear in `check_python`

KaiSzuttor: yes, looks good

KaiSzuttor:
@jngrad could you please the master and remove the WIP in the title if you think this is done

jngrad:
@RudolfWeeber @KaiSzuttor ready for review!

jngrad:
Not sure what's up with the EK tutorial test. I had the same error on the CIP pool machine. On the surface it looked like a read/write race condition, where a file is written to disk and visible with `os.listdir('.')` but cannot be opened for reading just yet, however adding a delay between writing and reading with `time.sleep(0.5)` didn't help.

KaiSzuttor:
btw you shouldn't perform all the other checks in your job or better add your check to some existing CI job

RudolfWeeber:
I think, the samples check with the empty myconfig failed, because the load_... sample are run before the store_... samples.

@KaiSzuttor, the tutorials are run in a special docker image at present. I think, we can run the one with the default myconfig on normal pull requests (but without check_python) and the ones with empty and maxset in the nightly schedule.;

jngrad:
@RudolfWeeber should be fixed now, see last commit message

KaiSzuttor:
Why should we check the tutorials and samples with different configs? We only want to make sure that interface changes don't break them. The CI already takes too long.

KaiSzuttor:
and why do we need an extra build for these checks? we can just exchange the image of an existing job and additionally check tutorials/samples

jngrad:
We use a dedicated docker image for the tutorials and samples, and we compile espresso with a different configuration in each tutorial job, so we cannot re-use artifacts.

jngrad:
The EK tutorial test should be fixed now.

doc/sphinx/testing.rst
```diff
@@ -0,0 +1,419 @@
...
+
+
+CMake
+^^^^^
+
+Configure tests in :file:`/src/core/unit_tests/CMakeLists.txt` using the syntax
+
+.. code-block:: cmake
+
+   unit_test(NAME <mytest> SRC <mytest.c> [<../dependency1.c>, ...] [NUM_PROC <N>])
```
fweik: This is not up-to-date: the `unit_test` cmake function now supports `DEPENDS`, with which cmake targets can be added as dependencies for the test. The way you describe should not be used.

fweik (changes requested):

doc/sphinx/testing.rst
jngrad: Text updated.

RudolfWeeber:
@jngrad can you please mark the tutorials tests for default, empty and no-gpu as schedule only in gitlab-ci.yml.
Then, I think, this is ready.

jngrad:
@RudolfWeeber you may mark the PR as reviewed

fweik (approved):

PR #2456, opened 2019-01-21
Backport Particle fold corner (#2447) to 4.0.1

RudolfWeeber:
* Testsuite: particle position folding corner cases

* core: Account for rounding error for dd_save_position_to_cell

* core: logical instead of bitwise or

* core: Comment why we accept certain particles

Fixes #

Description of changes:
 - 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


RudolfWeeber:
This cannot be backported at acceptable effort, since 4.0.1 still comes with the old particle sorting code. This has several more instances of fuzzy right-hand-side boundary detection.
Going to release without it.

PR #2457, opened 2019-01-22
Linker flags

fweik:
Fixes #

Description of changes:
 - 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


mkuron:
On Mac OS, it looks like you forgot to link the Cython-generated libraries against the Python library, but on Linux it passes.

fweik:
Yes, python modules in principle need not be linked against the python lib, because python takes care of that, but the MacOS linker does not allow undefined symbols in the shared object (this is what the flag changed). But I think it does not hurt to link against libpython. Let's see.

CMakeLists.txt
```diff
@@ -214,9 +214,7 @@ find_package(PythonInterp REQUIRED)
 
 if (WITH_PYTHON)
   find_package(Cython 0.23 REQUIRED)
-  execute_process(COMMAND ${PYTHON_EXECUTABLE}
-                -c "import distutils.sysconfig as cg; print(cg.get_python_inc())"
-                OUTPUT_VARIABLE PYTHON_INCLUDE_DIRS OUTPUT_STRIP_TRAILING_WHITESPACE)
+  find_package(PythonLibs 2.7 REQUIRED)
```
mkuron: Please revert this change. The previous version is necessary so `PythonInterpreter` and `PythonLibs` actually are the same Python. CMake should in theory take care of that, but there were always corner cases where that broke. `distutils.sysconfig` is 100% reliable though.

src/python/espressomd/CMakeLists.txt
```diff
@@ -102,7 +99,9 @@ foreach(cython_file ${cython_SRC})
           set_source_files_properties(${outputpath} PROPERTIES COMPILE_FLAGS "-Wno-pedantic -Wno-unused-variable")
       endif()
       set_target_properties(${basename} PROPERTIES CXX_CLANG_TIDY "")
-      target_link_libraries(${basename} PRIVATE EspressoConfig EspressoCore EspressoScriptInterface)
+      target_link_libraries(${basename} PRIVATE EspressoConfig EspressoCore EspressoScriptInterface ${PYTHON_LIBRARIES})
+      target_include_directories(${basename} PRIVATE SYSTEM ${PYTHON_INCLUDE_DIRS} ${NUMPY_INCLUDE_DIR})
```
mkuron: Or add the `-undefined dynamic_lookup` flag here exclusively to the Python modules. It seems like that is what most Python modules do. For some reason, the official Python recommendation is to not link modules against libpython.

mkuron (changes requested):

src/python/espressomd/CMakeLists.txt
fweik: Yes, that has the advantage that you always get the lib for the python executable which you are actually running. This needs binary compatibility, so it is probably a good idea.

mkuron: distutils also uses `-undefined dynamic_lookup`, so we are in good company: https://github.com/python/cpython/blob/92b8322e7ea04b239cb1cb87b78d952f13ddfebb/configure.ac#L2553

fweik: I think it is the right thing to do for python modules.

mkuron (approved):

PR #2461, opened 2019-01-23
4.0.1 NEWS and version update

RudolfWeeber:
Please give the release notes a final check


fweik (approved):

mkuron:
Once this is merged, you should merge the 4.0.1 branch into 4.0 and delete the 4.0.1 branch. Historically we only have branches for minor releases, not patch releases.

KaiSzuttor:
it doesn't hurt to keep it, right?

mkuron:
Having both a branch and a tag called 4.0.1 is a really bad idea.

mkuron:
Now we have a branch release-4.0.1 and a tag 4.0.1. This is inconsistent with how we previously handled it. Please push the commits to the 4.0 branch and delete release-4.0.1.

RudolfWeeber:
On Fri, Jan 25, 2019 at 09:25:40AM -0800, Michael Kuron wrote:
> Now we have a branch release-4.0.1 and a tag 4.0.1. This is inconsistent with how we previously handled it. Please push the commits to the 4.0 branch and delete release-4.0.1.
I don't think, this is an issue. The 4.0.1  release contains
backported stuff that is not part of the dev branch.
As long as I get the correct version when I checkout 4.0.1 and there
is no name conflict between a branch and a tag, I think it can stay.

mkuron:
I‘m just saying it‘s inconsistent. Having a branch for a patch also defeats the point of having a branch for every minor version.

PR #2462, opened 2019-01-24
cmake: Tag gpu tests, run tests in parallel.

fweik:
This adds a new target `make_check_python_gpu` to only run tests that require a gpu. Also the correct number of processes is set, and the test are running concurrently. This currently oversubscribes the gpu,
but that does not seem to be a problem.


testsuite/python/CMakeLists.txt
```diff
@@ -15,43 +15,43 @@ if(EXISTS ${MPIEXEC})
...
+            )
   else()
     add_test(${TEST_NAME} ${CMAKE_BINARY_DIR}/pypresso ${TEST_FILE})
   endif()
 
+  set_tests_properties(${TEST_NAME} PROPERTIES PROCESSORS ${TEST_NUM_PROC})
+
+  if(${TEST_MAX_NUM_PROC} LESS 2)
+    set_tests_properties(${TEST_NAME} PROPERTIES LABELS "${TEST_LABELS};serial")
+  elseif(${TEST_MAX_NUM_PROC} LESS 2)
```
jngrad:
```suggestion
  elseif(${TEST_MAX_NUM_PROC} LESS 3)
```

fweik:
@mkuron this brings down the runtime for the rocm tests to ~5 min, so we should be fine with the current time limit.

KaiSzuttor (approved):

jngrad:
There is a risk for two (or more) gpu-tests running simultaneously on the same GPU. It won't cause any failure, but it will slow both tests down. I handled this situation elsewhere with a property:
  https://github.com/espressomd/espresso/blob/f9e413c014b30f57ad7cb2b3d24ed55ed0ef8d03/testsuite/scripts/CMakeLists.txt#L34-L36

KaiSzuttor:
@jngrad I just saw your comment, if we run into issues related to oversubscribing or a serious performance drop you can open a PR to fix that.

jngrad:
Alright!

fweik:
I am aware of this, but it was still faster with oversubscribing the gpu. This can also be easily changed if needed.

PR #2463, opened 2019-01-25
Cuda10compat

KaiSzuttor:


KaiSzuttor:
:cry:  seems like hip does not support ``__all_sync``

fweik:
This increases the cuda requirement to cuda 9. What is needed for c++ 14?

KaiSzuttor:
We need cuda9 for c++14

mkuron (dismissed):

KaiSzuttor:
@mkuron can you again tell why we only require cuda 7 but need cuda 9 for c++14?

mkuron:
This should actually be changed to 9:
https://github.com/espressomd/espresso/blob/d9f7826f0a9b99d88b8942ab8511bccea2974ed1/CMakeLists.txt#L201

Then these lines can be deleted:
https://github.com/espressomd/espresso/blob/d9f7826f0a9b99d88b8942ab8511bccea2974ed1/CMakeLists.txt#L218-L220

However, don‘t touch the following lines, which are only used when clang++ is used instead of nvcc (in this case, only Clang needs to support C++14, which it did even before it supported CUDA).
https://github.com/espressomd/espresso/blob/d9f7826f0a9b99d88b8942ab8511bccea2974ed1/CMakeLists.txt#L152-L154

RudolfWeeber:
What is the status, here?

fweik (approved):

PR #2478, opened 2019-01-29
WIP: Remove ghost shifts

fweik:
Removes the shifts on the ghosts. So far normal MD works, LB coupling still needs fixing, and the layered cell system does not work.


fweik:
@RudolfWeeber do you have an idea why the virtual sites relative is broken?

bindgens1:
Thanks a lot!

pdebuyl:
Hi @fweik 

I have trouble making sense of this full PR. Would it be possible at all to have a PR against the current python branch for which only the ghost shift changes are visible? Or is there a subset of the commits here that I could read? Thanks!

fweik:
I'm not sure that I follow, everything in this PR is related to the ghost shifts (arguably except for the removal of the centor of mass virtual sites, but you can just ignore those two filed).

pdebuyl:
Hi @fweik I did look at the PR at some point in time where it was not in sync so the changes were huge.

The current version is ok, thanks for the reply :-)

pdebuyl:
One of the failures in the tests is `testsuite/python/cellsystem.py`. With two MPI processes, I get 

>  x: array([[0, 1],
>        [0, 1],
>        [0, 1],
>        [0, 1]])

for `get_pairs_`. From my reading of the code, this means that `link_cell` goes over all cell pairs (hence one finds i,j and j,i) and the results are collected across nodes (so that any node having both i and j (one as a normal particle the other as a ghost) will return the pairs).

I don't know enough about the cell iterators internals to go further unfortunately.

PR #2480, opened 2019-01-29
Core+Test: Changing the time step should not affect particle velocities

### LINKED ISSUE #2479: Changing the time step changes particle velocities
By the ratio of old_dt/new_dt. If the time step was never set, old_dt=-1.

RudolfWeeber:
Fixes #2479


fweik (approved):

KaiSzuttor (commented):
Please add some docstring what is being tested

KaiSzuttor (approved):

PR #2483, opened 2019-01-30
Early exit from pair force kernel

RudolfWeeber:
This adds a cutoff check to the beginning of the pair force kernel. Performance improvement is 10-15% for an LJ fluid (phi=0.2 and 0.4 tested).
The performance improvement is larger when more features are compiled in.

### DIFF
--- a/src/core/forces_inline.hpp
+++ b/src/core/forces_inline.hpp
@@ -270,18 +270,18 @@ inline void add_non_bonded_pair_force(Particle *p1, Particle *p2, double d[3],
   double torque1[3] = {0., 0., 0.};
   double torque2[3] = {0., 0., 0.};
   int j;
-  
+
   // Early exit if there is no interactoin to calculate
   // The exception for MMM2d is there, because the method assumes that
   // pairs within a cell system layer but outside the cutoff are considered
-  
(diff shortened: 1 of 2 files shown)


fweik (changes requested):
I think this should be integrated into `VerletCriterion`, which then should be used also for the normal pair calculation. In the long run we also need to have a more scalable solution for this.

RudolfWeeber:
On Thu, Jan 31, 2019 at 08:19:11AM +0000, Florian Weik wrote:
> fweik requested changes on this pull request.
> 
> I think this should be integrated into `VerletCriterion`, which then should be used also for the normal pair calculation. In the long run we also need to have a more scalable solution for this.
VerletCriterion checks for cutoffs including the skin, though. It
also stores (cutoff+skin)^2.
Would we extend VerletCriterion (NonBondedCriterion) with an
interface such as
bool include_in_verlet_list()
bool consider_for_pair_kernel()
bool consider_for_short_range_kernel()
and add cached values without skin?

fweik:
Ah you are right. Then I think this is ok for now.

fweik (approved):

PR #2484, opened 2019-01-31
ci: Limit tests with non-standard architectures to scheduled pipeline…

fweik:
… runs. Because they take to long.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -239,8 +239,8 @@ rocm-maxset:
   tags:
     - docker
     - linux
-  except:
-    - /^PR-.*/
+  only:
+    - scheduled
 
 ubuntu:arm64:
   <<: *arch_definition


mkuron:
Good idea, but nobody will notice if the build fails. You need to enable email notifications or write a bot that opens an issue when a scheduled build fails.

fweik:
Yes, I agree. Email notifications are enable but they go to `espressomd@icp.uni-stuttgart.de` which nobody reads.

fweik:
We could also consider a slack channel or similar.

mkuron:
I'd prefer Mattermost over Slack as that is already built into Gitlab. However we would still need to convince people to actually use it...

fweik:
We should discuss that independent of this issue. It also turns out that @KaiSzuttor gets the email for the pipeline status.

mkuron:
If this pull request gets merged before we have someone monitoring build failures, it's almost certain that builds will be broken within a few days...

fweik:
Doesn't matter, they are all read anyway because they take too long.

RudolfWeeber (approved):

RudolfWeeber:
Since Kai gets the notifications, I think this can go ahaed. In the normal developing process, both, the long lag times and the false positives are distracting.

KaiSzuttor:
well, i get an email for every job. Therefore, I don't read the notification.

fweik:
You can change it so that it only notifies you if it fails

PR #2485, opened 2019-01-31
core: Removed local_stress_tensor

fweik:
because it is broken and untested.


KaiSzuttor (approved):

RudolfWeeber:
On Thu, Jan 31, 2019 at 01:00:57AM -0800, Florian Weik wrote:
> because it is broken and untested.
What precisely is broken?

fweik:
Segfaulted when I tested it. Also the code looks suspicious and implements it's own incomplete virial calculation. This is not worthwhile keeping.

RudolfWeeber (approved):

PR #2494, opened 2019-02-01
WIP: Separate build from checks in CI pipeline

### LINKED ISSUE #2476: Improve CI pipeline parallelism
I think throughput of the CI could be improved if the build and test a split into two stages. Currently, for pull requests, if there are enough resources the runtime is basically the time for clang-6.0 (slowest job from 'build' stage) plus 'check_with_odd_no_of_proc'. If they would be split it would be only the clang-6.0 runtime (plus some overhead for downloading the artifacts).

jngrad:
Fixes #2476

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -4,6 +4,7 @@ stages:
   - permission
   - prepare
   - build
+  - checks
   - additional_checks
   - deploy
   - result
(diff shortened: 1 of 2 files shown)


jngrad:
The VirtualBox runners currently can't upload artifacts. The [docs](https://docs.gitlab.com/runner/executors/virtualbox.html) mentions that `gitlab-runner` must be installed inside the VM to enable that feature. The CI log files suggest the VM gitlab-runner is not up-to-date with the main gitlab-runner; in [this thread](https://gitlab.com/gitlab-org/gitlab-runner/issues/3752#note_116854576) the same error message was fixed with a VM update.

mkuron:
The runner is not currently installed in the macOS VMs. We should probably update them some time anyway, so I have put that on my to-do list for the next VM update.

However, this pull request is a needlessly complicated solution for #2476: you could achieve the same reduction in pipeline time by moving check_with_odd_no_of_proc to the build stage. This has the advantage of not burdening our GitLab server with all the build artifacts.

mkuron:
Also, your present solution does not run any tests on containers like CentOS and Debian. It should definitely run the particle test (which only takes a few seconds) to ensure that MPI is working and all libraries were linked correctly. That is why we always run that test even if testing is disabled.

jngrad:
Closed in favor of #2501

Issue #2145, opened 2018-08-09
Test coverage for tabulated angle potentials missing


RudolfWeeber:

jngrad:
There is some refactoring going on in #2497, with new integration tests in 5c50bd6 covering the tabulated angle potential, among others.

The tabulated dihedral force appears to be zero on all particles.

Bonded angle calculations seem based on the external angle convention. It can be a source of confusion with tabulated angles if the user forgets to flip the energy and force arrays. In addition, the tabulated force drops rapidly when the angle is almost equal to 0 or pi within 5 decimal places. Here is a minimal example:

```python
import espressomd
import numpy as np
system = espressomd.System(box_l=[10.0, 10.0, 10.0])
system.cell_system.skin = 0.4
system.time_step = .1
system.part.add(pos=[1,0,0])
system.part.add(pos=[0,0,0])
system.part.add(pos=[0,1,0])

# create a tabulated angle interaction, not flipping the arrays on purpose
tab_energy = [0.5 * (phi - np.pi / 3)**2 for phi in np.linspace(0, np.pi, 180)]
tab_force = 10 * np.abs(tab_energy)
angle_tabulated = espressomd.interactions.Tabulated(
    type='angle', energy=tab_energy, force=tab_force, min=0., max=np.pi)
system.bonded_inter.add(angle_tabulated)
system.part[1].add_bond((angle_tabulated, 0, 2))

epsilon = 0
#epsilon = 1e-4 # uncomment this line to erase the drop in force
a = []
for i, theta in enumerate(np.linspace(0 + epsilon, np.pi - epsilon, 180)):
    system.part[2].pos = [np.cos(theta), np.sin(theta), 0]
    system.integrator.run(recalc_forces=True, steps=0)
    E_sim = system.analysis.energy()["bonded"]
    F_sim = np.linalg.norm(system.part[0].f)
    j = -i - 1  # here use j = i  if the arrays were flipped
    E_ref = tab_energy[j]
    F_ref = tab_force[j]
    a.append((theta, E_sim, E_ref, F_sim, F_ref))

a = np.array(a)

from matplotlib import pyplot as plt
# plot the force
plt.plot(a[:,0], a[:,3], 'g')  # simulated force
plt.plot(a[:,0], a[:,4], 'b')  # expected force
plt.show()
# plot the energy
plt.plot(a[:,0], a[:,1], 'g')  # simulated energy
plt.plot(a[:,0], a[:,2], 'b')  # expected energy
[...]

PR #2497, opened 2019-02-04
Cleanup bonded interactions

### LINKED ISSUE #2445: Angle potentials have code duplication
angle harmonic, angle cossine and angle cossine2 are almost identical, yet have all the code duplicated.

jngrad:
Fixes #2445

Description of changes:
 - remove legacy features BOND_ANGLE_OLD, BONDED_IA_ANGLE_OLD
 - refactor duplicated code in angle potentials
 - update documentation

Got no significant difference in simulation time for a 200-beads polymer with AngleCosine.


src/core/bonded_interactions/angle_common.hpp
```diff
@@ -0,0 +1,104 @@
...
+ *  Common code for functions calculating angle forces.
+ */
+
+#include "grid.hpp"
+#include "utils.hpp"
+
+/** Compute the normalized vector between two points.
+ *  @param[in]  r_mid     Position of the middle particle
+ *  @param[in]  r_out     Position of the outer particle
+ *  @param[out] vec       Normalized vector
```
KaiSzuttor: Probably it's a good idea to remove the output parameters and instead return a tuple.

src/core/bonded_interactions/angle_common.hpp
```diff
@@ -0,0 +1,104 @@
...
+ *  @param[out] force1    Force on particle 1
+ *  @param[out] force2    Force on particle 2
+ *  @param[in]  vec1      Normalized vector for particle 1
+ *  @param[in]  vec2      Normalized vector for particle 2
+ *  @param[in]  d1i       Normalization constant for @p vec1
+ *  @param[in]  d2i       Normalization constant for @p vec2
+ *  @param[in]  cosine    Cosine of the angle
+ *  @param[in]  fac       Angle bending constant
+ */
+inline void calc_angle_force(double force1[3], double force2[3],
```
KaiSzuttor: please use ``Vector3d``

src/core/bonded_interactions/angle_cosine.hpp
```diff
@@ -31,46 +31,38 @@
...
+/** Computes the three-body angle interaction force.
+ *  @param[in]  p_mid     Second/middle particle.
+ *  @param[in]  p_left    First/left particle.
+ *  @param[in]  p_right   Third/right particle.
+ *  @param[in]  iaparams  Bonded parameters for the angle interaction.
+ *  @param[out] force1    Force on particle 1.
+ *  @param[out] force2    Force on particle 2.
+ *  @retval 0
+ */
+inline int calc_angle_cosine_force(Particle const *p_mid,
```
KaiSzuttor, line 55: please also use ``Vector3d`` and return instead of using output variables.

KaiSzuttor (changes requested):

src/core/bonded_interactions/angle_common.hpp
```diff
@@ -0,0 +1,104 @@
...
+#include "grid.hpp"
+#include "utils.hpp"
+
+/** Compute the normalized vector between two points.
+ *  @param[in]  r_mid     Position of the middle particle
+ *  @param[in]  r_out     Position of the outer particle
+ *  @param[out] vec       Normalized vector
+ *  @param[out] di        Normalization constant
+ */
+inline void calc_angle_vector(const Vector3d &r_mid, const Vector3d &r_out,
```
fweik: This function is not needed, you can just do `vec = get_mi_vector(r_mid, r_out)` at the call site, and then  do `vec.norm()` as needed.

fweik (changes requested):

src/core/bonded_interactions/angle_common.hpp
jngrad: Good idea.

src/core/bonded_interactions/angle_common.hpp
jngrad: This function will be removed as per fweik's suggestion.

src/core/bonded_interactions/angle_common.hpp
jngrad: I can use `Vector3d` here and inside the calling functions too for intermediate variables (the return variables are arrays for forces_inline.hpp). At some point in the future we should probably also use `Vector3d` in `/src/core/forces_inline.hpp`.

src/core/bonded_interactions/angle_cosine.hpp
jngrad: Do you mean `/src/core/forces_inline.hpp` should take the following C++11 form?
```c++
// current
bond_broken = calc_umbrella_pair_force(p1, p2, iaparams, dx, force);
// C++11
std::tie(bond_broken, force) = calc_umbrella_pair_force(p1, p2, iaparams, dx);
// C++17
auto [bond_broken, force] = calc_umbrella_pair_force(p1, p2, iaparams, dx);
```

jngrad:
Instead of splitting the angle force functions in two inlined functions and forwarding a long list of variables, there is now just one template function `calc_angle_generic_force()`, which is not inlined to avoid some overhead (the `inline` keyword causes a 0.8% slowdown and the template function still appears in the object dump). This approach is also slightly faster than the previous one with two inlined functions. The angle factor calculation is transferred to a lambda expression. Same design for `calc_angle_generic_3body_forces()`.

This PR causes an overall 1.2% slowdown, which is not caused by the lambdas nor by the `Vector3d`.

src/core/bonded_interactions/angle_cosine.hpp
KaiSzuttor: yes

KaiSzuttor:
how can inlining cause any overhead?

fweik:
@jngrad the function template is almost certainly inlined. It is neither true that functions with the `inline` keyword are always inlined nor that functions without it are not. The presence of a symbol in the object is also not an indicator  of whether a function was inlined, as the presence of symbols is governed by the linkage rules (that is by the function definition, and not by it's use). The only guaranteed effect of `inline` is to except it from ODR (one definition rule).

jngrad:
@fweik I didn't realize the linker would keep the symbols in the object. Good to know.

The `inline` keyword somehow creates an overhead for AngleCosine and AngleHarmonic. Here are the measurements for 30 runs of 100000 steps, with 95% error bars:

|           | AngleCossquare | AngleCosine   | AngleHarmonic |
| --------- | -------------- | ------------- | ------------- |
| PR        | 8.09 +/- 0.07  | 8.17 +/- 0.09 | 8.91 +/- 0.08 |
| PR+inline | 8.03 +/- 0.07  | 8.23 +/- 0.10 | 8.97 +/- 0.07 |
| original  | 8.01 +/- 0.07  | 8.03 +/- 0.04 | 8.80 +/- 0.08 |

src/core/bonded_interactions/angle_cosine.hpp
jngrad: I'll take care of that in another PR, as it requires updating the functions for pressure and non-bonded forces. Noted in #2502

KaiSzuttor: :+1:

src/core/TabulatedPotential.hpp
```diff
@@ -28,18 +28,37 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+ *  Forces and energies are evaluated by linear interpolation. The curves
+ *  @ref force_tab and @ref energy_tab must be sampled uniformly between
+ *  @ref minval and @ref maxval.
+ */
 struct TabulatedPotential {
+  /** Position on the x-axis of the first tabulated value. */
   double minval = -1.0;
+  /** Position on the x-axis of the last tabulated value. */
   double maxval = -1.0;
+  /** %Distance on the x-axis between tabulated values. */
```
fweik, line 42: `%`

src/core/bonded_interactions/angle_common.hpp
```diff
@@ -0,0 +1,98 @@
...
+/** Compute a three-body angle interaction force.
+ *  @param[in]  p_mid        Second/middle particle.
+ *  @param[in]  p_left       First/left particle.
+ *  @param[in]  p_right      Third/right particle.
+ *  @param[in]  forceFactor  Angle bending constant.
+ *  @param[out] force1       Force on particle 1.
+ *  @param[out] force2       Force on particle 2.
+ */
+template <typename ForceFactor>
+void calc_angle_generic_force(Particle const *p_mid, Particle const *p_left,
```
fweik: You can narrow the interface here: only the positions are needed, not the particles themselves.

src/core/bonded_interactions/angle_common.hpp
```diff
@@ -0,0 +1,98 @@
...
+  /* vector from p_left to p_mid */
+  auto vec1 = get_mi_vector(p_mid->r.p, p_left->r.p);
+  auto d1i = 1.0 / vec1.norm();
+  vec1 *= d1i;
+  /* vector from p_mid to p_right */
+  auto vec2 = get_mi_vector(p_right->r.p, p_mid->r.p);
+  auto d2i = 1.0 / vec2.norm();
+  vec2 *= d2i;
+  /* scalar product of vec1 and vec2 */
+  auto cosine = scalar(vec1, vec2);
```
fweik: Please use the operator overload and remove the include for `utils.hpp`.

src/core/bonded_interactions/angle_common.hpp
```diff
@@ -0,0 +1,98 @@
...
+ *  @param[in]  p_mid        Second/middle particle.
+ *  @param[in]  p_left       First/left particle.
+ *  @param[in]  p_right      Third/right particle.
+ *  @param[in]  forceFactor  Angle bending constant.
+ *  @param[out] force1       Force on particle 1.
+ *  @param[out] force2       Force on particle 2.
+ *  @param[out] force3       Force on particle 3.
+ */
+template <typename ForceFactor>
+void calc_angle_generic_3body_forces(Particle const *p_mid,
```
fweik: Here also the particles are not needed.

src/core/bonded_interactions/angle_common.hpp
```diff
@@ -0,0 +1,98 @@
...
+ *  @param[out] force1       Force on particle 1.
+ *  @param[out] force2       Force on particle 2.
+ */
+template <typename ForceFactor>
+void calc_angle_generic_force(Particle const *p_mid, Particle const *p_left,
+                              Particle const *p_right, ForceFactor forceFactor,
+                              double force1[3], double force2[3]) {
+  /* vector from p_left to p_mid */
+  auto vec1 = get_mi_vector(p_mid->r.p, p_left->r.p);
+  auto d1i = 1.0 / vec1.norm();
```
fweik: Most of the local variables here can be `const`

src/core/bonded_interactions/angle_common.hpp
```diff
@@ -0,0 +1,98 @@
...
+ *  @param[out] force2       Force on particle 2.
+ *  @param[out] force3       Force on particle 3.
+ */
+template <typename ForceFactor>
+void calc_angle_generic_3body_forces(Particle const *p_mid,
+                                     Particle const *p_left,
+                                     Particle const *p_right,
+                                     ForceFactor forceFactor, Vector3d &force1,
+                                     Vector3d &force2, Vector3d &force3) {
+  auto vec21 = -get_mi_vector(p_mid->r.p, p_left->r.p);
```
fweik: Most of the local variables can be `const`.

src/core/bonded_interactions/angle_cosine.hpp
```diff
@@ -31,166 +31,101 @@
...
-        vec21[j] / (vec21_magn * vec31_magn) - cos_phi * vec31[j] / vec31_sqr;
-  }
-
-  // note that F1 = -(F2 + F3)
-  for (j = 0; j < 3; j++) {
-    force1[j] = force1[j] - fac * (fj[j] + fk[j]);
-    force2[j] = force2[j] + fac * fj[j];
-    force3[j] = force3[j] + fac * fk[j];
-  }
+    auto pot_dep = K * (sin_phi * cos_phi0 - cos_phi * sin_phi0);
```
fweik: `const`

src/core/bonded_interactions/angle_cosine.hpp
```diff
@@ -31,166 +31,101 @@
...
+    auto fac = pot_dep / sin_phi;
```
fweik: `return`

src/core/bonded_interactions/angle_cosine.hpp
```diff
@@ -31,166 +31,101 @@
...
+  auto vec1 = get_mi_vector(p_mid->r.p, p_left->r.p);
+  double d1i = 1.0 / vec1.norm();
+  vec1 *= d1i;
+  /* vector from p_mid to p_right */
+  auto vec2 = get_mi_vector(p_right->r.p, p_mid->r.p);
+  double d2i = 1.0 / vec2.norm();
+  vec2 *= d2i;
   /* scalar product of vec1 and vec2 */
-  cosine = scalar(vec1, vec2);
+  double cosine = scalar(vec1, vec2);
```
fweik: You can directly do
```c++
auto const cosine = vec1.normalize() * vec2.normalize()
```
it might be worth while to put the cosine calculation into a utility function, potentially including the angle sanitization,
as this is reused here already and also useful in different places.

fweik (changes requested):
See inline comments, you should in general check that everything that can be `const` is.

src/core/TabulatedPotential.hpp
jngrad: The role of this symbol is to prevent the creation of an autolink to the `Distance` struct defined in `/src/core/short_range_loop.hpp` ([Doxygen doc](http://www.doxygen.nl/manual/faq.html#faq_class))

src/core/bonded_interactions/angle_common.hpp
jngrad: Done!

src/core/bonded_interactions/angle_common.hpp
jngrad: Done!

src/core/bonded_interactions/angle_common.hpp
jngrad: Done!

src/core/bonded_interactions/angle_common.hpp
jngrad: Good point.

src/core/bonded_interactions/angle_common.hpp
jngrad: Done.

src/core/bonded_interactions/angle_cosine.hpp
jngrad: Most callers need the inverse length of the vectors besides the cosine, obtained by the `.norm()` method. I didn't use the `.normalize()` method to avoid calling `.norm()` twice for each vector. The dependency to `utils.hpp` was removed.

src/core/bonded_interactions/angle_cosine.hpp
jngrad: Done.

src/core/bonded_interactions/angle_cosine.hpp
jngrad: I'm returning a temporary variable named `fac` to follow the naming convention in other `forceFactor` lambdas. Should I avoid that?

jngrad: The lambdas now use a `return` directly.

jngrad:
I think this is as far as refactoring should go in this PR. Some of the angle potentials will be slightly rewritten in a future PR anyway to address the observations in #2145. The small drop in performance is unfortunate, but the code is now more maintainable.

RudolfWeeber (approved):
This improves clarity alot. I'lll pull this into #2542 since there are overlapping changes.

PR #2499, opened 2019-02-06
WIP: Replaced dawannar with MDDS and made it parallel

fweik:
- bugfix for `Utils::Mpi::gatherv`.
- implementation of MPI_Allgatherv for non-mpi datatypes
- Removed `dawaanr` and replaced it by mdss, which `dawaanr` was a strict subset of
- Made mdss run in parallel.


RudolfWeeber:
The code looks good, but it is slower by a factor of 2 on a single node.
I think, the double calculation needs to be avoided.

With regards to the mpi function: So the bug was that for out-of-place gathers, the local contribution was ignored?

fweik:
Yes, but only for non-mpi types. I don't think it was used that way in Espresso, so no problem.

fweik:
I see considerable speedup now:

ranks |  PR  |  python
--------|--------|------------
 1 |  0.0263761997223 | 0.0472198963165
 2 | 0.0203853130341 |
 4 | 0.0131695985794 |
 6 | 0.0115113019943 |
 8 | 0.00787239074707 |
10| 0.00673069953918 |

RudolfWeeber:
Please add Doxygen to the functions.

Please add headings to the different part of the main force calculation funciton, such as
Calculate ias betweenn articles local to the node
calculate ia of a particle with its own images
calculate ia between particle and 2nd particle + its images
etc.


Shouldn't the all_gather go somewhere in utils?

fweik:
Can you please tell me for which functions it is not clear to you from their name what they do?

fweik:
@RudolfWeeber I added some documentation.

RudolfWeeber:
Something about the replica handling is wrong. The devitations with respect to p3m jump by 5 orders of magnituede (up) when you go from 0 to 2 replica. I pushed a failing test.

fweik:
I'm abandoning this for now due to time constraints.

RudolfWeeber:
@stekajack this is the dipolar pr. You'll have to examine the state of the code, first. IIRC, the final version was incorrect, if replica are included. You either need to figure out, why or roll back to a version without latency hiding, IIRC, but it was a long time ago.

stekajack:
> @stekajack this is the dipolar pr. You'll have to examine the state of the code, first. IIRC, the final version was incorrect, if replica are included. You either need to figure out, why or roll back to a version without latency hiding, IIRC, but it was a long time ago.

its gonna take me a while to get up to speed, but i will slowly chip away at the problem in the coming weeks :) thanks!

fweik:
@stekajack If you have questions, feel free to ask. I will answer to the extend that my memory servers me :-)

stekajack:
@fweik thank you for your offer, i will make sure to take you up on it if necessary :) Luckily for me, espresso gods got involved and the bugs are fixed and mdds is ported in to espresso 4.2. With some preliminary testing mdds proves to work well but some larger scale tests are necessary. I unfortunately have to pause with this rn, because i have a project submission deadline, but i hope in september to have done my part. Cheers

PR #2501, opened 2019-02-07
CI: move odd processor check to build stage.

### LINKED ISSUE #2476: Improve CI pipeline parallelism

KaiSzuttor:
Fixes #2476

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -351,8 +351,6 @@ check_with_odd_no_of_processors:
   stage: build
   image: gitlab.icp.uni-stuttgart.de:4567/espressomd/docker/cuda:9.0
   when: on_success
-  dependencies:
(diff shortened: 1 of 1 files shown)


jngrad:
You may also remove the artifact in `cuda-maxset`:
https://github.com/espressomd/espresso/blob/010b5f63b9b2ac44ca993e55028acac7e983210c/.gitlab-ci.yml#L173-L176

fweik (approved):

PR #2505, opened 2019-02-11
Fixed soft sphere cutoff calculation

fweik:
- Added `WCA` to maxset
- Fixed cutoff calculation for  `SOFT_SPHERE`
- Added function to reset all non-bonded interactions.


jngrad (approved):
Looks good to me.

PR #2506, opened 2019-02-11
Caliper: A Performance Analysis Toolbox in a Library

fweik:
For docs see https://llnl.github.io/Caliper/index.html, profiler annotations can be enabled by `WITH_PROFILER=ON`. The lib is available from github and installed in stow at the icp.


fweik:
We should probably write some documentation on how to use this an how to add annotations.

KaiSzuttor (approved):

PR #2511, opened 2019-02-13
WIP: mutually-compatible features checkpointing

### LINKED ISSUE #333: No testcase for LB checkpointing

### LINKED ISSUE #1848: Test pickle checkpointing
@davidsean @pkreissl @konradbreitsprecher

### LINKED ISSUE #2365: LB GPU test for non binary checkpointing missing

### LINKED ISSUE #2436: checkpoint testing needs a rewrite
In the current test framework for the checkpoints mutually exclusive features can not be tested. That's a fundamental issue and should be resolved in general.

### LINKED ISSUE #2555: Extend LB checkpointing tests
* test multiple lattice sites
* test more properties: population, density, velocity, stress, ... per lattice site
* loading a checkpointed LB fluid into an LB object with different lattice geometry should raise an error (maybe write lattice dimensions to the checkpoint file)

### LINKED ISSUE #2564: LB: incompatible agrid is handled to late (leads to segfault)

jngrad:
Fixes #2436, Fixes #2365, Fixes #333 

Description of changes:
 - use a single checkpoint script to test multiple mutually-compatible features
 - define semicolon-delimited lists of feature combinations in `/testsuite/python/CMakeLists.txt`
    - each combination is a hyphen-delimited list of mutually-compatible features
    - use a dot to add flags, e.g. `lb.gpu-p3m.cpu-lj` tests for LB on the GPU together with P3M on the CPU and LJ
    - for non-binary checkpointing, `lb.gpu-lb.ascii-p3m.cpu-lj` could be used, or `lb.gpu.ascii-p3m.cpu-lj` after a rewrite of the `modes` parser

@jainka, @KaiSzuttor 

Edit: this PR also fixes #1848, fixes #2555, fixes #2564


KaiSzuttor:
What is the status here?

jngrad:
The checkpointing tests now cover LB CPU+GPU (#333), ASCII+binary checkpointing (#2365) and are more flexible (#2436). Provided CI passes, I'd say we're good to merge and close all 3 tickets.

testsuite/python/CMakeLists.txt
```diff
@@ -59,8 +63,18 @@ function(PYTHON_TEST)
   set(python_tests ${python_tests} ${TEST_FILE} PARENT_SCOPE)
 endfunction(PYTHON_TEST)
 
-python_test(FILE save_checkpoint.py MAX_NUM_PROC 4)
-python_test(FILE test_checkpoint.py MAX_NUM_PROC 4 DEPENDS save_checkpoint)
+foreach(TEST_COMBINATION lb.cpu-p3m.cpu-lj;lb.gpu-p3m.cpu-lj)
```
KaiSzuttor, line 68: Could you please add a comment how to add new tests?

jngrad: Done.

jngrad:
Ticket #1848 can also be closed after this merge.

fweik:
@KaiSzuttor @jngrad What is still missing here?

jngrad:
@fweik @KaiSzuttor We should be able to merge cleanly now.

jngrad:
@KaiSzuttor in the current `python` branch, the `LB_parameters_gpu` struct holds the lattice dimensions, but the `LB_parameters` (for cpu) does not, so sometimes it is fetched from the `lattice` global variable:
https://github.com/espressomd/espresso/blob/b24864f4a56ce05356b898e51cfbaa53612f417b/src/core/grid_based_algorithms/lb_interface.cpp#L501
and other times it is calculated without rounding:
https://github.com/espressomd/espresso/blob/b24864f4a56ce05356b898e51cfbaa53612f417b/src/core/grid_based_algorithms/lb_interface.cpp#L648-L652
even though the `lattice` object uses rounding:
https://github.com/espressomd/espresso/blob/b24864f4a56ce05356b898e51cfbaa53612f417b/src/core/grid_based_algorithms/lattice.cpp#L39
Is there a reason for not using the lattice object directly in the CPU code?

jngrad:
Container `check_with_odd_no_of_processors` says the CPU checkpoint file cannot be written to disk. Can't reproduce the issue locally inside the Docker container. Not sure why MPI should interact here, the only change in the save function for the CPU LB was:
https://github.com/espressomd/espresso/blob/98c60a40634882ff55bc1cd34f2b3dfcf92f5c42/src/core/grid_based_algorithms/lb_interface.cpp#L794-L799

KaiSzuttor:
@jngrad I don't think that there is a reason for not using ```lblattice``` directly. I just missed that while refactoring the CPU LB.

jngrad:
@KaiSzuttor thanks! I'll make the change. I've also reproduced the traceback from the failing container. The issue shows up when adding a CPU LB to the actor list and running the script with `mpiexec -n 3`:
```python
lbf = espressomd.lb.LBFluid(agrid=0.5, visc=1.3, dens=1.5, tau=0.01)
system.actors.add(lbf)
```
The failure is at this line, for n = 0:
https://github.com/espressomd/espresso/blob/b24864f4a56ce05356b898e51cfbaa53612f417b/src/core/grid_based_algorithms/lbboundaries.cpp#L278

Have you ever seen this  MPI interaction before?

fweik:
Have you checked that `lb_fields` is actually allocated?

KaiSzuttor:
MnotWE:
``` python
import espressomd
import espressomd.lb

system = espressomd.System(box_l=[10.0, 12.0, 14.0])
system.cell_system.skin = 0.1
system.time_step = 0.01
lbf = espressomd.lb.LBFluid(agrid=0.5, visc=1.3, dens=1.5, tau=0.01)
system.actors.add(lbf)                                              
```

KaiSzuttor:
not yet clear to me, but the issue seems to be located in ```mpi_bcast_lb_params``` which is called when setting ```agrid```

KaiSzuttor:
ahh, ```agrid``` is incompatible with the local box

KaiSzuttor:
@fweik how to handle this? there is only a check in the lattice class that prints to ```runtimeErrorMsg```

fweik:
This would be the reason why most codes let the user choose the grid dimensions and calculate the grid constant and not the other way around. Not sure if we want to change this, but since we just made incompatible changes to the LB interface anyway now would be the time. What is the opinion of our UX officer @RudolfWeeber on this?

jngrad:
Ah ok, it was my understanding that espresso threw an error for `agrid` values incompatible with the number of nodes. For example using the MnotWE, with 5 nodes: `Exception: LB fluid activation: ERROR: box_l in direction 0 is too small`, with 4 nodes: runs ok, with 3 nodes: MPI traceback. Thanks! I would never have suspected the box size to be the issue.

jngrad:
Also, I get a similar MPI traceback when the box dimensions are not integer multiples of `agrid`; e.g. `agrid=0.50001`. This means floating point errors can cause a failure, too.

fweik:
@KaiSzuttor is also think there should be bounds checking with `ADDITIONAL_CHECKS` enabled.

fweik:
@jngrad yes, this is why the user should chose the integral value and not the fractional one.

KaiSzuttor:
if we'd the time to rewrite the lattice class and the interface...

jngrad:
@KaiSzuttor the sanity check should now be doing its job.

RudolfWeeber:
Ready to merge?

jngrad:
@RudolfWeeber there are still open points in https://github.com/espressomd/espresso/issues/2555#issuecomment-471038543

jngrad:
For reference: this PR also fixes #1848, fixes #2555, fixes #2564

src/core/grid_based_algorithms/lattice.cpp
```diff
@@ -45,17 +45,22 @@ int Lattice::init(double *agrid, double *offset, int halo_size, size_t dim) {
...
+      incompatible_agrid = true;
       runtimeErrorMsg() << "Lattice spacing agrid[" << dir << "]=" << agrid[dir]
                         << " is incompatible with local_box_l[" << dir
                         << "]=" << local_box_l[dir] << " ( box_l[" << dir
                         << "]=" << box_l[dir] << " node_grid[" << dir
                         << "]=" << node_grid[dir] << " )";
     }
   }
+  if (incompatible_agrid) {
+    return ES_ERROR;
```
KaiSzuttor: I think that does not help, since the return value is not checked... we could also directly return in after the ```runtimeErrorMsg```

KaiSzuttor (approved):

src/core/grid_based_algorithms/lattice.cpp
jngrad: Thanks! Both the `.at(n)` solution and the return value check solution worked; we now get a clean `Exception: LB fluid activation` instead of the full MPI traceback.

jngrad:
@RudolfWeeber it's ready to merge

PR #2512, opened 2019-02-13
Coulomb/Dipolar refactor

reinaual:
Extracted coulomb and dipole switch-statements into separate functions


src/core/CMakeLists.txt
```diff
@@ -53,6 +53,8 @@ set(EspressoCore_SRC
   topology.cpp
   tuning.cpp
   virtual_sites.cpp
+  coulomb_switch.cpp
```
fweik: Can you please call the files just `coulomb.{hpp,cpp}` and `dipole.{hpp,cpp}` and put them into `electrostatics_magnetostatics`.

fweik (changes requested):
Please also check where headers from the electrostatics/dipolar methods can be removed. The CI failure is an technical issue unrelated to the PR.

src/core/electrostatics_magnetostatics/coulomb.cpp
```diff
@@ -0,0 +1,745 @@
...
+#include "initialize.hpp"                             // on_ghost_flags_change
+#include "integrate.hpp"                              // skin
+#include "layered.hpp"                                // layer_h
+#include "nonbonded_interactions/reaction_field.hpp"  // Reaction_field_params
+#include "npt.hpp"                                    // nptiso
+#include "statistics.hpp"                             // Observable_stat
+
+#ifdef ELECTROSTATICS
+
+void pressure_inline_add_pair_pressure(Particle *p1, Particle *p2, double *d,
```
fweik: This has to go to the header so that it can be inlined.

src/core/electrostatics_magnetostatics/coulomb.cpp
```diff
@@ -0,0 +1,745 @@
...
+    maggs_init();
+    /* Maggs electrostatics needs ghost velocities */
+    on_ghost_flags_change();
+    break;
+  default:
+    break;
+  }
+}
+
+void forces_inline_calc_pair_coulomb_force(Particle *p1, Particle *p2,
```
fweik: This has to go to the header so that it can be inlined.

src/core/electrostatics_magnetostatics/coulomb.cpp
```diff
@@ -0,0 +1,745 @@
...
+    assert(!Scafacos::dipolar());
+    Scafacos::add_long_range_force();
+    break;
+#endif
+  default:
+    break;
+  }
+}
+
+void energy_inline_add_pair_coulomb_energy(Particle *p1, Particle *p2,
```
fweik: This has to go to the header so that it can be inlined.

src/core/electrostatics_magnetostatics/dipole.cpp
```diff
@@ -0,0 +1,420 @@
...
+  case DIPOLAR_P3M:
+    dp3m_init();
+    break;
+#endif
+  default:
+    break;
+  }
+}
+
+void forces_inline_calc_pair_dipole_force(Particle *p1, Particle *p2, double *d,
```
fweik: This has to go to the header so that it can be inlined.

src/core/electrostatics_magnetostatics/dipole.cpp
```diff
@@ -0,0 +1,420 @@
...
+#endif
+  case DIPOLAR_NONE:
+    break;
+  default:
+    runtimeErrorMsg() << "unknown dipolar method";
+    break;
+  }
+}
+
+void energy_inline_add_pair_dipole_energy(Particle *p1, Particle *p2, double *d,
```
fweik: This has to go to the header so that it can be inlined.

src/core/electrostatics_magnetostatics/coulomb.hpp
```diff
@@ -0,0 +1,51 @@
+#ifndef ESPRESSO_COULOMB_SWITCH_HPP
+#define ESPRESSO_COULOMB_SWITCH_HPP
+
+#include "statistics.hpp" // Observable_stat
+
+#ifdef ELECTROSTATICS
+
```
fweik, line 33: Can you please put everything here into a namespace `Coulomb`, and shorten the names e.g. `pressure_inline_add_pair_pressure` -> `add_pair_pressure` and so on. The same for dipolar.

fweik (changes requested):

fweik:
Please also integrate the changes from https://github.com/fweik/espresso/tree/reaction_field (removing the INTER_RF method).

fweik:
This already looks pretty good, I think the naming can still be improved. @KaiSzuttor or @RudolfWeeber can maybe give some input?

fweik:
Please merge again.

src/core/electrostatics_magnetostatics/coulomb.cpp
```diff
@@ -0,0 +1,555 @@
...
+#include "electrostatics_magnetostatics/p3m_gpu.hpp" // p3m_gpu_init
+#include "errorhandling.hpp"                         // runtime_error
+#include "initialize.hpp"                            // on_ghost_flags_change
+#include "layered.hpp"                               // layer_h
+
+#ifdef ELECTROSTATICS
+
+namespace Coulomb {
+
+void pressure_n_coulomb(int &n_coulomb) {
```
fweik: Can you please remove all the `coulomb`/`dipolar` parts from the names that are redundant with the namespaces?

fweik: You can also remove the prefixed like `pressue_` and `nonbonded_`, it should be clear by the name of the function what it doesn.

src/core/electrostatics_magnetostatics/coulomb.cpp
```diff
@@ -0,0 +1,555 @@
...
+    case COULOMB_MMM2D:
+      add_mmm2d_coulomb_pair_force(q1q2, d, dist2, dist, force);
+      break;
+    default:
+      break;
+    }
+  }
+}
+
+void icc_calc_long_range_force() {
```
fweik: This should be removed, icc should use the normal coulomb longrange function, the same for the pair forces

fweik (changes requested):

fweik:
Ok, now we are getting somewhere. Now please split up the control struct `coulomb` into a `coulomb` and a `dipolar` struct, so that there is no cross-dependency.

src/core/energy_inline.hpp
```diff
@@ -68,12 +67,16 @@
 #include "electrostatics_magnetostatics/mmm2d.hpp"
```
fweik: I think those explicit includes are not needed any more, `coulomb.hpp` and `dipolar.hpp` should be sufficient.

src/core/initialize.cpp
```diff
@@ -39,6 +39,7 @@
 #include "electrostatics_magnetostatics/p3m-dipolar.hpp"
```
fweik: Please check if the explicit includes are still needed.

src/core/nonbonded_interactions/nonbonded_interaction_data.cpp
```diff
@@ -56,13 +56,13 @@
 #include "electrostatics_magnetostatics/mmm2d.hpp"
```
fweik: These should not be needed any more

src/core/nonbonded_interactions/nonbonded_interaction_data.cpp
```diff
@@ -77,28 +77,31 @@
...
 /****************************************
  * variables
  *****************************************/
 int max_seen_particle_type = 0;
 std::vector<IA_parameters> ia_params;
 
-#if defined(ELECTROSTATICS) || defined(DIPOLES)
-Coulomb_parameters coulomb = {
 #ifdef ELECTROSTATICS
+Coulomb_parameters coulomb = {
```
fweik: Move to `coulomb.hpp`

src/core/nonbonded_interactions/nonbonded_interaction_data.cpp
```diff
@@ -77,28 +77,31 @@
...
     0.0,
     COULOMB_NONE,
-#endif
+};
+#endif // ELECTROSTATICS
+
 #ifdef DIPOLES
+Dipole_parameters dipole = {
```
fweik: Please move do `dipolar.hpp`

src/core/nonbonded_interactions/nonbonded_interaction_data.cpp
reinaual: should i also move the `enum`?

fweik: Yes, everything. Same for dipolar, so that only `#include "coulomb.hpp"` is needed to use it.

RudolfWeeber (dismissed):
Once this is in, we can also add high-level developer docs for adding new Coulomb/dipolar methods.
Concentrating these in one place is a big improvement.

fweik:
@RudolfWeeber I agree, this is also kind of and test on how a complete method interface, including short range contribution had to look like. This is of course not final yet, but it already makes many contact points visible.

RudolfWeeber:
Please merge the python branch. Then it can be merged.

fweik (approved):

PR #2513, opened 2019-02-13
Expose Shape.calc_distance() in Python

### LINKED ISSUE #2495: Py: Shape.calc_distance() not exposed in Python
although usable via call_method

jngrad:
Fixes #2495

There is unfortunately no Python docstring to document the `calc_distance` method. Maybe it could be added programmatically using the following logic in `/src/python/espressomd/script_interface.pyx`:

```python
class ScriptInterfaceHelper(PScriptInterface):
    _so_name = None
    _so_bind_methods = {}
    ...
    def generate_caller(self, method_name, docstring=""):
        def template_method(**kwargs):
            res = self.call_method(method_name, **kwargs)
            return res
        template_method.__doc__ = docstring
        return template_method

    def define_bound_methods(self):
        if not isinstance(self._so_bind_methods, dict):
            self._so_bind_methods = {x: "" for x in self._so_bind_methods}
        for method_name, method_doc in self._so_bind_methods.items():
            setattr(self, method_name, self.generate_caller(method_name, method_doc))
```
after which `help(shape.calc_distance)` generates:
```
Help on cython_function_or_method in module espressomd.script_interface:

template_method(**kwargs)
    Calculate the distance between a point and the shape surface.
```
which is not very helpful: the method name is shown as `template_method` and the docstring does not appear in the Sphinx documentation.

PR Checklist
------------
 - [x] Tests?
   - [x] Interface
   - [x] Core 
 - [x] Docs?

### DIFF
--- a/src/python/espressomd/shapes.py
+++ b/src/python/espressomd/shapes.py
@@ -18,7 +18,13 @@ from .script_interface import ScriptInterfaceHelper, script_interface_register
 
 
 @script_interface_register
-class Cylinder(ScriptInterfaceHelper):
+class Shape(ScriptInterfaceHelper):
+    _so_name = "Shapes::Shape"
+    _so_bind_methods = ("calc_distance",)
+
+
+@script_interface_register
+class Cylinder(Shape):
 
     """
     A cylinder shape.
@@ -41,11 +47,10 @@ class Cylinder(ScriptInterfaceHelper):
 
     """
     _so_name = "Shapes::Cylinder"
-    _so_bind_methods = ['calc_distance']
 
 
 @script_interface_register
-class Ellipsoid(ScriptInterfaceHelper):
+class Ellipsoid(Shape):
 
     """
     An ellipsoid.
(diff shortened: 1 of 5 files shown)


src/python/espressomd/shapes.py
```diff
@@ -41,6 +41,7 @@ class Cylinder(ScriptInterfaceHelper):
 
     """
     _so_name = "Shapes::Cylinder"
+    _so_bind_methods = ['calc_distance']
```
fweik: Couldn't you just introduce a base class an put the member there instead of repeating it for every shape?

jngrad: Done.

src/python/espressomd/shapes.py
```diff
@@ -18,7 +18,13 @@
 
 
 @script_interface_register
-class Cylinder(ScriptInterfaceHelper):
+class Shape(ScriptInterfaceHelper):
```
fweik: I don't think that Shape should be a script interface object. Just use multiple inheritance, don't use a base class here and don't set `_so_name` here. `Shapes::Shape` does not exist and this class can not be instantiated anyway.

fweik (changes requested):

src/python/espressomd/shapes.py
jngrad: Thanks!

fweik (approved):

PR #2514, opened 2019-02-13
Re-enable exotic architectures and set time limit for tests

### LINKED ISSUE #2475: Tests on emulated platforms take too long
There is a time limit of 3h for the jobs, for some of the tests that run in qemu that is too low.
I think it is sufficient to run them only scheduled, like every night, and not on every update.

mkuron:
Fixes #2475. I have also increased the timeout to 6h now.


fweik (approved):

mkuron:
The jobs have completed successfully, but collecting the code coverage fails for some unknown reason on some jobs: https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/pipelines/5545/failures

PR #2516, opened 2019-02-14
LB: make three point interpolation available to interface (shows that…

KaiSzuttor:


KaiSzuttor:
@hmenke any idea what's wrong here? Interested in fixing?

PR #2518, opened 2019-02-15
Don't call std::exit in shared libraries

### LINKED ISSUE #2249: rpmlint report
Update of #774

```
libEspresso4.i586: W: shared-lib-calls-exit /usr/lib/libEspressoCore.so.4 exit@GLIBC_2.0
This library package calls exit() or _exit(), probably in a non-fork()
context. Doing so from a library is strongly discouraged - when a library
function calls exit(), it prevents the calling program from handling the
error, reporting it to the user, closing files properly, and cleaning up any
state that the program has. It is preferred for the library to return an
actual error code and let the calling program decide how to handle the
situation.
```

mkuron:
Fixes #2249

@junghans

### DIFF
--- a/maintainer/CI/build_cmake.sh
+++ b/maintainer/CI/build_cmake.sh
@@ -231,7 +231,8 @@ cmd "make -k -j${build_procs}" || cmd "make -k -j1" || exit $?
 end "BUILD"
 
 # check for exit function, which should never be called from shared library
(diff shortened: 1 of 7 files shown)


junghans (approved):
Nice!

fweik (approved):

PR #2522, opened 2019-02-17
Suppress gcov output in CI

mkuron:
gcove produces thousands of lines of output so that the actual build output can only be seen in Gitlab CI after clicking the "Raw log" link. It seems like codecov finally updated to https://github.com/codecov/codecov-bash/issues/121, so we can use it now.

### DIFF
--- a/maintainer/CI/build_cmake.sh
+++ b/maintainer/CI/build_cmake.sh
@@ -279,9 +279,9 @@ if $with_coverage; then
     lcov -q --remove coverage.info '*/doc/*' --output-file coverage.info # filter out docs
     # Uploading report to CodeCov
     if [ -z "$CODECOV_TOKEN" ]; then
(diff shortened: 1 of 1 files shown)


mkuron:
Before: https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/75706. After: https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/75745.

mkuron:
Doesn‘t seem to have intended effect

hmenke:
@mkuron  At pairinteraction we have just [disabled `gcov` altogether](https://github.com/hmenke/pairinteraction/blob/05bcad7340e7599a43c7f6e597aa35d0f51d3b71/ci/build_docker.sh#L118) because we collect a coverage report with `lcov` and the [Codecov uploader can detect and use this report](https://travis-ci.org/hmenke/pairinteraction/jobs/493507970#L1403) directly.

PR #2524, opened 2019-02-18
Use the implementation-defined EOF value

jngrad:
Fixes the broken pipeline.

The following error message appeared in containers `ubuntu:s390x`, `ubuntu:arm64`, `ubuntu:armhf`, `ubuntu:ppc64le`:
```
[  1%] Building CXX object src/core/CMakeFiles/EspressoCore.dir/PdbParser.cpp.o
/builds/espressomd/espresso/src/core/PdbParser.cpp: In member function 'bool PdbParser::PdbParser::parse_itp_file(const string&)':
/builds/espressomd/espresso/src/core/PdbParser.cpp:120:49: error: comparison is always false due to limited range of data type [-Werror=type-limits]
             if (std::isspace(buf[0]) || (buf[0] == -1)) {
```

The cause is a mismatch between signed/unsigned types: the `EOF` value is implementation-dependent (typically -1 or 255), and the `buf` string's value_type `char` can or cannot be cast to a signed integer depending on the implementation. One solution is to cast `EOF` (or the result of `std::char_traits::eof()`) to the string's value_type.

### DIFF
--- a/src/core/PdbParser.cpp
+++ b/src/core/PdbParser.cpp
@@ -79,8 +79,6 @@ bool PdbParser::parse_pdb_file(const string &filename) {
 bool PdbParser::parse_itp_file(const string &filename) {
   ifstream file(filename.c_str());
   string tmp, buf;
-  const decltype(buf)::value_type EOF_value =
-      std::char_traits<decltype(buf)::value_type>::eof();
   itp_atom atom;
   std::size_t pos;
 
@@ -118,7 +116,7 @@ bool PdbParser::parse_itp_file(const string &filename) {
             buf = char(file.get());
 
             /* Ignore leading whitespace, check for end of file */
-            if (std::isspace(buf[0]) || (buf[0] == EOF_value)) {
+            if (std::isspace(buf[0]) || file.eof()) {
               continue;
             }
             /* End of atoms section */
@@ -158,7 +156,7 @@ bool PdbParser::parse_itp_file(const string &filename) {
             }
 
             /* Ignore leading whitespace, check for end of file */
-            if (std::isspace(buf[0]) || (buf[0] == EOF_value)) {
+            if (std::isspace(buf[0]) || file.eof()) {
               continue;
             }
 
--- a/src/core/PdbParser.cpp
+++ b/src/core/PdbParser.cpp
@@ -79,6 +79,8 @@ bool PdbParser::parse_pdb_file(const string &filename) {
 bool PdbParser::parse_itp_file(const string &filename) {
   ifstream file(filename.c_str());
   string tmp, buf;
+  const decltype(buf)::value_type EOF_value =
+      std::char_traits<decltype(buf)::value_type>::eof();
   itp_atom atom;
   std::size_t pos;
 
@@ -87,7 +89,7 @@ bool PdbParser::parse_itp_file(const string &filename) {
 
   while (file.good()) {
     try {
-      buf = char(file.get());
+      buf = file.get();
       /* Skip leading whitespace */
       if (std::isspace(buf[0]))
         continue;
@@ -113,11 +115,10 @@ bool PdbParser::parse_itp_file(const string &filename) {
 
         if (section == "atoms") {
           while (file.good()) {
-            buf = char(file.get());
+            buf = file.get();
 
-            /* Ignore leading whitespace, check for end of file (standard says
-             * EOF is "generally" -1) */
-            if (std::isspace(buf[0]) || (buf[0] == -1)) {
+            /* Ignore leading whitespace, check for end of file */
+            if (std::isspace(buf[0]) || (buf[0] == EOF_value)) {
               continue;
             }
             /* End of atoms section */
@@ -144,7 +145,7 @@ bool PdbParser::parse_itp_file(const string &filename) {
           itp_atomtype type;
           std::string type_name;
           while (file.good()) {
-            buf = char(file.get());
+            buf = file.get();
 
             /* Ignore leading whitespace */
             if (std::isspace(buf[0])) {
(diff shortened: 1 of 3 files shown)


mkuron (approved):

src/core/unit_tests/CMakeLists.txt
```diff
@@ -68,3 +68,4 @@ unit_test(NAME field_coupling_force_field SRC field_coupling_force_field_test.cp
 unit_test(NAME Span_test SRC Span_test.cpp)
 unit_test(NAME periodic_fold_test SRC periodic_fold_test.cpp)
 unit_test(NAME sendrecv_test SRC sendrecv_test.cpp DEPENDS Boost::mpi MPI::MPI_CXX)
+unit_test(NAME PdbParser_test SRC PdbParser_test.cpp DEPENDS EspressoCore)
```
KaiSzuttor: Could you please remove the Pdbparser from the EspressoCore target and make it a shared library. Then the test does not depend on the whole core anymore

fweik: If so it should be move out of core into an extra library.

src/core/PdbParser.cpp
```diff
@@ -79,6 +79,8 @@ bool PdbParser::parse_pdb_file(const string &filename) {
 bool PdbParser::parse_itp_file(const string &filename) {
   ifstream file(filename.c_str());
   string tmp, buf;
+  const decltype(buf)::value_type EOF_value =
```
fweik: Is there is specific reason for using `decltype` instead of just using `string` which is hard-coded one line above? Also you could use auto so you don't have to repeat the type.

jngrad: I tried `const auto EOF_value` first, but the auto is deduced as a signed int.  Not sure why. `decltype(buf)::value_type` is a char that cannot be cast to a signed int on the previously mentioned architectures. Here is a minimum working example:
```c++
#include <iostream>
#include <string>
int main(void) {
  std::string buf = "abcd";
  std::cout << typeid(unsigned char).name() << ": unsigned char" << std::endl;
  std::cout << typeid(signed char).name() << ": signed char" << std::endl;
  std::cout << typeid(char).name() << ": char" << std::endl;
  std::cout << typeid(unsigned int).name() << ": unsigned int" << std::endl;
  std::cout << typeid(signed int).name() << ": signed int" << std::endl;
  std::cout << typeid(int).name() << ": int" << std::endl;
  const auto EOF_value1 =
      std::char_traits<decltype(buf)::value_type>::eof();
  const decltype(buf)::value_type EOF_value2 =
      std::char_traits<decltype(buf)::value_type>::eof();
  std::cout << typeid(decltype(EOF_value1)).name() << ": const auto" << std::endl;
  std::cout << typeid(decltype(EOF_value2)).name() << ": const decltype(buf)::value_type" << std::endl;
  return 0;
}
```
Output:
```
h: unsigned char
a: signed char
c: char
j: unsigned int
i: signed int
i: int
i: const auto
c: const decltype(buf)::value_type
```

fweik: EOF is not of `value_type`, that is kind of the point. Also you only ever need to compare it to `value_type`,
I don't really see the problem. Also `std::char_traits<char>::int_type` (the return type of eof) is standardized as `int`, so chances are that the eof value will not fit into a char. (see https://en.cppreference.com/w/cpp/string/char_traits).

fweik: If think the proper way to do it is to use `std::char_traits::not_eof` anyway, instead of an `int` comparison.

mkuron: `buf[0]` is definitely of type `char`, so the comparison can only ever be true if `eof` fits into a `char`.

fweik: Yes, but it shouln'd be. `std::ifstream::get` returns an `int_type`, which is either a valid character, or `eof`, also an  `int_type`. This result can then be checked for `eof`.

fweik: I think we can all agree that this API is horrible, but if we want to make it portable, we can't make implicit assumptions such as that the result of `get` or `eof` fit into a char.

jngrad: The issue here is not about the actual value `EOF` since it is -1 by definition, and -1 == 255 (mod 256). How the compiler interprets the signedness of a char is also irrelevant, because the comparison operator will cast the value correctly. Without the -Wextra flag, it compiles just fine and the PdbParser unit test passes. I think the warning just says the compiler might try to optimize away part of the conditional based on the range of values it can hold before the cast. The manual cast of `EOF` helps the compiler figuring out what the developer's intention is.

jngrad: Hmm... let me take that back: the conditional doesn't cast correctly, `EOF` is not detected at all. Let's use `ifstream::eof()` instead.

fweik:
This is not correct, now it is checked if the stream is eof _after_ the get.

jngrad:
Doesn't `buf[0] == EOF_value` check *after* the get?

fweik:
Ah yes you are right, my bad.

src/core/unit_tests/CMakeLists.txt
jngrad: Fixed by simply linking to the cpp file.

fweik: @KaiSzuttor are you happy with this solution? Why shouldn't it be linked against EspressoCore?

src/core/unit_tests/CMakeLists.txt
```diff
@@ -68,3 +68,4 @@ unit_test(NAME field_coupling_force_field SRC field_coupling_force_field_test.cp
 unit_test(NAME Span_test SRC Span_test.cpp)
 unit_test(NAME periodic_fold_test SRC periodic_fold_test.cpp)
 unit_test(NAME sendrecv_test SRC sendrecv_test.cpp DEPENDS Boost::mpi MPI::MPI_CXX)
+unit_test(NAME PdbParser_test SRC PdbParser_test.cpp ../PdbParser.cpp)
```
KaiSzuttor, line 71: Actually I meant to create a new target for the PDB parser since it does not depend on anything else in the core. I think it's not the cleanest solution to use the relative path to the source file here but depend on the to be created PDB target.

PR #2526, opened 2019-02-20
cmake: separate lib for pdbparser.

KaiSzuttor:


jngrad (approved):
Perfect

fweik:
You should then move the code out of the core folder so that the folder layout reflects the dependencies. (This is so you can not just include the header without adding the dependency in cmake).

fweik:
You should then also move the test :-)

KaiSzuttor:
Have a unit test in the pdbparser directory?

fweik:
Yes, why would it be in `core`?

fweik:
Sorry to nag you again, but I think the folder layout should be either flat or have standard layout (include, src, tests, ...)

PR #2527, opened 2019-02-20
cmake: add cluster_analysis to the EspressoCore target.

KaiSzuttor:
Since it depends on the core.


fweik:
Why?

KaiSzuttor:
because we agreed on having a clean library structure.

fweik:
`EspressoCore` does not actually depend on `cluster_analysis`, only the other way round. This should stay it's own target, that depends on `EspressoCore`.

fweik (changes requested):
Don't add the cluster analysis to the `EspressoCore` target, remove dependency of `EspressoCore` on `cluster_analysis` and link `cluster_analysis` in the script interface where it is actually used. `cluster_analysis` can also be moved out of the core folder.

PR #2529, opened 2019-02-21
Caliper section markings

fweik:
Start and stop profiler sections from python.


src/python/espressomd/profiler.pyx
```diff
@@ -0,0 +1,31 @@
...
+
+
+def begin_section(name):
+    """
+    Start named section in profiler.
+
+    Parameters
+    ----------
+
+    name: obj:`str`
```
KaiSzuttor: ```name : obj:`str` ```

src/python/espressomd/profiler.pyx
```diff
@@ -0,0 +1,31 @@
...
+        Name of the section
+
+    """
+
```
KaiSzuttor, line 14: remove blank

src/python/espressomd/profiler.pyx
```diff
@@ -0,0 +1,31 @@
...
+    c_profiler.begin_section(name)
```
KaiSzuttor, line 15: should we check if the profiler is available?
```
IF NOT HAVE_CALIPER:
    raise ValueError("No profiler available.")
```

fweik: It does nothing if it's not available, which is fine.

src/python/espressomd/profiler.pyx
fweik: Does this affect the result?

KaiSzuttor: idk

fweik: Is this style documented somewhere? It does not produce a sphinx warning not to have the space

KaiSzuttor: https://www.sphinx-doc.org/en/master/usage/extensions/example_numpy.html

KaiSzuttor: https://numpydoc.readthedocs.io/en/latest/format.html

fweik: Can this be enforced automatically? Otherwise we should not require it.

KaiSzuttor: I think no.

KaiSzuttor (approved):

Issue #2521, opened 2019-02-16
CI failures due to coverage collection


fweik:
We need a solution for the `lcov`-related issues like here: https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/75623
No idea what the cause of this is. Rerunning usually fixes the issue, but not always.

fweik:
The problem persists: https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/76701, there is now a failure in almost every pipeline.

jngrad:
Difficult to say since the container exits on failure before the artifacts get uploaded, so we cannot browse the files. Tinkering with lcov locally suggests the error message is raised when the gcno file is missing:

| gcda    | gcno           | stderr |
| :---: | :---: | :--- |
| present | present        |        |
| present | missing        | `geninfo: ERROR: $path/Variant_test.cpp.gcno: could not open file` |
| present | empty          | `geninfo: WARNING: empty $path/Variant_test.cpp.gcno (skipped)` |
| present | corrupt head | `geninfo: ERROR: $path/Variant_test.cpp.gcno: found unrecognized gcno file magic` |
| present | corrupt tail | `geninfo: ERROR: $path/Variant_test.cpp.gcno: reached unexpected end of file` |
| missing | present        |        |
| empty   | present        | `$path/Variant_test.cpp.gcda:cannot open data file, assuming not executed` |

We could open a PR with a modified `/maintainer/CI/build_cmake.sh` that prints the list of gcno files at multiple stages, wait for the first container to fail, then inspect the log file to see at which stage in the script the gcno files get erased. It might also be the compiler failed to generate the gcno files.

fweik:
I *think* the runners can be run locally for debugging.

mkuron:
While a job is running, you can even click the "Debug" button on its GitLab page and get a terminal in your browser. Just add a `sleep 10000` to the end of the build script first so the terminal session doesn't get killed when the build finishes.

fweik:
This is still a problem (https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/76169)

fweik:
If fixed, #2533 has to be reverted.

KaiSzuttor:
May it be that there is interaction with ccache?

PR #2533, opened 2019-02-25
maintainer: Don't fail on coverage failure

fweik:
Don't error-exit if reporting the coverage does fail. Provisional solution to #2521.


KaiSzuttor:
I included an alternative error ignoring in #2537

fweik:
Closed in favor of #2537.

PR #2536, opened 2019-02-26
Small coldet refactor

fweik:
@RudolfWeeber can you please check this, especially the changes in `place_vs_and_relate_to_particle`.
The collision handling function is very hard to understand, that should probably be refactored at some point.


fweik:
@RudolfWeeber if you could give me any hint why this does not work that would be helpful. It's entirely not clear to me...

RudolfWeeber (approved):

PR #2542, opened 2019-02-28
core: Removed duplication of pair bond force calc.

fweik:
Description of changes:
 - Removed code duplication of pair bond force calculation between force and pressure calc
- Don't add noise bond to the NPT virial pressure
- Consider all pair bonds when calculating pressure.


RudolfWeeber:
@fweik @jngrad, I added test coverage for bond stress and angle bond force and stress to this one
@jngrad, to my understanding, an angle of phi=0 means 3 particles on a line.
The force calculation is quite unreadable to me, but seems to be correct.
To simplify it, one might just use a cross-product on the two sides of the " triangel" to get the rotation axis
and a cross-product of the rotation axis and a side of the triangle to get the force direction.
That can then be multiplied with the scalar force expression

fweik:
@KaiSzuttor

testsuite/python/interactions_bond_angle.py
```diff
@@ -57,85 +63,105 @@ def rotate_vector(self, v, k, phi):
...
+            np.testing.assert_allclose(
+                np.sum(np.copy(self.system.part[0:3].f), 0), [0, 0, 0], atol=1E-12)
+
+            # No pressure
+            self.assertAlmostEqual(
+                self.system.analysis.pressure()["bonded"], 0, delta=1E-12)
+            # Stress tensor trace=0
+            self.assertAlmostEqual(
+                np.trace(self.system.analysis.stress_tensor()["bonded"]), 0, delta=1E-12)
+
```
KaiSzuttor, line 121: could you please add a comment here how to formula for ```p_tensor_expected``` looks like?

testsuite/python/interactions_bonded.py
```diff
@@ -137,33 +87,64 @@ def test_coulomb(self):
...
             # force equals minus the counter-force  ...
-            self.assertTrue((f0_sim == -f1_sim).all())
+            np.testing.assert_allclose(f0_sim, -f1_sim, 1E-12)
             # and has correct value.
-            f1_sim_copy = np.copy(f1_sim)
-            np.testing.assert_almost_equal(f1_sim_copy, f1_ref)
+            np.testing.assert_almost_equal(f1_sim, f1_ref)
+
+            # Pressure
+            p_expected = 1. / 3. * \
```
KaiSzuttor, line 134: Also here, please add a formula for the pressure in a comment

KaiSzuttor (approved):

jngrad:
@RudolfWeeber here's my problem with angle forces:
```python
import espressomd
import numpy as np
system = espressomd.System(box_l=[10.0, 10.0, 10.0])
system.cell_system.skin = 0.4
system.time_step = .1
# bent water molecule, 90 degrees angle, lying in the XY plane
system.part.add(id=0, pos=[1, 1, 0], type=0) # hydrogen
system.part.add(id=1, pos=[0, 0, 0], type=0) # oxygen
system.part.add(id=2, pos=[1, -1, 0], type=0) # hydrogen
# harmonic force to restore the 2*pi/3 angle
angle_harmonic = espressomd.interactions.AngleHarmonic(bend=100, phi0=np.pi/3)
system.bonded_inter.add(angle_harmonic)
system.part[0].add_bond((angle_harmonic, 1, 2))
# calculate forces
system.integrator.run(recalc_forces=True, steps=0)
forces = np.array([system.part[0].f, system.part[1].f, system.part[2].f])
np.round(forces)
```
outputs:
```
array([[  0, -13,   0],   # hydrogen
       [-13,  13,   0],   # oxygen
       [ 13,   0,   0]])  # hydrogen
```
The molecule should experience a restorative force that pushes the first atom up and the third atom down to increase the angle. Also they should be pushed to the left, while the second atom gets pushed to the right. I expect something like this:
```
array([[ -5,  13,   0],
       [ 10,   0,   0],
       [ -5, -13,   0]])
```
This is what Gromacs and Amber do.

fweik:
@KaiSzuttor I'm getting mixed messages here, can this be merged or not?

KaiSzuttor:
@fweik documentation of the pressure expressions is not yet done

fweik:
Why did you approve then? :-)

KaiSzuttor:
it's like ```if (fweik_did_it) can_be_merged=true;``` :smiley:

fweik:
@KaiSzuttor that is for @RudolfWeeber

RudolfWeeber:
@KaiSzuttor, the requested doc should be in. Can we go ahead, here?

KaiSzuttor (approved):

PR #2543, opened 2019-03-01
Remove thermalized bond center of mass feature

fweik:
Description of changes:
 - Removed the center-of-mass thermalization from the thermalized bond

This does not gain anything and causes problems with reintroducing Lees-
Edwards.

Depends of #2542


KaiSzuttor:
whats the status here?

RudolfWeeber:
It is related to the Drude complex. My current understanding is that the same effect cannot be achieved by the other version of thermalized bond, alone.

Some kind of virtual site implementation would be needed. I suggest to leave it open but defer it.

fweik:
Apparently we are unable to decide on this. Closing.

PR #2551, opened 2019-03-04
WIP: Re-introducing three point coupling

### LINKED ISSUE #2517: LB: three point coupling gives wrong results

KaiSzuttor:
Fixes #2517

Description of changes:
 - reimplementation of three-point interpolation
 - added tests


src/core/grid_based_algorithms/lb_interpolation.cpp
```diff
@@ -8,10 +9,37 @@
...
 #include "lbgpu.hpp"
 
+namespace {
+
+InterpolationOrder interpolation_order = InterpolationOrder::linear;
+}
+
+void mpi_set_interpolation_order_slave(int order, int) {
+  if (order == 0) {
+    interpolation_order = InterpolationOrder::linear;
```
fweik: You can do `interpolation_order = InterpolationOrder{order}` without the switch, then you don't have to hard-code the values.

src/core/grid_based_algorithms/lb_interpolation.cpp
```diff
@@ -8,10 +9,37 @@
...
+    interpolation_order = InterpolationOrder::quadratic;
+  }
+}
+
 #if defined(LB) || defined(LB_GPU)
 
-namespace {
+void lb_lbinterpolation_set_interpolation_order(
+    InterpolationOrder const &interpolation_order) {
+  if (interpolation_order == InterpolationOrder::linear) {
```
fweik: I think it is cleaner to just do
```c++
switch(interpolation_order) {
  case InterpolationOrder::linear:
  case InterpolationOrder::quadratic:
    mpi_call(interpolation_order_slave, static_cast<int>(interpolation_order));
    mpi_set_interpolation_order_slave(static_cast<int>(interpolation_order), 0);
}
```
or skip the switch if the value checking is not needed. This way you can avoid hard-coding the values, and a compiler warning is emitted if one of the cases is not handled.

src/core/grid_based_algorithms/lb_interpolation.cpp
```diff
@@ -72,12 +100,21 @@ lb_lbinterpolation_get_interpolated_velocity_global(const Vector3d &pos) {
   if (lattice_switch & LATTICE_LB_GPU) {
 #ifdef LB_GPU
     Vector3d interpolated_u{};
-    lb_get_interpolated_velocity_gpu(folded_pos.data(), interpolated_u.data(),
-                                     1);
+    if (interpolation_order == InterpolationOrder::linear) {
```
fweik: `switch(interpolation_order)`

src/core/grid_based_algorithms/lb_interpolation.cpp
```diff
@@ -72,12 +100,21 @@ lb_lbinterpolation_get_interpolated_velocity_global(const Vector3d &pos) {
...
+                                       1, false);
+    } else if (interpolation_order == InterpolationOrder::quadratic) {
+      lb_get_interpolated_velocity_gpu(folded_pos.data(), interpolated_u.data(),
+                                       1, true);
+    }
     return interpolated_u;
 #endif
   } else if (lattice_switch & LATTICE_LB) {
 #ifdef LB
+    if (interpolation_order == InterpolationOrder::quadratic) {
```
fweik: `switch(interpolation_order)`

src/core/grid_based_algorithms/lb_interpolation.cpp
```diff
@@ -92,6 +129,10 @@ lb_lbinterpolation_get_interpolated_velocity_global(const Vector3d &pos) {
 #ifdef LB
 void lb_lbinterpolation_add_force_density(const Vector3d &pos,
                                           const Vector3d &force_density) {
+  if (interpolation_order == InterpolationOrder::quadratic) {
```
fweik: `switch(interpolation_order)`

src/core/grid_based_algorithms/lb_particle_coupling.cpp
```diff
@@ -168,13 +168,26 @@ void lb_lbcoupling_calc_particle_lattice_ia(bool couple_virtual) {
   ESPRESSO_PROFILER_CXX_MARK_FUNCTION;
   if (lattice_switch & LATTICE_LB_GPU) {
 #ifdef LB_GPU
-    if (lb_particle_coupling.couple_to_md && this_node == 0)
-      lb_calc_particle_lattice_ia_gpu(couple_virtual,
-                                      lb_lbcoupling_get_gamma());
+    if (lb_particle_coupling.couple_to_md && this_node == 0) {
+      if (lb_lbinterpolation_get_interpolation_order() ==
```
fweik: `switch(interpolation_order)`

src/python/espressomd/lb.pyx
```diff
@@ -174,6 +174,20 @@ cdef class HydrodynamicInteraction(Actor):
 
             return self._params
 
+        def set_interpolation_order(self, interpolation_order):
+            """ Set the order for the fluid interpolation scheme.
+
+            Parameters
+            ----------
+            interpolation_order : :obj:`int`
+                0 refers to linear interpolation, 1 to quadratic interpolation.
```
fweik: Why not directly use strings for the interpolation order?

src/core/grid_based_algorithms/lbgpu_cuda.cu
```diff
@@ -1740,87 +1747,19 @@ __device__ void calc_viscous_force(LB_nodes_gpu n_a, float *delta,
...
-      (delta[7] * delta_j[0]));
-  atomicAdd(
-      &(node_f.force_density[(1) * para->number_of_nodes + node_index[7]]),
-      (delta[7] * delta_j[1]));
-  atomicAdd(
-      &(node_f.force_density[(2) * para->number_of_nodes + node_index[7]]),
-      (delta[7] * delta_j[2]));
+                                LB_node_force_density_gpu node_f,
+                                bool three_point_coupling) {
+  int max_node = 8;
```
fweik: This information should be encoded into the type of `node_index` and `delta`, this is very brittle...

src/core/grid_based_algorithms/lbgpu_cuda.cu
```diff
@@ -2307,12 +2246,13 @@ __global__ void
 calc_fluid_particle_ia(LB_nodes_gpu n_a, CUDA_particle_data *particle_data,
                        float *particle_force, LB_node_force_density_gpu node_f,
                        LB_rho_v_gpu *d_v, bool couple_virtual,
-                       uint64_t philox_counter, float friction) {
+                       uint64_t philox_counter, float friction,
+                       float *lb_boundary_velocity, bool three_point_coupling) {
```
fweik: Since there are only two coupling schemes, and this code is performance relevant, I think the coupling scheme should be selected statically and code for both schemes without the branches should be generated. E.g. make the coupling scheme a template parameter and instantiate both versions of this kernel.

fweik (changes requested):

src/core/grid_based_algorithms/lb_interpolation.cpp
KaiSzuttor: this is c++17

fweik: Ah yeah sorry, you have to do `static_cast<InterpolationOrder>(order)`.

KaiSzuttor: what about ```interpolation_order = static_cast<InterpolationOrder>(order);```?

KaiSzuttor: this is probably not such a good idea:
https://gcc.godbolt.org/z/Re_etN

fweik: That is fine here because the value is generated from an enum.

fweik: If you don't want that you can send the enum value via mpi, they can be serialized...

KaiSzuttor:
@fweik

src/core/grid_based_algorithms/lb_interpolation.cpp
```diff
@@ -1,17 +1,39 @@
...
+  boost::mpi::broadcast(comm_cart, interpolation_order, 0);
+}
+
 #if defined(LB) || defined(LB_GPU)
 
-namespace {
+void lb_lbinterpolation_set_interpolation_order(
+    InterpolationOrder const &order) {
+  interpolation_order = order;
+  boost::mpi::broadcast(comm_cart, interpolation_order, 0);
```
fweik: The mpi calls need to be the other way around, this is a deadlock.

fweik (changes requested):

src/core/grid_based_algorithms/lb_interpolation.cpp
KaiSzuttor: Could you please explain?

fweik: You are blocking on ` boost::mpi::broadcast` before the callback is invoked.

fweik (approved):

RudolfWeeber:
I think, this pr was merged prematurely, I'm afraid.

This PR decreased the performance of the 2pt coupling by a factor of 3, 
I now have a time step of 1.2 sec instead of 0.4 sec for 500k particles.

If the reason for the drag cannot be removed, the changes to the 2pt coupling need to be reverted.

PR #2558, opened 2019-03-06
Device array

### LINKED ISSUE #2559: LB GPU coupling performance needs to be improved
@RudolfWeeber please provide a benchmark.

KaiSzuttor:
Array type that can be used in CUDA device code

Closes #2559.


src/core/grid_based_algorithms/lbgpu_cuda.cu
```diff
@@ -1408,11 +1408,12 @@ __device__ __inline__ float three_point_polynomial_larger_than_half(float u) {
 }
 
 template <std::size_t no_of_neighbours,
```
fweik: Now `no_of_neighbours` no longer needs to be a template parameter, and you can rely on function overloads instead of enable_if

src/core/utils/Array.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+#if defined(__CUDACC__)
+#define DEVICE_QUALIFIER __host__ __device__
+#else
+#define DEVICE_QUALIFIER
+#endif
+
+namespace Utils {
+
+template <typename T, std::size_t N> struct Array {
```
fweik, line 14: The array looks good to me.

KaiSzuttor:
@fweik do you know why non-constant access to the array cannot be constexpr?

fweik:
> The throw-expression is classified as prvalue expression of type `void`.

KaiSzuttor:
https://stackoverflow.com/questions/34280729/throw-in-constexpr-function

src/core/utils/Array.hpp
```diff
@@ -0,0 +1,70 @@
...
+  DEVICE_QUALIFIER constexpr iterator begin() { return &m_data[0]; };
+  DEVICE_QUALIFIER constexpr const_iterator cbegin() const {
+    return &m_data[0];
+  };
+  DEVICE_QUALIFIER constexpr iterator end() { return &m_data[N]; };
+  DEVICE_QUALIFIER constexpr const_iterator cend() const { return &m_data[N]; };
+  DEVICE_QUALIFIER constexpr bool empty() const { return size() == 0; }
+  DEVICE_QUALIFIER constexpr size_type size() const { return N; }
+  DEVICE_QUALIFIER constexpr size_type max_size() const { return N; }
+  DEVICE_QUALIFIER void fill(const value_type &value) {
```
fweik, line 62: This doen't need to be a member function

KaiSzuttor: No, but it is in ```std::array``` and I want to be as compatible to the interface as possible

fweik:
The rocm build seems to be still building c++11. maybe we should also do verbose make in the CI

src/core/utils/Array.hpp
KaiSzuttor: But ```fill``` could be ```constexpr```, right?

hmenke: …and `noexcept(std::is_nothrow_assignable<value_type &, const value_type &>::value)`

src/core/utils/Array.hpp
```diff
@@ -0,0 +1,78 @@
...
+  }
+  DEVICE_QUALIFIER constexpr reference operator[](size_type i) {
+    return m_data[i];
+  }
+  DEVICE_QUALIFIER constexpr const_reference operator[](size_type i) const {
+    return m_data[i];
+  }
+  DEVICE_QUALIFIER constexpr reference front() { return *begin(); }
+  DEVICE_QUALIFIER constexpr const_reference front() const { return *cbegin(); }
+  DEVICE_QUALIFIER constexpr reference back() {
```
fweik, line 43: Can you please check if 
```c++
std::enable_if<N > 0, reference> back() { ... }
```
works? The you could also remove the data member for `N == 0`.

fweik: Ah, for this to work back hast to be a template...

KaiSzuttor: ```warning: zero size arrays are an extension```

fweik: Yeah, the really absolutely proper way to to it is to put into a base class and specialize that, which then also gives you empty base optimization. Like so (warninge: not tested):

```c++
template<class T, size_t N>
struct array_storage {
  T m_data[N];
};

template<class T>
struct array_storage<T, 0> {};

template<class T, size_t N>
class Array : private array_storage<T, N> {
...
};
```

src/core/utils/Array.hpp
```diff
@@ -0,0 +1,78 @@
...
+  DEVICE_QUALIFIER constexpr size_type size() const { return N; }
+  DEVICE_QUALIFIER constexpr size_type max_size() const { return N; }
+  DEVICE_QUALIFIER void fill(const value_type &value) {
+    for (size_type i = 0; i < size(); ++i)
+      m_data[i] = value;
+  }
+  DEVICE_QUALIFIER static constexpr Array<T, N>
+  broadcast(const value_type &value) {
+    Array<T, N> ret{};
+    for (size_type i = 0; i < N; ++i) {
```
fweik, line 69: Isn't that what `fill` does?

KaiSzuttor: void functions cannot be constexpr since the return type has to be a LiteralType

RudolfWeeber:
This PR indeed seems to remove the performance penalty introduced with the changed lb gpu particle coupling.

src/core/utils/Array.hpp
KaiSzuttor: This actually does not work if we want to preserve aggregate initialization.

fweik: Yes, but you can make it a member.

KaiSzuttor:
close due to #2561

PR #2561, opened 2019-03-07
Vector array inheritance

KaiSzuttor:
contains #2558


src/core/unit_tests/Array_test.cpp
```diff
@@ -79,3 +82,27 @@ BOOST_AUTO_TEST_CASE(broadcast) {
   static_assert(a[1] == 5, "");
   static_assert(a[2] == 5, "");
 }
+
+BOOST_AUTO_TEST_CASE(serialization) {
```
fweik, line 87: Why do you use MPI here?

KaiSzuttor: To test serialization... of course I could also just serialize and deserialize it

KaiSzuttor:
@fweik

fweik (approved):

PR #2562, opened 2019-03-08
Fix nightly builds

mkuron:
Random123 by default refuses to run on ARM and S390 architectures, but supposedly works. Also, it requires a `#define` to generate 64-bit random numbers on 32-bit architectures. Finally, we move the i386 build to the non-nightly builds because it is not emulated and completes within a few minutes.


src/core/grid_based_algorithms/lb.cpp
```diff
@@ -50,6 +50,7 @@
 #include "utils/uniform.hpp"
 #include "virtual_sites/lb_inertialess_tracers.hpp"
 
+#define R123_USE_MULHILO64_C99 1
```
KaiSzuttor: this should go to random.hpp

mkuron: Why random.hpp? It needs to be defined before including the Philox header. For lb_particle_coupling.cpp, it actually has to go right where it is because the Philox header is included before anything else.

fweik: If this should be used everywhere with philox you could also put a define on the cmake target, instead of putting it into a header.

PR #2567, opened 2019-03-08
Callback refactor

fweik:
Description of changes:
 - This allows mpi callbacks with (almost) arbitraty arguements
- Simplified some of the  callbacks

There are lots of callbacks that can be simplified with this.


KaiSzuttor:
maybe a stupid question... why donwe have this master slave model at all? cant we have a python interpreter running on all nodes and let each instance call collective functions in the core?

fweik:
Well this is how it is.

KaiSzuttor:
i see

fweik:
It's there for historic reasons, this is just how the chose to implement it. If one wanted to change that the had to be a well defined parallel interface. This could then be either directly exposed e.g. to python, or via the script interface.

KaiSzuttor (approved):

PR #2568, opened 2019-03-08
LB: encapsulate defines into enums

### LINKED ISSUE #2556: LB: lattice_switch is ambiguous
The global variable `lattice_switch` is currently defined as:
https://github.com/espressomd/espresso/blob/0682e84736bd6d8b0131078a0551d35dd1a920c6/src/core/grid_based_algorithms/lattice.hpp#L34-L43

And it is used like this:
https://github.com/espressomd/espresso/blob/0682e84736bd6d8b0131078a0551d35dd1a920c6/src/core/grid_based_algorithms/lb_particle_coupling.cpp#L48-L59

A few issues here:
* the numerical values are not connected to the variable, causing documentation decay (the Doxygen blocks are already outdated)
* the `#ifdef` statements seem redundant with the global variable; shouldn't writing `lattice_switch = LATTICE_LB_GPU` when LB_GPU isn't compiled in raise an exception?
* is the statement `lattice_switch = LATTICE_LB & LATTICE_LB_GPU` legal (documentation says yes) and if so, in a switch, when should LB CPU evaluation happen before LB GPU evaluation?

jngrad:
Hard-coded integer values and macros are confusing, not logically connected to each other and prone to accidental redefines and documentation decay. `enum class` puts everything in scopes. Fixes #2556

### DIFF
--- a/src/core/communication.cpp
+++ b/src/core/communication.cpp
@@ -873,17 +873,17 @@ void mpi_sync_topo_part_info_slave(int, int) {
 
 /******************* REQ_BCAST_LBPAR ********************/
 
-void mpi_bcast_lb_params(int field, int value) {
+void mpi_bcast_lb_params(LBParam field, int value) {
 #ifdef LB
-  mpi_call(mpi_bcast_lb_params_slave, field, value);
-  mpi_bcast_lb_params_slave(field, value);
+  mpi_call(mpi_bcast_lb_params_slave, (int)field, value);
+  mpi_bcast_lb_params_slave((int)field, value);
 #endif
 }
 
 void mpi_bcast_lb_params_slave(int field, int) {
 #ifdef LB
   MPI_Bcast(&lbpar, sizeof(LB_Parameters), MPI_BYTE, 0, comm_cart);
-  lb_lbfluid_on_lb_params_change(field);
+  lb_lbfluid_on_lb_params_change((LBParam)field);
 #endif
 }
 
--- a/src/core/communication.cpp
+++ b/src/core/communication.cpp
@@ -875,15 +875,15 @@ void mpi_sync_topo_part_info_slave(int, int) {
 
 void mpi_bcast_lb_params(LBParam field, int value) {
 #ifdef LB
-  mpi_call(mpi_bcast_lb_params_slave, (int)field, value);
-  mpi_bcast_lb_params_slave((int)field, value);
+  mpi_call(mpi_bcast_lb_params_slave, static_cast<int>(field), value);
+  mpi_bcast_lb_params_slave(static_cast<int>(field), value);
 #endif
 }
 
 void mpi_bcast_lb_params_slave(int field, int) {
 #ifdef LB
   MPI_Bcast(&lbpar, sizeof(LB_Parameters), MPI_BYTE, 0, comm_cart);
-  lb_lbfluid_on_lb_params_change((LBParam)field);
+  lb_lbfluid_on_lb_params_change(static_cast<LBParam>(field));
 #endif
 }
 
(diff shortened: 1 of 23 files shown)


KaiSzuttor:
maybe you want so apply your changes onto #2566

src/core/communication.cpp
```diff
@@ -873,17 +873,17 @@ void mpi_sync_topo_part_info_slave(int, int) {
 
 /******************* REQ_BCAST_LBPAR ********************/
 
-void mpi_bcast_lb_params(int field, int value) {
+void mpi_bcast_lb_params(LBParam field, int value) {
 #ifdef LB
-  mpi_call(mpi_bcast_lb_params_slave, field, value);
-  mpi_bcast_lb_params_slave(field, value);
+  mpi_call(mpi_bcast_lb_params_slave, (int)field, value);
```
KaiSzuttor: please use ```static_cast```

src/core/forces.cpp
```diff
@@ -154,7 +154,7 @@ void force_calc() {
 
 #if defined(LB_GPU) || defined(LB)
 #ifdef LB
-  if (lattice_switch & LATTICE_LB) {
+  if (lattice_switch == ActiveLB::CPU) {
```
KaiSzuttor, line 157: as in #2566 id prefer to move the enum to the LB interface and only expose via an interfacefunction

src/core/grid_based_algorithms/lb_constants.hpp
```diff
@@ -0,0 +1,43 @@
...
+ *  Determine what actions have to take place upon change of the respective
+ *  parameter.
+ */
+enum class LBParam {
+  DENSITY,   /**< fluid density */
+  VISCOSITY, /**< fluid kinematic viscosity */
+  AGRID,     /**< grid constant for fluid lattice */
+  TAU,       /**< time step for fluid propagation */
+  /** friction coefficient for viscous coupling between particles and fluid */
+  FRICTION,
```
KaiSzuttor: friction is no longer part of the LB but only the coupling

src/core/grid_based_algorithms/lb_interface.cpp
```diff
@@ -416,43 +416,30 @@ double lb_lbfluid_get_tau() {
...
-    break;
-  case 2:
-    lattice_switch = LATTICE_LB_GPU;
-    mpi_bcast_parameter(FIELD_LATTICE_SWITCH);
-    break;
-  default:
-    throw std::invalid_argument("Invalid lattice switch.");
-  }
+  assert(local_lattice_switch >= 0 && local_lattice_switch <= 2);
+  lattice_switch = (ActiveLB)local_lattice_switch;
```
KaiSzuttor: ```static_cast```

KaiSzuttor (changes requested):

src/core/grid_based_algorithms/lb_constants.hpp
jngrad: both FRICTION and TAU aren't used anywhere else in the code; should I just delete them?

fweik: Yes

fweik:
In general, to get the full benefit of enums, they should be used with switch statements. E.g. then a compiler warning is issued if not all cases are handled.

src/core/communication.cpp
jngrad: done

src/core/grid_based_algorithms/lb_constants.hpp
jngrad: done

src/core/grid_based_algorithms/lb_interface.cpp

jngrad:
This PR has been transferred to fweik/espresso#9

fweik:
@jngrad that is already merged.

KaiSzuttor:
moved to #2576

PR #2570, opened 2019-03-09
Fix Random123 on architectures that do not provide 128-bit integers

mkuron:
Follow-up to #2562.

[s390x](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/82214) and [arm64](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/82210) were still failing. This pull request unconditionally enables `R123_USE_MULHILO64_C99`, which is used as a fallback if neither assembly instructions, intrinsics, nor 128-bit integers are available.


fweik:
Ah, I misunderstood the flag.

fweik (approved):

PR #2576, opened 2019-03-12
Lattice

### LINKED ISSUE #2556: LB: lattice_switch is ambiguous

KaiSzuttor:
edit: fixes #2556


src/core/communication.cpp
```diff
@@ -866,17 +866,17 @@ void mpi_sync_topo_part_info_slave(int, int) {
 
 /******************* REQ_BCAST_LBPAR ********************/
 
-void mpi_bcast_lb_params(int field, int value) {
+void mpi_bcast_lb_params(LBParam field, int value) {
 #ifdef LB
-  mpi_call(mpi_bcast_lb_params_slave, field, value);
-  mpi_bcast_lb_params_slave(field, value);
+  mpi_call(mpi_bcast_lb_params_slave, static_cast<int>(field), value);
```
KaiSzuttor, line 871: why do you cast? you can directly broadcast the enum class

jngrad: an enum class doesn't implicitly cast to an int, therefore we get `cannot convert ‘LBParam’ to ‘int’ for argument ‘2’ to ‘void mpi_call(void (*)(int, int), int, int)’`

fweik:
@jngrad please resolve the merge conflict.

jngrad:
I had the same issue when merging into fweik's branch. Not sure why git is making things so difficult.

src/core/grid_based_algorithms/lb_interface.hpp
```diff
@@ -58,14 +54,19 @@ void lb_lbfluid_set_rng_state(uint64_t counter);
 const Lattice &lb_lbfluid_get_lattice();
 
 /**
- * @brief Get the global variable lattice_switch which defines wether NONE, CPU
- * or GPU LB is active.
+ * @brief Get the global variable @ref lattice_switch.
+ *
+ * Function to be called from Cython.
+ */
+int lb_lbfluid_get_lattice_switch_cython();
```
fweik: Why is this function needed? scoped enums can be used in cython, please deal with cython stuff in the cython code.

src/core/grid_based_algorithms/lb_interface.cpp
```diff
@@ -1274,35 +1261,39 @@ void lb_lbnode_set_pop(const Vector3i &ind, const Vector19d &p_pop) {
...
+ActiveLB lb_lbfluid_get_lattice_switch() { return lattice_switch; }
+
+int lb_lbfluid_get_lattice_switch_cython() {
+  return static_cast<int>(lattice_switch);
+}
 
-void lb_lbfluid_on_lb_params_change(int field) {
-  if (field == LBPAR_AGRID) {
+void lb_lbfluid_on_lb_params_change(LBParam field) {
+  if (field == LBParam::AGRID) {
```
fweik: `switch(field)` ...

fweik (changes requested):

src/core/grid_based_algorithms/lb_interface.cpp

src/core/grid_based_algorithms/lb_interface.hpp
jngrad: I already looked for a better alternative last week, but Cython support for exported C enums is limited:
* `cdef enum ActiveLB: NONE = 0, CPU, GPU` in .pxd, then in .pyx:
   * `fun(ActiveLB.CPU)` -> `error: ‘CPU’ was not declared in this scope: fun(CPU)`
   * `fun(<ActiveLB> ActiveLB.CPU)` -> `error: ‘CPU’ was not declared in this scope: fun((enum ActiveLB)CPU)`
* `cdef enum enum_ActiveLB "ActiveLB": NONE = 0, CPU, GPU` in .pxd, then in .pyx:
   * `fun(enum_ActiveLB.CPU)` -> `error: ‘CPU’ was not declared in this scope: fun(CPU)`
* same observations with `ctypedef` instead of `cdef`
* there is a namespace for pure Python enums, though

Is the new solution with two enums (f009b72) satisfactory?

fweik: You can use them like this (example from the LB interface):

```cython
      cdef cppclass InterpolationOrder:
            pass

    cdef extern from "grid_based_algorithms/lb_interpolation.hpp" namespace "InterpolationOrder":
        cdef InterpolationOrder linear
        cdef InterpolationOrder quadratic
```

I think this should also work for return values.

jngrad: Ok, it should work now. It's kind of annoying the exported enum values cannot be grouped in a Python namespace (`cdef ActiveLB ActiveLB_CPU "CPU"` didn't work).

fweik: Yes, it's not nice. Did you see https://github.com/cython/cython/issues/1603#issuecomment-324602083 ?

jngrad: Saw it last week, but forgot to try it... Let's see if all compilers accept it.

jngrad: Looks like a no on Cython 0.23.4 (excepted container `empty`), but a yes on 0.25.2 and above. Should I just revert the last commit?

fweik: Yes, just revert to the solution without the namespace.

fweik (approved):

PR #2577, opened 2019-03-12
core: Moved obstrvable struct definitions to sep file

fweik:
These need to be included to add energy or pressure contributions.

The corresponding functions should be cleaned up too, but is out of scope for me.


KaiSzuttor (commented):
approved

PR #2582, opened 2019-03-13
Fix LB sample syntax

KaiSzuttor:


jngrad:
Thanks! But many samples are broken, eg. `/samples/ekboundaries.py`, `/samples/espresso_logo.py`, etc., with no easy way to detect that. Many were fixed and guarded by unit tests in #2452. We can keep backporting and forwardporting fragments of commits in #2452, one PR at at time (see #2571, #2433, and now this one), until #2452 becomes full of redundancies and merge conflicts. This is not sustainable.

pkreissl (approved):
Merge depending on the outcome of #2584 deciding the fate of these samples.

KaiSzuttor:
No, this can be merged independently #2584

PR #2583, opened 2019-03-13
WIP: Neighbor search

fweik:
- Abstraction for finding particles within a radius of a point.
- Made reaction ensemble faster

This should also be rolled out to

- [ ] collision detection
- [ ] other places I forgot

This final goal here is to make the implementation and storage
of the cells an implementation detail of the cell system, which is
not visible from the outside.


src/core/cells.cpp
```diff
@@ -67,6 +73,122 @@ double max_range = 0.0;
...
+/**
+ * @brief Search the neighborhood of x for particles
+ *        within a sphere around a point
+ *
+ * Search is limited up to the maxrange of the cell system,
+ * so it only find particles close than max range.
+ *
+ * @param x Point at which so search
+ * @param r Search radius.
+ * @return  List of particles nearby.
```
KaiSzuttor, line 85: List of particle ids

src/core/cells.cpp
```diff
@@ -68,6 +73,122 @@ double max_range = 0.0;
...
+
+  return accumulate(
+      search_cell->neighbors().all(), nearby(search_cell, {}),
+      [&](std::vector<int> parts, Cell *n) { return nearby(n, parts); });
+}
+
+void cells_find_nearby_particles_slave(int pnode, int) {
+  if (comm_cart.rank() == pnode) {
+    std::pair<Vector3d, double> params;
+    comm_cart.recv(0, 52, params);
```
jonaslandsgesell: Could we introduce a macro here instead of using a plain magic number 52? It would be more readable.

fweik: And what would you call the 'macro'? `NUMBER`?

jonaslandsgesell: Hehe, sorry i just saw that the magic number is the tag which is used for the MPI communication. If we would introduce a name here it would probably be function specific... So probably just forget the comment

RudolfWeeber:
This is a good addition, but lacks a test. 
Apart from a unit test (which would require some mocking), an alternative would be to use it in a feature, where the outcome is more directly dependent on correctly finding the pairs (collision detection or cluster analysis, maybe).

PR #2585, opened 2019-03-13
4.0.2rc

RudolfWeeber:
Description of changes:
 - 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


fweik:
This should be merged into the 4.0 branch, not 4.0.1-release, no?

RudolfWeeber:
On Thu, Mar 14, 2019 at 10:47:28AM +0000, Florian Weik wrote:
> This should be merged into the 4.0 branch, not 4.0.1-release, no?
I had no intention of merging. I picked the base branch such that
the diff makes sense and can be checked.

RudolfWeeber:
@jngrad could you please give the diff between 4.0.1 and this branch a final check and then set the release process in motion?
You can also make a decision about still including the visualizer corrections or deferring them for the next time. They will have to be tested manually in the 4.0.2 branch.

mkuron:
I have confirmed that #2751 and #2677 are safe to cherry-pick.

fweik:
@jngrad did you push to https://github.com/espressomd/espresso/tree/upstream/release-4.0.2 on purpose? Also if some of the ci pipelines are not actually needed please cancel them. They are running the emulated tests, an clog up the CI for hours.

jngrad:
I know, will `git push upstream :upstream/release-4.0.2 release-4.0.2` fix it?

fweik:
I *think* you can only push from local to remote, not remote to remote. But I guess you can just rename the branch on github?

jngrad:
The GitHub interface only enable branch deletion, not renaming. Solved it by deleting then pushing again with the right name.

fweik:
@KaiSzuttor @RudolfWeeber so far we only had branches for the minor versions, not sure how we want to go about this. I would have put the 4.0.2 on the 4.0 branch.

RudolfWeeber:
On Fri, Apr 26, 2019 at 02:35:44PM +0000, Florian Weik wrote:
> @KaiSzuttor @RudolfWeeber so far we only had branches for the minor versions, not sure how we want to go about this. I would have put the 4.0.2 on the 4.0 branch.
For 4.0.1, we made a branch, I think.
If I checkout 4.0, that's what I'd expect to get.

mkuron:
> so far we only had branches for the minor versions, not sure how we want to go about this. I would have put the 4.0.2 on the 4.0 branch.

I agree, but for some reason @KaiSzuttor insisted on keeping the 4.0 branch at 4.0.0 and introducing a branch release-4.0.1 which pointed to the same commit as the v4.0.1 tag. Common practice is that branches should be used for minor versions and tags for patch releases. That would have enabled useful things like allowing other people to post pull request against the release branch...

mkuron:
> If I checkout 4.0, that's what I'd expect to get.

Actually, you should expect to get the latest 4.0.x release plus any further bug fixes that have been merged to that branch. If you check out the 4.0.0 tag, you would get precisely that release.

fweik:
I agree with @mkuron, also I think in the end we did put 4.0.1 into the 4.0 branch (see https://github.com/espressomd/espresso/branches, no 4.0.1 branch).

jngrad:
@fweik there is a 4.0.1 branch: https://github.com/espressomd/espresso/branches/stale

fweik:
Ok, I still think that keeping patch level branches around is odd, after all the patch release should be compatible, and the previous version contains known bugs (hence the bugfix release), and should not be used anymore.

RudolfWeeber:
> Ok, I still think that keeping patch level branches around is odd, after all the patch release should be compatible, and the previous version contains known bugs (hence the bugfix release), and should not be used anymore.
OK, let's implement @mkuron's suggestion for 4.1. Hopefully, this will be the last 4.0.x release.

PR #2586, opened 2019-03-13
core: Header cleanup

fweik:


fweik:
@KaiSzuttor https://travis-ci.org/espressomd/espresso/jobs/505903357 any idea? I thought the LB was supposed to be deterministic now.

KaiSzuttor (approved):

PR #2595, opened 2019-03-14
Vector

fweik:
- removed no longer needed `Vector::dot`
- renamed `Vector::cross` to `vector_product` and made a free function
- constexpr constructors from c-arrays and std::initializer_list
- constexpr relational operators

### DIFF
--- a/src/core/utils/Vector.hpp
+++ b/src/core/utils/Vector.hpp
@@ -31,23 +31,23 @@
 
 #include "utils/Array.hpp"
 
-template <typename Scalar, std::size_t n>
-class Vector : public Utils::Array<Scalar, n> {
+template <typename T, std::size_t n>
+class Vector : public Utils::Array<T, n> {
 public:
-  using Utils::Array<Scalar, n>::at;
-  using Utils::Array<Scalar, n>::operator[];
-  using Utils::Array<Scalar, n>::front;
-  using Utils::Array<Scalar, n>::back;
-  using Utils::Array<Scalar, n>::data;
-  using Utils::Array<Scalar, n>::begin;
-  using Utils::Array<Scalar, n>::cbegin;
-  using Utils::Array<Scalar, n>::end;
-  using Utils::Array<Scalar, n>::cend;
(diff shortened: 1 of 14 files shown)


src/core/utils/Vector.hpp
```diff
@@ -287,6 +285,12 @@ template <size_t N, typename T> Vector<T, N> sqrt(Vector<T, N> const &a) {
   return ret;
 }
 
+template <class T>
+Vector<T, 3> vector_product(const Vector<T, 3> &a, const Vector<T, 3> &b) {
```
KaiSzuttor: constexpr

src/core/utils/Vector.hpp
```diff
@@ -287,6 +287,12 @@ template <size_t N, typename T> Vector<T, N> sqrt(Vector<T, N> const &a) {
...
```
KaiSzuttor: east const :fist:

fweik: polyconst braclets

fweik: We should decide on a convention though. I think `west const` is more what people are used to?

KaiSzuttor: there are many opinions on that: I'm in favor of east since ```const``` modifies the type on it's left unless there is none

fweik: With west const the rule is `const modifies what is on its left`

fweik: With `east const` the rule is `const modifies what is on its left`, if there is nothing, it's not `east const`. I like this argument for `east const`: [A foolish consistency](http://slashslash.info/2018/02/a-foolish-consistency/). The core guidline [say](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#nl26-use-conventional-const-notation) `west const`, but with weak arguments.

fweik:
@KaiSzuttor please review, I made everything east const just for you.

KaiSzuttor (approved):

PR #2598, opened 2019-03-15
Grid refactor

KaiSzuttor:


src/python/espressomd/lb.pyx
```diff
@@ -324,7 +324,7 @@ IF LB or LB_GPU:
             self.node[0] = key[0]
             self.node[1] = key[1]
             self.node[2] = key[2]
-            if not lb_lbnode_is_index_valid(self.node):
+            if not lb_lbnode_is_index_valid(<const Vector3i&>self.node):
```
fweik: Why is this needed ? `Vector3i` can bind to `Vector3i const &` (but not to `Vector3i const`)

src/python/espressomd/lb.pyx
```diff
@@ -371,7 +371,7 @@ IF LB or LB_GPU:
         property population:
             def __get__(self):
                 cdef Vector19d double_return
-                double_return = lb_lbnode_get_pop(self.node)
+                double_return = lb_lbnode_get_pop(<const Vector3i&> self.node)
```
fweik: ?

src/python/espressomd/lb.pyx
```diff
@@ -397,12 +397,12 @@ IF LB or LB_GPU:
                 cdef Vector19d double_return
                 for i in range(19):
                     double_return[i] = value[i]
-                lb_lbnode_set_pop(self.node, double_return)
+                lb_lbnode_set_pop(<const Vector3i&>self.node, double_return)
```
fweik: ?

src/python/espressomd/lb.pyx
```diff
@@ -397,12 +397,12 @@ IF LB or LB_GPU:
...
 
         property boundary:
             def __get__(self):
                 cdef int int_return
-                int_return = lb_lbnode_get_boundary(self.node)
+                int_return = lb_lbnode_get_boundary(<const Vector3i&>self.node)
```
fweik: ?

fweik (changes requested):

src/python/espressomd/lb.pyx
KaiSzuttor: I have no idea... i'm exploring the issue

KaiSzuttor: It seems like a bug in cython but I'm not sure what I have changed to trigger it

fweik: Did you look at the generated c++? This sometimes helps to understand what's going on.

KaiSzuttor: It's an error from cython directly not from the compiler

fweik: In part this is because your return types are `const` qualified

fweik:
I have a solution, if you did not fix it yet

fweik:
https://github.com/fweik/espresso/tree/grid_refactor

KaiSzuttor:
it had to do with some side-effects of the wildcard import of the globals but it's not clear to me yet

fweik:
In my branch it worked without changes on the imports or pxd.

KaiSzuttor:
seems like there is some UB in cython

fweik (approved):

PR #2604, opened 2019-03-15
Clang tidy(2)

KaiSzuttor:


fweik:
Please add the checks in cmake where possible so that it stays clean.

fweik (approved):

PR #2605, opened 2019-03-16
Doxygen cleanup

### LINKED ISSUE #2587: Make documentation of library interfaces visible
Several places in the code are guarded by `#ifdef SCAFACOS`, `#ifdef CUDA`, etc. These macros are categorized as either derived or external in `features.def`, so `gen_doxyconfig.py` ignores them, making several parts of the code invisible to Doxygen. For example, the `Scafacos Namespace Reference` page is mostly empty, and a CUDA method throws a Doxygen warning because it is implemented in a .cu file but its prototype declaration in the .hpp file is ignored due to `#ifdef CUDA`.

If there is no objection, I'll replace `defs.features` with `defs.allfeatures` in the following line:
https://github.com/espressomd/espresso/blob/318ffb5582d97f455ec9ad4699171226e99598c6/doc/doxygen/gen_doxyconfig.py#L62

jngrad:
Changes:
* Library interfaces are now visible in Doxygen (fixes #2587)
* Doxygen warnings:
   * some warnings were masking other warnings, there were ~30 warnings with doxygen-1.8.13 (Ubuntu 18) and 93 with doxygen-1.8.16 (latest GitHub version, has more sanity checks)
   * currently only 10 warnings remain, we probably won't go any lower than that:
      * 3 `member <NAME> belongs to two different groups. The second one found here will be ignored.`: happens when a function is declared in a group in the .hpp file but defined in another group in the .cpp file. This is usually caused by the "Private functions" and "Exported functions" groups in .hpp files.
      * 4 `documented symbol <NAME> was not declared or defined.`: known bug, fixed in 1.8.16
      * 3 `no uniquely matching class member found for <NAME>`: known bug, still not fixed
* Clang `-Wdocumentation` compiler warnings:
   * went from 2416 warnings to 0 (on maxset without CUDA)
   * only 61 of these warnings were unique, they propagated through `#include` recursively
   * this compiler option has limited support for Doxygen commands, so we won't add it to CMake (e.g. both `@param p1,p2 Details` and `@return @ref ES_OK` fail, see also [bug 17437](https://bugs.llvm.org/show_bug.cgi?id=17437))
* Parts of the code were changed without affecting the program logic:
   * fixed a mistake in two `runtime_error` messages (AutoParameters.hpp)
   * removed a duplicate function declaration written in the wrong file (bonded_tab.hpp)
   * used `ES_OK` and `ES_ERROR` instead of `0` and `1` when it made sense to do so (if macros were mixed with hard-coded values, or just to be consistent across related functions)
   * enforced constness in a few functions with mixed input/output pointer parameters of type `T*`: input is now `T const *const`, output is now `T *const`


jngrad:
# Coverxygen Report
Merging `#2605` into python will **increase** documentation coverage by 0.8%
```diff
@@                Coverage Diff                    @@
## Items          python   #2605    +/-   coverage ##
=====================================================
  Files            30.6%   30.6%   +0.0%   193/631
  Namespaces        2.9%    2.9%   +0.0%     2/69
  Typedefs          8.4%    9.2%   +0.8%    12/131
  Defines          23.7%   23.7%   +0.0%   123/519
  Unions          100.0%  100.0%   +0.0%     1/1
+ Structs          53.2%   55.2%   +1.9%    85/154
  Classes          23.0%   23.0%   +0.0%    44/191
    -> Variables   39.7%   40.2%   +0.4%   654/1628
    -> Friends     12.5%   12.5%   +0.0%     1/8
+ Enums            40.0%   45.0%   +5.0%     9/20
+   -> Values      50.7%   59.4%   +8.7%    82/138
  Functions        42.7%   43.1%   +0.4%  1661/3855
+   -> param       26.8%   28.1%   +1.3%  1567/5584
    -> tparam       4.3%    4.3%   +0.0%    17/392
=====================================================
+ Total            32.6%   33.4%   +0.8%  4451/13321
```

fweik (approved):

PR #2609, opened 2019-03-18
Fixed some issues found by clang-tidy-7

fweik:


src/core/cells.hpp
```diff
@@ -72,15 +72,18 @@
...
-#define CELL_STRUCTURE_NONEYET -1
-/** Flag indicating that the current cell structure will be used further on */
-#define CELL_STRUCTURE_CURRENT 0
-/** cell structure domain decomposition */
-#define CELL_STRUCTURE_DOMDEC 1
-/** cell structure n square */
-#define CELL_STRUCTURE_NSQUARE 2
-/** cell structure layered */
-#define CELL_STRUCTURE_LAYERED 3
+enum {
```
jngrad, line 71: Here the Doxygen syntax gets broken. The group description should become an enum description, e.g.:
```c++
/** Cell Structure */
enum {
  /** Flag indicating that there is no cell system yet. Only at the
   *  VERY beginning of the code startup.
   */
  CELL_STRUCTURE_NONEYET = -1,
  /** Flag indicating that the current cell structure will be used further on */
  CELL_STRUCTURE_CURRENT = 0,
  /** cell structure domain decomposition */
  CELL_STRUCTURE_DOMDEC = 1,
  /** cell structure n square */
  CELL_STRUCTURE_NSQUARE = 2,
  /** cell structure layered */
  CELL_STRUCTURE_LAYERED = 3
};
```

src/core/domain_decomposition.cpp
```diff
@@ -71,28 +61,11 @@ double max_skin = 0.0;
 /*@{*/
 
 /** Convenient replace for loops over all cells. */
```
jngrad: please remove the orphaned Doxygen lines here, and at lines 65, 67-68

src/core/electrostatics_magnetostatics/p3m-dipolar.cpp
```diff
@@ -1826,7 +1826,8 @@ static double dp3m_m_time(char **log, int mesh, int cao_min, int cao_max,
   /* the initial step sets a timing mark. If there is no valid r_cut, we can
      only try
      to increase cao to increase the obtainable precision of the far formula. */
-  do {
+  double tmp_time;
+  -do {
```
jngrad: extra minus sign

src/core/cells.hpp
```diff
@@ -67,20 +67,22 @@
...
+  CELL_STRUCTURE_CURRENT = 0,
+  /** cell structure domain decomposition */
+  CELL_STRUCTURE_DOMDEC = 1,
+  /** cell structure n square */
+  CELL_STRUCTURE_NSQUARE = 2,
+  /** cell structure layered */
+  CELL_STRUCTURE_LAYERED = 3
+};
+
 /*@}*/
```
jngrad: this line still needs to go, it is part of the old Doxygen group
```suggestion

```

src/core/MpiCallbacks.cpp
```diff
@@ -76,7 +76,12 @@ int MpiCallbacks::add(func_ptr_type fp) {
 
 void MpiCallbacks::remove(const int id) { m_callbacks.remove(id); }
 
-void MpiCallbacks::abort_loop() const { call(LOOP_ABORT, 0, 0); }
+void MpiCallbacks::abort_loop() const {
+  try {
+    call(LOOP_ABORT, 0, 0);
+  } catch (...) {
```
KaiSzuttor, line 82: why is that needed?

fweik: Because it's used in the destructor, which is not allowed to throw, but mpi can throw. When there is an error in MPI during shut down there is nothing we can do, so catch all should be fine.

KaiSzuttor (approved):

PR #2613, opened 2019-03-19
WIP utils: Make some math functions constexpr and mark them as side-effec…

fweik:
…t free.

`__attribute__((const))` means that the function is pure and side-effect free.

### DIFF
--- a/src/core/utils/attributes.hpp
+++ b/src/core/utils/attributes.hpp
@@ -1,0 +1,11 @@
+#ifndef UTILS_ATTRIBUTES_HPP
+#define UTILS_ATTRIBUTES_HPP
+
+/* __GNUC__ catches gcc, intel and clang,
+ * MSVS does not have the attribute */
+#if defined(__GNUC__)
+#define FUNCTION_ATTRIBUTE_CONST __attribute__((const))
+#else
+#define FUNCTION_ATTRIBUTE_CONST
+#endif
+#endif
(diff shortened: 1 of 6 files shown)


jngrad:
I would have expected the compiler to figure out this attribute by itself in the case of inline and template functions. But when compiling with `-Wsuggest-attribute=pure` and `-Wsuggest-attribute=const` to flag suitable candidates, the results surprised me at first glance:
```c++
template <typename T> double sqrlen(T const &v) { return scalar(v, v); }
```
is a candidate for pure, but
```c++
template <typename T> T sqr(T x) { return x * x; }
```
isn't, even with `-O3`. Looking at the documentation for attributes, this result is not so surprising, though. Should we use the const attribute in general, or only in situations where there is a gain in performance?

fweik:
I don't think that it will make a huge difference. I wouldn't expect the codegen to change for templates, but I haven't checked. There's a [c++ weekly episode](https://www.youtube.com/watch?v=8ZxGABHcu40) about this :-)

fweik:
I think this won't make a difference and is just confusing. I'll take it out. The change that I actually wanted was the one to `int_pow` to avoid the compiler warnings in debug mode.

jngrad:
Why won't gcc and clang throw a warning at this code?
```c++
int i = 0;
template <typename T> __attribute__((const)) T sqr(T x) { i += 1; return x * x; }
```

fweik:
It's a contract: you promise to the compiler that this function does not mutate global state (and therefore has referential transparency), the attribute is usefol exactly for situations where the implementation is not visible.

PR #2617, opened 2019-03-20
Span device qualifier

fweik:
and remove `CXX14_CONSTEXPR` macro


src/core/utils/device_qualifier.hpp
```diff
@@ -0,0 +1,16 @@
...
+#define UTILS_DEVICE_QUALIFIER_HPP
+
+#if defined(__CUDACC__)
+#define DEVICE_QUALIFIER __host__ __device__
+#else
+#define DEVICE_QUALIFIER
+#endif
+
+#ifdef __CUDACC__
+#define DEVICE_ASSERT(A) assert((A))
```
KaiSzuttor, line 9: Why not have all defines in one ```ifdef``` ?

fweik: Not the same logic

KaiSzuttor: it's binary logic that depends on ```__CUDACC__```, and both cases are handled at the top. What do I miss here?

fweik: one is `ifndef`

KaiSzuttor (commented):
good that you also think it's important :smile:

fweik:
@KaiSzuttor

KaiSzuttor (approved):

Issue #2618, opened 2019-03-20
Timeouts in the CI


fweik:
e.g. (times are from the CI)
```
 40/123 Test  #58: virtual_sites_tracers ...................   Passed  247.73 sec
[...]
 54/123 Test #122: lb_density ..............................   Passed  394.11 sec
```
with debug builds I regularly get timeouts (default timeout is 300s) even on a modern machine, this is too slow to be practical.

KaiSzuttor:
you mean the tests are too long or we should find out where the performance drop comes from... We knew before merging that LB will be slower

KaiSzuttor:
``` bash
82/123 Test  #66: lb ......................................   Passed   13.17 sec
```
with 

``` bash
cmake_params=-DCMAKE_BUILD_TYPE=Debug -DPYTHON_EXECUTABLE=/usr/bin/python2 -DWARNINGS_ARE_ERRORS=ON -DTEST_NP:INT=2  -DWITH_SCAFACOS=ON -DCMAKE_CXX_FLAGS=-Og
```

in https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/88590

How do you know that it is connected to the refactoring?

fweik:
I don't, but it is reasonable to assume that it is related? Could also some problem in the CI.

fweik:
looks like a CI problem, but the issue still needs to be addressed.

fweik:
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/88973 has failed due to timed out LB tests

KaiSzuttor:
It seems like whenever I run the tests, the issue is gone :disappointed: 
I don't have the time to run the tests for everybody, so that doesn't seem to be a solution

jngrad:
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/89108
```
 70/123 Test  #51: ek_eof_one_species_x ....................***Timeout 300.02 sec
 71/123 Test  #52: ek_eof_one_species_y ....................***Timeout 300.03 sec
 72/123 Test  #53: ek_eof_one_species_z ....................***Timeout 300.02 sec
121/123 Test #119: lb_thermostat ...........................***Timeout 300.02 sec
```

fweik:
Master is read again because of this:
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/pipelines/6362

KaiSzuttor:
[...]

PR #2622, opened 2019-03-21
Increase LB verbosity for CI debugging

KaiSzuttor:
See #2618


fweik:
What is the effect of the change?
Also:
```
 83/123 Test  #48: lb_stokes_sphere ........................   Passed  393.61 sec
```

KaiSzuttor:
it increases the verbosity... which can only be seen if the test fails. I wanted to see whether the GPU tests are responsible.

KaiSzuttor:
@fweik 
Lets merge that, so we can at least see if the GPU test takes to long due to oversubscription.

fweik (approved):

PR #2623, opened 2019-03-21
Observable template

jngrad:
sorry

patches #2621


fweik:
Why did the old code build?

jngrad:
Not sure, there isn't any `T` class/struct anywhere; you can even add `using C = T;` or `using D = UWX;` in `/src/script_interface/observables/CylindricalProfileObservable.hpp` and it will still compile just file. But it won't work in `/src/core/`. Maybe the `-Wall` flag in CMake is not inherited by the target that compiles the Python bindings?

jngrad:
File `/src/script_interface/observables/CylindricalProfileObservable.hpp` needs to be removed. It is not included anywhere, and the classes declared in it are actually already declared in `/src/script_interface/observables/initialize.cpp` and properly unit tested.

fweik (approved):

PR #2625, opened 2019-03-22
Demonstrate broken three-point interpolation in 4.0

KaiSzuttor:
@RudolfWeeber


testsuite/lb_poiseuille_4.0.py
```diff
@@ -0,0 +1,178 @@
...
+            self.system.box_l[0] - 2.0 * AGRID,
+            50)
+
+        cnt = 0
+        for x in x_values:
+            v_tmp = []
+            for y in range(int(self.system.box_l[1] + 1)):
+                for z in range(int(self.system.box_l[2] + 1)):
+                    v_tmp.append(
+                        self.lbf.get_interpolated_fluid_velocity_at_positions(np.array([[x, y * AGRID, z * AGRID],]))[0][2]
```
KaiSzuttor: the same test can be run on the current python branch by replacing this line with
``` python
self.lbf.get_interpolated_fluid_velocity_at_positions(np.array([[x, y * AGRID, z * AGRID],]), True)[0][2]
```

PR #2626, opened 2019-03-22
Demonstrate correctness of three-point interpolation.

KaiSzuttor:
see #2625


fweik:
Do you want to merge this, or is it just for demonstration purposes?

KaiSzuttor:
it was just for demonstration, the same core function is already tested in the python branch

PR #2627, opened 2019-03-22
Scriptinterface::Observables documentation

jngrad:
Description of changes:
- Improved user experience when creating new Observable classes
- Merged the Org-mode documentation of the ScriptInterface into Doxygen


jngrad:
Not sure what is wrong here. I retried `rocm-maxset` 3 times, but it always timed out on the `checkpoint_lb.gpu-p3m.cpu-lj_1` test. <strike>Yet the test skips the LB part completely with a fixture because the runner has no GPU, so the issue shouldn't be related to LB slowdown.</strike> I've pulled the image in Docker and ran the test locally in a loop around 1500 times over two hours, but couldn't reproduce the timeout. The container has no `Debug` mode, so we cannot test it live.

edit: these is actually an AMD GPU, so it might be related to LB slowdown. We'll see with Michael and Frank how to setup a dedicated runner with an AMD card.

fweik:
I am currently preparing an update for the script interface which will change the structure and add new documentation. I'll integrate this PR when I am done with that.

fweik (approved):

PR #2629, opened 2019-03-24
Vector3d support for ScriptInterface

### LINKED ISSUE #1687: Types named differently in C++ and Cython
E.g. int_list vs IntList. That is just confusing, they are the same type.

### LINKED ISSUE #2531: ScriptInterface variant should support Vector3d
Currently `Vector3d` is converted to a `std::vector<double>` when assigned to a `ScriptInterface::Variant`. This is not only wasteful, it also requires `Vector3d` to have an awkward implicit conversion operator to `std::vector`.

fweik:
Fixes #2531. Fixes #1687.
Description of changes:
 - Directly support Vector3d for script_interface parameters
 - Moved conversions from python to c++


src/python/espressomd/script_interface.pyx
```diff
@@ -165,30 +125,30 @@ cdef class PScriptInterface(object):
...
-        if < int > type == <int > DOUBLE:
-            return get[double](value)
-        if < int > type == <int > STRING:
-            return to_str(get[string](value))
-        if < int > type == <int > INT_VECTOR:
-            return get[vector[int]](value)
-        if < int > type == <int > DOUBLE_VECTOR:
-            return get[vector[double]](value)
-        if < int > type == <int > OBJECTID:
+        if is_type[bool](value):
```
KaiSzuttor, line 139: can we reduce code duplication by using type deduction here?

fweik: Can you elaborate on that?

KaiSzuttor: Ok, different question, can you elaborate on why it's necessary to check each type?

fweik: Because you have to map every type to python.

KaiSzuttor (approved):

PR #2631, opened 2019-03-26
Testsuite: Speedup some LB tests (for sanitizer builds, mainly)

RudolfWeeber:


fweik (approved):

testsuite/python/lb_poiseuille.py
```diff
@@ -91,11 +91,13 @@ def prepare(self):
                        int((self.system.box_l[2] / AGRID) / 2)]
         diff = float("inf")
         old_val = self.lbf[mid_indices].velocity[2]
+        self.system.time = 0
```
KaiSzuttor: why did you add this and the print?

testsuite/python/virtual_sites_tracers_common.py
```diff
@@ -219,8 +219,8 @@ def test_triel(self):
         # non-bonded should move apart by the flow (control group)
         # weakly-bonded should stretch somewhat
         # strongly-bonded should basically not stretch
-        self.assertGreater(dist1non, 2)
-        self.assertAlmostEqual(dist1weak, 1, delta=0.3)
+        self.assertGreater(dist1non, 1.5)
```
KaiSzuttor, line 222: is that criterion a magic number or where does it come from?

RudolfWeeber:
On Wed, Mar 27, 2019 at 02:37:02AM -0700, Kai Szuttor wrote:
> 
> 
> 
> > @@ -91,11 +91,13 @@ def prepare(self):
>                         int((self.system.box_l[2] / AGRID) / 2)]
>          diff = float("inf")
>          old_val = self.lbf[mid_indices].velocity[2]
> +        self.system.time = 0
> 
> why did you add this and the print?
To figure out the warmup time as a function of lb parametesr.
I'll remove it.

RudolfWeeber:
On Wed, Mar 27, 2019 at 02:37:59AM -0700, Kai Szuttor wrote:
> 
> 
> 
> > @@ -219,8 +219,8 @@ def test_triel(self):
>          # non-bonded should move apart by the flow (control group)
>          # weakly-bonded should stretch somewhat
>          # strongly-bonded should basically not stretch
> -        self.assertGreater(dist1non, 2)
> -        self.assertAlmostEqual(dist1weak, 1, delta=0.3)
> +        self.assertGreater(dist1non, 1.5)
> 
> is that criterion a magic number or where does it come from?
Yes. Basically it tests that un-bound particles are advected with the fluid wheras the bound ones stay at the desired distance.
Since now the system is propagated not that far, the free particles are advected less.
The way this is tested is not ideal, but that's for a different day.

KaiSzuttor (approved):

PR #2632, opened 2019-03-27
Clang tidy

KaiSzuttor:
added some more checks


CMakeLists.txt
```diff
@@ -524,7 +524,57 @@ if(WITH_CLANG_TIDY)
     set(
       CMAKE_CXX_CLANG_TIDY
       "${CLANG_TIDY_EXE}"
-      "-checks=*,clang-analyzer-alpha.*,-misc-unused-*,-android-*,-fuchsia-*,-google-*,-llvm-*,-modernize-*,-readability-*,-cppcoreguidelines-*,-clang-analyzer-deadcode*,-cert-*,-clang-analyzer-core.NullDereference,-clang-analyzer-core.DivideZero,-misc-misplaced-widening-cast,-misc-macro-parentheses, -clang-analyzer-security.FloatLoopCounter, -clang-analyzer-optin.mpi.MPI-Checker,-boost-use-to-string,-hicpp-*, modernize-use-nullptr, modernize-use-auto, modernize-pass-by-value, modernize-avoid-c-arrays, modernize-use-equals-default, modernize-deprecated-headers, readability-non-const-parameter"
+      "-checks=*,\
```
fweik: I think you can put the check list into a `.clang-tidy` file similar to the `.clag-format`, which is maybe more readable than to put it into cmake.

src/core/RuntimeErrorCollector.cpp
```diff
@@ -98,7 +98,7 @@ int RuntimeErrorCollector::count() const {
   int totalMessages;
   const int numMessages = m_errors.size();
 
-  all_reduce(m_comm, numMessages, totalMessages, std::plus<int>());
+  all_reduce(m_comm, numMessages, totalMessages, std::plus<>());
```
fweik: Can you check that boost.mpi can still detect that this is a MPI build-in operation with this?

fweik: There are multiples like this.

fweik (changes requested):

src/core/RuntimeErrorCollector.cpp
KaiSzuttor: How can I check that?

fweik: Look at the implementation

fweik: I can look it up

fweik: Looking at https://github.com/boostorg/mpi/blob/develop/include/boost/mpi/operations.hpp#L218, `std::plus<void>` is not yet recognized...

fweik: FYI: https://github.com/boostorg/mpi/pull/83, but that of course does not help us now.

KaiSzuttor: does this has a significant impact on performance?

fweik: Yes

fweik (approved):

PR #2633, opened 2019-03-28
Tentative fix for Lb boundary force

RudolfWeeber:
This changes the unit conversion forLbBoundary.get_force() to
agrid^4/(rho tau^2)
The origin of this factor is not clear to me, but it makes the outcome robust against lb parameter changes
The lb_shear test now contains checking of the boundary force (parallel to the shear direction) within 1%
The lb_stoks_sphere test now uses a much bigger box and a drag velocity well below the speed of sound. Accuracy is 3%. Bigger boxes and diffeent time steps/viscosities/densities work as well.
The runtime is still quite long.
Things to clarify:
What is lbpar.rho (density or mass per node?)
what are the units of the momentum nodes (momentum or momentum density?)

@jainka @KaiSzuttor


testsuite/python/lb_stokes_sphere.py
```diff
@@ -100,24 +103,25 @@ def size(vector):
...
-        self.system.integrator.run(200)
-
+        last_force = -1000.
         stokes_force = 6 * np.pi * KVISC * radius * size(v)
-
-        # get force that is exerted on the sphere
+        while True:
+            self.system.integrator.run(60)
+            force = np.linalg.norm(sphere.get_force())
+            print(force, last_force)
```
KaiSzuttor: please remove all prints

testsuite/python/lb_shear.py
```diff
@@ -28,12 +28,16 @@
...
-AGRID = 0.5
-VISC = 1.4
+AGRID = 0.6
+VISC = 3.2
 DENS = 2.3
 TIME_STEP = 0.01
-H = 30.
-SHEAR_RATE = 0.3
+# Box size will be H +2 AGRID to make room for walls.
+# The number of grid cells should be devisible by for and 3 in all directions
```
KaiSzuttor: typo "four"

testsuite/python/lb.py
```diff
@@ -163,6 +164,42 @@ def test_mass_momentum_thermostat(self):
         self.assertAlmostEqual(
             np.mean(all_temp_particle), self.params["temp"], delta=temp_prec_particle)
 
+    def test_stress_tensor(self):
```
KaiSzuttor, line 166: please add a test description

src/python/espressomd/observables.py
```diff
@@ -243,6 +243,18 @@ class LBVelocityProfile(Observable):
     _so_name = "Observables::LBVelocityProfile"
 
 
+@script_interface_register
+class LBFluidStress(Observable):
+    _so_name = "Observables::LBFluidStress"
+    """Calculates the average stress of the LB fluid for all nodes.
```
KaiSzuttor, line 249: docstring should be the first content of the body

src/python/espressomd/lb.pyx
```diff
@@ -263,6 +263,16 @@ IF LB:
             self._set_lattice_switch()
             self._set_params_in_es_core()
 
+        property stress_fluid:
```
KaiSzuttor: The observable is named "FluidStress", to be consistent name the property fluid_stress

src/core/grid_based_algorithms/lbboundaries.cpp
```diff
@@ -347,28 +347,24 @@ int lbboundary_get_force(void *lbb, double *f) {
...
-    f[0] = forces[3 * no + 0] * rho * agrid /
-           (tau * tau); // lbpar.tau; TODO this makes the units wrong and
-    f[1] = forces[3 * no + 1] * rho * agrid /
-           (tau * tau); // lbpar.tau; the result correct. But it's 3.13AM
-    f[2] = forces[3 * no + 2] * rho * agrid /
-           (tau * tau); // lbpar.tau; on a Saturday at the ICP. Someone fix.
+  }
+  const auto rho = lb_lbfluid_get_density();
+  const auto agrid = lb_lbfluid_get_agrid();
+  const auto tau = lb_lbfluid_get_tau();
```
KaiSzuttor: the unit conversion should *not* happen here, it should happen in the python interface

src/core/grid_based_algorithms/lb_interface.cpp
```diff
@@ -1273,13 +1265,64 @@ const Vector6d lb_lbnode_get_pi_neq(const Vector3i &ind) {
...
+
+    p[0] += p0;
+    p[2] += p0;
+    p[5] += p0;
+
+#endif
+  } else
+#ifdef LB
+      if (lattice_switch == ActiveLB::CPU) {
+    int number_of_nodes = lblattice.global_grid[0] * lblattice.global_grid[1] *
```
KaiSzuttor: you should define the variable right before usage

src/core/grid_based_algorithms/lb_interface.cpp
```diff
@@ -1273,13 +1265,64 @@ const Vector6d lb_lbnode_get_pi_neq(const Vector3i &ind) {
...
+                          lblattice.global_grid[2];
+
+    for (int i = 0; i < lblattice.global_grid[0]; i++) {
+      for (int j = 0; j < lblattice.global_grid[1]; j++) {
+        for (int k = 0; k < lblattice.global_grid[2]; k++) {
+          const Vector3i node{{i, j, k}};
+          p += lb_lbnode_get_pi(node);
```
KaiSzuttor, line 1327: this is insanely slow because you have "number of nodes" times the mpi communication. We should think about a generic solution to access all LB nodes

KaiSzuttor (changes requested):

KaiSzuttor:
on units:
mode 0: [mass unit] / [length unit]^3
modes 1-3: [mass unit] / [time unit] / [length unit]^2

src/core/grid_based_algorithms/lb_interface.cpp
RudolfWeeber: See timings in #2054. Walberla has its own efficient implementations, afaik.

RudolfWeeber:
@KaiSzuttor, I think I addressed everything

src/core/grid_based_algorithms/lb_interface.cpp
```diff
@@ -1273,13 +1265,64 @@ const Vector6d lb_lbnode_get_pi_neq(const Vector3i &ind) {
                              lblattice.halo_grid);
 
     mpi_recv_fluid(node, index, &rho, j, pi.data());
```
KaiSzuttor: ```p_pi.data()```

KaiSzuttor: and remove ```Vector6d pi{};```

src/core/grid_based_algorithms/lb_interface.cpp
```diff
@@ -1273,13 +1265,64 @@ const Vector6d lb_lbnode_get_pi_neq(const Vector3i &ind) {
...
 #endif // LB
   }
   return p_pi;
 }
 
+/** calculates the average stress of all nodes by iterating
+ * over all nodes and deviding by the number_of_nodes.
+ */
+const Vector6d lb_lbfluid_get_stress() {
+  Vector6d p{0, 0, 0, 0, 0, 0};
```
KaiSzuttor, line 1296: ```Vector6d p{};```

src/core/grid_based_algorithms/lb_interface.cpp
```diff
@@ -1273,13 +1265,64 @@ const Vector6d lb_lbnode_get_pi_neq(const Vector3i &ind) {
...
+                  [&p](LB_rho_v_pi_gpu &v) {
+                    for (int i = 0; i < 6; i++)
+                      p[i] += v.pi[i];
+                  });
+
+    // Normalize
+    p *= (1. / lbpar_gpu.number_of_nodes);
+
+    // Add equilibrium stress to the diagonal (in LB units)
+    double p0 = lb_lbfluid_get_density() * lbmodel.c_sound_sq;
```
KaiSzuttor: ```double const p0=```...

src/core/grid_based_algorithms/lb_interface.cpp
```diff
@@ -1273,13 +1265,64 @@ const Vector6d lb_lbnode_get_pi_neq(const Vector3i &ind) {
...
+    for (int i = 0; i < lblattice.global_grid[0]; i++) {
+      for (int j = 0; j < lblattice.global_grid[1]; j++) {
+        for (int k = 0; k < lblattice.global_grid[2]; k++) {
+          const Vector3i node{{i, j, k}};
+          p += lb_lbnode_get_pi(node);
+        }
+      }
+    }
+
+    int number_of_nodes = lblattice.global_grid[0] * lblattice.global_grid[1] *
```
KaiSzuttor: ```int const```

src/script_interface/lbboundaries/LBBoundary.hpp
```diff
@@ -65,9 +67,14 @@ class LBBoundary : public AutoParameters<LBBoundary> {
...
       // The get force method uses mpi callbacks on lb cpu
-      if (this_node == 0)
-        return m_lbboundary->get_force();
-
+      if (this_node == 0) {
+        const auto rho = lb_lbfluid_get_density();
+        const auto agrid = lb_lbfluid_get_agrid();
+        const auto tau = lb_lbfluid_get_tau();
+        const double unit_conversion =
+            agrid * agrid * agrid * agrid / rho / tau / tau;
```
KaiSzuttor, line 75: I'm not yet sure about the unit conversion:

```
[length unit]^7 / ([mass unit] [time unit]^2)
```

Looks weird to me...

KaiSzuttor: I'd expect:
```
[length unit] / [time unit]^2
```

RudolfWeeber:
Looking at 
lb_lbnode_get_velocity(),
the velocity is returned as
j / rho
Based on the various flow field tests (couette, poiseulle, ...), I think we can assume, that the velocity is correct.
So there are two possible units for j and rho:
1. j = momentum density (kg m /m^3 s) and rho = density (kg/m^3)
2. j = momentum (kg m/s) and rho = mass (kg)

@KaiSzuttor, I think what you posted above is variant 1.

Assuming variant 1 and that the "force" value stored in the boundary is of tyhe same units as j,
we would have to go from
momentum denisty 
to
momentum / time.
That would mean a multiplication with length^3/time
Then, to convert momentum from lb to md units, there would be an agrid/tau, analogous to the conversoin for the velocity in
p.pxd:python_lbnode_get_velocity()

So the difference between that and the factor used in this pr is 
1/mass

The bounce back code (lb.cpp:lb_bounce_back()) actually contains a multiplication by rho:
              population_shift -= lbpar.rho * 2 * lbmodel.c[i][l] *
                                  lbmodel.w[i] * lbfields[k].slip_velocity[l] /
                                  lbmodel.c_sound_sq;
            }

KaiSzuttor:
@RudolfWeeber I'm still not really convinced:
from momentum density to force we get a factor of [length]^3/[time]
from force in LB units to force in MD units we get a factor of agrid / tau^2.
That would leave us with agrid^4 / tau^3. This differs to the factor proposed in this PR by a factor of [length]^3/[mass]/[time]. Please correct me if I'm wrong.

RudolfWeeber:
My conclusion from offline discussion:
Momentum and force on the boundary (momentum that went into the boundary / time step) have the same (numeric) value in LB units, because the time step in lb units is 1.
The value stored in the boudary nodes is
mass_density * momentum_density

KaiSzuttor (approved):

Issue #2502, opened 2019-02-07
Use Vector3d and std::tuple in force functions


jngrad:
Use `Vector3d force` instead of `double force[3]` and `std::tuple` instead of output arguments in bonded/non-bonded functions called by force/pressure functions, i.e. from

```c++
// /src/core/bonded_interactions/umbrella.hpp
inline int calc_umbrella_pair_force(
    Particle const *p1, Particle const *p2,
    Bonded_ia_parameters const *ia_params,
    double const d[3], double force[3]) {
   ...
}
```
```c++
// /src/core/forces_inline.hpp
double force[3] = {0., 0., 0.};
...
bond_broken = calc_umbrella_pair_force(p1, p2, iaparams, dx, force);
```

to

```c++
#include <tuple>
inline std::tuple<int, Vector3d> calc_umbrella_pair_force(
    Particle const *p1, Particle const *p2,
    Bonded_ia_parameters const *ia_params,
    Vector3d const d) {
   ...
}
```
```c++
#include <tuple>
Vector3d force;
...
std::tie(bond_broken, force) = calc_umbrella_pair_force(p1, p2, iaparams, dx);
```

Maybe also use `bool` instead of `int` for  variable `bond_broken` and the return value of bonded forces.

jngrad:
C++14:
```c++
#include <tuple>

inline std::tuple<bool, Vector3d>
calc_pair_force(Particle const *p1, Particle const *p2,
                Bonded_ia_parameters const *ia_params) {
  if(d < cutoff) {
    return std::make_tuple(true, Vector3d{});
  }
  Vector3d force;
  /* ... */
  return std::make_tuple(false, force);
}

void main_loop() {
  Vector3d force;
  bool bond_broken;
  std::tie(bond_broken, force) = calc_pair_force(p1, p2, iaparams, dx);
}
```

C++17:
```c++
#include <tuple>

inline std::tuple<bool, Vector3d>
calc_pair_force(Particle const *p1, Particle const *p2,
                Bonded_ia_parameters const *ia_params) {
  if(d < cutoff) {
    return {true, {}};
  }
  Vector3d force;
  /* ... */
  return {false, force};
}

void main_loop() {
  auto [bond_broken, force] = calc_pair_force(p1, p2, iaparams, dx);
}
```

fweik:
The same should be done for the energy kernels.

jngrad:
Target:
```c++
boost::optional<double> tab_angle_energy(...);
[...]

PR #2634, opened 2019-03-28
Bond angle refactor

jngrad:
Each of the 4 bond angle (harmonic, cosine, cossquare, tabulated) force functions was defined twice, once for `pressure_inline.hpp` (aka `calc_angle_..._3body_forces()`) using the external angle convention, and once for `forces_inline.hpp` (aka `calc_angle_..._force()`) using the internal angle convention. Since the angle convention changes the equations for the angle energy and angle forces, the duplicated functions ended up diverging significantly in coding style, causing severe readability issues, even though they gave the same numerical results.

Changes:
   - all angle energy/force functions now use the internal angle convention
   - force calculation is done in functions `calc_angle_..._3body_forces()`, while functions  `calc_angle_..._force()` are just wrappers (commit 0cc3904, the diff is hard to read)
   - new observables to monitor bond distances, angles, and dihedrals during integration
   - new Doxygen page dedicated to the bonded interactions and their equations
   - tabulated angle potentials are now unit tested
   - removed unused `TABANGLEMINUS` (old debug macro to switch from internal to external convention in tabulated angle functions, I suppose)

Remarks:
   - the wrapper functions `calc_angle_..._force()` still use double[3] instead of Vector3d; these wrapper functions will be removed in the future when Vector3d becomes more widespread (#2502)
   - the dihedral angle functions will be refactored in a future PR, which will cause an API change
   - the Doxygen page with equations will be supplemented with more interactions in the future (e.g. dihedrals); it shouldn't be moved to Sphinx since it is only relevant to maintainers who need to understand where the formula in the C++ code come from

Currently, when the bond angle is &pi;, the behavior of angle force functions is inconsistent: AngleCossquare sets the forces to zero, AngleCosine sets the force to NaN, AngleHarmonic makes the two outer particles slowly collapse on the central particle. This is confusing, should we instead set the force to zero regardless of the potential?

### DIFF
--- a/doc/doxygen/Doxyfile
+++ b/doc/doxygen/Doxyfile
@@ -1262,7 +1262,7 @@ PAPER_TYPE             = a4wide
 # The EXTRA_PACKAGES tag can be to specify one or more names of LaTeX
 # packages that should be included in the LaTeX output.
 
-EXTRA_PACKAGES         =
+EXTRA_PACKAGES         = amsmath
 
 # The LATEX_HEADER tag can be used to specify a personal LaTeX header for
 # the generated latex document. The header should contain everything until
--- a/doc/sphinx/inter_bonded.rst
+++ b/doc/sphinx/inter_bonded.rst
@@ -11,7 +11,7 @@ the following syntax to activate and assign a bonded interaction::
     system.bonded_inter.add(bond)
     system.part[pid1].add_bond((bond, pid2...))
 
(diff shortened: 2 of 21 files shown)


fweik (approved):

jngrad:
@fweik should we just ignore the inconsistent behavior at &theta;=&pi;?

fweik:
I don't have an opinion on that.

PR #2637, opened 2019-04-02
WIP: async GPU force/torque back-transfer, launch kernels earlier

RudolfWeeber:
This uses the thrust pinned memory allocator for the vectors receiving gpu forces/torques on the host.
Then, the back-transfer can be asynchroneous.
The host force/torque vector have to remain globals, because the allocation of pinned memory takes at least as long as the data transfer itself.
A Python-level exit handler maeks sure, the vectors are de-allocated before cuda is unloaded (becuase the custom allocator calls cudaFree())
The performance benefits apply mostly to dense systems and probably also to systems with long range interactions.

### DIFF
--- a/src/core/cuda_common_cuda.cu
+++ b/src/core/cuda_common_cuda.cu
@@ -49,19 +49,16 @@ static CUDA_particle_seed *particle_seeds_device = nullptr;
 static CUDA_energy *energy_device = nullptr;
 
 CUDA_particle_data *particle_data_host = nullptr;
-std::vector<float> particle_forces_host;
 CUDA_energy energy_host;
 
-std::vector<float> particle_torques_host;
 #ifdef ENGINE
 std::vector<CUDA_v_cs> host_v_cs;
 #endif
 
-/**cuda streams for parallel computing on cpu and gpu */
-cudaStream_t stream[1];
-
 cudaError_t _err;
 cudaError_t CU_err;
+
+cudaStream_t stream[1];
 
 void _cuda_safe_mem(cudaError_t CU_err, const char *file, unsigned int line) {
   if (cudaSuccess != CU_err) {
@@ -185,8 +182,6 @@ void gpu_change_number_of_part_to_comm() {
                                      &global_part_vars_host,
                                      sizeof(CUDA_global_part_vars)));
 
-    // if the arrays exists free them to prevent memory leaks
-    particle_forces_host.clear();
     if (particle_data_host) {
       cuda_safe_mem(cudaFreeHost(particle_data_host));
       particle_data_host = nullptr;
@@ -206,9 +201,6 @@ void gpu_change_number_of_part_to_comm() {
 #ifdef ENGINE
     host_v_cs.clear();
 #endif
-#ifdef ROTATION
-    particle_torques_host.clear();
-#endif
 
 #ifdef ROTATION
     if (particle_torques_device) {
@@ -224,14 +216,8 @@ void gpu_change_number_of_part_to_comm() {
                                   global_part_vars_host.number_of_particles *
                                       sizeof(CUDA_particle_data),
                                   cudaHostAllocWriteCombined));
-      particle_forces_host.resize(3 *
-                                  global_part_vars_host.number_of_particles);
 #ifdef ENGINE
       host_v_cs.resize(global_part_vars_host.number_of_particles);
-#endif
-#if (defined DIPOLES || defined ROTATION)
-      particle_torques_host.resize(3 *
-                                   global_part_vars_host.number_of_particles);
 #endif
 
       cuda_safe_mem(cudaMalloc((void **)&particle_forces_device,
@@ -267,6 +253,8 @@ void gpu_change_number_of_part_to_comm() {
   }
 }
 
+cudaEvent_t forces_torques_dtoh;
+
 /** setup and call particle reallocation from the host
  *  Note that in addition to calling this function the parameters must be
  * broadcast with either:
@@ -277,6 +265,8 @@ void gpu_change_number_of_part_to_comm() {
  * nodes)
  */
 void gpu_init_particle_comm() {
+  cuda_safe_mem(
+      cudaEventCreateWithFlags(&forces_torques_dtoh, cudaEventDisableTiming));
   if (this_node == 0 && global_part_vars_host.communication_enabled == 0) {
     if (cuda_get_n_gpus() == -1) {
       runtimeErrorMsg()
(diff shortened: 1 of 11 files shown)


src/core/cuda_common_cuda.cu
```diff
@@ -336,26 +326,45 @@ void copy_part_data_to_gpu(ParticleRange particles) {
   }
 }
 
+std::unique_ptr<PinnedVectorHost<float>> particle_forces_host{
```
fweik, line 329: I really don't like this construction. I a way this is worse than manual memory management, because it pretends to be RAII, but really isn't. As far as I can see there is not reason for these buffers to exist statically, couldn't you just create them e.g. when the integration starts, and release them after.  ~~Depending on the cost of the allocation, it may even be feasible to create it in the force calculation.~~

src/core/cuda_interface.hpp
```diff
@@ -26,6 +26,12 @@
 #include "ParticleRange.hpp"
 #include "SystemInterface.hpp"
 
+#include "thrust/host_vector.h"
+#include "thrust/system/cuda/experimental/pinned_allocator.h"
+template <class T>
+using PinnedVectorHost = thrust::host_vector<
```
fweik, line 32: This should be called `PinnedHostVector` to keep it closer to thrusts name.

fweik (changes requested):

RudolfWeeber:
It is unfortunate that one cannot convince Python to actually call the destructor of the system class.
Then this could be done differently. As it is, I didn't find a working alternative to the exit handler.



For 1k particles, the alloc takes approx as long as an md time step with lj at 0.1 volume fraction (200us).
It scales much less then linearly with the number of particles

It would probably surprise users if
10x integrator.run(1) takes 20x the time than integrator.run(10)

The inner-most place one could move the vectors to is 
python_integrate(), then, the vectors would stay allocated for sytems with auto update accumulators, at least.
AFAIK, there eis no observable recorder (yet) which could store un-processed observable results, though.

fweik:
Can you please explain in more detail what the lifetime issue is? It's not totally clear to me.

fweik:
Also you seemed to have timed the allocation, could you please share your results and how you have measured that? It's not clear to me why a single allocation (or two) would be so expensive, my understanding is that the pinned allocation is just a normal allocation followed by a `mlock` call. This should not be expensive (if you are not out of RAM).

RudolfWeeber:
On Tue, Apr 02, 2019 at 04:07:08AM -0700, Florian Weik wrote:
> Can you please explain in more detail what the lifetime issue is? It's not totally clear to me.
For the evaluation of observables, the main integration function, integrate_vv(), is interupted, because ES switches to master-slave mode.
If there are auto update accumulators, integrate_vv is called n times with 1 step.

RudolfWeeber:
On Tue, Apr 02, 2019 at 04:16:05AM -0700, Florian Weik wrote:
> Also you seemed to have time the allocation, could you please share your results and how you have measured that? It's not clear to me why a single allocation (or two) would be so expensive, my understanding is that the pinned allocation is just a normal allocation followed by a `mlock` call. This should not be expensive (if you are not out of RAM).
I used nvprof.
Percentages are with regards to the total time of all cuda api calls on the host.
                    0.14%  471.65us         2  235.83us  8.4520us  463.20us  cudaMallocHost
                    0.09%  321.79us         2  160.89us  22.721us  299.07us  cudaFreeHost

So the total time for allocation and de-allocation is .7ms.
One time step (without alloc/free) takes 0.28ms (with lb without lj).

fweik:
Again: Can you please explain in more detail what the lifetime issue is? It's not totally clear to me.

RudolfWeeber:
On Tue, Apr 02, 2019 at 06:43:22AM -0700, Florian Weik wrote:
> Again: Can you please explain in more detail what the lifetime issue is? It's not totally clear to me.
I'm not sure, I understand the question.
If you are referring to where the need for the exit handler arises:
The thrust allocator used in the vectors calls cudaFree at destruction to release the pinned memory.
That fails, if the cuda runtime has been unloaded already.
AFAIK, we cannot control the oder in which shared objects are unloaded, so 
Cuda might be unloaded before libEspressoCore.
Then, the cudaFree()-call releasing the vectorss' memories throws an exception.

fweik:
Wouldn't you have the same problem  with `free()` from the shared object `libc`?

RudolfWeeber:
It is not clear to me.
There is, as far as I can tell, no explicit call unloading the cuda driver in Espresso.
Also, I could not find a global whose destruciotn would entail the unloading.
So I cannot really think about an other explanation than the unloading order of shared objects.
One odd thing is that
libEspressoCore.so and libEsperssoCua.so
are actually not linked against libcudart, but just against libcufft:

	linux-vdso.so.1 (0x00007ffebe3fa000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x0000145d6b140000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x0000145d6af3c000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x0000145d6ad34000)
	libcufft.so.9.1 => /usr/lib/x86_64-linux-gnu/libcufft.so.9.1 (0x0000145d63847000)
	libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x0000145d634be000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x0000145d63120000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x0000145d62f08000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x0000145d62b17000)
	/lib64/ld-linux-x86-64.so.2 (0x0000145d6b89b000)

fweik:
It seems to me that I forgot linking to cudart:

(From `/CMakeLists.txt`)

```cmake
        function(add_gpu_library)
          cuda_add_library(${ARGV})
          set_property(TARGET ${ARGV0} PROPERTY CUDA_SEPARABLE_COMPILATION ON)
          target_link_libraries(${ARGV0} PRIVATE ${CUDA_CUFFT_LIBRARIES})
        endfunction()
```

Could you please add it (like for the clang case) and check if this improves matters? (I probably thought that `cuda_add_library` does that automatically...)

RudolfWeeber:
So libcudart was linked statically already. Linking it dynamically did not help.

It is possible to avoid the exit hanlder by storing the vectors containig the pinned memory in a wrapper class which explicitly creates a cuda context. An releases it after destructing the vectors.

This solution still needs work, e.g., for switching gpus. It also needs to be moved/integrated with cuda init and device switching.
I'll do that, if we go with this solution.

fweik:
Ok I think we can go forward with this. In general I think we should seek to remove all the non-trivial globals rather sooner than later. We've had multiple issues with them.

fweik:
@RudolfWeeber it seems like there is a file missing here (where the feature detection in python went), could you pleas add that? I am currently working on the async forward communication and want to integrate this, but it is not working.

fweik:
@RudolfWeeber are you still looking into this?

RudolfWeeber:
>> <https://github.com/RudolfWeeber> @RudolfWeeber are you still looking into this?

IIRC, you said you wanted to integrate this into the non-blocking mpi-node -> gpu communication.

If that’s not the case, I will look into it once I’m back from Paris.

It also depends on the decision we take with regards to LBGPU.

Even if we don’t do the full thing, the early starting of the GPU methods could be merged independently.

fweik:
It seems to me that you forgot to check in some files.

RudolfWeeber:
>It seems to me that you forgot to check in some files.
Even including commit afded8?

What is still missing?

fweik:
Ah maybe there is just an guard missing. I'll have a look

KaiSzuttor:
@fweik is this obsolete by now?

fweik:
No. Let's keep this open for now.

PR #2640, opened 2019-04-03
utils: get_linear_index + cleanup

fweik:


KaiSzuttor:
its unfortunate that the unravel index uses a different convention, couldn't we merge your logic into the unravel/ravel index functions without a lot of effort?

fweik:
Since this is performance relevant in some places I'd keep the version with static dimension (=3). Also from a quick inspection I'd say that one is row-major and one is column-major, because why not?

KaiSzuttor:
yes it's fine to have both versions, but maybe we want to have names that imply that the functions do the same (besides column/row -majorness)

fweik:
The proper way I think (since `ravel_index` already takes a range) would be to keep one name and specialize it for `Vector`. Also neither the name nor documentation of the functions says what they actually do. Also `ravel_index` and `unravel_index` seem not to be common names for this, I'm aware of the numpy function, but otherwise these names apparently are not use much.

fweik:
@KaiSzuttor so?

KaiSzuttor:
I agree, maybe we can add a static switch for the row/column-major logic?

fweik:
@KaiSzuttor I agree that this should be refactored, but this PR still is an improvement, so we should merge it.

KaiSzuttor (approved):

PR #2641, opened 2019-04-03
Skip long-running (timeout-prone) tests on exotic archs

RudolfWeeber:


RudolfWeeber:
How can I manually trigger the exotic arch tests for this?

mkuron:
This is a terrible solution. The tests should be fixed, not skipped.

KaiSzuttor:
it's an intermediate solution

KaiSzuttor (approved):

RudolfWeeber:
While I understand the discomfort, as long as no PR with a better solution is provided,
this PR aims to resolve an untenable situation keeping most of the tests on uncommen architectures.
Except for coulomb_tuning, all tests exlucded here run in less than 10 seconds on my desktop. P3m_Tuning takes 21 seconds.
So the tests are not inherently over the top in terms of computational effort.

I'll gladly withdraw this PR, if there is an alternative one solving the frequent false-positives.

Issue #2488, opened 2019-01-31
Implement proper notifications for scheduled pipelines


fweik:
So that they are not forgotten.

fweik:
Maybe they can file an issue on github, gitlab allows to trigger webhools on pipeline events.

fweik:
This is maybe fixed by #2643.

KaiSzuttor:
it is fixed

PR #2643, opened 2019-04-03
Notification if merged PR build fails

KaiSzuttor:
will add a python only after it seems working

### DIFF
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -1,4 +1,5 @@
 set(EspressoCore_SRC
+  abs.cpp
   accumulators.cpp
   cells.cpp
   collision.cpp
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -1,5 +1,4 @@
 set(EspressoCore_SRC
-  abs.cpp
   accumulators.cpp
   cells.cpp
   collision.cpp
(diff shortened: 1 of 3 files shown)


src/core/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 set(EspressoCore_SRC
+  abs.cpp
```
fweik: Is this related?

KaiSzuttor: ~yes its the newly introduced “anti blockading system” for particles~

KaiSzuttor: just for making the build fail

fweik (approved):

mkuron:
Can this be adapted to also fix #2488?

fweik:
Shouldn't that just work? The filter is only on the branch...

mkuron:
Could you have it automatically close the issue if the build succeeds after someone manually presses the retry button in Gitlab CI?

KaiSzuttor:
we dont expect that to happen very frequently, so I dont think this is necessary for now.

hmenke:
I recommend using the [new issues API](https://gist.github.com/jonmagic/5282384165e0f86ef105) of GitHub instead of the regular REST API.  The former one is not rate-limited because it always requires authorization.

KaiSzuttor:
I didn't get what the advantage is

PR #2648, opened 2019-04-04
CI: automatically closing CI created issues

KaiSzuttor:
for convenience

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -390,8 +390,6 @@ notify_success:
   stage: cleanup
   script: bash maintainer/gh_close_issue.sh
   when: on_success
-  only:
-    - python
   tags:
     - linux
   dependencies:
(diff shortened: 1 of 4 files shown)


KaiSzuttor:
this won't work across pushes :cry:

mkuron:
Why don't you just use the GitHub API to search for the auto-created issue that references the correct pull request?

KaiSzuttor:
we currently only reference the pipeline in the issue

mkuron:
Then search all issues for the pipeline ID.

KaiSzuttor:
You don't know the pipeline ID

mkuron:
Why? `${CI_PIPELINE_ID}` doesn't change between retries.

KaiSzuttor:
But it changes if I push, right?

mkuron:
Yes. I was only talking about a pipeline that fails, but passes after someone clicks the "retry" button.

mkuron:
And if it's the master branch, just close all open auto-created issues.

KaiSzuttor:
that's already possible with this PR

Issue #2579, opened 2019-03-12
Deadlock in dipolar Barnes-Hutt on the GPU


RudolfWeeber:
On an NVidai rtx 2080.

This happens in `sortBH()`, at the device synchronize after launching the `sortKernel()`
The likely cause is that some threads enter an endless loop:

```cuda
__global__ __launch_bounds__(THREADS4, FACTOR4) void sortKernel() {
  int i, k, ch, dec, start, bottom;

  bottom = bottomd;
  dec = blockDim.x * gridDim.x;
  // Start from the end of the nnodesd == 8 * nbodiesd.
  // Reverse order is required now cause octant cells which are more close
  // to the root have a larger count of entities inside (countd[k]).
  // Particles should be sorted over all entities count in the tree array
  // representation made by treeBuildingKernel.
  k = bhpara->nnodes + 1 - dec + threadIdx.x + blockIdx.x * blockDim.x;

  // iterate over all cells assigned to thread
  while (k >= bottom) {
    start = bhpara->start[k];
    // Let's start from the root which has only startd=0 defined
    // in boundingBoxKernel. All other bodies and cells have -1.
    if (start >= 0) {
      for (i = 0; i < 8; i++) {
        ch = bhpara->child[k * 8 + i];
        if (ch >= bhpara->nbodies) {
          // child is a cell
          bhpara->start[ch] = start;  // set start ID of child
          start += bhpara->count[ch]; // add # of bodies in subtree
        } else if (ch >= 0) {
          // Child is a body.
          // This particle should be saved with a stepping over
          // a count of particles in the cells.
          // treeBuildingKernel already has ordered cells in a
          // linear array way. The sortKernel just order random particle
          // indices in the same order. Hence, they will be much faster accessed
          // by forceCalculationKernel and energyCalculationKernel.
          bhpara->sort[start] = ch; // record body in 'sorted' array
          start++;
        }
      }
      k -= dec; // move on to next cell
    }
    __syncthreads(); // throttle
  }
}
```

If `start = bhpara->start[k]`  is less than zero, k is not decremented in the loop.
The deadlock happens most of the time, but not always.
There seems to be an undefined behavior or race condition.
The card's specs are

```
--- device 0: ---
device name: GeForce RTX 2080
compute capability: 7.5
number of multiprocessors: 46
global memory: 8338604032 bytes
size of L2 cache: 4194304 bytes
maximum shared memory per block: 49152 bytes
maximum number of registers per block: 65536
warp size: 32
[...]

PR #2653, opened 2019-04-04
Only run gpu tests when a gpu is present

RudolfWeeber:
And expose the availability of a gpu via espressomd.gpu_available()

I tested this in an nvidia-based coker container without a gpu attached.
It would be of interest to test this in the CI. As long as gitlab does not support excluding runners by tag (not cuda), this would require tagging all non-cuda runners with a non_cuda tag or the like.


fweik:
Wouldn't it be easier to introduce a common base class for all tests, instead of adapting every test? This would also make it less likely to forget this in the future. A premium solution would also require all tests to state their minimal dependencies.

RudolfWeeber:
Quite a few tests would then need multiple inheritance (e.g., lb).

This is order-dependent in a rather unintuitivew way:

tests_base.py:
import unittest as ut


@ut.skipIf(True, "skip")
class SkipBase(ut.TestCase):
    pass

class Base2(ut.TestCase):
    def test(self):
        print self.__class__.__name__

tests.py:

from test_base import SkipBase, Base2
import unittest as ut


class test1(SkipBase, Base2):
    pass

class test1(Base2, SkipBase):
    pass




if __name__ == "__main__":
    ut.main()

This runs one test and skps the other.
Not deriving Base2 from ut.TestCase will ignore the 2nd test entirely,.
For test classes which directly derive from ut.TestCase, it would be possible
@ut.skipIf(not espressomd.gpu_available()...)
class EKTest(ut.TestCase)
would become
class EKTest(GpuTestCase)

For tests such as coulomb_cloud_wall in which only some test methods depend on a
gpu, nothing can be done via a base class.

I think what's there now is the most clear.

fweik:
```python
            if str(espressomd.cuda_init.CudaInitHandle().device_list[0]) == "Device 687f":
                print("skipping test on amd gpu")
                return
```

how is this maintainable?

mkuron:
@fweik, it's not. It just turns of this specific test in our CI environment. If you run outside of the Docker container, you'll get a proper device name and the test will (and is supposed to) fail.

This was supposed to be a stop-gap until #2579 and the GPU P3M are fixed, both of which I couldn't do because I don't understand the code.

RudolfWeeber:
On Thu, Apr 04, 2019 at 12:52:56PM +0000, Florian Weik wrote:
> ```python
>             if str(espressomd.cuda_init.CudaInitHandle().device_list[0]) == "Device 687f":
>                 print("skipping test on amd gpu")
>                 return
> ```
> 
> how is this maintainable?
It isn't. This check was introduced with the HIP support
in
b6c884f4b58a406270
We need to figure out the cause of the deadlock or remove the
methods.

This PR touched the check, since its execution has to bedlayed to a
point where it is clear that a gpu is present.

bogdan-tanygin:
Hi all. The above-mentioned issue can be potentially resolved for the Barnes-Hut (GPU) on latest Nvidia and AMD architectures upon #2663. Sorry that this has taken so long: a root cause is really cumbersome.

fweik:
So with #2663 those device checks can be removed? This still leaves the original design issue here.

psci2195:
> This still leaves the original design issue here.

I agree. https://github.com/espressomd/espresso/pull/2663#issuecomment-481457076

mkuron:
> As long as gitlab does not support excluding runners by tag (not cuda), this would require tagging all non-cuda runners with a non_cuda tag or the like.

I suggest doing `export CUDA_VISIBLE_DEVICES=none` instead as that does't require GitLab involvement and will work even if some day all our build machines have GPUs.

bogdan-tanygin:
> So with #2663 those device checks can be removed? This still leaves the original design issue here.

These safety checks from https://github.com/espressomd/espresso/commit/b6c884f4b58a40627031d525d075e53b4629ff4c have been removed for the BH magnetostatics only. Not for P3M-GPU: it requires "similar" (i.e. hardware architecture agnostic) fixes, I suppose.

RudolfWeeber:
@mkuron, thanks for suggesting that. The test does exactly that.
As far as I can tell, it works.

RudolfWeeber:
On Tue, Apr 09, 2019 at 07:20:50AM -0700, Florian Weik wrote:
> So with #2663 those device checks can be removed? This still leaves the original design issue here.
There is also a device check for p3m_gpu. This, you probably need to look at.
This is independent of this PR.

With regards to the design:
This PR corrects incorrectly stated preconditions for running some tests.
I'm not going to extend it into a testsuite wide refactoring.

So the design question here is how to state that the test requires a gpu.

As I mentioned in a previous comment, I do not find the (partially possible) outsourcing of the requirements checking into a base class more clear than an explicit 
@ut.skipIf() 
decorator.
So, unless there is a third, better, option, I'd prefer to stay with 
@ut.skipIf(not gpu_available())

The case of gpu code being built but no gpu being present is now checked in the CI, so if someone forgets the check in the future, we will know.

fweik (approved):

fweik:
I think a more terse syntax could be achieved if there was a feature (say, "GPU") present if there is an usable gpu. The it would be possible to combine the gpu check with the feature checks. But I also think it's not terribly important.

PR #2659, opened 2019-04-04
Update energy.cpp

KonradBreitsprecher:
Restore p3m sums to only account for particle charges (without ELC contributions).
Cures ELC sanity check for charge neutrality that is called in the beginning of each integration step.

Fixes mailing list issue.


fweik (dismissed):

KaiSzuttor:
can this be tested?

fweik:
This is hard to test, and this code will likely be removed soon anyway.

fweik (approved):

PR #2662, opened 2019-04-04
testsuite: Test both interpolaton code paths of p3m

fweik:
P3M can optionally use an interpolated window function, this tests both possible code paths.


testsuite/python/coulomb_cloud_wall.py
```diff
@@ -102,7 +102,15 @@ def compare(self, method_name, energy=True, prefactor=None):
     # Tests for individual methods
 
     if espressomd.has_features(["P3M"]):
-        def test_p3m(self):
+        def test_p3m_direct_caf(self):
```
KaiSzuttor, line 108: please add a docstring

fweik: They all run the same check with different methods, none of the other test functions have docstrings.

KaiSzuttor: while I can imagine what happens in ```test_scafacos_p3m``` its hard to grasp the meaning of ```test_p3m_direct_caf```

KaiSzuttor (changes requested):
once the tests have a short description it can be merged

KaiSzuttor (dismissed):

KaiSzuttor (approved):

Issue #2500, opened 2019-02-06
DipolarBarnesHut does not work with multiple cores


fweik:
The dipolar method is not properly set. If fixed, core limit on the test should be removed.

fweik:
Duplicate of #2579.

bogdan-tanygin:
Actually, this issue is not yet fixed, it is independent one. I've prepared a separate pull request. Is it possible to reopen it formally (if needed)?

PR #2663, opened 2019-04-04
 Barnes-Hut gpu: Volta, Turing arch and AMD Vega fixes of deadlocks

### LINKED ISSUE #2579: Deadlock in dipolar Barnes-Hutt on the GPU

bogdan-tanygin:
Fixes #2579, psci2195#223
Does not fix (a matter of a separate PR from me?): #2500

Description of changes:
- An issue was much more complex: deadlocks have been reproduced in different kernels (treeBuildingKernel(), summarizationKernel(), sortKernel()) depending on GPU brand and architecture and depending on parameters. These issues started only since the latest GPU architectures (Volta and Turing level). These issues do not reproduce till Pascal inclusive.
 - The fix was a securing the summarizationKernel() logic by additional iterations count and some additional thread synchronization measures in all complex kernels.
- A number of threads and blocks also have been more precisely aligned between multiprocessors. Threads count was more precisely aligned with the warp (very important since Volta).
- All the fixes have been validated on different GPUs of Pascal and Volta architectures: **GeForce GTX 1070**, **Tesla V100**. According to a research, the Volta fixes should be probably fine for the Turing as well (probably a reporter of #2579 could kindly run the test dawaanr-and-bh-gpu.py on the NVidai rtx 2080).
- The AMD-specific fixes have been introduced and validated on **Radeon Vega Frontier Edition**.

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


bogdan-tanygin:
+ according to the Gitlab, this works on the Kepler as well.

RudolfWeeber:
Thanks alot.
It also works on the NVidia RTX 2080.

Why are the extra __syncthreads() calls on the AMD card necessary?

bogdan-tanygin:
> Why are the extra __syncthreads() calls on the AMD card necessary?

The treeBuildingKernel() (and some further kernels) is based on a mutex, hence, this is a cuda critical section case. This kind of patterns are known to be sensitive to the GPU hardware architecture specific compilation. I've seen reports of the similar issue. So, as one usually goes in such cases, a syntactic "trick" has been found: it leads to the required machine code.. Even though the algorithm pseudo-code would not require this.

RudolfWeeber (approved):
Tricking the compiler is probably not a sustainable solution.
I'm merging this, so we can continue to run the test suite on machines with recent gpus.

bogdan-tanygin:
> Tricking the compiler is probably not a sustainable solution.
> I'm merging this, so we can continue to run the test suite on machines with recent gpus.

I agree with this. If the compiler issues will resume on new hardware architectures then this feature should be developed from scratch using non-mutex approach. Even though it could lead to some reduction of the performance. Basically, it can become equal to the existing DDS-GPU magnetostatics. Will see..

PR #2666, opened 2019-04-05
Philox RNG for langevin thermostat

KonradBreitsprecher:
- Removed d_random() RNG in thermostat, switched to Philox RNG
- Added 'seed' keyword in langevin thermostat interface.
- Extended langevin testcase
- Updated docs, testsuite, samples


src/core/communication.cpp
```diff
@@ -67,6 +67,7 @@
 #include "statistics.hpp"
 #include "statistics_chain.hpp"
 #include "swimmer_reaction.hpp"
+#include "thermostat.hpp"
```
fweik: Please move the callbacks to `thermostat.cpp`, and remove the include here. Also now callbacks with any signature are allowed, so you can pass directly in the counter value instead of doing an additional broadcast, similar to the LB.

src/core/integrate.cpp
```diff
@@ -355,6 +361,12 @@ void integrate_vv(int n_steps, int reuse_forces) {
       ghost_communicator(&cell_structure.update_ghost_pos_comm);
     }
 #endif
+
+    // Propagate langevin philox rng counter
+    if (n_steps > 0) {
```
fweik: The check is not needed, you can only get here if the condition is true.

src/core/thermostat.cpp
```diff
@@ -40,6 +43,8 @@ double temperature = 0.0;
 /** True if the thermostat should act on virtual particles. */
 bool thermo_virtual = true;
 
+int langevin_seed;
```
fweik: This seems unneeded.

fweik (changes requested):

testsuite/python/langevin_thermostat.py
```diff
@@ -64,6 +64,52 @@ def test_aa_verify_single_component_maxwell(self):
...
+    def test_langevin_seed(self):
+        """Test for RNG seed consistency."""
+        system = self.system
+        system.time_step = 0.01
+        system.part.clear()
+        system.part.add(pos=[0, 0, 0])
+        
+        kT = 1.1
+        gamma = 3.5
+        system.thermostat.set_langevin(kT=kT, gamma=gamma, seed=41)
```
KaiSzuttor, line 81: please also add a test that one cannot use the thermostat without seed

src/core/communication.cpp
KonradBreitsprecher: I try to do that but I'm not quite sure if I got it right. Please carefully review lines 83-108 in the thermostat.cpp that are:

```
std::unique_ptr<Utils::Counter<uint64_t>> langevin_rng_counter;

void mpi_bcast_langevin_rng_counter_slave(const uint64_t counter) {
  langevin_rng_counter = std::make_unique<Utils::Counter<uint64_t>>(counter);
}

REGISTER_CALLBACK(mpi_bcast_langevin_rng_counter_slave)

void mpi_bcast_langevin_rng_counter(const uint64_t counter) {
  mpi_call(mpi_bcast_langevin_rng_counter_slave, counter);
}

void langevin_rng_counter_increment() {
  if (thermo_switch & THERMO_LANGEVIN)
    langevin_rng_counter->increment();
}

bool langevin_is_seed_required() {
  /* Seed is required if rng is not initialized */
  return langevin_rng_counter == nullptr;
}

void langevin_set_rng_state(const uint64_t counter) {
  mpi_bcast_langevin_rng_counter(counter);
  langevin_rng_counter = std::make_unique<Utils::Counter<uint64_t>>(counter);
}
```

fweik (approved):

fweik:
@KaiSzuttor

KaiSzuttor (approved):

PR #2677, opened 2019-04-08
Visualization: dihedrals connectivity

jngrad:
Visualizers currently represent proper dihedral bonds (A-B-C-D connectivity, aka torsion angle) as improper dihedral bonds (A-B(-C)-D connectivity, aka out-of-plane angle), causing some confusion when setting up linear polymers. This PR introduces a conditional statement to correctly treat bond connectivity in dihedral interactions.

MWE: re-use `/samples/visualization_bonded.py`, with `box_l = 200`, `n_part = 20`, then add
```python
from espressomd.interactions import Dihedral
dihedral = Dihedral(bend=10, mult=1, phase=np.pi)
system.bonded_inter.add(dihedral)
for i in range(1, n_part - 3):
    system.part[i].add_bond((dihedral, system.part[i - 1].id, system.part[i + 1].id,
        system.part[i + 2].id))
```
right before the minimization step. Without this PR, the polymer looks like a spiderweb. With this PR, the polymer looks like beads on a string, and the magenta segments (bonds) perfectly overlap with the orange segments (dihedrals). I've tested it in OpenGL **but not in MayaVi** (cannot run the mayavi module in my environment because of an issue with the wx module, it has nothing to do with espresso).

@KonradBreitsprecher could you please run the MWE in both visualizers as part of the code review?


RudolfWeeber:
@fweik, why did you assign me?

fweik:
Because somebody has to manage this?

RudolfWeeber:
On Tue, Apr 09, 2019 at 05:43:08AM -0700, Florian Weik wrote:
> Because somebody has to manage this?
OK. I understood that differently. Manage, I can do.

KonradBreitsprecher:
Thanks for the MWE, will test it.

jngrad:
Update: the `mayavi2` GUI works in my environment, the issue only comes when espresso makes it call the wx plugin. `pip3 install wxPython --user` fails because gtk+-3.0 is missing.  `pip3 install -f https://extras.wxpython.org/wxPython4/extras/linux/gtk3/ubuntu-18.04/ wxPython --user` worked, but now there is an API error when loading the wx plugin from mayavi in espresso.

RudolfWeeber:
@KonradBreitsprecher ?

KonradBreitsprecher (dismissed):
Tested the openGl part, no more bond 'spiderweb' there.

jngrad:
@RudolfWeeber @KonradBreitsprecher I've removed the MayaVi fix since we cannot know if it works.

mkuron:
Try Ubuntu 16.04 in a virtual machine. That has working mayavi packages.

jngrad:
Tried Ubuntu 16.04 in a VM, mayavi still segfaults. Used mayavi 4.4.3 (from apt), then 4.5.0 (from pip), with traits 4.5.0, then 4.6.0.

mkuron:
The version from Apt used to work about a year ago. Does the minimal example from https://github.com/espressomd/espresso/issues/2704#issuecomment-482069178 also segfault?

mkuron:
By the way, the Mac version does not segfault.

jngrad:
I've used your MWE to test multiple versions of mayavi. Where can I find our VM image for Mac?

RudolfWeeber:
@jngrad, could you please re-introduce the Mayavi fix and test it on someone's Mac at the ICP (@KaiSzuttor, @mkuron).
Then this PR can be finalized.

mkuron (approved):

fweik (approved):

PR #2697, opened 2019-04-09
Removed some code duplication

fweik:
by making some utils math functions usable on the gpu.
Also replaced the macro constants by functions, also for gpu.


KaiSzuttor (dismissed):

fweik:
This does not work with HIP, but the error message is not really helpful...

fweik:
@KaiSzuttor the HIP stuff could be fixed, I also replaced the constant macros that collided with walberla (and should have been replaced anyway).

KaiSzuttor (approved):

PR #2713, opened 2019-04-11
boost::optional for LB philox counter

KaiSzuttor:


src/core/utils/Counter.hpp
```diff
@@ -8,10 +8,7 @@ template <typename T> class Counter {
 private:
   T m_val;
   T m_initial;
-
-private:
   friend class boost::serialization::access;
```
fweik, line 11: If the method is no longer private the fried declaration can be removed too.

KaiSzuttor: it's actually still private, I just removed a redundant ```private```

fweik (dismissed):

fweik:
@KaiSzuttor please look at this, seems to be a genuine error

fweik (approved):

PR #2714, opened 2019-04-11
WIP: core: particle_data: Getter for particle mass, set mass to zero for v…

### LINKED ISSUE #2276: mass-dependent observables shouldn't include virtual particles
which is wrong, because they are not degrees of freedom and do not carry momentum or mass.

fweik:
…irtual particles

~~@jonaslandsgesell needs a test~~

Fixes #2276.


jonaslandsgesell:
Can you please put the following line in the testcase virtual_sites_relative.py:

```
             # is the virtual particle mass 0
             self.assertEqual(system.part[cur_id].mass, 0)
```

below 

```
            # Was the particle made virtual
            self.assertEqual(system.part[cur_id].virtual, 1)
```

That would check if the particle mass which is returned is zero.

Just out of curiosity. What is the impact on the runtime?

fweik:
@RudolfWeeber if we do this and want to be consistent we have to also remove the possibility to thermostat virtual sites. The is no consistent way of Langevin dynamics with massless particles.

fweik:
@jonaslandsgesell can you please add the test?

fweik:
According to offline discussion the Langevin terms in espresso are independent of mass, so nothing else has to be changed.

jonaslandsgesell:
Can we document somewhere, why virtual particles have mass 0?

fweik:
Of course you can.

jonaslandsgesell:
Just think about whether people will know why the mass is 0 in 10 years, when you are not around anymore.

fweik:
We decided not to go forward with this due to its odd interface behavior.

PR #2717, opened 2019-04-12
Widom insertion test

jrfinkbeiner:


testsuite/python/widom_insertion.py
```diff
@@ -0,0 +1,105 @@
...
+    N0 = 1
+    temperature = 0.5
+    type_HA = 0
+    charge_HA = 0
+    lj_eps = 1.0
+    lj_sig = 1.0
+    lj_cut = 5.0
+    box_l = 20.0
+
+    target_mu_ex = -0.00227888  # analytical solution for this specific system
```
jonaslandsgesell: Can you actually do the calculation numerically in code instead of putting a magic number here? It would be better documented this way. Computation time for a weighted summation is marginal with numpy.

testsuite/python/widom_insertion.py
```diff
@@ -0,0 +1,105 @@
...
+        num_samples = 100000
+        for i in range(num_samples):
+            Widom.measure_excess_chemical_potential(
+                0)  # 0 for insertion reaction
+        mu_ex = Widom.measure_excess_chemical_potential(0)
+        deviation_mu_ex = abs(mu_ex[0] - target_mu_ex)
+        dev_comp_to_err = deviation_mu_ex - 1.5 * mu_ex[1] 
+        # error
+        self.assertLess(
+            dev_comp_to_err,
```
jonaslandsgesell: I would just use the deviation_mu_ex to judge whether the method is giving correct results. Think about num_samples=1, then the standard error mu_ex[1]  is high, but this scenario does not help you to find out whether your algorihtm correctly works or not.

testsuite/python/widom_insertion.py
```diff
@@ -0,0 +1,105 @@
...
+
+    target_mu_ex = -0.00227888  # analytical solution for this specific system
+
+    system = espressomd.System(box_l=np.ones(3) * box_l)
+    system.seed = system.cell_system.get_state()['n_nodes'] * [2]
+    np.random.seed(69)  # make reaction code fully deterministic
+    system.cell_system.skin = 0.4
+    volume = np.prod(system.box_l)  # cuboid box
+    
+    unimportant_exclusion_radius = 1.0
```
jonaslandsgesell: Yes widom insertion does not make use of the exclusion radius. So lets just put np.nan as argument for the exclusion radius (I plan to remove this argument from the widom insertion constructor).

jonaslandsgesell (changes requested):
Thanks Jan :) You can drop by if you have questions.

testsuite/python/widom_insertion.py
```diff
@@ -0,0 +1,131 @@
...
+import unittest as ut
+import numpy as np
+import espressomd  # pylint: disable=import-error
+from espressomd import reaction_ensemble
+from tests_common import lj_potential
+
+
+@ut.skipIf(not espressomd.has_features(["LENNARD_JONES"]),
+           "Features not available, skipping test!")
+class ReactionEnsembleTest(ut.TestCase):
```
KaiSzuttor: A better name is probably ```WidomInsertion```

testsuite/python/widom_insertion.py
```diff
@@ -0,0 +1,131 @@
...
+                                       # calculate excess chemical potential of
+                                       # the system from the integral values
+
+    system = espressomd.System(box_l=np.ones(3) * box_l)
+    system.seed = system.cell_system.get_state()['n_nodes'] * [2]
+    np.random.seed(69)  # make reaction code fully deterministic
+    system.cell_system.skin = 0.4
+    volume = np.prod(system.box_l)  # cuboid box
+    
+    unimportant_exclusion_radius = np.nan
```
KaiSzuttor: if it's not important, don't give it a name...

testsuite/python/widom_insertion.py
```diff
@@ -0,0 +1,131 @@
...
+    temperature = 0.5
+    type_HA = 0
+    charge_HA = 0
+    lj_eps = 1.0
+    lj_sig = 1.0
+    lj_cut = 5.0
+    box_l = 20.0
+    lj_shift = lj_potential(lj_cut, lj_eps, lj_sig, lj_cut + 1, 0.0)
+
+    def lj_potential_arr(r, lj_eps, lj_sig, lj_shift):
```
KaiSzuttor: could you please adapt the ```lj_potential``` function defined in ```tests_common.py``` for that. Maybe there are some customers for that.

testsuite/python/widom_insertion.py
```diff
@@ -0,0 +1,131 @@
...
+class ReactionEnsembleTest(ut.TestCase):
+
+    """Test the implementation of the widom insertion (part of the reaction ensemble)."""
+
+    # The excess chemical potential is calculated for identical particles into
+    # a 20 cubed box with a single particle, interacting via a LJ-potential
+    # (cut-off at 5 sigma).
+    N0 = 1
+    temperature = 0.5
+    type_HA = 0
```
KaiSzuttor: Constants are normally written all-capital in python according to [pep8](https://www.python.org/dev/peps/pep-0008/#id48)

testsuite/python/widom_insertion.py
```diff
@@ -0,0 +1,131 @@
...
+from tests_common import lj_potential
+
+
+@ut.skipIf(not espressomd.has_features(["LENNARD_JONES"]),
+           "Features not available, skipping test!")
+class ReactionEnsembleTest(ut.TestCase):
+
+    """Test the implementation of the widom insertion (part of the reaction ensemble)."""
+
+    # The excess chemical potential is calculated for identical particles into
```
KaiSzuttor: This comment should go into the docstring of the testclass

testsuite/python/widom_insertion.py
```diff
@@ -0,0 +1,131 @@
...
+    def lj_potential_arr(r, lj_eps, lj_sig, lj_shift):
+        """lennard jones potential (that works for array input)"""
+        r6 = r**(-6)
+        r12 = r6**2
+        ljsig6 = lj_sig**6
+        ljsig12 = ljsig6**2
+        return 4.0 * lj_eps * (ljsig12 * r12 - ljsig6 * r6) + lj_shift
+
+    r = np.linspace(1e-10, lj_cut, 1000)
+    integrateSphere = 4 * np.pi * np.trapz(
```
KaiSzuttor: can you come up with a self-explanatory variable name?

testsuite/python/widom_insertion.py
```diff
@@ -0,0 +1,131 @@
...
+        system = ReactionEnsembleTest.system
+        Widom = ReactionEnsembleTest.Widom
+        target_mu_ex = ReactionEnsembleTest.target_mu_ex
+
+        system.seed = system.cell_system.get_state()[
+            'n_nodes'] * [np.random.randint(5)]
+        num_samples = 100000
+        for i in range(num_samples):
+            Widom.measure_excess_chemical_potential(
+                0)  # 0 for insertion reaction
```
KaiSzuttor, line 111: what is a insertion reaction?

testsuite/python/widom_insertion.py
```diff
@@ -0,0 +1,131 @@
...
+    system.cell_system.skin = 0.4
+    volume = np.prod(system.box_l)  # cuboid box
+    
+    unimportant_exclusion_radius = np.nan
+    Widom = reaction_ensemble.WidomInsertion(
+        temperature=temperature,
+        exclusion_radius=unimportant_exclusion_radius)    
+
+    @classmethod
+    def setUpClass(cls):
```
KaiSzuttor: why do you use a class method here? you can just use the ```setUp``` method of the ```unittest``` module

testsuite/python/widom_insertion.py
```diff
@@ -0,0 +1,131 @@
...
+
+    r = np.linspace(1e-10, lj_cut, 1000)
+    integrateSphere = 4 * np.pi * np.trapz(
+        r**2 * np.exp(-lj_potential_arr(r,
+                                        lj_eps,
+                                        lj_sig,
+                                        lj_shift) / temperature),
+        x=r)  # numerical solution 
+    integreateRest = (box_l**3 - 4.0 / 3.0 * np.pi * lj_cut**3)
+                      # numerical solution for V_lj=0 => corresponds to the
```
KaiSzuttor: could you please move the comments above the code snippets

KaiSzuttor (commented):
Thanks @jrfinkbeiner, I added some comments, mostly concerning style.

testsuite/python/widom_insertion.py
jonaslandsgesell: It means you measure the excess chemical potential via inserting a particle, and not via removing one particle (Wich you can also do in a many particle system, i.e. the thermodynamic limit). This you would do via supplying a 1 instead of a zero.

testsuite/python/widom_insertion.py
jonaslandsgesell: It will be removed in the new interface so nobody is confused (other PR)

KaiSzuttor: how is this connected? just put np.nan in the argument instead of a variable

testsuite/python/widom_insertion.py
KaiSzuttor: comments that need a long explanation should maybe go. rather put a link to some documentation or reference then

testsuite/python/widom_insertion.py
jonaslandsgesell: This comment refers to an peculiarity of this test. You need to explain somewhere why the numerical calculation which is presented here is correct. I plan to introduce an even longer comment ;) and this goes not to the documentation in my mind

KaiSzuttor: yes, just move the comments above the code snippets

jonaslandsgesell (approved):

testsuite/python/widom_insertion.py
fweik: I think this is more an issue about the interface of thee class than the test. @KaiSzuttor if so you should open an issue about that, but I don't think this blocks this PR.

fweik (approved):

PR #2719, opened 2019-04-12
BH-gpu magnetostatics: multiple CPU support and refactoring

### LINKED ISSUE #2500: DipolarBarnesHut does not work with multiple cores

bogdan-tanygin:
Fixes #2500
In general, this fix is important cause it allows a user to parallelize both CPU and GPU calculations: EOM integration and magnetostatics respectively.

Description of changes:
 - Correct dipole prefactor broadcasting among the nodes
 - New test dds-and-bh-gpu.py is running over 4 CPUs: a comparison between BH and DDS gpu calculations. A new test has been needed because we cannot use dds-cpu as a baseline cause it does not support multiple CPUs. 
- Some technical debt has been caught up here as well: smart pointers based BH, the AMD GPU skip cleanup, per node test seeding. 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


fweik (approved):

fweik:
bors retry

PR #2722, opened 2019-04-13
Coulomb cleanup

fweik:
Some modernization of the coulomb interface:
- Removed unneeded parameters
- Split central and non-central forces (relevant for pressure calculation)
- Turned some output parameters into return values
- More signature consistency across the methods
- Replaced {coulomb,dipole}_cutoff globals


fweik:
@KonradBreitsprecher could you please have a look at the thole stuff? The old implementation seemed wrong to me, because the force calculation only removed interaction according to the thole charge fraction, but the energy calculation removed the whole interaction. Also the checks were not run.

KonradBreitsprecher (approved):
The thole calculation seems fine, energy is untested however. Will add issue...

fweik:
Yes, but the test fails :-)

src/core/nonbonded_interactions/nonbonded_interaction_data.cpp
```diff
@@ -286,11 +275,6 @@ static void recalc_maximal_cutoff_nonbonded() {
       if (max_cut_current < data->REACTION_range)
         max_cut_current = data->REACTION_range;
 #endif
-#ifdef THOLE
```
KonradBreitsprecher: Removing that shuts down the thole calculation completely.
I guess it should be
```
#ifdef THOLE    
      // If THOLE is active, use Coulomb cutoff 
      if (data->THOLE_scaling_coeff != 0)
        max_cut_current = std::max(max_cut_current, Coulomb::cutoff(box_l));
#endif
```
thole/drude tests worked with it.

KonradBreitsprecher (changes requested):

src/core/nonbonded_interactions/nonbonded_interaction_data.cpp
fweik: ~~This is only needed is thole should work for uncharged particles (the cutoff of the active electrostatic method is set anyway, but is not considered for uncharged pairs). I'm not sure if this should work, if not the a relative charge fraction could be used instead of the charges.~~

fweik: Works now, thank you.

RudolfWeeber (approved):

PR #2726, opened 2019-04-15
Sanity check for velocity less than LB Mach limits

### LINKED ISSUE #2702: Sanity check for LB parameters
The LB parameters prescribed for a simulation should be sanity-checked i.e. it should be ensured that these parameters are within the LB stability and Mach number limits. This should be tested in the core.

jainka:
Fixes #2702

Fixes #

Description of changes:
 - 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


src/core/grid_based_algorithms/lb_interface.cpp
```diff
@@ -1542,3 +1545,16 @@ Vector3d lb_lbfluid_calc_fluid_momentum() {
...
 }
 #endif
+
+/*** Sanity check if the velocity defined at LB boundaries is within the Mach
+number limits of the scheme i.e. u < 0.3**************************************/
+
+void lb_boundary_mach_check() {
+  for (auto it = 0; it < LBBoundaries::lbboundaries.size(); ++it) {
+    for (int v = 0; v < 3; ++v) {
+      if (LBBoundaries::lbboundaries[it]->velocity()[v] > 0.3) {
```
fweik: Shouldn't this check the absolute value? Also please check what units the boundary velocity is in, if I recall it correctly it's in MD unites, in which case a unit version would be needed here.

KaiSzuttor:
@jainka  could you please have a look at the CI failures and the comments

jainka:
> @jainka could you please have a look at the CI failures and the comments

The CI checks failed due to the failure in interpolation test. This PR will go in only after the fix #2729 is merged.

src/core/grid_based_algorithms/lb_interface.cpp
jainka: both the observations are right; I did not realize that. Will correct it.

KaiSzuttor:
This PR does not depend on #2729. Also the failure in [clang6](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/98781) is of different kind. The failures are most likely due to the MD-LB unit conversion that @fweik mentioned.

KaiSzuttor:
@jainka could you please have a look at the failing checks? e.g. https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/99338

jngrad:
@jainka please write `#endif // LB` if it was guarding LB, so that future contributors won't run into the same issue.

src/core/grid_based_algorithms/lb_interface.cpp
```diff
@@ -208,15 +208,20 @@ void lb_lbfluid_propagate() {
...
+      if (lb_reinit_particles_gpu()) {
+        lb_realloc_particles_gpu();
+        lb_reinit_particles_gpu.validate();
+      }
     }
 #endif
   } else if (lattice_switch == ActiveLB::CPU) {
 #ifdef LB
     lb_sanity_checks();
+    lb_boundary_mach_check();
```
KaiSzuttor, line 244: I'd prefer to have a central interface function like ```lb_lbfluid_sanity_checks``` that decides which functions to call (either CPU or GPU related)

KaiSzuttor (changes requested):

KaiSzuttor (dismissed):

KaiSzuttor (approved):

PR #2748, opened 2019-04-23
Additional LB test cases

RudolfWeeber:
* Buyancy force on a sphere in a box of fluid under gravity
* Momentum conservation in a fluid with an external volume force applied and a particle with an oposite force applied

@KaiSzuttor @mkuron Could you please take a look at the
lb_buoyancy_force test? It only passes, when the fluid density is set to 1.
On the other hand, the lb_shear and lb_stokes_sphere tests check boundary forces which do not (and should not) depend on the density. They pass when run with different densities.
Also, the application of an external volume force is tested in lb.py in test_a_ext_force...
Furthermore, to the lb_momentum_conservation.py test checks the balance between forces applied to the lb and those applied to a particle.


testsuite/python/lb_buoyancy_force.py
```diff
@@ -0,0 +1,117 @@
...
+        for i in range(3):
+            n = np.zeros(3)
+            n[i] = 1
+            self.system.lbboundaries.add(
+                lbboundaries.LBBoundary(shape=shapes.Wall(
+                                        normal=-n, dist=-(self.system.box_l[i] - AGRID))))
+
+            self.system.lbboundaries.add(lbboundaries.LBBoundary(
+                                         shape=shapes.Wall(
+                                         normal=--n, dist=AGRID)))
```
fweik: should be `normal=n` ?

testsuite/python/lb_momentum_conservation.py
```diff
@@ -0,0 +1,95 @@
...
+            self.system.integrator.run(100)
+
+            # Check that particle momentum =-fluid momenum
+            # up to the momentum trnasferred in 1/2 time step
+            np.testing.assert_allclose(
+                p.v * p.mass, -
+                    np.array(
+                        self.system.analysis.analyze_linear_momentum(
+                            include_particles=False)),
+                atol=np.linalg.norm(applied_force) * TIME_STEP * 0.55)
```
mkuron: What's this 0.55?

RudolfWeeber:
On Tue, Apr 30, 2019 at 07:57:24AM -0700, Michael Kuron wrote:
> mkuron commented on this pull request.
> 
> 
> 
> > +
> +        # Reach steady state
> +        self.system.integrator.run(700)
> +        v_final = np.copy(p.v)
> +
> +        for i in range(3):
> +            self.system.integrator.run(100)
> +
> +            # Check that particle momentum =-fluid momenum
> +            # up to the momentum trnasferred in 1/2 time step
> +            np.testing.assert_allclose(
> +                p.v * p.mass, -
> +                    np.array(
> +                        self.system.analysis.analyze_linear_momentum(
> +                            include_particles=False)),
> +                atol=np.linalg.norm(applied_force) * TIME_STEP * 0.55)
> 
> What's this 0.55?
In the currently implemented integration scheme, MD and LB momentum seem to be 1/2 time step apart. The difference between 0.5 and 0.55 is numeric tolerance.

fweik (approved):

fweik:
So @jngrad should this now be rebased? Most of the commits are broken.

jngrad:
The two merge commits will make the rebasing/squashing difficult. Not sure if this is worth the effort. I would merge it directly.

testsuite/python/lb_momentum_conservation.py
```diff
@@ -22,13 +22,13 @@
...
 # Define the LB Parameters
-TIME_STEP = 0.01
-AGRID = 0.5 
-KVISC = 6 
-DENS = 0.5
-
-F = 0.013
-BOX_SIZE = 10 * AGRID
+TIME_STEP = 0.1
+AGRID = 1.0
```
KaiSzuttor, line 26: Did you use ```AGRID``` equal to unity on purpose?

PR #2749, opened 2019-04-23
Observable for stress contribution due to friction

bindgens1:
Fixes #

Description of changes:
Added observable and test case for the friction contribution to the stress in a DPD fluid. 
Only works on one node, mpi reduce is still missing. 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


src/core/observables/DPDStress.hpp
```diff
@@ -0,0 +1,22 @@
...
+#include "pressure.hpp"
+#include <vector>
+
+namespace Observables {
+
+class DPDStress : public Observable {
+public:
+  int n_values() const override { return 9; };
+  std::vector<double> operator()(PartCfg &partCfg) const override {
+    std::vector<double> res(n_values());
```
fweik: Why is this needed?

src/core/pressure.cpp
```diff
@@ -171,6 +171,28 @@ void pressure_calc(double *result, double *result_t, double *result_nb,
              p_tensor_non_bonded.data_nb.n, MPI_DOUBLE, MPI_SUM, 0, comm_cart);
 }
 
+const Vector9d dpd_stress() {
+  Vector9d dpd_stress = Vector9d::broadcast(0);
```
fweik: You can just write `Vector3d dpd_stress = {};`

src/core/pressure.cpp
```diff
@@ -171,6 +171,28 @@ void pressure_calc(double *result, double *result_t, double *result_nb,
...
+
+  if (max_cut > 0) {
+    short_range_loop(Utils::NoOp{},
+      [&dpd_stress](Particle &p1, Particle &p2, Distance &d) {
+      Vector3d f;
+      IA_parameters *ia_params = get_ia_param(p1.p.type, p2.p.type);
+      f = dpd_pair_force(&p1, &p2, ia_params, d.vec21.data(), sqrt(d.dist2), d.dist2);
```
fweik: Please use `auto const f = ...` and remove the previous declaration of `f`.

fweik (changes requested):
Except for some minor points this looks good to me

src/core/pressure.cpp
```diff
@@ -171,6 +171,28 @@ void pressure_calc(double *result, double *result_t, double *result_nb,
              p_tensor_non_bonded.data_nb.n, MPI_DOUBLE, MPI_SUM, 0, comm_cart);
 }
 
+const Vector9d dpd_stress() {
```
fweik: Please move this to `dpd.cpp`

src/core/pressure.hpp
```diff
@@ -67,6 +67,8 @@ int observable_compute_stress_tensor(int v_comp, double *A);
 
 void update_pressure(int v_comp);
 
+const Vector<double, 9> dpd_stress();
```
fweik: Please move this to `dpd.hpp`

src/core/observables/DPDStress.hpp
bindgens1: I actually looked at `observables/StressTensor.hpp` to get an example.

bindgens1:
@pdebuyl 
I think this is now working as expected

@fweik 
Any concerns?

fweik:
Sorry, it was quicker to just fix the points I had.

fweik:
@RudolfWeeber could you please have a look, especially if the test is correct.

pdebuyl:
@fweik in 87cd1b0afe8bc5af7168d2823b99fc78ec85ae9c the computation changes from "during regular force computation" to a repeat of the `short_range_loop`.

We decided to compute the DPD stress during the regular call to DPD force to make sure that we were using the correct random number. Can we be sure that this is working properly in the current state of the observable?

RudolfWeeber:
On Thu, May 09, 2019 at 09:17:45AM -0700, Pierre de Buyl wrote:
> @fweik in 87cd1b0afe8bc5af7168d2823b99fc78ec85ae9c the computation changes from "during regular force computation" to a repeat of the `short_range_loop`.
> 
> We decided to compute the DPD stress during the regular call to DPD force to make sure that we were using the correct random number. Can we be sure that this is working properly in the current state of the observable?
Stress calculation and force calculation runs should not be entangled.
Getting the same random numbers would b accomplished by using a counter based random number generator in DPD. This is being worked on, afaik.

fweik:
There is multiple issues here:
1. You can not use the DPD forces from the force calculation, because we are doing
    Velocity Verlet, and the forces are calculated with velocities between time steps,
    so the result would not be consistent with the other observables.
2. In light of that: why did the test past with your implementation?
3. Is the instantaneous stress of interest to you? If you are averaging in the end
    it just adds noise, and can be left out.

bindgens1:
The test passes because there is a manual integrator.run(steps=0, recalculate_forces=True) to guarantee that everything is at a defined stage and up-to-date in python. Otherwise the test doesn't pass and I can't imagine designing a test case with velocities from the python interface at half a time step.

I think the version like this is nice because it avoids any issues with the random numbers. Furthermore: Are there any customers besides Pierre and me for DPD? I think, one could write a statement for the user guide to tell people what's going on concerning the half time step issue. From the outside nothing has changed and it has the benefit to not rerun everything. With the high number densities required in DPD simulations this is really a benefit.

Otherwise we could go back to the version I proposed in the first place ([here](https://github.com/espressomd/espresso/pull/2749/commits/43e0d7ec95a71f77785fae3f1ac391147d66dd59)). However, there were issues to get it running on multiple nodes.

fweik:
@bindgens1 Ok, now it's not calculating the noise, and does not affect the state of the system. Does this work for you now?

@RudolfWeeber please have a look.  I've also cleaned it up so that I can actually understand what the kernel is doing. It's not perfect obviously but I still think it's an improvement.

fweik (dismissed):

bindgens1:
I think, it should be possible to measure some of the results presented in https://aip.scitation.org/doi/pdf/10.1063/1.1883163?class=pdf

However, they also have plots for the viscosity at different noise amplitudes. I know that the dissipative forces and the noise are connected via fluctuation dissipation (???) but it is not clear to me, if we can replicate this data with your implementation. Thus, I would suggest to give it a try.

@pdebuyl What do you think?

fweik:
I am not sure that I follow. Maybe I am really stupid here, but from my understanding there is no contribution of the noise to the expectation value of the stress tensor, it's sampling zero-mean white noise on top of the dissipative part?

bindgens1:
> I am not sure that I follow. Maybe I am really stupid here, but from my understanding there is no contribution of the noise to the expectation value of the stress tensor, it's sampling zero-mean white noise on top of the dissipative part?

I totally agree on that. There should be no contribution of the noise on average when calculating the stress. I rechecked in the paper and their noise amplitude includes the friction coefficient. Thus the additional contribution in due to the friction and should be included in the dissipative part. Sorry, I am stupid!

fweik:
You also have to think about correlation functions, but I *think* it does not contribute there either. Also as you mentioned the FDT kind of says that the noise does not add new information. Technically it is not difficult to add the noise again, but I *think* it is not a good idea.

bindgens1:
> You also have to think about correlation functions, but I think it does not contribute there either. Also as you mentioned the FDT kind of says that the noise does not add new information. Technically it is not difficult to add the noise again, but I think it is not a good idea.

I was confused by this plot. The noise just causes the error bars to get bigger. The shift in the mean values should be due to friction. So it should be sufficient as it is.

image

fweik:
The value of the dissipative contributions are also influenced by the noise amplitude (e.g. if you increase the kT at fixed gamma). Just because there is not _additional_ noise added to the observable does not mean that a cold system is sampled. Also the disspative part will of course fluctuate. The graph that you are showing is for the viscosity, which is related to the stress by a Green-Kubo relation (it's the integral of the stress acf), that is why it's not enough when it is zero on average, it also has to be uncorrelated, which I think it is.

bindgens1:
I think we are on the same page then. The best is probably to put it to a test to check, if we can replicate the results of the paper. As an initial try, Green Kubo in an unsheared system should be fine. If this works out, we can use it to test the LE.

fweik:
@RudolfWeeber

pdebuyl:
We just reviewed the observable together with @bindgens1 and are glad with the current version :-)

We did preliminary test for the value of the Green-Kubo viscosity and things seem ok.

bindgens1:
Hey!

Anybody an idea why intel18 compiler is failing? I couldn't find the issue in the code.
CylindricalLBFluxDensityProfileAtParticlePositions.cpp seems to be the same for python and this branch.

RudolfWeeber:
Maybe, the evaluate(partcfg&) method in PidObservable should be protected rather than private?

RudolfWeeber (dismissed):
Once the CI passes, we can go ahead with this.

fweik:
They should all be private. But I don't think the compiler is right here...

bindgens1:
@pdebuyl and I tried the intel compiler on the cluster in Leuven. Even with the same compile flags as you use we didn't get it to fail. We didn't change anything in this file and are unsure what to do.

fweik:
I think it has to spread, because the warning is caused in the derived class.

bindgens1:
Thanks a lot for taking care of this!

For the remaining issues:
The CI tests that are failing are not restricted to this specific branch. The fails occur in fails that we didn't touch and are also present in the other pull request.

bindgens1:
This looks like a problem fo the CI for me.

fweik (dismissed):

bindgens1:
@fweik 

I tried to fix the merge regression but unsure if this is the way to go.
It builds and the tests pass on my PC. Please have a look, if this is done correctly.

fweik (approved):

fweik:
bors cancel

pdebuyl:
:tada:

PR #2753, opened 2019-04-25
Testsuite: Speedup for sanitizer builds

RudolfWeeber:


testsuite/python/lb_interpolation.py
```diff
@@ -74,7 +74,7 @@ def test_interpolated_velocity(self):
 
         """
         self.set_boundaries()
-        self.system.integrator.run(3000)
+        self.system.integrator.run(1800)
```
KaiSzuttor: you have to increase this number...

fweik:
@RudolfWeeber can you please address the one remaining problem? The faster we can speed this up the better...

RudolfWeeber:
I'd have expected the lb interpolation test to be deterministic...
I now pushed a version which even passes when the tolerance is lowerd by a factor of 10 - on my machine, that is...

fweik:
It looked deterministic to me: the deviation was exactly the same in all jobs.

fweik:
If PARTICLE_ANISOTROPY is disabled you are comparing a vector to a scalar in `  File "/builds/espressomd/espresso/build/testsuite/python/langevin_thermostat.py", line 145, in test_02__friction_trans`, this is why this test fails with empty config.

fweik (approved):

PR #2768, opened 2019-04-27
CI: Don't override number of build processes on cuda builds

RudolfWeeber:
This resulted in an over-subscription of available cores.


fweik:
Why does this script exist at all? To call `nvidia-smi`?

fweik (approved):

PR #2772, opened 2019-04-29
CI: mark files as up to date for sphinx/doxygen build.

### LINKED ISSUE #2764: check_sphinx runs the build again
See https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/104578,
from my understanding of the config it should get the the build folder from
the `cuda-maxset` job by artifact, but apparently this is not working.

KaiSzuttor:
Fixes #2764

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -249,10 +249,6 @@ cuda-no-gpu:
   script:
     - export myconfig=maxset hide_gpu=true python_version=3 test_timeout=900
     - bash maintainer/cuda_build.sh
-  artifacts:
-    paths:
-    - build/
-    expire_in: 1 week
   tags:
     - docker
     - linux
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -393,7 +393,7 @@ check_sphinx:
   when: on_success
   script:
     - cd ${CI_PROJECT_DIR}/build
-    - make sphinx
+    - make -t && make sphinx
   artifacts:
     paths:
     - build/doc/sphinx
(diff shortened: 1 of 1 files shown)


RudolfWeeber:
Once this works, several other jobs can make use of it, e.g., 

cuda-no-gpu

and the odd number of cores job

KaiSzuttor:
@RudolfWeeber  ```cuda-no-gpu``` is a weird CI job, it stores the build directory, but no other job depends on this artifact. I cannot see what this has to do with the timestamp issue in sphinx/doxygen

KaiSzuttor:
also you cannot have dependencies within a stage and we observed that waiting for the whole stage just to omit the build is actually slower at the end.

fweik (approved):
👍

fweik:
I think the no-gpu job can move to the additional checks stage and use the cuda-maxset artifacts, instead of running the same build twice

RudolfWeeber:
On Mon, Apr 29, 2019 at 01:50:57AM -0700, Kai Szuttor wrote:
> @RudolfWeeber  ```cuda-no-gpu``` is a weird CI job, it stores the build directory, but no other job depends on this artifact. I cannot see what this has to do with the timestamp issue in sphinx/doxygen
That job storing an artefact was a copy and past error on my part.
This job (and the odd number of cores job) can re-use the cuda-maxset build and just run the tests.

KaiSzuttor:
yes, but you would have to wait for the "build" stage to finish, although you might have enough resources to start the jobs you mentioned. As history taught us this approach is less efficient.

PR #2774, opened 2019-04-29
Use a single build script

jngrad:
The `cuda_build.sh` used to be a separate build script for GPU jobs, but it is now just a two-line wrapper for `build_cmake.sh`. It has been in that state for almost 2 years and keeps getting thinned out, to the point where it would make more sense to merge both files.

### DIFF
--- a/maintainer/CI/build_cmake.sh
+++ b/maintainer/CI/build_cmake.sh
@@ -87,7 +87,7 @@ if [ -z "$cxx_flags" ]; then
     fi
 fi
 
-if [[ ! -z ${with_coverage+x} ]]; then
+if [ ! -z ${with_coverage+x} ]; then
   bash <(curl -s https://codecov.io/env) &> /dev/null;
 fi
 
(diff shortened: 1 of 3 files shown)


maintainer/CI/build_cmake.sh
```diff
@@ -42,6 +42,8 @@ function end {
 
 # execute and output a command
 # handle environment variables
+[ -z "$cuda_job" ] && cuda_job="false"
+[ $cuda_job = "true" ] && [ -n "${CI_PROJECT_DIR}" ] && srcdir="${CI_PROJECT_DIR}"
```
fweik, line 46: Why is this needed?

jngrad: `cuda_build.sh` overrides the user-provided `$srcdir` with `${CI_PROJECT_DIR}`, but if the script is not run in CI, `$srcdir` becomes an empty string, then gets the value `pwd`.
This behavior can be confusing when running the script on a local machine, so I added an extra `[ -n "${CI_PROJECT_DIR}" ]` to avoid overriding `$srcdir` when the script is not run in CI.

fweik: Yes, why does it overwrite srcdir? Isn't that just the default?

jngrad: It probably has to do with the documentation jobs (`deploy_sphinx_documentation`, `check_sphinx`, etc.) which start with the command `cd ${CI_PROJECT_DIR}/build`. We could explicitly pass `srcdir="${CI_PROJECT_DIR}"` to the job from which the documentation jobs depend and drop the overwriting line from the build script.

fweik (approved):

PR #2775, opened 2019-04-29
Fix formulas in doxygen

KaiSzuttor:


jngrad:
I've been using a local copy of MathJax 2.7.5 for a while, however for a live website, shouldn't we use a rolling version instead, i.e. `https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.5/latest.js`, as shown in the [MathJax docs](http://docs.mathjax.org/en/latest/start.html#using-a-content-delivery-network-cdn)? This would require a sed expression to replace `MathJax.js` by `latest.js`.

jngrad:
Example:
```bash
mathjax_cdn=`grep '^MATHJAX_RELPATH' ${DOXYGEN_OUT} | sed -r 's|^MATHJAX_RELPATH\s*=\s*(.+)/?\s*$|\1|'`
mathjax_cdn_old="src=\"${mathjax_cdn}/MathJax.js"
mathjax_cdn_new="src=\"${mathjax_cdn}/latest.js"
sed -ri "s|${mathjax_cdn_old}|${mathjax_cdn_new}|" doc/doxygen/html/*.html
```

(If the URL was declared in CMake, the block could be simplified to just the sed line as a CMake target.)

To test it:
```bash
sed -r "s|${mathjax_cdn_old}|${mathjax_cdn_new}|" doc/doxygen/html/bondedIA.html > doc/doxygen/html/bondedIA.html.bak
head -n24 doc/doxygen/html/bondedIA.html*
```

For example, using `.../libs/mathjax/2.7.4/latest.js` correctly redirects to 2.7.5.

Other than that, everything's fine.

KaiSzuttor:
is there any reason why we cannot fix the version and update if necessary?

jngrad:
We do not need to look out for MathJax bugfix and browser-compatibility updates when there is a redirect. This is the approach taken by the Sphinx team. It's up to you.

jngrad (approved):

PR #2778, opened 2019-04-30
Make call to nvidia-smi conditional on cuda tag

### LINKED ISSUE #2776: CI build failed for merged PR
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/pipelines/7001

jngrad:
Fixes #2776 (regression introduced by #2774)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -175,7 +175,7 @@ cuda-maxset:
   stage: build
   image: gitlab.icp.uni-stuttgart.de:4567/espressomd/docker/cuda:9.0
   script:
-    - export cuda_job=true myconfig=maxset with_coverage=true python_version=3 test_timeout=900
+    - export cuda_job=true myconfig=maxset with_coverage=true python_version=3 test_timeout=900 srcdir=${CI_PROJECT_DIR}
     - bash maintainer/CI/build_cmake.sh
   artifacts:
     paths:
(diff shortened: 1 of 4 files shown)


fweik (approved):

fweik (changes requested):
It think this most robust way is to call it only when it exists, like so:
```bash
command -v nvidia-smi && nvidia-smi
```

jngrad:
But then the nvidia-smi splash screen will be displayed on all jobs, even those that don't require or explicitly forbid the use of a GPU, if the runner happens to have a GPU card.

KaiSzuttor:
@jngrad that does not matter, it's for debugging purposes only

jngrad:
Nevermind, didn't see all non-gpu jobs just failed. The conditional had no effect.

fweik (dismissed):

fweik:
@jngrad the active matter tutorial test failed in staging. Could you please look at this, it does not seem to be deterministic, but it should be if I understand your testing harness correctly... (https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/106664)

jngrad:
did you cancel it?

fweik:
We both did restart it and then I canceled one, there is still a job running it is just not correctly shown in the UI. Should be fine

fweik:
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/106680

jngrad:
the CI interface is confusing, on this run from the same pipeline it shows cancel:
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/106681
I'm currently fixing the tutorial

jngrad:
ok, makes sense
I increased the sampling time for the tutorial test that failed and decreased the sensitivity of a sample test that failed in another PR last week.

fweik (approved):

PR #2781, opened 2019-04-30
Add support for gdb in ipypresso

hirschsn:
Ipypresso's flag "--gdb" does not work on my system because ipypresso is just a wrapper script executing python with a bunch of parameters. Nevertheless, support of --gdb can be added to ipypresso by a slightly more complicated gdb setup.

This PR adds this support using gdb's exec-wrapper functionality. The implementation is quite hacky by inserting a check for ipython into pypresso.cmakein. Let me know if the functionality itself is something that you want to have upstream. If so, we also could figure out a less hacky solution.

Description of changes:
 - Use ipython as gdb's exec-wrapper, not directly as executable

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?

### DIFF
--- a/src/python/pypresso.cmakein
+++ b/src/python/pypresso.cmakein
@@ -54,7 +54,7 @@ fi
 case $1 in
     --gdb)
         shift
-        expr match @PYTHON_FRONTEND@ '.*ipython.*' && exec gdb -ex "set print thread-events off" -ex "set exec-wrapper sh -c 'exec /usr/bin/ipython \"$@\"'" --args /usr/bin/python $@
+        expr match @PYTHON_FRONTEND@ '.*ipython.*' && exec gdb -ex "set print thread-events off" -ex "set exec-wrapper sh -c 'exec \"@IPYTHON_EXECUTABLE@\" \"$@\"'" --args "@PYTHON_EXECUTABLE@" $@
         exec gdb --args @PYTHON_FRONTEND@ $@
         ;;
     --lldb)
@@ -76,7 +76,7 @@ case $1 in
     --gdb=*)
         options="${1#*=}"
         shift
-        expr match @PYTHON_FRONTEND@ '.*ipython.*' && exec gdb -ex "set print thread-events off" -ex "set exec-wrapper sh -c 'exec /usr/bin/ipython \"$@\"'" ${options} --args /usr/bin/python $@
+        expr match @PYTHON_FRONTEND@ '.*ipython.*' && exec gdb -ex "set print thread-events off" -ex "set exec-wrapper sh -c 'exec \"@IPYTHON_EXECUTABLE@\" \"$@\"'" ${options} --args "@PYTHON_EXECUTABLE@" $@
         exec gdb ${options} --args @PYTHON_FRONTEND@ $@
         ;;
     --lldb=*)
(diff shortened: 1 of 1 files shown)


KaiSzuttor:
@mkuron could you please have a look?

KaiSzuttor:
@mkuron

mkuron (changes requested):
Paths shouldn't be hardcoded. Use the respective CMake variables.

hirschsn:
@mkuron You're right. Is there a possibility to get the path of the python executable from cmake if the frontend is ipython?

mkuron:
There are `IPYTHON_EXECUTABLE` and `PYTHON_EXECUTABLE`.

jngrad:
Here's a MWE to reproduce the bug:
```
./ipypresso --gdb script.py
GNU gdb (Ubuntu 8.1-0ubuntu3) 8.1.0.20180409-git
"/usr/bin/ipython": not in executable format: File format not recognized
(gdb) run
Starting program:  script.py
No executable file specified.
Use the "file" or "exec-file" command.
```
And the result after this PR:
```
./ipypresso --gdb script.py   # assuming script.py fails
GNU gdb (Ubuntu 8.1-0ubuntu3) 8.1.0.20180409-git
Reading symbols from /usr/bin/python2...Reading symbols from /usr/lib/debug/.build-id/fd/3904306c73383fb371287416257b82d6a3363b.debug...done.
done.
(gdb) run
Starting program: /usr/bin/python2 script.py

process 3301854 is executing new program: /usr/bin/python2.7
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[Inferior 1 (process 3301854) exited normally]
```
which is the expected behavior: ipypresso should give the same output as pypresso.

The regex is not necessary and won't work here because there is no guarantee the IPython program is called "ipython". We already support Jupyter as a fallback when IPython is not found, and users are free to override that choice with `cmake -DIPYTHON_EXECUTABLE=~/coding/my-own-ipynb-interpreter/bin ..`

You can simply do the following:
```cmake
    --gdb)
        shift
        [ "@PYTHON_FRONTEND@" = "@IPYTHON_EXECUTABLE@" ] && exec gdb -ex "set print thread-events off" -ex "set exec-wrapper sh -c 'exec @IPYTHON_EXECUTABLE@ \"$@\"'" --args "@PYTHON_EXECUTABLE@" $@
        exec gdb --args @PYTHON_FRONTEND@ $@
```

hirschsn:
@jngrad  Thanks for your suggestion, that's right. Changed it.

jngrad:
With 556501f there is a difference in behavior for ipypresso:
`./ipypresso --gdb` then `(gdb) run script.py`: starts ipython prompt without executing the script
`./ipypresso --gdb script.py` then `(gdb) run`: does not start ipython prompt but executes the script
Probably not related to the changes in this PR: `./pypresso --gdb` sometimes spans multiple threads.

hirschsn:
@jngrad thanks, again, for your thoroughness. Accidentally, the $@ was evaluated too early. Now the sh of the exec-wrapper evaluates it and arguments to run should work.

jngrad (approved):

PR #2783, opened 2019-04-30
Removed LB feature

fweik:
LB does not have a performance impact anymore, so it does not need to be a compile time feature.


KaiSzuttor:
@fweik have you checked that there is now performance impact by means of performance measures before and after this PR?

fweik:
It was a feature because the coupling added to the particle data, this is no longer the case.

KaiSzuttor (approved):

PR #2786, opened 2019-05-01
Replaced LB_GPU by CUDA

fweik:


fweik:
Needs #2787 fixed first.

KaiSzuttor (approved):

PR #2793, opened 2019-05-03
Fix sample lbf.py

### LINKED ISSUE #2792: Sample lbf.py requires EXTERNAL_FORCES
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/pipelines/7069

jngrad:
Fixes #2792

Sample `lbf.py` requires `EXTERNAL_FORCES`

### DIFF
--- a/samples/lbf.py
+++ b/samples/lbf.py
@@ -39,12 +39,12 @@ print("""
 =======================================================
 """)
 
+required_features = ["EXTERNAL_FORCES"]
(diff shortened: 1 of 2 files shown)


fweik (approved):

fweik:
The error did not show up before because it never run in configs without `EXTERNAL_FORCES`, because those also did not have `LB`, if that makes any sense.

PR #2795, opened 2019-05-04
Add more documentation checks

### LINKED ISSUE #2661: More robust documentation checks in CI
The Sphinx and Doxygen documentations keep piling up errors again. These issues are trivial to fix in the PRs where they emerge in the first place, whereas fixing them weeks after a merge involves a fair amount of guesswork (which can go wrong, as explained in the last meeting with the #2623 case) and has the potential to cause merge conflicts in other opened PRs. CI already fails when Sphinx warnings are raised, to some extent. We should make it more robust and do the same for Doxygen.

* Sphinx: the `-W` option turns warnings into errors in the `check_sphinx` container, so issues are difficult to introduce. Unfortunately, some syntax errors are silently ignored by Sphinx. For example, a link to a non-existing Python function won't generate a warning:
  ```rest
  :meth:`espresso.electrostatics.P3M.Tune`
  ```
  There are currently several such examples in the documentation. Sometimes it's just a typo, sometimes the function moved and the paragraph must be updated accordingly. These errors can be spotted with a regex on the HTML output because Sphinx doesn't use line wrapping.
* Doxygen: warnings aren't turned into errors, so the `check_doxygen` container can't fail. Some warnings are not caused by the code and can be filtered out with a regex. `dot` warnings can probably be removed with a command in the Doxyfile. Some errors are silently ignored, though:
  ```c++
  /** @brief Some function
   *  @return Number of neighbors.
   *  @param p  First particle
   *  @param dist
   *  Detailed explanation of the algorithm.
   */
   void foo(Particle &p, double dist);
  ```
  Here the second param has no description, which can be used as a trick to silence a warning about undocumented parameters, because Doxygen will automatically use the subsequent paragraph `Detailed ...` as the description for parameter `dist`. This behavior can be unpredictable: if paragraph `Detailed...` starts with any Doxygen command for paragraph-level markup, the second parameter description remains empty. This error can be spotted with a regex. The superfluous `@return` will raise a warning in a future release of Doxygen.

jngrad:
Closes #2661

Description of changes:
- broken links in Sphinx are now treated as errors
   + mostly incorrect namespaces, typos and incorrect names (e.g. `integer` instead of `int`)
- some Doxygen warnings are now treated as errors
   + warnings caused by known Doxygen bugs are ignored
   + warnings that don't change the html output are ignored
   + only unique warnings are displayed
- developers can use `maintainer/CI/doc_warnings.sh` and `maintainer/CI/dox_warnings.sh` to list Sphinx resp. Doxygen issues


maintainer/CI/dox_warnings.py
```diff
@@ -0,0 +1,97 @@
...
+# filter out non-critical warnings
+warnings = {}
+for (filepath, lineno, warning), warning_list in raw_warnings.items():
+    if re.search('^member \S+ belongs to two different groups\. '
+                 'The second one found here will be ignored\.$', warning):
+        # happens when a function is declared in a group in the .hpp file but
+        # defined in another group in the .cpp file; this is usually caused by
+        # the "Private functions" and "Exported functions" groups in .hpp files
+        continue
+    if re.search('^documented symbol `\S+\' was not declared or defined\.$', warning):
```
KaiSzuttor, line 64: I'm wondering if all this is maintainable... @fweik what do you think?

doc/sphinx/magnetostatics.rst
```diff
@@ -171,11 +171,11 @@ calculations. The feature ``SCAFACOS_DIPOLES`` has to be added to
 :file:`myconfig.hpp` to activate this feature. At the time of this writing (Feb
```
pkreissl: This should probably go: "At the time of this writing (Feb 2018)"

jngrad:
Progress: from 136 Doxygen warnings to 0, from 66 Sphinx broken links to 0. The scripts worked as expected: no container failure, previous espresso-ci bot messages automatically removed.

Next step: merge python into this branch, check the espresso-ci bot posts the new warnings, fix the new warnings, then merge before anyone else can merge.

doc/sphinx/magnetostatics.rst
jngrad: done

jngrad:
The syntax checker implemented in 68f1a5b correctly identified the broken roles ([logfile](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/110107)). They were corrected in cde9732, and now the trace only shows `Found no broken link requiring fixing` ([logfile](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/110135)).

We won't make these syntax errors fail CI because they are difficult to spot in the HTML output, and it's not always obvious why they fail (e.g. the `:rtype:`appeared in the HTML output because a Sphinx external plugin incorrectly parsed `Returns` section headings in `cuda_init.pyx`).

fweik (approved):

PR #2799, opened 2019-05-06
observables: remove cylindrical LB observables (not save to use).

KaiSzuttor:
It turned out that the usage of this cylindrical observables is not save (sampling/performance issues).


fweik:
Can you please elaborate what the problem is?

KaiSzuttor:
The LB fluid is sampled on a regular grid although the histogram is polar.
Two issues arise from this:
- two have a sample for the smallest bin the sampling grid resolution has to be very high
- even if there is a hit for every bin, the sampling artifacts are quite large. You can see this in the sample script.

This is also not trivial to fix because it's (at least to me) not clear how to distribute the sampling points.
Alternatives are the ```LBVelocityProfile``` observable that can be used to create a cylindrical histogram afterwards..

fweik:
Can't the sampling points be picked from a regular cylindrical gird?

fweik:
Also plain radial profiles (without angular resolution) don't have this problem right? I think this is by far the most common case an is much easier. In general these obs have the advantage that they reduce the amount of data a lot, so if possible I'd like to keep some form of them.

KaiSzuttor:
you would want to have a constant sampling density in the bins. This is non-trivial to accomplish.

For a radial profile you would still need to sample different angles of course.

fweik:
@KaiSzuttor I assume this is obsolete now?

KaiSzuttor:
yes

PR #2800, opened 2019-05-06
Remove OpenMPI reference from LB swimmer documentation

mkuron:
This is not specific to OpenMPI, but applies to any MPI implementation


fweik:
I am not sure that this statement is even true.

pkreissl (approved):

fweik (approved):

PR #2803, opened 2019-05-06
core: particle_data: Made external forces/torques deactivable again

fweik:


fweik:
@KaiSzuttor  The clang-tidy warning is a false positive, this is why this was broken.

jngrad:
is it just a matter of removing this line?
https://github.com/espressomd/espresso/blob/1645121ee47d6a5cb12222f753e0adc36411f09a/.clang-tidy#L28

fweik:
Yes, but the original change was made automatically by `clang-tidy -fix`, which means all places where this was changed with `Utils::Vector` have to be checked. The replaced version is not equivalent...

fweik:
The offending PR was #2632. The rest looks fine. This is arguably a bug in the clang tidy rule. We could also mark this two place `NOLINT`, I think in general the rule is useful. @KaiSzuttor what do you think?

fweik:
For the record: This has been fixed in clang tidy, as far as I can tell...

jngrad:
it doesn't look too bad:
```bash
git diff -U1 454d62036f^ HEAD | grepdiff 'Vector3d{}' --output-matching=hunk | colordiff | less -R
```
then `/Vector3d\{\}` and `/\.empty` show only 3 occurences: the two you already fixed, and one in `src/core/unit_tests/field_coupling_fields_test.cpp` that is fixed with `Vector3d{0.0, 0.0, 0.0}`.

fweik:
No, it's fine; I checked.

fweik:
@RudolfWeeber

RudolfWeeber:
Please add a corresponding test to particle.py

fweik:
It's not observable.

RudolfWeeber (approved):
Right.

PR #2811, opened 2019-05-07
mpi callbacks with return values

fweik:
- added possibility to return values from mpi callbacks
  1. by reducing over all nodes
  2. for callbacks with `boost::optional` return value returning
      the value from one node
- modernization of some of the callback code
-  refactor of galilei transform stuff and tests thereof


RudolfWeeber (approved):

jngrad:
bors r=fweik

PR #2816, opened 2019-05-08
Fix segfault on empty nodes for mindist calculation

hirschsn:
Fixes a segfault for the mindist calculation if there is a node without particles.

Description of changes:
 - Check for empty partCfg and return inf in this case.

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


fweik:
Thank you! Turns out that this can be further simplified -> #2817.

PR #2817, opened 2019-05-08
Mindist

fweik:
Follow up on #2816.


jngrad (changes requested):
add the new python test to CMakeLists.txt

hirschsn:
Oops, my bad, sorry. :)

testsuite/python/CMakeLists.txt
```diff
@@ -196,6 +196,7 @@ python_test(FILE observable_chain.py MAX_NUM_PROC 4)
 python_test(FILE mpiio.py MAX_NUM_PROC 4)
 python_test(FILE gpu_availability.py MAX_NUM_PROC 1 LABELS gpu)
 python_test(FILE features.py MAX_NUM_PROC 1)
+python_test(FILE analyze_distance_empty.py MAX_NUM_PROC 256)
```
jngrad: `MAX_NUM_PROC` should be 1 (just like analyze_distance.py) because the function being tested is not likely to break when more CPU cores are used

testsuite/python/analyze_distance_empty.py
```diff
@@ -0,0 +1,28 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+import unittest as ut
+import espressomd
+
+
+class AnalyzeDistanceEmpty(ut.TestCase):
+    system = espressomd.System(box_l=[1.0, 1.0, 1.0])
+
+    def test_min_dist(self):
+        self.assertEqual(self.system.analysis.min_dist(), float("inf"))
```
pkreissl: `numpy.inf`

fweik:
@jngrad

jngrad (approved):

PR #2818, opened 2019-05-08
LB observable

KaiSzuttor:


src/core/observables/LBProfileObservable.hpp
```diff
@@ -19,12 +19,43 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+                                       x * sampling_delta_x);
+          m_sample_positions.push_back(sampling_offset_y +
+                                       y * sampling_delta_y);
+          m_sample_positions.push_back(sampling_offset_z +
+                                       z * sampling_delta_z);
+        }
+      }
+    }
+  }
+  std::vector<double> m_sample_positions;
```
fweik: Shouldn't this be a `Vector3d` or similar?

src/utils/include/utils/coordinate_transformation.hpp
```diff
@@ -47,9 +48,8 @@ inline Vector3d transform_pos_to_cylinder_coordinates(const Vector3d &pos,
 /** \brief Transform the given 3D velocity to cylinder coordinates with
  * longitudinal axis aligned with axis parameter.
  */
-inline Vector3d transform_vel_to_cylinder_coordinates(const Vector3d &vel,
-                                                      const std::string &axis,
-                                                      const Vector3d &pos) {
+inline Vector3d transform_velocity_cartesian_to_cylinder(
```
fweik: Why is there a special transform for velocities?

src/utils/include/utils/Vector.hpp
```diff
@@ -324,6 +324,14 @@ template <typename T, size_t N> struct decay_to_scalar<Vector<T, N>> {
 };
 
 template <typename T> struct decay_to_scalar<Vector<T, 1>> { using type = T; };
-} // namespace Utils
 
+template <size_t N, typename T>
+std::ostream &operator<<(std::ostream &out, Vector<T, N> const &a) {
```
fweik: This
- adds an extra space at the end
- is not tested
- is not templated on the outstream.

src/utils/include/utils/sampling.hpp
```diff
@@ -0,0 +1,93 @@
...
+    for (auto const phi : phi_range) {
+      sampling_positions.push_back(Vector3d{{*r_range.begin(), phi, z}});
+    }
+  }
+
+  // Scale the number of samples for larger bins
+  auto arc_length = [delta_phi, delta_r](int r_bin) {
+    return delta_phi * (r_bin + 1) * delta_r;
+  };
+  auto n_phi_samples = [arc_length](int r_bin) {
```
fweik, line 65: So `n_phi_samples(i) = i + 1` ? You can actually make this a function of r instead of the count, then you don't need the iterators later.

fweik (changes requested):

src/core/observables/LBProfileObservable.hpp
KaiSzuttor: Maybe, it's not part of the changes for this PR

src/utils/include/utils/coordinate_transformation.hpp
KaiSzuttor: Because you need to transform the velocities into cylinder coordinates

src/utils/include/utils/Vector.hpp
KaiSzuttor: I can remove it...

src/utils/include/utils/coordinate_transformation.hpp
fweik: The function should be called `transform_cartesian_to_cylinder` right?

KaiSzuttor: and how do I distinguish from the function for coordinates?

fweik: Is that a different transformation?

KaiSzuttor: yes

fweik:
Can you please also look at the change in the class hierarchy from #2830, and see that this is compatible?

src/core/observables/LBProfileObservable.hpp
fweik: Still, though...

KaiSzuttor:
should not be a problem...

src/core/observables/LBProfileObservable.hpp
KaiSzuttor: I did it, just for you... :+1:

src/utils/include/utils/sampling.hpp
KaiSzuttor: i find it more readable as it is now

KaiSzuttor:
@fweik could you please have a look at the failing accumulator test? It seems related to the script interface.

KaiSzuttor:
@fweik maybe it's easier for you to merge this

fweik:
Inheritance from `ProfileObservable` has little benefit, but does greatly increase the complexity. You should solve that differently.

fweik:
There wasn't anything wrong with your code: There was a latent bug in the python code that was triggered by your changes because you introduced a class that is not default constructible.

KaiSzuttor:
Why did the build of ```make_lin_range``` not fail in the already merged PR? Are the utils plus test actually build and run in the CI? It seems like the build issue on redhat only now pops up because I use the function... So ```make_lin_range``` is not compatible with boost 1.53.0 at least.

fweik:
Good question, but they are run somewhere, see https://github.com/espressomd/espresso/issues/2825

KaiSzuttor:
Ah no checks ror centos 7

mkuron:
@espresso-ci, have you become self-aware?

KaiSzuttor:
I just pressed the response button, an whoops I became espresso-ci

fweik:
That's actually how a lot of AI works ^^

KaiSzuttor:
```-G``` disables all optimization for device code

fweik:
@KaiSzuttor please fix the sphinx error, than this is good to go.

KaiSzuttor:
the sphinx error is not specific to this PR but is rather connected to an update of sphinx in the container

KaiSzuttor:
@fweik

KaiSzuttor:
i think you have to approve

fweik (approved):

PR #2821, opened 2019-05-09
Catch ihipException when no GPU is available

### LINKED ISSUE #2747: test_python_bindings calls CUDA function even if no GPU available
https://gitlab.icp.uni-stuttgart.de/espressomd/docker/-/jobs/101847 was built with AMD GPU support but the test was run on a machine without an appropriate GPU. That's fine because in the espressomd/docker repo we only run trivial tests. test_python_bindings however seems to call CUDA already when the `espressomd` module is imported (`mpi_init` -> `on_program_start` -> `cuda_init` -> `cudaStreamCreate`), which should not happen.

jngrad:
Fixes #2747

### DIFF
--- a/src/core/cuda_init_cuda.cu
+++ b/src/core/cuda_init_cuda.cu
@@ -39,7 +39,21 @@ static const int computeCapabilityMinMinor = 0;
 
 const char *cuda_error;
 
-void cuda_init() { cudaStreamCreate(&stream[0]); }
+void cuda_init() {
+#if defined(__HIPCC__) and not defined(__CUDACC__) and                         \
+    HIP_VERSION_PATCH <= 19171 /* i.e. <= v2.4.0 */
+  // Catch an exception that causes `import espressomd` to crash in
+  // Python when no compatible GPU is available (fixed in HIP v2.5.0)
+  try {
+    cudaStreamCreate(&stream[0]);
+  } catch (std::exception &) {
+    // The ihipException is not part of the public HIP library,
+    // so we catch its parent class std::exception instead
+  }
+#else
+  cudaStreamCreate(&stream[0]);
+#endif
+}
 
 /// get the number of CUDA devices.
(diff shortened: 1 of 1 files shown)


fweik (approved):

mkuron (commented):
What about `cudaGetDeviceCount`? Does that not throw exceptions?

jngrad:
`ihipException` is thrown in `ihipInit()` (`hip_hcc.cpp`), which is called only once via `std::call_once` in `hip_impl::hip_init()` (`hip_hcc.cpp`), which is called by most cuda functions via the macro `HIP_INIT_API` (`hip_hcc_internal.h`).

`cudaGetDeviceCount()` used to crash the python session because I called it before `cudaStreamCreate()`. Now that the exception is caught, all cuda functions can be called without throwing `ihipException`. For example, `cudaGetDeviceCount()` now just hangs indefinitely.

I used gdb to find the source of the hang. The python session cannot be stopped by Ctrl+C, only Ctrl+Z, which isn't caught by gdb so it stops both the python session and gdb. Instead, I ran the script in the background and attached its pid in gdb. The trace looks like a deadlock:
```
(gdb) attach 4885
Attaching to program: /usr/bin/python2, process 4885
[New LWP 4888]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
0x00007f7662678a4f in futex_wait (private=0, expected=1, futex_word=0x7f765f1f7a78 <hip_impl::hip_init()::hip_initialized>)
    at ../sysdeps/unix/sysv/linux/futex-internal.h:61
61	../sysdeps/unix/sysv/linux/futex-internal.h: No such file or directory.
```
I used the docker cmd: `run --user root --volume /scratch/jgrad:/local:rw --cap-add=SYS_PTRACE -it hip:gdb bash` to avoid the gdb error `ptrace: Operation not permitted.`.

PR #2826, opened 2019-05-10
Increase make_lin_space test tolerance

### LINKED ISSUE #2825: CI build failed for merged PR
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/pipelines/7217

jngrad:
Fixes #2825

Set tolerance to 1e-13% (aka 1e-15), like in `src/utils/tests/interpolation_test.cpp`.

### DIFF
--- a/src/utils/tests/make_lin_space_test.cpp
+++ b/src/utils/tests/make_lin_space_test.cpp
@@ -44,7 +44,7 @@ BOOST_AUTO_TEST_CASE(make_lin_space_test) {
 
     auto const dx = (stop - start) / (num - 1);
     for (int i = 0; i < values.size(); i++) {
-      BOOST_CHECK_CLOSE(values.at(i), start + i * dx, 1e-14);
+      BOOST_CHECK_CLOSE(values.at(i), start + i * dx, 1e-13);
     }
   }
 
@@ -64,7 +64,7 @@ BOOST_AUTO_TEST_CASE(make_lin_space_test) {
 
     auto const dx = (stop - start) / num;
     for (int i = 0; i < values.size(); i++) {
-      BOOST_CHECK_CLOSE(values.at(i), start + i * dx, 1e-14);
+      BOOST_CHECK_CLOSE(values.at(i), start + i * dx, 1e-13);
     }
   }
 }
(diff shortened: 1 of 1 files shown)


jngrad:
ran without failure 2000 times in `ubuntu:ppc64le`

fweik:
Why would you expect this to be non-deterministic?! Also I think this should use `std::numeric_limits<double>::epsilon()` to be more portable, no?

fweik:
Btw, is there documentation somewhere on how to run these containers?

jngrad:
I did not look at the source code for the function(s) that was(were) tested.
There is no documentation for the containers. Here's my cmd:
`docker run --user espresso --volume /scratch/jgrad/espresso-fork:/local:rw -it gitlab.icp.uni-stuttgart.de:4567/espressomd/docker/ubuntu:ppc64le bash`

fweik:
Ah, the emulation is completely transparent, that is nice!

fweik:
Why did you change the macro?

jngrad:
`BOOST_CHECK_CLOSE` uses [percentage tolerance](https://www.boost.org/doc/libs/1_70_0/libs/test/doc/html/boost_test/utf_reference/testing_tool_ref/assertion_boost_level_close.html), don't we need absolute tolerance when using epsilon?

fweik (approved):

jngrad:
btw are we there checking precision at 0.001 ppq (parts-per-quadrillion, 1e-15)?
https://github.com/espressomd/espresso/blob/75c39277a88dba622cde083efa11fc35c934ec3d/src/utils/tests/interpolation_test.cpp#L143

fweik:
Yes, because the result should be correct to machine precision.

jngrad:
but epsilon is divided by 100

fweik:
Yes, it was not clear to me that it is percentage.

fweik:
Actually also checking for epsilon is not optimal, but that usually does the trick (Maybe `0. or std::nextafter(0., 1.) or std::nextafter(0., -1.)`...)

jngrad:
Something like this? `BOOST_CHECK_SMALL(std::fabs(a - b), std::nextafter(0., 1.));`
Replacing epsilon everywhere would require a dedicated PR, let us merge this one first.

fweik:
I don't think this is worth the effort. ~Also what you wrote would fail if the result is bitwise identical?~

jngrad:
bors r=fweik

PR #2830, opened 2019-05-11
Time series + Accumulator improvements

fweik:
So I tried to use the observables...

- Simple accumulator that just records the values of an observable
- Only drop out of the intergration loop for updates if needed
- Only pull particle config if needed
- Removed one unneeded diamond pattern from the Observables class hierarchy

the last point is a considerable performance improvement if the partCfg isn't actually needed...
Overall the code is still pretty bad, but I can see some progress now...

### DIFF
--- a/src/core/observables/CylindricalProfileObservable.hpp
+++ b/src/core/observables/CylindricalProfileObservable.hpp
@@ -19,14 +19,12 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #ifndef OBSERVABLES_CYLINDRICALPROFILEOBSERVABLE_HPP
 #define OBSERVABLES_CYLINDRICALPROFILEOBSERVABLE_HPP
 
-#include <cmath>
-
-#include "Observable.hpp"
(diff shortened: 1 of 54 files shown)


KaiSzuttor (approved):

src/core/observables/CylindricalProfileObservable.hpp
```diff
@@ -19,14 +19,12 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-
-#include "Observable.hpp"
 #include <utils/Vector.hpp>
 
-namespace Observables {
+#include <string>
 
-class CylindricalProfileObservable : virtual public Observable {
+namespace Observables {
+class CylindricalProfileObservable {
```
KaiSzuttor, line 27: I've overlooked this in the review... why is this not an observable anymore?

fweik: Why would it? Nothing in this class is related to the observable interface

PR #2838, opened 2019-05-15
Added missing include for ```std::abs```

KaiSzuttor:


fweik (approved):

KaiSzuttor:
I accidentally created a branch in the espresso repository... we have to delete that branch after merge

PR #2839, opened 2019-05-15
maintainer: enable checks for centos.

KaiSzuttor:
since this is the linux with the oldest boost version, we should run the checks in this container.


KaiSzuttor:
a number of checks fail due to an incompatible numpy version in the centos container

PR #2840, opened 2019-05-15
Lin space

### LINKED ISSUE #2837: Utils function not compatible with required boost version
```Utils::make_lin_space``` does not build with boost version 1.53 which is currently the version we require. By inspection of the boost repo, I suppose we need at least 1.57 (from this version on at least the namespaces exist).

### PR DESCRIPTION
**Pull request by fweik**
Maybe fixes #2837.

### CODE REVIEW

**Comment by KaiSzuttor** [comment]
as I said, I already tried that...

**Comment by fweik** [comment]
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/114263
boost issue seems gone, but numpy is too old too

**Review by mkuron — DISMISSED** [review]

**Comment by mkuron** [comment]
The Numpy thing can easily be fixed by replacing `numpy.linalg.norm(A, axis=1)` with ` numpy.sqrt(numpy.einsum('ij,ij->i',A,A))`. That already works in Numpy 1.6.0.

**Comment by KaiSzuttor** [comment]
this is hard to read, I'd prefer to have a utility function that wraps this logic for older numpy versions

**Comment by fweik** [comment]
```patch
diff --git a/src/utils/include/utils/math/make_lin_space.hpp b/src/utils/include/utils/math/make_lin_space.hpp
index fa2198d06..6145048a3 100644
--- a/src/utils/include/utils/math/make_lin_space.hpp
+++ b/src/utils/include/utils/math/make_lin_space.hpp
@@ -45,8 +45,21 @@ auto make_lin_space(T start, T stop, size_t number, bool endpoint = true) {
   using boost::make_iterator_range;
   using boost::make_transform_iterator;
 
+  struct X {
+       using result_type = T;
+
+       T dx;
+       T start;
+
+       T operator()(size_t i) const {
+          return start + i * dx;
+       }
+  };
+
   auto const dx = (stop - start) / (number - endpoint);
-  auto x = [dx, start](size_t i) { return start + i * dx; };
+//  auto x = [dx, start](size_t i) { return start + i * dx; };
+
+      auto const x = X{dx, start};
 
   return make_iterator_range(
       make_transform_iterator(make_counting_iterator(size_t(0)), x),
```

src/utils/include/utils/raster.hpp
```diff
@@ -22,19 +23,19 @@ namespace Utils {
 template <class T, class F>
```
KaiSzuttor, line 23: where is ```T``` actually used?

KaiSzuttor (approved):

PR #2841, opened 2019-05-15
Tag CI jobs that cannot run in the cloud

mkuron:
I'm going to try to set up some runners on https://www.bw-cloud.org/ soon.


KaiSzuttor:
Could you please add documentation in the wiki for all of the CI changes you will perform

mkuron:
Will do that once it's running. I can't (or shouldn't) test it before this PR is merged.

jngrad (approved):

PR #2849, opened 2019-05-20
maintainer: add a job with the minimum required boost version

### LINKED ISSUE #2837: Utils function not compatible with required boost version

KaiSzuttor:
Fixes #2837

### DIFF
--- a/src/core/accumulators/Correlator.cpp
+++ b/src/core/accumulators/Correlator.cpp
@@ -298,7 +298,7 @@ void Correlator::initialize() {
     m_correlation_args[0] = m_correlation_args[0] * m_correlation_args[0];
     m_correlation_args[1] = m_correlation_args[1] * m_correlation_args[1];
     m_correlation_args[2] = m_correlation_args[2] * m_correlation_args[2];
-    fprintf(stderr, "args2: %f %f %f\n", m_correlation_args[0],
+    fprintf(stderr, "args^2: %f %f %f\n", m_correlation_args[0],
             m_correlation_args[1], m_correlation_args[2]);
     if (dim_A % 3)
       throw std::runtime_error(init_errors[18]);
--- a/src/core/accumulators/Correlator.cpp
+++ b/src/core/accumulators/Correlator.cpp
@@ -298,8 +298,6 @@ void Correlator::initialize() {
     m_correlation_args[0] = m_correlation_args[0] * m_correlation_args[0];
     m_correlation_args[1] = m_correlation_args[1] * m_correlation_args[1];
     m_correlation_args[2] = m_correlation_args[2] * m_correlation_args[2];
-    fprintf(stderr, "args^2: %f %f %f\n", m_correlation_args[0],
-            m_correlation_args[1], m_correlation_args[2]);
     if (dim_A % 3)
       throw std::runtime_error(init_errors[18]);
     m_dim_corr = dim_A / 3;
--- a/src/core/io/writer/h5md_core.hpp
+++ b/src/core/io/writer/h5md_core.hpp
@@ -27,15 +27,15 @@
 #include "cells.hpp"
 #include "global.hpp"
 
+#include "communication.hpp"
 #include <algorithm>
+#include <boost/filesystem.hpp>
 #include <fstream>
+#include <h5xx/h5xx.hpp>
 #include <iostream>
 #include <mpi.h>
 #include <string>
 #include <unordered_map>
-#include <boost/filesystem.hpp>
-#include "communication.hpp"
-#include <h5xx/h5xx.hpp>
 
 namespace Writer {
 namespace H5md {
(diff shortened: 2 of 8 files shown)


KaiSzuttor:
this needs to be merged before we can test the new container in the docker PR. After that we can activate the job in the CI.

KaiSzuttor:
@fweik any idea, what is happening there?
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/116554

mkuron:
> https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/116554
> mpiexec noticed that process rank 1 with PID 0 on node coyote exited on signal 4 (Illegal instruction).

That also happened last night in the i386 container: https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/116520 (#2851). Sounds like the compiler is generating instructions not supported by the processor?

KaiSzuttor:
@jngrad I don't understand the failing samples test. If I locally run the sample ```observables_correlators.py``` I get the exact same error for the current python branch. However, the CI did only detect the issue in this PR. Could you please have a look into that?

jngrad:
@KaiSzuttor found it: `Accumulators::Correlator::set_correlation_args()` in the core is never called by the interface constructor `ScriptInterface::Accumulators::Correlator::Correlator()`, so the value `[10, 10, 10]` passed to argument `args` are not copied to `m_correlation_args`, which is left uninitialized, usually with values `[0., 0., 0.]`, so no throw (and returns wrong numerical results).
https://github.com/espressomd/espresso/blob/0520af905d8914c284b672e9c0493414a063c66b/src/script_interface/accumulators/Correlator.hpp#L43-L53

KaiSzuttor:
But why is this not an issue in the current python branch?

src/core/io/writer/h5md_core.hpp
```diff
@@ -27,17 +27,15 @@
 #include "cells.hpp"
 #include "global.hpp"
 
+#include "communication.hpp"
```
fweik: Please clean up the includes. I think the core headers are only needed in the implementation..

fweik (changes requested):

jngrad:
The scenario `corr_operation="fcs_acf"` is untested, so only jobs with sample checks enabled will run it via the sample `observables_correlators.py`. When printing the member `correlation_args`, I get either `0.000000 0.000000 0.000000` or `0.000000 0.000000 -7732675196468477331125564344973318597912019529669651378575435354452340506624.000000` followed by our RuntimeError.  `T m_data[N];` (the internal data structure for `Vector3d correlation_args`) should be implicitly zero-initialized, so there must be an issue with pointers. From my side the RuntimeError frequency is rather low and very sensitive to how many print/printf functions I add in the core class or in the sample script.

fweik (dismissed):

fweik:
The correlator code is unreadable and incomprehensible, I don't think it is worth while investigating any issues in it.

fweik:
@jngrad you wrote
> Accumulators::Correlator::set_correlation_args() in the core is never called by the interface constructor ScriptInterface::Accumulators::Correlator::Correlator()

The constructor of `ScriptInterface::Accumulators::Correlator::Correlator()` is not where the contained object is initialized, but by the call of `make_shared` in `ScriptInterface::Accumulators::Correlator::construct`...

jngrad:
Yes I realized it after posting my message. There is no `args` argument in the core constructor so the call to the core initializer method checks an uninitialized core member. Providing a default argument still doesn't fix the RuntimeError though.

jngrad:
I may have found a solution, let me check this on a clean source.

src/core/accumulators/Correlator.cpp
```diff
@@ -298,7 +298,7 @@ void Correlator::initialize() {
     m_correlation_args[0] = m_correlation_args[0] * m_correlation_args[0];
     m_correlation_args[1] = m_correlation_args[1] * m_correlation_args[1];
     m_correlation_args[2] = m_correlation_args[2] * m_correlation_args[2];
-    fprintf(stderr, "args2: %f %f %f\n", m_correlation_args[0],
+    fprintf(stderr, "args^2: %f %f %f\n", m_correlation_args[0],
```
fweik: This should probably be removed.

fweik (changes requested):

src/core/accumulators/Correlator.cpp
jngrad: Removed.

fweik (approved):

jngrad:
`coulomb_cloud_wall` is failing repeatedly

KaiSzuttor:
seems to be some UB?

fweik:
No, it's still `Illegal instruction`, so most likely some 64-bit code got in there

KaiSzuttor:
the printed error is not necessarily the actual if you google around...

KaiSzuttor:
https://stackoverflow.com/questions/10354147/find-which-assembly-instruction-caused-an-illegal-instruction-error-without-debu

jngrad:
bors r=fweik

PR #2853, opened 2019-05-23
Different architectures: switch Python2 to Python3

jngrad:
Next PR for #2694

Containers arm64, armhf, ppc64le, i386, s390x were updated today to Ubuntu 19.04 disco with Python3 (espressomd/docker#86).

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -171,9 +171,9 @@ fedora:
     - docker
     - linux
 
-#ubuntu:1404 not needed: used in cuda:8.0
 #ubuntu:1604 not needed: used in cuda:9.0
 #ubuntu:1804 not needed: default used in non-CUDA builds
+#ubuntu:1904 not needed: base image for builds on different architectures
 
 ### Builds with CUDA
 
@@ -304,7 +304,7 @@ rocm-maxset:
   script:
     - export with_cuda=false test_timeout=900 check_skip_long=true
     - export OMPI_MCA_btl_vader_single_copy_mechanism=none
-    - export myconfig=maxset
+    - export myconfig=maxset python_version=3
     - bash maintainer/CI/build_cmake.sh
   tags:
     - docker


jngrad:
same issue as #2849: `coulomb_cloud_wall` is failing repeatedly on `i386/coyote6` and `maxset/coyote3`

jngrad:
Other PRs won't be able to pass CI until this PR is merged since `i386` doesn't support Python2 anymore. The `Illegal instruction` bug can be reproduced in the Debug terminal of a running  `i386` job, with and without MPI. The `coulomb_cloud_wall` test fails at this line precisely:
https://github.com/espressomd/espresso/blob/0520af905d8914c284b672e9c0493414a063c66b/testsuite/python/coulomb_cloud_wall.py#L171-L177
The `espressomd.electrostatics.Scafacos` instantiation is fine, it's the adding to the actors list that fails.
@RudolfWeeber @fweik @KaiSzuttor

jngrad:
It was tedious without `gdb`, but here is the reconstructed traceback, from the point of view of the Scafacos object after being added to the list of actors:
```
self._activate()
self._activate_method()
self._set_params_in_es_core()
set_parameters() // in the core
set_params_safe()
tune()
scafacos->tune()
fcs_tune() // error!
```
Line 151:
https://github.com/espressomd/espresso/blob/fa4123d396158132197c8d268fa40d9b32be465d/src/scafacos/Scafacos.cpp#L148-L152
The call to `handle_error` is fine, it's the call to `fcs_tune` that fails. Arguments seem ok: `charges.size()` is `300`, `positions` is `4.96758 0.0381298 0.827927 ... 2 5.5`, `charges` is `-1 1 -1 ... 1 -1`.

KaiSzuttor:
We should definitely *not* install the newest scafacos but instead a statically released version. Otherwise we also test scafacos, which is obviously not what we want.

RudolfWeeber:
The dipoles branch being used did not receive an update, though.

So, there seems to be a different issue.

fweik:
This seems to be an upstream issue, so I think the proper solution is to remove scafacos from the i386 container.

jngrad:
We have this illegal instruction on all `ubuntu-python3:18.04`-based containers, not just `i386`.

fweik:
Which instruction is the offending one?

jngrad:
`shrx`

fweik:
pfft supports the `--enable-portable-binary` flag, which switches the compiler flag from `-march=native` to `-mtune=native`, which makes it portable. With this the generate binary seems portable (
` objdump -d .libs/libpfft.so.0.0.0 | grep shrx | wc -l` returns 0).

KaiSzuttor:
sounds good, could you pls open a PR in the docker repo?

fweik:
I think it is as easy as adding the flag to the top level configure, but I'm still waiting for the compile to finish...

jngrad:
`objdump -d /usr/local/lib/libpfft.so.0.0.0 | grep shrx | wc -l` also returns 0 in an image failing due to `shrx`, this instruction is actually found in the fftw3_mpi library, try instead `objdump -d /usr/local/lib/libfcs_fftw3_mpi.so | grep shrx | wc -l`

fweik:
fftw supports the same flag

jngrad:
This worked, all failing `coulomb_cloud_wall` jobs that got restarted in the last 2 hours [now pass CI](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs), except for i386 which requires the updated `.gitlab-ci.yml` in this PR. Can someone approve it? I'll then group merge it with the boost PR and the docs PR.

KaiSzuttor (approved):

jngrad:
bors r=KaiSzuttor

PR #2877, opened 2019-05-31
Ek params cleanup

fweik:


fweik:
@KaiSzuttor

src/core/grid_based_algorithms/lbgpu_cuda.cu
```diff
@@ -3209,9 +3202,6 @@ void lb_coupling_set_rng_state_gpu(uint64_t counter) {
 
 void lb_fluid_set_rng_state_gpu(uint64_t counter) {
   rng_counter_fluid_gpu = Utils::Counter<uint64_t>(counter);
-#ifdef ELECTROKINETICS
-  ek_set_rng_state(counter);
```
KaiSzuttor, line 3213: is this happening somewhere else now?

fweik: No, but I think there is also no way to set the seed for the EK. The way it was it's broken, because it will use exactly the same random numbers for LB and EK

KaiSzuttor (approved):

PR #2885, opened 2019-06-06
utils: Fixed DEVICE macros for HIP

fweik:


KaiSzuttor:
actually I already tried that but there are a lot of functions that do not seem to work with hip

KaiSzuttor:
e.g. ```throw``` is not allowed

fweik:
It's still more correct? But in general it looks like the HIP stuff is does not scale. I think the only maintainable option is to do only HIP (which can run on all hardware), or stick to CUDA for now.

mkuron:
> I think the only maintainable option is to do only HIP (which can run on all hardware)

Actually, HIP's CUDA wrapper is just a big header file that is used with nvcc. So the problem would just be the other way round.

> or stick to CUDA for now

Since our next compute cluster likely won't have Nvidia cards due to their price, that's not a good idea. I guess we just have to `#ifdef` our way around HIP's peculiarities and file bugs at https://github.com/ROCm-Developer-Tools/HIP.

mkuron (approved):

KaiSzuttor:
@mkuron in the current state we are not able to maintain HIP with espresso. The hipcc compiler bug left us with a green CI although there were severe issues during compilation. Until we have a maintainable state we have to remove it from the development branch

mkuron:
That compiler bug is already fixed in the next version of HIP. If we don‘t use HIP and report bugs to them, it might never get these issues fixed.

If we remove it from CI now, it will be completely broken by the time we re-introduce it in a few years or months, requiring lots of work.

fweik:
It is already broken, so potential decay is not an argument.

mkuron:
The tests on the master branch pass. That breakage only affects the GPU P3M, which is not currently tested on AMD because I didn't understand the code enough to make it work.

So potential decay is actually an argument here. The LB and all related stuff works perfectly, as proven by the passing tests. And I don't see why the present issue shouldn't be easily fixable. HIP and CUDA are similar, but not the same, so a few special-case `#ifdef`s are expected to be necessary.

PR #2886, opened 2019-06-06
Swimmer reactions

### LINKED ISSUE #2498: Multiple issues with swimmer reaction, decide faith
Currently swimmer reactions does not properly handle ghosts and pair discovery: It does it's own n2 pair finding but does not take into account the facts that pairs can be found multiple times, that some of the reacting particles could be ghosts and does not synchronize reactions between nodes. Also the implementation is rather confusing and the details are not documented. In the current state we can not keep it.

fweik:
Fixes #2498.

This was broken in multiple ways (see #2498) and seems abandoned.
Also It uses an interface (access to verlet lists and cells) that we do not
plan to provide in the future.

Description of changes:
 - Remove SWIMMER_REACTSION feature and functionality


mkuron:
@hmenke @jdegraaf

hmenke:
I think this is a fine addition to the ongoing effort of removing everything from ESPResSo that performs actual physics until only an over-engineered pile of interface code is left. :+1:

mkuron:
@hmenke, joking aside, #2498 raises some valid points that question the physics of the implementation and haven't been fixed.

KaiSzuttor:
also no physics is better than broken physics 😋

KaiSzuttor:
tutorial has to be removed

fweik:
I don't have a strong opinion about this. Is somebody want's to fix it we can leave it in and mark the feature as experimental. It is just not straight forward because the algorithm does directly use the cells.

RudolfWeeber:
The ticket reporting the issues with SWIMMEr_REACTION was opened four months ago.
So far no-one has reported that they use this feature in a scientific project and no-one has comitted to fixing it.

Unless that changes by Tuesday, I will aprove this PR removing the feature.

hmenke:
This conversation makes it very clear that my earlier comment was not a joke at all.

hmenke (approved):
Go ahead and merge.

RudolfWeeber (approved):

Issue #2892, opened 2019-06-10
MMM1d and MMM1DGPU  forces are not calculated in short-range loop


jonaslandsgesell:
This is due to a wrong value for the coulomb cutoff (?) which is zero in both cases. Corresponds to #1663 which probably allowed this to happen.

jonaslandsgesell:
This is another example which shows off wrong results (all forces are zero, the results are written to forces_mmm1d.txt):

```` 
from __future__ import print_function
import espressomd
from espressomd.interactions import *
from espressomd import electrostatics
#from espressomd.electrostatics import MMM1DGPU
from espressomd.electrostatics import MMM1D
import numpy as np
import sys

N0=40 
bjerrum= 2
cylinder_height=30.966
radius=16.2284848063

box_l=2*radius
box_l_z=cylinder_height

system = espressomd.System(box_l=[box_l, box_l, box_l_z])
system.cell_system.set_n_square()
system.time_step = 0.01
system.cell_system.skin = 0.4
system.periodicity=[0,0,1]
system.seed=np.random.randint(20000)
np.random.seed=42


#############################################################
#  Setup System                                             #
#############################################################

for i in range(N0):
    system.part.add(pos=np.random.random(3),q=-1)
    system.part.add(pos=np.random.random(3),q=1)

old_pos=np.loadtxt("pos.txt")
system.part[:].pos=old_pos

old_q=np.loadtxt("q.txt")
system.part[:].q=old_q

charge_tot=np.sum(system.part[:].q)
if(abs(charge_tot)>1e-3):
    raise ValueError("System is not neutral.")

if(bjerrum>0):
    mmm1d = MMM1D(prefactor=bjerrum, maxPWerror=1e-5)
#    mmm1d=MMM1DGPU(prefactor=bjerrum, maxPWerror=1e-5)
    system.actors.add(mmm1d)

print(mmm1d._get_params_from_es_core())
old_pos=np.loadtxt("pos.txt")
system.part[:].pos=old_pos

old_q=np.loadtxt("q.txt")
system.part[:].q=old_q

system.integrator.run(steps=0)
np.savetxt("forces_mmm1d.txt", system.part[:].f)
````

[pos.txt](https://github.com/espressomd/espresso/files/3275673/pos.txt)
[q.txt](https://github.com/espressomd/espresso/files/3275674/q.txt)

RudolfWeeber:
Can mmm1d be tested with sth for which the expected result can be computed in python directly 

[...]

PR #2893, opened 2019-06-11
Mmm1d test

### LINKED ISSUE #1663: MMM1D lacks integration test
cpu and gpu

jonaslandsgesell:
Fixes #1663 @mkuron @fweik 

The tests failed however because forces and energies were not calculated, related to #2892. Works now


src/core/electrostatics_magnetostatics/coulomb.cpp
```diff
@@ -118,6 +118,10 @@ double cutoff(const Utils::Vector3d &box_l) {
     return dh_params.r_cut;
   case COULOMB_RF:
     return rf_params.r_cut;
+  case COULOMB_MMM1D:
+    return mmm1d_params.far_switch_radius_2;
```
jonaslandsgesell: this maybe needs to be std::sqrt()...

fweik:
Please document where the reference data  is from

fweik:
The GPU method should not declare a cutoff, since it does not use short-range calculation, and from what I can tell from the implementation the CPU version should not be cut off. The switch between short- and far-formula happens within the method.

jonaslandsgesell:
The reference data is taken from the old tcl branch where it was stored in a blockfile. Unfortunately, I cannot tell how it was created.

testsuite/python/mmm1d.py
```diff
@@ -20,33 +20,98 @@
...
+    system.cell_system.skin = 0.4
+    system.cell_system.set_n_square()
+    system.thermostat.set_langevin(kT=0,gamma=1, seed=8)
 
+    pid_target, pos_x_target, pos_y_target, pos_z_target, q_target, f_x_target, f_y_target, f_z_target = np.loadtxt(
+        tests_common.abspath("data/mmm1d_data.txt"), unpack=True)
+    vec_f_target = np.stack((f_x_target, f_y_target, f_z_target), axis=-1)
+    energy_target = -7.156365298205383
+    num_particles = pid_target.shape[0]
+    pos_x_target=np.mod(pos_x_target,10)
```
jonaslandsgesell: @fweik Hey Floh. What changed in the implementation of setting particles that setting them into the main box is needed to obtain the correct result?
Will the MMM1D calculation break if the particle leaves the main box and "enters" an image box? It seems like the implementation is sensitive to the periodic image... (This means the test fails if you comment out all lines with np.mod() )...

jngrad (approved):
The tests look good to me. The two points raised by Jonas should be addressed/discussed.

fweik:
bors r=jngrad

testsuite/python/mmm1d.py
jonaslandsgesell: the particles were set outside the non-periodic box. this resulted in wrong results. If Espresso would throw an error like suggested in #2905 this would have been obvious

jonaslandsgesell:
@jngrad can you rerun the bors build?

jngrad:
bors retry

PR #2901, opened 2019-06-12
hip: partially fix build.

KaiSzuttor:


KaiSzuttor:
it seems that hip is not capable of handling aggregate initialization of the ```Array``` class

src/utils/include/utils/device_qualifier.hpp
```diff
@@ -1,7 +1,11 @@
 #ifndef UTILS_DEVICE_QUALIFIER_HPP
 #define UTILS_DEVICE_QUALIFIER_HPP
 
-#if defined(__CUDACC__)
+#if defined(__CUDACC__) || defined(__HIPCC__)
+#define GPUCOMPILER
```
fweik: I think it would be more consistent to do

```c++
#if defined(__CUDACC__)
#define DEVICE_COMPILER cudacc
#elif  defined(__HIPCC__)
#define DEVICE_COMPILER hipcc
#endif 

#if defined(DEVICE_COMPILER)
#define DEVICE_THROW(E) throw (E)
#else 
#define DEVICE_THROW(E)
#endif 
```

KaiSzuttor:
``` bash
/builds/espressomd/espresso/src/utils/include/utils/Array.hpp:31:45: error: no matching constructor for initialization of 'detail::Storage<float, 6UL>'
template <typename T, std::size_t N> struct Array {
                                            ^~~~~
/builds/espressomd/espresso/src/utils/include/utils/Array.hpp:16:45: note: candidate constructor (the implicit copy constructor) not viable: no known conversion from 'float [6]' to 'const Utils::detail::Storage<float, 6>' for 1st argument
template <typename T, std::size_t N> struct Storage {
                                            ^
/builds/espressomd/espresso/src/utils/include/utils/Array.hpp:16:45: note: candidate constructor (the implicit move constructor) not viable: no known conversion from 'float [6]' to 'Utils::detail::Storage<float, 6>' for 1st argument
template <typename T, std::size_t N> struct Storage {
                                            ^
/builds/espressomd/espresso/src/utils/include/utils/Array.hpp:16:45: note: candidate constructor (the implicit copy constructor) not viable: no known conversion from 'float [6]' to 'float [6]' for 1st argument
template <typename T, std::size_t N> struct Storage {
                                            ^
/builds/espressomd/espresso/src/utils/include/utils/Array.hpp:16:45: note: candidate constructor (the implicit default constructor) not viable: requires 0 arguments, but 1 was provided
```

any idea how to fix that? @mkuron @fweik

fweik:
@KaiSzuttor maybe use a different compiler? ^^

fweik:
Ok, so now we are back to `hipErrorUnknown` ?

KaiSzuttor:
no, we are still at the error above

fweik:
So, what do we do now?

KaiSzuttor:
I'm out of ideas, tbh. We have to wait for @mkuron

KaiSzuttor:
probably related:

https://github.com/RadeonOpenCompute/hcc/issues/860

fweik:
Overall I'd say that hcc/hip seems not production ready, maybe we should retry later.

KaiSzuttor:
it also seems that hcc is deprecated:

https://github.com/RadeonOpenCompute/hcc#deprecation-notice

KaiSzuttor:
However the limitation we hit:

https://github.com/ROCm-Developer-Tools/HIP/blob/91f82ce541b51ad1106a28e951b410a256e61f62/docs/markdown/hip_bugs.md#errors-related-to-no-matching-constructor

has been removed in ROCm-Developer-Tools/HIP@4f363df159740277d11ec8224bf17937dbb30c03

at least in the documentation...

KaiSzuttor:
I'm not sure what we do now... probably we disable CI for now and hope that our hipster is not hit by a truck in the meantime? 🚜

fweik:
I think we let @RudolfWeeber decide :-)

mkuron:
The current master builds fine with rocm, so it's not obvious to me why this pull request is needed.

We've worked around CUDA compiler limitations many times before, so we can probably figure out a workaround for this HIP compiler bug too. CUDA and HIP are both not fully C++-compliant, so it's possible that we've hit something here that they don't sufficiently support. I'll try to re-write the offending lines with a bit less C++.

> it also seems that hcc is deprecated

We don't use HCC directly. Both HIP and HC wrap around AMD's low-level API.

fweik:
Superseded by #2937.

PR #2902, opened 2019-06-12
Put global box length and periodicity into class, pass as argument

### LINKED ISSUE #2675: Output parameter version of `get_mi_vector` should be removed
to discourage bad code where the result is not `const`.

fweik:
Fixes #2675.

Description of changes:
 - Introduce type for global box geometry
- Pass box geometry as argument to some functions
- de-feature `PARTIAL_PERIODIC`


fweik:
@RudolfWeeber @jngrad any input on why the virtual sites and bond angle stuff is not working anymore is welcome. It's not obvious to me.

src/core/electrostatics_magnetostatics/mmm2d.cpp
```diff
@@ -295,6 +296,19 @@ void MMM2D_setup_constants() {
   }
 }
 
+static void layered_get_mi_vector(double res[3], double const a[3],
```
KaiSzuttor, line 299: why did you move the function to mm2d?

fweik: Because it's only used there

fweik:
Sorry forgot something.

jngrad:
@fweik I don't see what you mean by virtual sites and bonds, on my machine I can't even get the PR to compile: the Cython code is unable to cast `Vector3d` to `const Vector3d` nor to choose which overloaded `make_array_locked()` to import from `utils.pxd`.

fweik:
Some of those are bugs in the cython compiler that are fixed in newer versions. Works here with cython 0.28.2 (from pip).

fweik:
The doxygen warning seems to be a false positive to me, the cited line does not contain a `\ref` command.

jngrad:
Let me have a look

jngrad:
ah the prototype `static void dp3m_init_a_ai_cao_cut();` is declared twice at lines 98 and 131. The offending ref is in the first prototype. But both Doxygen comments don't match at all, one of them must be wrong.

fweik:
Ah ok, I somehow failed to see how the error message is related to that...

jngrad:
You still need to remove the `@ref` at line 96.

src/core/BoxGeometry.hpp
```diff
@@ -0,0 +1,48 @@
...
+   *
+   * @param coord The coordinate to set the periodicity for.
+   * @param val True if this direction should be periodic.
+   */
+  void set_periodic(unsigned coord, bool val) { m_periodic.set(coord, val); }
+
+  /**
+   * @brief Check periodicity in direction.
+   *
+   * @param coord Direction to check
```
pkreissl, line 27: Typo "if".

src/core/BoxGeometry.hpp
```diff
@@ -0,0 +1,48 @@
...
+   * @return true iff periodic in direction.
```
pkreissl, line 28: Comment refers to this line.

fweik: https://en.wikipedia.org/wiki/If_and_only_if :-)

pkreissl: Iff this was indeed intentional, leave as is ;-)

fweik:
@KaiSzuttor @RudolfWeeber

jngrad (approved):

RudolfWeeber (approved):

PR #2903, opened 2019-06-12
Clarification of virtual sites relative update rule

fweik:
I think this makes it clearer what happens (at least to me ^^). I'm also not
so sure about the displacement limit and communication pattern, but will
address that in a separate PR.


RudolfWeeber (approved):
bors r+

PR #2906, opened 2019-06-13
core: remove hip support.

KaiSzuttor:
@RudolfWeeber


src/core/cuda_wrapper.hpp
```diff
@@ -1,64 +1,11 @@
...
-
-#endif
-
-#if defined(__HIPCC__) and not defined(__CUDACC__) // AMD-only
-
-#define make_uint3 dim3
-
-#endif
+#define DYNAMIC_SHARED(type, var) extern __shared__ type var[];
+#define SYMBOL(X) X
```
fweik: Please stick to the `DEVICE_` naming scheme.

src/core/cuda_wrapper.hpp
```diff
@@ -1,64 +1,11 @@
...
-#define cudaStreamCreate hipStreamCreate
-#define cudaStreamDestroy hipStreamDestroy
-#define cudaStream_t hipStream_t
-#define cudaSuccess hipSuccess
-#define __all_sync __all
-
-#else // Nvidia via CUDA
-
-#define hipLaunchKernelGGL(kernel, blocks, threads, mem, stream, ...)          \
+#define LaunchKernel(kernel, blocks, threads, mem, stream, ...)          \
```
fweik: This should have an all uppper name to indicate that it is macro, maybe `DEVICE_LAUNCH`?

fweik (changes requested):

RudolfWeeber:
I have the impression that hip is not ready for use in ESpresso yet.
Looking at the issues and prs in the hip repository at github, things seem to be very much in flow.
I feel that with our limited resources we cannot commit to keeping hip support in Espresso working at all times. 

There are simply too many items to work on in Espresso propper for us to be dealing with bugs in tooling to that extent.

Support for AMD Gpus would be a very nice thing to have, but I think, for now we are stuck with Cuda.
We can continue to look at Hip in a features branch, such that we are not forced to keep it in sync with the development head at all times. Once the state of hip has converged a bit, we can and should retry.

A final decisoin will have to be made next week.

fweik:
We decided to defer this decision for 6 weeks.

PR #2909, opened 2019-06-13
LB: removed remove total momentum.

### LINKED ISSUE #2876: LB CPU: Momentum calculation and node velocity getter inconsistent
lb_calc_fluid_momentum() adds 1 node force density to the current velocity, whereas the node velocity getter adds 1/2 node force density.
I assume, the latter is correct

KaiSzuttor:
Fixes #2876


fweik:
Please add a test for the linear momentum.

fweik (changes requested):
Please add a test for the linear momentum.

KaiSzuttor:
@fweik

fweik (dismissed):

KaiSzuttor:
@fweik

fweik (approved):

RudolfWeeber:
Sorry, but I think this needs more discussion.
I agree that an lbfluid method should not change the particles, but I also think the ability to add momentumto the fluid as a whole should remain. For the particles, that canprobably be done in two lines of Python, but for the fluid, probably sth in the core is needed.

KaiSzuttor:
the functions that were present only allowed to add global momentum to the fluid. I don't see any use-case for that.

KaiSzuttor:
also, I only removed code that has been used in the ```remove_total_momentum``` function. Unless there is a use-case for the core functionality (which I don't see), we should remove unused code.

RudolfWeeber:
Offline discussion yielded that we add a generator which let's one iterate over lb fluid nodes from Python.
Then, adding a velocity offset to the fluid can be written easily in simulation scripts.

PR #2912, opened 2019-06-13
Clarify which dipolar methods are MPI-enabled

### LINKED ISSUE #2828: Exceptions in dipolar direct sum methods on cpu
Classes for dipolar direct sum methods on cpu aren't very verbose in their exceptions. For example in class `DipolarDirectSumCpu`:
https://github.com/espressomd/espresso/blob/75c39277a88dba622cde083efa11fc35c934ec3d/src/python/espressomd/magnetostatics.pyx#L270-L274
The exception doesn't say why class instantiation failed, even though `dawaanr_set_params()` could either print to stderr or return different error codes depending on the issue. Same observation for `DipolarDirectSumWithReplicaCpu`.

The user guide and core should also clarify which magnetostatics methods are MPI-enabled. For example, class `DipolarDirectSumCpu` cannot be instantiated when more than 1 MPI thread is detected, but the documentation doesn't reflect that limitation. Reported by Ricky on the mailing list.

jngrad:
Fixes #2828

The error messages are more verbose and the documentation now mentions which classes are not MPI-enabled. The bibliographic issue was already fixed in #2857. This PR branches from #2910.

### DIFF
--- a/doc/sphinx/magnetostatics.rst
+++ b/doc/sphinx/magnetostatics.rst
@@ -132,9 +132,9 @@ To use the methods, create an instance of either :class:`~espressomd.magnetostat
 
 For testing purposes, a variant of the dipolar direct sum is available which adds periodic copies to the system in periodic directions (:class:`~espressomd.magnetostatics.DipolarDirectSumWithReplicaCpu`).
 
-The dipolar direct sum methods do not support MPI parallelization. They
-are not intended to do simulations, but rather to check the results
-you get from more efficient methods like P3M.
+The dipolar direct sum methods are very slow, and the CPU-based implementations
+do not support MPI parallelization. They are not intended to do simulations, but
+rather to check the results you get from more efficient methods like P3M.
 
 
 
--- a/doc/sphinx/magnetostatics.rst
+++ b/doc/sphinx/magnetostatics.rst
@@ -132,9 +132,9 @@ To use the methods, create an instance of either :class:`espressomd.magnetostati
 
 For testing purposes, a variant of the dipolar direct sum is available which adds periodic copies to the system in periodic directions (:class:`espressomd.magnetostatics.DipolarDirectSumWithReplicaCpu`).
 
-As it is very slow, this method is not intended to do simulations, but
-rather to check the results you get from more efficient methods like
-P3M.
+The dipolar direct sum methods do not support MPI parallelization. They
+are not intended to do simulations, but rather to check the results
+you get from more efficient methods like P3M.
 
 
 
--- a/doc/sphinx/magnetostatics.rst
+++ b/doc/sphinx/magnetostatics.rst
@@ -130,12 +130,15 @@ To use the methods, create an instance of either :class:`~espressomd.magnetostat
   system.actors.add(dds)
 
 
-For testing purposes, a variant of the dipolar direct sum is available which adds periodic copies to the system in periodic directions (:class:`~espressomd.magnetostatics.DipolarDirectSumWithReplicaCpu`).
-
-The dipolar direct sum methods are very slow, and the CPU-based implementations
-do not support MPI parallelization. They are not intended to do simulations, but
+For testing purposes, a variant of the dipolar direct sum is available which
+adds periodic copies to the system in periodic directions:
+:class:`~espressomd.magnetostatics.DipolarDirectSumWithReplicaCpu`.
+As it is very slow, this method is not intended to do simulations, but
 rather to check the results you get from more efficient methods like P3M.
 
+:class:`~espressomd.magnetostatics.DipolarDirectSumCpu` and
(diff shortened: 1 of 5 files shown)


doc/sphinx/magnetostatics.rst
```diff
@@ -111,30 +111,30 @@ Due to the long-range nature of dipolar interactions, Direct summation with mini
...
   system.actors.add(dds)
 
 
-For testing purposes, a variant of the dipolar direct sum is available which adds periodic copies to the system in periodic directions (:class:`espressomd.magnetostatics.DipolarDirectSumWithReplicaCpu`).
+For testing purposes, a variant of the dipolar direct sum is available which adds periodic copies to the system in periodic directions (:class:`~espressomd.magnetostatics.DipolarDirectSumWithReplicaCpu`).
 
-As it is very slow, this method is not intended to do simulations, but
-rather to check the results you get from more efficient methods like
-P3M.
+The dipolar direct sum methods are very slow, and the CPU-based implementations
```
RudolfWeeber: The slowness comment was meant for the direct sum with replica spcifically. For open boundary systems, there is no choice but to use direct summation (except P2NFFT provided by the Scafacos library, which is more difficult to use).

src/core/electrostatics_magnetostatics/magnetic_non_p3m_methods.cpp
```diff
@@ -389,6 +389,7 @@ double magnetic_dipolar_direct_sum_calculations(int force_flag,
 
 int dawaanr_set_params() {
   if (n_nodes > 1) {
+    runtimeErrorMsg() << "MPI parallelization not supported by DAWAANR.";
```
RudolfWeeber: Please adjsut the method names in the error message to mirror the Python class names

src/core/electrostatics_magnetostatics/magnetic_non_p3m_methods.cpp
```diff
@@ -402,14 +403,15 @@ int dawaanr_set_params() {
 
 int mdds_set_params(int n_cut) {
   if (n_nodes > 1) {
+    runtimeErrorMsg() << "MPI parallelization not supported by MDDS.";
```
RudolfWeeber: Same here

doc/sphinx/magnetostatics.rst
jngrad: Thanks! It's fixed.

src/core/electrostatics_magnetostatics/magnetic_non_p3m_methods.cpp
jngrad: Why? This core function is unaware of the Python interface. When this message is sent, Python displays it at the end of the traceback as `Exception: Could not activate magnetostatics method DipolarDirectSumCpu: ERROR: MPI parallelization not supported by DAWAANR.` This is done by:
https://github.com/espressomd/espresso/blob/76be86ca7678afbf302805ed5dcaf57137e89569/src/python/espressomd/magnetostatics.pyx#L273-L275

jngrad: If your real concern is that the `DAWAANR` abbreviation is irrelevant to the user, we could just remove the `"by DAWAANR"` at the end of the error message.

jngrad: I think I get it: the core function prefix `dawaanr_` is a C-style namespace, right? And DAWAANR was renamed into DipolarDirectSum in the Python interface for convenience. If that's the case, removing `"by DAWAANR"` from the message would be sufficient in my opinion, and will avoid any confusion if we decide to rename magnetostatics Python classes in the future.

RudolfWeeber:
jngrad commented on this pull request.

  _____  

In src/core/electrostatics_magnetostatics/magnetic_non_p3m_methods.cpp <https://github.com/espressomd/espresso/pull/2912#discussion_r293870213> :

> @@ -389,6 +389,7 @@ double magnetic_dipolar_direct_sum_calculations(int force_flag,
 
 int dawaanr_set_params() {
   if (n_nodes > 1) {
+    runtimeErrorMsg() << "MPI parallelization not supported by DAWAANR.";

I think I get it: the core function prefix dawaanr_ is a C-style namespace, right? And DAWAANR was renamed into DipolarDirectSum in the Python interface for convenience. If that's the case, removing "by DAWAANR" from the message would be sufficient in my opinion, and will avoid any confusion if we decide to rename magnetostatics Python classes in the future.

 

Let’s don’t keep secret what exactly does not support parallelization. 

I consider the name of the Python classes final as they actually say what the methods do (as opposed to DAWAANR), so a good error message would be more important than totally strict core/interface separation.

RudolfWeeber (approved):

PR #2916, opened 2019-06-13
Installation instructions for Python 3

RudolfWeeber:
This adds preliminary instructions to build with Python3. 
Once Python 2 support is removed, the docs have to be touched again, because they can then be simplified.

@jngrad 

 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?

### DIFF
--- a/doc/sphinx/installation.rst
+++ b/doc/sphinx/installation.rst
@@ -188,7 +188,9 @@ If you have installed the requirements (see section :ref:`Requirements
 create a build directory and call ``cmake`` and ``make`` (optional steps
 which modify the build process are commented out).
 To chose the correct Python version, run
+
 .. code-block:: bash
+
     python -V
     python3 -V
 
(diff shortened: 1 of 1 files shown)


doc/sphinx/installation.rst
```diff
@@ -178,15 +186,22 @@ Quick installation
 If you have installed the requirements (see section :ref:`Requirements
 <requirements>` ) in standard locations, to compile, it is usually enough to
 create a build directory and call ``cmake`` and ``make`` (optional steps
-which modify the build process are commented out):
+which modify the build process are commented out).
+To chose the correct Python version, run
+.. code-block:: bash
```
jngrad, line 192:
```suggestion

.. code-block:: bash

```
I think there should always be a newline character both before and after a `code-block` directive.

jngrad (changes requested):

KaiSzuttor:
@jngrad

jngrad (approved):

PR #2917, opened 2019-06-14
Lb nodes generator

RudolfWeeber:
@KaiSzuttor
This also moves lbfluid(cpu).stress to the base class since it also applies to lbgpu.


testsuite/python/lb.py
```diff
@@ -111,10 +111,8 @@ def test_mass_momentum_thermostat(self):
         # Cache the lb nodes
         lb_nodes = []
         n_nodes = int(self.params['box_l'] / self.params['agrid'])
-        for i in range(n_nodes):
-            for j in range(n_nodes):
-                for k in range(n_nodes):
-                    lb_nodes.append(self.lbf[i, j, k])
+        for n in self.lbf.nodes():
+            lb_nodes.append(n)
```
KaiSzuttor: you can omit this step now, just use the generator directly

KaiSzuttor (changes requested):
please remove the unnecessary loop in the test, besides that LGTM

KaiSzuttor (approved):

PR #2930, opened 2019-06-19
utils: tensor plus test.

KaiSzuttor:


src/utils/include/utils/tensor.hpp
```diff
@@ -0,0 +1,99 @@
...
+#include <numeric>
+#include <vector>
+
+#include <boost/range/algorithm/transform.hpp>
+#include <boost/range/numeric.hpp>
+
+/**
+ * @brief Heap-allocated tensor class.
+ * @tparam T data type
+ * @tparam t_rank number of dimensions of the tensor
```
fweik: Does not exist

src/utils/include/utils/tensor.hpp
```diff
@@ -0,0 +1,99 @@
...
+  using const_reference = T const &;
+  using pointer = T *;
+  using const_pointer = const value_type *;
+  using iterator = typename std::vector<T>::iterator;
+  using const_iterator = typename std::vector<T>::const_iterator;
+
+  /**
+   * @brief Creates a tensor of given rank and sizes
+   * @param dimensions Size for each dimension
+   * @param rank Rank of the tensor
```
fweik: Does not exist

src/utils/include/utils/tensor.hpp
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  /**
+   * @brief Element access.
+   * @param indices index for each dimension
+   */
+  const_reference
+  operator()(std::initializer_list<std::size_t> const &indices) const {
+    auto const index = boost::inner_product(m_strides, indices, 0);
+    assert(index < m_data.size());
```
fweik: This should check per component

src/utils/include/utils/tensor.hpp
```diff
@@ -0,0 +1,99 @@
...
+  pointer data() noexcept { return m_data.data(); }
+
+  const_pointer data() const noexcept { return m_data.data(); }
+
+  auto rank() const noexcept { return m_rank; }
+
+  auto dimensions() const noexcept { return m_dimensions; }
+
+private:
+  std::size_t m_rank;
```
fweik: This is redudent, `size() == m_dimensions.size()`

src/utils/include/utils/tensor.hpp
```diff
@@ -0,0 +1,99 @@
...
+  std::vector<T> m_data;
+  std::vector<std::size_t> m_dimensions;
```
fweik: This can be reconstructed from the total size and the strides

src/utils/include/utils/tensor.hpp
```diff
@@ -0,0 +1,99 @@
...
+   */
+  explicit Tensor(std::initializer_list<std::size_t> const &dimensions)
+      : m_rank(dimensions.size()),
+        m_dimensions(dimensions.begin(), dimensions.end()) {
+    m_data =
+        std::vector<T>(std::accumulate(dimensions.begin(), dimensions.end(), 1,
+                                       std::multiplies<std::size_t>()));
+
+    boost::transform(dimensions, std::back_inserter(m_strides),
+                     [stride = 1](size_t dim) mutable {
```
fweik: should be `[stride = size_t(1)]`

fweik (changes requested):

src/utils/include/utils/tensor.hpp
KaiSzuttor: yes but I need the dimensions every time I want to check the index, therefore it's better to store the extends.

src/utils/include/utils/tensor.hpp
```diff
@@ -0,0 +1,124 @@
...
+    }
+  }
+
+  std::size_t
+  calc_linear_index(std::initializer_list<std::size_t> const &indices) {
+    return boost::inner_product(m_strides, indices, 0);
+  }
+
+  std::vector<T> m_data;
+  std::vector<std::size_t> m_extends;
```
fweik: That should be `m_extents` right?

KaiSzuttor: yep

KaiSzuttor:
@jngrad there seems to be an issue with doxygen

maintainer/CI/dox_warnings.sh
```diff
@@ -66,6 +66,14 @@ fi
 # find @param without description
 grep -Prn '^[\ \t]*(?:\*?[\ \t]+)?[@\\]t?param(?:\[[in, out]+\])?[\ \t]+[a-zA-Z0-9_\*]+[\ \t]*$' "${srcdir}/src" > doc/doxygen/empty-params.log
 
+# find @param of arguments not recognized in class methods
+rm -f doc/doxygen/class_methods.log
```
jngrad:
```suggestion
echo "" > doc/doxygen/class_methods.log
```

src/utils/include/utils/tensor.hpp
```diff
@@ -0,0 +1,128 @@
...
+#include <vector>
+
+#include <boost/range/algorithm/transform.hpp>
+#include <boost/range/numeric.hpp>
+
+namespace Utils {
+
+/**
+ * @brief Heap-allocated tensor class.
+ * @tparam T data type
```
fweik, line 17: Doesn't exist

src/utils/include/utils/tensor.hpp
```diff
@@ -0,0 +1,128 @@
...
+ */
+template <typename T> class Tensor {
+public:
+  using value_type = T;
+  using reference = T &;
+  using const_reference = T const &;
+  using pointer = T *;
+  using const_pointer = const value_type *;
+  using iterator = typename std::vector<T>::iterator;
```
fweik: I think you should just fix that the storage is contiguous (and thus `Tensor` is a [ContiguousContainer]( https://en.cppreference.com/w/cpp/named_req/ContiguousContainer)), and just use pointers here, instead of leaking how you manage that memory.

src/utils/include/utils/tensor.hpp
```diff
@@ -0,0 +1,128 @@
...
+  reference operator()(std::initializer_list<std::size_t> const &indices) {
+    validate_indices(indices);
+    return m_data[calc_linear_index(indices)];
+  }
+
+  reference front() { return *begin(); }
+
+  const_reference front() const { return *cbegin(); }
+
+  reference back() { return *(end() - 1); }
```
fweik: Why don't you just return `back()` from your storage implementation?

src/utils/include/utils/tensor.hpp
```diff
@@ -0,0 +1,128 @@
...
+
+  const_reference back() const { return *(cend() - 1); }
```
fweik: See above

src/utils/include/utils/tensor.hpp
```diff
@@ -0,0 +1,128 @@
...
+private:
+  void
+  validate_indices(std::initializer_list<std::size_t> const &indices) const {
+    if (indices.size() != m_extents.size()) {
+      throw std::runtime_error("Number of indices have to match rank");
+    }
+    // check if any index exceeds an extend.
+    auto const res =
+        std::mismatch(indices.begin(), indices.end(), m_extents.begin(),
+                      [](std::size_t a, std::size_t b) { return a < b; });
```
fweik: This looks suspiciously like `std::less`

src/utils/include/utils/tensor.hpp
```diff
@@ -0,0 +1,128 @@
...
+
+private:
+  void
+  validate_indices(std::initializer_list<std::size_t> const &indices) const {
+    if (indices.size() != m_extents.size()) {
+      throw std::runtime_error("Number of indices have to match rank");
+    }
+    // check if any index exceeds an extend.
+    auto const res =
+        std::mismatch(indices.begin(), indices.end(), m_extents.begin(),
```
fweik: There is a range version in boost.

src/utils/include/utils/tensor.hpp
```diff
@@ -0,0 +1,128 @@
...
+  const_reference operator()(size_ts... indices) const {
+    return operator()({static_cast<std::size_t>(indices)...});
+  }
+
+  template <typename... size_ts> reference operator()(size_ts... indices) {
+    return operator()({static_cast<std::size_t>(indices)...});
+  }
+
+  const_reference
+  operator()(std::initializer_list<std::size_t> const &indices) const {
```
fweik, line 80: I think the interface can be made clearer here if `operator[]` is used for single argument access instead. This is also hat the proposed `mdspan` and `boost::multi_array` does. E.g. use a `const_reference operator[](initializer_list<...>)` or similar. Also I'm not sure that explicitly using initializer_list here as the only type with an explicit overload leads to awkward code if you already have an array: you either have to construct an `initializer_list`, which is just noise, or explicitly destructure it, which is complicated if you don't know the rank statically. You could either template the type, or at least use an explicit version with `Utils::Span<const size_t>` which will implicitly construct from all containers that have `.data()` and `.size()`. In general I think it is also not common client code that the indices are literals, so that the curly braces syntax isn't really used that often (only looking at unit tests may be a bit misleading here).

src/utils/include/utils/tensor.hpp
```diff
@@ -0,0 +1,128 @@
...
+
+  const_iterator cend() const noexcept { return m_data.end(); };
+
+  pointer data() noexcept { return m_data.data(); }
+
+  const_pointer data() const noexcept { return m_data.data(); }
+
+  auto rank() const noexcept { return m_extents.size(); }
+
+  auto extents() const noexcept { return m_extents; }
```
fweik: I *think* this should be `auto const&`

fweik (changes requested):

src/utils/include/utils/tensor.hpp
```diff
@@ -0,0 +1,128 @@
...
+    return operator()({static_cast<std::size_t>(indices)...});
+  }
+
+  template <typename... size_ts> reference operator()(size_ts... indices) {
+    return operator()({static_cast<std::size_t>(indices)...});
+  }
+
+  const_reference
+  operator()(std::initializer_list<std::size_t> const &indices) const {
+    validate_indices(indices);
```
fweik: I'm not sure that the default access should be range checked, this is not e.g. like `std::array` or `std::vector` behave. Here there is a special penalty to it because you need to load additional unrelated heap memory that you would not need otherwise. I think this should be an `assert`, and you should provide an `at` member that does a range checked version so that the caller can decide.

fweik (changes requested):

src/utils/include/utils/tensor.hpp
KaiSzuttor: what do you mean?

src/utils/include/utils/tensor.hpp
KaiSzuttor: > To provide multidimensional array access semantics, e.g. to implement a 3D array access a[i][j][k] = x;, operator[] has to return a reference to a 2D plane, which has to have its own operator[] which returns a reference to a 1D row, which has to have operator[] which returns a reference to the element. To avoid this complexity, some libraries opt for overloading operator() instead, so that 3D access expressions have the Fortran-like syntax a(i, j, k) = x;

from https://en.cppreference.com/w/cpp/language/operators

KaiSzuttor:
@fweik

src/utils/include/utils/tensor.hpp
```diff
@@ -0,0 +1,194 @@
...
+
+  iterator begin() noexcept { return &(*m_data.begin()); };
+
+  const_iterator begin() const noexcept { return &(*m_data.begin()); };
+
+  const_iterator cbegin() const noexcept { return &(*m_data.begin()); };
+
+  iterator end() noexcept { return &(*m_data.end()); };
+
+  const_iterator end() const noexcept { return &(*m_data.end()); };
```
fweik: You are not allowed to dereference the end iterator, it needs not to be valid, just say `data() + size()`, which is also less noisy.

src/utils/include/utils/tensor.hpp
```diff
@@ -0,0 +1,194 @@
...
+  /**
+   * @brief Throw if invalid indices.
+   * @param indices Indices to be checked.
+   * @tparam Type of container holding the indices.
+   * @throws std::runtime_error if indices are invalid.
+   */
+  template <typename Container>
+  void validate_indices(Container const &indices) const {
+    if (not valid_indices(indices)) {
+      throw std::runtime_error("Invalid indices");
```
fweik, line 179: `std::vector` throws `std::out_of_range`

fweik (changes requested):
I did not look at the doxygen stuff, because whatever...

src/utils/include/utils/tensor.hpp
KaiSzuttor: ok, but it's not only checking for out of range

KaiSzuttor:
actually I think you should have a look at the doxygen stuff, because you objected when I once said that this is hard to maintain...

src/utils/include/utils/tensor.hpp
fweik: Because it also checks the length? It's still what I would expect

fweik:
I don't think that there is currently another way to check the doxygen stuff. This of course has to be maintained, but I feel that it already helped considerably to improve the doxs. It is also not really
technical debt because there is nothing depending on this: We can tomorrow decide to remove it without
affecting anything else. Since it does not break every other day (I actually  think this is the first time), I'd
say it's worth it for now.

fweik:
I think the discussion of the type design is done, this can be used as is, but should only be merged when needed. Closing for now.

PR #2932, opened 2019-06-21
Improve maintainability of build_cmake.sh

jngrad:
The number of pitfalls in a shell script can be reduced if Google's [Shell Style Guide](https://google.github.io/styleguide/shell.xml#Variable_expansion) is followed:
- no more cryptic `if [ ! -z "${var+x}" ]` to test if a variable is undeclared (also, the conditional was used incorrectly and caused the error message `line 94: curl: command not found` in Travis-CI without triggering the trap)
- use CMake-like boolean values and test explicitly if a variable is equal to  `true` or `false` (we had a mix of `true`, `yes` and empty strings)
- every variable is brace-quoted
- every string variable is quoted (no more `build_dir="build maxset"; cd ${build_dir}` causing `bash: cd: too many arguments`)

### DIFF
--- a/maintainer/CI/build_cmake.sh
+++ b/maintainer/CI/build_cmake.sh
@@ -40,36 +40,71 @@ function end {
     echo "=================================================="
 }
 
-# execute and output a command
-# handle environment variables
-[ -z "${insource}" ] && insource="false"
-[ -z "${srcdir}" ] && srcdir=`pwd`
-[ -z "${cmake_params}" ] && cmake_params=""
-[ -z "${with_fftw}" ] && with_fftw="true"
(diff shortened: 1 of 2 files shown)


fweik (approved):

fweik:
I think next time something is added to this script we should consider transforming it into python.

jngrad:
I'm not looking forward to rewriting this script as a series of Python `subprocess` commands...

fweik:
There is things like the [`sh`](https://amoffat.github.io/sh/) module that let's you use shell commands as if they were python functions. That is a nice thing to know in general. Maybe we'll just refrain to add further stuff to the ci script then ^^

Issue #2900, opened 2019-06-12
Remove remaining manual memory management


fweik:
E.g. replace `Utils::malloc` and `Utils::realloc` by `std::vector` or similar.

jonaslandsgesell:
I ll remove what is obvious

RudolfWeeber:
This is merged, right?

fweik:
No, not resolved.

jngrad:
coding day: identify places where malloc and realloc are still used

jngrad:
`grep -r --color -P '(m|re)alloc(?=\()' src/core`

Most matches are in `src/core/grid_based_algorithms`, which will be replaced by waLBerla. The only exception is `src/core/ParticleList.hpp`, where the container size grows linearly (a std::vector grows exponentially).

fweik:
This, and then there is `Utils::List`, which is used for the bond and exclusion lists. Those have the advantage of being smaller than most implementations of `std::vector` if they are empty (because most impls store 3 pointers, and `List` is a pointer and two ints). In principle both can be replaced by boosts vector, which is the same size and has a customizable growth strategy, but this requires a newer boost version than what we currently require. (It's not clear which one).

jngrad:
The LB GPU mallocs leak memory, the `EK_DEBUG` feature doesn't compile and the EK VTK files are broken.

KaiSzuttor:
> The LB GPU mallocs leak memory, the `EK_DEBUG` feature doesn't compile and the EK VTK files are broken.

i think we can live with the leak for now, given that this part of the code will get replaced soon

PR #2933, opened 2019-06-24
removed some more allocs

jonaslandsgesell:
Partially adresses #2900 again


src/python/espressomd/analyze.pyx
```diff
@@ -820,9 +816,8 @@ class Analysis(object):
 
         cdef double * rh = NULL
```
KaiSzuttor: can be removed

fweik:
@KaiSzuttor could you please look at this? I think in some places there should be some cleaning up around the vector replacements should be done.

fweik:
@jonaslandsgesell you don't have to update the PR by hand if there is no conflict that needs resolving. We'll do that automatically when merging.

src/core/grid_based_algorithms/lb.cpp
```diff
@@ -372,13 +373,14 @@ static void halo_push_communication(LB_Fluid &lbfluid,
...
 
   if (local_node_grid[0] > 1) {
-    MPI_Sendrecv(sbuf, count, MPI_DOUBLE, snode, REQ_HALO_SPREAD, rbuf, count,
-                 MPI_DOUBLE, rnode, REQ_HALO_SPREAD, comm_cart, &status);
+    MPI_Sendrecv(sbuf.data(), count, MPI_DOUBLE, snode, REQ_HALO_SPREAD,
+                 rbuf.data(), count, MPI_DOUBLE, rnode, REQ_HALO_SPREAD,
+                 comm_cart, &status);
   } else {
-    memmove(rbuf, sbuf, count * sizeof(double));
+    memmove(rbuf.data(), sbuf.data(), count * sizeof(double));
```
KaiSzuttor: ```std::copy```

KaiSzuttor (changes requested):
please remove the ```memmoves```, otherwise looks good

src/core/grid_based_algorithms/lb_interface.cpp
```diff
@@ -1121,7 +1121,7 @@ double lb_lbnode_get_density(const Utils::Vector3i &ind) {
                            ind[2] * lbpar_gpu.dim_x * lbpar_gpu.dim_y;
     static std::vector<LB_rho_v_pi_gpu> host_print_values(0);
 
-    if (host_print_values.size()==0)
+    if (host_print_values.size() == 0)
```
KaiSzuttor: ```.empty()```

KaiSzuttor:
https://whatthecommit.com 😄

fweik:
@jonaslandsgesell please stop updating the branch without a reason

jonaslandsgesell:
> @jonaslandsgesell please stop updating the branch without a reason

If you have a look at the diff you see that I commited some small changes which Kai requested:
https://github.com/espressomd/espresso/pull/2933/commits/33df87018df43579d99a1ae4a758ffb728becc7a

the word up seems to mean something different to you than to me ;)

KaiSzuttor:
i think @fweik rather means to in general avoid a lot of pushes that only contain minor changes because each push triggers the CI

fweik:
They boy who cried 'up'

KaiSzuttor:
it's all about push ups 💪

KaiSzuttor:
```/builds/espressomd/espresso/src/core/grid_based_algorithms/lb_interface.cpp:1201:9: error: the 'empty' method should be used to check for emptiness instead of 'size' [readability-container-size-empty,-warnings-as-errors]
    if (host_print_values.size() == 0)
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        host_print_values.empty()
```
please fix those

KaiSzuttor:
I just assisted in merging because it was me causing conflicts ;)

KaiSzuttor (approved):

src/core/grid_based_algorithms/lb_interface.cpp
```diff
@@ -1115,16 +1111,15 @@ const Utils::Vector6d lb_lbnode_get_stress_neq(const Utils::Vector3i &ind) {
   Utils::Vector6d stress{};
   if (lattice_switch == ActiveLB::GPU) {
 #ifdef CUDA
-    static LB_rho_v_pi_gpu *host_print_values = nullptr;
-    if (host_print_values == nullptr)
-      host_print_values =
-          (LB_rho_v_pi_gpu *)Utils::malloc(sizeof(LB_rho_v_pi_gpu));
+    static std::vector<LB_rho_v_pi_gpu> host_print_values(0);
```
fweik: Why isn't this just a stack variable?

src/core/grid_based_algorithms/lbgpu.cpp
```diff
@@ -102,7 +102,7 @@ LB_parameters_gpu lbpar_gpu = {
     0.0};
 
 /** this is the array that stores the hydrodynamic fields for the output */
-LB_rho_v_pi_gpu *host_values = nullptr;
+std::vector<LB_rho_v_pi_gpu> host_values(0);
```
fweik, line 105: Here too, why is this a `vector`?

src/core/statistics.cpp
```diff
@@ -51,7 +51,7 @@
 
 /** Previous particle configurations (needed for offline analysis and
     correlation analysis) */
-double **configs = nullptr;
+std::vector<std::vector<double>> configs;
```
fweik, line 54: This should be a `std::vector<Utils::Vector3d>`, the inner size is known statically. This is really wasteful...

src/core/statistics_chain.cpp
```diff
@@ -45,14 +46,12 @@ void update_mol_ids_setchains() {
   }
 }
 
-void calc_re(PartCfg &partCfg, double **_re) {
-  int i;
-  double dx, dy, dz;
+std::vector<double> calc_re(PartCfg &partCfg) {
```
fweik: Shouldn't be on the heap, size is known

src/core/statistics_chain.cpp
```diff
@@ -64,26 +63,27 @@ void calc_re(PartCfg &partCfg, double **_re) {
...
   re[0] = dist / tmp;
   re[2] = dist2 / tmp;
   re[1] = sqrt(re[2] - re[0] * re[0]);
   re[3] = sqrt(dist4 / tmp - re[2] * re[2]);
+  return re;
 }
 
-void calc_rg(PartCfg &partCfg, double **_rg) {
-  int i, j, p;
+std::vector<double> calc_rg(PartCfg &partCfg) {
```
fweik: Shouldn't be on the heap, the size is known.

src/core/statistics_chain.cpp
```diff
@@ -106,26 +106,25 @@ void calc_rg(PartCfg &partCfg, double **_rg) {
...
   rg[3] = sqrt(r_G4 / tmp - rg[2] * rg[2]);
+  return rg;
 }
 
-void calc_rh(PartCfg &partCfg, double **_rh) {
-  int i, j, p;
-  double dx, dy, dz, r_H = 0.0, r_H2 = 0.0, *rh = nullptr, ri = 0.0, prefac,
-                     tmp;
-  *_rh = rh = Utils::realloc(rh, 2 * sizeof(double));
+std::vector<double> calc_rh(PartCfg &partCfg) {
```
fweik: Shouldn't be on the heap, the size is known.

fweik (changes requested):

KaiSzuttor:
@fweik you are right, but by matter of efficiency regarding the effort to merge and due to the fact that all these functions are not in the hot loop plus they don't use manual memory management anymore, I suggest we merge this soon

fweik:
This does not need any serious changes, the types can just be switched and the resizing removed.
I really would like to not have code like this, it's just confusion because everybody reading this will
wonder why it is done that way.

KaiSzuttor:
ping @jonaslandsgesell

src/core/grid_based_algorithms/lb_interface.cpp
jonaslandsgesell: yep

src/core/grid_based_algorithms/lbgpu.cpp
jonaslandsgesell: There is multiple cals of resize for host_values in lb_interface.cpp :
`host_values.resize(lbpar_gpu.number_of_nodes);`
so the size of host_values is not statically known.

src/core/statistics.cpp
jonaslandsgesell: If the number of particles changes in a simulation then the inner size is not statically known but changes during the runtime: `  configs[n_configs].resize(3 * n_part_conf);` (in analyze_configs() in statistics.cpp)

One could introdce a `std::vector<std::vector<Utils::Vector3d>> configs`,
which one would enlarge via configs[n_configs].resize(n_part_conf) and adress like `configs[n_configs][particle_id]=p.r.p`, however I am not sure what the benefit is compared to:
`  int i = 0;
  for (auto const &p : partCfg) {
    configs[n_configs][3 * i + 0] = p.r.p[0];
    configs[n_configs][3 * i + 1] = p.r.p[1];
    configs[n_configs][3 * i + 2] = p.r.p[2];
    i++;
  }`

src/core/statistics_chain.cpp
jonaslandsgesell: yep

src/core/statistics_chain.cpp
jonaslandsgesell: yep

src/core/statistics_chain.cpp
jonaslandsgesell: yep

src/core/grid_based_algorithms/lb_interface.cpp
```diff
@@ -1052,10 +1052,7 @@ double lb_lbnode_get_density(const Utils::Vector3i &ind) {
 #ifdef CUDA
     int single_nodeindex = ind[0] + ind[1] * lbpar_gpu.dim_x +
                            ind[2] * lbpar_gpu.dim_x * lbpar_gpu.dim_y;
-    static std::vector<LB_rho_v_pi_gpu> host_print_values(0);
-
-    if (host_print_values.empty())
-      host_print_values.resize(1);
+    static std::array<LB_rho_v_pi_gpu, 1> host_print_values;
```
fweik: array of size 1 looks suspicious. Why is this an array?

jonaslandsgesell: True. It was an array before, I just kept the memory access like it was. Removed the array now.

KaiSzuttor:
@jonaslandsgesell could you please resolve the conflict and make @fweik happy?

mkuron:
Please merge the latest master to fix the rocm build.

RudolfWeeber:
Can we go ahead with this?

src/core/grid_based_algorithms/lbgpu.cpp
fweik: Ah sorry, I did not realize that this is different from `host_print_values`

src/core/statistics.cpp
fweik: The inner dimension is, as you are saying, `3`. The advantage would be that e.g. it's not possible to have an incompatible size, also the code get less noise, because there is no need for elementwise copy.

fweik (approved):

src/core/statistics.cpp
jonaslandsgesell: How about saving this for an independent PR. The corresponding changes would require further changes in the way configs[I] is handled? This is not related to memory management though.

fweik:
Yes, I did already approve. But you wrote that you can not see tye benefit
of that, so I pointed it out to you.

PR #2935, opened 2019-06-24
armhf: prevent OpenMPI issue on bwCloud

### LINKED ISSUE #2934: OpenMPI incompatible with 32bit builds using overlay2 backend
The `ubuntu:armhf` container cannot use OpenMPI on the bwCloud (see espressomd/docker#108). In the rare event where a scheduled pipeline must be restarted during the day, while many jobs are already running, there is a chance the `ubuntu:armhf` job will be assigned to a bwCloud machine, which will cause a failure:
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/128008

jngrad:
Fixes #2934

The `ubuntu:armhf` container cannot run on bwCloud.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -332,6 +332,10 @@ ubuntu:arm64:
 
 ubuntu:armhf:
   <<: *arch_definition
+  tags:
+    - docker
+    - linux
+    - icp
 
 ubuntu:i386:
   <<: *arch_definition


mkuron:
This is only a temporary and insufficient fix. As discovered in #2934, it affects all emulated containers and the i386 container, due to their use of a newer OpenMPI version than the regular build containers. Once our new build machines with AMD Epyc CPUs are delivered, this problem will also appear outside of bwCloud. So we actually need to figure out why newer OpenMPI versions don't work inside Docker on AMD.

jngrad:
The other exotic archs could both build and run tests on bwCloud:
https://gitlab.icp.uni-stuttgart.de/espressomd/docker/-/jobs/128171
https://gitlab.icp.uni-stuttgart.de/espressomd/docker/-/jobs/127926
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/128010
You show in #2934 the issue is more widespread, but as far as CI is concerned, we've only seen it with armhf on bwCloud+elk. While we investigate the source of the OpenMPI issue, we should either deliver a temporary fix to stop the frequent job failures, or make the slow exotic archs fully manual including on the release branches as proposed by @RudolfWeeber in his annotation of the espresso meeting summary.

mkuron:
I‘ll try to track down the OpenMPI issue tomorrow. I think some of bwCloud is on AMD and most is Intel, that‘s why we don‘t see it consistently. I forgot about elk though — that has an older Intel CPU, so I don‘t know how it fits into the pattern.

PR #2937, opened 2019-06-24
hip: fully fix build

mkuron:
Fixes #2901.
Fixes #2906.
@KaiSzuttor 

I know it's an ugly workaround, but it's the best I could come up with. You can't use variadic template functions here.

Workaround from https://github.com/ROCm-Developer-Tools/HIP/blob/91f82ce541b51ad1106a28e951b410a256e61f62/docs/markdown/hip_bugs.md#errors-related-to-no-matching-constructor, found via https://github.com/RadeonOpenCompute/hcc/issues/860

Hopefully the underlying bug goes away once AMD drops HCC and uses Clang directly.


KaiSzuttor:
i don't particularly like it

mkuron:
Neither do I, but this is not a beauty contest. It's the official workaround and should soon be taken care of by the compiler.

KaiSzuttor:
it is, kind of. elegance and maintainability have quite some correlation.

KaiSzuttor:
So, in general the question is not, *if* we are able to support HIP any longer but how large the time costs are for the developers. At some point we have to make a cut and wait for the HIP framework to be stable and more mature. The bug list you referenced above was last touched two years ago, which lets me wonder if they are still committed to fix those.

src/utils/include/utils/Array.hpp
```diff
@@ -125,6 +125,42 @@ template <typename T, std::size_t N> struct Array {
...
+  __attribute__((annotate("serialize"))) void
+  __cxxamp_serialize(Kalmar::Serialize &s) const {
+    for (int i = 0; i < N; ++i)
+      s.Append(sizeof(T), &m_storage[i]);
+    for (int i = N; i < 10; ++i)
+      s.Append(sizeof(T), T());
+  }
+
+  __attribute__((annotate("user_deserialize"))) void
+  Foo(T x0, T x1, T x2, T x3, T x4, T x5, T x6, T x7, T x8, T x9)[[cpu]][[hc]] {
```
fweik: What is the function of this method, and why is it called `Foo`?

mkuron: This is the deserialization method and it needs as many arguments as there were serialized parameters. `Foo` is a copy-and-paste leftover (it should actually be the name of the class, to be another constructor). It seems like it is resolved via the annotation attribute, so the name does not matter.

I think I may have found a simpler solution, will update the pull request soon.

RudolfWeeber (approved):
Itś clear that the workaround is not pretty, but since we deferred a decision to remove HIP support, this PR should be merged.

src/utils/include/utils/Array.hpp
```diff
@@ -125,6 +125,14 @@ template <typename T, std::size_t N> struct Array {
     return ret;
   }
 
+#ifdef __HCC__
+  // workaround for https://github.com/RadeonOpenCompute/hcc/issues/860
+  __attribute__((annotate("serialize"))) void
```
KaiSzuttor, line 130: @RudolfWeeber can you explain what these lines do?

mkuron: They just force the compiler into a certain overload resolution. Function prototypes are sufficient since we don't actually need to serialize data.

RudolfWeeber:
„KaiSzuttor commented on this pull request.

  _____  

In src/utils/include/utils/Array.hpp <https://github.com/espressomd/espresso/pull/2937#discussion_r300457808> :

> @@ -125,6 +125,14 @@ template <typename T, std::size_t N> struct Array {
     return ret;
   }
 
+#ifdef __HCC__
+  // workaround for https://github.com/RadeonOpenCompute/hcc/issues/860
+  __attribute__((annotate("serialize"))) void

@RudolfWeeber <https://github.com/RudolfWeeber>  can you explain what these lines do?

The comment and the ifdef makes it clear what it is for, and as long as we keep HIP support, this needs to be merged.

I don’t see a benefit in delaying.

If the HIP project is active, the bug will be fixed and the workaround no longer needed. If not, HIP support will have to be removed in any case.

KaiSzuttor:
why don't we have merged then already?

PR #2939, opened 2019-06-25
Documentation of VV, NPT, Thermostats

### LINKED ISSUE #2680: Integration scheme not documented
The integration scheme for Langevin dynamics is not documented, at least I could not find something in the documents or code comments. LD can be done in different ways with a Verlet scheme, in Espresso it is even more complex because it also supports dissipative and inertia coupling to LB, DPD, anisotropy and more, as well as combinations of those. I think this is a central point that should absolutely be documented an well understood.

In my opinion the following points should be addressed, not only documenting what
_is_ implemented, but also explaining why it is done in a specific way if there is a choice:

- Discretization of Langevin Dynamics, including extensions like anisotropy, body and
  space fixed external foces, more complex drift terms like DPD, citing literature sources for the     validity of the scheme.
- Discretization scheme for NpT LD, including in combination with other extensions,
  citing literature sources for the     validity of the scheme.
- Discretization of the dissipative coupling with LB, citing literature sources for the validity
  of the coupled scheme.
- Discretization of the inertia coupling with LB, citing literature sources for the validity of
  the coupling scheme.

Without having clarity about these points, we can in my opinion not say with confidence
what the code actually calculates, and whether the numeric calculations approximate
some physical model in any limit. 
This understanding is also needed to properly handle initial forces in the future.

christophlohrmann:
Fixes #2680 

Description of changes:
 - Added sphinx documentation for velocity verlet, NPT
- Added/Updated sphinx documentation for Langevin, NPT, LB, DPD thermostat
- Found some papers as reference

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [x] Docs?


doc/sphinx/inter_non-bonded.rst
```diff
@@ -559,60 +559,78 @@ DPD interaction
...
+that contains velocity-dependent friction and noise according 
+to a temperature set by :py:attr:`espressomd.thermostat.Thermostat.set_dpd()`
+. The parameters are
 
-The interaction consists of a dissipative force :math:`\vec{F}_{ij}^{D}` and
-a random force :math:`\vec{F}_{ij}^R`, and is decomposed into a component
-parallel and perpendicular to the distance vector of the particle pair :math:`\vec{F}_{ij}`.
-The parameters for the two parts can be chosen independently.
-The force contributions of the parallel part are
+* ``gamma``
```
KaiSzuttor, line 572: where is the temperature parameter?

doc/sphinx/lb.rst
```diff
@@ -121,53 +121,53 @@ load in the particles with the correct forces, and use::
...
-To enable the LB thermostat, use::
+To get interpolated velocity values between lattice nodes, the function::
 
-    sys.thermostat.set_lb(LB_fluid=lbf, seed=123, gamma=1.5)
+    lb.get_interpolated_velocity(pos = [1.1,1.2,1.3])
+    
+with a single position  ``pos`` as an argument can be used. 
+For the GPU fluid :class:`espressomd.lb.LBFluidGPU`
+also :py:attr:`espressomd.lb.LBFluidGPU.get_interpolated_fluid_velocity_at_positions()`
+is available, which expects a numpy array of positions as an argument.
```
KaiSzuttor, line 139: could you please also mention the issue of interpolating velocities near boundaries?

doc/sphinx/inter_non-bonded.rst
christophlohrmann: The temperature is set by the dpd thermostat (Thermostat.set_dpd()) for all particles. The friction parameters etc are set by the dpd interaction

KaiSzuttor: ah it's maybe confusing because you mention the thermostat and directly after that you continue with "The parameters are"

christophlohrmann: I will clarify that

doc/sphinx/inter_non-bonded.rst
```diff
@@ -559,60 +559,78 @@ DPD interaction
...
     system.non_bonded_inter[type1, type2].dpd.set_params(**kwargs)
 
-This command defines a velocity-dependent interaction
-between particles of the types ``type1`` and ``type2``. For a description of the input arguments
-see :class:`espressomd.interactions.DPDInteraction`. The interaction
-only has an effect if the DPD thermostat activated, and acts according to the
-temperature of the thermostat.
+This command defines an interaction between particles of the types ``type1`` and ``type2``
+that contains velocity-dependent friction and noise according 
+to a temperature set by :py:attr:`espressomd.thermostat.Thermostat.set_dpd()`
```
jngrad: A mix of `:py:attr:` and `:py:func:` is used here and elsewhere to link to class methods. Although the HTML output looks the same, the documentation would probably be easier to maintain if we have a consistent use of `:meth:` for class methods, `:attr:` for class members and `:func:` for free functions, without the `:py:` prefix.

doc/sphinx/system_setup.rst
```diff
@@ -263,64 +263,135 @@ Best explained in an example::
...
+ 
+.. math:: <\overline{\eta}> = 0 , <\overline{\eta}\overline{\eta}> = 1/dt
 
 The keyword ``seed`` controls the state of the random number generator (Philox
 Counter-based RNG) and is required on first activation of the thermostat. It
 can be omitted in subsequent calls of ``set_langevin()``. It is the user's
 responsibility to decide whether the thermostat should be deterministic (by
 using a fixed seed) or not (by using a randomized seed).
 
+The diffusion coeffictient :math:`D` of the particle can be obtained by the Einstein-Smoluchowski relation
```
jonaslandsgesell: We should discuss to what degree the plain Einstein-Smoluchowski relation is applicable to the gamma which is set in the thermostat. In my mind what is presented in the text only holds for ideal systems. Interactions in the system will modify the DIffusion coefficient -- otherwise there would never be a need to measure diffusion coefficients using the mean squared displacement or a green kubo formula in MD simulations.

christophlohrmann:
Concerning the dimensions enabled/disabled for scaling in the NPT algorithm:

The corresponding code has not been touched since the original commit from 2004.
The author (Ira Cooke) did not describe this method in any of his papers and I cannot
find his dissertation on the Max Planck homepage. Also I did not find any other paper 
using that method.

Should I just mark that option as not tested and leave it like that?

doc/sphinx/system_setup.rst
fweik: I agree with @jonaslandsgesell, the documentation should not make statements about the diffusion coefficient here.

jngrad:
mark NPT as experimental and update "Usage Status = None" in:
https://github.com/espressomd/espresso/blob/python/doc/sphinx/introduction.rst#available-simulation-methods

doc/sphinx/system_setup.rst
christophlohrmann: @jonaslandsgesell I would remove the mention of the diffusion coefficient entirely, as @fweik suggested.

jonaslandsgesell: Hey Christoph. Why not removing it?yes, I mean for ideal (noninteracting ) particles one should observe D=kT/gamma_thermostat. So we state that gamma is the friction of the thermostat somewhere?

christophlohrmann: We say that gamma is the bare friction coefficient for the Langevin equation and any physics beyond that should be the user's responsibility.

jonaslandsgesell: Sounds good!

RudolfWeeber (approved):
bors r+

RudolfWeeber:
I checked NPT against Monte Carlo volume moves. It seems to work, in principle.

The PR improves the docs alot. I'd suggest not to defer much longer and instead open a new pr if there is something to add.

jngrad:
bors r=RudolfWeeber

PR #2941, opened 2019-06-25
Move intel and qemu builds to tags and release branches only

### LINKED ISSUE #2942: Implemented CI changes discussed in Offline Meeting
[The changes we agreed on in the offline meeting](https://github.com/espressomd/espresso/wiki/Espresso-meeting-2019-06-25#simplifying-the-ci-infrastructure) should be implemented. See also https://github.com/espressomd/docker/issues/110.

mkuron:
As discussed earlier. The regex matches any branch that contains a dot, e.g. _4.0_. Fixed #2942.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -325,7 +325,8 @@ rocm-maxset:
     - docker
     - linux
   only:
-    - schedules
+    - /.*\..*/
+    - tags
 
 ubuntu:arm64:
   <<: *arch_definition
@@ -395,6 +396,9 @@ intel:18:
     - linux
     - cuda
     - icp
+  only:
+    - /.*\..*/
+    - tags
 
 check_sphinx:
   <<: *global_job_definition


jngrad:
Didn't we mean to keep the `i386` because it was fast enough?

fweik:
@jngrad i386 is a regular CI job, not a scheduled one

jngrad:
right, YAML overwrites the `only:` node

fweik:
My understanding was that we remove intel completely. This would allow to also remove the support from cmake, which would e.g. help to clean up the build flags.

KaiSzuttor:
moving the intel tests to release does not reduce the complexity, also the CI for espresso in the intel container is not at all an issue. @fweik  and me had in mind to remove the intel support completely as discussed offline

mkuron:
> remove the intel support completely as discussed offline

That's not how I understood our conclusion. I thought we would treat it the same as the qemu builds. We actually have someone who uses the Intel compiler (@bindgens1), it's the default compiler on some clusters, and it has the same frontend as the Cray compiler, which is the default (and best supported) compiler on Cray machines.

KaiSzuttor:
Do we mirror tags to gitlab? I'd use ```when: manual```

RudolfWeeber:
Manually triggering would be enough, I think

mkuron:
We mirror all tags and branches. These builds should definitely run automatically on tags, so we need `only: tags`. Unfortunately that can't be combined with `when: manual`, so we need to match the release branches via a regex.

RudolfWeeber:
On Tue, Jun 25, 2019 at 07:11:51AM -0700, Michael Kuron wrote:
> We mirror all tags and branches. These builds should definitely run automatically on tags, so we need `only: tags`. Unfortunately that can't be combined with `when: manual`, so we need to match the release branches via a regex.
I'd go for manual. The branch chosing rule can lead to surprises and  releases are manual processes anyway. Also, the CI has to be run, before the tag is created.

KaiSzuttor:
sorry if I'm missing something here but we really don't get rid of any complexity of our CI here. IIRC before the meeting nobody of the core developers intended to just remove the jobs for the espresso CI. Can please somebody enlighten me?

RudolfWeeber:
I’d also have like to see the emulated architectures removed entirely.

However, I can live with the compromise with both the jobs in the docker repo and the ones in the espressomd repo being set to manual-only.

Then, we don’t have to put up with it during the normal development process.

Also, the weekly CI rebuilds will have a much higher chance of going through.

 

At the time of release, the stuff can be run once and a decision be made based on the magnitude of issues.

mkuron:
At the meeting, we agreed to keep both intel and qemu but run them on releases only, see https://github.com/espressomd/espresso/wiki/Espresso-meeting-2019-06-25#simplifying-the-ci-infrastructure.

KaiSzuttor:
IIRC, we decided to keep a single 32 bit build without quemu, because it will simplify the CI for the docker repo. I personally don't care when the tests using those containers are run... Maybe that was not clear in the meeting.

mkuron:
The 32-bit container was supposed to stay in the regular CI, as it is now. The CI is made faster by making the QEMU builds there manual.

.gitlab-ci.yml
```diff
@@ -325,7 +325,8 @@ rocm-maxset:
     - docker
     - linux
   only:
-    - schedules
+    - /.*\..*/
```
KaiSzuttor, line 328: please make the job ```when: manual``` and remove the ```only```

KaiSzuttor (changes requested):

RudolfWeeber:
On Tue, Jun 25, 2019 at 10:36:04PM -0700, Michael Kuron wrote:
> At the meeting, we agreed to keep both intel and qemu but run them on releases only, see https://github.com/espressomd/espresso/wiki/Espresso-meeting-2019-06-25#simplifying-the-ci-infrastructure.
I understood the outcome differently and have annotated the minutes accordingly.

KaiSzuttor:
closed due to #2963

PR #2946, opened 2019-06-26
cmake: add soversion to libH5mdCore

junghans:
Found on Fedora package.

This fix might not be needed on master as libs get installed in python directory if I recall correctly.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -229,6 +229,7 @@ ubuntu:wo-dependencies:
   tags:
     - docker
     - linux
+  when: manual
 
 ubuntu:arm64:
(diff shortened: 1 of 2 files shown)


KaiSzuttor:
CI is broken for 4.0

RudolfWeeber:
Since we will not provide another bugfix release for 4.0, we might as well disable all checks except ubuntu with python 2/3 on x86_64.

jngrad:
There is no `bors.toml` file on 4.0.

fweik:
Ah of course. Thank you!

jngrad:
I would have made the same mistake, actually :smile:

PR #2947, opened 2019-06-27
Make particle sorting unique for nsquare

fweik:
The nsq cell system did not have a unique mapping of
particles to nodes/cells (same thing), which was causing me trouble. Now they are mapped
to the node by their id, which also allowed to simplify the particle exchange considerably and
a global mapping of positions to nodes is no longer needed.

Description of changes:
 - Allow particle sorting by other properties than position
 - Sort by particle id for nsquare cell system
 - Removed CellStructure::position_to_node

### DIFF
--- a/src/core/MpiCallbacks.hpp
+++ b/src/core/MpiCallbacks.hpp
@@ -544,9 +544,9 @@ public:
    *
    * This method can only be called on the head node.
    */
-  template <class R, class... Args>
+  template <class R, class... Args, class... ArgRef>
   auto call(Result::OneRank, boost::optional<R> (*fp)(Args...),
-            Args... args) const -> std::remove_reference_t<R> {
+            ArgRef... args) const -> std::remove_reference_t<R> {
 
     const int id = m_func_ptr_to_id.at(reinterpret_cast<void (*)()>(fp));
     call(id, args...);
--- a/src/core/communication.hpp
+++ b/src/core/communication.hpp
@@ -142,7 +142,7 @@ void mpi_reshape_communicator(std::array<int, 3> const &node_grid,
  *  \param node  the node to attach it to.
  *  \param pos   the particles position.
  */
-void mpi_place_particle(int node, int id, double pos[3]);
+void mpi_place_particle(int node, int id, const Utils::Vector3d &pos);
 
 /** Issue REQ_PLACE: create particle at a position on a node.
  *  Also calls \ref on_particle_change.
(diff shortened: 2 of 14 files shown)


fweik:
@KaiSzuttor

src/core/nsquare.cpp
```diff
@@ -34,8 +34,9 @@
 
 Cell *local;
 
-Cell *nsq_position_to_cell(const Utils::Vector3d &pos) { return local; }
-int nsq_position_to_node(const Utils::Vector3d &) { return this_node; }
+Cell *nsq_position_to_cell(int id) {
```
KaiSzuttor: isn't this name misleading?

src/core/particle_data.cpp
```diff
@@ -833,26 +833,17 @@ void prefetch_particle_data(std::vector<int> ids) {
                            std::make_move_iterator(parts.begin()));
 }
 
-int place_particle(int part, double p[3]) {
-  int retcode = ES_PART_OK;
+int place_particle(int part, const double *p_) {
```
KaiSzuttor: why not ```Utils::Vector3d```? and why that name?

src/core/nsquare.cpp
fweik: Ah, I missed that one

src/core/particle_data.cpp
fweik: Changing the type would require updating all the call sites, which is out of scope.

src/core/MpiCallbacks.hpp
```diff
@@ -544,9 +544,9 @@ class MpiCallbacks {
    *
    * This method can only be called on the head node.
    */
-  template <class R, class... Args>
+  template <class R, class... Args, class... ArgRef>
```
KaiSzuttor, line 547: do you think that we should make sure that ```Args``` and ```ArgRef``` are mutually convertible via SFINAE?

fweik: There is no other overload that could be picked up, so actually substitution failure is an error here.

src/core/particle_data.cpp
fweik: As a token of good will I changed some other signatures ^^

src/core/communication.hpp
```diff
@@ -142,7 +142,7 @@ void mpi_reshape_communicator(std::array<int, 3> const &node_grid,
  *  \param node  the node to attach it to.
  *  \param pos   the particles position.
  */
-void mpi_place_particle(int node, int id, double pos[3]);
+void mpi_place_particle(int node, int id, const Utils::Vector3d &pos);
```
KaiSzuttor, line 145: you totally lost track on east const 😞

KaiSzuttor (approved):

PR #2949, opened 2019-06-27
LB mpi callback

KaiSzuttor:


KaiSzuttor:
@fweik 

```
note: candidate template ignored: deduced
      conflicting types for parameter 'Args' (<const Utils::Vector<int, 3> &> vs. <Utils::Vector<int,
      3>>)
  auto call(Result::OneRank, boost::optional<R> (*fp)(Args...),
```

any idea?

KaiSzuttor:
the ```is_local``` check seems not to work properly... but I don't want to fiddle around with the unreadable lattice class.

fweik:
What do you mean? The condition is correct no?

KaiSzuttor:
the lb unit test hangs even when run on a single core, I assume because the ```is_local``` function returns false.

fweik:
This would cause an assertion failure in a debug build

KaiSzuttor:
really? which ```assert``` should trigger here?

fweik:
https://github.com/espressomd/espresso/blob/python/src/core/MpiCallbacks.hpp#L163
and https://github.com/espressomd/espresso/blob/python/src/core/MpiCallbacks.hpp#L556 on the head node

KaiSzuttor:
here it its:
```
Communication::MpiCallbacks::call(Communication::Result::OneRank, boost::optional<T> (*)(FArgs ...), Args ...) const [with R = double; FArgs = {const Utils::Vector<int, 3ul>&}; Args = {Utils::Vector<int, 3ul>}; std::remove_reference_t<T> = double]: Assertion `1 == boost::mpi::all_reduce(m_comm, static_cast<int>(!!local_result), std::plus<>())' failed
```
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/130841

RudolfWeeber:
:.,$d
@KaiSzuttor, the Walberla branch on my github repo might be of interest. As I understnad it, the parallelization scheme is the same for LBCPU and Walberla.

-- 
Dr. Rudolf Weeber
Institute for Computational Physics
Universität Stuttgart
Allmandring 3
70569 Stuttgart
Germany
Phone: +49(0)711/685-67717
Email: weeber@icp.uni-stuttgart.de
http://www.icp.uni-stuttgart.de/~icp/Rudolf_Weeber

KaiSzuttor:
@RudolfWeeber yes it is, but actually the issue seems to be my lack of willingness to understand the lattice class

fweik:
The problem is:
```
bool Lattice::is_local(const Vector3i&) const index 0 0 11 local_index_offset -1 -1 -1 grid 12 12 12
```
:-)

KaiSzuttor:
yes, my motivation got as far as printing those values aswell...

fweik:
Just check it by the position, that should work. (like `offset + index * agrid >= my_left` and so on). That matches the *style* of the rest of the class

src/core/grid_based_algorithms/lb_interface.cpp
```diff
@@ -17,88 +18,93 @@ ActiveLB lattice_switch = ActiveLB::NONE;
...
- */
-void mpi_send_fluid(int node, int index, double rho, Utils::Vector3d const &j,
-                    Utils::Vector6d const &pi) {
-  if (node == this_node) {
-    lb_calc_n_from_rho_j_pi(index, rho, j, pi);
-  } else if (0 == this_node) {
-    mpi_call(mpi_send_fluid, node, index, rho, j, pi);
+
+template <typename Kernel>
+void lb_set(Utils::Vector3i const &index, Kernel &&kernel) {
```
fweik: There is no need for the forwarding reference here, just say `Kernel kernel`

src/core/grid_based_algorithms/lb_interface.cpp
```diff
@@ -17,88 +18,93 @@ ActiveLB lattice_switch = ActiveLB::NONE;
...
 
-REGISTER_CALLBACK(mpi_send_fluid)
-
-void mpi_recv_fluid_slave(int node, int index) {
-  if (node == this_node) {
-    double data[10];
-    lb_calc_local_fields(index, &data[0], &data[1], &data[4]);
-    MPI_Send(data, 10, MPI_DOUBLE, 0, SOME_TAG, comm_cart);
+template <typename Kernel>
+auto lb_calc(Utils::Vector3i const &index, Kernel &&kernel) {
```
fweik: There is no need for the forwarding reference here, just say `Kernel kernel`

src/core/grid_based_algorithms/lb_interface.cpp
```diff
@@ -17,88 +18,93 @@ ActiveLB lattice_switch = ActiveLB::NONE;
...
+  using R = decltype(kernel(index));
+  if (lblattice.is_local(index)) {
+    return boost::optional<R>(kernel(index));
   }
+  return boost::optional<R>();
 }
 
-REGISTER_CALLBACK(mpi_recv_fluid_slave)
+template <class Kernel>
+auto lb_calc_fluid_kernel(Utils::Vector3i const &index, Kernel &&kernel) {
```
fweik: also no `&&`

fweik (changes requested):
looks good except for some minor cosmetics

fweik (approved):

PR #2951, opened 2019-06-27
Require Python 3

mkuron:
This is needed for #2950 (otherwise the Apple-provided Python 2.7 is picked up), but must not be merged until all CI builds have been moved to Python 3.


RudolfWeeber:
I already updated the installation instructions further below in the UG,
such that the Python binary is explicitly specified.

Once we actually amek the jump (soon), the "from __future__" sutuff needs to be removed as well.

PR #2957, opened 2019-06-28
WIP: Disable ghost shifts for domain decomposition

fweik:
Description of changes:
 - Refactor of DD neighbor finding 
 - Removed ghost shift for DD

MD works with this, LB coupling is broken, as expected.


fweik:
@RudolfWeeber: This is merged into #2976, to think we should try to get this into mainline independently?

PR #2965, opened 2019-07-01
utils: memory order as parameter for get_linear_index

fweik:
This is mostly to document what the memory order is, and I need
row major for the p3m. @jngrad do you have an idea how to document
the tag parameter in doxygen? I'd rather not name it because it's
not used...

### DIFF
--- a/src/utils/include/utils/index.hpp
+++ b/src/utils/include/utils/index.hpp
@@ -85,6 +85,8 @@ get_linear_index(int a, int b, int c, const Vector3i &adim,
     return a + adim[0] * (b + adim[1] * c);
   case MemoryOrder::ROW_MAJOR:
     return adim[1] * adim[2] * a + adim[2] * b + c;
+  default:
+    throw std::runtime_error("Unknown memory order");
   }
 }
 
--- a/src/utils/include/utils/index.hpp
+++ b/src/utils/include/utils/index.hpp
@@ -84,7 +84,6 @@ inline int get_linear_index(int a, int b, int c, const Vector3i &adim,
   return a + adim[0] * (b + adim[1] * c);
 }
 
-
 /** get the linear index from the position (@p a,@p b,@p c) in a 3D grid
  *  of dimensions @p adim, , usiung row major memory order.
  *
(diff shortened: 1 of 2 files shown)


jngrad:
@fweik unnamed parameters cannot be documented via `@param`. You can either mention in the long description how the last argument is just a tag for the function template, or name the argument, e.g. `major_order`. In the future with C++17 these 3 functions could probably be condensed into a single function template using `if(std::is_same<T, MemoryOrder::RowMajor>::value)`.

src/utils/include/utils/index.hpp
```diff
@@ -58,28 +58,51 @@ inline void unravel_index(InputIterator dimensions_begin,
   }
 }
 
+namespace MemoryOrder {
+constexpr struct RowMajor {
+} row_major;
+constexpr struct ColumnMajor { } coulumn_major; } // namespace MemoryOrder
```
jngrad: typo + formatting
```suggestion
constexpr struct ColumnMajor {
} column_major;
} // namespace MemoryOrder
```

fweik: Actually this is what clang-format wants

src/utils/include/utils/index.hpp
```diff
@@ -58,28 +58,51 @@ inline void unravel_index(InputIterator dimensions_begin,
...
+
 /*************************************************************/
 /** \name Three dimensional grid operations                  */
 /*************************************************************/
 /*@{*/
 
 /** get the linear index from the position (@p a,@p b,@p c) in a 3D grid
- *  of dimensions @p adim.
+ *  of dimensions @p adim, usiung column major memory order.
```
jngrad:
```suggestion
 *  of dimensions @p adim, using column major memory order.
```

fweik:
Well there is no real reason here for static dispatch, see updated version.

RudolfWeeber (approved):

jngrad:
But the PR didn't pass CI... We still need to replace the switch statement by an if/else.

RudolfWeeber:
Sorry, didn notice that.

RudolfWeeber:
bors cancel

fweik:
I like the switch better

jngrad (approved):

PR #2970, opened 2019-07-03
Local box 

fweik:
Description of changes:
 - Put info about local box into struct
 - Get node grid info from communicator instead of storing
 - some wrappers around cartesian communicator functions


KaiSzuttor:
this will be fun to merge with #2949

src/core/LocalBox.hpp
```diff
@@ -0,0 +1,24 @@
...
+#include <utils/Vector.hpp>
+
+template <class T> class LocalBox {
+public:
+  /** Dimensions of the box a single node is responsible for. */
+  Utils::Vector<T, 3> local_box_l = {1, 1, 1};
+  /** Left (bottom, front) corner of this nodes local box. */
+  Utils::Vector<T, 3> my_left_ = {0, 0, 0};
+
+  auto const &my_left() const { return my_left_; }
```
KaiSzuttor: please document all members

src/core/LocalBox.hpp
```diff
@@ -0,0 +1,24 @@
+#ifndef ESPRESSO_LOCALBOX_HPP
+#define ESPRESSO_LOCALBOX_HPP
+
+#include <utils/Vector.hpp>
+
+template <class T> class LocalBox {
+public:
```
KaiSzuttor: wouldn't it be cleaner to make the members private and have a constructor such that an instance of ```LocalBox``` cannot be in an inconsistent state

KaiSzuttor (changes requested):

fweik:
@KaiSzuttor I made some changed to `Lattice` in the merge

jngrad:
Not sure why CI had trouble pulling images. It affected other PRs as well. It is back to normal now.

src/core/LocalBox.hpp
```diff
@@ -0,0 +1,39 @@
...
+  /** @brief Boundary information for the local box.
+   *
+   * This returns for each of the faces of the local box if
+   * it is a boundary of the simulation box. The format is
+   * as follows:
+   *  (x low, x high, y low, y high, z low, z high).
+   *
+   * @return Array with boundary information.
+   */
+  Utils::Array<int, 6> const &boundary() const { return m_boundaries; }
```
KaiSzuttor, line 36: why not bool?

src/core/LocalBox.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+template <class T> class LocalBox {
+  Utils::Vector<T, 3> m_local_box_l = {1, 1, 1};
+  Utils::Vector<T, 3> m_lower_corner = {0, 0, 0};
+  Utils::Vector<T, 3> m_upper_corner = {1, 1, 1};
+  Utils::Array<int, 6> m_boundaries = {};
+
+public:
+  LocalBox() = default;
+  LocalBox(Utils::Vector<T, 3> const &lower_corner,
```
KaiSzuttor, line 14: this looks much better now

src/core/grid.cpp
```diff
@@ -75,88 +60,64 @@ int map_position_node_array(const Utils::Vector3d &pos) {
...
+      get<1>(cart_shift(comm, 0, -1)), get<1>(cart_shift(comm, 0, +1)),
+      get<1>(cart_shift(comm, 1, -1)), get<1>(cart_shift(comm, 1, +1)),
+      get<1>(cart_shift(comm, 2, -1)), get<1>(cart_shift(comm, 2, +1)),
+  };
+}
 
-  map_node_array(node, node_pos.data());
-  for (dir = 0; dir < 3; dir++) {
-    int buf;
+LocalBox<double> regular_decomposition(const BoxGeometry &box,
```
KaiSzuttor, line 86: there is no chance I get a test for this free function?

src/core/grid.cpp
```diff
@@ -75,88 +60,64 @@ int map_position_node_array(const Utils::Vector3d &pos) {
...
+  return Utils::Mpi::cart_rank(comm_cart, im);
 }
 
-int calc_node_neighbors(int node) {
+Utils::Vector3i calc_node_pos(const boost::mpi::communicator &comm) {
+  return Utils::Mpi::cart_coords<3>(comm, comm.rank());
+}
 
-  int dir, neighbor_count;
+Utils::Vector<int, 6>
```
KaiSzuttor, line 74: and maybe a test for that function?

src/core/grid.hpp
```diff
@@ -91,44 +63,30 @@ extern Utils::Vector3d my_right;
...
+ */
+Utils::Vector<int, 6> calc_node_neighbors(const boost::mpi::communicator &comm);
+
+/**
+ * @brief Calculate the position of node in topology.
+ *
+ * @param comm Cartesian communicator
+ * @return Index of node in grid.
+ */
+Utils::Vector3i calc_node_pos(const boost::mpi::communicator &comm);
```
KaiSzuttor, line 83: why don't we use https://www.boost.org/doc/libs/1_68_0/doc/html/boost/mpi/cartesian_communicator.html
?

src/core/grid_based_algorithms/lattice.cpp
```diff
@@ -117,7 +116,7 @@ void Lattice::map_position_to_lattice(const Utils::Vector3d &pos,
 
 Utils::Vector3i Lattice::local_index(Utils::Vector3i const &global_index) const
     noexcept {
-  return {global_index[0] - node_pos[0] * grid[0] + halo_size,
-          global_index[1] - node_pos[1] * grid[1] + halo_size,
-          global_index[2] - node_pos[2] * grid[2] + halo_size};
+  return {global_index[0] - local_index_offset[0] + halo_size,
```
KaiSzuttor, line 119: we are almost there... having a readable ```Lattice``` class

src/core/grid_based_algorithms/lb_boundaries.cpp
```diff
@@ -223,13 +223,15 @@ void lb_init_boundaries() {
...
     int the_boundary = -1;
-    map_node_array(this_node, node_domain_position.data());
+
+    auto const node_pos = calc_node_pos(comm_cart);
+
     const auto lblattice = lb_lbfluid_get_lattice();
-    offset[0] = node_domain_position[0] * lblattice.grid[0];
-    offset[1] = node_domain_position[1] * lblattice.grid[1];
-    offset[2] = node_domain_position[2] * lblattice.grid[2];
+    offset[0] = node_pos[0] * lblattice.grid[0];
```
KaiSzuttor, line 232: we'd have users for https://en.wikipedia.org/wiki/Hadamard_product_(matrices)
all over the code base... how cool espresso would be with an implementation for ```Vector```

src/core/grid_based_algorithms/lb_particle_coupling.cpp
```diff
@@ -149,6 +149,9 @@ Utils::Vector3d lb_viscous_coupling(Particle *p,
 namespace {
 bool in_local_domain(Utils::Vector3d const &pos) {
   auto const lblattice = lb_lbfluid_get_lattice();
+  auto const my_left = local_geo.my_left();
```
KaiSzuttor, line 152: i think this whole function should go to the lattice class, I saw a very similar function before

src/core/virtual_sites/lb_inertialess_tracers.cpp
```diff
@@ -45,6 +45,21 @@ void GetIBMInterpolatedVelocity(const Utils::Vector3d &p, double *v,
 
 bool *isHaloCache = nullptr;
 
+namespace {
+bool in_local_domain(Utils::Vector3d const &pos) {
```
KaiSzuttor, line 49: here it is... if this would be in the lattice class... or we check the node-locality independent on the lattice?

KaiSzuttor (changes requested):
looks good except for the few comments

src/core/LocalBox.hpp
fweik: Because the sign bit is used to indicate the direction of the normal.

fweik: This can probably be helped later, but one step at a time...

src/core/grid.hpp
fweik: Because that is not available in all boost versions that we support. There is a price for everything...

src/core/grid_based_algorithms/lattice.cpp
fweik: Maybe it will go away soon anyway: the p3m contains similar logic, but with more features. If there is a quick way to reuse that I'll do it. That also includes halo comm, so the lb halo stuff (for the coupling) would also go away. Let's see.

src/core/virtual_sites/lb_inertialess_tracers.cpp
fweik: In it's current form it depends on lb globals, so there is some work to be done here. That was clearly out of scope.

fweik: So abstractly what is missing is a function that checks if a point is in a rectangle (and some more box functions, like volume)

fweik: [just geometry stuff](https://www.boost.org/doc/libs/1_70_0/libs/geometry/doc/html/geometry/reference/algorithms.html)

fweik:
@KaiSzuttor I added some tests. Obvious there is much more work to do here, but I still think it's an improvement.

KaiSzuttor (approved):
thanks for the tests

jngrad:
The picture `doc/doxygen/figs/directions.gif` is not used in the Doxygen blocks anymore. If it has become irrelevant, we can delete it.

fweik:
@jngrad the information is still valid, and I'm going to reuse it in another PR. Let's keep it for now.

fweik:
bors r=KaiSzuttor

PR #2976, opened 2019-07-08
WIP: Lees-Edwards boundary conditions

bindgens1:
Description of changes:
 - Reimplements LE boundary conditions

Depends on:
- #2957 
- #2543

This pull request is based on #2957 and the removal of the ghost shifts therein.


fweik:
Did you get it to work?

bindgens1:
> Did you get it to work?

Yes, we got it to work.
There was an abs missing in the distance check for get_mi_vector (fbafc81)

That's why it failed before with more than one node.

KaiSzuttor:
@fweik could you please merge the python branch?

fweik:
No, sorry. Also I think there is no point before the LB issue isn't addressed.

KaiSzuttor:
@RudolfWeeber wanted me to look at the LB issue and to estimate the effort we would have to fix it.

fweik:
Well I don't want to do it.

KaiSzuttor:
ok, but could you comment on what you think how much work this would be?

fweik:
What, the merge?

KaiSzuttor:
no, the LB fix

fweik:
I don't know how to fix it or if it can be done with the current coupling at all. So no, I can't estimate the effort. Please also see the [comment in the wiki](https://github.com/espressomd/espresso/wiki/Offline-Espresso-meeting) regarding the issue.

RudolfWeeber:
I merged the current Python branch.

I'm trying to separate the cached LeesEdwardsState (containing current positional offset and shear velocity) from the LeesEdwardsProtocol holding the parameters (such as frequency, amplitude, t_0).

While doing that, I noticed something odd about the times used to calculate the LE state.

Assuming, we start at t=t_0:
* The initial force calculation is run at (le_pos_offset(t=t_0), le_shear_velocity(t=t_0)).

Within the loop:
  * propagate_vel_pos() does not used the cached state. Instead it uses velocity(t=t_0), offset(t_0 +dt/2)
  * After propagate_vel_pos(), sim_time is incremented by a time step.
  * Before force_calc(), the cached velocity is updated to (t_0 +dt +0.5). 
  I think, this should be 0.5 * dt. 
  Also, do you actually want t_0 +dt +0.5*dt or should it be t_0 +0.5 dt?
  * propagate_pos_finalize_p_inst(), does not use cached values. Instead:
  velocity(t_0 +dt +dt). Position is not used.

@bindgens1, can you please recheck this?

bindgens1:
I just read the code again and added line number to make sure we are talking about the same:

> I merged the current Python branch.
> 
> I'm trying to separate the cached LeesEdwardsState (containing current positional offset and shear velocity) from the LeesEdwardsProtocol holding the parameters (such as frequency, amplitude, t_0).
> 
> While doing that, I noticed something odd about the times used to calculate the LE state.
> 
> Assuming, we start at t=t_0:
> 
> * The initial force calculation is run at (le_pos_offset(t=t_0), le_shear_velocity(t=t_0)).

We are saving the time when the LE protocol is switched on. I don't think that there should be a problem here.

> 
> Within the loop:
> 
> * propagate_vel_pos() does not used the cached state. Instead it uses velocity(t=t_0), offset(t_0 +dt/2)

In this function (line: 697...) velocity and position are propagated. Since we are unable to determine the exact point in time when a particle will cross the boundary, we assume it's a the mid of the time_step. 

> * After propagate_vel_pos(), sim_time is incremented by a time step.
> * Before force_calc(), the cached velocity is updated to (t_0 +dt +0.5).
>   I think, this should be 0.5 * dt.
>   Also, do you actually want t_0 +dt +0.5*dt or should it be t_0 +0.5 dt?

Are you referring to line 338? 
Yes, there is indeed something off. In case we take t_0 + 0.5 * dt as the reference for the DPD force calculation we have to subtract half a time step here. I think the point to propagate the time in the integration loop has changed in general. I remember seeing it further down. We want: t_0 +dt  - 0.5*dt

> * propagate_pos_finalize_p_inst(), does not use cached values. Instead:
>   velocity(t_0 +dt +dt). Position is not used.

This should be le_vel(sim_time) / 2. Same reasoning as above: The system time is propagated earlier and not at the end of the int loop.

> @bindgens1, can you please recheck this?

Does this solve the questions?

bindgens1:
I just checked: I think the propagation of sim_time was changed in commit 8e60c78fe7 in February. Since we did our last merge (before the one with Flo's changes to the ghost shifts) earlier we didn't see the changes and din't pay attention.

bindgens1:
@RudolfWeeber 

Should I update that part?

RudolfWeeber:
On Wed, Aug 07, 2019 at 06:51:21AM -0700, bindgens1 wrote:
> @RudolfWeeber 
> 
> Should I update that part?
Yes. Please do so.

RudolfWeeber:
I did the following:
* Split lees edwars protocol into two classes:
  * The protocol contians the input parameters such as shear rate and shear direction
  * The cache contains the current shear velocity and positional offset
* The cache is still needed, because within get_mi_vector and vel_diff calls it is not known, which simulation times to apply for velocities and positions
* The active Lees-Edwards protocol and cache are now stored in BoxGeometry. This way, the recently aquired unit-testability of get_mi_vector() is maintained.
* There are three Lees Edwards protocols: Off, LinearShear, OscillatoryShear. These are mirrored by corresponding classes in the script interface
```
from espressomd.lees_edwares import LinearShear
system.lees_edwards.protocol = Linearshear(initial_pos_offset=0.1, shear_velocity=1.2,shir_direction=1,shear_plane_normal=2)
print(system.lees_edwards.shear_velocity)
```
* @fweik, as we discussed, I used a variant rather than inheritance. This lead to quite a lot of additional code though. Shared pointers (or bare pointers) are still needed, to maintain the relationship between the parameters in the Python object and those in the core.
* The cahced positional offset and shear velocity are now updated, when system.sim_time is changed. Probably there are still cases, where the cache is not updated befor system.integrator.run(0). E.g., energy calculation. This will have to be addressed, when a solution for #3079 is available
* The test case is adapted and passes. Sebastian, can you please run one of your dpd fluids to make sure you get the same result as well. I'm not totally convinced, the current tests cover everything happening in the integration loop.

There is still some cleanup to do before merging, but this should no longer affect the Python interface.

What is the intended sign convention for system.velocity_difference()
Should it be the same as system.distance_vec()?

fweik:
@RudolfWeeber I don't see why you need reference semantics for the protocol. I'd make the
parameters constructor only, and set the protocol by value in the core. That is way clearer, you
don't need the awkward shared pointer (plus you don't have to explain to me what the exact ownership relation between the interface and the core is :-)), you can get away with a single
interface class and it is not nullable where it shouldn't be. Also you can get more terse code by using lambdas instead of named visitors. I also find this more readable because the visitor is defined at the point of usage. I didn't look at any other code.

bindgens1:
@RudolfWeeber 
@fweik 
I can run the diffusion tests again and compare them to the previous results.

Where did the protocol with the constant offset go? It was called step strain in the old version. It's useful to have a comparison to oscillatory tests since the difference between step strain and oscillatory tests is just an FT. If we can test this with a polymer + solvent it's a further hint for the correct implementation.

RudolfWeeber:
>Where did the protocol with the constant offset go? It was called step strain in the old version. It's useful to have a comparison to oscillatory tests since the >difference between step strain and oscillatory tests is just an FT.

LinearShear(initial_pos_offset=step_size,shear_velocity=0)

If you find it very important, the constant step can be reintroduced as a dedicated protocol. It doesn’t really add functionality, though. Maybe a hint in the docs will suffice.

 

With regards to the test, re-running and comparing one frequency with oscillatory shear will be enough, I think.

Just to be on the safe side.

 

@fweik, you’re right. For LE, it is probably sufficient that protocol changes can occur be re-assigning to lees_edwards.protocol. It’s a different matter for shapes and the like where there is some merit in reference symantics.

This will also remove the outdated-cache-problem, as a re-assignment to lees_edwards.protocol already forces a cache update.

 

I’d like to keep the per-protocol interface classes, as this avoids the 

If-this-parameter-is-that-than-the-following-other-parameters-are-needed 

Syntax.

fweik:
@RudolfWeeber  I''d use a constructor similar to
```c++
void construct(VaraiantMap const& params) {


if(

**Comment by fweik** [comment]
@RudolfWeeber  I''d use a constructor similar to
```c++
void construct(VaraiantMap const& params) {
  auto const &procotol = get<string>(params.at("protocol"));

  if(protocol == "linear") {
    set_protocol(make_from_params<Protocol::Linear>(params, "param1", "param2"));
  }  else if (protocol == "other_protocol") {
     ...
  } else {
     throw;
  }
}
```

which I think has less overhead than multiple overhead classes. But whatever works for you.

RudolfWeeber:
When is the image box reset needed?
This should probably be triggered by the user, otherwise a change of the LE protocol might have surprising condquences.
Is there a need for a core function, or does
system.part[:].pos =system.part[:].pos_folded
do the same thing?
Alternatively, we could make the image_box property writable, i.e.,
system.part[:].image_box =(0,0,0)

bindgens1:
@RudolfWeeber 
The box image reset is important to reconstruct the physical trajectories of the particles. We want to have a defined state so we can track changes in the image boxes as well as the background movement of the periodic boxes. 
An approach without the reset of the image boxes would be much more complicated (if possible at all).

The image reset of the boxes can already be initialised via the python interface. One can also think about a check similar to the Verlet lists. If the image boxes are not reseted a warning could be thrown.

bindgens1:
I corrected the velocity used for the force calculation. According to Groot & Warren 1997 (DPD, bridging the gap...) Eq. 9 half an offset is needed for the velocity. This coincides with their choice of lambda = 1/2.

bindgens1:
> What is the intended sign convention for system.velocity_difference()
> Should it be the same as system.distance_vec()?

It think this is the most reasonable approach to keep things consistent?

bindgens1:
I guess, we can close this now?

PR #2980, opened 2019-07-09
P3M local mesh and halo comm

fweik:
Description of changes:
 - Factor out p3m local mesh calculation and reuse in dipolar p3m
 - Factor out p3m halo communication and reuse in dipolar p3m


fweik:
I give up

PR #2878, opened 2019-05-31
Lbgpu node vel

fweik:
Description of changes:
 - Factored out velocity getter, and obey boundary velocity for
   both interpolation schemes,
- Removed some globals
- Leak less

PR #2982, opened 2019-07-09
Reduce excessive loop unrolling in lbgpu velocity interpolation

mkuron:
This caused excessive register usage, especially when combined with thrust. Issue discovered by @fweik in #2878.

It turns out that this is a problem for CUDA too, it just exhibits a different behavior. Instead of crashing like on HIP, CUDA just produces a large binary and slower code. In a perfect world, the compiler should display a warning, but I guess neither AMD nor Nvidia operate in a perfect world.

### DIFF
--- a/src/core/grid_based_algorithms/lbgpu_cuda.cu
+++ b/src/core/grid_based_algorithms/lbgpu_cuda.cu
@@ -1418,9 +1418,11 @@ velocity_interpolation(LB_nodes_gpu n_a, float *particle_position,
 
   int cnt = 0;
   float3 interpolated_u{0.0f, 0.0f, 0.0f};
-#pragma unroll
+#pragma unroll 1
   for (int i = 0; i < 3; ++i) {
+#pragma unroll 1
     for (int j = 0; j < 3; ++j) {
+#pragma unroll 3
       for (int k = 0; k < 3; ++k) {
         auto const x =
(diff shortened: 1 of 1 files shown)


src/core/grid_based_algorithms/lbgpu_cuda.cu
```diff
@@ -1418,9 +1418,11 @@ velocity_interpolation(LB_nodes_gpu n_a, float *particle_position,
 
   int cnt = 0;
   float3 interpolated_u{0.0f, 0.0f, 0.0f};
-#pragma unroll
+#pragma unroll 1
```
KaiSzuttor, line 1421: can't you just remove the ```pragma``` directive here?

> If the ICE evaluates to 1, the compiler will not unroll the loop

from https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#pragma-unroll

Or do you want to force the compiler not to unroll here?

mkuron: I wanted to force it to not unroll, just to be on the safe side. During my (rather limited) testing, I noticed that the compilers (nvcc, hipcc, clang) sometimes do some automatic unrolling, even on this long-bodied loop.

KaiSzuttor (approved):

PR #2984, opened 2019-07-10
Don't unroll loops with nontrivial loop bodies

mkuron:
Simple assignments and arithmetic are fine to unroll, but if there's a function call or a nested loop, unrolling is probably not useful

Follow-up to #2982.


KaiSzuttor (approved):

RudolfWeeber:
bors retry

PR #2986, opened 2019-07-10
Debian9/Intel18/Clang6/ROCm: switch Python2 to Python3

### LINKED ISSUE #2694: Prepare docker images for supported Distros with Python 3

jngrad:
Final PR for #2694

Images Debian9/Intel18/Clang6/ROCm now have a Python3 version (espressomd/docker#104)

Closes #2694
Closes #2951

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -235,7 +235,7 @@ if(WITH_CUDA)
   endif()
 endif(WITH_CUDA)
 
-find_package(PythonInterp REQUIRED)
+find_package(PythonInterp 3 REQUIRED)
 
 if(WITH_PYTHON)
   find_package(Cython 0.23 REQUIRED)
--- a/maintainer/CI/build_cmake.sh
+++ b/maintainer/CI/build_cmake.sh
@@ -94,7 +94,7 @@ set_default_value check_skip_long false
 set_default_value make_check_tutorials false
 set_default_value make_check_samples false
 set_default_value make_check_benchmarks false
-set_default_value python_version 2
+set_default_value python_version 3
 set_default_value with_cuda true
 set_default_value build_type "Debug"
 set_default_value with_ccache false
--- a/maintainer/CI/build_cmake.sh
+++ b/maintainer/CI/build_cmake.sh
@@ -94,7 +94,6 @@ set_default_value check_skip_long false
 set_default_value make_check_tutorials false
 set_default_value make_check_samples false
 set_default_value make_check_benchmarks false
-set_default_value python_version 3
 set_default_value with_cuda true
 set_default_value build_type "Debug"
 set_default_value with_ccache false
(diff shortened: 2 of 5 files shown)


maintainer/CI/build_cmake.sh
```diff
@@ -94,7 +94,7 @@ set_default_value check_skip_long false
 set_default_value make_check_tutorials false
 set_default_value make_check_samples false
 set_default_value make_check_benchmarks false
-set_default_value python_version 2
```
KaiSzuttor, line 97: does it make sense to have this variable at all after requiring version 3?

jngrad: At the moment we still allow compilation with Python2 bindings, we just offer no guarantee that it will work. Removing the ability to use Python2 means removing all `from __future__` imports, simplifying a few exceptions statements, removing `object` from classes, simplifying the CMake logic, removing the line you just flagged from GitLab CI and Travis CI files, etc.

CMakeLists.txt
```diff
@@ -235,7 +235,7 @@ if(WITH_CUDA)
   endif()
 endif(WITH_CUDA)
 
-find_package(PythonInterp REQUIRED)
+find_package(PythonInterp 3 REQUIRED)
```
KaiSzuttor, line 238: > At the moment we still allow compilation with Python2 bindings, we just offer no guarantee that it will work. Removing the ability to use Python2 means removing all from __future__ imports, simplifying a few exceptions statements, removing object from classes, simplifying the CMake logic, removing the line you just flagged from GitLab CI and Travis CI files, etc.

How does that work out with that line?

maintainer/CI/build_cmake.sh
mkuron: I think we can remove it from the CI scripts already. Users shouldn’t be running those anyway.

CMakeLists.txt
jngrad: This line calls a function that returns early instead of looking for a python interpreter when we pass `-DPYTHON_EXECUTABLE=$(which python2)` to cmake. The build script for CI always passes this variable.

maintainer/CI/build_cmake.sh
jngrad: I've removed the variable.

KaiSzuttor (approved):

CMakeLists.txt
KaiSzuttor: that's not true... the cmake script will fail if you call it with ```-DPYTHON_EXECUTABLE=$(which python2)```. That also makes sense since you want to restrict users to python3

jngrad: my bad, I was using the wrong branch to test the CMake behavior. No more Python2 with this line.

PR #2987, opened 2019-07-11
Fix lbb

### LINKED ISSUE #2983: Invalid code in lb_boundaries.cpp
To reproduce the compile error on `clang:6.0`:
```bash
cp ../maintainer/configs/maxset.hpp myconfig.hpp
cmake -DWITH_CUDA=OFF -DWITH_CLANG_TIDY=ON -DPYTHON_EXECUTABLE=$(which python2) ..
make
```
```
/local/espresso/src/core/grid_based_algorithms/lb_boundaries.cpp:308:5: error: do not use 'else' after 'return' [readability-else-after-return,-warnings-as-errors]
  } else {
    ^~~~~~
/local/espresso/src/core/grid_based_algorithms/lb_interface.cpp:741:14: error: Value stored to 'agrid' during its initialization is never read [clang-analyzer-deadcode.DeadStores,-warnings-as-errors]
  auto const agrid = lb_lbfluid_get_agrid();
             ^
2 warnings treated as error
```

This function is not valid and needs refactoring:
https://github.com/espressomd/espresso/blob/a2110311295c4c03495c601f57378589ac1866b0/src/core/grid_based_algorithms/lb_boundaries.cpp#L286

Issues:
* uneven number of closing braces
    + there should probably be one after line 316 to close the else
    + or the else statement should be inside the `#if defined(LB_BOUNDARIES)` block
* clang-tidy: `do not use 'else' after 'return' [readability-else-after-return]`
* <strike>the comment above the function is probably outdated</strike>, sorry, didn't realize it's for the argument
* the final return value should be `ES_OK` instead of 0

For the second error, `auto const agrid` could be moved inside the `#ifdef CUDA` block:
https://github.com/espressomd/espresso/blob/a2110311295c4c03495c601f57378589ac1866b0/src/core/grid_based_algorithms/lb_interface.cpp#L733

The fix will have to be manually tested since this combination of features is not covered in CI.

KaiSzuttor:
Fixes #2983


src/core/grid_based_algorithms/lb_boundaries.cpp
```diff
@@ -282,42 +282,32 @@ void lb_init_boundaries() {
   }
 }
 
-// TODO dirty hack. please someone get rid of void*
-int lbboundary_get_force(void *lbb, double *f) {
+Utils::Vector3d lbboundary_get_force(LBBoundary const *const lbb) {
```
fweik: On `const` less...

fweik (changes requested):

src/core/grid_based_algorithms/lb_boundaries.cpp
KaiSzuttor: why?

fweik:
Passing by constant value makes no sense

KaiSzuttor:
what am I missing? I want the pointer and the pointed to object to be const

fweik:
You are passing the pointer by value. Passing by const value is not good.

KaiSzuttor:
can you please give me ref.? I cannot see why this is not a good idea.

jngrad:
I'd also be interested in the reason. I'm currently refactoring the pressure/force/energy files and one minor part of the task consists in marking pointers as `const *const` whenever possible. This helps in functions creating pointers to bonded particles, where we don't want to accidentally re-assign a pointer passed as argument. I only know the rule that `const *const` should only appear in function implementations and `const*` in their declaration, to improve  readability and mask implementation details.

KaiSzuttor:
ping @fweik

fweik:
To me this is just line noise, but if you want to have it you can. The arguments that you have presented all seem to say the argument should be a reference and not a pointer...

fweik (approved):

PR #2988, opened 2019-07-11
Detailed balance exclusion radius

jonaslandsgesell:
This PR relaxes the conditions the exclusion radius has to meet. It now does not need to be small (i.e. comparable to the short ranged repulsive potential) anymore since now detailed balance is strictly obeyed during sampling.

### DIFF
--- a/samples/grand_canonical.py
+++ b/samples/grand_canonical.py
@@ -68,7 +68,7 @@ system.cell_system.max_num_cells = 2744
 # type 1 = A-
 # type 2 = H+
 
-for i in range(int(cs_bulk*box_l**3)):
+for i in range(int(cs_bulk * box_l**3)):
     system.part.add(pos=np.random.random(3) * system.box_l, type=1, q=-1)
     system.part.add(pos=np.random.random(3) * system.box_l, type=2, q=1)
 
--- a/samples/grand_canonical.py
+++ b/samples/grand_canonical.py
@@ -54,7 +54,7 @@ np.random.seed(seed=system.seed)
 system.time_step = 0.01
 system.cell_system.skin = 0.4
 temperature = 1.0
-system.thermostat.set_langevin(kT=temperature, gamma=1.0, seed=42)
+system.thermostat.set_langevin(kT=temperature, gamma=.5, seed=42)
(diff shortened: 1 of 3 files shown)


jonaslandsgesell:
@

src/core/reaction_ensemble.hpp
```diff
@@ -157,7 +157,7 @@ class ReactionAlgorithm {
                                                int particle_number_of_type,
                                                bool use_wang_landau);
 
-  bool particle_inserted_too_close_to_another_one;
+  bool particle_inside_exclusion_radius_was_touched;
```
KaiSzuttor: ```particle_inside_exclusion_radius_has_been_touched``` if you want a sentence as a variable

KaiSzuttor:
@thepith can you have a look?

samples/grand_canonical.py
```diff
@@ -86,7 +83,7 @@
             cutoff=lj_cut, shift="auto")
 
 RE = reaction_ensemble.ReactionEnsemble(
-    temperature=temperature, exclusion_radius=1.0, seed=3)
+    temperature=temperature, exclusion_radius=5.0, seed=3)
```
thepith: I do not think that setting the exclusion radius to 5 is a good thing in the samples. This might lead to people actually using this value.

KaiSzuttor: why is that not a good thing?

jonaslandsgesell: With the new changes it is allowed and produces correct physical results (which was previously not the case, back then the exlusion radius needed to fit the short ranged repulsive potential). However, increasing the exclusion radius reduces the acceptance rate... Therefore you will need more compute time to get uncorrelated data in the particle numbers of exchangeable species.

src/core/reaction_ensemble.hpp
KaiSzuttor: ping

samples/grand_canonical.py
thepith: Looks good now

KaiSzuttor (approved):

PR #2991, opened 2019-07-16
Reduce usage of local_cells

### LINKED ISSUE #2899: Reduce uses of the local_cells global
Many methods directly use the `local_cells` global, in most of the cases to iterate over the particles.
This exposes implementation details of the cell system to the methods, which isn't actually needed.
For those methods the `ParticleRange` to be iterated over should be passed as an argument.

KaiSzuttor:
Part of the solution to #2899

### DIFF
--- a/src/core/forces.hpp
+++ b/src/core/forces.hpp
@@ -30,6 +30,7 @@
  *  Implementation in forces.cpp.
  */
 
+#include "ParticleRange.hpp"
 #include "actor/Actor.hpp"
 #include "actor/ActorList.hpp"
 #include "bonded_interactions/bonded_interaction_data.hpp"
--- a/src/core/forces.hpp
+++ b/src/core/forces.hpp
@@ -45,10 +45,10 @@ extern ActorList forceActors;
 
 /** initialize real particle forces with thermostat forces and
     ghost particle forces with zero. */
-void init_forces();
+void init_forces(const ParticleRange &particles);
 
 /** Set forces of all ghosts to zero */
-void init_forces_ghosts();
+void init_forces_ghosts(const ParticleRange &particles);
 
 /** Calculate forces.
  *
@@ -60,10 +60,11 @@ void init_forces_ghosts();
  *  <li> Calculate long range interaction forces
  *  </ol>
  */
-void force_calc();
+void force_calc(const ParticleRange &particles);
 
 /** Check if forces are NAN */
-void check_forces();
+void check_forces(const ParticleRange &particles,
+                  const ParticleRange &ghost_particles);
 
 /** Calculate long range forces (P3M, MMM2d...). */
 void calc_long_range_forces();
--- a/src/core/integrate.cpp
+++ b/src/core/integrate.cpp
@@ -106,20 +106,20 @@ volatile std::sig_atomic_t ctrl_C = 0;
 /** Propagate the velocities. Integration step 1 of the Velocity Verlet
    integrator:<br>
     \f[ v(t+0.5 \Delta t) = v(t) + 0.5 \Delta t f(t)/m \f] */
-void propagate_vel();
+void propagate_vel(const ParticleRange &particles);
 /** Propagate the positions. Integration step 2 of the Velocity
    Verletintegrator:<br>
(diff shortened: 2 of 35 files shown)


fweik:
Did you consult with @reinaual, I think he was also looking into this.

KaiSzuttor:
no, I'm not yet into mind reading

fweik:
Yeah, better work organization/mind reading technique is on the agenda for when I'm back this fall. For now I'm afraid you have to talk to other people every now and then...

KaiSzuttor:
actually I think there is no magic behind writing a comment to an issue, also when we assign issues to somebody

src/core/forces.hpp
```diff
@@ -60,10 +61,11 @@ void init_forces_ghosts();
  *  <li> Calculate long range interaction forces
  *  </ol>
  */
-void force_calc();
+void force_calc(const ParticleRange &particles);
```
fweik, line 64: `force_calc` actually needs access to the cells and to the particle index. To be clear, it should take something like a `CellStructure` as a parameter, which hopefully at some point in the future will encapsulate everything.

src/core/integrate.cpp
```diff
@@ -234,11 +234,11 @@ void integrate_vv(int n_steps, int reuse_forces) {
       langevin_rng_counter_increment();
     }
 
-    force_calc();
+    force_calc(local_cells.particles());
 
     if (integ_switch != INTEG_METHOD_STEEPEST_DESCENT) {
 #ifdef ROTATION
-      convert_initial_torques();
+      convert_initial_torques(local_cells.particles());
```
fweik, line 241: Since the `ParticleRange` is used multiple times in this function, you should put it into local variables (after the resort), and reuse it. The range actually has state (because it caches its size which can not be calculated in constant time), so this has performance benefits.

src/core/rattle.cpp
```diff
@@ -49,16 +49,17 @@ int n_rigidbonds = 0;
 
 /** Calculates the corrections required for each of the particle coordinates
     according to the RATTLE algorithm. Invoked from \ref correct_pos_shake()*/
-void compute_pos_corr_vec(int *repeat_);
+void compute_pos_corr_vec(int *repeat_, const ParticleRange &particles);
```
fweik, line 52: Same comment as for `force_calc` applies, also the range should be reused withing thre function.

src/core/rattle.cpp
```diff
@@ -268,31 +272,33 @@ void apply_vel_corr() {
...
+  for (auto &p : particles)
     revert(p);
 
-  for (auto &p : ghost_cells.particles())
+  for (auto &p : ghost_particles)
     revert(p);
 }
 
-void correct_vel_shake() {
+void correct_vel_shake(const ParticleRange &particles,
```
fweik, line 289: Same comment as for `force_calc` applies, also the range should be reused withing thre function.

fweik (changes requested):
LGTM, except for a few cases where I think the interface is too narrow. In these cases
I think a `CellStructure` should be passed instead (or a `CellPList` where sufficient).

src/core/forces.hpp
KaiSzuttor: you mean, we should also extract the ```cell_structure``` global and the ```ghost_cells```?

src/core/integrate.cpp
KaiSzuttor: what do you mean with "after the resort"?

src/core/forces.hpp
fweik: You can either leave the globals, and refactor this latter, or put the `{local, global}_cells`, potentially as dummy methods that return the globals into `CellStructure` and pass that.
They should not passed next to each other, because you actually have to have the cells that belong to the cell structure that is passed, and the particles that belong to those cells.

src/core/integrate.cpp
fweik: This function function calls the particle resorting. This can invalidate the particle ranges, so after the resort the ranges have to be updated.

KaiSzuttor:
handed over to @reinaual

PR #2992, opened 2019-07-16
core: LB: add force density contribution to node velocity.

### LINKED ISSUE #2870: LB CPU: Node velocity and interpolated velocity inconsistancy
The node velocity getter in lb.cpp includes the +ext_force_density/2 term,
wheras the node_v() getter in lb_interpolation.cpp does not.
The test_a_ext_force_density() test in lb.py fails for lb cpu if an interpolated velocity is checked in additoin to the node velocity.

KaiSzuttor:
Fixes #2870


KaiSzuttor:
we should be careful here... this changes the force on a particle if a force density acts on the fluid

KaiSzuttor:
@RudolfWeeber please have a look

fweik (dismissed):

KaiSzuttor:
@RudolfWeeber what has to be clearified here?

KaiSzuttor:
https://journals.aps.org/pre/pdf/10.1103/PhysRevE.65.046308 Eq. 19

KaiSzuttor:
https://arxiv.org/pdf/1406.2211.pdf

fweik:
@KaiSzuttor looking at this again, couldn't you just include the complete force density (also for the CPU)?
That would be even more correct, and the infrastructure for this does already exist (because the tracers use it). Then this could be even a step towards the grand unified coupling scheme. If you don't want to do it, please open an issue, should be done at some point.

RudolfWeeber:
> <https://github.com/KaiSzuttor> @KaiSzuttor looking at this, couldn't you just include the complete force density (also for the CPU)?
>That would be even more correct, and the infrastructure for this does already exist (because the tracers use it). Then this could be even a step towards the >grand unified coupling scheme. If you don't want to do it, please open an issue, should be done at some point.

The Walberla implementation uses the full force,I.e., volume force + forces from the particle coupling.

fweik:
A think at the send we should get `Algorithm 3` from Ulf's paper that Kai posted. This supports all 3 possible couplings in the same scheme and is shown to be 2nd order accurate (which is better than what we have now).

src/core/forces.cpp
```diff
@@ -143,6 +143,8 @@ void force_calc(CellStructure &cell_structure) {
   // Must be done here. Forces need to be ghost-communicated
   immersed_boundaries.volume_conservation();
 
+  lb_reset_force_densities();
```
KaiSzuttor: @itischler I think I lost track of what we wanted to achieve... why would we want to reset the force densities here?

src/core/integrate.cpp
```diff
@@ -385,6 +385,7 @@ void integrate_vv(int n_steps, int reuse_forces) {
     MPI_Bcast(&nptiso.p_inst_av, 1, MPI_DOUBLE, 0, comm_cart);
   }
 #endif
+  lb_reset_force_densities();
```
KaiSzuttor: ... and here?

PR #2998, opened 2019-07-19
Benchmarks reliability

jngrad:
Since 4.0 a few things changed in the CMake workflow (either in espresso CMake files, or in newer CMake releases) that broke the benchmark functionality:
- changing myconfig.hpp doesn't always cause a full project recompilation (rare)
- ctest sometimes runs benchmarks in parallel

The benchmark bash scripts were also brittle. They were both refactored based on `build_cmake.sh`:
- trap mechanism with clear error message
- another trap with cleanup action
- simplified CMake logic
- ctest runs benchmarks in serial
- erase /build/src before each compilation

Other changes:
- don't print debug energy values in the CSV output
- some column names were changed
- the [wiki documentation](https://github.com/espressomd/espresso/wiki/Development#benchmarking) was updated accordingly
    - and it now discusses condor on ICP machines
- reduce risk of LJ simulation crashing with a longer energy minimization
- add a benchmark for bonded interactions
- print more info in stdout (config file being processed, energy during LJ minimization and simulation)

### DIFF
--- a/maintainer/benchmarks/CMakeLists.txt
+++ b/maintainer/benchmarks/CMakeLists.txt
@@ -59,15 +59,17 @@ function(PYTHON_BENCHMARK)
     endif()
     list(REMOVE_AT BENCHMARK_CONFIGURATIONS 0)
     foreach(nproc IN LISTS BENCHMARK_CONFIGURATIONS)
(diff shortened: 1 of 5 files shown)


RudolfWeeber (approved):
bors r+

PR #3003, opened 2019-07-24
Recalc cutoff on skin change

KonradBreitsprecher:
- Corrected wrong ifdef p3m for mmm2d in coulomb.cpp
- In event.cpp: For mmm2d, the cutoff depends on the skin (see `Coulomb::cutoff()`). 
A skin change so far did not respect this from what I saw.


fweik (changes requested):
I don't think any of this is actually needed, can you please check if it's sufficient to just set the curoff to a finit constant? I think this just in place so that the short-range loop is not skipped, so any finite value, or potentially `0.` will do. @RudolfWeeber do you remember? Also this needs a test to check the behavior.

fweik:
Maybe I am confusing this with ELC, but it still does not make sens: the maximal cutoff is used to calculate layer_h, so there is a circular dependency here. Also I fail to see how any of this is related to the skin. This is all very confused.

RudolfWeeber:
My understanding is that the mmm2d kernel is executed for all pairs in a layer. As Flo said, the cutoff just has to be a positive epsilon, otherwise the short range loop is skipped entirely.

I’d only change that, if it actually can be shown that it fixes something.

fweik:
What the method actually needs is cutoff infinity in x and y direction, and none in z direction (This would also force a compatible cell system, as this can only be supported by n2 and layered). This can not be expressed because we only consider a scalar cutoff. The hackish solution to this implemented is to set the cutoff to an arbitrary finite value, and have layered not check the value. So I think you should set it to `std::max(box_l[0], box_l[1])` to make the cutoff incompatible with dd. 
(What it really should be is `{inf, inf, INACTIVE}`).

KonradBreitsprecher:
Ok using a small eps for MMM2D in Coulomb:cutoff cured my MWE:
```
import espressomd
import numpy as np
from espressomd import electrostatic_extensions
box_l = 10.0
system = espressomd.System(box_l=[box_l]*3)
system.time_step = 0.01
system.cell_system.set_layered(n_layers=20, use_verlet_lists=False)
system.periodicity = [1, 1, 0]
system.part.add(pos=(1, 1, 1), q=1)
system.part.add(pos=(9, 9, 9), q=-1)
mmm2d = espressomd.electrostatics.MMM2D(prefactor=1.0, maxPWerror=1e-3)
system.actors.add(mmm2d)
system.cell_system.skin=0.4
```
Without the PR I got:
ERROR: layered: maximal interaction range 0.9 larger than layer height 0.5
fired by cells_on_geometry_change() which is now gone.

src/core/electrostatics_magnetostatics/coulomb.cpp
```diff
@@ -106,11 +106,11 @@ double cutoff(const Utils::Vector3d &box_l) {
   switch (coulomb.method) {
   case COULOMB_MMM1D:
     return std::numeric_limits<double>::infinity();
+  case COULOMB_MMM2D:
+    return 1e-40;
```
fweik: Please don't use a magic number, this should be `std::numeric_limits<double>::min()`

fweik (changes requested):

RudolfWeeber (approved):

RudolfWeeber (approved):
bors r+

fweik:
There is still no test.

RudolfWeeber:
On Tue, Aug 06, 2019 at 01:08:42AM -0700, Florian Weik wrote:
> There is still no test.
Sry. Overlooked that. @konradbreitsprecher?

RudolfWeeber:
@KonradBreitsprecher, could you please provide a test, then this can go into 4.1.

KonradBreitsprecher:
I'm not getting the error anymore for the python branch (maximal interaction range larger than layer height), sth else has fixed that. Not sure what to test, for now I added a test that checks if mm2d changes the max_cut from zero to nonzero and does a test integration.

RudolfWeeber:
Do I understand correctly that this is no longer needed?

fweik:
Aye

PR #3008, opened 2019-07-25
Removed from __future__

### LINKED ISSUE #3004: Remove from __future__ in Python code
since we require Py3 now.

jrfinkbeiner:
Fixes #3004


jngrad (approved):

mkuron:
There is some more Py2-specific code, mainly the use of the six module, the unicode C string wrapper and the Cython 2/3 flags in CMake. That should be removed too.

PR #3013, opened 2019-07-25
python: make checkpointing of LB thermostat possible again.

### LINKED ISSUE #2652: LB thermostat cannot be checkpointed
```set_lb``` requires a LBFluid instance which is not present at the time the state for the thermostat is set

KaiSzuttor:
Fixes #2652

Description of changes:
 - store the LB fluid instance in the state of the LB thermostat
 - test LB thermostat checkpointing


RudolfWeeber (approved):
bors r+

PR #3014, opened 2019-07-25
Unequal time step checks and test

### LINKED ISSUE #2989: Unequal MD and LB Timesteps: Forbid or understand and test

christophlohrmann:
Fixes #2989 

Description of changes:
 - added checks that LB tau > MD time_step
 - added checks that LB tau is integer multiple of MD time_step
 - added test for LB-particle-coupling with different time steps

PR Checklist
------------
 - [x] Tests?
   - [x] Interface
   - [ ] Core 
 - [ ] Docs?

### DIFF
--- a/src/core/communication.cpp
+++ b/src/core/communication.cpp
@@ -483,6 +483,22 @@ void mpi_set_time_step_slave(double dt) {
 REGISTER_CALLBACK(mpi_set_time_step_slave)
 
 void mpi_set_time_step(double time_s) {
+  if (time_s <= 0.)
+    throw std::invalid_argument("time_step must be > 0.");
+  if (lb_lbfluid_get_lattice_switch() !=  ActiveLB::NONE){
+    auto eps = std::numeric_limits<float>::epsilon();
+    auto tau = lb_lbfluid_get_tau();
+    if ((tau - time_s)/(tau + time_s) < -eps)
+      throw std::invalid_argument("MD time_step ("+ std::to_string(time_s) +
+                                   ") must be =< LB tau ("+ std::to_string(tau)
+                                   + ")");
+    auto factor = tau / time_s;
+    if (fabs(round(factor)-factor)/factor > eps)
+      throw std::invalid_argument("LB tau ("+ std::to_string(tau) +
+                                  ") must be integer multiple of "
+                                  "MD time_step ("+ std::to_string(time_s)
+                                  + "). Factor is " + std::to_string(factor));
+  }
   mpi_call_all(mpi_set_time_step_slave, time_s);
 }
 
--- a/src/core/communication.cpp
+++ b/src/core/communication.cpp
@@ -485,19 +485,20 @@ REGISTER_CALLBACK(mpi_set_time_step_slave)
 void mpi_set_time_step(double time_s) {
   if (time_s <= 0.)
     throw std::invalid_argument("time_step must be > 0.");
-  if (lb_lbfluid_get_lattice_switch() !=  ActiveLB::NONE){
+  if (lb_lbfluid_get_lattice_switch() != ActiveLB::NONE) {
     auto eps = std::numeric_limits<float>::epsilon();
     auto tau = lb_lbfluid_get_tau();
-    if ((tau - time_s)/(tau + time_s) < -eps)
-      throw std::invalid_argument("MD time_step ("+ std::to_string(time_s) +
-                                   ") must be =< LB tau ("+ std::to_string(tau)
-                                   + ")");
+    if ((tau - time_s) / (tau + time_s) < -eps)
+      throw std::invalid_argument("MD time_step (" + std::to_string(time_s) +
+                                  ") must be =< LB tau (" +
+                                  std::to_string(tau) + ")");
     auto factor = tau / time_s;
-    if (fabs(round(factor)-factor)/factor > eps)
-      throw std::invalid_argument("LB tau ("+ std::to_string(tau) +
+    if (fabs(round(factor) - factor) / factor > eps)
+      throw std::invalid_argument("LB tau (" + std::to_string(tau) +
                                   ") must be integer multiple of "
-                                  "MD time_step ("+ std::to_string(time_s)
-                                  + "). Factor is " + std::to_string(factor));
+                                  "MD time_step (" +
+                                  std::to_string(time_s) + "). Factor is " +
+                                  std::to_string(factor));
   }
   mpi_call_all(mpi_set_time_step_slave, time_s);
 }
--- a/src/core/communication.cpp
+++ b/src/core/communication.cpp
@@ -485,21 +485,8 @@ REGISTER_CALLBACK(mpi_set_time_step_slave)
 void mpi_set_time_step(double time_s) {
   if (time_s <= 0.)
     throw std::invalid_argument("time_step must be > 0.");
-  if (lb_lbfluid_get_lattice_switch() != ActiveLB::NONE) {
-    auto eps = std::numeric_limits<float>::epsilon();
-    auto tau = lb_lbfluid_get_tau();
-    if ((tau - time_s) / (tau + time_s) < -eps)
-      throw std::invalid_argument("MD time_step (" + std::to_string(time_s) +
-                                  ") must be =< LB tau (" +
-                                  std::to_string(tau) + ")");
-    auto factor = tau / time_s;
-    if (fabs(round(factor) - factor) / factor > eps)
-      throw std::invalid_argument("LB tau (" + std::to_string(tau) +
-                                  ") must be integer multiple of "
-                                  "MD time_step (" +
-                                  std::to_string(time_s) + "). Factor is " +
-                                  std::to_string(factor));
-  }
+  if (lb_lbfluid_get_lattice_switch() != ActiveLB::NONE)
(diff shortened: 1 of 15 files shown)


src/python/espressomd/lb.pyx
```diff
@@ -30,7 +30,8 @@ from . import cuda_init
 from copy import deepcopy
 from . import utils
 from .utils import array_locked, is_valid_type
-from .utils cimport make_array_locked
+from .utils cimport make_array_locked, numeric_limits
+from globals cimport time_step as global_time_step
```
KaiSzuttor: better call it ```md_time_step```

src/python/espressomd/lb.pyx
```diff
@@ -70,10 +71,21 @@ cdef class HydrodynamicInteraction(Actor):
...
+        if (tau <= 0.0):
+            raise Exception("LB_FLUID tau not set")
+        elif global_time_step > 0:
+            if (tau - global_time_step) / (abs(tau) + abs(global_time_step)) < -numeric_limits[float].epsilon():
+                raise ValueError(
+                    "LB_time_step ({}) must be >= Time Step ({})".format(tau, global_time_step))
+            factor = tau / global_time_step
+            if abs(round(factor) - factor) / factor > numeric_limits[float].epsilon():
+                raise ValueError(
+                    "LB_time_step must be integer multiple of Time step. Factor is {}".format(factor))
```
KaiSzuttor: the LB time step is called ```tau```, maybe also print ```MD time step```

testsuite/python/lb.py
```diff
@@ -326,6 +326,57 @@ def test_a_ext_force_density(self):
         for n in self.lbf.nodes():
             np.testing.assert_allclose(
                 np.copy(n.velocity), fluid_velocity, atol=1E-6)
+    
+    @utx.skipIfMissingFeatures("EXTERNAL_FORCES")
+    def test_unequal_time_step(self):
```
KaiSzuttor, line 331: please add a short docstring to the test

src/python/espressomd/lb.pyx
```diff
@@ -70,10 +71,21 @@ cdef class HydrodynamicInteraction(Actor):
...
-                raise ValueError("Density must be one positive double")
-
+        elif not (self._params["dens"] > 0.0 and (is_valid_type(self._params["dens"], float) or is_valid_type(self._params["dens"], int))):
+            raise ValueError("Density must be one positive double")
+        
+        tau = self._params["tau"]
+        if (tau <= 0.0):
+            raise Exception("LB_FLUID tau not set")
+        elif global_time_step > 0:
+            if (tau - global_time_step) / (abs(tau) + abs(global_time_step)) < -numeric_limits[float].epsilon():
```
KaiSzuttor: it would probably be cleaner to check that in the core, *e.g.* in ```lb_interface.cpp``` in ```lb_lbfluid_on_lb_params_change```

KaiSzuttor (changes requested):

src/python/espressomd/lb.pyx
christophlohrmann: do you want the check of the `time_step` in the core, too? (`mpi_set_time_step()` in `communication.cpp`) ?

KaiSzuttor: yes

RudolfWeeber (approved):
bors r+

src/core/communication.cpp
```diff
@@ -483,6 +483,22 @@ void mpi_set_time_step_slave(double dt) {
 REGISTER_CALLBACK(mpi_set_time_step_slave)
 
 void mpi_set_time_step(double time_s) {
+  if (time_s <= 0.)
+    throw std::invalid_argument("time_step must be > 0.");
+  if (lb_lbfluid_get_lattice_switch() !=  ActiveLB::NONE){
```
KaiSzuttor: could you please move this to a free function and correctly ifdef for the LB feature.

src/core/communication.cpp
```diff
@@ -483,6 +483,22 @@ void mpi_set_time_step_slave(double dt) {
...
+    auto eps = std::numeric_limits<float>::epsilon();
```
KaiSzuttor: ```auto const eps = std::numeric_limits<double>::epsilon();```

src/core/grid_based_algorithms/lb_interface.cpp
```diff
@@ -481,8 +481,22 @@ const Utils::Vector3d lb_lbfluid_get_ext_force_density() {
 }
 
 void lb_lbfluid_set_tau(double tau) {
-  if (tau <= 0)
-    throw std::invalid_argument("tau has to be positive.");
+  if (tau <= 0.)
+    throw std::invalid_argument("LB tau has to be positive.");
+  extern double time_step;
```
KaiSzuttor: you can then use the free function here for the same validation to reduce code duplication

KaiSzuttor (changes requested):
sorry, please use an independent function, then its good

src/core/communication.cpp
jngrad: there is no `LB` macro, could you precise what you meant?

KaiSzuttor: Ah right, I forgot

RudolfWeeber:
>  void mpi_set_time_step(double time_s) {
This callback will no longer be present with Walberla. The LB Timestep is a constructor argument of the LB Fluid and cannot be changed safely after the fact in any case.

Can the check be called somewhere else, like in the lb sanity checks in lb_interface.cpp?

KaiSzuttor:
why can't it be changed safely afterwards? the LB only depends on ```tau``` in a very few lines of code

mkuron:
> This callback will no longer be present with Walberla. The LB Timestep is a constructor argument of the LB Fluid and cannot be changed safely after the fact in any case.

The Lb timestep only changes the unit conversion going into (e.g. viscosity) and out of Walberla (e.g. velocity getter). Walberla doesn‘t need to know about our unit system.

christophlohrmann:
On a side note: Is there a reason that
https://github.com/espressomd/espresso/blob/62e360e3b4866be8ef0f43af5fb2616c6e2f7444/src/core/grid_based_algorithms/lb_interface.cpp#L504
is called with that argument when setting tau?

RudolfWeeber:
On Fri, Jul 26, 2019 at 06:14:07AM -0700, Kai Szuttor wrote:
> why can't it be changed safely afterwards? the LB only depends on ```tau``` in a very few lines of code
I should have added "safely". It's clear to me that it could be implemented. I'd have to be convinced of the continued validity of any existing flwo field, double buffered force fields, boundary forces, etc., though.
The meaning of velocities in lb units would suddenly change. 
Or one would one of those rescale-on-time-step-chane magics that we removed from Espresso.

I think, the only safe thing to do is re-instance a clean LB fluid with the new time step.

The existing LBs only call the time step broadcast once at construction. It is also not possible to change it later.

In any case, at construction of an LB Walberla fluid, the MPI broadcast routine is not called. The LbWalberla interface class cannot be instanced without a time step, because then, you'd have an LB Fluid for which the unit conversions in the particele coupling and other places are undefined.

Please move the checks to the lb_sanity_checks(). They should probably be called at the end of an lb instanceing from Python anyway.

KaiSzuttor:
maybe you should do a grep on tau, it's really not that bad

RudolfWeeber:
>maybe you should do a grep on tau, it's really not that bad
I did that. The issue is not the number of taus appearing in the code but the re-scaling that would be needed to keep the meaning (in MD units) of the flow field and derived observables invariant under LB time step changes.

 

Unless someone explains why the ability to change LB time steps in the middle of a simulation is important and provides the re-scaling + tests for the existing LBs, I’d rather not work on it for Walberla.

This is, however, orthogonal to this PR. The time step consistency check needs to be in the sanity checks in any case. Otherwise, you can change the MD time step after setting up the LB.

src/core/grid_based_algorithms/lbgpu.cpp
```diff
@@ -32,6 +32,7 @@
 #include "debug.hpp"
 #include "global.hpp"
 #include "grid.hpp"
+#include "grid_based_algorithms/lb.hpp"
```
KaiSzuttor: you are now introducing a dependency of the GPU LB on the CPU LB... maybe move the check to ```lb_interface.hpp```

KaiSzuttor:
thanks, looks good to me except for the comment above

KaiSzuttor (approved):

jngrad:
bors r= KaiSzuttor

jngrad:
@KaiSzuttor something went wrong with bors, the merge did not go as expected on the `staging.tmp` branch, so nothing was pushed to the `staging` branch and CI could not pull the commit. Here are the last 3 commits on  `staging.tmp`:
```
[ci skip][skip ci][skip netlify] -bors-staging-tmp-3014
[ci skip][skip ci][skip netlify]
Merge #3026
```
The `Merge #3026` commit was successful and pushed to `staging`, but then we have two unusual `[ci skip]` commits for this PR, one of which is not a merge.

jngrad:
bors retry

jngrad:
bors r= KaiSzuttor

PR #3024, opened 2019-07-26
Build and deploy tutorials in CI

### LINKED ISSUE #1487: Build and deploy tutorials in CI
The job on gitlab ci that builds and deploys sphinx and doxygen should also do the same with the tutorial pdf's.

jngrad:
Fixes #1487

Description of changes:
- introduce a CUDA 10.1 job
- simplify the export statements for notebooks with a single `nb_export` macro:
    - conversion to Python scripts is mandatory for tutorial tests
    - conversion to HTML (with or without running cells) is mandatory for the website
    - having the `html_export`/`python_export` macro pair caused code duplication in CMake
- CI/CD:
    - compile and deploy LaTeX tutorials as PDF to the website
    - run and deploy IPython tutorials as HTML to the website
    - don't deploy IPython tutorials as Python to the website
    - url: http://espressomd.org/wordpress/documentation/
- cleanup of LaTeX tutorials:
    - remove regenerable files (i.e. .pdf and .toc files) since they were always outdated and are now generated by CI/CD
    - remove the LaTeX template subfolder, since we decidedto write future tutorials in Jupyter
- the steps required to create a new Jupyter tutorial are now [explained in the wiki](https://github.com/espressomd/espresso/wiki/Documentation)


jngrad:
The notebooks were successfully synchronized with the website. The job took 20 min. The [error message in tutorial 08](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/141641) is expected and totally fine.

In our discussions during the coding day, @RudolfWeeber suggested we put this job in manual and @KaiSzuttor  that we create a dedicated weekly schedule.

KaiSzuttor:
BTW: could we add a check that there exists a link to the tutorial on the respective espressomd.org webpage? *E.g.* like this:
``` python
import re
import requests
from bs4 import BeautifulSoup

r = requests.get('http://espressomd.org/wordpress/documentation/')
soup = BeautifulSoup(r.content, 'html.parser')
tutorial_link = soup.find_all('a', href=re.compile('^http://espressomd.org/html/tutorials_html/11-'))
if not tutorial_link:
    print("Link to tutorial not found.")
    exit(1)                                                                                          
```

.gitlab-ci.yml
```diff
@@ -484,6 +524,19 @@ deploy_doxygen_documentation:
     - cd ${CI_PROJECT_DIR}/build/doc/doxygen/html &&
       rsync -avz --delete -e "ssh -i ${HOME}/.ssh/espresso_rsa" ./ espresso@elk.icp.uni-stuttgart.de:/home/espresso/public_html/html/dox
 
+deploy_tutorials:
+  extends: .deploy_base
+  dependencies:
+    - run_tutorials
+  script:
+    - cd ${CI_PROJECT_DIR}/build/doc/tutorials
+    - 'grep -Poe " src=(.)figures/.+?\1" */*.html > deploy_list.txt'
```
KaiSzuttor: what is happening here? The usage of regex should be documented somehow...

jngrad: Yes, that's quite inelegant. I'll write a proper Python script without regex for this job.

jngrad:
I thought about warnings for missing links on the website. But then everytime we create a new tutorial, we need to either write a broken link to the website before running CI/CD, or manually run the tutorial locally + filter the images in bash + scp the  files to elk with the correct ssh key + prevent anyone else from running CI/CD during that time otherwise the newly scp'ed tutorial will be erased. Removing or renaming a tutorial will cause the same headache, with extra if/else code in the check script. Not sure if a missing link check is worth adding a new layer of coupling between CI/CD and the website, for which only I and Ingo have credentials.

.gitlab-ci.yml
jngrad: Fixed by 638377d.

jngrad:
@KaiSzuttor @mkuron We now have 2 schedules:
- the nightly builds now run only 6 days a week
- the weekly builds run on day 7, same as nightly builds but also deploys the tutorials (34879e6)

doc/tutorials/html_runner.py
```diff
@@ -0,0 +1,104 @@
...
+        cells_code = [code]
+    # create new cells
+    cell_md = nbformat.v4.new_markdown_cell(source='Solution from ' + filepath)
+    nb['cells'].append(cell_md)
+    for cell_code in cells_code:
+        cell_code = nbformat.v4.new_code_cell(source=cell_code.strip())
+        nb['cells'].append(cell_code)
+
+# substitute global variables and disable OpenGL/Mayavi GUI
+unique_string = '\n##h7OZVDMsDxUPhqoVy2Mh0BTcoaFVbMUqEYCHYCDH\n'
```
KaiSzuttor: what is that string for?

jngrad: A comment line added between cells, helps splitting the source code back into notebook cells after variable substitution. It would be cleaner to use the uuid module instead of a hard-coded line.

jngrad: Fixed in 010d470

KaiSzuttor (approved):

PR #3025, opened 2019-07-28
Make LB coupling work with any skin

fweik:
Description of changes:
 - Particle coupling forces are now calculated by the
   node in whose LB domain the particle position is,
   and not by the node that owns the particle as before.
 - This now works independent of skin.
 - Also fixes the wrong sanity check in the LB, before you
   could use it with an incompatible skin if verlet lists were
   disabled, which would then crash.


src/core/grid_based_algorithms/lb_particle_coupling.cpp
```diff
@@ -147,17 +145,58 @@ Utils::Vector3d lb_viscous_coupling(Particle *p,
...
+/**
+ * @brief Check if a position is within the local LB domain
+ *       plus halo.
+ *
+ * @param pos Position to check
+ *
+ * @return True iff the point is inside of the domain.
+ */
+bool in_local_halo(Vector3d const &pos) {
+  auto const halo = 0.5 * lb_lbfluid_get_lattice().agrid;
```
KaiSzuttor: ```lb_lbfluid_get_agrid()```

src/core/grid_based_algorithms/lb_particle_coupling.cpp
```diff
@@ -180,9 +219,12 @@ void add_swimmer_force(Particle &p) {
   }
 }
 #endif
+
 } // namespace
 
-void lb_lbcoupling_calc_particle_lattice_ia(bool couple_virtual) {
+void lb_lbcoupling_calc_particle_lattice_ia(
+    bool couple_virtual, const ParticleRange &local_particles,
+    const ParticleRange &more_particles) {
```
KaiSzuttor, line 237: what are ```more_particles``` supposed to be? Please add a docstring

KaiSzuttor (changes requested):
Any idea why the swimmer test fails? We really should get rid of those data driven tests, they don't give any insight.

fweik:
I didn't even notice because on my machine it's skipped because vtk is missing.

fweik:
I don't get it. I also don't understand why there is a ghost communication for ENGINE right before the coupling. Maybe I don't understand the algorithm.

RudolfWeeber:
The halo region might have to include the swimmer->second-force-center distance.
I'm also not sure about parallelizaiton. As the secnd force center might be on a different MPI rank than the corresponding particle, the criterion for coupling a particle might have to be adapted.
@hmenke?

fweik:
The `force centers` only act on the fluid. They did couple for all particles in the halo region before, and should now. That is not the issue I think. This is some sort comm synchronization issue between ghosts and real particles, but I don't see it. I'll not further look into this, maybe we can revisit if new information comes up.

As as side note, all of this could be avoided if the force centers were just implemented with virtual sites.

RudolfWeeber:
>As as side note, all of this could be avoided if the force centers were just implemented with virtual sites.

We may have to do this for Walberla in any case. That would probably also make testing this easier.

fweik:
Yes, the reason for this PR is that I saw what you were doing for walberla and thought this might help you.

RudolfWeeber:
I took another look. I have no idea how to proceed until we have a test for a singe integration step with an engine particle.

It is also not clear why the engine gpu and engine cpu tests use different reference data. If I run the engine gpu with the cpu reference data, the error is even bigger than the change cause dby this PR.

RudolfWeeber:
@christophlohrmann, based on your observations, should we merge this PR?

christophlohrmann:
I can't speak about other things, but this PR fixes the issue of the missing second force centers when crossing periodic boundaries, so it's definitely better than the current state

src/core/grid_based_algorithms/lb_particle_coupling.cpp
fweik: Well propose a better name. I'd prefer just calling the function twice for each range, but that is not possible because of the GPU stuff.

RudolfWeeber:
The fix even goes so far, that the engine_lb on the cpu can use the same reference data as on the gpu.
@hmenke, is this supposed to work on the cpu in parallel. It is not clear to me, why it should, but the test passes with 4 and 8 cores.

RudolfWeeber (approved):

fweik:
Originally there was no parallel support. I allowed it at some point because the tests did pass and it should have worked. I did not check the tests though.

RudolfWeeber:
@KaiSzuttor

src/core/grid_based_algorithms/lb_particle_coupling.cpp
KaiSzuttor: okay different question: do you think that a dev in 3 years still knows what```more_particles``` are supposed to be without looking into the source code?

RudolfWeeber:
@KaiSzuttor, Flo answered the question about more_particles on Aug 10. I suggest to merge as is.

KaiSzuttor (approved):

PR #3029, opened 2019-07-30
Cleanup Sphinx documentation

jngrad:
Clean up Sphinx documentation of files affected by #3026:
- apply numpydoc style (indentation level of param descriptions, show allowed param values)
- remove duplicated docstrings, add missing words
- use LaTeX math to display equations of constraints
- fix broken Sphinx syntax (typos, unparsed `.. notes::` directives, missing links, etc.).

### DIFF
--- a/src/python/espressomd/collision_detection.pyx
+++ b/src/python/espressomd/collision_detection.pyx
@@ -40,7 +40,7 @@ class CollisionDetection(ScriptInterfaceHelper):
     def validate(self):
         """Validates the parameters of the collision detection.
 
-           This is called automatically on parameter change
+        This is called automatically on parameter change
 
         """
         return self.call_method("validate")
@@ -60,39 +60,42 @@ class CollisionDetection(ScriptInterfaceHelper):
 
         Parameters
         ----------
-        mode : One of "off", "bind_centers", "bind_at_point_of_collision", "bind_three_particles", "glue_to_surface"
-               Collision detection mode
+        mode : str, \{"off", "bind_centers", "bind_at_point_of_collision", "bind_three_particles", "glue_to_surface"\}
+            Collision detection mode
 
         distance : :obj:`float`
-               Distance below which a pair of particles is considered in the collision detection
+            Distance below which a pair of particles is considered in the
+            collision detection
 
-        bond_centers : Instance of :class:`espressomd.interactions.BondedInteraction`
-               Bond to add between the colliding particles
+        bond_centers : :obj:`espressomd.interactions.BondedInteraction`
+            Bond to add between the colliding particles
 
-        bond_vs :  Instance of :class:`espressomd.interactions.BondedInteraction`
-               Bond to add between virtual sites (for modes using virtual sites)
+        bond_vs : :obj:`espressomd.interactions.BondedInteraction`
+            Bond to add between virtual sites (for modes using virtual sites)
 
         part_type_vs : :obj:`int`
-               Particle type of the virtual sites being created on collision (virtual sites based modes)
+            Particle type of the virtual sites being created on collision
+            (virtual sites based modes)
 
         part_type_to_be_glued : :obj:`int`
-               particle type for "glue_to_surface" mode. See user guide.
+            particle type for ``"glue_to_surface"`` mode. See user guide.
 
         part_type_to_attach_vs_to : :obj:`int`
-               particle type for "glue_to_surface" mode. See user guide.
+            particle type for ``"glue_to_surface"`` mode. See user guide.
 
         part_type_after_glueing : :obj:`int`
-               particle type for "glue_to_surface" mode. See user guide.
+            particle type for ``"glue_to_surface"`` mode. See user guide.
 
         distance_glued_particle_to_vs : :obj:`float`
-               Distance for "glue_to_surface" mode. See user guide.
+            Distance for ``"glue_to_surface"`` mode. See user guide.
 
-        bond_three_particles : Instance of :class:`espressomd.interactions.BondedInteraction`
-               First angular bond for the "bind_three_particles" mode. See user guide
+        bond_three_particles : :obj:`espressomd.interactions.BondedInteraction`
+            First angular bond for the ``"bind_three_particles"`` mode. See
+            user guide
 
         three_particle_binding_angle_resolution : :obj:`int`
-              Resolution for the angular bonds (mode "bind_three_particles").
-              Resolution+1 bonds are needed to accommodate the case of a 180 degrees
+            Resolution for the angular bonds (mode ``"bind_three_particles"``).
+            Resolution+1 bonds are needed to accommodate the case of a 180 degrees
 
         """
 
@@ -102,8 +105,9 @@ class CollisionDetection(ScriptInterfaceHelper):
 
         # Completeness of parameter set
         if not (set(kwargs.keys()) == set(self._params_for_mode(kwargs["mode"]))):
-            raise Exception("Parameter set does not match mode. ", kwargs[
-                            "mode"], "requires ", self._params_for_mode(kwargs["mode"]))
+            raise Exception("Parameter set does not match mode. ",
+                            kwargs["mode"], "requires ",
+                            self._params_for_mode(kwargs["mode"]))
 
         # Mode
         kwargs["mode"] = self._int_mode[kwargs["mode"]]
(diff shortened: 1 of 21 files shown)


src/python/espressomd/collision_detection.pyx
```diff
@@ -60,39 +60,42 @@ class CollisionDetection(ScriptInterfaceHelper):
...
-               First angular bond for the "bind_three_particles" mode. See user guide
+        bond_three_particles : :obj:`espressomd.interactions.BondedInteraction`
+            First angular bond for the ``"bind_three_particles"`` mode. See
+            user guide
 
         three_particle_binding_angle_resolution : :obj:`int`
-              Resolution for the angular bonds (mode "bind_three_particles").
-              Resolution+1 bonds are needed to accommodate the case of a 180 degrees
+            Resolution for the angular bonds (mode ``"bind_three_particles"``).
+            Resolution+1 bonds are needed to accommodate the case of a 180 degrees
```
pkreissl: [...] the case of ~a~ 180 degrees [...]
or add missing words ([...] of a 180 degrees turn?)

src/python/espressomd/constraints.py
```diff
@@ -290,8 +289,7 @@ def field_from_fn(cls, box_size, grid_spacing, f, codim=None):
 
     @classmethod
     def field_coordinates(cls, box_size, grid_spacing):
-        """Returns an array of the coordinates of  the grid
-        points required.
+        """Returns an array of the coordinates of the grid points required.
```
pkreissl, line 292: Why is this not 

"""
Returns an array of the coordinates of the grid points required.

src/python/espressomd/interactions.pyx
```diff
@@ -581,19 +581,19 @@ IF LJCOS:
             return(self._params["epsilon"] > 0)
 
         def set_params(self, **kwargs):
-            """ Set parameters for the Lennard-Jones Cosine2 interaction.
```
pkreissl, line 584: Again, add line break after """ ?

src/python/espressomd/pair_criteria.py
```diff
@@ -22,7 +22,11 @@ class _PairCriterion(ScriptInterfaceHelper):
 
     def decide(self, p1, p2):
         """Makes a decision based on the two particles specified.
```
pkreissl, line 24: And again..

src/python/espressomd/particle_data.pyx
```diff
@@ -1041,16 +1042,17 @@ cdef class ParticleHandle:
...
-            The content of the torque and omega variables are meaningless, for the co-ordinates for which
+            Switches the particle's rotational degrees of freedom in the
+            Cartesian axes in the body-fixed frame. The content of the torque
+            and omega variables are meaningless for the co-ordinates for which
             rotation is disabled.
 
             The default is not to integrate any rotational degrees of freedom.
 
-            rotation : (int,int,int)
+            rotation : list of :obj:`int`
```
pkreissl, line 1052: Should the dimension of the list of `int`s be indicated?

src/python/espressomd/particle_data.pyx
```diff
@@ -2041,25 +2041,22 @@ Set quat and scalar dipole moment (dipm) instead.")
...
-            a keyword arguments in which the keys are names of particle properties
-            and the values are the values to filter for. E.g.,::
+        Parameters:
 
-                type=0,q=1
+        Either: a keyword arguments in which the keys are names of particle
+        properties and the values are the values to filter for. E.g.,::
 
-        Or:
+            type=0,q=1
```
pkreissl: missing space

src/python/espressomd/constraints.py
jngrad: Writing the brief on the same line as the triple quotes is not forbidden by the numpydoc style guide: https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard

pkreissl: No, sure, was just wondering as you changed that in other places.

src/python/espressomd/particle_data.pyx
jngrad: The dimension is almost never written for arrays, in most places I saw simply ```array_like :obj:`int` ```. The numpydoc style guide shows examples of both `array_like` and `list of` but none with the dimensions: https://numpydoc.readthedocs.io/en/latest/format.html#sections Even in Doxygen the size of C arrays is never mentioned. I think it would be more helpful to write the dimensions, but keeping that information up-to-date everytime the code changes is probably not possible.

pkreissl: At least in obvious cases like things that are and will stay 3d, I would add the dimension. The information that some kind of `array_like`/`list of` is expected without dimension could be rather non-intuitive/confusing to plain ('non programmer') users of espresso, I imagine...

src/python/espressomd/constraints.py
jngrad: Right, I have done that in 2 or 3 places, either the single line was longer than 80 chars, or it was followed by an indented paragraph. The latter is really annoying, because inconsistent line indentation in a docstring will be interpreted by Sphinx as a definition list, causing the text to be formatted differently depending on the browser. In that case I introduce a newline after the triple quote and align everything until the definition list disappears from the HTML output.

src/python/espressomd/particle_data.pyx
pkreissl: Also, for `numpy` `array_like` inputs usually can have an arbitrary dimension or at least length. That does not hold for espresso...

src/python/espressomd/collision_detection.pyx
jngrad: based on [the user guide](http://espressomd.org/html/doc/advanced_methods.html#creating-bonds-when-particles-collide), point `"bind_three_particles"`, maybe they meant this:
```suggestion
            Resolution+1 bonds are needed to accommodate the case of 180 degrees
            angles
```

src/python/espressomd/particle_data.pyx
jngrad: I agree that stating the dimension would be more helpful, yet we don't do it (torque_lab, rinertia, omega_body, etc. don't state their dimension) and we probably can't maintain it in the long term (we could re-purpose `check_type_or_throw_except()` to automatically inject the dimension in the docstring, if we don't mind the additional complexity...) and we probably won't agree on which properties/arguments are obvious candidates for stating their dimension.

My impression is that `array_like` is used as a synonym of `list of` in espresso. Shapes for example use `array_like` while particle properties use `list of`. I think it's a historical artifact, because now we accept 3-tuples, 3-lists and 3-np.arrays interchangeably thanks to implicit casts. If the user makes a mistake, an error message is printed with the actual dimension that is expected.

src/python/espressomd/particle_data.pyx
jngrad:
```suggestion
            system.part.select(type=0, q=1)
```

src/python/espressomd/particle_data.pyx
```diff
@@ -2041,25 +2041,22 @@ Set quat and scalar dipole moment (dipm) instead.")
...
 
-            a function taking a ParticleHandle as argument and returning True if
-            the particle is to be filtered for. E.g.,::
+        Or: a function taking a ParticleHandle as argument and returning True if
+        the particle is to be filtered for. E.g.,::
 
-                lambda p: p.pos[0]<0.5
+            lambda p: p.pos[0]<0.5
```
jngrad:
```suggestion
            system.part.select(lambda p: p.pos[0] < 0.5)
```

pkreissl (approved):

src/python/espressomd/particle_data.pyx
jngrad: offline discussion: @pkreissl will take care of this in a new PR after this one is merged.

fweik (approved):

RudolfWeeber:
> 
>-            rotation : (int,int,int)
>+            rotation : list of :obj:`int`

>offline discussion:  <https://github.com/pkreissl> @pkreissl will take care of this in a new PR after this one is merged.

In case that’s not there yet, a link might be added to the section on rotational degrees of freedom in the user’s guide, both, here and for the other rotation-related properties.

Issue #3034, opened 2019-07-31
Narrow signatures of interaction kernels


fweik:
Pair interactions should only take those parameters that are really needed for the calculation,
so that their dependencies are visible from the signature. E.g.
```c++
inline double morse_pair_energy(const Particle *p1, const Particle *p2,
                                const IA_parameters *ia_params,
                                const double d[3], double dist);
```
can be narrowed to
```c++
inline double morse_pair_energy(const IA_parameters *ia_params,
                                const double d[3], double dist);
```
To this end also the debug macros should be removed.

PR #3039, opened 2019-07-31
Simplify energy/force/pressure code

jngrad:
Changes:
- use Vector3d in energy/force/pressure calculation
    - affects bonded IAs, non-bonded IAs, electrostatics, magnetostatics, constraints, shapes
    - refactor code based on Vector3d-specific functions (scalar product, dot product, cross product, Hadamard product, norm)
- reduce code complexity substantially in functions involving dipoles

Required for #2502. No significant change in performance for LJ and P3M benchmarks with 1k particles (around 2% speed-up in some cases).

### DIFF
--- a/src/core/bonded_interactions/angle_cosine.hpp
+++ b/src/core/bonded_interactions/angle_cosine.hpp
@@ -74,21 +74,12 @@ calc_angle_cosine_3body_forces(Particle const *p_mid, Particle const *p_left,
  *  @param[out] f_right   Force on @p p_right.
  *  @retval false
  */
-inline bool calc_angle_cosine_force(Particle const *p_mid,
-                                    Particle const *p_left,
-                                    Particle const *p_right,
-                                    Bonded_ia_parameters const *iaparams,
-                                    double f_mid[3], double f_left[3],
-                                    double f_right[3]) {
-
-  Utils::Vector3d f_mid_v, f_left_v, f_right_v;
-  std::tie(f_mid_v, f_left_v, f_right_v) =
+inline bool calc_angle_cosine_force(
+    Particle const *p_mid, Particle const *p_left, Particle const *p_right,
+    Bonded_ia_parameters const *iaparams, Utils::Vector3d &f_mid,
+    Utils::Vector3d &f_left, Utils::Vector3d &f_right) {
+  std::tie(f_mid, f_left, f_right) =
       calc_angle_cosine_3body_forces(p_mid, p_left, p_right, iaparams);
-  for (int i = 0; i < 3; ++i) {
-    f_mid[i] = f_mid_v[i];
-    f_left[i] = f_left_v[i];
-    f_right[i] = f_right_v[i];
-  }
   return false;
 }
 
--- a/src/core/bonded_interactions/angle_cosine.hpp
+++ b/src/core/bonded_interactions/angle_cosine.hpp
@@ -46,9 +46,10 @@ int angle_cosine_set_params(int bond_type, double bend, double phi0);
  *  @return Forces on the second, first and third particles, in that order.
  */
 inline std::tuple<Utils::Vector3d, Utils::Vector3d, Utils::Vector3d>
-calc_angle_cosine_3body_forces(Particle const *p_mid, Particle const *p_left,
-                               Particle const *p_right,
-                               Bonded_ia_parameters const *iaparams) {
+calc_angle_cosine_3body_forces(Particle const *const p_mid,
+                               Particle const *const p_left,
+                               Particle const *const p_right,
+                               Bonded_ia_parameters const *const iaparams) {
 
   auto forceFactor = [&iaparams](double const cos_phi) {
     auto const sin_phi = sqrt(1 - Utils::sqr(cos_phi));
@@ -75,9 +76,9 @@ calc_angle_cosine_3body_forces(Particle const *p_mid, Particle const *p_left,
  *  @retval false
  */
 inline bool calc_angle_cosine_force(
-    Particle const *p_mid, Particle const *p_left, Particle const *p_right,
-    Bonded_ia_parameters const *iaparams, Utils::Vector3d &f_mid,
-    Utils::Vector3d &f_left, Utils::Vector3d &f_right) {
+    Particle const *const p_mid, Particle const *const p_left,
+    Particle const *const p_right, Bonded_ia_parameters const *const iaparams,
+    Utils::Vector3d &f_mid, Utils::Vector3d &f_left, Utils::Vector3d &f_right) {
   std::tie(f_mid, f_left, f_right) =
       calc_angle_cosine_3body_forces(p_mid, p_left, p_right, iaparams);
   return false;
@@ -91,9 +92,10 @@ inline bool calc_angle_cosine_force(
  *  @param[out] _energy   Energy.
  *  @retval false
  */
-inline bool angle_cosine_energy(Particle const *p_mid, Particle const *p_left,
-                                Particle const *p_right,
-                                Bonded_ia_parameters const *iaparams,
+inline bool angle_cosine_energy(Particle const *const p_mid,
+                                Particle const *const p_left,
+                                Particle const *const p_right,
+                                Bonded_ia_parameters const *const iaparams,
                                 double *_energy) {
   auto const vectors =
       calc_vectors_and_cosine(p_mid->r.p, p_left->r.p, p_right->r.p, true);
--- a/src/core/electrostatics_magnetostatics/mdlc_correction.cpp
+++ b/src/core/electrostatics_magnetostatics/mdlc_correction.cpp
@@ -19,15 +19,14 @@
   along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 /** \file
- * code for calculating the MDLC (magnetic dipolar layer
- *correction).
+ *  code for calculating the MDLC (magnetic dipolar layer correction).
  *  Purpose:   get the corrections for dipolar 3D algorithms
  *             when applied to a slab geometry and dipolar
- *	      particles. DLC & co
+ *             particles. DLC & co
  *  Article:   A. Brodka, Chemical Physics Letters 400, 62-67 (2004).
  *
- *	      We also include a tuning function that returns the
- *	      cut-off necessary to attend a certain accuracy.
+ *             We also include a tuning function that returns the
+ *             cut-off necessary to attend a certain accuracy.
  *
  *  Restrictions: the slab must be such that the z is the short
  *                direction. Otherwise we get trash.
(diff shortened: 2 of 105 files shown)


src/core/bonded_interactions/angle_cosine.hpp
```diff
@@ -100,9 +92,10 @@ inline bool calc_angle_cosine_force(Particle const *p_mid,
  *  @param[out] _energy   Energy.
  *  @retval false
  */
-inline bool angle_cosine_energy(Particle const *p_mid, Particle const *p_left,
-                                Particle const *p_right,
-                                Bonded_ia_parameters const *iaparams,
+inline bool angle_cosine_energy(Particle const *const p_mid,
```
KaiSzuttor, line 95: why do we have this strange function signature for all those bonded potentials? if a bond can break, can't we just handle this within the calculation and instead return the energy? If we want to keep the structure, we could also return a ```boost::optional```

jngrad: We need to forward the information about broken bonds to the calling function, which will produce a runtime error message that will be gathered on the master MPI node to stop the simulation (I think). In C, I would personally prefer having the broken bond as a side-effect of the function, i.e. `inline double angle_cosine_energy(..., bool *bond_broken)`. We're going to rewrite these signatures as `std::tuple<double, bool>` soon, and probably also remove the bool in those potentials where bonds can't break.

jngrad: Seems like the body of this conditional cannot be reached because the same check is performed earlier to exit the function prematurely:
https://github.com/espressomd/espresso/blob/de20f8e94cc5a69adf3ab916e92b23cf49cc3032/src/core/energy_inline.hpp#L348
The force and energy functions probably need some restructuring...

jngrad: Actually, since the energy is undefined for a broken bond, we could consider returning `std::nan` as a value and have a NaN check trigger the runtime error message. This could also be done in force kernels. What do you think?

src/core/electrostatics_magnetostatics/mdlc_correction.cpp
```diff
@@ -70,120 +67,92 @@ inline double g1_DLC_dip(double g, double x) {
...
-      auto const dip = p.calc_dip();
-      node_sums[0] += dip[0];
-      node_sums[1] += dip[1];
-      node_sums[2] += dip[2];
+      node_sums += p.calc_dip();
     }
   }
 
-  MPI_Allreduce(node_sums, tot_sums, 3, MPI_DOUBLE, MPI_SUM, comm_cart);
+  MPI_Allreduce(node_sums.data(), tot_sums.data(), 3, MPI_DOUBLE, MPI_SUM,
```
fweik, line 89: Just as a general note: There is a return value version of `MPI_Allreduce` in boost.mpi, which can be used like `auto const tot_sums = boost::mpi::all_reduce(comm_cart, node_sums, std::plus<>{})`, which allows the sum to be `const`. Also in general I think boost.mpi should be preferred, because the types and length are deduced, which is less noisy and leave fewer places for mistakes.

src/core/bonded_interactions/angle_cosine.hpp
KaiSzuttor: Why do we need to forward this information at all? If we encounter a broken bond, we can write to ```runtimeErrorMessage``` directly in the function.

Concerning the NaN approach:
CPP reference:

A common use case for optional is the return value of a function that may fail. As opposed to other approaches, such as std::pair<T,bool>, optional handles expensive-to-construct objects well and is more readable, as the intent is expressed explicitly.

Thats exactly our usecase, right?  Why not use the boost alternative to that?

KaiSzuttor: that would also be more expressive, since the caller can see from the return value that the calculation may fail

fweik: +1 for optional. The main benefit of `boost::optional` is that the caller has to check and  handle the failure case. When returning a pair or sentinel value like nan this can just be ignored, which would obfuscate the cause of the problem.

jngrad: Just tried `std::nan`, `std::optional` (C++17) and `boost::optional` on a toy example and all give the exact same assembly code. The boost version however adds a lot of extra symbols (objdump: 3420 lines, vs. 832 for the `std::nan`/`std::optional` solutions), is that fine?

jngrad: When reproducing the structure of `energy_inline.hpp`, the `std::optional`/`boost::optional` are identical and more efficient than `std::nan`: [godbolt-NaN](https://godbolt.org/z/su_sK-), [godbolt-optional](https://godbolt.org/z/ues41l), [godbolt-boost](https://godbolt.org/z/o2HzDl)

fweik: Let's just make it an optional then.

fweik: Regarding your example: I think if you have already check if the optional has a value you should use `boost::optional::get` for value extraction, instead of value. This makes all of the exception [code go away](https://godbolt.org/z/iWQ07w) because it can never throw.

jngrad: Looks much better! I think this `boost::optional` refactoring should be done in the PR where we move from output parameters to tuples, e.g.
```c++
boost::optional<double> tab_angle_energy(...);
boost::optional<std::tuple<Vector3d, Vector3d, Vector3d>> calc_tab_angle_force(...);
```
I already have the tuple version of the code in an old commit, I just need to remove the bool and add the boost::optional around each tuple. And fix a bug in that old commit.

jngrad: The tuple version without `boost::optional` ([fix-2502-p1b...fix-2502-p2](https://github.com/jngrad/espresso/compare/fix-2502-p1b...fix-2502-p2)) is now passing all tests, but the LJ script from the benchmarks shows it's 23% slower. The creation of all these intermediate `Vector3d` is probably the culprit. I'll investigate that and check if `boost::optional` has any effect on performance.

src/core/bonded_interactions/angle_cosine.hpp
```diff
@@ -46,9 +46,10 @@ int angle_cosine_set_params(int bond_type, double bend, double phi0);
  *  @return Forces on the second, first and third particles, in that order.
  */
 inline std::tuple<Utils::Vector3d, Utils::Vector3d, Utils::Vector3d>
-calc_angle_cosine_3body_forces(Particle const *p_mid, Particle const *p_left,
-                               Particle const *p_right,
-                               Bonded_ia_parameters const *iaparams) {
+calc_angle_cosine_3body_forces(Particle const *const p_mid,
```
KaiSzuttor, line 49: actually I think we should not have particles as arguments but only their positions.

src/core/bonded_interactions/angle_cosine.hpp
```diff
@@ -46,9 +46,10 @@ int angle_cosine_set_params(int bond_type, double bend, double phi0);
...
+                               Particle const *const p_left,
+                               Particle const *const p_right,
+                               Bonded_ia_parameters const *const iaparams) {
```
KaiSzuttor, line 52: And we want to have bond-specific parameter types

src/core/bonded_interactions/angle_cosine.hpp
```diff
@@ -74,21 +75,12 @@ calc_angle_cosine_3body_forces(Particle const *p_mid, Particle const *p_left,
...
-                                    Bonded_ia_parameters const *iaparams,
-                                    double f_mid[3], double f_left[3],
-                                    double f_right[3]) {
-
-  Utils::Vector3d f_mid_v, f_left_v, f_right_v;
-  std::tie(f_mid_v, f_left_v, f_right_v) =
+inline bool calc_angle_cosine_force(
+    Particle const *const p_mid, Particle const *const p_left,
+    Particle const *const p_right, Bonded_ia_parameters const *const iaparams,
+    Utils::Vector3d &f_mid, Utils::Vector3d &f_left, Utils::Vector3d &f_right) {
```
KaiSzuttor, line 81: those output parameters are evil

jngrad: agreed, they have already been dealt with in [fix-2502-p2](https://github.com/jngrad/espresso/compare/fix-2502-p1b...fix-2502-p2#diff-94e6cd4fc1e0b199e0902e54d014c802)

src/core/bonded_interactions/angle_cosine.hpp
jngrad: this means exposing the internal implementation of particles in `pressure_inline.hpp`:
```diff
   switch (iaparams->type) {
   case BONDED_IA_ANGLE_HARMONIC:
     std::tie(f_mid, f_left, f_right) =
-        calc_angle_harmonic_3body_forces(p_mid, p_left, p_right, iaparams);
+        calc_angle_harmonic_3body_forces(p_mid->r.p, p_left->r.p, p_right->r.p, iaparams);
     break;
```

src/core/bonded_interactions/angle_cosine.hpp
jngrad: this means exposing the internal implementation of Bonded_ia_parameters in `pressure_inline.hpp`:
```diff
   switch (iaparams->type) {
   case BONDED_IA_ANGLE_HARMONIC:
     std::tie(f_mid, f_left, f_right) =
-        calc_angle_harmonic_3body_forces(p_mid, p_left, p_right, iaparams);
+        calc_angle_harmonic_3body_forces(p_mid, p_left, p_right, &iaparams->p);
     break;
```

src/core/bonded_interactions/angle_cosine.hpp
KaiSzuttor: i think all these functions don't actually need a particle but really just a position

src/core/bonded_interactions/angle_cosine.hpp
KaiSzuttor: no, I thought of splitting ```Bonded_ia_parameters``` into separate structs, I think @fweik has proposed this already

jngrad: this should probably be addressed in a dedicated PR

fweik: @KaiSzuttor you are confusing this with the non-bonded parameters.

src/core/bonded_interactions/angle_cosine.hpp
fweik: There is issue #3034 for this, can be dealt with later.

fweik:
@KaiSzuttor you should maybe look at the other tickets in https://github.com/espressomd/espresso/projects/6

fweik (approved):

fweik:
I think this can be merged now.

jngrad:
bors r=fweik

PR #3043, opened 2019-08-01
Sphinx Doc: Replace `array_like` and 'list of' with `(N,) array_like of` where appropriate.

pkreissl:
Follow up on PR #3029.
As discussed offline this PR changes the code documentation such that the non-specific `array_like` inputs are replaced by n-tuples where appropriate.

Description of changes:
 - Array-like input parameters of a fixed expected length are documented as `<n>-tuple of :obj:<type>`. This style is used by numpy-doc for fixed-length inputs, see e.g. the `axis` parameter in [numpy.linalg.norm](https://docs.scipy.org/doc/numpy/reference/generated/numpy.linalg.norm.html).
 - Array-like input parameters with variable length are kept as `array_like`.
 - Return arrays are usually of type `ndarray`. This is updated where appropriate.

### DIFF
--- a/doc/sphinx/constraints.rst
+++ b/doc/sphinx/constraints.rst
@@ -318,8 +318,8 @@ Pictured is an example constraint with a ``SimplePore`` shape created with ::
 
 The resulting surface is a stomatocyte shaped boundary.
 This command should be used with care.
-The position can be any point in the simulation box and is set via the 3-tuple parameter ``center``.
-The orientation of the (cylindrically symmetric) stomatocyte is given by an ``axis`` (a 3-tuple of :obj:`float`),
+The position can be any point in the simulation box and is set via the (3,) array_like parameter ``center``.
+The orientation of the (cylindrically symmetric) stomatocyte is given by an ``axis`` (a (3,) array_like of :obj:`float`),
 which points in the direction of the symmetry axis and does not need to be normalized.
 The parameters: ``outer_radius``, ``inner_radius``, and ``layer_width``, specify the shape of the stomatocyte.
 Here inappropriate choices of these parameters can yield undesired results.
@@ -431,10 +431,10 @@ The resulting surface is a section of a hollow cone.
 The parameters ``inner_radius`` and ``outer_radius`` specifies the two radii .
 The parameter ``opening_angle`` specifies the opening angle of the cone (in radians, between 0 and :math:`\pi/2` ), and thus also determines the length.
 
-The orientation of the (cylindrically symmetric) cone is specified with the parameter ``axis`` (a 3-tuple of :obj:`float`),
+The orientation of the (cylindrically symmetric) cone is specified with the parameter ``axis`` (a (3,) array_like of :obj:`float`),
 which points in the direction of the symmetry axis, and does not need to be normalized.
 
-The position is specified via the 3-tuple ``center`` and can be any point in the simulation box.
+The position is specified via the (3,) array_like ``center`` and can be any point in the simulation box.
 
 The ``width`` specifies the width.
 This shape supports the ``direction`` parameter, ``+1`` for outward and ``-1`` for inward.
--- a/doc/sphinx/constraints.rst
+++ b/doc/sphinx/constraints.rst
@@ -318,8 +318,8 @@ Pictured is an example constraint with a ``SimplePore`` shape created with ::
 
 The resulting surface is a stomatocyte shaped boundary.
 This command should be used with care.
-The position can be any point in the simulation box and is set via the array_like parameter ``center``.
-The orientation of the (cylindrically symmetric) stomatocyte is given by an array_like ``axis``,
+The position can be any point in the simulation box and is set via the 3-tuple parameter ``center``.
+The orientation of the (cylindrically symmetric) stomatocyte is given by an ``axis`` (a 3-tuple of :obj:`float`),
 which points in the direction of the symmetry axis and does not need to be normalized.
 The parameters: ``outer_radius``, ``inner_radius``, and ``layer_width``, specify the shape of the stomatocyte.
 Here inappropriate choices of these parameters can yield undesired results.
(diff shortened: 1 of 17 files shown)


KaiSzuttor:
i think we should not change this... ```tuple``` is a python datatype and has the specific behaviour of being immutable, which is not a requirement for our parameters. Also this keyword is nowhere documented in the [numpydoc docstring guide](https://numpydoc.readthedocs.io/en/latest/format.html). Whereas I agree that ```array_like``` is too generic, ```tuple``` would be too specific and actually wrong. I think we should go with something like ```array_like of length X```

mkuron:
> tuple is a python datatype and has the specific behaviour of being immutable

Immutability is not relevant when referring to argument types, so this does not matter here. The numpy documentation seems to use tuple for fixed-length arrays, but this is indeed not documented in numpy's documentation documentation.

`array_like of length X` is not used in the numpy documentation anywhere. The only thing I could find is `array_like of ints` (as opposed to normal `array_like`s which contain floating-point numbers). Some complex examples found include `(…,N,…) array_like`, `2-D ndarray of float or tuple of 1-D array, shape (M, D)`, `(array-like, array_like)`, `array_like` with the required fixed length given in the argument description, `(M,) array_like`.

How about using `(M,) array_like`?

KaiSzuttor:
> Immutability is not relevant when referring to argument types, so this does not matter here.

Exactly, that's why not to use ```tuple```.

KaiSzuttor:
> How about using (M,) array_like?

+1

fweik:
As far as python terminology goes, I think [sequence](https://docs.python.org/3.7/glossary.html#term-sequence) is correct no?

KaiSzuttor:
[numpydoc terminology](https://numpy.org/devdocs/glossary.html#term-array-like) says ```array_like``` is what we refer to

jngrad:
@KaiSzuttor according to the definition, `array_like` is so broad it is also a synonym for scalar, which is not what we usually mean by `array_like` in espresso. Numpy docs write `2-tuple of ints` when the dimension is fixed (ex: [np.linalg.norm](https://docs.scipy.org/doc/numpy/reference/generated/numpy.linalg.norm.html)).

Concerning the immutability, I think `tuple` expresses the fact that a function won't modify the content of the passed variable, which can then be re-used safely in other parts of the simulation script. Do we have any function in the Python interface where an argument is an output parameter?

KaiSzuttor:
> Concerning the immutability, I think tuple expresses the fact that a function won't modify the content of the passed variable, which can then be re-used safely in other parts of the simulation script. Do we have any function in the Python interface where an argument is an output parameter?

If you document the parameter as ```tuple``` you actually require a python data type that we do not require. That's what I wanted to say.

jngrad:
@pkreissl in `maintainer/CI/doc_warnings.sh` you'll need to declare the new numpy types, using:
```bash
regex_sphinx_broken_link='<code class=\"xref py py-[a-z]+ docutils literal notranslate\"><span class=\"pre\">(?!(int|float|bool|str|object|list|tuple|dict|(?:numpy\.|np\.)?(?:nd)?array)<)[^<>]+?</span></code>(?!</a>)'
```

KaiSzuttor:
> @pkreissl in maintainer/CI/doc_warnings.sh you'll need to declare the new numpy types

could you please explain, why?

jngrad:
This regex gathers all ``:obj:`<type>` `` that Sphinx failed to parse. This helps detecting typos and links to classes/functions that were removed in a PR. But native Python types and numpy types are false positives (we didn't write docstrings for them in espresso), so we need to filter them out.

KaiSzuttor:
so, what do we do about this now? discuss in the offline meeting?

jngrad:
in the situation where we explicitly require a tuple, e.g. to be passed as key to a `dict`, we would write ``:obj:`tuple`(:obj:`str`, :obj:`int`)`` which would be displayed as **tuple**(**str**, **int**), similar to the standard notation in other style guides: `Tuple[str, int]` in [PEP 484](https://www.python.org/dev/peps/pep-0484/)/[Google](http://google.github.io/styleguide/pyguide.html#3199-tuples-vs-lists), `(string, int)` in old versions of PyCharm, or `tuple of (str, int)` in [pandas](https://python-sprints.github.io/pandas/guide/pandas_docstring.html#parameter-types).

For me, the ``n-tuple of :obj:`int` `` displayed as "n-tuple of **int**" is transparent, but looking more closely at the [numpy.linalg.norm](https://docs.scipy.org/doc/numpy/reference/generated/numpy.linalg.norm.html), their `n-tuple` notation actually means the Python `tuple` type, and not a *n*-tuple sequence (`np.linalg.norm([[0,1],[1,0]], axis=[0,1])` raises an error). Using the `n-tuple of ...` notation would therefore depart from standard practice, and we shouldn't rely on a boldface font to distinguish between a mathematical tuple and a Python tuple.

I'm getting more convinced that ``(3,) array_like of :obj:`float` `` should be used for e.g. particle positions, similar to what is done in the numpy docs ([numpy.rot90](https://docs.scipy.org/doc/numpy/reference/generated/numpy.rot90.html)). We can briefly discuss it in the meeting, and if everyone agrees I'll do the change with sed.

pkreissl:
Would be fine for me.

KaiSzuttor:
> in the situation where we explicitly require a tuple, e.g. to be passed as key to a dict, we would write :obj:`tuple`(:obj:`str`, :obj:`int`) which would be displayed as tuple(str, int)

do you have an example in espresso?

KaiSzuttor:
we should go with ```(N,) array_like of :obj:`float` ```

jngrad:
The `add_bond()` method requires a tuple ([docs](http://espressomd.org/html/doc/espressomd.html?highlight=add_bond#espressomd.particle_data.ParticleHandle.add_bond)), although a list is fine too because of the two casts. Side-note: the `add_bond()` code sample does not run and conflicts with the second code example in [7. Bonded interactions](http://espressomd.org/html/doc/inter_bonded.html).

jngrad (approved):

KaiSzuttor:
bors r=jngrad

PR #3045, opened 2019-08-02
LB: extract globals from functions

KaiSzuttor:

### DIFF
--- a/src/core/grid_based_algorithms/lb.cpp
+++ b/src/core/grid_based_algorithms/lb.cpp
@@ -108,7 +108,8 @@ extern constexpr const std::array<std::array<int, 19>, 19> e_ki_transposed = {
 } // namespace
 
 #ifdef ADDITIONAL_CHECKS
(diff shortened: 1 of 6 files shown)


src/core/grid_based_algorithms/lb.cpp
```diff
@@ -945,34 +961,36 @@ inline void lb_collide_stream() {
 #endif
 
   Lattice::index_t index = lblattice.halo_offset;
-  for (int z = 1; z <= lblattice.grid[2]; z++) {
-    for (int y = 1; y <= lblattice.grid[1]; y++) {
-      for (int x = 1; x <= lblattice.grid[0]; x++) {
+  for (int z = 0; z < lblattice.grid[2]; z++) {
```
fweik: Please either change this back so that `x, y, z` corresponds to the correct index, or remove them altogether, I think a single loop would work too.

fweik (changes requested):

fweik (approved):

PR #3051, opened 2019-08-04
Generate Sphinx documentation for visualizers in CI

jngrad:
Description of changes:
- CI now generates Sphinx documentation for the OpenGL and Mayavi visualizers
   - [4.0.2 docs](http://espressomd.org/html/doc4.0.2/espressomd.html#module-espressomd.visualization): module is fully documented
   - [devel docs](http://espressomd.org/html/doc/espressomd.html#module-espressomd.visualization): module is currently just 2 import errors, fixed by 63adc57 + 6aa2782 
   - @christophlohrmann 
- fixes the broken OpenGL visualizer (69314fa)
   - error: `UnboundLocalError: local variable 'types' referenced before assignment`

### DIFF
--- a/src/python/espressomd/visualization_opengl.pyx
+++ b/src/python/espressomd/visualization_opengl.pyx
@@ -336,12 +336,6 @@ class openGLLive:
         IF not ROTATION:
             self.specs['director_arrows'] = False
 
-        IF not CUDA:
-            self.specs['LB_draw_velocity_plane'] = False
-            self.specs['LB_draw_boundaries'] = False
-            self.specs['LB_draw_nodes'] = False
-            self.specs['LB_draw_node_boundaries'] = False
-
         IF not LB_BOUNDARIES and not LB_BOUNDARIES_GPU:
             self.specs['LB_draw_boundaries'] = False
             self.specs['LB_draw_node_boundaries'] = False
--- a/src/python/espressomd/visualization_opengl.pyx
+++ b/src/python/espressomd/visualization_opengl.pyx
@@ -1579,13 +1579,7 @@ class openGLLive:
         # LOOK FOR LB ACTOR
         if self.specs['LB_draw_velocity_plane'] or self.specs['LB_draw_nodes'] or self.specs['LB_draw_node_boundaries']:
             for a in self.system.actors:
-                types = [espressomd.lb.LBFluid]
-                IF CUDA:
-                    types.append(espressomd.lb.LBFluidGPU)
-
-                # if type(a) == espressomd.lb.LBFluidGPU or type(a) ==
-                # espressomd.lb.LBFluid
-                if type(a) in types:
+                if isinstance(a, espressomd.lb.HydrodynamicInteraction):
                     # if 'agrid' in pa:
                     self.lb_params = a.get_params()
                     self.lb = a
--- a/src/python/espressomd/visualization_opengl.pyx
+++ b/src/python/espressomd/visualization_opengl.pyx
@@ -1579,7 +1579,7 @@ class openGLLive:
         # LOOK FOR LB ACTOR
         if self.specs['LB_draw_velocity_plane'] or self.specs['LB_draw_nodes'] or self.specs['LB_draw_node_boundaries']:
             for a in self.system.actors:
-                types = [types.append(espressomd.lb.LBFluid)]
+                types = [espressomd.lb.LBFluid]
                 IF CUDA:
                     types.append(espressomd.lb.LBFluidGPU)
 
--- a/src/python/espressomd/visualization_opengl.pyx
+++ b/src/python/espressomd/visualization_opengl.pyx
@@ -191,21 +191,21 @@ class openGLLive:
 
     """
 
-    def __init__(self, system, **kwargs):
-        # MATERIALS
-        self.materials = {
-            'bright': [0.9, 1.0, 0.8, 0.4, 1.0],
-            'medium': [0.6, 0.8, 0.2, 0.4, 1.0],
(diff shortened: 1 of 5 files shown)


jngrad:
@mkuron I find the solution in 6aa2782 inelegant, do you remember what the following conditional was supposed to fix?
https://github.com/espressomd/espresso/blob/cefc3b954cdc8c11915ee1d46f2dd8fbb62a50af/src/python/espressomd/visualization.pyx#L5-L8

src/python/espressomd/visualization_opengl.pyx
```diff
@@ -1579,7 +1579,7 @@ class openGLLive:
         # LOOK FOR LB ACTOR
         if self.specs['LB_draw_velocity_plane'] or self.specs['LB_draw_nodes'] or self.specs['LB_draw_node_boundaries']:
             for a in self.system.actors:
-                types = [types.append(espressomd.lb.LBFluid)]
+                types = [espressomd.lb.LBFluid]
```
KaiSzuttor: why is that happening in a loop?

christophlohrmann (approved):

src/python/espressomd/visualization_opengl.pyx
jngrad: the new syntax in ea9dd76 is easily extensible: `if isinstance(a, (espressomd.lb.HydrodynamicInteraction, espressomd.walberla.lb.LB))`

src/python/espressomd/visualization_opengl.pyx
```diff
@@ -1579,13 +1579,7 @@ class openGLLive:
...
         if self.specs['LB_draw_velocity_plane'] or self.specs['LB_draw_nodes'] or self.specs['LB_draw_node_boundaries']:
             for a in self.system.actors:
-                types = [espressomd.lb.LBFluid]
-                IF CUDA:
-                    types.append(espressomd.lb.LBFluidGPU)
-
-                # if type(a) == espressomd.lb.LBFluidGPU or type(a) ==
-                # espressomd.lb.LBFluid
-                if type(a) in types:
+                if isinstance(a, espressomd.lb.HydrodynamicInteraction):
```
KaiSzuttor: have you tried running ```samples/visualization_lbboundaries.py```? I get a segmentation fault.

jngrad: nice catch, I had only tried `visualization_poiseuille.py` and `visualization_charged.py`

KaiSzuttor: it's probably not related to your PR but we should open an issue...

jngrad: my solution is probably wrong, because it includes EK while the original code didn't: `cdef class Electrokinetics(HydrodynamicInteraction)`

jngrad: the fix (50d1087) stopped working after I pushed it! `LBBoundaries` inherits from `ScriptObjectRegistry`, there shouldn't be any interference

jngrad: even better: when printing the actor, one gets:
```
WARNING: Recalculating forces, so the LB coupling forces are not included in the particle force the first time step. This only matters if it happens frequently during sampling.
LBFluid({'agrid': 1.0, 'dens': 1.0, 'ext_force_density': array([0.   , 0.003, 0.   ]), 'visc': 1.0, 'bulk_visc': -1.0, 'tau': 0.1, 'seed': None, 'kT': 0.0})
  File "../samples/visualization_poiseuille.py", line 68, in <module>
    visualizer.run(1)
  File "visualization_opengl.pyx", line 572, in espressomd.visualization_opengl.openGLLive.run
  File "visualization_opengl.pyx", line 579, in espressomd.visualization_opengl.openGLLive.start
  File "visualization_opengl.pyx", line 1588, in espressomd.visualization_opengl.openGLLive._init_espresso_visualization
  File "actors.pyx", line 86, in espressomd.actors.Actor.get_params
  File "lb.pyx", line 174, in espressomd.lb.HydrodynamicInteraction._get_params_from_es_core
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```
because `make_array_locked()` returns a `np.ndarray` in `property ext_force_density` (fixed in 536056b)

jngrad: I'm also pretty confident that the change at line 339 is incorrect: https://github.com/espressomd/espresso/commit/2255848556d048726b927e9d11c19ef50e5ab715#diff-aff7378e83b614ca98de80a01b3adaa5 ; with `LB` being enabled by default, we should not reset these user-provided flags to `False` if there is no GPU LB because the user may have set them to `True` to display a CPU LB (fixed in befba50)

jngrad: I still can't find the issue with `LBBoundaries`. Since the exact same backtrace is obtained from the current `python` branch, let's fix that bug in a separate issue: #3056

jngrad:
@KaiSzuttor any idea why `check_sphinx` rebuilt the project from scratch? [logfile](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/145361)

KaiSzuttor:
no, don't know

jngrad:
Fixed in 2ecd6ba, ready for review

src/python/espressomd/visualization_opengl.pyx
jngrad: this is now happening outside the loop

KaiSzuttor (approved):

PR #3053, opened 2019-08-05
Max cut

fweik:
Description of changes:
 - Replaced some globals related to cutoff calculation
 - Store range limits in cell_structure
 - Consistently use `INACTIVE_CUTOFF`/`-1.` for cutoffs
   of disabled kernels.

Further refactoring is blocked by the fact that the short-range kernels
with different cutoffs are fused together, and by the unclear resposibility
for the verlet list. The performance hacks that mix everything together
also do not help.


KaiSzuttor:
we should probably deactivate this clang-tidy warning...

fweik:
I think the point is valid.

RudolfWeeber:
LGTM, except one doxygen thing

>@@ -125,7 +117,8 @@ extern int min_num_cells;
>  *                cells_on_geometry_change.
>  *  @param grid   Number of nodes in each spatial dimension.
>  */
>-void dd_on_geometry_change(int flags, const Utils::Vector3i &grid);
>+void dd_on_geometry_change(int flags, const Utils::Vector3i &grid,
>+                           double range);
Please add doxygen for range

RudolfWeeber (approved):
bors r+

jngrad:
@RudolfWeeber there is a merge conflict, `bors` will fail and block processing of any other PR in the queue

jngrad:
bors retry

PR #3054, opened 2019-08-05
Coldet force recalc

### LINKED ISSUE #3027: Evaluating forces executes collisions
Running `System.integrator.run(steps=0)` does run the collision detection, and therefor mutates the system. I think it should not, as system mutation should only occur if the system is propagated. Otherwise we have results depending on how often you looked at the system,
which would create our very own measurement problem.
Removed the collision handling in the initial force calculation breaks some tests. @RudolfWeeber is this behavior desired? I'm sure the collisions can be tested without this...

RudolfWeeber:
Fixes #3027

### DIFF
--- a/testsuite/python/collision_detection.py
+++ b/testsuite/python/collision_detection.py
@@ -615,7 +615,7 @@ class CollisionDetection(ut.TestCase):
                                               bond_three_particles=2, three_particle_binding_angle_resolution=res, distance=cutoff)
         self.get_state_set_state_consistency()
         
-        self.s.time_step=1E-6
+        self.s.time_step = 1E-6
         self.s.integrator.run(1, recalc_forces=True)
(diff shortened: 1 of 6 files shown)


testsuite/python/collision_detection.py
```diff
@@ -79,7 +80,7 @@ def test_bind_centers(self):
         self.s.part.add(pos=(0, 0, 0), id=0)
         self.s.part.add(pos=(0.1, 0, 0), id=1)
         self.s.part.add(pos=(0.1, 0.3, 0), id=2)
-        self.s.integrator.run(0)
+        self.s.integrator.run(1)
```
fweik, line 83: Wouldn't it be cleaner to add an interface function to explicitly execute the collision detection ?
Like
```c++
void collide() {
 on_observable_calc();
 short_range_loop(...);
 handle_collisions();
}
```
should work right?

fweik (approved):

PR #3061, opened 2019-08-07
Avoid unneeded reallocations of ghost data

fweik:
So I looked at profiles of the particle exchange code, and noticed
some things... In some situations the can improve the performance
considerably, so I think this should go into the release.
This has some overlap with #2296, but the approach is similar, 
and this is slightly cleaner and can be merged.
It's slightly hackish, but I think this is the best we can do with
the current `Cell` data structure.

Description of changes:
 - Realloc ghost particles only if needed
 - Fixed memleak in resort code
 - Fixed corner-case of Utils::List::resize
 - Removed exclusions from ghosts

### DIFF
--- a/src/core/ghosts.cpp
+++ b/src/core/ghosts.cpp
@@ -165,11 +165,9 @@ void prepare_send_buffer(GhostCommunication *gc, int data_parts) {
         }
         if (data_parts & GHOSTTRANS_POSSHFTD) {
           /* ok, this is not nice, but perhaps fast */
-          auto *pp = reinterpret_cast<ParticlePosition *>(insert);
-          int i;
-          *pp = pt->r;
-          for (i = 0; i < 3; i++)
-            pp->p[i] += gc->shift[i];
+          auto pp = pt->r;
+          pp.p += gc->shift;
+          memcpy(insert, &pp, sizeof(pp));
           insert += sizeof(ParticlePosition);
         } else if (data_parts & GHOSTTRANS_POSITION) {
           memcpy(insert, &pt->r, sizeof(ParticlePosition));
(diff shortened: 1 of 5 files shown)


src/core/ghosts.cpp
```diff
@@ -221,26 +222,41 @@ void prepare_send_buffer(GhostCommunication *gc, int data_parts) {
...
-    }
+
+  /* initialize new particles */
+  if(old_cap < cell->max) {
+    auto new_parts = make_span(cell->part + old_cap, cell->max - old_cap);
+    std::uninitialized_fill(new_parts.begin(), new_parts.end(), Particle{});
+  }
+
+  for(auto &p : Utils::make_span(cell->part, cell->n)) {
+    /* Keep heap if any, but set the logical size to 0 */
```
jngrad: From a maintainability point of view, wouldn't it be more convenient for this loop to rely on a `Particle::reset()` method, instead of copy-pasting the ctor? If we ever add a new struct member or remove an ifdef inside `struct Particle`, and forget about that code duplication in `ghosts.cpp`, the compiler won't generate a warning. This is also exposing too much of the `Particle` implementation details.

fweik: Actually, this isn't needed at all.

fweik: In principle no initialization should be needed at all, because the fields should not be accessed before they are updated. In reality multiple methods break if I do this. This is not a good sign.

fweik:
So I figured it out.

jngrad (approved):

jngrad:
bors actually failed on retry because of an incorrect commit in GitLab

jngrad:
bors retry

Issue #2544, opened 2019-03-01
NPT should use normal pressure calculation


fweik:
Currently NPT uses it's own pressure calculation integrated into the force calculation. This is a duplication and has diverged, so that NPT does not use the same pressure as calculated by the "normal" pressure calculation. It should just use the pressure kernels.

jngrad:
Is this fixed by #3756?

fweik:
No, unfortunately not.

jngrad:
The NpT code has since been entirely rewritten with two barostats, one based on the MTK scheme and one based on the Andersen scheme. Since they evaluate the pressure at the half time step, they cannot produce the same pressure as the one calculated by the analysis method. The analysis method was refactored several times to reduce code duplication.

PR #3062, opened 2019-08-07
Integration loop cleanup part 1

RudolfWeeber:
* Move ghost comm logic into methods (rattle, virtual sites)
* Outsource some NPT and SIG_INT stuff into functions
* Don't re-check and communicate LB coupling state at every time step


src/core/integrate.cpp
```diff
@@ -241,37 +260,29 @@ void integrate_vv(int n_steps, int reuse_forces) {
...
+         v(t+0.5*dt) = v(t) + 0.5*dt * a(t)
+         p(t + dt)   = p(t) + dt * v(t+0.5*dt)
+         NOTE: Depending on the integration method Step 1 and Step 2
+         cannot be combined for the translation.
+      */
+      velocity_verlet_step_1_and_2(particles);
     }
 
+    /* Propagate time: t = t+dt */
+    sim_time += time_step;
```
fweik: Energy minimization should not propagate time.

src/core/integrate.cpp
```diff
@@ -167,6 +173,20 @@ void integrate_ensemble_init() {
...
 
 /************************************************************/
+/** @brief Step 1 and 2 of Velocity Verlet scheme
+         v(t+0.5*dt) = v(t) + 0.5*dt * a(t)
+         p(t + dt)   = p(t) + dt * v(t+0.5*dt)
+         NOTE: Depending on the integration method Step 1 and Step 2
+         cannot be combined for the translation.
+*/
+void velocity_verlet_step_1_and_2(ParticleRange &particles) {
+  if (integ_switch == INTEG_METHOD_NPT_ISO) {
```
fweik: please keep the `integ_switch` stuff in the calling context. This does not belong here, the function does not need to know how it is used.

src/core/rattle.cpp
```diff
@@ -170,16 +170,19 @@ void app_pos_correction(const ParticleRange &particles) {
   } // for i loop
 }
 
-void correct_pos_shake(const ParticleRange &particles) {
+void correct_pos_shake(ParticleRange const &particles) {
+  cells_update_ghosts();
   int repeat_, cnt = 0;
   int repeat = 1;
 
+  // Note: The particle range has to be re-obtained from the cell_structure
```
fweik: This is not true, they are invalidated by a resort. It's enough to get them again after `cells_update_ghosts`, it's not needed after every ghost comm. Why would it?

fweik (changes requested):

src/core/integrate.cpp
```diff
@@ -241,37 +260,29 @@ void integrate_vv(int n_steps, int reuse_forces) {
...
-       cannot be combined for the translation.
-    */
-    if (integ_switch == INTEG_METHOD_NPT_ISO) {
-      propagate_vel(particles);
-      propagate_pos(particles);
-
-      /* Propagate time: t = t+dt */
-      sim_time += time_step;
-    } else if (integ_switch == INTEG_METHOD_STEEPEST_DESCENT) {
+    if (integ_switch == INTEG_METHOD_STEEPEST_DESCENT) {
```
fweik: I think this could be made clearer here by introducing an other level of abstraction:
What we are doing here is running a two-step integration scheme, the actual implementation
is determined by `integ_switch`. Subsequently there should be a function here that is called
`step_1` or `intgrator_step1` or similar, that den switches on `integ_switch`. Then, after the
force calc there should be the corresponding `step_2` function. The fact that for VV the steps
are divided further is of no concern to the caller. For one-step integrators, like steepest descent, the 2nd step is a noop.

fweik (dismissed):

src/core/integrate.cpp
fweik: This is fundamental thing here, time and time steps belong to a specific integrator. There are integrators that have no concept of time, like steepest descent.

RudolfWeeber:
@fweik, something in the ghost comm does seem to invalidate the iterator. With below changes, the rigid_bond test segfaults
```

diff --git a/src/core/integrate.cpp b/src/core/integrate.cpp
index 91fe233ed..f1a2181a8 100644
--- a/src/core/integrate.cpp
+++ b/src/core/integrate.cpp
@@ -283,7 +283,10 @@ void integrate_vv(int n_steps, int reuse_forces) {
     /**Correct those particle positions that participate in a rigid/constrained
      * bond */
     if (n_rigidbonds) {
+      particles = cell_structure.local_cells().particles();
       correct_pos_shake(cell_structure.local_cells().particles());
+      // A resort may have been triggered.
+      particles = cell_structure.local_cells().particles();
     }
 #endif
 
diff --git a/src/core/rattle.cpp b/src/core/rattle.cpp
index 090ab07f3..2d7154012 100644
--- a/src/core/rattle.cpp
+++ b/src/core/rattle.cpp
@@ -175,14 +175,12 @@ void correct_pos_shake(ParticleRange const &particles) {
   int repeat_, cnt = 0;
   int repeat = 1;
 
-  // Note: The particle range has to be re-obtained from the cell_structure
-  // after ghost updates
   while (repeat != 0 && cnt < SHAKE_MAX_ITERATIONS) {
-    init_correction_vector(cell_structure.local_cells().particles());
+    init_correction_vector(particles);
     repeat_ = 0;
     compute_pos_corr_vec(&repeat_, cell_structure.local_cells().particles());
     ghost_communicator(&cell_structure.collect_ghost_force_comm);
-    app_pos_correction(cell_structure.local_cells().particles());
+    app_pos_correction(particles);
     /**Ghost Positions Update*/
     ghost_communicator(&cell_structure.update_ghost_pos_comm);
     if (this_node == 0)
```

fweik:
No, that has to have another reason.

fweik:
Also I'm currently looking into the ghost comm, if you wait for my PR this will go away. This would also avoid a ton of merge conflicts.

RudolfWeeber:
So, from the main integration loop, there are now calls to
integrator_step_1() and integrator_step_2()
These contain a switch(integ_switch) statement, each.

Individual integrators are now in src/core/integrators
these provide hooks like
velocity_verlet_step_1(), etc.

Currently, there are three integrators
* steepest descent
* velocity verlet
* velocity verlet with NPT

NPT constitues an independent integrator, because the equations of motion are changed.
The Langevin thermostat in its current implementation only adds forces, but uses the standard velocity verlet steps. Whether that will stay that way long-term is a different matter.

If there are no objections against the general structure, I'm going to clean up and also move the remaining NPT details out of the main integration loop.
That's about as far as I currently want to take this project.

I also think, this PR will be a pre-requiste for the Brownian dynamics work.

src/python/espressomd/integrate.pyx
```diff
@@ -145,18 +145,22 @@ cdef class Integrator:
...
-        if (integrate_set_npt_isotropic(ext_pressure, piston, direction[0], direction[1], direction[2], cubic_box)):
-            handle_errors("Encountered errors setting up the NPT integrator")
+        IF NPT:
+            self._method = "NPT"
+            self._isotropic_npt_params['ext_pressure'] = ext_pressure
+            self._isotropic_npt_params['piston'] = piston
+            self._isotropic_npt_params['direction'] = direction
+            self._isotropic_npt_params['cubic_box'] = cubic_box
+            if "NPT" not in espressomd.code_info.features():
+                raise Exception("NPT is not compiled in")
```
jngrad: this conditional is now redundant

src/core/integrate.cpp
```diff
@@ -166,7 +153,43 @@ void integrate_ensemble_init() {
...
+  default:
+    throw std::runtime_error("Unknown value for integ_switch");
+  }
+  return false;
+}
+
+void integrator_step_2(ParticleRange &particles) {
+  switch (integ_switch) {
+  case INTEG_METHOD_STEEPEST_DESCENT:
+    // Nothgin
```
fweik, line 149: typo

src/core/integrators/velocity_verlet_inline.hpp
```diff
@@ -0,0 +1,80 @@
...
+        /* Propagate velocity: v(t+dt) = v(t+0.5*dt) + 0.5*dt * a(t+dt) */
+        p.m.v[j] += 0.5 * time_step * p.f.f[j] / p.p.mass;
+#ifdef EXTERNAL_FORCES
+      }
+#endif
+    }
+  }
+}
+
+void velocity_verlet_step_1(const ParticleRange &particles) {
```
fweik: Needs to be `inline` if it stays in the header

src/core/integrators/velocity_verlet_inline.hpp
```diff
@@ -0,0 +1,80 @@
...
+  velocity_verlet_propagate_vel_pos(particles);
+  sim_time += time_step;
+}
+
+void velocity_verlet_step_2(const ParticleRange &particles) {
```
fweik: Needs to be `inline` if it stays in the header

src/core/integrators/velocity_verlet_inline.hpp
```diff
@@ -0,0 +1,80 @@
+#ifndef INTEGRATORS_VELOCITY_VERLET_HPP
+#define INTEGRATORS_VELOCITY_VERLET_HPP
+
+#include "ParticleRange.hpp"
+#include "config.hpp"
+#include "particle_data.hpp"
+
+inline void velocity_verlet_propagate_vel_pos(const ParticleRange &particles) {
```
fweik, line 14: Why do you keep the implementations in a header?

fweik (commented):
I don't agree with the naming. There are different concepts mixed here, and the names are actually upright misleading. NPT e.g. also is integrated by velocity verlet. What really is selected is the equation of motion (e.g. the dynamics). I'm not sure if there is a term for the combination of a specific numeric scheme and an equation of motion, but if not these things should be named by their dynamics, not by the numeric scheme. If I decide tomorrow to Langevin dynamcis with leap-frog, would you rename everything? Or if I integrate the energy minimization by velocity verlet? Also the naming by ensemble is not great,
because it depends on what you are actually doing whether it is any ensemble at all, e.g. you say
`set_NVE` in Espresso, but actually neither of those quantities is is conserved nor is the system in equilibrium. I think some basic respect for statistical mechanics would suit us well here.

RudolfWeeber:
>I don't agree with the naming. There are different concepts mixed here, and the names are actually upright misleading. NPT e.g. also is integrated by velocity >verlet. What really is selected is the equation of motion (e.g. the dynamics). I'm not sure if there is a term for a specific numeric scheme, but if not these things >should be named by their dynamics, not by the numeric scheme.

Since the kernels in integrators/ implement a specific integration scheme (such as velocity verlet), I think this is what they should be called.

 

The values that `integ_switch` can take need to be renamed to match the kernels.

 

NPT should be `velocity_verlet_npt` because that is what the equations of motinos are.

 

Otherwise, please suggest alternative names for the kernels and values of integ_switch.

 

Accomodating schemes with more force calculations such as Runge-Kutta is beyond the scope of this PR. Still, disentangling the velocity verlet and velocity verlet NPT kernels from the main loop makes adding those easier in the future.

fweik:
I think all the names should be on the same level of abstraction, and from the same category.

fweik:
Just as a side: Espresso can of course do Brownian Dynamics today. It's activated by setting all velocities to zero, the integrator to steepest descent and the thermostat to Langevin. If the system
is in equilibrium this draws configurations from the NVT equilibrium distribution. If this was clear
from how things a named and documented in Espresso you are a true expert, but I don't think
it is reasonable to expect this level of expertise from an average user or a student starting in the
code base.

RudolfWeeber:
>Just as a side: Espresso can of course to Brownian Dynamics today. It's activated by setting all velocities to zero, the integrator to steepest descent and the >thermostat to Langevin.

Wikipeida says:

\dot{X}(t) = - \frac{D}{k_B T} \nabla U(X) + \sqrt{2 D} R(t)

So, in equilibrium, yes, probably. As long as \nable U(x) is non-zero, maybe not.

 

With regards to naming:

In my opinion, the current names approximately describe, what the functions do. NPt actually alters the equation of motion of the particles.

 

Having said that, I’m also not particularly attached to the current names. Do you have an alternative suggestion?

fweik:
It does exactly do BD, in and out of equilibrium. But whether this results in a NVT distribution of configurations depends on the system. You can't prescribe the ensemble like this, because this depends on other factors. What actually is set is the dynamics, which might result in a specific
ensemble if some other independent conditions are met. This is why I would prefer to use
the dynamics as names (= equation of motion). NpT is also Langevin dynamics, the barostat
works by adding an additional degree of freedom (the piston), which is also following Langevin
dynamics. In principle this would also work with say an Anderson thermostat on the piston (don't do this at home). In general I think that these are very central concepts of (cg-)MD and that it would be useful to contemplate on them until we have clarity about them. I'm not happy with my personal lack of understanding in this area, because we're supposed to be the experts here, and obviously at least I am not. I think a lack of understanding of the core concepts will always come back and bite us, as is regularly illustrated in Espresso.

That being said, the rest of the structure is ok for me now. Please grep for `integ_switch` and move all of this to the integration (e.g. there is some in the force calculation). Also please address
the other points raised. Also please remove the asserts on `integ_switch`.

RudolfWeeber:
I moved out more NPT stuff, but I'ḿ leaving the integrator setup functions in integrate.cpp.
NPT stuff is still called from a lot of places. This will only go away, once NPT uses the standard stress calculation (#2544).

fweik:
We should maybe start to write down the refactoring dependency graph somewhere, not sure if there is a tool for this somewhere. This would also help to make it visible how substantial the mess is that we are in.

RudolfWeeber:
That's, how far I'd like to take it at present.
I'm sorry, the diff is hard to read. Most of it is code being moved around:
* the propagation kernels from integrate.cpp wen toto 
  * integrators/velocity_verlet_ineline.hpp
  * integrators/velocity_verlet_npt.cpp/hpp
In the individual kernels the checks and feature guards for NPT have been removed.

* A lot of npt specific code from integrate.cpp and a few places in the force calculation were extracted into functions in npt.cpp

* The responsibility for ghost communication related to virtual sites was moved into the virtual site implementations

* The main integration loop now has two propagation hooks, one before, and one after the force calculation.
* The switching of kernels based on integ_dwitch happens in integrate.cpp.


Outlook:
The next step would be to disentangle force calculation and npt-pressure calculation.

KaiSzuttor:
seems like you forgot to commit a file:
```/builds/espressomd/espresso/src/core/integrate.cpp:62:47: fatal error: integrators/velocity_verlet_npt.hpp: No such file or directory```

File: src/core/integrate.cpp [hunk 10]
```diff
@@ -130,46 +117,46 @@ void integrator_sanity_checks() {
...
 
-#ifdef NPT
+/************************************************************/
 
-void integrator_npt_sanity_checks() {
-  if (integ_switch == INTEG_METHOD_NPT_ISO) {
-    if (nptiso.piston <= 0.0) {
-      runtimeErrorMsg() << "npt on, but piston mass not set";
-    }
+bool integrator_step_1(ParticleRange &particles) {
```
KaiSzuttor, line 125: please add a short docstring for this function

src/core/integrate.cpp
```diff
@@ -130,46 +117,46 @@ void integrator_sanity_checks() {
...
+  default:
+    throw std::runtime_error("Unknown value for integ_switch");
   }
+  return false;
 }
-#endif /*NPT*/
 
-/************************************************************/
-void integrate_ensemble_init() {
+void integrator_step_2(ParticleRange &particles) {
```
KaiSzuttor, line 146: please also add a docstring for this function

src/core/integrate.cpp
```diff
@@ -329,47 +274,36 @@ void integrate_vv(int n_steps, int reuse_forces) {
...
-      // Set global to notify Python of signal.
-      set_py_interrupt = true;
-
-      // Break the integration loop
+      notify_sig_int();
       break;
     }
-  }
+
+  } // for(step ...)
```
KaiSzuttor: what is this comment for?

src/core/integrators/velocity_verlet_inline.hpp
```diff
@@ -0,0 +1,79 @@
+#ifndef INTEGRATORS_VELOCITY_VERLET_HPP
+#define INTEGRATORS_VELOCITY_VERLET_HPP
+
+#include "ParticleRange.hpp"
+#include "config.hpp"
+#include "particle_data.hpp"
+#include "rotation.hpp"
+
+inline void velocity_verlet_propagate_vel_pos(const ParticleRange &particles) {
```
KaiSzuttor, line 14: it would also be nice to have a docstring mentioning which part of VV is performed here (maybe a reference to literature)

src/core/integrators/velocity_verlet_inline.hpp
```diff
@@ -0,0 +1,79 @@
...
+    if (Utils::sqr(p.r.p[0] - p.l.p_old[0]) +
+            Utils::sqr(p.r.p[1] - p.l.p_old[1]) +
+            Utils::sqr(p.r.p[2] - p.l.p_old[2]) >
+        skin2)
+      set_resort_particles(Cells::RESORT_LOCAL);
+  }
+}
+
+inline void
+velocity_verlet_propagate_vel_final(const ParticleRange &particles) {
```
KaiSzuttor, line 54: it would also be nice to have a docstring mentioning which part of VV is performed here (maybe a reference to literature)

src/core/integrators/velocity_verlet_inline.hpp
```diff
@@ -0,0 +1,79 @@
...
+        /* Propagate velocity: v(t+dt) = v(t+0.5*dt) + 0.5*dt * a(t+dt) */
+        p.m.v[j] += 0.5 * time_step * p.f.f[j] / p.p.mass;
+#ifdef EXTERNAL_FORCES
+      }
+#endif
+    }
+  }
+}
+
+inline void velocity_verlet_step_1(const ParticleRange &particles) {
```
KaiSzuttor, line 75: it would also be nice to have a docstring mentioning which part of VV is performed here (maybe a reference to literature)

src/core/integrators/velocity_verlet_npt.cpp
```diff
@@ -0,0 +1,193 @@
...
+#include "communication.hpp"
+#include "grid.hpp"
+#include "integrate.hpp"
+#include "nonbonded_interactions/nonbonded_interaction_data.hpp"
+#include "npt.hpp"
+#include "particle_data.hpp"
+#include "thermostat.hpp"
+#include "utils/math/sqr.hpp"
+
+void velocity_verlet_npt_propagate_vel_final(const ParticleRange &particles) {
```
KaiSzuttor, line 16: it would also be nice to have a docstring mentioning which part of VV is performed here (maybe a reference to literature)

KaiSzuttor (changes requested):
the structure looks good to me, however, I think we should have at least a short docstring for each of the functions

RudolfWeeber:
@KaiSzuttor @fweik 
Kai, high level integrator docs are in the UG.

KaiSzuttor (approved):

KaiSzuttor:
@fweik

jngrad:
workaround if there is no other job in the bors queue: update this branch, then merge manually

KaiSzuttor:
I can just dismiss the review, if @fweik is not available the next days

fweik:
There is apparently no way to retract the review, so just dismiss it.

KaiSzuttor:
you could just approve ;)

fweik:
Well that would look like endorsement :-)

jngrad:
bors retry

PR #3063, opened 2019-08-07
Separate types for Tabulated bonds

### LINKED ISSUE #3048: Separate types for Tabulated bonds
Currently there is one bond type for all tabulated bonds (linear, angle, ...), and the actual type is discriminated using an extra parameter. Because of this the number of interaction partners for the bonds can not be statically determined, which would allow for cleaner and more efficient code. This can be solved by introducing separate types for the different kinds of tabulated bonds, and derive/CRTP from a common class.

jngrad:
Fixes #3048

There is still a single C++ tabulated struct, however the union now has 3 tabulated tags and the Python interface has 3 classes (the old one was renamed to avoid silently breaking the API).


jngrad:
Offline discussion with @RudolfWeeber: this change can go in 4.1

fweik:
That was quick. Looks good to me, if the CI passes. I'll give variant a quick shot later, should be easy now.

fweik (approved):

PR #3065, opened 2019-08-08
Reduce usage of local_cells

reinaual:
Fixes part of #2899


src/core/virtual_sites/virtual_sites_com.hpp
```diff
@@ -35,11 +35,11 @@ void update_mol_vel_particle(Particle *);
 
 // Distribute forces that have accumulated on virtual particles to the
 // associated real particles
-void distribute_mol_force();
+void distribute_mol_force(const ParticleRange &particles)
```
jngrad: The semi-colon disappeared here. Files that include this header won't compile.

jngrad (changes requested):

src/core/virtual_sites/virtual_sites_com.hpp
jngrad: You'll also have to undo the style patch in that file.

jngrad (approved):

jngrad:
bors r-
(I'll resolve the merge conflict here, it seems to be just a newline character)

PR #3068, opened 2019-08-09
WIP: Non-bonded interactions refactor

### LINKED ISSUE #3060: Separate generic properties from interaction potentials
The actual interaction potentials should be separated from generic properties like
cutoff, offset and shift. This is independent from each other, would give all potentials
shifts and cutoff and would reduce code duplication. We also had bugs in the past
because potentials did not correctly calculate the cutoff if there was a shift. 
Here is a rough draft of what I have in mind:

```c++
template<class Potential>
struct CetralPotential : Potential {
   double shift, offset, cutoff;

   double energy(double r) {
       auto const r_eff = r + offset;
      if(r_eff <= cutoff) 
        return Potential::U(r) + shift;
      else
        return 0.;
   }
};

struct Harmonic {
  double k;
   double U(double r) {
       return -k*r*r;
   }
};

struct IA_Parameters {
  [...]
  CentralPotential<Harmonic> harmonic;
  [...]
}
```

This also allows for easier reuse of the potential functions (observe how
the potential knows nothing about how it is used). If all the potentials have
the same interface, they can also be used in a homogeneous way, e.g.
we can have

```c++
using IA_Parameters = std::tuple<Potentials...>;
IA_Parameters ia_params;

double energy = 0.;
Utils::for_each(ia_params, [r, energy](auto const& ia) {
  energy += ia.energy(r);
});  
```

Some more thought should go into this I think, but I think something along these
lines is a good starting point.

KaiSzuttor:
Fixes #3060

This is for discussing ideas for the refactoring.

I refactored LennardJones into an independent Utils function. This has to be done for all interactions which is quite some work. After that, we could then switch to the tuple that I wrote down for demonstration purposes in ```central_potential.hpp```


RudolfWeeber:
I'd like to take part in this discussion, but I'd also ask that the discussion be postponed until after the 4.1 release. 
There are still more than 40 tickets open for the release.

fweik:
Again, how to assign and store the parameters is a separate issue. Please don't mix those things.

KaiSzuttor:
> Again, how to assign and store the parameters is a separate issue. Please don't mix those things.

the branch is called ```nb_ia_refactor```  and not ```fix_issue_3060```... it's not a big deal to pull that apart
the issue just reminded me on the design flaw in that part of the code and I wanted to discuss that within this PR. I think the refactoring of the potential is quite clear

src/utils/include/utils/hash/Cantor.hpp
```diff
@@ -0,0 +1,34 @@
...
+#define UTILS_HASH_CANTOR
+
+#include <utility>
+
+namespace Utils {
+namespace Hash {
+
+struct CantorPairing {
+  std::size_t cantor_pairing(int k1, int k2) const {
+    return static_cast<std::size_t>(0.5 * (k1 + k2) * (k1 + k2 + 1) + k2);
```
fweik, line 11: This should be `((k1 + k2) * (k1 + k2 + 1)) / 2` to avoid the promotion to double, then also the cast can be removed.

src/utils/include/utils/interaction/LennardJones.hpp
```diff
@@ -0,0 +1,30 @@
...
+#define UTILS_LENNARD_JONES
+
+#include "utils/Vector.hpp"
+#include "utils/math/int_pow.hpp"
+#include "utils/math/sqr.hpp"
+
+namespace Utils {
+namespace Interaction {
+
+struct LennardJones {
```
fweik, line 11: I can't see how this is a utility function, why can't we just keep it in non_bonded_interactions? If the potentials are at some point really reused, they can be their own library.

fweik:
I think the first step would be to look at all the short-range stuff in Espresso, see what types of interactions there are and what their requirements are, and then decided on how to group them.
Then implementation can start.

PR #3071, opened 2019-08-09
Non bonded interaction container

KaiSzuttor:
Description of changes:
 - use ```std::unordered_map``` with symmetric hash for pair instead of ```std::vector```

### DIFF
--- a/src/utils/include/utils/hash/Cantor.hpp
+++ b/src/utils/include/utils/hash/Cantor.hpp
@@ -1,0 +1,32 @@
+#ifndef UTILS_HASH_CANTOR
+#define UTILS_HASH_CANTOR
+
+#include <utility>
+
+namespace Utils {
+namespace Hash {
+
+struct CantorPairing {
+  std::size_t cantor_pairing(int k1, int k2) const {
+    return static_cast<std::size_t>(0.5 * (k1 + k2) * (k1 + k2 + 1) + k2);
+  }
+
+  std::size_t operator()(std::pair<int, int> const &p) const {
+    // make it symmetric
+    if (p.second > p.first) {
+      return cantor_pairing(p.first, p.second);
+    }
+    return cantor_pairing(p.second, p.first);
+  }
+};
+
+struct CantorCompare {
+  bool operator()(std::pair<int, int> const &lhs, std::pair<int, int> const &rhs) const {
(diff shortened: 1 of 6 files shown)


fweik:
I'd expect a vector to have way better performance. What is you rational of using a hash table over a vector?

src/utils/include/utils/hash/Cantor.hpp
```diff
@@ -0,0 +1,34 @@
...
+#define UTILS_HASH_CANTOR
+
+#include <utility>
+
+namespace Utils {
+namespace Hash {
+
+struct CantorPairing {
+  std::size_t cantor_pairing(int k1, int k2) const {
+    return static_cast<std::size_t>(0.5 * (k1 + k2) * (k1 + k2 + 1) + k2);
```
fweik, line 11:
```c++
return ((k1 + k2) * (k1 + k2 + 1))/2 + k2);
```

KaiSzuttor:
a map from pair to interaction parameter seems much more expressive to me than ussing a flat matrix that uses double the amount of entries and is directly exposed. Do you really think that such a sinple hash function has measurable impact on the performance?

fweik:
No, I think that `std::unordered_map` has  terrible performance. Also a matrix is also a map from pairs to the parameters. This is about if this should be optimized for space (hash table) or for access time (vector). Sparse or dense. This is also independent of whether it is symmetrical or not. You can do the dense one with only an upper triangle just fine.

fweik:
See #3072

KaiSzuttor:
i should have read this http://scottmeyers.blogspot.com/2015/09/should-you-be-using-something-instead.html?m=1

fweik:
(off-topic: I thought [C++Now 2018: You Can Do Better than std::unordered_map](https://www.youtube.com/watch?v=M2fKMP47slQ) was a nice talk on implementation details on hash tables and performance)

KaiSzuttor:
> See #3072

I think this still leaves the caller with the burden to use an index into an upper triangular matrix. I think it would be cleaner to hide this implementation detail and have a symmetric mapping of the indices: ```flat_index(i, j, n) == flat_index(j, i, n)```  with ```i,j``` &isin; ```[0, n)```

fweik:
If you want an interface, you can always define one.

PR #3074, opened 2019-08-11
utils: boost::mpi::is_mpi_datatype specializations for Utils types

fweik:
Description of changes:
 - Defining `boost::mpi::is_mpi_datatype` for serializable datatype from `namespace Utils`

This can avoid some copies when sending those datatype over mpi.


fweik:
Argh

PR #3075, opened 2019-08-12
Store non-bonded interaction parameters only once per pair

fweik:
Description of changes:
 - Index function for upper triangular matrices
 - Store non-bonded interaction parameters only once per pair


KaiSzuttor:
could you please adapt ```realloc_ia_params```

fweik:
Obviously `max_seen_particle_type` is not the largest seen particle type...

KaiSzuttor (approved):
it's definitely an improvement

PR #3077, opened 2019-08-12
Make force/energy kernels pure functions

### LINKED ISSUE #2502: Use Vector3d and std::tuple in force functions

### LINKED ISSUE #3033: Replace output parameters by return values for energy kernels
Same as #2502, but for the energies.

jngrad:
Fixes #2502, fixes #3033

Description of changes:
- make force/energy kernels pure functions
- use `boost::optional` for breakable bonded IAs


jngrad:
I'm opening this as WIP because reviews might find ways to improve the code further.

OIF global forces and some ELC kernels aren't pure functions.

Here's the LJ performance:

|  commit |            label | empty | default | maxset |
| ------- | ---------------- | ----- | ------- | ------ |
| 2d5e097 |  boost::optional |  65%  |  108%   | 114%   |
| 7e10ad0 |           python |  61%  |  100%   | 105%   |

I'm not displaying the P3M results because the non-deterministic nature of the tuning function is responsible for the large performance fluctuations (up to 27 percentage points) observed in previous benchmarks.

KaiSzuttor:
If we'd really wanted to improve the performance we would probably need to use some kind of expression templates?

RudolfWeeber:
On Fri, Aug 23, 2019 at 03:29:29AM -0700, Kai Szuttor wrote:
> If we'd really wanted to improve the performance we would probably need to use some kind of expression templates?
We should probably also figure out, whether it is really necessary to be able to use more than one short-rnage kernel for a given pair of particle types.
I personally never did that and I'm not aware of a resonable application.

jngrad:
I don't see how to improve performance further in this PR. The commit 3ea4ac6 managed to soften the performance impact of making the kernels pure functions, however it will need to be reverted in espresso 5: once we refactor the non-bonded IAs logic using switches or variants to only evaluate those kernels that are active, this summation will go away.

|  commit |   label | empty | default |  maxset |
| ------- | ------- | ----- | ------- | ------- |
| 3ea4ac6 |  reduce |  67%  |   105%  |   113%  |
| 6374bc7 |  merge  |  68%  |   109%  |   115%  |
| e2e7c5d |  python |  63%  |   100%  |   105%  |

edit: revert 3ea4ac6 + 4d71264 in espresso 5

src/core/bonded_interactions/bonded_coulomb.hpp
```diff
@@ -44,34 +44,30 @@ int bonded_coulomb_set_params(int bond_type, double prefactor);
...
  *  @param[in]  iaparams  Interaction parameters.
  *  @param[in]  dx        %Distance between the particles.
- *  @param[out] force     Force.
- *  @retval false
  */
-inline bool calc_bonded_coulomb_pair_force(
-    double const q1q2, Bonded_ia_parameters const *const iaparams,
-    Utils::Vector3d const &dx, Utils::Vector3d &force) {
+inline boost::optional<Utils::Vector3d>
+calc_bonded_coulomb_pair_force(double const q1q2,
```
KaiSzuttor: why is there a ```calc``` before the force calculating function's names but not for the energy kernels? Maybe we should drop the prefix.

src/core/bonded_interactions/bonded_coulomb.hpp
```diff
@@ -44,34 +44,30 @@ int bonded_coulomb_set_params(int bond_type, double prefactor);
...
+                               Bonded_ia_parameters const *const iaparams,
```
KaiSzuttor: can't we pass the parameters by ```const &```?

src/core/bonded_interactions/bonded_coulomb_sr.hpp
```diff
@@ -46,39 +46,30 @@ int bonded_coulomb_sr_set_params(int bond_type, double q1q2);
...
- *  @param[out] _energy   Energy.
- *  @retval false
  */
-inline bool
-bonded_coulomb_sr_pair_energy(Particle const *const p1,
-                              Particle const *const p2,
-                              Bonded_ia_parameters const *const iaparams,
-                              Utils::Vector3d const &dx, double *_energy) {
+inline boost::optional<double> bonded_coulomb_sr_pair_energy(
+    Particle const *const p1, Particle const *const p2,
```
KaiSzuttor: can we pass the particles by ```const &```?

src/core/bonded_interactions/bonded_tab.hpp
```diff
@@ -89,19 +87,18 @@ inline bool calc_tab_bond_force(Bonded_ia_parameters const *const iaparams,
...
+inline boost::optional<double>
+tab_bond_energy(Bonded_ia_parameters const *const iaparams,
+                Utils::Vector3d const &dx) {
   auto const *tab_pot = iaparams->p.tab.pot;
   auto const dist = dx.norm();
 
   if (dist < tab_pot->cutoff()) {
-    *_energy = tab_pot->energy(dist);
-    return false;
+    auto const energy = tab_pot->energy(dist);
```
KaiSzuttor: just return here

src/core/bonded_interactions/bonded_tab.hpp
```diff
@@ -64,21 +64,19 @@ int tabulated_bonded_set_params(int bond_type,
...
+calc_tab_bond_force(Bonded_ia_parameters const *const iaparams,
+                    Utils::Vector3d const &dx) {
   auto const *tab_pot = iaparams->p.tab.pot;
   auto const dist = dx.norm();
 
   if (dist < tab_pot->cutoff()) {
     auto const fac = tab_pot->force(dist) / dist;
-    force = fac * dx;
-    return false;
+    auto const force = fac * dx;
```
KaiSzuttor: just return here, the variable ```force``` is not needed

src/core/bonded_interactions/bonded_tab.hpp
```diff
@@ -176,8 +166,8 @@ inline bool tab_angle_energy(Utils::Vector3d const &r_mid,
 #else
   auto const phi = acos(cos_phi);
 #endif
-  *_energy = iaparams->p.tab.pot->energy(phi);
-  return false;
+  auto const energy = iaparams->p.tab.pot->energy(phi);
```
KaiSzuttor: return here

src/core/bonded_interactions/bonded_tab.hpp
```diff
@@ -263,9 +242,8 @@ inline bool tab_dihedral_energy(Utils::Vector3d const &r1,
   auto const *tab_pot = iaparams->p.tab.pot;
   calc_dihedral_angle(r1, r2, r3, r4, v12, v23, v34, v12Xv23, &l_v12Xv23,
                       v23Xv34, &l_v23Xv34, &cos_phi, &phi);
-  *_energy = tab_pot->energy(phi);
-
-  return false;
+  auto const energy = tab_pot->energy(phi);
```
KaiSzuttor: return here

src/core/bonded_interactions/dihedral.hpp
```diff
@@ -197,12 +188,16 @@ dihedral_energy(Utils::Vector3d const &r1, Utils::Vector3d const &r2,
 
   calc_dihedral_angle(r1, r2, r3, r4, v12, v23, v34, v12Xv23, &l_v12Xv23,
                       v23Xv34, &l_v23Xv34, &cosphi, &phi);
+  /* dihedral angle not defined - force zero */
+  if (phi == -1.0) {
+    return {};
+  }
 
-  *_energy =
+  auto const energy =
```
KaiSzuttor: return here

src/core/bonded_interactions/fene.hpp
```diff
@@ -59,31 +57,29 @@ inline bool calc_fene_pair_force(Bonded_ia_parameters const *const iaparams,
     fac = 0.0;
   }
 
-  force = fac * dx;
+  auto const force = fac * dx;
```
KaiSzuttor: return here

src/core/bonded_interactions/fene.hpp
```diff
@@ -59,31 +57,29 @@ inline bool calc_fene_pair_force(Bonded_ia_parameters const *const iaparams,
...
   if (dr >= iaparams->p.fene.drmax) {
-    return true;
+    return {};
   }
 
-  auto energy = -0.5 * iaparams->p.fene.k * iaparams->p.fene.drmax2;
-  energy *= log((1.0 - dr * dr * iaparams->p.fene.drmax2i));
-  *_energy = energy;
-  return false;
+  auto const energy = -0.5 * iaparams->p.fene.k * iaparams->p.fene.drmax2 *
```
KaiSzuttor: return here

src/core/bonded_interactions/harmonic.hpp
```diff
@@ -60,29 +58,28 @@ inline bool calc_harmonic_pair_force(Bonded_ia_parameters const *const iaparams,
   } else {
     fac = 0;
   }
-  force = fac * dx;
+  auto const force = fac * dx;
```
KaiSzuttor: return here

src/core/bonded_interactions/harmonic.hpp
```diff
@@ -60,29 +58,28 @@ inline bool calc_harmonic_pair_force(Bonded_ia_parameters const *const iaparams,
...
   auto const dist = dx.norm();
 
   if ((iaparams->p.harmonic.r_cut > 0.0) &&
       (dist > iaparams->p.harmonic.r_cut)) {
-    return true;
+    return {};
   }
 
-  *_energy =
+  auto const energy =
```
KaiSzuttor: return here

src/core/bonded_interactions/harmonic_dumbbell.hpp
```diff
@@ -46,60 +46,56 @@ int harmonic_dumbbell_set_params(int bond_type, double k1, double k2, double r,
...
   auto const dhat = dx / dist;
   auto const da = vector_product(dhat, director);
   auto const torque = iaparams->p.harmonic_dumbbell.k2 * da;
   auto const diff = dhat - director;
 
-  *_energy = 0.5 * iaparams->p.harmonic_dumbbell.k1 *
-                 Utils::sqr(dist - iaparams->p.harmonic.r) +
-             0.5 * iaparams->p.harmonic_dumbbell.k2 * (torque * diff);
-  return false;
+  auto const energy = 0.5 * iaparams->p.harmonic_dumbbell.k1 *
```
KaiSzuttor: return here

src/core/bonded_interactions/quartic.hpp
```diff
@@ -42,46 +42,42 @@ int quartic_set_params(int bond_type, double k0, double k1, double r,
...
   }
 
   auto const dr = dist - iaparams->p.quartic.r;
   auto const fac =
       (iaparams->p.quartic.k0 * dr + iaparams->p.quartic.k1 * dr * dr * dr) /
       dist;
-  force = -fac * dx;
-
-  return false;
+  auto const force = -fac * dx;
```
KaiSzuttor: return here

src/core/bonded_interactions/quartic.hpp
```diff
@@ -42,46 +42,42 @@ int quartic_set_params(int bond_type, double k0, double k1, double r,
...
-    return true;
+    return {};
   }
 
   double dr2 = Utils::sqr(dist - iaparams->p.quartic.r);
 
-  *_energy = 0.5 * iaparams->p.quartic.k0 * dr2 +
-             0.25 * iaparams->p.quartic.k1 * Utils::sqr(dr2);
-  return false;
+  auto const energy = 0.5 * iaparams->p.quartic.k0 * dr2 +
```
KaiSzuttor: return here

src/core/bonded_interactions/subt_lj.hpp
```diff
@@ -43,28 +43,23 @@ int subt_lj_set_params(int bond_type);
...
-                                    Utils::Vector3d const &dx,
-                                    Utils::Vector3d &force) {
-  auto const neg_dir = -dx;
-  add_lj_pair_force(iaparams, neg_dir, neg_dir.norm(), force);
-
-  return false;
+inline boost::optional<Utils::Vector3d>
+calc_subt_lj_pair_force(IA_parameters const *const iaparams,
+                        Utils::Vector3d const &dx) {
+  auto const force = -lj_pair_force(iaparams, dx, dx.norm());
```
KaiSzuttor: return here

src/core/bonded_interactions/subt_lj.hpp
```diff
@@ -43,28 +43,23 @@ int subt_lj_set_params(int bond_type);
...
- *  @retval false
  */
-inline bool subt_lj_pair_energy(IA_parameters const *const iaparams,
-                                Utils::Vector3d const &dx, double *_energy) {
-  *_energy = -lj_pair_energy(iaparams, dx.norm());
-  return false;
+inline boost::optional<double>
+subt_lj_pair_energy(IA_parameters const *const iaparams,
+                    Utils::Vector3d const &dx) {
+  auto const energy = -lj_pair_energy(iaparams, dx.norm());
```
KaiSzuttor: return here

src/core/bonded_interactions/umbrella.hpp
```diff
@@ -48,31 +48,29 @@ inline double umbrella_force_r(double k, int dir, double r, double distn) {
...
-inline bool umbrella_pair_energy(Bonded_ia_parameters const *const ia_params,
-                                 Utils::Vector3d const &d, double *_energy) {
+inline boost::optional<double>
+umbrella_pair_energy(Bonded_ia_parameters const *const ia_params,
+                     Utils::Vector3d const &d, double *_energy) {
   auto const distn = d[ia_params->p.umbrella.dir];
-  *_energy = 0.5 * ia_params->p.umbrella.k *
-             Utils::sqr(distn - ia_params->p.umbrella.r);
-  return false;
+  auto const energy = 0.5 * ia_params->p.umbrella.k *
```
KaiSzuttor: return here

src/core/electrostatics_magnetostatics/coulomb_inline.hpp
```diff
@@ -52,30 +52,34 @@ inline Utils::Vector3d central_force(double const q1q2,
...
 
-  if (q1q2 == 0)
-    return;
+  if (q1q2 == 0) {
+    return std::make_tuple(Utils::Vector3d{}, Utils::Vector3d{},
+                           Utils::Vector3d{});
+  }
 
-  force += central_force(q1q2, d, dist);
+  Utils::Vector3d force = central_force(q1q2, d, dist);
```
KaiSzuttor: ```auto const force```

src/core/electrostatics_magnetostatics/p3m-dipolar.hpp
```diff
@@ -221,11 +221,12 @@ void dp3m_shrink_wrap_dipole_grid(int n_dipoles);
 /** Calculate real space contribution of p3m dipolar pair forces and torques.
  *  If NPT is compiled in, it returns the energy, which is needed for NPT.
  */
-inline double dp3m_add_pair_force(Particle *p1, Particle *p2,
-                                  Utils::Vector3d const &d, double dist2,
-                                  double dist, Utils::Vector3d &force) {
+inline std::tuple<double, Utils::Vector3d, Utils::Vector3d, Utils::Vector3d>
+dp3m_add_pair_force(Particle const *const p1, Particle const *const p2,
+                    Utils::Vector3d const &d, double dist2, double dist) {
```
mkuron: Function needs to be renamed from `add` to `calc`.

src/core/electrostatics_magnetostatics/p3m-dipolar.hpp
```diff
@@ -268,19 +270,23 @@ inline double dp3m_add_pair_force(Particle *p1, Particle *p2,
...
 #endif
-    return fac * (mimj * B_r - mir * mjr * C_r);
+    auto const _energy = fac * (mimj * B_r - mir * mjr * C_r);
+#else
+    auto const _energy = 0.0;
 #endif
+    return std::make_tuple(_energy, force, torque1, torque2);
   }
-  return 0.0;
+  return std::make_tuple(0.0, Utils::Vector3d{}, Utils::Vector3d{},
```
KaiSzuttor, line 288: why is there sometimes an energy returned and sometimes not and why is it ever returned if it's not used anywhere in the code?

src/core/electrostatics_magnetostatics/dipole_inline.hpp
```diff
@@ -8,33 +8,36 @@
...
 }
 
 // energy_inline
-inline void add_pair_energy(Particle const *const p1, Particle const *const p2,
-                            Utils::Vector3d const &d, double dist, double dist2,
-                            Observable_stat &energy) {
+inline double add_pair_energy(Particle const *const p1,
+                              Particle const *const p2,
+                              Utils::Vector3d const &d, double dist,
+                              double dist2) {
```
mkuron: Again, this function now calculates instead of adding.

src/core/electrostatics_magnetostatics/dipole_inline.hpp
```diff
@@ -8,33 +8,36 @@
...
+calc_pair_force(Particle const *const p1, Particle const *const p2,
+                Utils::Vector3d const &d, double dist, double dist2) {
+  Utils::Vector3d force{}, torque1{}, torque2{};
   switch (dipole.method) {
 #ifdef DP3M
   case DIPOLAR_MDLC_P3M:
     // fall trough
   case DIPOLAR_P3M: {
+    double eng;
+    std::tie(eng, force, torque1, torque2) =
```
KaiSzuttor, line 21: if there was not this energy part in the returned tuple we could just do ```return dp3m_add_pair_force(p1, p2, d, dist2, dist)```

KaiSzuttor (changes requested):

KaiSzuttor:
could you also please merge the python branch?

src/core/electrostatics_magnetostatics/icc.cpp
```diff
@@ -72,11 +72,14 @@ void force_calc_iccp3m(const ParticleRange &particles,
...
+  auto forces = Coulomb::calc_pair_force(p1, p2, d, dist);
 
-  p1->f.f += force;
-  p2->f.f -= force;
+  p1->f.f += std::get<0>(forces);
+  p2->f.f -= std::get<0>(forces);
+#ifdef P3M
+  p1->f.f += std::get<1>(forces);
+  p2->f.f += std::get<2>(forces);
+#endif
```
mkuron, line 80: This still modifies the particle. And I think we also add the same force down in forces_inline.hpp.

src/core/electrostatics_magnetostatics/magnetic_non_p3m_methods.cpp
```diff
@@ -70,51 +54,29 @@ double calc_dipole_dipole_ia(Particle *p1, const Utils::Vector3d &dip1,
...
-    p2->f.f[1] -= dipole.prefactor * ffy;
-    p2->f.f[2] -= dipole.prefactor * ffz;
-
-// Torques
+    auto const dd = pe4 * pe2;
+
+    // Forces
+    auto const ff = ab * dr + cc * dip1 + dd * dip2;
+    p1->f.f += dipole.prefactor * ff;
+    p2->f.f -= dipole.prefactor * ff;
```
mkuron: Again, this also modifies the particle.

mkuron (approved):
Looks ok to me. Note however that I did not match up every single `p->f.f += ...` from before with where it is now though. Those I did look at matched up, so I think this is reasonably safe to merge.

src/core/electrostatics_magnetostatics/p3m-dipolar.hpp
KaiSzuttor: i'd suggest to drop both: ```dp3m_pair_force```

src/core/bonded_interactions/bonded_coulomb.hpp
jngrad: I've now removed the prefix in force/energy kernels

src/core/bonded_interactions/bonded_coulomb.hpp
jngrad: Yes, and this made the integration loop a lot more homogeneous. Only the cell system and the principal loop of files `*_inline.hpp` use Particle pointers now, because in these parts of the code we don't always know in advance if a particle exists.

src/core/bonded_interactions/bonded_tab.hpp
jngrad: Done

src/core/electrostatics_magnetostatics/p3m-dipolar.hpp
jngrad: I don't know why NPT requires energies during the calculation of forces. More worrisome here is the code duplication of the energy calculation in `dp3m_add_pair_force()` and `dp3m_add_pair_energy()`, the latter containing comments ([1](https://github.com/espressomd/espresso/blob/54bbb885120e166c0a6945ef196bcacabf10a971/src/core/electrostatics_magnetostatics/p3m-dipolar.hpp#L311), [2](https://github.com/espressomd/espresso/blob/54bbb885120e166c0a6945ef196bcacabf10a971/src/core/electrostatics_magnetostatics/p3m-dipolar.hpp#L336-L337)) suggesting the energy formula used in the former ([3](https://github.com/espressomd/espresso/blob/54bbb885120e166c0a6945ef196bcacabf10a971/src/core/electrostatics_magnetostatics/p3m-dipolar.hpp#L280-L282)) is wrong.

src/core/electrostatics_magnetostatics/dipole_inline.hpp
jngrad: that's inelegant for sure, but we need the energy for NPT on the next line, and reducing the use of output parameters is our goal. I don't think there is a performance impact (https://godbolt.org/z/drOVeJ).

src/core/electrostatics_magnetostatics/p3m-dipolar.hpp
jngrad: Thanks! I've now removed the `add` part.

src/core/electrostatics_magnetostatics/dipole_inline.hpp
jngrad: I've now removed the prefix.

src/core/electrostatics_magnetostatics/icc.cpp
jngrad: True, it's not a pure function, but it's also not a force kernel. It's only a helper function that is not visible outside this file. The forces here are obtained from `Coulomb::pair_force(p1, p2, d, dist);`, which is indeed called in `forces_inline.hpp`. I might be wrong, but I think this helper function is used by [`iccp3m_iteration()`](https://github.com/espressomd/espresso/blob/54bbb885120e166c0a6945ef196bcacabf10a971/src/core/electrostatics_magnetostatics/icc.cpp#L94) to progressively update virtual charges while making sure the resulting particle force doesn't cancel the external electric field. Unfortunately this function isn't documented.

src/core/electrostatics_magnetostatics/magnetic_non_p3m_methods.cpp
jngrad: Same remark as above.

RudolfWeeber:
>I don't know why NPT requires energies during the calculation of forces. 

The dipolar pressure is proportional to energy / volume.

 

I suggest not to defer this PR for NPT-related reasons.That needs to be refactored, anyway, which is a major project in itself.

The dipolar isotropic NPT-pressure may well be wrong. I don’t think, it’s tested.

KaiSzuttor (approved):

PR #3078, opened 2019-08-12
Detect unknown features in the testsuite

jngrad:
Long-term solution to #3076: `@utx.skipIfMissingFeatures()` will throw an exception on unknown features.

### DIFF
--- a/src/config/check_myconfig.py
+++ b/src/config/check_myconfig.py
@@ -14,7 +14,6 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
-from sys import argv
 import sys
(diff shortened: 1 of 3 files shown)


fweik:
I'm reasonably sure that we already have code for this.

jngrad:
There's something similar in [check_myconfig.py](https://github.com/espressomd/espresso/blob/python/src/config/check_myconfig.py), but it's only called during compilation.

fweik:
`has_features` throws on unknown features, so it used to be checked.
(Directly above in __init__.py)

fweik (approved):

jngrad:
bors r=fweik

jngrad:
bors retry

PR #3080, opened 2019-08-13
testsuite: add test for is_valid_type.

KaiSzuttor:
Currently it is possible to give ```nan``` as values to the particle properties because in python a ```nan``` is of type ```float```. This PR adds a check for ```None```, ```inf``` and ```nan``` to the ```is_valid_type``` and tests for the correct behavior.


testsuite/python/utils.py
```diff
@@ -0,0 +1,20 @@
+import numpy as np
+
+import unittest as ut
+import espressomd.utils
+
+
+class UtilsTest(ut.TestCase):
```
jngrad: this file should be in `CMakeLists.txt` :-)

KaiSzuttor: ah right

jngrad (approved):

fweik:
`is_valid_type` is a horrible name for the function. It checks values, and not types, also it returns false for most valid types. Regarding the value, `inf` is a valid value that should be settable.

KaiSzuttor:
> Regarding the value, inf is a valid value that should be settable

why?

fweik:
Because you don't know if `inf` is a valid value for all  properties of all methods present and future.
In many cases it even behaves as expected out of the box. E.g. you could set `mass` to infinity, which
would fix the particle. `gamma` infinity could mean tracer coupling, ... Also I don't see a downside. Why do you want to exclude that?

src/python/espressomd/particle_data.pyx
```diff
@@ -151,6 +151,8 @@ cdef class ParticleHandle:
 
         def __set__(self, _pos):
             cdef double mypos[3]
+            if np.isnan(pos).any() or np.isinf(pos).any():
```
fweik: Why do you want to check that? I'm not sure that I follow. What scenario is this check for?

KaiSzuttor: it seems that the callbacks do not handle ```nan```. Please try:
``` python
import espressomd

system = espressomd.System(box_l=[10.0]*3)
system.part.add(pos=[float('nan')]*3)     

```

KaiSzuttor: since ```pos``` is the only particle property where the target node depends on the value of the property we should validate before calling the callback.

KaiSzuttor:
bors r=jngrad

jngrad:
bors r-
the webhook failed again

Issue #3037, opened 2019-07-31
Turn interaction parameter setters into constructor and separate storage selection from setter


fweik:
Depends on #3036 for non-bonded, #3019 for bonded.

Currently the functions setting the interaction parameters also select on which
interaction struct they operate, e.g. they take the types and the use globals to
access the interaction parameters struct, same for bonds where the global data
is accessed by bond id. The data selection can be separated from the parameter
setting to get better separation of concerns and reusability. E.g. this would allow
to reuse the same interaction implementation as bonded and non-bonded IA.

This can work by turning the set_parameters function into a constructor for the
corresponding type and let the caller decide what data to operate on. E.g. for
the harmonic bond, which currently looks like this:
```c++
int harmonic_set_params(int bond_type, double k, double r, double r_cut) {
  if (bond_type < 0)
    return ES_ERROR;

  make_bond_type_exist(bond_type);

  bonded_ia_params[bond_type].p.harmonic.k = k;
  bonded_ia_params[bond_type].p.harmonic.r = r;
  bonded_ia_params[bond_type].p.harmonic.r_cut = r_cut;
  bonded_ia_params[bond_type].type = BONDED_IA_HARMONIC;
  bonded_ia_params[bond_type].num = 1;

  /* broadcast interaction parameters */
  mpi_bcast_ia_params(bond_type, -1);

  return ES_OK;
}
```

could be turned into
```c++
struct Harmonic {
  Harmonic(double k, double r, double r_cut);
};
```
and there could common function for all bonds that
does the data selection and broadcasting like
```c++
template<class Bond>
void set_bond_params(int bond_type, Bond bond) {
  make_bond_type_exist(bond_type);
  bonded_ia_params.at(bond_type).p = bond;

  /* broadcast interaction parameters */
  mpi_bcast_ia_params(bond_type, -1);
}
```

jngrad:
Bonded and non-bonded interactions are now implemented as shared pointers in the core.

PR #3081, opened 2019-08-13
New logic for default parameters in IA classes

### LINKED ISSUE #3069: Rethink set_default_params()
The bonded IA classes in Python have a method `set_default_params()` that is _supposed_ to set default values. It is unclear to me what this method is actually doing in the current state, because:
* it's followed by a call to `_params.update()` that will override them by those stored in the core
  https://github.com/espressomd/espresso/blob/c2e8604c4d914b70616133be32989f9203baedaf/src/python/espressomd/interactions.pyx#L1904-L1907
* it's often setting a default value to required parameters, which by definition can't have a default
  https://github.com/espressomd/espresso/blob/c2e8604c4d914b70616133be32989f9203baedaf/src/python/espressomd/interactions.pyx#L2125-L2135
* they don't always provide sensible values, like this angle potential where an angle of 0 or pi would lead to NaN values in the force functions because of the division by `sin(phi) = 0`:
  https://github.com/espressomd/espresso/blob/c2e8604c4d914b70616133be32989f9203baedaf/src/python/espressomd/interactions.pyx#L2975-L2979
* one of them has a typo
  https://github.com/espressomd/espresso/blob/c2e8604c4d914b70616133be32989f9203baedaf/src/python/espressomd/interactions.pyx#L2549-L2553
  which is really confusing:
  ```python
  di = espressomd.interactions.Dihedral(mult=2, bend=5, phase=0)
  # mult should be 2, but it shows 1 on the left
  di.params
  ```
  ```
  {"mult'": 1, 'bend': 5, 'phase': 0, 'mult': 2}
  ```

---

This method should return a dict of sensible default values, only be called once in the class constructor, and its keys should be checked for typos. To do all at once, one could replace this line:
https://github.com/espressomd/espresso/blob/c2e8604c4d914b70616133be32989f9203baedaf/src/python/espressomd/interactions.pyx#L1862
with e.g.:
```python
params = get_default_params()
params.update(kwargs)
self.params = params
```

jngrad:
Fixes #3069

Change internals of bonded/non-bonded IA Python classes:
- `valid_keys()` and `required_keys()` now return sets instead of tuples/lists/strings (strings can lead to cryptic error messages since they are iterable)
- `default_params()` and `set_default_params()` now return values only for optional parameters
- wrong keys (typos or unknown keys) in optional parameters were fixed
- added tests for a few common IAs to check the new syntax and verify default parameters have the correct name and don't overlap with required parameters, which can't have default values

### DIFF
--- a/src/python/espressomd/interactions.pyx
+++ b/src/python/espressomd/interactions.pyx
@@ -295,13 +295,7 @@ IF LENNARD_JONES == 1:
             """Python dictionary of default parameters.
 
             """
-            return {
-                "epsilon": 0.,
-                "sigma": 0.,
-                "cutoff": 0.,
-                "shift": 0.,
-                "offset": 0.,
-                "min": 0.}
+            return {"offset": 0., "min": 0.}
 
         def type_name(self):
             """Name of interaction type.
@@ -380,9 +374,7 @@ IF WCA == 1:
             """Python dictionary of default parameters.
 
             """
-            return {
-                "epsilon": 0.,
-                "sigma": 0.}
+            return {}
 
         def type_name(self):
             """Name of interaction type.
@@ -487,18 +479,7 @@ IF LENNARD_JONES_GENERIC == 1:
             """Python dictionary of default parameters.
 
             """
-            return {
-                "epsilon": 0.,
-                "sigma": 0.,
-                "cutoff": 0.,
-                "shift": 0.,
-                "offset": 0.,
-                "e1": 0,
-                "e2": 0,
-                "b1": 0.,
-                "b2": 0.,
-                "delta": 0.,
-                "lam": 0.}
+            return {"delta": 0., "lam": 0.}
 
         def type_name(self):
             """Name of interaction type.
@@ -592,7 +573,7 @@ IF LJCOS:
                 Interaction length scale.
             cutoff : :obj:`float`
                 Cutoff distance of the interaction.
-            offset : :obj:`float`
+            offset : :obj:`float`, optional
                 Offset distance of the interaction.
             """
             super().set_params(**kwargs)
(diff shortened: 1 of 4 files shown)


src/python/espressomd/interactions.pyx
```diff
@@ -487,18 +479,7 @@ IF LENNARD_JONES_GENERIC == 1:
...
-                "cutoff": 0.,
-                "shift": 0.,
-                "offset": 0.,
-                "e1": 0,
-                "e2": 0,
-                "b1": 0.,
-                "b2": 0.,
-                "delta": 0.,
-                "lam": 0.}
+            return {"delta": 0., "lam": 0.}
```
KaiSzuttor: why are only those two parameters default ones? Shouldn't we reflect the facts from the core here?

jngrad: * `lam` is now 1 by default to reflect the core
* this line is now guarded by `LJGEN_SOFTCORE`
* keeping them as optional allows this generic LJ class to behave the same way when `LJGEN_SOFTCORE` is enabled, unless the user provides custom values for `lam` and `delta`

jngrad:
@KaiSzuttor is anything missing?

KaiSzuttor (approved):
LGTM in the future we probably want to have constructors for the interactions in the core. Or at least on the python level.

jngrad:
Agreed. The Python class should only take care of checking variable types, the core should take care of checking values. Core constructors are planned in #3037.

jngrad:
bors r=KaiSzuttor

jngrad:
bors cancel

jngrad:
bors r=KaiSzuttor

PR #3082, opened 2019-08-14
Re-enable CMake option WITH_PYTHON

jngrad:
When looking for errors in the core with Clangs' scan-build tool, it is convenient to disable Cython, whose generated C++ code triggers many warnings that are outside our control. With this change, scan-build works out-of-the-box with our CMake logic:

```bash
# all reports
scan-build cmake .. -DWITH_CUDA=OFF
# skip Cython reports
scan-build cmake .. -DWITH_CUDA=OFF -DWITH_PYTHON=OFF
# build
scan-build -k -o static-analysis/ make -j$(nproc)
scan-build -k -o static-analysis/ make -j$(nproc) check_unit_tests
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,8 +56,7 @@ set(CMAKE_FIND_FRAMEWORK LAST)
 # User input options
 ######################################################################
 
-set(WITH_PYTHON yes)
-
+option(WITH_PYTHON "Build with Python bindings"  ON)
 option(WITH_GSL    "Build with GSL support"  ON)
 option(WITH_CUDA   "Build with GPU support"  ON)
 option(WITH_HDF5   "Build with HDF5 support" ON)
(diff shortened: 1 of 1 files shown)


fweik:
Couldn't you just do `make EspressoCore` ?

jngrad:
Yes, that gives the same output. But target `EspressoCore` is defined two levels down in the CMake logic (`src/core/CMakeLists.txt`), so you first need to find it, and also be familiar enough with our CMake logic to understand which other directories in `src/*` will also be visited by the static analyzer. In the snippet above, scan-build blends in the CMake command seamlessly.

mkuron (approved):

KaiSzuttor (approved):

jngrad:
bors r=KaiSzuttor

PR #3085, opened 2019-08-15
PersistenceAngles observable

KaiSzuttor:
New observable plus test.


src/core/observables/ChainAngles.hpp
```diff
@@ -0,0 +1,66 @@
...
+#include <utils/Vector.hpp>
+
+#include <cmath>
+#include <vector>
+
+namespace Observables {
+
+/** Calculate bond angles between particles in a polymer.
+ *  For @f$ n @f$ bonded particles, return the @f$ n-2 @f$ angles along the
+ *  chain, in radians.
```
jngrad: copy-paste error: you're actually calculating `n-2` angles averages between particles that are `n`-bond apart

src/python/espressomd/observables.py
```diff
@@ -388,6 +388,23 @@ class ParticleAngles(Observable):
     _so_name = "Observables::ParticleAngles"
 
 
+@script_interface_register
+class ChainAngles(Observable):
+
+    """Calculates mutual bond angles for chained particles with given ids.
+
+    The *i*-th  value of the result contains the angle between bonds that
+    are separated by *i* bonds. The angles are average over the chain.
```
jngrad:
```suggestion
    are separated by *i* bonds. The angles are averaged over the chain.
```

testsuite/python/observable_chain.py
```diff
@@ -181,5 +196,11 @@ def place_particles(bl, offset):
                         res_obs_chain, [dih1, dih2], decimal=9,
                         err_msg="Data did not agree for observable ParticleDihedrals")
 
+    def test_chain_angles(self):
+        self.system.part.clear()
+        self.system.part.add(pos= np.array([np.linspace(0, self.system.box_l[0], 20)] * 3).T + np.random.random((20, 3)))
```
jngrad, line 204: This test alters the state of the class. If this test starts before the two others, they could fail. From [unittest/Organizing test code](https://docs.python.org/3/library/unittest.html#organizing-test-code):
> The order in which the various tests will be run is determined by sorting the test method names with respect to the built-in ordering for strings.

This can happen if someone renames the method to `test_ChainAngles` for consistency with the other methods in this class. I tried it and it still passed, because the original test methods work on particle slices. I think it would be safer if you didn't create particles in your test, increased `n_parts` to 20 and only updated the positions of these `n_parts` particles. Alternatively, you could move the creation of particles from the `setUpClass()` to the individual test methods, and call `self.system.part.clear()` in a `tearDown()`.

src/core/observables/ChainAngles.hpp
```diff
@@ -0,0 +1,66 @@
...
+    std::vector<Utils::Vector3d> bond_vectors(no_of_bonds);
+    auto get_bond_vector = [&](auto index) {
+      return get_mi_vector(partCfg[ids()[index + 1]].r.p,
+                           partCfg[ids()[index]].r.p, box_geo);
+    };
+    for (int i = 0; i < no_of_bonds; ++i) {
+      auto const tmp = get_bond_vector(i);
+      bond_vectors[i] = tmp / tmp.norm();
+    }
+    // calculate angles between neighbouring bonds, next beighbours, etc...
```
jngrad:
```suggestion
    // calculate angles between neighbouring bonds, next neighbours, etc...
```

doc/sphinx/analysis.rst
```diff
@@ -710,6 +710,8 @@ The following observables are available:
      :math:`\frac{1}{\sum_i m_i} \left( \sum_i m_i v^x_i, \sum_i m_i v^y_i, \sum_i m_i v^z_i\right)`
    - ComForce: Sum of the forces on the particles
      :math:`\sum_i f^x_i, \sum_i f^y_i, \sum_i f^z_i`
+   - ChainAngles: Angles between bonds. The `i`-th value in the result vector corresponds to the angle between
+     bonds that are separated by `i` bonds.
```
jngrad: Wouldn't a name `PersistenceAngles` or something similar help differentiate this observable from `ParticleAngles`, which also calculates chain angles but without average?

jngrad (changes requested):

src/core/observables/PersistenceAngles.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+#include "PidObservable.hpp"
+#include <utils/Vector.hpp>
+
+#include <cmath>
+#include <vector>
+
+namespace Observables {
+
+/** Calculate bond angles between in a polymer.
```
jngrad:
```suggestion
/** Calculate bond angles between particles in a polymer.
```

src/core/observables/PersistenceAngles.hpp
```diff
@@ -0,0 +1,66 @@
...
+ *  For @f$ n @f$ bonded particles, return the @f$ n-2 @f$ angles along the
+ *  chain that are seperated by 1 bond, 2 bonds, ..., in radians.
```
jngrad:
```suggestion
 *  chain that are separated by 1 bond, 2 bonds, ..., in radians.
```
also, at index `i=0`, `i=1`, ..., the angle is for particles separated by 2 bonds, 3 bonds, ... (or equivalently for bonds separated by 0 bond, 1 bond, ..., as you wrote in analysis.rst); the current formulation seems ambiguous

src/script_interface/observables/initialize.cpp
```diff
@@ -137,8 +138,9 @@ void initialize() {
   REGISTER_PID_OBS(ComVelocity);
   REGISTER_PID_OBS(ComForce);
   REGISTER_PID_OBS(ParticleDistances);
-  REGISTER_PID_OBS(ParticleAngles);
+  REGISTER_PID_OBS(BondAngles);
   REGISTER_PID_OBS(ParticleDihedrals);
```
jngrad: Here we might as well rename `ParticleDistances -> BondLengths` and `ParticleDihedrals -> BondDihedrals`

jngrad (changes requested):

src/script_interface/observables/initialize.cpp
KaiSzuttor: I think ```ParticleDistances``` is the correct name. It doesn't need to be a bondlength you want to track.

KaiSzuttor:
@jngrad we had the same idea, we should avoid pushing at the same time to that PR

KaiSzuttor:
@jngrad I think we should not name it ```BondLengths``` as I commented above. Please open a separate PR for the renaming.

jngrad:
sorry! GitHub didn't show your answer in the review box

KaiSzuttor:
@jngrad could you please approve if your happy?

jngrad (approved):

KaiSzuttor:
bors r=jngrad

PR #3091, opened 2019-08-21
Tests and Bugfixes for the GB potential

### LINKED ISSUE #2922: Gay-Berne has low test coverage
The Gay-Bernen potential is not properly tested and documented.

Kischy:
Fixes #2922

Description of changes:
 - Added tests for the gb potential for the force and torque
 - Fixed bugs in gay_berne.hpp
 - Removed condition in gay_berne.hpp

PR Checklist
------------
 - [x] Tests?
   - [ ] Interface
   - [x] Core 
 - [ ] Docs?

### DIFF
--- a/testsuite/python/interactions_non-bonded.py
+++ b/testsuite/python/interactions_non-bonded.py
@@ -614,10 +614,7 @@ class InteractionsNonBondedTest(ut.TestCase):
             
         def rotate_part(particle):
             particle.rotate(axis=(1, 2, 3), angle=0.3)
-            particle.rotate(axis=(1, -2, -4), angle=1.2)
-                
-        def get_simulation_energy():
-            return self.system.analysis.energy()["non_bonded"]
+            particle.rotate(axis=(1, -2, -4), angle=1.2)                
         
         def get_reference_energy(gb_params, r, director1, director2):
             k_1, k_2, mu, nu, sigma_0, epsilon_0, cut = gb_params;            
@@ -647,6 +644,18 @@ class InteractionsNonBondedTest(ut.TestCase):
             
             return E_ref
         
+        def get_reference_force(gb_params, r, dir1, dir2):
+                
+            force_ref = numpy.zeros(3)        
+            
+            for i in range(3):
+                force_ref[i] = tests_common.calc_derivative(lambda x : get_reference_energy(gb_params, x, dir1, dir2),
+                                                             x=r,
+                                                             axis=i)
+            
+            
+            return force_ref
+            
         
         
         
@@ -663,7 +672,9 @@ class InteractionsNonBondedTest(ut.TestCase):
         setup_system(gb_params)        
         
         p1 = self.system.part[0]
-        p2 = self.system.part[1]     
+        p2 = self.system.part[1]   
+        
+        delta = 1.0e-7
         
         for i in range(111):
             
@@ -678,11 +689,25 @@ class InteractionsNonBondedTest(ut.TestCase):
             
             
             # Calc energies
-            E_sim = get_simulation_energy()
+            E_sim = self.system.analysis.energy()["non_bonded"]
             E_ref = get_reference_energy(gb_params, r, director1, director2)  
             
             # Test energies
-            self.assertAlmostEqual(E_sim, E_ref)   
+            self.assertAlmostEqual(E_sim, E_ref, delta=delta)   
+            
+            # Calc forces
+            f1_sim = p1.f
+            f2_sim = p2.f
+            
+            f1_ref = get_reference_force(gb_params, r, director1, director2)
+            
+            # Test forces
+            # force equals minus the counter-force 
+            self.assertTrue((f1_sim == -f2_sim).all())            
+            # compare force to reference force
+            self.assertAlmostEqual(f1_sim[0], f1_ref[0], delta=delta)
+            self.assertAlmostEqual(f1_sim[1], f1_ref[1], delta=delta)
+            self.assertAlmostEqual(f1_sim[2], f1_ref[2], delta=delta)
             
 
         
--- a/testsuite/python/interactions_non-bonded.py
+++ b/testsuite/python/interactions_non-bonded.py
@@ -668,14 +668,14 @@ class InteractionsNonBondedTest(ut.TestCase):
 
         def get_reference_force(gb_params, r, dir1, dir2):          
             force_ref = calc_partial_derivatives(
-                    lambda x: get_reference_energy(gb_params, x, dir1, dir2),
+                lambda x: get_reference_energy(gb_params, x, dir1, dir2),
                     x0=r, dx=1.0e-7)
 
             return -force_ref
 
         def get_reference_torque(gb_params, r, dir1, dir2):
             force_in_dir1 = calc_partial_derivatives(
-                    lambda x: get_reference_energy(gb_params, r, x, dir2),
+                lambda x: get_reference_energy(gb_params, r, x, dir2),
                     x0=dir1, dx=1.0e-7)
 
             return numpy.cross(-dir1, force_in_dir1)
--- a/testsuite/python/interactions_non-bonded.py
+++ b/testsuite/python/interactions_non-bonded.py
@@ -17,7 +17,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 import espressomd
-import numpy
+import numpy as np
 import unittest as ut
 import unittest_decorators as utx
 import tests_common
@@ -27,11 +27,11 @@ class InteractionsNonBondedTest(ut.TestCase):
     system = espressomd.System(box_l=[1.0, 1.0, 1.0])
     box_l = 10.
 
-    start_pos = numpy.random.rand(3) * box_l
-    axis = numpy.random.rand(3)
-    axis /= numpy.linalg.norm(axis)
+    start_pos = np.random.rand(3) * box_l
+    axis = np.random.rand(3)
+    axis /= np.linalg.norm(axis)
     step = axis * 0.01
-    step_width = numpy.linalg.norm(step)
+    step_width = np.linalg.norm(step)
 
     def setUp(self):
 
@@ -80,7 +80,7 @@ class InteractionsNonBondedTest(ut.TestCase):
             b1=lj_b1, b2=lj_b2, e1=lj_e1, e2=lj_e2, shift=lj_shift)
 
         E_ref = tests_common.lj_generic_potential(
-            r=numpy.arange(1, 232) * self.step_width, eps=lj_eps, sig=lj_sig,
+            r=np.arange(1, 232) * self.step_width, eps=lj_eps, sig=lj_sig,
             cutoff=lj_cut, offset=lj_off, b1=lj_b1, b2=lj_b2, e1=lj_e1,
             e2=lj_e2, shift=lj_shift)
 
@@ -122,7 +122,7 @@ class InteractionsNonBondedTest(ut.TestCase):
                                                           sigma=wca_sig)
 
         E_ref = tests_common.lj_generic_potential(
-            r=numpy.arange(1, 232) * self.step_width, eps=wca_eps, sig=wca_sig,
+            r=np.arange(1, 232) * self.step_width, eps=wca_eps, sig=wca_sig,
                 cutoff=wca_cutoff, shift=4. * wca_shift)
 
         for i in range(231):
@@ -613,9 +613,9 @@ class InteractionsNonBondedTest(ut.TestCase):
 
             """
             partial_x = lambda x: (func(x0 + x) - func(x0 - x)) / (
-                2.0 * numpy.linalg.norm(x))
-            delta = numpy.array([dx, 0.0, 0.0])
-            return numpy.array([partial_x(numpy.roll(delta, i)) for i in range(3)])
+                2.0 * np.linalg.norm(x))
+            delta = np.array([dx, 0.0, 0.0])
+            return np.array([partial_x(np.roll(delta, i)) for i in range(3)])
 
         def setup_system(gb_params):
             k_1, k_2, mu, nu, sigma_0, epsilon_0, cut = gb_params
@@ -640,7 +640,7 @@ class InteractionsNonBondedTest(ut.TestCase):
 
         def get_reference_energy(gb_params, r, director1, director2):
             k_1, k_2, mu, nu, sigma_0, epsilon_0, cut = gb_params
-            r_cut = r * cut / numpy.linalg.norm(r)
+            r_cut = r * cut / np.linalg.norm(r)
 
             E_ref = tests_common.gay_berne_potential(
                 r, director1, director2, epsilon_0, sigma_0, mu, nu, k_1, k_2)
@@ -660,7 +660,7 @@ class InteractionsNonBondedTest(ut.TestCase):
                 lambda x: get_reference_energy(gb_params, r, x, dir2),
                     x0=dir1, dx=1.0e-7)
 
-            return numpy.cross(-dir1, force_in_dir1)
+            return np.cross(-dir1, force_in_dir1)
 
         # actual tests of the gb potential
 
--- a/testsuite/python/interactions_non-bonded.py
+++ b/testsuite/python/interactions_non-bonded.py
@@ -588,32 +588,32 @@ class InteractionsNonBondedTest(ut.TestCase):
 
         self.system.non_bonded_inter[0, 0].gaussian.set_params(eps=0.)
         
-        
-    
-        
     # Test the gay berne potential and the resulting force and torque
     @utx.skipIfMissingFeatures("GAY_BERNE")
     def test_gb(self):    
         
-        
         # helper function definitions # 
         def setup_system(gb_params):        
-            k_1, k_2, mu, nu, sigma_0, epsilon_0, cut = gb_params;
+            k_1, k_2, mu, nu, sigma_0, epsilon_0, cut = gb_params
             
             self.system.part.clear()            
-            self.system.part.add(id=0, pos=(1,2,3), rotation=(1,1,1), type=0)
-            self.system.part.add(id=1, pos=(2.2,2.1,2.9), rotation=(1,1,1), type=0)
+            self.system.part.add(
+                id=0, pos=(1, 2, 3), rotation=(1, 1, 1), type=0)
+            self.system.part.add(
+                id=1, pos=(2.2, 2.1, 2.9), rotation=(1, 1, 1), type=0)
             
-            self.system.non_bonded_inter[0, 0].gay_berne.set_params(sig=sigma_0, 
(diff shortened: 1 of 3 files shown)


jngrad:
So far we've always used analytical solutions to calculate reference forces. Using a numerical approximation of the gradient is an interesting approach. I'm actually surprised that forces calculated with an integral step of 2e-7 can pass a `self.assertAlmostEqual` test at precision 1e-7. I've started the tests on non-amd64 architectures (pipeline [8879](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/pipelines/8879)) to see if there is a risk of numerical instability.

After re-organizing my version of the force function to match the nomenclature of the variables in the energy function, I still can't follow the force code. For example, in the calculation of epsilon, at line:
https://github.com/espressomd/espresso/blob/d79e9c14545827b07096a0e7af41227df48cb5e6/src/core/nonbonded_interactions/gay_berne.hpp#L70-L72
I would expect variable `E2` to be raised to the power of `mu - 1` in a subsequent variable to evaluate the derivative (d epsilon / d r), but I can't find it. It would be very helpful if we had the formula of the GB force, possibly with the intermediate steps to obtain it from the derivative of energy. We could write them with LaTeX in a Doxygen page, like we did for the [derivation of the bond angle force](http://espressomd.org/html/dox/bondedIA.html#bondedIA_angle_force).

Kischy:
> I've started the tests on non-amd64 architectures (pipeline 8879) to see if there is a risk of numerical instability.

I saw that two of them failed. This [one](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/152120) failed because the reaction_ensemble test had a timeout. The [second one](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/152124) failed because of different timeouts and fails, non of which was the interactions_non-bonded.py. Do you know why those failed?

> So far we've always used analytical solutions to calculate reference forces. Using a numerical approximation of the gradient is an interesting approach. I'm actually surprised that forces calculated with an integral step of 2e-7 can pass a self.assertAlmostEqual test at precision 1e-7.

I think this way testing derivatives can be very powerful, if the analytical solution is long or complicated. Although it is much slower to calculate the derivative this way, it is not so prone to error. To advertise the approach a little bit more: If I had just implemented the same analytical solution used in the c++ code for testing I would not have found the error. I borrowed the idea from machine learning, see [here](http://ufldl.stanford.edu/tutorial/supervised/DebuggingGradientChecking/) and [here](https://stackoverflow.com/questions/47506521/what-exactly-is-gradient-checking).

> I would expect variable E2 to be raised to the power of mu - 1 in a subsequent variable to evaluate the derivative (d epsilon / d r), but I can't find it. 

It is a little bit confusing. The change in power comes with this line: https://github.com/espressomd/espresso/blob/d79e9c14545827b07096a0e7af41227df48cb5e6/src/core/nonbonded_interactions/gay_berne.hpp#L77

And with this line:
https://github.com/espressomd/espresso/blob/d79e9c14545827b07096a0e7af41227df48cb5e6/src/core/nonbonded_interactions/gay_berne.hpp#L78

And also with these lines:
https://github.com/espressomd/espresso/blob/d79e9c14545827b07096a0e7af41227df48cb5e6/src/core/nonbonded_interactions/gay_berne.hpp#L89-L90

> It would be very helpful if we had the formula of the GB force, possibly with the intermediate steps to obtain it from the derivative of energy.

If you want I can send you my jupyter notebook (using python 3) or the python script, where I did the calculations of the derivative. 

>  We could write them with LaTeX in a Doxygen page, like we did for the derivation of the bond angle force.

The problem is that I do not know what paper the orginal author used, so I can't cite anything. I will try to work out a new nomenclatur for the problem and then we can put it in latex code, if you want.

jngrad:
The non-amd tests are known to sometimes false trigger, that's why we don't run them automatically. The gradient approach seems to work, though. Your [UFLDL link](http://ufldl.stanford.edu/tutorial/supervised/DebuggingGradientChecking/) states that an epsilon of 1e-4 will usually provide results at 4 digits precision, but in my experience a test is safer when the reference force is at least one order of magnitude more precise than the test threshold delta. Maybe @KaiSzuttor can comment on that.

The book or article used by the original author is unknown, all we know is this code is 16 years old. Given the notation in the energy function and the [LaTeX formula in the docs](http://espressomd.org/html/doc/inter_non-bonded.html#gay-berne-interaction), I would bet on [Cleaver *et al.* 1996](https://doi.org/10.1103/PhysRevE.54.559), equations 3-4 and 7-9.

Here is [my modified version](https://github.com/jngrad/espresso/blob/Bugfix_Tests_Gay_Berne_Issue_2922-nomencl/src/core/nonbonded_interactions/gay_berne.hpp) of the GB force function. Although it looks more verbose, when compiled with `-O3` it generates 23 less lines of assembly code in gcc and 13 less in Clang (9% resp. 6% less). We could probably reduce the assembly code further with more refactoring based on the analytical expression of the force. I've looked at several sources on the GB potential, but none of them give the expression of its derivative in Cartesian space. If you already worked out the derivation, you can share it here by attaching a file.

Kischy:
> The non-amd tests are known to sometimes false trigger, that's why we don't run them automatically. The gradient approach seems to work, though. Your UFLDL link states that an epsilon of 1e-4 will usually provide results at 4 digits precision, but in my experience a test is safer when the reference force is at least one order of magnitude more precise than the test threshold delta. Maybe @KaiSzuttor can comment on that.

Okay, I will change the epsilon of my force calculation and update the pull request.

> Here is my modified version of the GB force function. Although it looks more verbose, when compiled with -O3 it generates 23 less lines of assembly code in gcc and 13 less in Clang (9% resp. 6% less). We could probably reduce the assembly code further with more refactoring based on the analytical expression of the force. I've looked at several sources on the GB potential, but none of them give the expression of its derivative in Cartesian space. If you already worked out the derivation, you can share it here by attaching a file.

I worked on a better nomenclature just now too. If it is okay with you I would combine my version with yours and we see how it fits. _What do you think?_

Kischy:
I can't attach the files here, because the file type is not supported by github. I will send them to you by mail.

Kischy:
> Here is my modified version of the GB force function.

I think you're version is the best we can do. Can you just add the changed version to my pull request or should I do so?

Kischy:
> Your UFLDL link states that an epsilon of 1e-4 will usually provide results at 4 digits precision, but in my experience a test is safer when the reference force is at least one order of magnitude more precise than the test threshold delta. 

I added a commit that changes the delta value to 1x10E-6. For some reason it does not show here yet.

jngrad:
You can use my version. The simplest way to do it is to merge my branch into your branch (this is a trivial fast-forward), then add the new commit that changes the precision. Doing it the other way around (adding the precision commit, then merging my branch) would require a complex merge commit.

Also, I would not change delta to a bigger value, instead I would change epsilon to 1x10E-8.

jngrad:
You probably forgot to push the new commit to the GitHub repository. If it's ok with you, I can add my new version into your branch on GitHub. Then on you computer, you'll have to discard the precision commit (for example with `git reset --soft HEAD^` to rewind the branch history to d79e9c1), then pull your GitHub branch into your local repository.

Kischy:
No, it is already pushed. I'll merge your commit, change the delta and epsilon and push again.

Kischy:
> Also, I would not change delta to a bigger value, instead I would change epsilon to 1x10E-8.

The value of 1x10E-8 does not work, probably because of the numerical errors mentioned in the [UFLDL link ](http://ufldl.stanford.edu/tutorial/supervised/DebuggingGradientChecking/). I think 1x10E-7 is the lowest we can go here. Therefore I would recommend changing the delta value to 1x10E-6. Is that fine with you or not?

testsuite/python/tests_common.py
```diff
@@ -596,6 +596,20 @@ def gay_berne_potential(r_ij, u_i, u_j, epsilon_0, sigma_0, mu, nu, k_1, k_2):
     return 4. * epsilon * (rr**-12 - rr**-6)
 
 
+def calc_derivative(func, x, eps=1.0e-7, index=None):
```
KaiSzuttor: could you please use [```scipy.misc.derivative```](https://docs.scipy.org/doc/scipy/reference/generated/scipy.misc.derivative.html). If that does not suit your needs you should definitely add a docstring to this function. Actually we could even use [```sympy```](https://www.sympy.org/en/index.html). But we probably don't want to depend on that.

jngrad:
I've just discussed it with Kai, and we're both fine with a larger delta.

Kischy:
> 
> 
> I've just discussed it with Kai, and we're both fine with a larger delta.

Thanks for the update.

testsuite/python/tests_common.py
Kischy: I will add a docstring. [scipy.misc.derivative](https://docs.scipy.org/doc/scipy/reference/generated/scipy.misc.derivative.html) does not allow for a multidimensional input.

KaiSzuttor: https://docs.scipy.org/doc/numpy/reference/generated/numpy.vectorize.html

Kischy: Does that help? I need to manipulate the element at a specific index while the rest stays the same.

Kischy: I added a docstring. I could have thrown an error if someone gives a multidimensional input and index=None, but I think it would be a "overkill". In the docstring it is stated, that the index needs to be specified if x is an array.

Kischy:
Everything should be okay now. Do you need something else concerning issue #2922?

jngrad:
Note: it's unnecessary to merge python into this branch in the absence of merge conflicts, we have a bot to do that during the merge.

Kischy:
> 
> 
> Note: it's unnecessary to merge python into this branch in the absence of merge conflicts, we have a bot to do that during the merge.

Ok, sorry about that.

KaiSzuttor:
I think the confusion arises from the fact that the ```tests_common.gay_berne_potential``` returns an array and not a scalar as one would expect from a potential. I think we should not try to create a special derivative function that handles this case.

jngrad:
I'm also confused by the derivative function, in that it's actually a partial derivative function that expects a single N-dimensional coordinate, while scipy.misc.derivative expects a N-dimensional array of points scattered on the x-axis. The need for a partial derivative function comes from the multi-dimensional nature of the input values for the GB energy kernel. I'm currently rewriting the derivative function. I couldn't find an equivalent function in numpy and scipy.

KaiSzuttor:
can you explain why the potential is not a scalar?

Kischy:
The potential is scalar.

jngrad:
The potential is a scalar (linalg.norm without axis returns a scalar):
https://github.com/espressomd/espresso/blob/a910bab0e1c64e4fb443e76af546e4570253bb39/testsuite/python/tests_common.py#L594-L596
The input however isn't.

Kischy:
> I'm also confused by the derivative function, in that it's actually a partial derivative function that expects a single N-dimensional coordinate, while scipy.misc.derivative expects a N-dimensional array of points scattered on the x-axis. The need for a partial derivative function comes from the multi-dimensional nature of the input values for the GB energy kernel. I'm currently rewriting the derivative function. I couldn't find an equivalent function in numpy and scipy.

You are right. It is a partial derivative function. Did you already start the rewriting? I did the same thing half an hour ago.

KaiSzuttor:
this is not a scalar

https://github.com/espressomd/espresso/blob/a910bab0e1c64e4fb443e76af546e4570253bb39/testsuite/python/tests_common.py#L588

right?

Kischy:
> 
> 
> this is not a scalar
> 
> https://github.com/espressomd/espresso/blob/a910bab0e1c64e4fb443e76af546e4570253bb39/testsuite/python/tests_common.py#L588
> 
> right?

No, this gives a scalar for epsilon.

KaiSzuttor:
maybe we should save developer time and talk before code :smile:

Kischy:
Yeah, thats a good idea. :D

jngrad:
I'll upload my version of the partial derivative in a separate branch.

Kischy:
Okay, than I will upload mine here and you can pick which one you like better.

jngrad:
Here is my version: [calc_partial_derivative()](https://github.com/jngrad/espresso/blob/9cac5045d450d27d6f6127245ce8443f19c7138e/testsuite/python/tests_common.py#L599)

Kischy:
> 
> 
> Here is my version: [calc_partial_derivative()](https://github.com/jngrad/espresso/blob/9cac5045d450d27d6f6127245ce8443f19c7138e/testsuite/python/tests_common.py#L599)

Okay, I think a combination of our versions might be best. Mine calculates all the derivatives in one go. _What do you think?_ (Before I start coding again :D)
Nice Docstring by the way.

jngrad:
What do you mean? The unit test hasn't changed: [9cac504::interactions_non-bonded.py](https://github.com/jngrad/espresso/commit/9cac5045d450d27d6f6127245ce8443f19c7138e#diff-8080230a7bbebfde7eab64f61e78f499). The new version can also compute the derivative over two dimensions at once, e.g. `f(x,y)=(x^2+y^2)/2`, gradient: `df(x,y)=x+y`, implementation: `calc_partial_derivative(lambda x: 0.5*x[0]**2 + 0.5*x[1]**2, [1,1], variables=[0,1])` returns 2.0

Kischy:
I changed the test functions and put a for loop into the function to calculate all the partial derivatives of the function at point x0 at once.

Kischy:
It's just another implementation, I am fine with yours if you want. Can you merge it into this pull request or do I need to do that?

jngrad:
We should first decide on the interface. Currently my function works for a single coordinate in N-dimensional space, whereas yours seems to work on a numpy array of points in N-dimensional space. That would be an extension of scipy.misc.derivative. Is that correct?

Kischy:
I would go for a implementation like this:

```python
def calc_partial_derivatives(func, x0, dx=1.0e-7):    
    """
    Approximate all the partial derivatives of a function at a point x0 
    using the two-point central difference formula with spacing dx.

    Parameters
    ----------
    func: `function` obj
        function from which the derivative is calculated
    x0: (N) array_like of obj:`float`
        Point in N-dimensional space where the derivatives are calculated
    dx: :obj:`float`, optional
        Spacing
    Returns
    -------
    array_like of obj:`float` or `float` (if x0 is unidimensional)
        the partial derivatives    
    """       
    
    unidimensional = not np.array(x0).shape
    if unidimensional:
        x0 = [x0]        
        
    derivs = []
    
    for i, _ in enumerate(x0):
        x_plus = np.array(x0, dtype=float)
        x_minus = np.array(x0, dtype=float)        
        x_plus[i] += dx
        x_minus[i] -= dx
        
        derivs.append((func(x_plus) - func(x_minus)) / (2.0 * dx))    
        
    if unidimensional:
        derivs = derivs[0]
    else:
        derivs = np.array(derivs)

    return derivs
```


This is a combination of both our implementations.

Kischy:
> That would be an extension of scipy.misc.derivative. 

I'm not sure. I would say an extension of the scipy.misc.derivative would be a vectorized version. As you said before this function calculates partial derivatives and therefore it does something different, I think.

Kischy:
```python
def calc_partial_derivatives(func, x0, dx=1.0e-7):    
    """
    Approximate all the partial derivatives of a function at a point x0 
    using the two-point central difference formula with spacing 2dx.

    Parameters
    ----------
    func: :obj:`function`
        function from which the derivative is calculated
    x0: (N,) array_like of :obj:`float`
        Point in N-dimensional space where the derivatives are calculated
    dx: :obj:`float`, optional
        Spacing
    Returns
    -------
    array_like of obj:`float` or :obj:`float` (if x0 is unidimensional)
        the partial derivatives    
    """       
    
    unidimensional = not np.array(x0).shape
    if unidimensional:
        x0 = [x0]        
        
    derivs = []
    
    for i, _ in enumerate(x0):
        x_plus = np.array(x0, dtype=float)
        x_minus = np.array(x0, dtype=float)        
        x_plus[i] += dx
        x_minus[i] -= dx
        
        derivs.append((func(x_plus) - func(x_minus)) / (2.0 * dx))    
        
    if unidimensional:
        derivs = derivs[0]
    else:
        derivs = np.array(derivs)

    return derivs
```

Kischy:
Styling rules again. Sorry.

jngrad (approved):

testsuite/python/interactions_non-bonded.py
```diff
@@ -588,40 +588,167 @@ def test_gaussian(self):
 
         self.system.non_bonded_inter[0, 0].gaussian.set_params(eps=0.)
 
+    # Test the Gay-Berne potential and the resulting force and torque
     @utx.skipIfMissingFeatures("GAY_BERNE")
     def test_gb(self):
+
+        # helper function definitions
+        def calc_partial_derivatives(func, x0, dx=1.0e-7):
```
KaiSzuttor: aka gradient

Kischy: Ok, did you rename it and update the docstring?

jngrad: Yes: 377b216

Kischy: ty

KaiSzuttor:
I was not yet done...

KaiSzuttor:
I think we should have all reviewers approve the PR before merging

Kischy:
You changed the interface of the gradient function so that `x0: (3,)`? Does that matter that much now that it is specific to the gb_test?

Could you also update the Docstring?
`Point in N-dimensional space where the derivatives are calculated`
to
`Point in 3-dimensional space where the derivatives are calculated`

KaiSzuttor:
the scope of the function is limited to ```test_gb```. Have you ever calculated a gradient in more than three dimensions?

Kischy:
> the scope of the function is limited to test_gb.

That's why I thought that it doesn't matter.

> Have you ever calculated a gradient in more than three dimensions?

No

KaiSzuttor:
I thought it's just more readable

Kischy:
Yeah, you are right.

Kischy:
Can you exchange every "derivative" with "gradient" in the Docstring? I think than we are done. :)

KaiSzuttor:
sorry for the long discussion, but I think we overall improved readability. Thanks for your contribution!

Kischy:
@KaiSzuttor  I like your implementation of the gradiant function, it is quite compact. However it might be slower, which does not matter in this case. Just keep that in mind, if you ever use it heavily somewhere else.

Kischy:
> 
> 
> sorry for the long discussion, but I think we overall improved readability. Thanks for your contribution!

Yes, it looks better now. Thank you. I'm glad I could help. And no worries about the discussion.

KaiSzuttor:
> However it might be slower

Why?

Kischy:
> > However it might be slower
> 
> Why?

There will be one lambda for each call to the function. That will make it slower in comparison. I shouldn't have brought it up, because it does not matter.

PR #3092, opened 2019-08-21
Core: Explicitly include debug.hpp for FORCE_TRACE()

RudolfWeeber:
Build on a compute cluster failed without this. Include what you use...

Fixes #

Description of changes:
 - 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?

### DIFF
--- a/src/core/electrostatics_magnetostatics/coulomb.cpp
+++ b/src/core/electrostatics_magnetostatics/coulomb.cpp
@@ -6,6 +6,7 @@ double coulomb_cutoff;
 
 #ifdef ELECTROSTATICS
 #include "communication.hpp"
+#include "debug.hpp"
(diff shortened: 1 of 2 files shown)


jngrad (approved):
LGTM. However keep in mind we're currently removing debug macros in project [Interaction Kernel Refactoring](https://github.com/espressomd/espresso/projects/6), see e.g. 0c63f76. `FORCE_TRACE` seems useless, and `GHOST_FORCE_TRACE` is unused, so we will probably remove both of them in the future.

fweik:
I think we should remove all the debug macros. They are less powerful than a debugger and add a lot of coupling.

KaiSzuttor:
bors r=jngrad

Issue #3095, opened 2019-08-21
Remove debug macros


fweik:
Remove the debug macros from `debug.hpp` and the corresponding includes. They are less powerful than a debugger and add a lot of noise and coupling.

fweik:
This can probably be automated with a clever combination  of `unifdef` and `grep`.

jngrad:
<s>Please cherry-pick  0c63f76 first! One debug macro is misnamed and can fool tooling.</s> Update: dealt with in #3098. Use grep on [maintainer/configs](https://github.com/espressomd/espresso/tree/python/maintainer/configs), too, and update [doc/sphinx/installation.rst](https://github.com/espressomd/espresso/blob/python/doc/sphinx/installation.rst) accordingly.

RudolfWeeber:
Once the debug macros are gone, we can also remove the nocheck_maxset CI job.

jngrad:
Obvious candidates were removed in #3098. The 12 remaining ones are located in parts of the core I'm less familiar with. These grep commands print the number of occurrences of debug/trace macros:
```bash
for f in `grep -Phor '[A-Z_]+_DEBUG' src maintainer doc testsuite | sort | uniq`; do echo $f; grep -r $f src maintainer doc testsuite | wc -l; done
for f in `grep -Phor '[A-Z_]+_TRACE' src maintainer doc testsuite | sort | uniq`; do echo $f; grep -r $f src maintainer doc testsuite | wc -l; done
```

jngrad:
Some debug messages could be converted to comments. I've seen a few helpful `CELL_TRACE` printf messages that clarified what the code was doing when moving particles to neighboring cells.

jngrad:
The `debug.hpp` include is required in some places to get access to the global variable `this_node`. It's not always possible to use `communication.hpp` because we cannot include `mpi.h` in CUDA code, e.g. in `EspressoSystemInterface.hpp`.

There is a global variable `regular_exit` that doesn't seem used. The `CUDA_DEBUG` guards a call to `cudaDeviceSynchronize()`, if it's helpful for debugging GPU code please let me know, otherwise I'll remove it in the final PR.

fweik:
Using `this_node` but not needing access to a communicator sounds odd.

PR #3098, opened 2019-08-23
Remove legacy debug macros

jngrad:
Fixes partially #3095

Remove debug macros that are unused, trivial, or preventing further energy/force kernel refactoring.


fweik:
Can you please also check for now-unneeded includes of `debug.hpp`, this should be almost all of them, so I think you can just automatically remove all of them, and re-add where needed. (My best guess is that it's only needed in `cells.cpp` for the sorting check).

jngrad:
Unfortunately the majority of them still need to include `debug.hpp` because of this global variable:
https://github.com/espressomd/espresso/blob/852b1a0c4728041ffce660a6bb7071176b13508f/src/core/debug.hpp#L49

fweik:
Files that use this should include `communication.hpp`, also `s/this_node/comm_cart.rank()/g'. But maybe not in this PR.

jngrad:
Let's do this change in another PR, I need this PR merged to move on with kernel refactoring.

fweik:
Agreed.

fweik (approved):

jngrad:
bors r=fweik

PR #3100, opened 2019-08-23
Narrow signatures of interaction kernels

### LINKED ISSUE #3034: Narrow signatures of interaction kernels

jngrad:
Fixes #3034


fweik (approved):

jngrad:
bors r=fweik

PR #3105, opened 2019-08-26
Fix Constraint checkpointing

### LINKED ISSUE #2943: Some constraints cannot be indexed
Espresso segfaults when iterating or indexing `system.constraints` after some specific constraints have been added. The offending constraints are otherwise working as expected, for example the Gravity object in the MNWE at the bottom of this message correctly drives particles in the right direction. But as soon as the user or any function (e.g. the visualizer) indexes or iterates the list of constraints, it segfaults. Trace:

```
[takin:640566] Signal: Segmentation fault (11)
[takin:640566] Signal code: Address not mapped (1)
[takin:640566] Failing at address: (nil)
[takin:640566] [ 0] /lib/x86_64-linux-gnu/libc.so.6(+0x3ef20)[0x14e3d6360f20]
[takin:640566] [ 1] /work/jgrad/espresso/build/src/python/espressomd/script_interface.so(+0x2d286)[0x14e3cab68286]
[takin:640566] [ 2] /work/jgrad/espresso/build/src/python/espressomd/script_interface.so(+0x2d5aa)[0x14e3cab685aa]
[takin:640566] [ 3] /work/jgrad/espresso/build/src/python/espressomd/script_interface.so(+0x13280)[0x14e3cab4e280]
[takin:640566] [ 4] /work/jgrad/espresso/build/src/python/espressomd/script_interface.so(+0x151ec)[0x14e3cab501ec]
[takin:640566] [ 5] /usr/bin/python3(_PyObject_FastCallKeywords+0x19c)[0x57ec0c]
...
[takin:640566] *** End of error message ***
Segmentation fault (core dumped)
```

There is no error message, unless the script is executed in gdb, where `Exception <class 'RuntimeError'> Type does not have a target` pops up everywhere, starting from:
```
#0  0x000015554978054f in __pyx_pf_10espressomd_16script_interface_16PScriptInterface_12call_method (__pyx_v_self=__pyx_v_self@entry=0x1555200319f8, 
    __pyx_v_method=Python Exception <class 'RuntimeError'> Type does not have a target.: 
, __pyx_v_kwargs=__pyx_v_kwargs@entry={}) at /work/jgrad/espresso/build/src/python/espressomd/script_interface.cpp:3720
```
where `script_interface.cpp:3720` corresponds to this line:
https://github.com/espressomd/espresso/blob/3c598176e31fed0d7f549bcdb4b90483b3d61844/src/python/espressomd/script_interface.pyx#L71

Here is a MNWE reproducing the error for two constraints (I didn't check all of them):
```python
import espressomd
import espressomd.shapes
import espressomd.constraints
import numpy as np

system = espressomd.System(box_l=3*[2])

print('shape constraint')
cylinder = espressomd.shapes.Cylinder(center=3*[1], axis=[0, 1, 0],
                                      direction=-1, radius=.2, length=.2)
system.constraints.add(shape=cylinder)
for c in system.constraints: print(type(c))

print('magnetic field constraint')
H_constraint = espressomd.constraints.HomogeneousMagneticField(H=[0, 1, 0])
system.constraints.add(H_constraint)
for c in system.constraints: print(type(c))

# segfaults
print('gravity constraint')
G_constraint = espressomd.constraints.Gravity([0,1,0])
system.constraints.add(G_constraint)
for c in system.constraints: print(type(c))

# segfaults
print('potential field constraint')
P_constraint = espressomd.constraints.PotentialField(
    default_scale=1.0, particle_scales=[(0, 0.5)],
    field=np.array([[[[1.]]]]), grid_spacing=3*[4.])
system.constraints.add(P_constraint)
for c in system.constraints: print(type(c))
```

Output:
```
shape constraint
<class 'espressomd.constraints.ShapeBasedConstraint'>
magnetic field constraint
<class 'espressomd.constraints.ShapeBasedConstraint'>
<class 'espressomd.constraints.HomogeneousMagneticField'>
gravity constraint
[takin:642548] *** Process received signal ***
[takin:642548] Signal: Segmentation fault (11)
[takin:642548] Signal code: Address not mapped (1)
[takin:642548] Failing at address: (nil)
...
```

jngrad:
Fixes #2943

Enable checkpointing of all `Constraint` classes which are not based on shapes. This also solves the iterator issue in #2943.

API change: it is now required to instantiate all `Constraint` classes with named arguments instead of positional arguments.

### DIFF
--- a/src/core/field_coupling/fields/Interpolated.hpp
+++ b/src/core/field_coupling/fields/Interpolated.hpp
@@ -108,7 +108,7 @@ public:
   template <class Q = T>
   typename std::enable_if<codim == 1, std::vector<Q>>::type
   field_data_flat() const {
-    T const *data = reinterpret_cast<const T *>(m_global_field.data());
+    auto const *data = reinterpret_cast<T const *>(m_global_field.data());
     return std::vector<T>(data, data + m_global_field.num_elements());
   }
 
--- a/src/core/field_coupling/fields/Interpolated.hpp
+++ b/src/core/field_coupling/fields/Interpolated.hpp
@@ -104,26 +104,10 @@ public:
     return {m_global_field.shape(), m_global_field.shape() + 3};
   }
 
-  /** Serialize scalar field */
-  template <class Q = T>
-  typename std::enable_if<codim == 1, std::vector<Q>>::type
-  field_data_flat() const {
+  /** Serialize field */
+  std::vector<T> field_data_flat() const {
     auto const *data = reinterpret_cast<T const *>(m_global_field.data());
-    return std::vector<T>(data, data + m_global_field.num_elements());
-  }
-
-  /** Serialize vector field */
-  template <class Q = T>
-  typename std::enable_if<codim != 1, std::vector<Q>>::type
-  field_data_flat() const {
-    auto const &data = m_global_field.data();
-    std::vector<T> data_flat(codim * m_global_field.num_elements());
-    for (size_t i = 0; i < m_global_field.num_elements(); ++i) {
-      for (size_t j = 0; j < codim; ++j) {
-        data_flat[i * codim + j] = data[i][j];
-      }
-    }
-    return data_flat;
+    return std::vector<T>(data, data + codim * m_global_field.num_elements());
   }
 
   /*
--- a/src/core/field_coupling/fields/Interpolated.hpp
+++ b/src/core/field_coupling/fields/Interpolated.hpp
@@ -51,13 +51,20 @@ void deep_copy(boost::multi_array<T, 3> &dst,
 } // namespace detail
 
 /**
- * @brief A vector field interpolated from a regular grid.
+ *  @brief A vector or scalar field interpolated from a regular grid.
  *
- * This is an interpolation wrapper around a boost::multi_array,
- * which can be evaluated on any point in space by spline interpolation.
+ *  This is an interpolation wrapper around a boost::multi_array,
+ *  which can be evaluated on any point in space by spline interpolation.
+ *
+ *  @tparam T      Underlying type of the field values, see @ref value_type
+ *  @tparam codim  Dimension of the field: 3 for a vector field,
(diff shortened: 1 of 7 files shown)


jngrad:
@fweik I'm not sure how to interpret the ASAN error `stack-use-after-scope` [here](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/153553). It's triggered by:

https://github.com/espressomd/espresso/blob/e2e7c5db0a71b5270734ec2443a0f8408467ce99/src/script_interface/constraints/fields.hpp#L144-L150

When adding a printf to show `data_ptr[0]`, it becomes an ASAN `heap-use-after-free` error. Is the `boost::multi_array` returned by `this_().field_data()` holding an array that has been freed by the time we run the lambda in the `AutoParameter` constructor above? Here is the `field_data()` function:
https://github.com/espressomd/espresso/blob/e2e7c5db0a71b5270734ec2443a0f8408467ce99/src/core/field_coupling/fields/Interpolated.hpp#L94

This issue didn't show up before because `field_params_impl<Interpolated<T, codim>>::params()` isn't tested.

jngrad:
The two ASAN errors `stack-use-after-scope` and `heap-use-after-free` are now gone in the Clang version installed in the docker image after 74ff66e. On the older Clang version installed at the ICP, an ASAN error is still raised but it doesn't show a backtrace nor an error message, only the string `"((IsAligned((uptr)ptr, alignment))) != (0)" (0x0, 0x0)` (fixed by [D44404](https://reviews.llvm.org/D44404) in our Docker image). Still not sure why the `boost::multi_array` would be memcopied somewhere else and the original array freed in the middle of a simulation setup.

jngrad:
Nevermind, the cryptic error message fixed by D44404 pops up everywhere when running the core unit tests of the python branch with our outdated Clang compiler. We can safely ignore it.

RudolfWeeber (approved):

jngrad:
bors r=RudolfWeeber

Issue #3104, opened 2019-08-25
drude gpu sample fails CI


jngrad:
Two samples have been randomly failing tests recently:
* Grand Canonical has large deviations from target concentration (for the last 2 weeks)
* Drude in BMIM PF6 has broken bonds (for the last 2 days)

jngrad:
Drude is still failing:
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/156474
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/154727

jngrad:
Finally found the source of the error: bad P3M parameters from the tuning function. There seems to be a pattern in P3M parameters from simulations that crash: the `r_cut` value is almost twice as small as in simulations that run fine. To reproduce it locally or on a coyote with `ubuntu-python3:cuda-10.1`:
```bash
make local_samples
cd testsuite/scripts/samples
sed -i  "/system.actors.add(p3m)/i p3m._params = {'cao': 7, 'inter': 32768, 'r_cut': 2.5491526892051883, 'alpha': 1.286486160729783, 'accuracy': 0.0009884728050820963, 'mesh': [120, 120, 120], 'epsilon': 0.0, 'mesh_off': [0.5, 0.5, 0.5], 'tune': True, 'check_neutrality': True, 'prefactor': 1389.3612645, 'alpha_L': 47.67025070635568, 'r_cut_iL': 0.06879447050636864, 'cao_cut': [0.0, 0.0, 0.0], 'a': [0.0, 0.0, 0.0], 'ai': [0.0, 0.0, 0.0], 'inter2': 0, 'cao3': 0, 'additional_mesh': [0.0, 0.0, 0.0]}" local_samples/drude_bmimpf6.py
rm -f local_samples/drude_bmimpf6_gpu_processed.py; ../../../pypresso test_drude_bmimpf6_with_gpu.py
```

The LJ sigmas are in the range 3.4-5.0, in simulations that don't crash the tuned P3M `r_cut` is in the range of 3.3-5.1, while in crashed simulations `r_cut` is in the range 2.5-2.8. If this is the real cause, we could add a lower bound to `r_cut` in the parameters of the tuning function.

Note: checkpointing the particle positions/forces/velocities obtained from a bad tuning (i.e., the simulation eventually crashed) into a simulation with good P3M parameters doesn't lead to a crash.

fweik:
[...]

PR #3107, opened 2019-08-28
Fix failing grand_canonical sample test and add documentation to samples

jonaslandsgesell:
This PR adds documentation in to the sample files:

* widom_insertion.py
* wang_landau_reaction_ensemble.py
* grand_canonical.py
* reaction_ensemble.py

The PR also fixes partly #3104: The problem with the failing test was that the excess chemical potential did not match the submitted concentration.
I now provide a matching pair of concentration and excess chemical potential.

### DIFF
--- a/samples/wang_landau_reaction_ensemble.py
+++ b/samples/wang_landau_reaction_ensemble.py
@@ -17,7 +17,9 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 """
-This sample simulates the Wang-Landau Reaction Ensemble for a harmonic bond.
+This example script simulates two reacting monomers which are bonded via a harmonic potential.
+The script aborts as soon as the abortion criterion in the Wang-Landau algorithm is met.
+The Wang-Landau potential which is updated during the simulation is written to the file WL_potential_out.dat.
 """
 import numpy as np
 
--- a/samples/wang_landau_reaction_ensemble.py
+++ b/samples/wang_landau_reaction_ensemble.py
@@ -19,13 +19,13 @@
 """
 This example script simulates two reacting monomers which are bonded via a harmonic potential.
 The script aborts as soon as the abortion criterion in the Wang-Landau algorithm is met.
-With the setup of the Wang-Landau algorithm in this script you sample the density of states of a three dimensional reacting harmonic oscillator 
-as a function of the two collective variables which are introduced are the degree of association and the potential energy.
 The Wang-Landau simulation runs until the Wang-Landau potential is converged and then raises a Warning that it has converged, effectively aborting the simulation.
-The recorded Wang-Landau potential (which is updated during the simulation is written to the file WL_potential_out.dat) 
-contains the density of states. You can view the converged Wang-Landau potential e.g. via plotting with gnuplot: sp "WL_potential_out.dat". 
+With the setup of the Wang-Landau algorithm in this script you sample the density of states of a three dimensional reacting harmonic oscillator 
+as a function of the two collective variables 1) degree of association and 2) potential energy.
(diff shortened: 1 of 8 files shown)


testsuite/scripts/samples/test_grand_canonical.py
```diff
@@ -20,7 +20,7 @@
 
 sample, skipIfMissingFeatures = importlib_wrapper.configure_and_import(
     "@SAMPLES_DIR@/grand_canonical.py", n_int_cycles=51, n_int_steps=5,
-    warm_n_times=10, warm_steps=50, cmd_arguments=[0.01, 0.01])
+    warm_n_times=10, warm_steps=50, cmd_arguments=[8.523659461370000200e-04, -3.365753246398797693e-01])
```
jngrad: how many significant figures do we really need here?

samples/wang_landau_reaction_ensemble.py
```diff
@@ -17,7 +17,9 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 """
-This sample simulates the Wang-Landau Reaction Ensemble for a harmonic bond.
+This example script simulates two reacting monomers which are bonded via a harmonic potential.
+The script aborts as soon as the abortion criterion in the Wang-Landau algorithm is met.
+The Wang-Landau potential which is updated during the simulation is written to the file WL_potential_out.dat.
```
jngrad: Just to check my understanding: we have a neutral monomer (type 3) and a base monomer (type A-) connected by a bond, and two protons (type H+) to set the pH. The WLRE algorithm changes the base monomer into its conjugated acid (type AH) and back using MC moves and stores the degree of association ñ and the chemical potential µ to two files.

testsuite/scripts/samples/test_grand_canonical.py
jonaslandsgesell: 8.523659461370000200e-04 was the concentration I used for determining the excess chemical potential. This is therefore exact. 
The value for the excess chemical potential has a standard error of  +/- 1.590969027075880404e-03 therefore I cropped the result to the third digit after the point: -0.336

samples/wang_landau_reaction_ensemble.py
jonaslandsgesell: The situtation is the following: We have two bonded monomers. These can change their type. They are either of type A- or of type HA. When they are of type A- they have released the corresponding H+ ion. The pH would have to be determined through the reweighted H+ concentration after the simulation is over.
The Wang-Landau algorithm is a histogram reweighting algorithm and drives the system out of equilibrium via applying a Wang-Landau potential. This is not directly connected to the chemical potential. The Wang-Landau potential either represents the canonical partition function, the density of states or the a mixture of both, depending on the acceptance probabilities you impose on the system.
A detailed explaination can be found in https://pubs.acs.org/doi/full/10.1021/acs.jctc.6b00791 .

testsuite/scripts/samples/test_grand_canonical.py
jngrad: you're providing a concentration with 19 significant figures... either WLRE relies on chaos theory, or you're not familiar with Avogadro's constant :) please trim that down, espresso runs on multiple CPU architectures with different machine precision, your value doesn't even fit inside a `long double` on a 64bit machine

samples/wang_landau_reaction_ensemble.py
jngrad: Thanks!

jonaslandsgesell (commented):
Looks good to me

jngrad (approved):

PR #3109, opened 2019-08-29
Use directed acyclic graphs in gitlab CI

RudolfWeeber:


jngrad:
For reference: DAG in GitLab-CI, [short explanation](https://gitlab.com/gitlab-org/gitlab-ce/issues/47063), [documentation](https://docs.gitlab.com/ee/ci/directed_acyclic_graph/)

RudolfWeeber:
As far as I can tell from yesterday's build, itś working.

jngrad (approved):

jngrad:
This could also help in our docker repository, where stage N only starts after the slowest job at stage N-1 is complete. This would give us faster feedback during a PR and faster deployment after a merge. The only drawback I can think of is the possibility of a race condition during CI of a merge: a job at stage 1 might not start until a job at stage 3 is complete that requires a runner with the same tags, causing some images to be deployed later than expected. This concern is however irrelevant to the espresso repository.

PR #3022, opened 2019-07-26
Mmm2d

fweik:
Description of changes:
 - Replaced direct cell access in mmm2d by using `local_cells`
 - Removed two globals from the layered cellsystem

This was the last direct usage of the cells outside of the cell systems,
which makes it now possible to hide how the actual storage is implemented.

PR #2725, opened 2019-04-15
WIP: ELC and MMM2D code duplication

reinaual:
Fixes #2279

PR #3112, opened 2019-08-31
More efficient ELC/MMM2D calculations

jngrad:
Apply the subset of #2725 that reduces the number of multiplications in ELC and MMM2D:
```
c[i] = q * a[i] * b[i]
f[k] = p * (c[i] - c[i+1] + c[i+2] - c[i+3] + ...)
```
becomes
```
c[i] = a[i] * b[i]
f[k] = p * q * (c[i] - c[i+1] + c[i+2] - c[i+3] + ...)
```
This is a follow-up to our MMM2D discussion in the [ES meeting of 2019-08-06](https://github.com/espressomd/espresso/wiki/Espresso-meeting-2019-08-06#handle-long-standing-prs).

Further refactoring of ELC and MMM2D has already taken place in #3022.


jngrad:
Looking at the assembly code generated for mmm2d.cpp before and after the PR, we actually increased the number of multiplication operations from 740 to 836, because the original charge multiplications are still performed in the call to `addscale_vec` that replaces the original call to `add_vec`. After discussing it with @reinaual, we decided not to move forward with this PR.

PR #3113, opened 2019-08-31
Change type of True/False flags from integer to boolean

jngrad:
Makes code more readable, helps distinguish bool flags from int-based bitfield flags, and avoids mistakes like `auto c = (mixed_flag == 1) ? a : b;` (what happens when `mixed_flag` is 2?).


RudolfWeeber:
```

             def __set__(self, _v):
                 if is_valid_type(_v, int):
-                    set_particle_virtual(self._id, _v) 
+                    set_particle_virtual(self._id, < bint > _v)
                 else:
-                    raise ValueError("virtual must be an integer >= 0.")
+                    raise ValueError("virtual must be a boolean.")
```

Shouldn't the is_valid_type() be changed as well. It doesn't seem to be in line with the error msg

I assume, this qualifies as an interface change.
Would this go into 4.1? Or would we delay it, since the next version of ES might have more broad interface changes in any case?

jngrad:
Actually, we cannot use `is_valid_type(_v, bool)` because the Python `bool` type isn't recognized by Cython and the libcpp `bool` type cannot be mixed with Python code. For this reason, Cython provides the `bint` type (abbreviation of "boolean integer"), which automatically converts any integer/boolean Python type (including np.int/np.bool) and any C++ integer type to a Python True/False value.

For example, this Cython code:
```python
def __get__(self):
    cdef const bool * x = NULL
    pointer_to_virtual(self.particle_data, x)
    return x[0]
```
compiles to:
```c++
  bool const *__pyx_v_x;
  __pyx_v_x = NULL;
  pointer_to_virtual(__pyx_v_self->particle_data, __pyx_v_x);
  __pyx_r = __Pyx_PyBool_FromLong((__pyx_v_x[0]));
  return __pyx_r;
```

Therefore there is no interface change: any script that uses `virtual=1` will still work.

jngrad:
I think this can be included in 4.1, it makes the ghost, integration and mpi code slightly easier to follow.

RudolfWeeber:
>Actually, we cannot use is_valid_type(_v, bool) because the Python bool type isn't recognized by Cython and the libcpp bool type cannot be mixed >with Python code. For this reason, Cython provides the bint type (abbreviation of "boolean integer"), which automatically converts any integer/boolean >Python type (including np.int/np.bool) and any C++ integer type to a Python True/False value.

>For example, this Cython code:

>def __get__(self):
>    cdef const int * x = NULL
>    pointer_to_virtual(self.particle_data, x)
>    return x[0]

>compiles to:

>  bool const *__pyx_v_x;
>  __pyx_v_x = NULL;
>  pointer_to_virtual(__pyx_v_self->particle_data, __pyx_v_x);
>  __pyx_r = __Pyx_PyBool_FromLong((__pyx_v_x[0]));
>  return __pyx_r;
I see. 
 

>Therefore there is no interface change: any script that uses virtual=1 will still work.

But system.part.select(virtual=1) would return virtual particles before and an empty  set afterwards?

That would be a rather surprising silent interface change.

jngrad:
I haven't thought about particle selection and other functions based on the virtual flag, mainly because the Cython bint type [automatically casts to the Python bool type from C++ int/bool types](https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#types) and the Python bool type automatically casts to the Python int type in arithmetic operations. In fact, many Cython class methods that return booleans [actually have the bint type](https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#built-in-functions). To take your example:
```python
system.part.add(pos=3*[5], virtual=True)
system.part.add(pos=3*[5], virtual=False)
system.part.add(pos=3*[5], virtual=1)
system.part.add(pos=3*[5], virtual=0)

system.part.select(virtual=1).id
system.part.select(virtual=0).id
system.part.select(virtual=True).id
system.part.select(virtual=False).id
```
returns the same output before and after this PR. You could of course design a piece of code such that it would fail silently after this change, e.g. `return isinstance(system.part[0].virtual, bool)`, but it's not obvious to me under which circumstances you would need this type test. If you'd rather have this interface change in espresso 5, I can revert the `virtual` change, since it's the only one that is accessible from the Python interface, and still merge the rest.

RudolfWeeber:
Since the particle selection and property comparison still works, I think, this can go ahead.

RudolfWeeber (approved):

PR #3117, opened 2019-09-04
WIP: LB particle coupling

### LINKED ISSUE #2870: LB CPU: Node velocity and interpolated velocity inconsistancy

### LINKED ISSUE #2875: LB CPU+GPU: ext_force_density ignored in 1st time step

KaiSzuttor:
Fixes #2870
Fixes #2875

Description of changes:
 - add the force density contribution to all LB fluid velocity related functions
 - buffer the force densities caused by coupled particles for reuse


src/core/grid_based_algorithms/lb.cpp
```diff
@@ -948,17 +950,9 @@ inline void lb_collide_stream() {
...
 #endif // LB_BOUNDARIES
 
-#ifdef VIRTUAL_SITES_INERTIALESS_TRACERS
-  // Safeguard the node forces so that we can later use them for the IBM
-  // particle update
-  // In the following loop the lbfields[XX].force are reset to zero
-  // Safeguard the node forces so that we can later use them for the IBM
-  // particle update In the following loop the lbfields[XX].force are reset to
-  // zero
   for (int i = 0; i < lblattice.halo_grid_volume; ++i) {
```
fweik, line 953: I think the extra sweep is not needed, this can be included into the update loop. You should probably time this, but should be faster.

KaiSzuttor:
we will defer to applying the planed changes to walberla only

PR #3123, opened 2019-09-05
ELC Coverage, integer flags converted

### LINKED ISSUE #2685: Identify (and resolve) missing coverage for ELC

reinaual:
Should fix #2685 for the neutral case.
Also converted integer flags to boolean in ELC and MMM2D.


jngrad (approved):
LGTM

Issue #3134, opened 2019-09-05
DAG in GitLab-CI shortcomings


jngrad:
The new CI dependency graph introduced in #3109 (later modified in #3133) doesn't always work.

For example when stage 1 fails, stage 3 still runs instead of going directly to stage 4:
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/pipelines/9012

When some jobs in stage 2 fail, stage 3 can't start and we never reach stage 4:
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/pipelines/9023

jngrad:
Offline discussion with @RudolfWeeber: we might have to re-create the scenario described above in PRs and tweak the DAG until it's fixed.

jngrad:
It looks like `check_cuda_maxset_no_gpu` is sometimes rebuilding the core and Python interface instead of recycling the artifacts, which increases the runtime  from 5 min to 20 min:
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/159169

jngrad:
This is no longer an issue on the `python` branch. Closing.

PR #3136, opened 2019-09-06
Fix issues with Gitlab CI dag

RudolfWeeber:
Fixes partly #3134


jngrad (approved):
The CI logic has improved, but we still go through stage 3 when stage 1 fails:
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/pipelines/9059

PR #3141, opened 2019-09-07
Remove debug macros

jngrad:
Fixes partially #3095

Remove debug macros in the integration loop, thermostat, halo, cell system and ghost manager that are preventing further refactoring of the core. Some debug messages were converted to comments. The code was also cleaned up, mostly to limit the scope of intermediate variables, make pointers const and change integer flags to the bool type.


src/core/communication.hpp
```diff
@@ -36,7 +36,7 @@
  *  the action issued. If applicable, the second parameter is the node
  *  number of the slave this request is dedicated to.
  *
- *  To add new actions (e.g. to implement new Tcl commands), do the
+ *  To add new actions (e.g. to implement new Python features), do the
```
KaiSzuttor: this does not depend on the script language to reach the interface in the core. just call it ```interface functionality``` or similar.

RudolfWeeber (approved):
bors r+

jngrad:
bors retry

PR #3144, opened 2019-09-09
external force density in first integration step

### LINKED ISSUE #2875: LB CPU+GPU: ext_force_density ignored in 1st time step

KaiSzuttor:
Fixes #2875

Description of changes:
 - initialize the force density member with the external force density set by the user
 - if node velocity is set from the interface, the force density on this node is reset


KaiSzuttor:
ready for review @RudolfWeeber

RudolfWeeber (changes requested):
The call
```
    lbfields = lb_get_initialized_fields(lbpar, lblattice);
```
might result in the fields array to be allocated twice (once the previous value of fields, once the one instanced in the funciton.).
Probably better pass fields as a reference to the funciton.
Also, please add doxgen for teh funciotn since it has some non-obvious logic.

There is a lot of unrelated formatting in there. Please check that your autopep8 args and version match the one in the CI.

KaiSzuttor:
sorry, I forgot to enable some formatting flags... the version is the same as in the CI

RudolfWeeber (approved):
bors r+

Issue #3145, opened 2019-09-09
system.analysis.linear_momentum is fishy


RudolfWeeber:
because it uses
mass * velocity +Force/2 *timestep / mass (statistics.cpp:predict_momentum_particles())
for the particle momentum, independently of wheter or not the fluid momentum is requested or not.

I think, the F/2 in the particle momentum may be wrong even in conjunction with LB particle coupling, since:
* not all the forces on the particles originate from the LB coupling
* It is not clear to me, whether the LB coupling forces on the particles are behind by 1/2 dt in the first place.

I suggest to delete this function and expose the fluid momentum to Python directly.
Particle momentum can be calculated in Python using
np.matmul(system.part[:].v.T,system.part[:].mass)
or a 2-line loop.

KaiSzuttor:
resolved in #3156

RudolfWeeber:
Is that change already mentioned in the release notes in the ES wiki? Otherwise, please do so.

PR #3146, opened 2019-09-09
Omit virtual sites in mass related analysis

### LINKED ISSUE #2276: mass-dependent observables shouldn't include virtual particles

RudolfWeeber:
Fixes #2276.

Center of mass, angular momentum, ComVelocity, ComForce, ComPos.

I left analysis.linear_mometnum() alone due to unclear interactions with the LB particle coupling (#3145)

There is also an unrelated fix where the interface (not result) of DipolarDirectSumWithReplicaCpu was tested even if EXPERIMENTAL_FEAUTERS was not complied in. It is not clear to me, why that was not detected by the test suite.


RudolfWeeber:
Fixes #2276.

jngrad:
I've updated your first message with `Fixes #2276`, because GitHub only looks for that command in the first post.

jonaslandsgesell:
Did you look at the kinetic pressure?

src/python/espressomd/analyze.pyx
```diff
@@ -1054,15 +1056,30 @@ class Analysis:
...
+    
+        Parameters
+        ----------
+        p_type : :obj:`int`
+            Particle :attr:`~espressomd.particle_data.ParticleHandle.type` for
+            which to calculate the center of mass.
+    
+        Returns
+        -------
+        array of :obj:`float`
```
KaiSzuttor: ```(3,) array_like of :obj:`float` ```

testsuite/python/analyze_mass_related.py
```diff
@@ -60,17 +65,41 @@ def i_tensor(self, ids):
                  - np.outer(p.pos - com, p.pos - com))
         return I
 
-    def test(self):
+    def test_itensor(self):
         # Particles of type 0
-        I0 = self.i_tensor(range(0, 10, 1))
+        I0 = self.i_tensor(self.system.part.select(
+            lambda p: (not p.virtual) and p.type == 0).id)
```
KaiSzuttor, line 74: this only works ```IF VIRTUAL_SITES```, *i.e.* ```espressomd.has_features("VIRTUAL_SITES")```

testsuite/python/analyze_mass_related.py
```diff
@@ -60,17 +65,41 @@ def i_tensor(self, ids):
...
             I0, self.system.analysis.moment_of_inertia_matrix(p_type=0), atol=1E-9)
         # type=1
-        I1 = self.i_tensor(range(20, 30, 1))
+        I1 = self.i_tensor(self.system.part.select(type=1).id)
         self.assertTrue(
             np.allclose(I1, self.system.analysis.moment_of_inertia_matrix(p_type=1), atol=1E-9))
 
+    def test_center_of_mass(self):
+        no_virtual_type_0 = self.system.part.select(
+            lambda p: (not p.virtual) and p.type == 0)
```
KaiSzuttor, line 85: this only works ```IF VIRTUAL_SITES```, *i.e.* ```espressomd.has_features("VIRTUAL_SITES")```

testsuite/python/analyze_mass_related.py
```diff
@@ -60,17 +65,41 @@ def i_tensor(self, ids):
...
+            com += p.pos * p.mass
+        com /= np.sum(no_virtual_type_0.mass)
+
+        np.testing.assert_allclose(
+            com,
+            self.system.analysis.center_of_mass(p_type=0))
+    
+    def test_angularmomentum(self):
+        no_virtual_type_0 = self.system.part.select(
+            lambda p: (not p.virtual) and p.type == 0)
```
KaiSzuttor, line 110: this only works ```IF VIRTUAL_SITES```, *i.e.* ```espressomd.has_features("VIRTUAL_SITES")```

KaiSzuttor (changes requested):

KaiSzuttor:
please check https://github.com/espressomd/espresso/blob/b814c25207937d478ed0fd6aefc7f351e4c37c34/src/core/pressure_inline.hpp#L233-L251

src/python/espressomd/analyze.pyx
RudolfWeeber: It returns an ndarray. Do you still think, it should be array_like?

testsuite/python/analyze_mass_related.py
jngrad: the `virtual` property of particles now return False if `VIRTUAL_SITES` is not compiled in

jngrad:
Should particles be guarded by `if(p.p.virtual)` only if they come from a `PartCfg`, or in every case? For example the center of mass in [src/core/statistics_chain.cpp#L70-L102](https://github.com/espressomd/espresso/blob/72084f5ee912a581a00fea7b6b77bcaab1f36f82/src/core/statistics_chain.cpp#L70-L84) is not guarded.

src/python/espressomd/analyze.pyx
KaiSzuttor: yes, why not?

jonaslandsgesell: Unrelated to your question: Do we want to state a warning in the description of the function that the angular momentum is not conserved when PBCs are applied? See e.g. https://arxiv.org/abs/1312.7008

fweik:
The proper solution to this is to separate the particle/DoF selection from the specific calculation. Adding a condition everywhere is not a good solution because it does not scale (if you want do change e.g. what makes a particle virtual you have to change many unrelated places), mixes concerns (why does the center of mass momentum calculation have to now what virtual sites are and how to identify them) and is bug prone (because for every new thing you want to calculate you can forget anew to exclude virtual sites). This is exactly how incidental complexity is picked up over time until every change requires expert knowledge and a lot of typing, as is the case for this project.
It's also not clear what should be included where. E.g. should virtual sites be included at all in observables? Currently e.g. the sum of all forces in `ParticleForce` is not the total forces, because
the forces on virtual sites are counted twice.

RudolfWeeber:
>The proper solution to this is to separate the particle/DoF selection from the specific calculation. Adding a condition everywhere is not a good solution because >it does not scale (if you want do change e.g. what makes a particle virtual you have to change many unrelated places), mixes concerns (why does the center of >mass momentum calculation have to now what virtual sites are and how to identify them) and is bug prone (because for every new thing you want to calculate >you can forget anew to exclude virtual sites). 

Agreed. A more general solution is, however, not feasible within the deadline for 4.1 and I’d rather not ship with this issue un-fixed.

 

The entire analysis/observables complex should be a focus for the next release.

E.g., most stuff under system.analysis could be an observables, observables should now about result shape, most of the chain analysis stuff could be represented by observables.

RudolfWeeber:
>Do we want to state a warning in the description of the function that the angular momentum is not conserved when PBCs are applied? See e.g. > <https://arxiv.org/abs/1312.7008> https://arxiv.org/abs/1312.7008

No. That is a physics, not a software issue, in my view.

fweik:
@RudolfWeeber I think your last comment is for the wrong issue.

fweik:
That was #2714, I think we could not find a nice interface for that. Maybe one could look into that again. This is of course also a hack, but a better one :-)

RudolfWeeber:
Throwing, when attempting to calculate the gyration radius on chains including virtual sites.
It is not clear, whether VS should be excluded only for the center of mass, or also from the gyration radius part.

src/core/event.cpp
```diff
@@ -456,3 +448,18 @@ void on_ghost_flags_change() {
   }
 #endif
 }
+
+void update_dependent_particles() {
+  if (recalc_forces) {
```
KaiSzuttor: please extract this flag to a function parameter

RudolfWeeber:
On Wed, Sep 11, 2019 at 03:24:19PM +0000, Kai Szuttor wrote:
> 
> 
> 
> > @@ -456,3 +448,18 @@ void on_ghost_flags_change() {
>    }
>  #endif
>  }
> +
> +void update_dependent_particles() {
> +  if (recalc_forces) {
> 
> please extract this flag to a function parameter
Not sure, I see the benefit. The call sites know nothing about it, either.
I removed the condition, instead. If someone calls this, they probably want the virtual sites to be updated.

RudolfWeeber:
@jngrad, @KaiSzuttor, I think, I addressed everything.

jngrad (commented):
there are two places left without a virtual guard: [galilei.cpp](https://github.com/espressomd/espresso/blob/3b0c1931bbe27306421c0e6976d6c243c55f92c0/src/core/galilei.cpp#L56-L75), [Mass.hpp](https://github.com/espressomd/espresso/blob/3b0c1931bbe27306421c0e6976d6c243c55f92c0/src/core/field_coupling/couplings/Mass.hpp#L28-L31)

jngrad (approved):
ok from my side

RudolfWeeber:
@KaiSzuttor?

src/core/pressure_inline.hpp
```diff
@@ -225,23 +225,26 @@ inline void add_three_body_bonded_stress(Particle const &p1) {
...
+      virials.data.e[0] += Utils::sqr(time_step) * p1.m.v.norm2() * p1.p.mass;
+    }
 
-  /* ideal gas contribution (the rescaling of the velocities by '/=time_step'
-   * each will be done later) */
-  for (int k = 0; k < 3; k++)
-    for (int l = 0; l < 3; l++)
-      p_tensor.data.e[k * 3 + l] +=
-          (p1.m.v[k] * time_step) * (p1.m.v[l] * time_step) * p1.p.mass;
+    /* ideal gas contribution (the rescaling of the velocities by '/=time_step'
```
KaiSzuttor, line 241: this comment contains outdated information

KaiSzuttor (approved):
LGTM

PR #3155, opened 2019-09-11
Fix regressions found by exotic archs

jngrad:
Part of the [release checklist](https://github.com/espressomd/espresso/wiki/release-checklist#for-all-releases): run tests on exotic architectures and intel compiler.

### DIFF
--- a/maintainer/CI/build_cmake.sh
+++ b/maintainer/CI/build_cmake.sh
@@ -136,7 +136,7 @@ if [ ${with_ccache} = true ]; then
     cmake_params="${cmake_params} -DWITH_CCACHE=ON"
 fi
 
-command -v nvidia-smi && nvidia-smi
+command -v nvidia-smi && nvidia-smi || true
 if [ ${hide_gpu} = true ]; then
     echo "Hiding gpu from Cuda via CUDA_VISIBLE_DEVICES"
     export CUDA_VISIBLE_DEVICES=""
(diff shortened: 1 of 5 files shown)


jngrad:
481ebc7 fixes this issue:
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/158686

The s390x arch failed on coyote11
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/159999
The periodicity errors are reproducible on my machine and on a bwCloud machine. The global variable is not broadcasted correctly, because the code logic in [global.cpp](https://github.com/espressomd/espresso/blob/python/src/core/global.cpp) casts the `std::bitset<3>` to an int, and broadcasts a pointer to the object with size `MPI_INT`, when in reality the bits are stored in an `unsigned long`. Adding a new case for `MPI_UNSIGNED_LONG` fixes that regression. It's curious it only happens on this architecture, where the size of int and long is the same as on amd64, namely 4 resp. 8. The wrong half of the register, i.e. 0x0000, should be broadcast on some of the other archs too.

mkuron:
s390x is the only big-endian architecture, that‘s why we test on it. Could you please push a fix for this regression?

KaiSzuttor (approved):

jngrad:
bors r=KaiSzuttor

PR #3151, opened 2019-09-10
cleanup of LB interface.

KaiSzuttor:

PR #3160, opened 2019-09-12
Improve readability of LB shear test and add unit conversion to global LB stress property

KaiSzuttor:
~~Note that there is still no test for the global stress for the LB.~~

This PR contains #3151


RudolfWeeber (approved):
bors r+

PR #3161, opened 2019-09-12
Core: do_non_bonded, capture by reference rather than value

RudolfWeeber:
This fixes a performance regression from #3077, I think.
Time per step fro lj benchmark went down from 0.78 to 0.38 ms


jngrad (approved):

fweik:
@RudolfWeeber was this in a debug built?

RudolfWeeber:
>was this in a debug built?

No. First thing I checked.

Would you have expected the compiler to figure out that the copy wasn’t needed?

fweik:
Yes, because `Particle` only has compiler generated special members, so I'd expect it to figure out that the copy does not have side effects...

fweik:
Btw this should only capture the id, not the whole particle. This would also have avoided the regression...

```c++
 std::none_of(p1.el.begin(), p1.el.end(),
                      [id2 = p2.p.identity](int id) { return id2 == id; });
```

Issue #3147, opened 2019-09-09
Coulomb tuning fails CI on AMD GPU


jngrad:
Over the last 2 days:
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/157535
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/157613
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/157493
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/157874
Restarting the job often fixes it.

RudolfWeeber:
Wasn’t there a hack in the testsuite preventing p3m_gpu from running on the AMD GPU. That was, IIRC, for a specific card ID. Now, we have a different card in the new CI machine.

mkuron:
Is there a reason why we never looked at why the tuning selects a bad set of parameters?

mkuron:
To reproduce this issue locally, use lama (which has an AMD GPU) and run two instances of the test simultaneously. When only one is running, the issue does not occur. 
```bash
for i in $(seq 100); do
    echo $i
    ../../pypresso coulomb_tuning.py > tuning_${i}a.log 2>&1 &
    ../../pypresso coulomb_tuning.py > tuning_${i}b.log 2>&1 &
    wait
done
```

PR #3163, opened 2019-09-12
P3M GPU: don't use mesh sizes with radix > 5 on AMD GPUs

mkuron:
This fixes the problem that P3M tuning sometimes produced parameters that didn't give correct results. For reasons not entirely clear to me, P3M and rocFFT produce wrong results when using mesh sizes that have prime factors greater than 5. For other reasons not clear to me, such mesh sizes are only chosen by the tuning when multiple instances of Espresso are running simultaneously (see https://github.com/espressomd/espresso/issues/3147#issuecomment-530761965).

This is the real fix for #3147. I ran it 100 times. Previously I would get about 10 failures, now I get zero.

### DIFF
--- a/src/core/electrostatics_magnetostatics/p3m_gpu_cuda.cu
+++ b/src/core/electrostatics_magnetostatics/p3m_gpu_cuda.cu
@@ -751,7 +751,7 @@ void p3m_gpu_init(int cao, const int mesh[3], double alpha) {
       if (mesh[0] % 7 == 0 || mesh[0] % 11 == 0 || mesh[0] % 13 == 0 ||
           mesh[1] % 7 == 0 || mesh[1] % 11 == 0 || mesh[1] % 13 == 0 ||
           mesh[2] % 7 == 0 || mesh[2] % 11 == 0 || mesh[2] % 13 == 0) {
-        throw std::string("Mesh size not supported");
+        throw std::runtime_error("Mesh size not supported");
       }
 #endif
 
@@ -774,7 +774,7 @@ void p3m_gpu_init(int cao, const int mesh[3], double alpha) {
                       FFT_PLAN_FORW_FLAG) != CUFFT_SUCCESS ||
           cufftPlan3d(&(p3m_gpu_fft_plans.back_plan), mesh[0], mesh[1], mesh[2],
                       FFT_PLAN_BACK_FLAG) != CUFFT_SUCCESS) {
-        throw std::string("Unable to create fft plan");
+        throw std::runtime_error("Unable to create fft plan");
       }
     }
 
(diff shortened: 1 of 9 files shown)


RudolfWeeber:
Please avoid repeating the forbidden prime numbers. Probably a combination of std::any_of and bool hip_fft_forbidden_prime_factor()
Please add a comment explaining the workaround at the point where you introduce it.

mkuron:
I thought about that, but it's significantly less legible and more code than what I have now. Who can look at
```
std::any_of(hip_fft_forbidden_prime_factors().begin(), hip_fft_forbidden_prime_factors().end(), [&mesh](int f){ return std::any_of(mesh.begin(), mesh.end(), [f](int n){ return n % f == 0; });} )
```
and tell me at first glance what it does? Also, it's not like people constantly add new radices to rocFFT (or any other FFT library). In fact, I don't think any library implements radix-17 or beyond.

I have added comments now though.

RudolfWeeber (approved):
bors r+

src/core/electrostatics_magnetostatics/p3m_gpu_cuda.cu
```diff
@@ -744,6 +744,17 @@ void p3m_gpu_init(int cao, const int mesh[3], double alpha) {
...
 
     if ((p3m_gpu_data_initialized == 0) && (p3m_gpu_data.mesh_size > 0)) {
+#if defined(__HIPCC__) and not defined(__CUDACC__)
+      // rocFFT gives wrong P3M results for mesh sizes whose prime factors are
+      // not 2, 3 or 5. So we check for the other supported prime factors and
+      // throw an error if they are used with HIP.
+      if (mesh[0] % 7 == 0 || mesh[0] % 11 == 0 || mesh[0] % 13 == 0 ||
+          mesh[1] % 7 == 0 || mesh[1] % 11 == 0 || mesh[1] % 13 == 0 ||
+          mesh[2] % 7 == 0 || mesh[2] % 11 == 0 || mesh[2] % 13 == 0) {
+        throw std::string("Mesh size not supported");
```
KaiSzuttor: ```throw std::runtime_error("Mesh size not supported");```

KaiSzuttor: Why do you throw ```std::string```?

mkuron: copy-and-paste error from 23 lines further down... I just noticed that there are a few more places in Espresso that make the same mistake and should probably be changed to a `runtime_error`.

RudolfWeeber:
bors cancel

KaiSzuttor (approved):

jngrad (approved):

jngrad:
bors r=jngrad,KaiSzuttor,RudolfWeeber

Issue #3157, opened 2019-09-11
Split the Particle definition from the rest of particle_data.hpp


RudolfWeeber:
To avoid circular includes

Issue #2964, opened 2019-07-01
Some alignment of quaternion formalism with the referenced works


bogdan-tanygin:
In regards to all the key quaternion related equations in the rotation.cpp... There are references sonnenschein1985 and Omelyan1998 though the code and Martys1999 in the code header and in the documentation. A comparison of the formulae provides a match between these 3 works. For the source code: if one compares each component index of the quaternion and such indices of the torque, omega and other Cartesian (hereinafter, it means 3D space) entities then there is a mismatch with the reference work formulae. Of course, all the formalism works fine now. A quick analysis result: this is just a formal redefinition of the quaternion space with some other alignments: the 4-dim space axes permutation and some inversion. Note, that the Cartesian components are the same. This is why most of formulae looks different. Probably, the real original reference work was different or some other formalism has been used initially. Just one example of practical confusion of this mismatch (there are others as well: with the director, etc):

- from Es. [documentation](http://espressomd.org/html/doc/running.html#integrating-rotational-degrees-of-freedom): _For a value of (1,0,0,0), the body and space frames coincide._

- from Eq. 3 of sonnenschein1985: see the image (the rotation matrix):
![image](https://user-images.githubusercontent.com/12170193/60463243-1b41e500-9c54-11e9-858b-567030b94c65.png)
Obviously, the ksi is 0-th element of the quaternion. If we set 1 there (from q=1 0 0 0 value) then we have the rotational matrix diagonal elements: -1 1 -1. The Eq. 3 will require q=0 0 0 1 in this case. Actually, we can see the the Eq. 3 is different compare to the matrix in define_rotation_matrix().

The same for other equations. Nothing serious, but it, probably, leads to some confusions when someone will try to use the quaternions from Es. in a context of other equations and the formalism and simulation methods which are based on the conventions of the Martys1999, other references with such formalism.

Unless, I seriously misunderstood something, please correct me if this is the case... I'm ready to help: either documentation based or the source code adaptations (permute the indices in the Q formulae, maybe align some tests, etc). You can decide, but I suggest the latter solution cause it is really sophisticated to work with the existing quaternion space when trying to develop some tailored extra features of scientific computing... At least for me :-)

bogdan-tanygin:
a bit related discussions: #1048, #1854

bogdan-tanygin:
this can be related to the notation convention options: either 1st or last component of the quaternion is real. However, probably, this does not change my above comments ;-)

RudolfWeeber:
[...]

Issue #2289, opened 2018-09-30
Replace custom rotation and vector code by boost.qvm


fweik:
[boost.qvm](https://www.boost.org/doc/libs/1_68_0/libs/qvm/doc/index.html) has most of the vector/matrix/quaternion methods that we need. With this we could drop most of our own quaternion and vector code, plus we would gain more things, like matrix-vector ops. Boost.qvm can work with out types by providing some simple template specializations, so this could be rolled out gradually and easily.

This would need to increase the required boost version to 1.62.0.

PR #3164, opened 2019-09-12
Separate quaternion algebra from particle rotation

jngrad:
Description of changes:
- extract quaternion algebra from the `particle_data.hpp` and `rotation.hpp` files
   - removes code duplication caused by a (now resolved) circular dependency (see #3157)
   - makes it possible to replace the quaternion code by a dedicated library, e.g. [boost:qvm::quat](https://www.boost.org/doc/libs/1_68_0/libs/qvm/doc/index.html) or [boost::math::quaternion](https://www.boost.org/doc/libs/1_62_0/libs/math/doc/html/quaternions.html) in the core (see #2289), [rowan](https://rowan.readthedocs.io/en/latest/) in the interface (see #2964)
- simplify code around call sites using Vector4d arithmetic, `std::tuple`, Particle references
- documentation cleanup


RudolfWeeber:
I'd suggest deferring this until after 4.1. The freeze is already a few days overdue and I'd like to focus on release-relevant issues, right now.

src/core/quaternion.hpp
```diff
@@ -0,0 +1,94 @@
...
+  auto const dm = d.norm();
+
+  // null vectors cannot be converted to quaternions
+  if (dm < ROUND_ERROR_PREC) {
+    return {1, 0, 0, 0};
+  }
+
+  // Calculate angles
+  auto const d_xy = sqrt(d[0] * d[0] + d[1] * d[1]);
+  double theta2, phi2;
```
fweik: `T`

src/core/quaternion.hpp
```diff
@@ -0,0 +1,94 @@
...
+ *  if the director is the null vector.
+ */
+template <class T>
+Utils::Vector<T, 4>
+convert_director_to_quaternion(Utils::Vector<T, 3> const &d) {
+
+  auto const dm = d.norm();
+
+  // null vectors cannot be converted to quaternions
+  if (dm < ROUND_ERROR_PREC) {
```
fweik: This assumes `dm` is `double`, please use `std::numeric_limits`.

src/core/quaternion.hpp
```diff
@@ -0,0 +1,94 @@
...
+    // Here the director is co-linear with the z-azis
+    // We need to distinguish between (0, 0, +d_z) and (0, 0, -d_z)
+    theta2 = (d[2] > 0) ? 0 : Utils::pi() / 2;
+    phi2 = 0;
+  } else {
+    // Here we take care of all other directions
+    // We suppose that theta2 = theta/2 and phi2 = (phi - pi/2)/2,
+    // where angles theta and phi are in spherical coordinates
+    theta2 = acos(d[2] / dm) / 2;
+    phi2 = ((d[1] > 0) ? 1 : -1) * acos(d[0] / d_xy) / 2 - Utils::pi() / 4;
```
fweik: `Utils::sgn(d[1])`, `Utils::pi<T>()`

src/core/quaternion.hpp
```diff
@@ -0,0 +1,94 @@
...
+    return {1, 0, 0, 0};
+  }
+
+  // Calculate angles
+  auto const d_xy = sqrt(d[0] * d[0] + d[1] * d[1]);
+  double theta2, phi2;
+  if (d_xy == 0) {
+    // Here the director is co-linear with the z-azis
+    // We need to distinguish between (0, 0, +d_z) and (0, 0, -d_z)
+    theta2 = (d[2] > 0) ? 0 : Utils::pi() / 2;
```
fweik: `Utils::pi<T>()`

src/core/quaternion.hpp
```diff
@@ -0,0 +1,94 @@
...
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+#ifndef QUATERNION_H
+#define QUATERNION_H
+/** \file
+ *  Quaternion algebra.
+ */
+
+#ifdef ROTATION
```
fweik: This is not needed, please remove. (Also observe that the header is not self-contained, because the config is not included).

fweik (changes requested):
Please move `quaternion.hpp` to utils and add unit tests for the three functions in there.

src/core/quaternion.hpp
jngrad: thanks!

src/core/quaternion.hpp
jngrad: fixed

src/core/quaternion.hpp
jngrad: `Utils::sgn(d[1])` won't work here because it returns a value in the set {-1, 0, 1}. If `d[1]` is zero `phi2` will be -pi/2 instead of +pi/2 and the resulting quaternion will have wrong signs, causing the python test `collision_detection` to fail.

src/core/quaternion.hpp
jngrad: fixed

jngrad:
@fweik I personally wouldn't move that file to utils because it's not re-usable. The corner case with +/-pi/2 shows this quaternion implementation only works because the calling sites were tailored for it. Writing a C++ unit test would not really make sense here, because only the final result of the rotation is of interest (already covered by Python tests), and it would make maintenance harder (we once had a C++ tabulated bond with the wrong sign, and a test that made sure the result had the wrong sign, because the python logic was tailored for it). Coming up with a unit test is also annoying because the quaternion formalism is not standard, and the fact that a [unit test was hard-coded inside a calling site](https://github.com/espressomd/espresso/blob/61026c799a93e10e2e0900b43b5abb785c69dbdb/src/core/virtual_sites.cpp#L119-L125) to check the rotation inversion implementation is correct every time the function is called is odd.

fweik:
Please add the tests.

fweik:
Testing the multiplication is easy, just check the multiplication table. For the conversion functions you should check that they are inverse to each other. It's sufficient that they encode the direction, not how. If anything, when there is client code that depends on a specific behavior of the function, especially it is non-standard, that gives more reason to add a test, so that behavior is kept. So you should also add a test for the sign convention.
Also the file should be moved, it's generic math and not related to MD or the espressos core, so it is a better fit for utils.

src/utils/tests/quaternion_test.cpp
```diff
@@ -0,0 +1,107 @@
...
+  /* self-consistency */
+  using Utils::convert_quaternion_to_director;
+  for (int i = -2; i < 3; ++i) {
+    for (int j = -2; j < 3; ++j) {
+      for (int k = -2; k < 3; ++k) {
+        if (i == 0 and j == 0 and k == 0) {
+          continue;
+        }
+        // here the case where j is zero is also tested
+        Vector3d input{{(double)i, (double)j, (double)k}};
```
KaiSzuttor: ```static_cast<double>(i)```...

jngrad: thanks!

src/core/quaternion.hpp
jngrad: The signum edge case is now documented and tested.

fweik (approved):

jngrad:
bors r=fweik

PR #3165, opened 2019-09-13
Never throw strings

mkuron:
Found by @KaiSzuttor in https://github.com/espressomd/espresso/pull/3163#discussion_r324096198.


mkuron:
moved to #3163

PR #3167, opened 2019-09-13
fix for zero particle numbers

jonaslandsgesell:
Fixes divide by zero if there is no count in the rdf. This may happen if for example all particles of a given type are removed from the system during a grand canonical simulation. (Reported by Andrea)

### DIFF
--- a/src/core/statistics.cpp
+++ b/src/core/statistics.cpp
@@ -271,6 +271,8 @@ void calc_part_distribution(PartCfg &partCfg, int const *p1_types, int n_p1,
       }
     }
   }
+  if (cnt == 0)
+    return;
 
   /* normalization */
   *low /= (double)cnt;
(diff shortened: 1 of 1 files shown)


pkreissl (approved):

KaiSzuttor (approved):

src/core/statistics.cpp
```diff
@@ -336,7 +336,8 @@ void calc_rdf(PartCfg &partCfg, int const *p1_types, int n_p1,
     r_out = r_in + bin_width;
     bin_volume = (4.0 / 3.0) * Utils::pi() *
                  ((r_out * r_out * r_out) - (r_in * r_in * r_in));
-    rdf[i] *= volume / (bin_volume * cnt);
+    if (cnt != 0)
+      rdf[i] *= volume / (bin_volume * cnt);
```
jngrad: Wait, what's the point of executing this for loop if `cnt` is zero? `cnt` should be constant in that scope, therefore we can directly write `if(cnt == 0) return;` above the for loop. This would be obvious if the temporary variables were declared inside the loops with const-qualifier ([C++ core guidelines: keep scopes small](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es5-keep-scopes-small)).

jngrad (changes requested):

jngrad (approved):

jngrad:
bors r=jngrad,KaiSzuttor

PR #3174, opened 2019-09-16
remove undocumented side effect in MC moves

jonaslandsgesell:
This part of the code was useful when sampling around a cylinder is performed.
Since it is undocumented and not always desired I remove it.

### DIFF
--- a/src/core/reaction_ensemble.cpp
+++ b/src/core/reaction_ensemble.cpp
@@ -632,17 +632,8 @@ Utils::Vector3d ReactionAlgorithm::get_random_position_in_box() {
         std::sqrt(m_uniform_real_distribution(
             m_generator)); // for uniform disk point picking in cylinder
     double phi = 2.0 * Utils::pi() * m_uniform_real_distribution(m_generator);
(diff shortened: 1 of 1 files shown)


pkreissl (approved):

RudolfWeeber (approved):
Should this go into 4.1?

jonaslandsgesell:
I would prefer it! It is unexpected behaviour I would say. I used it at some time but then forgot about it...

jngrad:
this looks like a silent change to me, if this is a bugfix we should mention it in the release notes

jonaslandsgesell:
I used it for sampling in correspondence with a charged rod, where this was sensible. Sampling a system which does not include a rod but another charged object (like a flexible polymer) may be giving wrong results. That is why I removed this code now.

RudolfWeeber:
I'd consider this a bugfix which should go in. Jonas, can you please add a corresponding entry in the draft release notes in the Espresso Wiki on github.

jonaslandsgesell:
Okay I added a note on the change.

Issue #3001, opened 2019-07-23
ELC energy wrong for non-neutral systems


KonradBreitsprecher:
Dragging a particle through the box in a non-neutral system (testscript below), the ELC energy jumps close to the boundaries (inside the elc_params.space_layer):

![image](https://user-images.githubusercontent.com/3974659/61691944-fc071680-ad2c-11e9-8783-e4ca7d3cff53.png)

The current testcase (elc_vs_mmm2d_nonneutral.py) fails to cover this case because the particles don't enter the space layer (not dragged close enough to the walls). 

I made the following observations:
- Forces are ok (agree with MMM2D)
- Neutral systems are ok
- The error has be related to one of the many parts, where additional contributions from the space layer appear (`p.r.p[2] < elc_params.space_layer`).
- I could not narrow down the error by finding the single contribution that causes a discontinuity. In fact, several parts of the energy calc. cause jumps at the space layer, and probably fail to cancel each other for non-neutral system. 
- Another hot candidate is the part of the energy calculation in coulomb.cpp, where ELC messes with the system charges and P3M sums.

I don't see through the ELC energy calculation (both non-neutral and neutral) and
I'm afraid the best way out of this is to not support ELC for non-neutral systems.

Test script:

```
from __future__ import print_function
import unittest as ut
import espressomd
import numpy as np
import espressomd.electrostatics
from espressomd import electrostatic_extensions


class ELC_vs_MMM2D_neutral(ut.TestCase):
    # Handle to espresso system

    system = espressomd.System(box_l=[1.0, 1.0, 1.0])
    acc = 1e-6
    elc_gap = 8.0
    box_l = 10.0
    bl2 = box_l * 0.5
    system.time_step = 0.01
    system.cell_system.skin = 0.1

    def test_elc_vs_mmm2d(self):
        elc_param_sets = {
            "inert": {"gap_size": self.elc_gap, "maxPWerror": self.acc, "check_neutrality": False},
            "dielectric": {
                "gap_size": self.elc_gap,
                "maxPWerror": self.acc,
                "delta_mid_bot": 0.1,
                "delta_mid_top": 0.9,
                "check_neutrality": False},
            "const_pot_0": {
                "gap_size": self.elc_gap,
                "maxPWerror": self.acc,
                "const_pot": 1,
                "pot_diff": 0.0,
                "check_neutrality": False},
            "const_pot_1": {
                "gap_size": self.elc_gap,
                "maxPWerror": self.acc,
[...]

PR #3178, opened 2019-09-17
Disable ELC for non-neutral systems with dielectric contrast

reinaual:
Disables the parameter combination in ELC causing #3001


jngrad (approved):

jngrad:
Thanks! You can add a line in the [4.1 release notes](https://github.com/espressomd/espresso/wiki/es_41_release_notes#changed-and-removed-functionality) to document this change

KaiSzuttor:
why do we keep testsuite/python/elc_vs_mmm2d_nonneutral.py?

PR #3182, opened 2019-09-20
Support AMD's aomp compiler

mkuron:
As discussed in https://github.com/espressomd/espresso/pull/2901#issuecomment-501290067, AMD has deprecated the hcc compiler a while ago and is upstreaming HIP support into Clang. Until an official Clang version with HIP support is released (probably Clang 10 coming next year), AMD is providing their own version with all patches merged: https://github.com/ROCm-Developer-Tools/aomp.

I have checked whether Espresso builds correctly with that, adapted the build system, and fixed some minor things. Essentially, compiling HIP like this is the same thing as using Clang to compile CUDA code without nvcc; it just targets a different chip architecture. This means that HCC workarounds like https://github.com/espressomd/espresso/pull/2937/files#diff-ae9f214b6d7b9a9ac803d3be6eff2c51R128-R135 will be able to go away soon.

To use it, install AOMP and the build Espresso with
```
cmake .. \
    -DCMAKE_CXX_COMPILER=/usr/lib/aomp_$(aompversion)/bin/clang++ \
    -DCMAKE_C_COMPILER=/usr/lib/aomp_$(aompversion)/bin/clang \
    -DCUDA_NVCC_EXECUTABLE=/usr/lib/aomp_$(aompversion)/bin/clang++
```


RudolfWeeber:
I'll take a look

RudolfWeeber:
I was not able to build that on lama @ icp. Cmake complained about libraries in /opt/rocm shadowing the ones from aomp in /usr/lib/aomp_0.7.5

RudolfWeeber:
If we go ahead with this, support for the old hip compiler needs to be removed and installation docs updated.

mkuron:
The old one is still the “official” one, while aomp is only in beta,

RudolfWeeber:
I think supporting two tool chains for AMd is out of the question from a cost/benefit perspective.

Since we consider Espresso’s support for AMD cards experimental anyway, I’d personally not care about the Aomp stuff being marked beta, as long as it is clear that that’s the way they are going.

While the old hipcc stuff was not marked beta, it forced us to put nasty hacks into the code.

 

So, once the build issues are sorted out, I’d go for it.

PR #3184, opened 2019-09-20
some modifications to constant-pH tutorial

### LINKED ISSUE #3211: Check reaction ensemble tutorial with current HEAD

kosovan:
Fixes #3211 

TODO (will do):

- [x] First , explain how the reaction works, then discuss the acceptance

- [x]  Define pH, pK, activity and its relation chemical potential and concentration

- [x]  Refer to original literature for the acceptance, refer to review paper for discussion of peculiarities and pitfalls

- [x] Include all physical parameters of the simulated system (box_l, N0, pKa, pH_range) in one block at the beginning of the script

- [x] Set N0 and concentration in SI units,  calculate box_l in sigma

- [x] restrict the range of pH values to pH in [pK-2:pK+2]

- [x] Remove the discussion of standard error of the mean, refer to Janke instead (statistical analysis is not the learning goal of this tutorial)

- [x] Samples of alpha are correlated. Therefore, apply blocking method or some other analysis of correlated samples. 

- [x] Remove the comparison of histograms, or define a learning goal for this part. Currently, the comparison suggests that constant-pH results differ from theory, while the actual reason for this is insufficient sampling. For a sufficiently long simulations, they should not differ. This is very confusing, especially if it is not explained.

Description of changes:
 - 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


jonaslandsgesell:
Please go ahead

KaiSzuttor:
can you estimate when this tutorial will be done? We probably want to do 4.1.1 release this week that includes your tutorial.

kosovan:
I am working on it now. I think that the tutorial is quite clean now. 

I suggest that @jonaslandsgesell and @thepith should review the tutorial. 

Still, I would like to add the conversion of units to the SI, and comparison to the reaction ensemble. We can decide later make that part a separate tutorial.

In any case, I want to finish all the work by Thursday.

thepith:
The tutorial is usable as is, only the point
>  Set N0 and concentration in SI units, calculate box_l in sigma

is missing.

In the future we can think of expanding the tutorial to Reaction Ensemble.

jonaslandsgesell:
Thank you. The tutorial looks good to me.
One or two minor things:

- Can you please put the error estimation alogrithm into a function which just returns the estimated error? This would encapsulate the algorithm logically.
I imagine a function which just takes num_As as argument and returns the error.

- Regarding the 95% Confidence interval of the estimated mean value: If the mean value itself is distributed according to a normal distribution, then the 95% CI is given by a z-score of 1.96 (and not 2). Could you link to Wikipedia: https://en.wikipedia.org/wiki/1.96 ? In an ipython notebook you do this via [blue text](url)

PS: Please clear the cells in order to avoid future merge confilicts. (You can do this in the ipython notebook menu: Kernel-> Restart&Clear output, and then saving the notebook)

PPS: maybe you can also run a spell check before final submission?

PPPS: could you also delete the Reaction ensemble tutorial then (the other folder)?

kosovan:
@jonaslandsgesell  @KaiSzuttor 
I am sorry but I cannot do more on this worksheet now. My computer crashed on Thursday, and since then I have been busy with recovering. I resolved some merge conflicts in a text editor but cannot compile espresso and run the notebook on my system. Please, check if it runs cleanly for you before the merge.

kosovan:
I think that it is ready to be merged and used in the tutorials at Espresso school.

jonaslandsgesell:
Can you please merge my fix for the tests:

https://github.com/kosovan/espresso/pull/2

jonaslandsgesell:
The tutorial tests still fail. See CI.

jonaslandsgesell:
Please try to merge my new PR, it might fix the test case again

doc/tutorials/12-constant_pH/si_units.py
```diff
@@ -0,0 +1,37 @@
+import numpy as np
+import scipy.constants as const
+
+# physical constants
+# kg atomic mass unit
+amu = const.physical_constants['atomic mass constant'][0]
```
KaiSzuttor: if this file should be part of espresso in the future, we should not pollute the namespace with such global variables that are just aliases.

doc/tutorials/12-constant_pH/si_units.py
```diff
@@ -0,0 +1,37 @@
...
+N_Av = const.N_A  # Avogadro 1/mol
+
+
+class si_units():
```
KaiSzuttor: please adhere to the naming convention, namely ```CamelCase``` for classes.

doc/tutorials/12-constant_pH/si_units.py
```diff
@@ -0,0 +1,37 @@
...
+        for p in required_params:
+            if(p not in self.params.keys()):
+                raise ValueError("\n" + p
+                                 + " is missing.\nParameters that need to be defined for SI units: "
+                                 + str(required_params) + "\n")
+        # conversion factor for the unit of length
+        self.sigma_to_nm = self.params['sigma_in_nm']
+        # conversion factor for the unit of concentration
+        self.conc_N_sigma_to_mol_L = 1.0e-3 / \
+            (N_Av * (self.params['sigma_in_nm'] * 1e-9)**3)
```
KaiSzuttor: instead of ```N_Av``` just explicitly use ```scipy.constatns.N_A```

doc/tutorials/12-constant_pH/si_units.py
KaiSzuttor: ```amu``` is actually newer used anywhere

KaiSzuttor (changes requested):

doc/tutorials/12-constant_pH/si_units.py
kosovan: @KaiSzuttor would you like it to be calles SIunits or SiUnits? I vote for the first option because it reads easier, although it is not quite camelcase.

doc/tutorials/12-constant_pH/si_units.py
```diff
@@ -0,0 +1,33 @@
+import numpy as np
+# physical constants
+import scipy.constants
+#as const
+
+class SIunits():
```
KaiSzuttor: actually there already exists a package that may be doing what you are looking for:
https://pint.readthedocs.io/

KaiSzuttor: *e.g.*:
``` python
#!/usr/bin/env python3

import scipy.constants
import pint

ureg = pint.UnitRegistry()

sigma = 5 * ureg.nanometer
tau = 1.0 * ureg.picosecond

c = 0.1 * ureg.mol / ureg.litre
box_volume = 10 * sigma * 20 * sigma * 5 * sigma
N_part = int((c * box_volume).to(ureg.mole).magnitude * scipy.constants.N_A)
```

doc/tutorials/12-constant_pH/si_units.py
```diff
@@ -25,9 +25,9 @@ def convert(self, from_value, from_unit, to_unit, exponent=1.0):
         if (from_unit == "nm" and to_unit == "sigma"):
             return from_value / self.sigma_to_nm
         if (from_unit == "mol/L" and to_unit == "N/sigma^3"):
-            return from_value * self.conc_N_sigma_to_mol_L
-        if (to_unit == "mol/L" and from_unit == "N/sigma^3"):
             return from_value / self.conc_N_sigma_to_mol_L
+        if (to_unit == "mol/L" and from_unit == "N/sigma^3"):
+            return from_value * self.conc_N_sigma_to_mol_L
```
jonaslandsgesell: Does this now return that in order to convert from 1/sigma^3 (with sigma=0.355nm) to mol/l, I need to multiply with roughly 37.1? Just checking for a result I know...

kosovan: To get the answer, just add the following line anywhere after initialization of the SI_units class:
`print(SI.convert(1.0,'N/sigma^3','mol/L'))`

Result: 37.11624488606752

jonaslandsgesell: Great! That s easy.

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
```diff
@@ -166,95 +297,53 @@
...
-    "        numAs.append(system.number_of_particles(type=type_A))\n",
-    "    degrees_of_dissociation.append(np.mean(numAs) / N0)\n",
-    "    std_dev_degree_of_dissociation.append(np.std(numAs, ddof=1) / N0)\n",
-    "    print(\"occurred particle numbers of type A \", sorted(set(numAs)))\n",
-    "    histogram = np.bincount(np.array(numAs).astype(int), minlength=N0 + 1)\n",
-    "    histograms.append(histogram / float(len(numAs)))\n",
-    "    if(abs(pH - 1.0) < 1e-4):\n",
-    "        print(\"acceptance rate dissociation\", RE.get_acceptance_rate_reaction(0))\n",
-    "        print(\"acceptance rate association\", RE.get_acceptance_rate_reaction(1))"
+    "        if(np.random.random()<PROB_REACTION): \n",
```
jonaslandsgesell: I don't think this is needed. You can run MD and MC moves sequentially. This is normal practice. Sequentiality will only affect how fast one converges to equilibrium. In the limit of infinite runtime the exact sequence of MC/MD moves does not matter anymore. https://link.springer.com/chapter/10.1007/978-3-642-41315-5_23 See figure 1.

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
```diff
@@ -166,95 +297,53 @@
...
+    "            RE.reaction(N_acid + 1) # should be at least one reaction attempt per particle\n",
+    "        else:\n",
+    "            system.integrator.run(steps=1000);\n",
```
jonaslandsgesell: You should make using the integrator conditional on USE_WCA in order to speed up the ideal case:

```
        else:
            if USE_WCA==True:
                system.integrator.run(steps=1000);
```

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
kosovan: It is common practice but I am not aware of any proof that it is correct. In general, performing MC moves sequentially instead of a random order breaks detailed balance. If there is enough randomness in the system, this breaking might be weak and hardly noticeable. Sot it is not advisable to do it like that.

Even if we routinely break this rule, we should not teach others to do so.

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
kosovan: Yes, we can do it.

doc/tutorials/12-constant_pH/si_units.py
kosovan: I will soon commit a different approach, that uses the pint package and internally checks for correct dimensionality.

doc/tutorials/12-constant_pH/si_units.py
kosovan: DONE

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
jonaslandsgesell: I am not aware of a proof on how detailed balance is broken by sequential MC/MD moves. Are you?

doc/tutorials/12-constant_pH/si_units.py
kosovan: fixed

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
kosovan: Some years back, Olaf Lenz was showing an example of that at the ICP seminar.  I cannot immediately reproduce it now.

Are you aware that a random order could break detailed balance? If not, then we should keep the random order as a safe choice.

kosovan:
@jonaslandsgesell @thepith Please, have a look at my last modifications regarding the SI units. If you do not find an error, then I think this tutorial is ready for a merge.

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
```diff
@@ -105,19 +105,23 @@
...
     "from espressomd import reaction_ensemble\n",
     "\n",
-    "# class for working with SI units (this example works only with a limited set of units)\n",
-    "from si_units import SIunits\n",
-    "# Initialize the class by supplying the SI values of some internal units\n",
-    "# sigma=0.40 nm is a commonly used  particle size in coarse-grained simulations\n",
-    "SI=SIunits(sigma_in_nm=0.355, T_in_K=293.15);\n",
+    "\n",
+    "import scipy.constants # physical constants\n",
+    "import pint # class for working with units and dimensions\n",
```
jonaslandsgesell: This is a desired change. However this breaks the tests of the tutorials. Please also update the sphinx documentation regarding the installation requirements.

@KaiSzuttor  we need dockerfiles with pint installed. How do we proceed here. Update the Docker images?

jonaslandsgesell (changes requested):
The test cases need to be fixed. This requires Docker images to be updated to include pint.

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
jonaslandsgesell: I do not know published literature that prohibits sequentially using MD and MC moves.

The paper "A rigorous procedure for combining molecular dynamics and Monte Carlo simulation algorithms" demonstrates that the correct probability distribution is sampled for the canonical (NVT) case:
https://www.sciencedirect.com/science/article/pii/S0301010400002469

They perform MDMC steps. One MDMC step consists of 

1. m MC steps
2. integrating n timesteps with MD

In section 3 they show that this is a valid approach.

thepith: The problem is that you are proposing a sequential MC scheme. While this can be correct, it can also fail. See the discussion in [1]:

> The implications of our theorem are significant. First, the sequential updating approach, widely practiced and nearly as widely criticized, is correct. All that is required is that the
sweep leave the Boltzmann distribution invariant, in addition to leading to regular sampling. Indeed, there are numerical reports that the sequential updating approach leads not only
to correct sampling but also to more efficient sampling in a variety of contexts, ranging from image processing to membrane physics. There are some simple cases, however, for
which random Metropolis updating is ergodic and regular, but sequential Metropolis updating is not. One is the Ising model at infinite temperature, where the configurations produced by sequential updating simply oscillate between the initial configuration and the configuration with all spins flipped. Another example is the 2x2 Ising model at any temperature. If the spins are updated in the order [[1 2][3 4]], then the initial configurations [[+ -][+ +]] and [[+ +][- +]] lead to sequences that also oscillate between spin–flipped configurations. In these examples, our theorem does not guarantee convergence to the Boltzmann distribution, because assumption (4) is violated. If, however, the individual spin–flip moves within the sweep were performed with a probability (1 - a), as in Eq. (5), then convergence to the Boltzmann distribution would occur in both of these examples.

This shows that sequential updating can break the MC simulation. Therefore, we should be careful how we teach it.

The paper you cite they show that you can combine configurational MC with MD. But they do not discuss whether this is true when you combine "orthogonal" algorithms like cpH and MD (where one samples the configurational space and the other the composition space).

It might well be that the sequential updating of the cpH and the MD move leads to correct sampling. But what is the advantage to do this in this tutorial? I do not see any real advantage, but the big disadvantage that we teach an advanced method which is not generally valid. If we want to use the sequential updating, then we should discuss why we want use it, why it works and when does not work. All of this is out of scope of this tutorial.

The random mixing of the moves is the most basic way how to do it, and in this tutorial we should stick to be basics when it comes to such subjects.

[1] Manousiouthakis, V. I.; Deem, M. W. Strict Detailed Balance Is Unnecessary in Monte Carlo Simulation. The Journal of Chemical Physics 1999, 110 (6), 2753–2756. https://doi.org/10.1063/1.477973.

kosovan:
@jonaslandsgesell can you please check our latest changes? 

In my opinion, the tutorial is now ready to be merged. 

There is still a pending review from @KaiSzuttor

jngrad:
The tutorial test fails on my machine due to the use of unknown variable `degrees_of_dissociation`.

kosovan:
> The tutorial test fails on my machine due to the use of unknown variable `degrees_of_dissociation`.

Because the file `test.py` still runs an old version of the tutorial. I will fix that.

Update: Sorry, I initially misunderstood what was going on. This file existed only locally on my computer. Now I think I know what the trouble is, and will try to fix it.

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
jngrad: `pint` was installed in docker and the requirements updated in 4d655b8

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
```diff
@@ -166,130 +314,251 @@
...
+    "The plot shows that at intermediate pH the concentration of $\\mathrm{B^+}$ ions is approximately equal to the concentration of $\\mathrm{M^+}$ ions. Only at one specific $\\mathrm{pH}$ the concentration of $\\mathrm{B^+}$ ions is equal to the concentration of $\\mathrm{M^+}$ ions. This is the pH one obtains when dissolving the weak acid $\\mathrm{A}$ in pure water.\n",
     "\n",
-    "\\begin{equation}\n",
-    "    p(N_\\mathrm{A})=\\frac{{N_0 \\choose N_\\mathrm{A}} 10^{(\\mathrm{pH-p}K)N_\\mathrm{A}}}\n",
-    "        {\\sum_{N=0}^{N_0} {N_0 \\choose N} 10^{(\\mathrm{pH}-\\mathrm{p}K)N}}\n",
-    "\\end{equation}"
+    "In an ideal system, the ions missing in the simulation have no effect on the results. In an interacting system, the presence of ions in the box affects the properties of other parts of the system. Therefore, in an interacting system this discrepancy is harmless only at intermediate pH. The effect of the small ions on the rest of the system can be estimated from the overall the ionic strength.\n",
+    "$$ I = \\frac{1}{2}\\sum_i c_i z_i^2 $$"
```
jonaslandsgesell: This is not quite true. This depends on the observable you are looking at. In an ideal system the pressure is given by P=\sum_i c_i kT. If the concentration of some species i is wrongly represented, then the pressure which is observed is already wrong. This can be significant in the cpH ensemble at low or high pH.  Please adapt the sentence to reflect this fact.

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
```diff
@@ -166,130 +314,251 @@
...
+   "metadata": {
+    "scrolled": false
+   },
    "outputs": [],
    "source": [
-    "import scipy.stats\n",
-    "def ideal_degree_of_dissociation(pH, pK):\n",
-    "    return 1. / (1 + 10**(pK - pH))\n",
+    "av_c = av_alpha*c_acid # average concentration of B+ in mol/L\n",
+    "err_c = err_alpha*c_acid # error in the average concentration\n",
```
jonaslandsgesell: Please drop the comments they easily get outdated.

jonaslandsgesell:
The new figures are quite helpful in understanding the limitations of the cpH ensemble. Thank you for contributing those.

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
jonaslandsgesell: I would drop the comment, or say "module for working with units and dimensions"

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
kosovan: @jonaslandsgesell I do not understand. What could become outdated in these comments?

jonaslandsgesell: the information that av_c is in mol/l is redundant, because this is carried over from c_acid. This can become outdated if someone changes the unit of c_acid.
You might consider variable names which are more self-explaining. Then the comment is also not needed anymore.

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
kosovan: This is a more general issue, namely what is the meaning of "pressure" in implicit solvent simulations. One could claim that it is the osmotic pressure, but then one needs to specify what is on the other side of the osmotic membrane. This discussion is for sure outside the scope of the reaction toturial

I can address the comment by explicitly stating that the ions do not affect the ionization degree.

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
kosovan: I will make the variable names more explicit.

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
kosovan: ok

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
jonaslandsgesell: The instantaneous "kinetic" virial pressure tensor component i,j is given by:
image,
The last line is the ensemble averaged kinetic virial pressure tensor component i,j. Via some lines of math using the Maxwell-Boltzmann distribution of velocity components one obtains <P_i,j>=\sum_k c_k KT (where k runs over all ion species).

The instantaneous kinetic virial pressure tensor is returned by the pressure tensor command in Espresso.
The result of this observable (and also the pressure observable) will be affected if there are e.g. H+ ions missing at low pH in the cpH ensemble. Therefore the statement  "the ions missing in the simulation have no effect on the results [in an ideal system]" does not hold in general.

jonaslandsgesell:
This is fine for me to merge. Thank you for your improvements.

kosovan:
@jonaslandsgesell @KaiSzuttor 
I think that this PR can be merged now.
Github says that there is one pending change request from @KaiSzuttor but I cannot find it. Can anybody please point me to what is still missing?

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
```diff
@@ -4,114 +4,190 @@
...
-    "for i in range(N0):\n",
-    "    system.part.add(pos=np.random.random(3) * system.box_l, type=type_HA)"
+    "c_acid=concentration(1e-3,'mol/L') # 0.05 mol/L is a reasonable concentration that could be used in experiments\n",
+    "# Using the constant-pH method is safe if Ionic_strength > max(10**(-pH), 10**(-pOH) ) and C_salt > C_acid\n",
+    "c_salt=concentration(2*c_acid) # additional salt to control the ionic strength\n",
+    "# In the ideal system, concentration is arbitrary (see Henderson-Hasselbalch equation)\n",
+    "# but it is important in the interacting system\n",
+    "N_acid = 20  # number of titratable units in the box\n",
+    "\n",
+    "PROB_REACTION=0.5 # select the reaction move with 50% probability \n",
```
KaiSzuttor: could you please check the coding style of this part? Especially:
* please use spaces in assignments: ```a = 3```
* please one assignment per line ~~```a=3; b=2```~~
* is the global system seed needed her? I thought only NPT is left using a global seed...

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
```diff
@@ -120,20 +196,75 @@
...
-    "pK = -np.log10(K)"
+    "# create the particles\n",
+    "##################################################\n",
+    "# we need to define bonds before creating polymers\n",
+    "from espressomd.interactions import HarmonicBond\n",
+    "hb= HarmonicBond(k=30, r_0=1.0)\n",
+    "system.bonded_inter.add(hb)\n",
+    "\n",
+    "# create the polymer composed of ionizable acid groups, initially in the ionized state\n",
+    "from espressomd import polymer\n",
```
KaiSzuttor: please move all imports to the top

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
```diff
@@ -120,20 +196,75 @@
...
+    "polymers = polymer.positions(n_polymers=1,\n",
+    "                             beads_per_chain=N_acid,\n",
+    "                             bond_length=0.9, seed=23)\n",
+    "for p in polymers:\n",
+    "    for i, m in enumerate(p):\n",
+    "       id = len(system.part)\n",
```
KaiSzuttor: I think if you can come up with more expressive variable names it would help to make it more readable:
*e.g.*: 
``` python
for p in polymer:
    for index, monomer_pos in enumerate(p):
```

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
```diff
@@ -120,20 +196,75 @@
...
+    "\n",
+    "\n",
+    "# if needed, set up and tune the Coulomb interaction\n",
+    "if USE_ELECTROSTATICS:\n",
+    "    print (\"set up and tune p3m, please wait....\")\n",
+    "    from espressomd import electrostatics\n",
+    "    p3m = electrostatics.P3M(prefactor=Bjerrum.to('sigma').magnitude, accuracy=1e-3)\n",
+    "    system.actors.add(p3m)\n",
+    "    p3m_params = p3m.get_params()\n",
+    "    for key in list(p3m_params.keys()):\n",
```
KaiSzuttor: this can probably simplified to ```print(p3m.get_params())```

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
```diff
@@ -120,20 +196,75 @@
...
+    "                epsilon=1.0, sigma=1.0,\n",
+    "                cutoff=2**(1.0 / 6), shift=\"auto\")\n",
+    "\n",
+    "# run a steepest descent minimization to relax overlaps\n",
+    "system.integrator.set_steepest_descent(\n",
+    "    f_max=0, gamma=0.1, max_displacement=0.1)\n",
+    "system.integrator.run(20)\n",
+    "system.integrator.set_vv()  # to switch back to velocity Verlet\n",
+    "\n",
+    "# run a short integration before using the electrostatics\n",
```
KaiSzuttor: the comment should explain why there are integrations before setting up electrostatics

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
```diff
@@ -4,114 +4,190 @@
...
+    "# unfortunately, pint module cannot handle cube root of m**3, so we need to explicitly set the unit\n",
+    "N_salt = int(c_salt*Box_V*N_A) # number of salt ion pairs in the box\n",
+    "# print the values of dependent parameters to check for possible rounding errors\n",
+    "print (\"N_salt: {0:.1f}, N_acid: {1:.1f}, N_salt/N_acid: {2:.7f}, c_salt/c_acid: {3:.7f}\".format(\n",
+    "    N_salt, N_acid, 1.0*N_salt/N_acid, c_salt/c_acid))\n",
+    "\n",
+    "n_blocks=16  # number of block to be used in data analysis\n",
+    "desired_block_size = 10  # desired number of samples per block\n",
+    "num_samples = int(n_blocks * desired_block_size / PROB_REACTION) # number of reaction samples per each pH value\n",
+    "pHmin=pK-offset # lowest pH value to be used\n",
```
KaiSzuttor: please check the naming convention for variables: ```long_variable_name``` instead of ```longvariablename```

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
```diff
@@ -4,114 +4,190 @@
...
+    "TYPE_HA = 0; q_HA = 0\n",
+    "TYPE_A  = 1; q_A  =-1\n",
+    "TYPE_B  = 2; q_B  =+1\n",
+    "TYPE_Na = 3; q_Na =+1\n",
+    "TYPE_Cl = 4; q_Cl =-1\n",
+    "\n",
+    "# acidity constant\n",
+    "pK = 4.88\n",
+    "K = 10**(-pK)\n",
+    "offset=2.0 # range of pH values to be used pK +/- offset\n",
```
KaiSzuttor: if this is a constant please use capital letters for the name

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
```diff
@@ -4,114 +4,190 @@
...
+    "num_pHs=15 # number of pH values\n",
```
KaiSzuttor: constant?

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
```diff
@@ -166,95 +314,254 @@
...
-    "    histograms.append(histogram / float(len(numAs)))\n",
-    "    if(abs(pH - 1.0) < 1e-4):\n",
-    "        print(\"acceptance rate dissociation\", RE.get_acceptance_rate_reaction(0))\n",
-    "        print(\"acceptance rate association\", RE.get_acceptance_rate_reaction(1))"
+    "        if np.random.random() < PROB_REACTION:\n",
+    "            RE.reaction(N_acid + 1) # should be at least one reaction attempt per particle\n",
+    "        elif USE_WCA:                \n",
+    "            system.integrator.run(steps=1000)\n",
+    "        numAs_current.append(system.number_of_particles(type=TYPE_A))\n",
+    "        #print (\"NP: A {}, HA {}\".format(system.number_of_particles(type=TYPE_A),system.number_of_particles(type=TYPE_HA)))\n",
```
KaiSzuttor: please remove the commented out code

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
```diff
@@ -166,95 +314,254 @@
...
+    "def ideal_alpha(pH, pK):\n",
+    "    return 1. / (1 + 10**(pK - pH))\n",
     "\n",
-    "pHs = np.linspace(1, 8, num=15)\n",
-    "degrees_of_dissociation = []\n",
-    "std_dev_degree_of_dissociation = []\n",
-    "histograms = []\n",
-    "histogram_edges = range(0,N0+1)\n",
+    "# empty lists as placeholders for collecting data\n",
+    "numAs = [] # number of A- species observed at each sample\n",
```
KaiSzuttor: please come up with a more expressive variable name which maybe makes it possible to get rid of the comment:
*e.g.* ```num_of_A```?

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
```diff
@@ -166,95 +314,254 @@
...
+    "constitute a Markov chain. It is a sequence of realizations of a random process, where\n",
+    "the next value in the sequence depends on the preceding one. Therefore,\n",
+    "the subsequent values are correlated. To estimate statistical error of the averages\n",
+    "determined in the simulation, one needs to correct for the correlations.\n",
     "\n",
-    "\\begin{equation}\n",
-    "    z = \\frac{\\bar{\\alpha} - \\mu_\\alpha}{\\sigma_n} \\sqrt{n} \\sim \\mathcal{N}(0,1)\n",
-    "\\end{equation}\n",
+    "Here, we will use a rudimentary way of correcting for correlations, termed the binning method.\n",
+    "We refer the reader to specialized literature for a more sophisticated discussion, for example [Janke2002]. The general idea is to group a long sequence of correlated values into a rather small number of blocks, and compute an average per each block. If the blocks are big enough, they\n",
```
KaiSzuttor: have you checked that the citation works?

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
```diff
@@ -166,95 +314,254 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# statistical analysis of the results\n",
+    "def block_analyze(input_data,n_blocks=16):\n",
+    "    # statistical analysis of the results\n",
```
KaiSzuttor: one of the comments needs to go

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
```diff
@@ -166,95 +314,254 @@
...
+    "    err_data=np.sqrt(block_var/(n_blocks-1))\n",
+    "    # estimate autocorrelation time using the formula given by Janke\n",
+    "    # this assumes that the errors have been correctly estimated\n",
+    "    tau_data=np.zeros(av_data.shape)\n",
+    "    for val in range(0,av_data.shape[0]):\n",
+    "        if av_data[val] == 0:\n",
+    "            tau_data[val]=-1.0 # unphysical value marks a failure to compute tau\n",
+    "        else:\n",
+    "            tau_data[val]=0.5*block_size*n_blocks/(n_blocks-1)*block_var[val]/(av2_data[val]-av_data[val]*av_data[val])\n",
+    "    return av_data, err_data, tau_data, block_size\n",
```
KaiSzuttor: please also check this code for formatting and style

KaiSzuttor (changes requested):
i've added some comments on the code... I think it's not a lot of work. If my comments are addressed this can be merged.

RudolfWeeber:
https://jupyter-contrib-nbextensions.readthedocs.io/en/latest/nbextensions/code_prettify/README_autopep8.html

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
kosovan: I made the variable name more explicit but still prefer to keep the comment.

kosovan:
@KaiSzuttor I hope now it is ready.

KaiSzuttor:
thanks @kosovan  I think it's fine now, there are still some inconsistencies with respect to style but I think we have to check for some tools to automatically do this. There are plugins for jupyter to apply such style changes automatically. I'll try it and will push to this PR.

KaiSzuttor:
@kosovan please check again, I think I only touched style plus removed the output

kosovan:
I think that everything is OK.

KaiSzuttor (approved):

kosovan:
> -1 Rejected by code reviews

Unfortunately, I do not understand what this means .

KaiSzuttor:
> Unfortunately, I do not understand what this means .

never mind that's our merge-bot that did not want to merge because CI was not green yet.

jonaslandsgesell (approved):

jngrad:
@KaiSzuttor what is bors merging right now? There are two CI pipelines for `staging` currently running: this PR merge and yesterday's merge, the latter having replaced this PR merge in the `staging` history.

PR #3188, opened 2019-09-21
Simplify CI/CD

jngrad:
Description of changes:
- use fewer docker images (`cuda:10.1`, `cuda:tutorials`, `ubuntu-python3:18.04` were merged into the new `ubuntu-python3:cuda-10.1` in espressomd/docker#134)
- generate the Sphinx website in a container with ScaFaCoS
   * reduces the number of regexes in the Sphinx check script
   * simplifies the release checklist: no need to manually build the documentation on a machine with ScaFaCoS installed, instead just post the sphinx artifacts of the release branch CI
- don't build the Doxygen website in PRs: the job cannot fail and doesn't provide any feedback, there is already another job `style_doxygen` that provides feedback
- upgrade Debian9 job to Debian10

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -125,12 +125,10 @@ no_rotation:
 
 ### Builds with different Distributions
 
-#debian:8 removed: similar to ubuntu:1404
-
-debian:9:
+debian:10:
   <<: *global_job_definition
   stage: build
-  image: gitlab.icp.uni-stuttgart.de:4567/espressomd/docker/debian-python3:9
+  image: gitlab.icp.uni-stuttgart.de:4567/espressomd/docker/debian-python3:10
   script:
     - export with_cuda=false
     - export myconfig=maxset make_check=false
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -181,7 +181,7 @@ fedora:
 cuda10-maxset:
   <<: *global_job_definition
   stage: build
-  image: gitlab.icp.uni-stuttgart.de:4567/espressomd/docker/cuda:10.1
+  image: gitlab.icp.uni-stuttgart.de:4567/espressomd/docker/ubuntu-python3:cuda-10.1
   script:
     - export myconfig=maxset with_coverage=false test_timeout=900 srcdir=${CI_PROJECT_DIR}
     - export cmake_params="-DIPYTHON_EXECUTABLE=$(which jupyter)"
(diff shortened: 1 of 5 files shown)


RudolfWeeber (approved):

.gitlab-ci.yml
```diff
@@ -399,7 +399,7 @@ check_sphinx:
   script:
     - sed -i 's/ or "DISPLAY" in os.environ/ or True/' ${CI_PROJECT_DIR}/src/python/espressomd/visualization.pyx
     - cd ${CI_PROJECT_DIR}/build
-    - make -t && make sphinx
+    - make -t && rm src/python/espressomd/visualization.* && make sphinx
```
KaiSzuttor, line 402: why do we need this?

.gitlab-ci.yml
```diff
@@ -181,7 +179,7 @@ fedora:
 cuda10-maxset:
   <<: *global_job_definition
   stage: build
-  image: gitlab.icp.uni-stuttgart.de:4567/espressomd/docker/cuda:10.1
+  image: gitlab.icp.uni-stuttgart.de:4567/espressomd/docker/ubuntu-python3:cuda-10.1
```
KaiSzuttor, line 182: what naming convention are we following for the container images?

.gitlab-ci.yml
jngrad: We have to change the code in a Cython file, otherwise it cannot be imported in python in a docker container. This was done by a sed expression. Then `make -t` made everything up-to-date, and CMake printed the splash screen with the project details and rebuilt the modified Cython file. But with the new container, CMake doesn't realize the Cython file was modified after `make -t` and doesn't rebuild it, causing this [CI error](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/164414). If you delete the .so files then do `make -t`, CMake won't rebuild it either, causing an import error. So we delete the file after `make -t` to trigger the rebuild.

.gitlab-ci.yml
jngrad: We merged the cuda and scafacos dockerfiles. Because scafacos requires a build script, and docker can only access files in the current folder, we can either duplicate the scafacos build file in the cuda folder, or move the cuda dockerfile in the ubuntu folder. I opted for the latter. Once this PR is merged, I'll move the cuda9 dockerfile in the ubuntu folder, delete the old cuda:10 and cuda:tutorial dockerfiles, and the cuda folder.

PR #3189, opened 2019-09-21
Upgrade pep8 version to the one from Ubuntu 18.04

### LINKED ISSUE #2306: Formatting: Upgrade pep8 version to the one from Ubuntu 18.4?

jngrad:
Fixes #2306

Update to autopep8 v1.3.4 with pycodestyle v2.3.1

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -45,7 +45,7 @@ status_pending:
 style:
   <<: *global_job_definition
   stage: prepare
-  image: gitlab.icp.uni-stuttgart.de:4567/espressomd/docker/clang:6.0
+  image: gitlab.icp.uni-stuttgart.de:4567/espressomd/docker/clang-python3:6.0
   dependencies: []
   before_script:
     - git submodule deinit .
--- a/doc/tutorials/02-charged_system/scripts/nacl_units_confined.py
+++ b/doc/tutorials/02-charged_system/scripts/nacl_units_confined.py
@@ -109,7 +109,8 @@ def combination_rule_sigma(rule, sig1, sig2):
         return ValueError("No combination rule defined")
 
 
-for s in [["Cl", "Na"], ["Cl", "Cl"], ["Na", "Na"], ["Na", "Electrode"], ["Cl", "Electrode"]]:
+for s in [["Cl", "Na"], ["Cl", "Cl"], ["Na", "Na"],
+          ["Na", "Electrode"], ["Cl", "Electrode"]]:
     lj_sig = combination_rule_sigma(
         "Berthelot", lj_sigmas[s[0]], lj_sigmas[s[1]])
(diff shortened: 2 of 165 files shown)


jngrad:
Note: the diff of the first 3 commits is unreadable on GitHub but readable in the terminal with command `git diff --word-diff -w python..jngrad/fix-2306`
The other commits are fine. There is no merge conflict with #3188.

Not sure why I got a style patch from CI in 442b505, I'm using the same version of autopep8 and pycodestyle as in the docker container (same MD5 sum), and ran it with `fix_style.sh`.

RudolfWeeber (approved):

mkuron:
It seems like the new version of autopep8 changed the maximum line length from 80 to 74, causing many files to get re-wrapped. I think we should have changed that default to keep the changes minimal. But it's too late now because it's already merged.

jngrad:
`autopep8 --help` and `pycodestyle --help` both show a default value of 79 chars. Sometimes the line wrap inserted a newline right after an opening square or round bracket and I manually moved the left part to the next line, which can give the impression that the line wrap is smaller that 79 (e.g. [here](https://github.com/espressomd/espresso/pull/3189/commits/deb5e617f10e430e6e2e7bb2f898be4e3ea51de6#diff-f51571162683afe1105aee2043d5d7b4)). Did you spot a line wrapped at 74 chars in the first two commits?

mkuron:
Well, the lines consisting entirely of `#`s used to have 80 of them and now they only have 74.

jngrad:
Oh I see now, the PEP requires [comments and docstrings to be wrapped at 72 chars](https://www.python.org/dev/peps/pep-0008/#maximum-line-length). `autopep8` only applied it on comments, not on docstrings. There is no option to tweak that number, and there is no error or warning code associated to it, so we cannot exclude that rule nor change that number. Looking into the `autopep8` code, it is [hardcoded for comments](https://github.com/hhatto/autopep8/blob/120537a051d7f3cbbb5c4ede19b9e515156bd3d1/autopep8.py#L3333-L3336), but it's not implemented for docstrings.

RudolfWeeber:
On Mon, Sep 23, 2019 at 05:42:37AM -0700, Jean-Noël Grad wrote:
> Oh I see now, the PEP requires [comments and docstrings to be wrapped at 72 chars](https://www.python.org/dev/peps/pep-0008/#maximum-line-length). `autopep8` only applied it on comments, not on docstrings. There is not option to tweak that number, and there is no error or warning code associated to it, so we cannot exclude that rule nor change that number. Looking into the `autopep8` code, it is [hardcoded for comments](https://github.com/hhatto/autopep8/blob/120537a051d7f3cbbb5c4ede19b9e515156bd3d1/autopep8.py#L3333-L3336), but it's not implemented for docstrings.
I'd say, we take it as it is. In the end, formatting is a matter of taste.

PR #3192, opened 2019-09-23
Fix deprecated code for recent CUDA versions

### LINKED ISSUE #3187: Guard CUDA-specific statement in Barnes Hut
The following conditional macro is undefined behavior:
https://github.com/espressomd/espresso/blob/03536720f4ac54a3ea201ed9f71265a13d6d8cd5/src/core/actor/DipolarBarnesHut_cuda.cu#L845-L850
because `CUDA_VERSION` expands to nothing. To get it we need to `#include <cuda.h>` somewhere, and then we need to guard the conditional above with an additional check for CUDA via Nvidia. It is used twice in that file and causes the warning `__all() is deprecated in favor of __all_sync() and may be removed in a future release` during compilation in the cuda10 job ([logfile](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/164292)).

@mkuron do you know which if branch (`__all_sync` or `__all`) should be taken on HIP/AMD?

jngrad:
Fixes #3187

### DIFF
--- a/src/core/actor/DipolarBarnesHut_cuda.cu
+++ b/src/core/actor/DipolarBarnesHut_cuda.cu
@@ -842,7 +842,7 @@ __global__ __launch_bounds__(THREADS5, FACTOR5) void forceCalculationKernel(
 
             // Check if all threads agree that cell is far enough away (or is a
             // body, i.e. n < nbodiesd).
-#if CUDA_VERSION >= 9000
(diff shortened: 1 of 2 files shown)


mkuron (approved):

KaiSzuttor (approved):

jngrad:
bors r=KaiSzuttor

jngrad:
bors r=KaiSzuttor

PR #3196, opened 2019-09-24
Active matter tutorial update

### LINKED ISSUE #3208: Check active matter tutorial with current HEAD

christophlohrmann:
Fixes #3208

Description of changes:
 - Small corrections to the new seed handling of thermostats 
- Changed time step in the rectification_simulation so it does not crash anymore when using higher velocities
- changed mass in enhanced_diffusion so the difference between active and passive particles is more visible in the MSD

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


doc/tutorials/06-active_matter/EXERCISES/enhanced_diffusion.py
```diff
@@ -109,7 +104,7 @@
                      tau_lin=16)
     system.auto_update_accumulators.add(msd)
 
-## Exercise 3 ##
+## Exercise 2 ##
```
KaiSzuttor, line 100: After exercise 3 follows exercise 2?

christophlohrmann: I removed ex 1 and kept the order of the rest. The new 3 pops up earlier in the text, but you first have to do ex 2 at the bottom of the file before 3 makes sense

christophlohrmann: You first have to do 2 at the bottom of the file before 3 makes sense

KaiSzuttor: Shouldn't you then change the order?

christophlohrmann: I can't write the `for ...` thing after the loop body

KaiSzuttor: I'm confused

christophlohrmann: I'll rephrase the exercise

KaiSzuttor:
LGTM, there are some things that can be improved, if you want to:

* in https://github.com/espressomd/espresso/blob/08c45fc5178317db4685d01efb7a730a2d2efbdf/doc/tutorials/06-active_matter/EXERCISES/enhanced_diffusion.py#L31-L33
 it's probably a matter of taste, but I think its much clearer to read ```espressomd.observables.ParticlePositions``` compared to just ```ParticlePositions```. So I would just do 
  ```python
  import espressomd
  import espressomd.observables
  ```

* in https://github.com/espressomd/espresso/blob/08c45fc5178317db4685d01efb7a730a2d2efbdf/doc/tutorials/06-active_matter/EXERCISES/enhanced_diffusion.py#L35-L36
 just use ```espressomd.assert_features(["ENGINE", "ROTATION"])```

* in https://github.com/espressomd/espresso/blob/08c45fc5178317db4685d01efb7a730a2d2efbdf/doc/tutorials/06-active_matter/EXERCISES/enhanced_diffusion.py#L41-L44
 just use ```os.makedirs(outdir, exist_ok=True)```

* in https://github.com/espressomd/espresso/blob/08c45fc5178317db4685d01efb7a730a2d2efbdf/doc/tutorials/06-active_matter/EXERCISES/enhanced_diffusion.py#L50-L52
 we should probably use  sth. like 
   ```python
   parser.add_argument('vel', type=float, help='Velocity of active particle.')
   args = parser.parse_arguments()
   ```

* in https://github.com/espressomd/espresso/blob/08c45fc5178317db4685d01efb7a730a2d2efbdf/doc/tutorials/06-active_matter/EXERCISES/enhanced_diffusion.py#L58-L60
 PLEASE USE CAPITAL LETTERS FOR CONSTANTS

* in https://github.com/espressomd/espresso/blob/08c45fc5178317db4685d01efb7a730a2d2efbdf/doc/tutorials/06-active_matter/EXERCISES/enhanced_diffusion.py#L124
 I think it's not obvious that the actual MSD is ```msd.result()[:, 2] + msd.result()[:, 3] + msd.result()[:, 4]``` as I already saw ESPResSo  people being confused about this.

doc/tutorials/06-active_matter/06-active_matter.tex
```diff
@@ -352,16 +352,16 @@ \subsection{\label{sub:edtut}The Enhanced-Diffusion Tutorial}
 
 pos_id = ParticlePositions(ids=[0])
 msd = Correlator(obs1=pos_id,
-                   corr_operation="square_distance_componentwise",
+                   corr_operation="scalar_product",
```
KaiSzuttor: are you sure about this?

christophlohrmann: sorry you are right, we need a subtraction in there

christophlohrmann: I'll leave it as it was then. The correlator output as three components is described in the pdf manual

KaiSzuttor: but it's not documented whether the msd is the sum or the average of the three components

KaiSzuttor (approved):

PR #3197, opened 2019-09-24
LB tutorial refactoring

### LINKED ISSUE #3205: Check LB tutorial with current HEAD

KaiSzuttor:
Fixes #3205 

* removed stokes drag part due to the result being possibly inconsistent under parameter changes
* refactored the simulation code in the other parts


doc/tutorials/04-lattice_boltzmann/04-lattice_boltzmann_part1.ipynb
```diff
@@ -326,178 +324,27 @@
...
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "import espressomd\n",
-    "espressomd.assert_features(['CUDA', 'LB_BOUNDARIES_GPU'])\n",
```
jngrad, line 380: We probably need an `assert_features` here. Tutorial tests rely on that assertion to know if the test can be run or not (otherwise we would have to define the list of assertions in the test itself, and make sure it is always up-to-date). You might pass the tutorial test in the PR CI pipeline, but as soon as it is merged, the main branch CI will run additional tutorial jobs, including one with an empty myconfig file. Can this tutorial run without any feature?

doc/tutorials/04-lattice_boltzmann/04-lattice_boltzmann_part1.ipynb
```diff
@@ -326,178 +324,27 @@
     "### The <tt>LBBoundary</tt> class\n",
     "\n",
     "The <tt>LBBoundary</tt> class represents a boundary on the <tt>LBFluid</tt>\n",
-    "lattice. It is dependent on the classes of the module <tt>espressomd.shapes</tt> as it derives its geometry from them. For the initialization, the arguments <tt>shape</tt> and <tt>velocity</tt> are supported. The <tt>shape</tt> argument takes an object from the <tt>shapes</tt> module and the <tt>velocity</tt> argument expects a list, tuple or array containing 3 floats. Setting the <tt>velocity</tt> will results in a slip boundary condition.\n",
+    "lattice. It dependents on the classes of the module <tt>espressomd.shapes</tt> as it derives its geometry from them. For the initialization, the arguments <tt>shape</tt> and <tt>velocity</tt> are supported. The <tt>shape</tt> argument takes an object from the <tt>shapes</tt> module and the <tt>velocity</tt> argument expects a list, tuple or array containing 3 floats. Setting the <tt>velocity</tt> will results in a slip boundary condition.\n",
```
jngrad:
```suggestion
    "lattice. It depends on the classes of the module <tt>espressomd.shapes</tt> as it derives its geometry from them. For the initialization, the arguments <tt>shape</tt> and <tt>velocity</tt> are supported. The <tt>shape</tt> argument takes an object from the <tt>shapes</tt> module and the <tt>velocity</tt> argument expects a list, tuple or array containing 3 floats. Setting the <tt>velocity</tt> will results in a slip boundary condition.\n",
```

doc/tutorials/04-lattice_boltzmann/scripts/04-lattice_boltzmann_part3_solution.py
```diff
@@ -0,0 +1,109 @@
+import espressomd
+espressomd.assert_features(['LENNARD_JONES'])
```
jngrad: You might want to add tests for these new scripts. And remove the test for the deleted script.

doc/tutorials/04-lattice_boltzmann/04-lattice_boltzmann_part1.ipynb
KaiSzuttor: they only depend on ```LB_BOUNDARIES_GPU``` which depends on the external feature ```CUDA```. Of course this tutorial can only be run with a gpu.

KaiSzuttor:
there is some weird interaction between ```importlib_wrapper.py``` enforcing the ```Agg``` backend for plots and the ```plt.ion()``` command:

``` bash
ValueError: Data has no positive values, and therefore can not be log-scaled.
```

which only occurs for the wrapped script.

KaiSzuttor:
probably related https://github.com/matplotlib/matplotlib/issues/9800

jngrad:
The error disappears if you don't provide a figsize, or if you don't use `plt.ion()`. We could update `disable_matplotlib_gui()` to remove `plt.ion()`.

KaiSzuttor:
> The error disappears if you don't provide a figsize, or if you don't use plt.ion()

both should be possible in a tutorial

jngrad:
Of course. But the wrapper script needs to remove something. Removing `plt.ion()` in wrapped tutorials solves the issue and does not create side-effects in other tutorial tests on my machine.

KaiSzuttor:
okay, could you please push the regex magic to this branch?

fweik:
I have a general question about the import wrapper thing: Wouldn't it be easier (and more robust) to use pythons parser instead of rolling our own? This is available from the build-in `ast` module, which
provides `ast.parse` to produce the AST for a piece of python code, which can then be searched for `Import` statements.
(I'm not sure that your regex even captures the whole [grammar for regex statements](https://docs.python.org/3.7/reference/simple_stmts.html#import), and in general you can not write a python parser with regex because they are not powerful enough: consider `exec("import" + " mathplotlib")`)

KaiSzuttor:
@jngrad already looked into that and me too:

you probably mean something like

```python
#!/usr/bin/env python3

import argparse
import ast

parser.add_argument("source_file", type=str)
args = parser.parse_args()

class VariableTransformer(ast.NodeTransformer):

    def __init__(self, variable_name, new_value):
        self.variable_name = variable_name
        self.new_value = new_value

    def visit_Assign(self, node):
        if node.targets[0].id == self.variable_name:
            node.value = ast.Num(n=self.new_value)
            ast.fix_missing_locations(node)
            return node                                                
        return node


class GetMatplotlibPyplotAlias(ast.NodeVisitor):
    def visit_Import(self, node):
        if node.names[0].name == "matplotlib.pyplot":
            self.matplotlibpyplot_alias = node.names[0].asname

class PltIonTransformer(ast.NodeTransformer):

    def __init__(self, matplotlibpyplot_alias):
        self.matplotlibpyplot_alias = matplotlibpyplot_alias

    def visit_Expr(self, node):
        if hasattr(node.value.func, "value"):
            if node.value.func.value.id == self.matplotlibpyplot_alias:
                if node.value.func.attr == "ion":
                    node.value.func.attr = "ioff"
                    ast.fix_missing_locations(node)
        return node

with open(args.source_file, 'r') as source_file:
    source_code = source_file.read()
    code = ast.parse(source_code)
    transformer = VariableTransformer("MY_CONSTANT", 20)
    transformed_code = transformer.visit(code)
    visitor = GetMatplotlibPyplotAlias()
    visitor.visit(transformed_code)
    transformer = PltIonTransformer(visitor.matplotlibpyplot_alias)
    transformed_code = transformer.visit(transformed_code)
    exec(compile(transformed_code, '<string>', 'exec'))

```

jngrad:
I've starting replacing some of the regex logic by AST in one of my local branches a few months ago. This is still a work in progress, with a few obstacles (e.g. `ast` shows the line number and column number where a statement starts but not where it ends, so removing multi-line statements requires first the implementation of a backtrack mechanism). It would be definitely more maintainable and correct than regular expressions.

fweik:
@jngrad I don't think there is need to do textual replacement, you can modify the ast and then directly run it, as in the example from @KaiSzuttor, otherwise there is e.g. the [`astor`](https://github.com/berkerpeksag/astor) module for code generation, which might also be handy during development.

fweik:
Please either replace the regex or open an issue to do it later. I don't we should keep this, because this means that tutorials would have be written in an undocumented python dialect. This also leads
to very opaque behavior: if I understand your code correctly `import[space]matplotlib` and 
`import[space][space]matplotlib` would behave differently, which is rather unexpected.

KaiSzuttor:
@fweik could you please have a look at the tutorial too? I think it's more or less done by now.

fweik:
What do you want feedback on? On the style of the scripts?

KaiSzuttor:
mostly for the changed content

fweik:
I have no experience with creating didactic material, so I don't really feel qualified. I can check for factual correctness if you'd like.

KaiSzuttor:
> I have no experience with creating didactic material, so I don't really feel qualified. I can check for factual correctness if you'd like.

yes please.

doc/tutorials/04-lattice_boltzmann/04-lattice_boltzmann_part1.ipynb
```diff
@@ -312,10 +311,10 @@
     "The possible properties are:\n",
     "\n",
     "+ <tt>velocity</tt>: the fluid velocity (list of three floats)\n",
-    "+ <tt>pi</tt>: the stress tensor (list of six floats: $\\Pi_{xx}$, $\\Pi_{xy}$, $\\Pi_{yy}$, $\\Pi_{xz}$, $\\Pi_{yz}$, and $\\Pi_{zz}$)\n",
-    "+ <tt>pi_neq</tt>: the nonequilibrium part of the stress tensor, components as above.\n",
+    "+ <tt>stress</tt>: the pressure tensor (list of six floats: $\\Pi_{xx}$, $\\Pi_{xy}$, $\\Pi_{yy}$, $\\Pi_{xz}$, $\\Pi_{yz}$, and $\\Pi_{zz}$)\n",
```
fweik, line 300: Is it true that `stress` sets the pressure tensor? (They are equal up to sign). If so we should reconsider the interface and potentially warn the user about this.

doc/tutorials/04-lattice_boltzmann/04-lattice_boltzmann_part1.ipynb
```diff
@@ -326,178 +325,27 @@
...
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "### System size dependence\n",
+    "Note that all used variables are inherited from previous examples. This will create a wall with a surface normal of $(1, 0, 0)$ at a distance of 1 from the origin of the coordinate system in direction of the normal vector. The wall exhibits a slip boundary condition with a velocity of $(0, 0, 0.01)$. For a no-slip boundary condition, leave out the velocity argument or set it to zero. Please refer to the user guide for a complete list of constraints.\n",
     "\n",
-    "Measure the drag force for a fixed radius but varying system size. Does the drag force increase of decrease with the system size? Can you find a qualitative explanation?"
+    "Currently only the so-called *link bounce back* method is implemented, where the effective hydrodynamic boundary is located midway between boundary and fluid node."
```
fweik: I think it would be better to say "Espresso implements the link bounce back method ..." as none of the other methods are discussed anywhere. This assumes context that is not in the text.

doc/tutorials/04-lattice_boltzmann/04-lattice_boltzmann_part2.ipynb
```diff
@@ -47,16 +47,16 @@
...
     "We want to determine the diffusion coefficient from the mean square\n",
-    "distance that a particle travels in the time $t$. For large $t$ it is\n",
-    "be proportional to the time and the diffusion coefficient occurs as \n",
+    "distance that a particle travels in the time $t$. For large $t$ the mean square distance is\n",
+    "proportional to the time and the diffusion coefficient occurs as a \n",
     "prefactor:\n",
     "\n",
     "\\begin{equation}\n",
-    "  \\frac{\\partial \\langle r^2 \\left(t\\right)\\rangle}{\\partial t} = 2 d D. \n",
+    "  \\frac{\\partial \\langle \\vec{r}^2 \\left(t\\right)\\rangle}{\\partial t} = 2 d D. \n",
```
fweik: I think it would be clearer (and more common) to actually put down the formula for the MSD instead of the derivative. This would also correspond more closely to how it is calculated.

doc/tutorials/04-lattice_boltzmann/04-lattice_boltzmann_part2.ipynb
```diff
@@ -88,88 +88,97 @@
...
+    "import matplotlib.pyplot as plt\n",
+    "plt.rcParams.update({'font.size': 22})\n",
+    "plt.ion() # switch on interactive mode\n",
+    "\n",
+    "plt.figure(figsize=(10,10))\n",
+    "plt.xlabel('tau [$\\Delta t$]')\n",
+    "plt.ylabel('msd [$\\sigma^2$]')\n",
+    "for index, r in enumerate(msd_results):\n",
+    "    # adding up componentwise MSDs\n",
+    "    msd = r[1:100, 2] + r[1:100, 3] + r[1:100, 4]\n",
```
fweik, line 172: You should consider explaining the numbers here.

fweik (dismissed):
The introduction part (which you did not touch) is pretty bad, you should maybe take a look.
It would help to think about what the target audience is and what the required knowledge is
for the tutorials. E.g. the text makes some fringe statement about information theory, but does
not explain how any of the content discussed is related to hydrodynamics. It's also contradictory
(e.g. `f` can not be the probability density to find a molecule and normalized to the total mass
at the same time, as the text states).

doc/tutorials/04-lattice_boltzmann/04-lattice_boltzmann_part1.ipynb
KaiSzuttor: yes, that's true. See https://github.com/espressomd/espresso/blob/d14acdb8642e3546b95a4a3ad9880df6422bf0b5/testsuite/python/lb_shear.py#L139-L157

fweik:
Well, this fits our style...

doc/tutorials/04-lattice_boltzmann/04-lattice_boltzmann_part2.ipynb
```diff
@@ -88,88 +88,97 @@
...
-    "loops = 40000\n",
-    "steps = 10\n",
-    "time_step = 0.01\n",
-    "runtime = loops * steps * time_step\n",
-    "box_l = 16\n",
-    "\n",
-    "lb_friction = 1.0  # The friction has been set to be 1.0 as default\n",
-    "                   # Change this value for further simulations\n",
+    "import logging\n",
+    "logging.basicConfig(level=logging.INFO)\n",
```
jngrad: The default settings of the `logging` module were designed with terminals in mind. All logs are streamed to stderr to help filtering logs from print statements with piping and redirects. In Jupyter, stderr is printed on a red background to alert the user about something critical that does not interrupt the flow of the program (flow interruptions like exceptions are formatted differently). In my opinion, user experience would improve if we didn't display the INFO category in red. This can be achieved by:
```python
import sys
import logging
logging.basicConfig(level=logging.INFO, stream=sys.stdout)
```

doc/tutorials/04-lattice_boltzmann/04-lattice_boltzmann_part2.ipynb
```diff
@@ -88,88 +88,97 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import matplotlib.pyplot as plt\n",
+    "plt.rcParams.update({'font.size': 22})\n",
+    "plt.ion() # switch on interactive mode\n",
```
jngrad: It is a known limitation in the Jupyter backend that the `plt.show()` command prints a string representation of the python object rather than creating a canvas when it's in the same cell as `import matplotlib.pyplot`, forcing the user to execute the cell a second time to get the plot displayed. There are two ways to remediate this: either split the cell to have the import statement separated from the plotting command, or inserting a `%matplotlib inline` before the import statement. If you choose the latter, you'll need to add the following snippet before the `return code` statement of `disable_matplotlib_gui()`:
```python
    # remove magic function
    code = code.replace(
        "get_ipython().run_line_magic('matplotlib', 'inline')", "")
```

KaiSzuttor: i won't touch this... we have this issue in all tutorials and we should just be aware of the bug.

KaiSzuttor: hm, read the doc again... I think we should globally check all tutorials for this

jngrad (approved):

KaiSzuttor:
bors r=jngrad

Issue #3194, opened 2019-09-23
Python3 linter


jngrad:
`pylint` was recently removed (b720b5c1) because it only checked for one warning that had become a false positive in Python3. We can re-enable it (and move it to the style check job) if there is a need for rules that cannot be enforced by `autopep8`. A few candidates: `class-naming-style`, `const-naming-style`. Feel free to propose more rules here. The full list can be found in [Pylint features](http://pylint.pycqa.org/en/latest/technical_reference/features.html).

jngrad:
`pylint` can't run on Cython files, therefore whenever a .py file loads a function from a .pyx file, a warning is generated, usually one of E0001/E0401/E0611/E1101. The solution is to whitelist rules using `--disable=all --enable=...`. Most of the `pylint` logic has already been written down in the `fix_style.sh` script that runs in the style job, it's only a matter of reverting commit f2ea574 and optionally updating the python script that posts messages on GitHub to show the log. We should also pin the version of `pylint` in the dockerfile.

I think the following rules should always be enforced:

* `W0102`: mutable values in optional arguments: `def foo(a=[]):` to `def foo(a=()):`
* `W0401`: wildcard imports, e.g. `from a import *` to `from a import (b, c, d\n e, f, g\n)`
* `W0611`: unused imports
* `W0614`: unused imports from a wildcard
* `W1505`: deprecated `assertRaisesRegexp()`
* `R1701`: merge type checks: `isinstance(x, y) or isinstance(x, z)` to `isinstance(x, (y, z))`
* `R1714`: merge value checks: `if x == 1 or x == 2:` to `if x in (1, 2):`
* `R1707`: trailing comma tuple: `return a,` to either `return a` or `return (a,)`
* `C0201`: iterate dict keys directly
* `C0202`: use `cls` instead of `self` in class methods
* `C0325`: superfluous parentheses: `if(a == 2):` to `if a == 2:`
* `W1652,W1651,W1649,W1657,W1660,W1658,W1659,W1623,W1622,W1620,W1621,W1645,W1624,W1625,`
  `W1611,W1601,W1602,W1603,W1604,...`: using Python2 features that have been deprecated or removed in Python3

The following rules should be run manually and periodically by maintainers to detect candidates for cleanup or refactoring:

* `W0511`: show todo
* `W0612`: unused variable
* `W0613`: unused arguments
* `W1401`: use raw strings in regex patterns
* `R0401`: cyclic imports
* `R0912`: too many branches
[...]

PR #3203, opened 2019-09-25
Code cleanup with pylint

jngrad:
Follow-up to #3194

Description of changes:
- remove unused imports
- rewrite wildcard imports as explicit imports
- replace mutable optional parameters by immutable equivalents
- fix LB thermostat checkpointing mechanism
- cleanup and simplify conditional statements
- refactor code with numpy and new espresso methods
- remove unused variables/arguments
- rename unused loop variables from `i` to `_`
- fix broken numpy commands in parts of the testsuite that aren't executed
- remove trailling whitespaces


KaiSzuttor:
are you sure we want to defer to 5.0? I think backporting bugfixes after the 4.1 release could be harder...

RudolfWeeber:
Please defer. Otherwise all tutorials have to be rechecked

jngrad:
Yes we should defer it to espresso 5. The bugs found here are not executed in our CI pipeline otherwise we would have got syntax errors. I've also not fixed bugs that require more than a one-liner change.

In espresso 4.1, the sources of syntax errors and runtime errors are:
```
src/python/espressomd/MDA_ESP/__init__.py:194:18:
    E0602: Undefined function 'triclinic_vectors()'
src/config/defines.py:24:4: W0102:
    Dangerous default value [] as argument
testsuite/python/data/gen_coulomb_2d_ref_data.py:18:0:
    W0622: Redefining built-in 'int', 'float', 'object', 'sum', 'any', 'range', ...
```
edit: fixes for the following bugs have been backported to 4.1 in f8e36f0:<s>
```
testsuite/python/observables.py:81:32:
    E0602: Undefined function 'average()'
testsuite/python/observables.py:83:32:
    : Incorrect call to function 'sum()'
testsuite/python/tests_common.py:87:0:
    W0102: Dangerous default value [] as argument
testsuite/python/virtual_sites_tracers_common.py:37:4:
    W0102: Dangerous default value [] as argument
```
</s>

RudolfWeeber:
I'm not opposed to the code changes in this PR. 
I'd like to make a general remark though.
I think, the benefit of code beautification not done by automated formatting scripts needs to be  weighed really carefully against the hassle caused in terms of merge conflicts.
I'm in favor of removing dangerous stuff such as wildcard imports into the current namespace.
In my opinion, stuff like isinstance refactoring or import sorting does not provide enough benefit, though.
The fact that something is in the pep rules doesn't necessarily mean following the rules is the best use of limited develpoer time.

This is particularly so for future autmated lint checking in PRs. There, rules should be added on an individual basis, only.

mkuron:
> I'm in favor of removing dangerous stuff such as wildcard imports into the current namespace.
> In my opinion, stuff like isinstance refactoring or import sorting does not provide enough benefit, though.

I agree. New formatting rules should have limited impact on existing code. However, the changes in this pull request are rather small, increase code legibility and unlikely to cause major merge conflicts. So I think it should be merged.

jngrad:
@RudolfWeeber thanks for your feedback! Here is my perspective:
- rules should be added on an individual basis: I strongly support this idea. Hopefully the dedicated issue #3194 will attract feedback that can be used in es meetings to decide on the rules to enforce. Whenever the C++ core is refactored/cleaned-up based on compiler warnings and Clang-format / Clang-tidy rules, we add those to the CMake logic to prevent the re-introduction of bad code. We can do the same with Python, progressively if need be. In https://github.com/espressomd/espresso/issues/3194#issuecomment-535100226 I break down which of the pylint rules used in this PR should be enforced, and which ones shouldn't.
- import sorting: I have no strong opinion on whether to enforce it or not. We already enforce it automatically in C++, and it can be done automatically in Python too, so there shouldn't be any loss of developer time if we choose to do it. We should just make sure that `espressomd.has_features` don't get moved below the import statements, that parenthesis imports don't get transformed to a list of single-import statements and that pylint comments don't get removed.
- refactoring isinstance and friends: there are several spots in the code, especially in the testsuite, where the same code logic was used despite being inefficient, longer and less legible that the refactored version. Most likely these came from copy-pasting blocks of code "that just work". This is one aspect of technical debt. On quite a few occasions I lost time figuring out the intent of the original author in convoluted code blocks, only to realize from the git history it was just the result of minor fixes piling up in the same spot.
- code indentation and formatting: these ones cannot be enforced, since pylint does not apply them automatically. They should be done by autopep8, unfortunately it is not a perfect tool. My plan is to manually apply pylint formatting rules to the files *already touched* by `fix_style.sh` when creating a commit named "Formatting", so as to avoid wasting time in code reviews.
- merge conflicts: I agree it's a major issue with such large scale refactors. I'm not a big fan of them, and whenever I do one, I locally merge the most recent PRs to detect merge conflicts, and either reduce the scope of my refactor to exclude the conflicting files, or if I can't, communicate about it in the es meetings (ex: ["if any merge conflict occurs in PRs, @jngrad can help" in Vector3d refactor, August 6](https://github.com/espressomd/espresso/wiki/Espresso-meeting-2019-08-06)). Sometimes conflicts go under the radar, recently a bors merge failed because of an extra newline character, and a pointer refactor caused conflicts in your PR and introduced a regression in a lambda. Managing large scale refactors is difficult and time-consuming. At the same time, we need them to improve the global state of espresso. Large-scale formatting PRs like this one are the worst in terms of added value divided by conflict resolving time, that's why it is best to do them after a code freeze. Most of the work has been done, now we just have to select a small subset of those rules we think should be enforced. The other rules can be checked by maintainers periodically, because the issues they detect tend to concentrate in poorly maintained parts of the code like OIF.

KaiSzuttor:
maybe we should add a ```.pylintrc``` after the discussion is finished.

jngrad:
I suggest we merge this PR in priority after the end of the summer school.

jngrad:
@KaiSzuttor @christophlohrmann the NPT thermostat checkpointing might be broken too:
https://github.com/espressomd/espresso/blob/9dc1e0606165237913b594dc71a1f081b94d624d/src/python/espressomd/thermostat.pyx#L108-L110
does not match the function signature:
https://github.com/espressomd/espresso/blob/9dc1e0606165237913b594dc71a1f081b94d624d/src/python/espressomd/thermostat.pyx#L419

jngrad:
alright, fixing NPT revealed other bugs in the checkpointing mechanism for integrators. I'll have to open a separate PR, here is not the right place.

jngrad:
The checkpointing bugfixes have their own PR now. The only relevant bugfix commit left in this PR for 4.1.1 (3b716e5) was cherry-picked in the release candidate. This is now ready for review.

KaiSzuttor (approved):

jngrad:
bors r=KaiSzuttor

PR #3212, opened 2019-09-26
Remove unused force reduction function

hirschsn:
The function `reduce_forces_sum` was made obsolete by a538ebb

Fixes #

Description of changes:
 - Remove unused function

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?

PR #3216, opened 2019-09-26
Refactor ghosts.cpp

hirschsn:
More refactoring that builds upon #3212 

Description of *major* changes:
 - remove GhostCommunication::mpi_comm as it is not used in ghosts.cpp,
 - make GhostCommunication::part_lists a std::vector,
 - remove static variables s_buffer and r_buffer,
 - factor out memory handling,
 - change loops to range based for,
 - use boost::mpi.
 - Replace the manual poststore and prefetch loops by find_ifs

Mainly, ghosts.cpp now defines CommBuf, which is a container for the data to be sent or received as well as two classes (Archiver and BondArchiver), that insert and extract the memory from CommBuf.

Some of these changes, like the removal of static variables, is necessary for my implementation of asynchronous ghost communication.

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


hirschsn:
That should be it. Sorry about the unannounced work. I was about to PR my ghost adaptions, so they don't always require a merge if ghosts.cpp changed.

But then I thought while I was at it, I could also go ahead and refactor the file a bit.

I think, now that the type punning and memory handling is factored out and the structure of ghosts_communicator() is a bit easier, the whole file should be easier to comprehend.

RudolfWeeber:
Since you and @fweik know the ghost communication code best, could we organize an offline meeting where we go through the ghost communication together, so everone is on the same page?

hirschsn:
@RudolfWeeber Sure! Just propose a time and date.

fweik:
I'll have a look at the changes today, and then we can decide how to proceed with this.

fweik:
This looks good to me. I'd plan to do something similar in the future but did not get around to it yet, so thank you :-). I've made some cosmetic changes and made a clearer separation between the two archiver classes so that they can be tested better in the future. Please let me now if you are d'accord with this, I did not mean to hijack your PR.

RudolfWeeber:
@hirschsn, @fweik, either during the summer school or the week after. Letś organize it offline, once Iḿ back on the 7th.

hirschsn:
@fweik Feel free to hijack any PR you like. :) I'm totally fine with your changes, they definitely improve this PR. Thanks!
@RudolfWeeber just ping me when you're back.

src/core/ghosts.cpp
```diff
@@ -107,87 +161,60 @@ int calc_transmit_size(GhostCommunication *gc, int data_parts) {
...
-void prepare_send_buffer(GhostCommunication *gc, int data_parts) {
+static void prepare_send_buffer(CommBuf &s_buffer, GhostCommunication &gc,
+                                int data_parts) {
   /* reallocate send buffer */
-  n_s_buffer = calc_transmit_size(gc, data_parts);
-  if (n_s_buffer > max_s_buffer) {
-    max_s_buffer = n_s_buffer;
-    s_buffer = Utils::realloc(s_buffer, max_s_buffer);
-  }
+  s_buffer.resize(calc_transmit_size(gc, data_parts));
```
KaiSzuttor: a lot of the code would be easier to read if the variable names would be just a little more descriptive: *e.g.*:
 * ```send_buffer``` instead of ```s_buffer```,
 * ```rec_buffer``` instead of ```r_buffer```,
 * ```ghost_comm``` instead of ```gc```,
 * ```archiver``` instead of ```ar```,
 * ```bond_archiver``` instead of ```bar```,
 * ```bond_list``` instead of ```bl```,
 * ```part``` instead of ```pt```,
 * ```part_list``` instead of ```cur_list```

src/core/ghosts.cpp
```diff
@@ -29,29 +29,90 @@
...
 #include <vector>
 
 /** Tag for communication in ghost_comm. */
 #define REQ_GHOST_SEND 100
 
-static int n_s_buffer = 0;
-static int max_s_buffer = 0;
-/** send buffer. Just grows, which should be ok */
-static char *s_buffer = nullptr;
+struct CommBuf {
```
KaiSzuttor: please add a docstring for the struct

KaiSzuttor: also think about using ```class``` instead of ```struct``` here. [core guideline](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c8-use-class-rather-than-struct-if-any-member-is-non-public)

src/core/ghosts.cpp
```diff
@@ -405,159 +391,114 @@ void ghost_communicator(GhostCommunicator *gc, int data_parts) {
...
-      case GHOST_SEND: {
-        MPI_Send(s_buffer, n_s_buffer, MPI_BYTE, node, REQ_GHOST_SEND,
-                 comm_cart);
-        int n_bonds = s_bondbuffer.size();
-        if (!(data_parts & GHOSTTRANS_PROPRTS) && n_bonds > 0) {
+    } else if (poststore) {
+      /* send op; write back delayed data from last recv, when this was a
+       * prefetch send. */
+      /* find previous action where we recv and which has PSTSTORE set */
+      auto postst_gcn = std::find_if(
```
KaiSzuttor: please find a readable name for this variable

KaiSzuttor (changes requested):

hirschsn:
@KaiSzuttor Thanks for the suggestions. Done.

I now implemented the following convention in both, ghosts.hpp and ghosts.cpp:

 - "GhostCommunication" variables are always "ghost_comm".
 - "GhostCommunicator" variables are always named "gcr".

The latter one is not very descriptive but that is by choice. Otherwise there would be too much similarity to ghost_communication/ghost_comm/... which is also irritating. If anyone can think of a better combination of names, please tell me.

Also, I sneaked in two more changes: Another pointer->reference conversion and the removal of two superfluous pointer variables in cell_cell_transfer. Could you please take another look at these?

KaiSzuttor (approved):

jngrad:
bors r=KaiSzuttor

PR #3220, opened 2019-09-27
Correctly sort imports

KaiSzuttor:


jngrad:
This will interfere with 2a50ae5 in #3203 where many unused imports were removed.

KaiSzuttor:
I used your branch as a base now

PR #3221, opened 2019-09-27
Make unit tests respect Cmake MPIEXEC_*FLAGS

hirschsn:
There are two cmake flags called MPIEXEC_PREFLAGS and MPIEXEC_POSTFLAGS.
This commit makes the unit test include these two variables in the creation of
the cmdlines. These variables are useful for passing parameters to mpiexec, e.g. MPIEXEC_PREFLAGS="--bind-to;none".

Fixes #

Description of changes:
 - 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


cmake/unit_test.cmake
```diff
@@ -34,9 +34,9 @@ function(UNIT_TEST)
     # OpenMPI 3.0 and higher checks the number of processes against the number of CPUs
     execute_process(COMMAND ${MPIEXEC} --version RESULT_VARIABLE mpi_version_result OUTPUT_VARIABLE mpi_version_output ERROR_VARIABLE mpi_version_output)
     if (mpi_version_result EQUAL 0 AND mpi_version_output MATCHES "\\(Open(RTE| MPI)\\) ([3-9]\\.|1[0-9])")
-      add_test(${TEST_NAME} ${MPIEXEC} -oversubscribe ${MPIEXEC_NUMPROC_FLAG} ${TEST_NUM_PROC} ${CMAKE_CURRENT_BINARY_DIR}/${TEST_NAME})
+        add_test(${TEST_NAME} ${MPIEXEC} -oversubscribe ${MPIEXEC_NUMPROC_FLAG} ${TEST_NUM_PROC} ${MPIEXEC_PREFLAGS} ${CMAKE_CURRENT_BINARY_DIR}/${TEST_NAME} ${MPIEXEC_POSTFLAGS})
     else()
-      add_test(${TEST_NAME} ${MPIEXEC} ${MPIEXEC_NUMPROC_FLAG} ${TEST_NUM_PROC} ${CMAKE_CURRENT_BINARY_DIR}/${TEST_NAME})
+        add_test(${TEST_NAME} ${MPIEXEC} ${MPIEXEC_NUMPROC_FLAG} ${TEST_NUM_PROC} ${MPIEXEC_PREFLAGS} ${CMAKE_CURRENT_BINARY_DIR}/${TEST_NAME} ${MPIEXEC_POSTFLAGS})
```
jngrad: Lines 37 and 39 can probably be combined into one line by adding this construction on top of the file (before line 1):
https://github.com/espressomd/espresso/blob/fc08e8a1095c1221284ba2e322ccf03439dd42c2/testsuite/python/CMakeLists.txt#L7-L15

jngrad (commented):
To get the same MPI command line syntax in the entire project, the new flags could be added here:
https://github.com/espressomd/espresso/blob/fc08e8a1095c1221284ba2e322ccf03439dd42c2/maintainer/benchmarks/CMakeLists.txt#L63-L65

hirschsn:
@jngrad thanks for your suggestions. Done. (I did not add the oversubscribe code at the beginning of the file but rather before the command line. Let me know if you want it at the beginning or if this is okay.)

jngrad:
Moving the block to the top would avoid calling mpiexec in a subprocess every time a new test is configured.  We already do it for the testsuite and benchmarks. This might also help refactoring the CMake logic in the future, by having only one occurrence of this block of code in the entire project, somewhere in the top-level CMake file after the MPI import.

hirschsn:
@jngrad that's definitely a good reason. I hadn't thought about that.

jngrad (approved):
LGTM

PR #3226, opened 2019-10-01
cmake: add soversion to libH5mdCore

junghans:
Just #2946 again, not why it didn't make it in v4.1


mkuron (approved):
Thanks. Seems like we forgot to cherry-pick it into the master branch. We'll definitely need to do that now.

KaiSzuttor (approved):

KaiSzuttor:
@junghans could you please explain why this is needed?

mkuron:
Generating a Fedora RPM throws a warning if a library isn't versioned, if I remember correctly. An unversioned library is named libSomething.so, while a versioned library is named libSomething.so.4 with a symlink libSomething.so pointing to it.

pdbreader.so, mpiio.so and pdbparser.so are also missing versions.

junghans:
In general shared libraries should be versioned, but now as everything gets installed in the pythondir we could drop the soversion completely.

And there is another bug as pdbreader and mpiio don't get installed at all.

junghans:
See #3228 for replacement.

PR #3228, opened 2019-10-02
Fix cmake install

junghans:
Supersedes #3226

Description of changes:
- fix the CMake build system to include libraries in the list of installed files and test it in a dedicated CI job
- fix the versioning of Cython .so shared objects
- simplify the CMake tests


mkuron (approved):

mkuron:
Needs cherry pick into master branch too.

junghans:
> Needs cherry pick into master branch too.

The merge seems to work as well.

junghans:
Ping

CMakeLists.txt
```diff
@@ -29,14 +29,6 @@ endif()
...
 
 set(PROJECT_VERSION "4.1.0")
-string(REGEX
-       REPLACE "^([1-9]+)\\.[0-9]+.*$"
-               "\\1"
-               SOVERSION
-               "${PROJECT_VERSION}")
-if(NOT ${SOVERSION} MATCHES "^[1-9]+$")
-  message(FATAL_ERROR "Could not determine SOVERSION from ${PROJECT_VERSION}")
-endif(NOT ${SOVERSION} MATCHES "^[1-9]+$")
```
jngrad, line 39: Why is SOVERSION not needed anymore?

src/core/io/mpiio/CMakeLists.txt
```diff
@@ -1,3 +1,4 @@
 add_library(mpiio SHARED mpiio.cpp)
 target_include_directories(mpiio PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
 target_link_libraries(mpiio PRIVATE EspressoConfig EspressoCore MPI::MPI_CXX)
+install(TARGETS mpiio LIBRARY DESTINATION ${PYTHON_INSTDIR}/espressomd)
```
jngrad, line 4: What is the purpose of the new `install()` line here and in the other submodules?

jngrad (commented):
Please edit the original post to explain what this PR is achieving.

CMakeLists.txt
junghans: All libraries get to installed in the python site-packages directory and in there so versioning isn’t required anymore. Plus half of the new library miss it anyhow.

src/core/io/mpiio/CMakeLists.txt
junghans: These libraries were missing from the installation and hence made the espressomd module unusable.

jngrad:
@junghans Thanks for reporting this issue. I will add a test to detect this type of regression.

@mkuron While trying to reproduce the issue on Ubuntu, I found another flaw in the build system: the `CMAKE_INSTALL_RPATH` is not updated when the user provides a custom `DESTDIR` during installation. This should have been caught by our CMake tests, unfortunately I did not configure the destdir test properly. I've investigated the issue further and came to the conclusion that we cannot actually use `make install DESTDIR="path/to/install"` and should warn the user explicitly against using that command, and only use `cmake .. -DCMAKE_INSTALL_PREFIX="path/to/install"`. Here is a MWE:
```bash
# with DESTDIR and no CMAKE_INSTALL_PREFIX, PATH is right but RPATH is wrong
cmake ..
make -j$(nproc)
make install DESTDIR="/tmp/espresso-1"
```
```
-- Installing: /tmp/espresso-1/usr/local/lib/python3.6/site-packages/espressomd/EspressoScriptInterface.so
-- Set runtime path of "/tmp/espresso-1/usr/local/lib/python3.6/site-packages/espressomd/EspressoScriptInterface.so" to "/usr/local/lib/python3.6/site-packages/espressomd"
```
```bash
# with no DESTDIR and CMAKE_INSTALL_PREFIX, PATH is right and RPATH is right
cmake .. -DCMAKE_INSTALL_PREFIX="/tmp/espresso-2"
make -j$(nproc)
make install
```
```
-- Installing: /tmp/espresso-2/lib/python3.6/site-packages/espressomd/EspressoScriptInterface.so
-- Set runtime path of "/tmp/espresso-2/lib/python3.6/site-packages/espressomd/EspressoScriptInterface.so" to "/tmp/espresso-2/lib/python3.6/site-packages/espressomd"
```
```bash
# with DESTDIR and CMAKE_INSTALL_PREFIX, PATH is wrong and RPATH is wrong
cmake .. -DCMAKE_INSTALL_PREFIX="/tmp/espresso-3"
make -j$(nproc)
make install DESTDIR="/tmp/espresso-4"
```
```
-- Installing: /tmp/espresso-4/tmp/espresso-3/lib/python3.6/site-packages/espressomd/EspressoScriptInterface.so
-- Set runtime path of "/tmp/espresso-4/tmp/espresso-3/lib/python3.6/site-packages/espressomd/EspressoScriptInterface.so" to "/tmp/espresso-3/lib/python3.6/site-packages/espressomd"
```
When RPATH is wrong, we get the following error message:
```
$> ls /tmp/espresso-1/usr/local/lib/python3.6/site-packages/espressomd/EspressoCore.so 
/tmp/espresso-1/usr/local/lib/python3.6/site-packages/espressomd/EspressoCore.so
$> /tmp/espresso-1/usr/local/bin/pypresso 
Python 3.6.8 (default, Oct  7 2019, 12:59:55) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import espressomd
  File "<stdin>", line 1, in <module>
  File "/tmp/espresso-1/usr/local/lib/python3.6/site-packages/espressomd/__init__.py", line 23, in <module>
    import espressomd._init
ImportError: EspressoCore.so: cannot open shared object file: No such file or directory
```

jngrad:
I have written most of the logic for the new test, however I can't install 4.1.1rc with or without this PR:
```bash
cmake .. -DCMAKE_INSTALL_PREFIX=/tmp/espresso-unit-tests
make -j$(nproc)
make install
./pypresso -c 'from espressomd.cluster_analysis import ClusterStructure;print("ok1")'
/tmp/espresso-unit-tests/bin/pypresso -c 'import espressomd;from espressomd.cluster_analysis import ClusterStructure;print("ok2")'
```
```
ok1
  File "<string>", line 1, in <module>
ImportError: dynamic module does not define module export function (PyInit_cluster_analysis)
```
The `cluster_analysis.so` file is a C++ shared object, which is not meant to be imported from Python. To access its classes, we use the ScriptInterfaceHelper in `cluster_analysis.py`. In the build folder, the .so and .py files are in different folders so there is no name collision. However, when we install espresso, the .so and .py end up in the same folder and Python thinks the .so file is a Cython shared object, which is expected to have a `PyInit_*` function. When I rename `cluster_analysis.py` to e.g. `clusters_analysis.py`, there is no import error.

@KaiSzuttor could this be a regression of our efforts to restructure the core folders?

mkuron:
> @mkuron While trying to reproduce the issue on Ubuntu, I found another flaw in the build system: the `CMAKE_INSTALL_RPATH` is not updated when the user provides a custom DESTDIR during installation.

Just get rid of `DESTDIR`. That's non-standard anyway and everyone should use `CMAKE_INSTALL_PREFIX` instead.

junghans:
`DESTDIR` is really only for packaging, i.e. to make a tarball with a certain `RPATH` encoded without installing it there, so don't use it here.

I would replace `make install` by `cmake --install` to allow for testing with ninja as well.

jngrad:
> I would replace `make install` by `cmake --install` to allow for testing with ninja as well.

Nice feature! But this seems to be [CMake 3.15-specific](https://cmake.org/cmake/help/v3.15/release/3.15.html?highlight=--install#command-line), which has been released very recently. We still support 3.4, and consider bumping to 3.10 in the next release.

junghans:
> > I would replace `make install` by `cmake --install` to allow for testing with ninja as well.
> 
> Nice feature! But this seems to be [CMake 3.15-specific](https://cmake.org/cmake/help/v3.15/release/3.15.html?highlight=--install#command-line), which has been released very recently. We still support 3.4, and consider bumping to 3.10 in the next release.

Then if clause it depending on the CMale version, in CMake-3.4 you could at least use `cmake --build` instead of `make all`.

jngrad (approved):

jngrad:
This PR was merged into the open PR for the 4.1.1 release candidate.

PR #3229, opened 2019-10-02
cmake: generate version.hpp at compile time to avoid rebuilds

junghans:
It annoyed me for years, that if I touch a CMakeLists.txt everything gets rebuild for no reason.


junghans:
CI fail not related to this change.

junghans:
Duplicate of #3259

PR #3234, opened 2019-10-05
Fix mpiio with stdlibc++ range checking

### LINKED ISSUE #3230: mpiio tests fails on i686 with espresso-4.1.0
```
130/135 Test #127: mpiio .........................................***Failed    1.83 sec
/usr/include/c++/9/bits/stl_vector.h:1042: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = int; _Alloc = std::allocator<int>; std::vector<_Tp, _Alloc>::reference = int&; std::vector<_Tp, _Alloc>::size_type = unsigned int]: Assertion '__builtin_expect(__n < this->size(), true)' failed.
===================================================================================
=   BAD TERMINATION OF ONE OF YOUR APPLICATION PROCESSES
=   PID 22870 RUNNING AT buildvm-12.phx2.fedoraproject.org
=   EXIT CODE: 134
=   CLEANING UP REMAINING PROCESSES
=   YOU CAN IGNORE THE BELOW CLEANUP MESSAGES
===================================================================================
YOUR APPLICATION TERMINATED WITH THE EXIT STRING: Aborted (signal 6)
This typically refers to a problem with your application.
Please see the FAQ page for debugging suggestions
```

Details [here](https://github.com/espressomd/espresso/files/3683130/build_i686.log.txt)

mkuron:
Fixes #3230. Reported by @junghans.

When mpiio was used but no bonds were present, we would still try to copy zero bonds from a zero-length vector. This triggered an assertion when stdlibc++ range checking was enabled.

Please tag for cherry-picking into 4.1.1.

### DIFF
--- a/src/core/io/mpiio/mpiio.cpp
+++ b/src/core/io/mpiio/mpiio.cpp
@@ -446,7 +446,7 @@ void mpi_mpiio_common_read(const char *filename, unsigned fields) {
 
     for (int i = 0; i < nlocalpart; ++i) {
       int blen = boff[i + 1] - boff[i];
-      if(blen) {
+      if (blen) {
         auto &bl = local_particles[id[i]]->bl;
         bl.resize(blen);
         std::copy_n(&bond[boff[i]], blen, bl.begin());
--- a/src/core/io/mpiio/mpiio.cpp
+++ b/src/core/io/mpiio/mpiio.cpp
@@ -446,9 +446,11 @@ void mpi_mpiio_common_read(const char *filename, unsigned fields) {
 
     for (int i = 0; i < nlocalpart; ++i) {
       int blen = boff[i + 1] - boff[i];
-      auto &bl = local_particles[id[i]]->bl;
-      bl.resize(blen);
-      std::copy_n(&bond[boff[i]], blen, bl.begin());
+      if(blen) {
+        auto &bl = local_particles[id[i]]->bl;
+        bl.resize(blen);
+        std::copy_n(&bond[boff[i]], blen, bl.begin());
+      }
     }
   }
 
--- a/src/core/io/mpiio/mpiio.cpp
+++ b/src/core/io/mpiio/mpiio.cpp
@@ -446,11 +446,9 @@ void mpi_mpiio_common_read(const char *filename, unsigned fields) {
 
     for (int i = 0; i < nlocalpart; ++i) {
       int blen = boff[i + 1] - boff[i];
-      if (blen) {
-        auto &bl = local_particles[id[i]]->bl;
-        bl.resize(blen);
-        std::copy_n(&bond[boff[i]], blen, bl.begin());
-      }
+      auto &bl = local_particles[id[i]]->bl;
+      bl.resize(blen);
+      std::copy_n(bond.begin() + boff[i], blen, bl.begin());
     }
   }
 


jngrad (approved):

src/core/io/mpiio/mpiio.cpp
```diff
@@ -446,9 +446,11 @@ void mpi_mpiio_common_read(const char *filename, unsigned fields) {
 
     for (int i = 0; i < nlocalpart; ++i) {
       int blen = boff[i + 1] - boff[i];
-      auto &bl = local_particles[id[i]]->bl;
-      bl.resize(blen);
-      std::copy_n(&bond[boff[i]], blen, bl.begin());
+      if (blen) {
+        auto &bl = local_particles[id[i]]->bl;
+        bl.resize(blen);
+        std::copy_n(&bond[boff[i]], blen, bl.begin());
```
fweik: I think it  would be better to use `std::vector::data` instead of taking the address of the first element. That makes the intent clearer, and works also with an empty vector, and hence also does not need the check.

fweik (changes requested):

src/core/io/mpiio/mpiio.cpp
mkuron: I don‘t think `bond.data() + boff[i]` is that much clearer than `&bond[boff[i]].` Most people reading code tend to be confused by pointer arithmetic like this.

fweik: I don't agree, this is not incidental, but the offset calculation is the actual logic here. Also getting rid of the control structure is worth it. We can also do iterator arithmetic instead if you think that is less scary.

hirschsn:
I tend to think that the C++ implementation used in #3230 is broken in this case. The standard states for copy_n:
"Effects: For each non-negative integer i < n, performs *(result + i) = *(first + i)."

If an implementation for copy_n asserts: `__n < this->size()`, the implementation forgot the corner case: `__n == 0`. Or am I missing something here?

Nonetheless, this PR looks good because it is an easy workaround for the problem.

src/core/io/mpiio/mpiio.cpp
```diff
@@ -446,9 +446,11 @@ void mpi_mpiio_common_read(const char *filename, unsigned fields) {
 
     for (int i = 0; i < nlocalpart; ++i) {
       int blen = boff[i + 1] - boff[i];
-      auto &bl = local_particles[id[i]]->bl;
-      bl.resize(blen);
-      std::copy_n(&bond[boff[i]], blen, bl.begin());
+      if (blen) {
```
hirschsn: I'd actually suggest a `if (blen > 0)` here, because `blen` is not a bool and not an int pretending to be.

mkuron:
@hirschsn, the assertion is not inside `std::copy_n`, but in `std::vector::operator[]`. And that's standard-compliant because subscripting beyond bounds is undefined.

src/core/io/mpiio/mpiio.cpp
mkuron: using iterator arithmetic now

fweik (approved):

hirschsn:
@mkuron I see. That makes sense. :)

Wouldn't `bond.begin() + off` where `off == 0` in this case fix the problem already?

NVM, I see that's what you've done.

mkuron:
Build is failing because of https://github.com/espressomd/espresso/issues/3235, which is fixed by https://github.com/espressomd/docker/pull/139.

jngrad:
bors r=fweik

jngrad:
bors r-
(failed CI)

jngrad:
bors r=fweik

PR #3236, opened 2019-10-07
ESS2019 installation guide updates

mkuron:
Lessons learned today:

- We require MPI 3 because we depend on const-correctness in a few places. That means that OpenMPI 1.6.5 and lower are not supported anymore.
- Installing the ROCm driver breaks access to /dev/kfd, causing hwloc initialization during `mpiexec` to hang. Rebooting helps.
- Add matplotlib, ipython and jupyter to the Mac install guide.
- Homebrew now defaults to Python 3, requires manually enabling cython, and it's unclear whether the hdf5 package still supports MPI (https://github.com/Homebrew/homebrew-core/issues/26974)
- Anaconda (~/anaconda[23]) and python.org packages (/Library/Python and /usr/local/bin) are also sources of conflict

Please tag for the 4.1.1 release

### DIFF
--- a/doc/sphinx/installation.rst
+++ b/doc/sphinx/installation.rst
@@ -114,6 +114,8 @@ ROCm SDK to make use of GPU computation:
     sudo apt update
     sudo apt install libnuma-dev rocm-dkms rocblas rocfft rocrand rocthrust
 
+After installing the ROCm SDK, please reboot your computer.
+
 .. _Installing Requirements on Mac OS X:
 
 Installing Requirements on Mac OS X
@@ -132,7 +134,8 @@ following commands:
     sudo port selfupdate
     sudo port install cmake python37 py37-cython py37-numpy \
       openmpi-default fftw-3 +openmpi boost +openmpi +python37 \
-      doxygen py37-opengl py37-sphinx py37-pip gsl hdf5 +openmpi
+      doxygen py37-opengl py37-sphinx py37-pip gsl hdf5 +openmpi \
+      py37-matplotlib py37-ipython py37-jupyter
     sudo port select --set cython cython37
     sudo port select --set python3 python37
     sudo port select --set pip pip37
@@ -145,12 +148,11 @@ Alternatively, you can use Homebrew.
     sudo xcode-select --install
     sudo xcodebuild -license accept
     /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
-    brew install cmake python@3 cython boost boost-mpi fftw \
-      doxygen gsl
-    brew install hdf5 --with-mpi
-    brew install numpy --without-python@2
-    ln -s /usr/local/bin/python2 /usr/local/bin/python
-    pip install --user PyOpenGL
+    brew install cmake python cython boost boost-mpi fftw \
+      doxygen gsl numpy ipython jupyter
+    brew install hdf5
+    brew link --force cython
+    pip install PyOpenGL matplotlib
 
 Note: If both MacPorts and Homebrew are installed, you will not be able to
 run |es|. Therefore, if you have both installed, please uninstall one
--- a/doc/sphinx/installation.rst
+++ b/doc/sphinx/installation.rst
@@ -73,8 +73,8 @@ To make |es| run on 18.04 LTS, its dependencies can be installed with:
 .. code-block:: bash
 
     sudo apt install build-essential cmake cython3 python3-numpy \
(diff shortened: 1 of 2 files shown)


jonaslandsgesell:
If the full path of the build directory contains a space then the pypresso/ipypresso scripts do not work. When setting the variable PYTHONPATH the path to the espressomd module needs to be set in quotes in the scripts. I can describe this offline.

mkuron:
@jonaslandsgesell, that‘s fixed in #3238 already.

doc/sphinx/installation.rst
```diff
@@ -114,52 +114,96 @@ ROCm SDK to make use of GPU computation:
...
+
 .. code-block:: bash
 
     sudo xcode-select --install
     sudo xcodebuild -license accept
     sudo port selfupdate
     sudo port install cmake python37 py37-cython py37-numpy \
       openmpi-default fftw-3 +openmpi boost +openmpi +python37 \
-      doxygen py37-opengl py37-sphinx py37-pip gsl hdf5 +openmpi
+      doxygen py37-opengl py37-sphinx py37-pip gsl hdf5 +openmpi \
```
KaiSzuttor: I think we should put all python dependencies into ```/requirements.txt``` and let the users install them via ```pip install -r requirements.txt```.
That would mean to add the following packages to ```/requirements.txt```:
  * Cython
  * PyOpenGL
  * numpy
  * sphinx
  * matplotlib
  * ipython
  * jupyter

KaiSzuttor (changes requested):

doc/sphinx/installation.rst
mkuron: I don't agree. If the packages are in the system package manager, they should be installed from there. I don't even have pip installed on my laptop because everything is available in MacPorts.

KaiSzuttor: we have to put the python dependencies there anyway and we are able to restrict the versions if any has a bug

KaiSzuttor: > If the packages are in the system package manager, they should be installed from there

why?

KaiSzuttor: > . I don't even have pip installed on my laptop because everything is available in MacPorts

we explicitly tell the user to install pip

pdebuyl: For what it's worth, even on Linux (where many packages are available from the system) I almost always use a venv with the packages for espresso as I find it simpler (also, no changes because of system updates). For execution only (no matplotlib etc), I install always install numpy, cython, and h5py in the venv.

mkuron: If you install other packages from the package manager that require different versions (i.e. the versions provided by the package manager), you'll run into trouble. The only way around is with virtualenv, but we definitely shouldn't confuse people with that. If we want to blacklist versions, we should do that on the CMake or Python level.

KaiSzuttor: can you come up with an example where a package from the package manager has a python package dependency if itself is not a python package?

mkuron: Removed pip references as discussed offline with @KaiSzuttor.

RudolfWeeber:
> I think we should put all python dependencies into ```/requirements.txt``` and let the users install them via ```pip install -r requirements.txt```.
> That would mean to add the following packages to ```/requirements.txt```:
I think mixing distro-provided and pip-provided packages is a source for all kind of trouble during the installation.
For distros, where it is possible, we should use distro packges alone.

RudolfWeeber:
> we explicitly tell the user to install pip
We probably shouldn't, unless there is a need to install something from pip.

RudolfWeeber:
> For what it's worth, even on Linux (where many packages are available from the system) I almost always use a venv with the packages for espresso as I find it simpler (also, no changes because of system updates). For execution only (no matplotlib etc), I install always install numpy, cython, and h5py in the venv.
I think that is only simpler for experts. For beginners its confusing to have to think about different environments for different python projects.

RudolfWeeber:
> can you come up with an example where a package from the package manager has a python package dependency if itself is not a python package?
Numpy is the classic. Using a custom numpy installed via pip for Espresso reasons and distro-provided numpy-using packages.
When everithing is distro-provided, there is a good chance its compatible with each other and across different python apps.

KaiSzuttor (approved):

jngrad:
bors r=KaiSzuttor

jngrad:
bors r-
(failed CI)

jngrad:
bors r=KaiSzuttor

PR #3239, opened 2019-10-08
Added test criteria for the charged_system-2 tutorial

reinaual:


testsuite/scripts/tutorials/test_02-charged_system-2.py
```diff
@@ -27,6 +28,22 @@
...
+        avg = 0
+        for i in range(1, 3):
+            pos = np.flatnonzero(tutorial.res[:, i] > 0)
+            self.assertGreater(tutorial.res[pos[0], 0], tutorial.wall_margin)
+            self.assertLess(tutorial.res[pos[-1], 0],
+                            tutorial.box_z - tutorial.wall_margin)
+
+            avg += np.mean(tutorial.res[pos, i])
+        avg *= 0.5
+        self.assertAlmostEqual(avg, 0.0148, delta=5e-4)
```
jonaslandsgesell: What is the origin of the magic number? Can you place the calculation in the script or refer to a source where it is published?

reinaual: It is the mean particle density per bin of the histogram which i originally planned to compare against each bin value, which is not possible without sufficient sampling. I just noticed that this mean value is constant anyways, which is only fluctuating because i neglected the 0s. I will remove this comparison since it does not give any insight into the system.

RudolfWeeber (approved):

jngrad (approved):
ran 300 times on coyote10 without failure

jngrad:
bors r=RudolfWeeber,jngrad

PR #3240, opened 2019-10-08
Benchmark: Ferrofluid benchmark (derived from LJ one)

RudolfWeeber:
I did not include it into the benchmark cmake target, as it is not such a common scenario outside the ICP


fweik (approved):

jngrad:
bors r=fweik

PR #3241, opened 2019-10-08
WIP: Stokesian dynamics

hmenke:
TODO:
- [x] Test case
- [x] Communication of rotations
- [ ] Thermalization
- [ ] Docs
- [ ] Lubrication corrections (units)


testsuite/python/stokesian_dynamics.py
```diff
@@ -0,0 +1,59 @@
...
+                x_ref = self.data[i,2*n]
+                y_ref = self.data[i,2*n+1]
+                x = self.pos[:,3*n]
+                y = self.pos[:,3*n+1]
+
+                if y_ref < -555:
+                    continue
+
+                idx = np.abs(y - y_ref).argmin()
+                self.assertTrue(np.abs(y_ref - y[idx]) < 0.5)
```
mkuron: Add an assert for x too.

mkuron:
Test case for the thermalization will be added next week. I already have it running locally, only the temperature is off by a factor of 2. Other than that, everything seems fine.

RudolfWeeber:
I think, this is a good addition to Espresso.

As long as it's marked WIP, just a few comments:
* Collecting the particles via partCfg does not work in parallel, neither do the partiel property setters defined in particle_data.hpp. You need some MPI-gather and scatter based approach, instead.
* Do I understand correctly that this requires CUDA for thrust, even with the CPU implementation?
* Can you please add some doxygen to device_matrix.hpp and similar files?

RudolfWeeber:
My understanding of the current state of affairs is the following:
This PR still needs more work in some areas. These are:
* more extensive testing
* collection of particle pos/vel/omega and back-distribution of vel and omega that do not break the synchronized parallelization scheme
* making the CPu solver independent from Cuda components (thrust)
* docs

@mkuron explained in the ESpresso offline meeting that he aand @hmenke currently do not have the time to address these issues.

Currently, most student assistants are involved with teaching. We will try to find someone to work on this PR in the term break. Since the interface to the Espresso core is quite narrow, there should not be major merge conflicts.

hmenke:
Making the CPU solver independent of Thrust is impossible, because that breaks my current single-source architecture.  I exploit the fact that Thrust's algorithms can be instantiated for either CPU or GPU through a template parameter.  I really don't want to litter my code with `#ifdef` statements.
It should also be noted that Thrust is independent of CUDA and could be imported into the ESPResSo repo as a git submodule from https://github.com/thrust/thrust.  Since it is header-only, there are also no additional build steps needed.

RudolfWeeber:
I understood from @mkuron, that it could be done in a less invasive fashion.

The dependence on thrust for the CPU version complicates matters. I think, Espresso currently uses different thrust versions depending on whether HIP or CUDA is being used as GPU-backend.

src/core/stokesian_dynamics/sd_interface.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+    std::size_t i = 0;
+    for (auto const &p : parts) {
+      id[i] = p.p.is_virtual ? -1 : p.p.identity;
+      x_host[6 * i + 0] = p.r.p[0];
+      x_host[6 * i + 1] = p.r.p[1];
+      x_host[6 * i + 2] = p.r.p[2];
+      x_host[6 * i + 3] = 0;
+      x_host[6 * i + 4] = 0;
+      x_host[6 * i + 5] = 0;
```
biermanncarl: Explicitly copy orientations from the espresso particle list for clarity. 
Also I guess it is required if particles are rotationally non-invariant (e.g. particles with a dipole)?

jngrad:
The rebased SD was merged. Closing.

PR #3244, opened 2019-10-10
Update docs 4.1

jngrad:
Provide links to Dockerfiles, fix links in 4.1 release notes, fix typos.


jngrad:
@christophlohrmann 
* Do you know why the first two lists of [`running.html#isotropic-npt-thermostat`](http://espressomd.org/html/doc4.1.0/running.html#isotropic-npt-thermostat) are an almost exact copy of [`system_setup.html#isotropic-npt-thermostat`](http://espressomd.org/html/doc4.1.0/system_setup.html#isotropic-npt-thermostat)?
* Why does `system_setup.html` give a circular link to itself instead of pointing to your far more detailed NPT discussion?
* `running.html` says all directions fluctuate by default. Function signature says `[0, 0, 0]` by default. Shouldn't we write in the docstring that `0` means dimension is not fixed and `1` is fixed?
* Why is the friction coefficient gammaV described as the piston mass Q?
  https://github.com/espressomd/espresso/blob/9dc1e0606165237913b594dc71a1f081b94d624d/src/python/espressomd/thermostat.pyx#L429-L431

christophlohrmann (approved):
Looks good to me

fweik (approved):

jngrad:
bors r=fweik

PR #3257, opened 2019-10-16
Remove PDB parser feature

jngrad:
The consensus offline at the ICP and online in #1441 is to drop support of the PDB parser feature in favor of the dedicated python package MDAnalysis.


fweik:
LGTM, @RudolfWeeber what do you think?

fweik:
Do you thinks we should consider adding more docs/examples on how to use Espresso with MDAnalysis? I'd think a lot of hours and hand-rolled code could be avoided if this was more well known...

fweik (approved):

jngrad:
Should this be part of 4.1.1? MDAnalysis is far more reliable than the pdbparser, which would break on the first occurrence of two values not separated by a whitespace (typical cases: large atom ids for ions in PDB frames extracted from Gromacs trajectories, altLoc or iCode fields in protein models).

fweik:
Thats a breaking change in a bugfix release, I think we should refrain from that.

jngrad:
bors r=fweik

jngrad:
bors r-
(failed merge on staging)

jngrad:
bors r=fweik

PR #3258, opened 2019-10-17
CMake minor fixes

jngrad:
Description of changes:
- change next milestone to 4.2
- load `GNUInstallDirs` to make standard GNU paths accessible from CMake variables
- simplify CMake logic and install in `python3.X` folder instead of the deprecated `python3` folder
- add extra check to make sure install paths are correctly configured (all python and shared object files must be inside the package `espressomd`)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -242,7 +242,7 @@ if(WITH_PYTHON)
     execute_process(
       COMMAND
         ${PYTHON_EXECUTABLE} -c
-        "import distutils.sysconfig as cg; print(cg.get_python_lib(1,0,prefix='${CMAKE_INSTALL_EXEC_PREFIX}'))"
+        "import distutils.sysconfig as cg; print(cg.get_python_lib(prefix='${CMAKE_INSTALL_PREFIX}', plat_specific=True, standard_lib=False).replace('${CMAKE_INSTALL_PREFIX}/', '', 1))"
       OUTPUT_VARIABLE PYTHON_INSTDIR
       OUTPUT_STRIP_TRAILING_WHITESPACE)
   endif(NOT PYTHON_INSTDIR)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,6 +20,7 @@
 
 cmake_minimum_required(VERSION 3.4)
 include(FeatureSummary)
+include(GNUInstallDirs)
 project(ESPResSo)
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -20,6 +20,7 @@
 
 cmake_minimum_required(VERSION 3.4)
 include(FeatureSummary)
+include(GNUInstallDirs)
```
fweik, line 23: Probably the block
```cmake
if(NOT DEFINED BINDIR)
  set(BINDIR "bin")
endif(NOT DEFINED BINDIR)
```
in line 401 should be removed now. Did you check that the paths provided by `GNUInstallDirs` are used for all install targets? I did a quick grep and did find nothing, but I only checked the `install` commands, not sure
if this effects rpaths and so on....

jngrad: Thanks! I overlooked that one. The `install()` commands have to be configured such that submodules get installed in the python `espressomd` folder, if not the new test in c43d057 will detect that (it searches for misplaced files in `CMAKE_INSTALL_PREFIX`). The GNU variables inherit from `CMAKE_INSTALL_PREFIX`, so we should use them as often as possible.

fweik (approved):

jngrad:
I've a few more improvements.

jngrad:
The failures in CI are actually a bug in the build system, not a regression. Working on it.

CMakeLists.txt
```diff
@@ -241,7 +242,7 @@ if(WITH_PYTHON)
     execute_process(
       COMMAND
         ${PYTHON_EXECUTABLE} -c
-        "import distutils.sysconfig as cg; print(cg.get_python_lib(1,0,prefix='${CMAKE_INSTALL_EXEC_PREFIX}'))"
+        "import distutils.sysconfig as cg; print(cg.get_python_lib(prefix='${CMAKE_INSTALL_PREFIX}', plat_specific=True, standard_lib=False).replace('${CMAKE_INSTALL_PREFIX}/', '', 1))"
```
jngrad, line 245: @junghans this change might affect your packaging workflow

junghans: Thanks for asking, but it won't as `PYTHON_INSTDIR` is set for packaging, so this call isn't actually used.

jngrad:
bors r=fweik

PR #3259, opened 2019-10-17
cmake: generate version.hpp at compile time to avoid rebuilds

junghans:
Make version.hpp a proper target. This prevents CMake from rebuilding all source files when a minor change is introduced anywhere in the docs, tests, or CMakeLists.txt files. Significantly reduce the number of source files including version.hpp to limit the number of objects to rebuild when the commit or branch changes.


fweik:
@jngrad can you look at this? I'm not sure if this at all the right way to capture the version, that was kind of an ad hoc solution to fix a problem.

jngrad:
The current solution fails CI because `@PROJECT_VERSION@` expands to an empty string in this branch, so h5md cannot extract the version from the derived macro `ESPRESSO_VERSION`. I'll have a look.

junghans:
@fweik fixed.

jngrad:
@junghans let me know if we can proceed with the merge. As a side-note, merging python into a PR isn't necessary in the absence of merge conflicts, `bors` will do it automatically.

junghans (commented):
LGTM

fweik (approved):

jngrad:
bors r=fweik

PR #3261, opened 2019-10-18
Make version.hpp a proper dependency

jngrad:
Fixes #3259

This prevents CMake from rebuilding all source files when a minor change is introduced anywhere in the docs, tests, or CMakeLists.txt files.


fweik (approved):

junghans:
This PR won't help if you change a file after running cmake, e.g.
```
cmake ..
git stash pop
make
```

cmake won't get run another time and hence the `version.hpp` will still say `CLEAN` and not `DIRTY`.

jngrad:
You're right, I've to rework my solution. Probably starting from your PR.

jngrad:
Merged into #3259.

PR #3268, opened 2019-10-21
core: Removed external force switching flags

fweik:
Description of changes:
 - Removed flags to switch external force/torque on and
   of per particle. This does not actually improved performance
   and creates some noise.
- Not user visible.


src/core/particle_data.hpp
```diff
@@ -62,20 +62,10 @@ enum {
 /** \ref ParticleProperties::ext_flag "ext_flag" value for particle subject to
  *  an external force
  */
-#define PARTICLE_EXT_FORCE 1
-/** \ref ParticleProperties::ext_flag "ext_flag" value for fixed coordinate
- *  coord. */
```
jngrad, line 67: did you remove the correct docstring?

fweik: No, I'll address that in a separate pull request.

jngrad (approved):
LGTM

fweik:
bors r=jngrad

Issue #3271, opened 2019-10-22
Merge the two MinimizeEnergy interfaces


jngrad:
There are currently two Python interfaces for the steepest descent (SD) algorithm in the core:
* `MinimizeEnergy`: has type checks, default values, can be checkpointed, no interruption handling
* `Integrator`: has no type checks, no default values, can be checkpointed, interruption handling

When using `Integrator`, the previous integrator state is overridden in both the core and interface, and needs to be restored after the minimization ends. When using `MinimizeEnergy`, the previous integrator state is overridden only in the core, which means running `system.integrator.run()` afterwards is ill-defined. The structure of class `Integrator` is also quite heterogeneous and not easily extensible. In particular, the numerous `if self._method == X` introduce too much coupling between the different types of integrators.

It would make more sense to refactor `integrate.pyx` with the design of `interactions.pyx`: an abstract class `Integrator` from which `SD`/`NPT`/`VV` inherit, and a class `CurrentIntegrator` (used as `System.integrator = CurrentIntegrator()`) that contains  an instance of `Integrator`. Adding or removing an integrator algorithm would only be a matter of writing/removing a class.

RudolfWeeber:
Agreed. Furthermore, the term 'minimize energy' should not appear, since that is not, what it does.

fweik:
I think if we are considering a (major) interface change here, we should evaluate if the integrator should really be a property of the system. Mixing algorithmic aspects with the definition of the system is not co-aligned with the mental model suggested by statistical mechanics. I've always disliked how the current interface mixes the declaration of what the system is with what you do to it.
If we are willing to break the interface we should put some thought into make it good for once.

To be more verbose: The `System` class  is more incidental than it is designed or thought out. What does it mean? The `System` in the model sense (e.g. the `Hamiltonian` or the `Liouvillian`)? The simulation model? The current state of that model? It has to be clear that those basic elements of the API are what Espresso means to the users, so it is something we should put serious consideration into. It is the language in which people formulate their thesis work in, their research.
I'm by no means an expert user of the software, but I've never heard somebody calling it's interface intuitive, and it stands to reason that this would mean that the interface is not catering to peoples expectations. I've worked with and on the software for more years than I'd care to admit, and still I could not answer really what Espressos _terms_ mean. I know it's easy to dismiss such points as idealistic or esoteric, but I'm convinced that if we try to define concepts that in a way closer to science in our simulation tool we could make even more friends and users.  

[...]

PR #3274, opened 2019-10-23
Fix velocity Verlet setter

jngrad:
Not for direct merging.

Fixes the bug found in #3271

When setting up a system with the NpT integrator, then switch to the VV integrator, the system keeps the NpT integrator in the core. To reproduce the error, take e.g. `samples/visualization_npt.py` and replace the call to `system.integrator.set_isotropic_npt` by the following:
```
system.integrator.set_isotropic_npt(ext_pressure=1.0, piston=0.01)
print(system.box_l)
system.integrator.set_vv()  # script interface: velocity Verlet, core: NpT
system.integrator.run(100)
print(system.box_l)
exit(0)  # exit now to avoid starting the visualizer
```
output:
```
[10. 10. 10.]
[10.00003915 10.00003915 10.00003915]
```

### DIFF
--- a/src/python/espressomd/integrate.pyx
+++ b/src/python/espressomd/integrate.pyx
@@ -50,6 +50,8 @@ cdef class Integrator:
         self._method = state['_method']
         if self._method == "STEEPEST_DESCENT":
             self.set_steepest_descent(state['_steepest_descent_params'])
+        elif self._method == "VV":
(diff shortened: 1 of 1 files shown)


RudolfWeeber (approved):

jngrad:
Merged in 4.1.1 as 0865e1dae4

PR #3276, opened 2019-10-24
Clean-up data_parts for ghost communication

fweik:
Description of changes:
 - Removed data_parts left-overs from cell systems
 - Tuned GHOSTTRANS_* constants into enum
 - Calculate data_parts for vs relative on the fly.

### DIFF
--- a/src/core/ghosts.hpp
+++ b/src/core/ghosts.hpp
@@ -125,22 +125,21 @@ further details.
 /** \name Transfer data classes, for \ref GhostCommunication::type */
 /************************************************************/
 /*@{*/
-/// transfer \ref ParticleProperties
-#define GHOSTTRANS_PROPRTS 1
-/// transfer \ref ParticlePosition
-#define GHOSTTRANS_POSITION 2
-/// transfer \ref ParticleMomentum
-#define GHOSTTRANS_MOMENTUM 8
-/// transfer \ref ParticleForce
-#define GHOSTTRANS_FORCE 16
-
-/// resize the receiver particle arrays to the size of the senders
-#define GHOSTTRANS_PARTNUM 64
-
-#ifdef ENGINE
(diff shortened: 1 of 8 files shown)


src/core/ghosts.hpp
```diff
@@ -125,22 +125,21 @@ further details.
...
-#define GHOSTTRANS_FORCE 16
-
-/// resize the receiver particle arrays to the size of the senders
-#define GHOSTTRANS_PARTNUM 64
-
-#ifdef ENGINE
-/// transfer \ref ParticleParametersSwimming
-#define GHOSTTRANS_SWIMMING 128
-#endif
+enum : unsigned {
```
jngrad, line 126: type mismatch: change `GhostCommunication::type` to `unsigned int`
https://github.com/espressomd/espresso/blob/3923430b5cd560ba523292068d40dc92d04d7014/src/core/ghosts.hpp#L149-L151

src/core/ghosts.hpp
```diff
@@ -125,22 +125,21 @@ further details.
 /** \name Transfer data classes, for \ref GhostCommunication::type */
 /************************************************************/
 /*@{*/
```
jngrad: since the values are now grouped inside an enum, you can remove the keyword `\name`, the `/****************/` and the block delimiters `/*@{*/` `/*@}*/`

jngrad (changes requested):

src/core/ghosts.hpp
fweik: Those are unrelated

jngrad: Nevermind, I misread `GHOSTTRANS_*` values as `GHOST_*` values

jngrad (approved):
Thanks!

fweik:
bors r=jngrad

PR #3277, opened 2019-10-24
Removed ENGINE shear torque calculation

fweik:
@RudolfWeeber @christophlohrmann is this what you had in mind?
If so the simplification is quite significant because all of the custom
communication code for ENGINE can be removed. Also the LB coupling
should now work in parallel.

Description of changes:
 - Removed ENGINE shear torque calculation
 - Moved swimmer parameters into ParticleProperties
 - Removed custom communiication code for ENGINE
 - Adaped docs


christophlohrmann:
Yes that's the one, thank you.
The ``engine_lb.py`` test has two occurrences of ``rotational_friction`` which can be removed, too:
``test_rotfric_exception()`` in line 137 and
``test_particle_torques()`` in line 164

fweik:
Oh, then why did the test pass?

RudolfWeeber (approved):

fweik:
bors r=RudolfWeeber

PR #3279, opened 2019-10-25
Replaced some manual memory allocations

reinaual:
Partially adresses #2900

### DIFF
--- a/src/core/PartCfg.hpp
+++ b/src/core/PartCfg.hpp
@@ -53,12 +53,12 @@ public:
       return {make_indirect_iterator(begin), make_indirect_iterator(begin)};
     }
 
-    auto begin =
-        skip_it(local_particles.data(), local_particles.data() + max_seen_particle + 1,
-                SkipIfNullOrGhost());
-    auto end =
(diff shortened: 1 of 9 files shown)


RudolfWeeber:
@reinaual please make the static analysis happy

jngrad:
@reinaual Let's discuss it tomorrow, it might be a bug that went un-noticed by Clang until now. Also, the Clang version on the ICP machines seems to output different errors.

jngrad:
Summary of Wednesday's offline discussion: the change from an array of pointers to a vector of vectors helped Clang detect a scenario where the array would be used un-initialized. The exact sequence of conditional branches that leads to this scenario is however complex to analyze in the current state of `fft.cpp`, and the path doesn't seem to be taken in the multiple tests of the testsuite (probably due to the low coverage of the `if (not group)` branch).

@reinaual if you run out of ideas, I suggest you partially revert b867d6a831, e.g. by merging 06122545c4. This way your contribution will remain in the history and can be restored in the future, when the code that initializes the FFT arrays becomes more readable.

jngrad (approved):
Thanks! This makes the code more readable.

PR #3287, opened 2019-10-31
LB Boundary force: fix unit conversion and tests

RudolfWeeber:
Reference solutions for forces on shear planes and for stokes drag need dynamic viscosity, but LB uses kinetmatic viscosity.
This switches the tests to dynamic viscosity and removes a 1/rho from the unit conversion in the script interface.
Also, a force balance test is added, which does not depend on the density.

Fixes #

Description of changes:
 - 

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?


testsuite/python/lb_shear.py
```diff
@@ -98,9 +97,14 @@ def check_profile(self, shear_plane_normal, shear_direction):
         """
         self.system.lbboundaries.clear()
         self.system.actors.clear()
+        try:
```
fweik: Why is the try/catch needed?

fweik (approved):

PR #3291, opened 2019-11-04
Factor out and test memcpy serialization

fweik:
Description of changes:
 - This splits off the memcpy serialization used in the ghost communication and adds unit test
   for it.
 - Added customization point to opt-in to memcpy serialization for types that are not [TriviallyCopyable](https://en.cppreference.com/w/cpp/named_req/TriviallyCopyable), but serializes
 to a bounded size, like `boost::optional`. This is so we can use `optional` and `variant` in `ParticleProperties`, which is planned.

### DIFF
--- a/src/core/ghosts.cpp
+++ b/src/core/ghosts.cpp
@@ -34,9 +34,11 @@
 #include "errorhandling.hpp"
 #include "particle_data.hpp"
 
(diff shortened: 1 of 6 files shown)


RudolfWeeber (approved):

jngrad:
bors r=RudolfWeeber

Issue #3242, opened 2019-10-08
Shell linter


jngrad:
We've recently had a few issues troubleshooting bash scripts that relied on the inherent flexibility of the shell syntax, resulting in silent bugs that went undetected, verbose bugs that took a while to correctly diagnose, and shell scripts that are generally hard to maintain (#3238, #2932, #2998).

Our shell scripts could be made less brittle by following some of the recommendations in the Google [Shell Style Guide](https://google.github.io/styleguide/shell.xml) and using Dash with POSIX syntax in simple scripts that don't require Bash-specific features. The task of modernizing all our shell scripts can be assisted by the [`shellcheck`](https://github.com/koalaman/shellcheck) linter. It can also detect common POSIX violations when using option `--shell=dash` and looking for rule [`SC2039`](https://github.com/koalaman/shellcheck/wiki/SC2039); Bash scripts with few to no violations can be candidates for a downgrade to Dash. Below are a few rules that are relevant to the shell scripts in the 4.1 release. `shellcheck` does not have a whitelisting feature though, which would be helpful to prevent the re-introduction of specific errors in CI.

Dangerous errors:

* [`SC2148`](https://github.com/koalaman/shellcheck/wiki/SC2148): Missing shebang
* [`SC2061`](https://github.com/koalaman/shellcheck/wiki/SC2061): Quote the parameter to -name, e.g. `find . -name *.so` to `find . -name '*.so'`
* [`SC2166`](https://github.com/koalaman/shellcheck/wiki/SC2166): Change `if [ a -o b ]` to `if [ a ] || [ b ]`: POSIX standard for utilities `test` and `[` is unspecified when called with >4 arguments ([`test`/OPERANDS in IEEE Standard 1003.1](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html))

Readability:
* [`SC2116`](https://github.com/koalaman/shellcheck/wiki/SC2116): Useless echo, e.g. `f=$(echo "/path/${filename}")` to `f="/path/${filename}"`
* [`SC2002`](https://github.com/koalaman/shellcheck/wiki/SC2002): Useless cat, e.g. `lines=$(cat a.log | wc -l)` to `lines=$(wc -l < a.log)`
* [`SC2006`](https://github.com/koalaman/shellcheck/wiki/SC2006): Legacy subshell syntax, change ``var=`..` `` to `var=$(..)`

Miscellaneous:
[...]

PR #3292, opened 2019-11-04
Pypresso: Remove "test -a" and quote $@

hirschsn:
Description of changes:
 - Remove `test -a` since calls to test with more than 4 arguments are unspecified: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html
 - Quote `$@` to avoid splitting $@. Currently `./pypresso --gdb "test me.py"` will fail because tries to find the file `test`. Quoting avoids this.

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?

### DIFF
--- a/src/python/pypresso.cmakein
+++ b/src/python/pypresso.cmakein
@@ -14,13 +14,13 @@ else
 fi
 export PYTHONPATH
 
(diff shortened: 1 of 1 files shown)


jngrad (approved):
Thanks! We really need to work on #3242.

Issue #3093, opened 2019-08-21
Decide on minimum required versions for dependencies after 4.1


RudolfWeeber:

jngrad:
* Python (currently >= 3.3), Cython (currently >= 0.23)
* CMake: discussion in #3090
* boost: versions below 1.67 make the [CMake logic convoluted](https://github.com/espressomd/espresso/blob/d14acdb8642e3546b95a4a3ad9880df6422bf0b5/CMakeLists.txt#L378-L403),  dockerfiles require patches
* python packages: numpy, scipy, pylint, vtk, etc. should be written to [`requirements.txt`](https://github.com/espressomd/espresso/blob/python/requirements.txt)
* MPI via OpenMPI: depending on the OS, a different minimum version is required

jngrad:
Concerning version numbers for python dependencies, the version numbers of the corresponding packages in the Ubuntu 16.04 repository (which we currently support in the Clang 6 image) would offer a conservative estimate, unless stated otherwise, for example `pint` is pinned to a higher version due to numpy-related compatibility bugs in the version available on Ubuntu.

hmenke:
Also #1644 requires at least Boost 1.65.1, because it's now using Boost.Spirit X3 which is relatively new.

jngrad:
Update: a discussion took place in [es meeting 2019-11-26](https://github.com/espressomd/espresso/wiki/Espresso-meeting-2019-11-26)
We'll probably need to increase the boost requirements in two steps: first 1.58, which we have to support until the Ubuntu 20.04 LTS release (we need to support Ubuntu 16.04 LTS until then), then 1.65. Python versions 3.3 and 3.4 are only supported on CentOS SCLo.

jngrad:
Once we get rid of `clang:6.0` in CI (which uses 1.58), we should be able to bump to 1.65, which is used by `osx` in CI. We'll then be able to simplify CMake and remove all `#if BOOST_VERSION`. Python itself and Python packages will need to be kept in sync with NEP 29 (#3421), which should not be difficult considering we support Ubuntu LTS versions for 3 years.

RudolfWeeber:
Based on Ubuntu 20:
* Remove Cuda 9 asupport
* Cuda 11 with coverage in every build
* CMake 3.16
* Cython 0.29
* Python 3.6
* Boost 1.66

Drop Centos build. Clusters use custom compilers and libraries via modules anyway.

jngrad:
[...]

PR #3300, opened 2019-11-11
List Python3 dependencies

jngrad:
Follow-up to #3093

List the minimal version numbers that are supported for dependent Python modules.

Please note `requirements.txt` does not have to list strict dependencies, optional packages are also allowed. Comments after a package name must have a whitespace before the `#` symbol.

### DIFF
--- a/doc/sphinx/contributing.rst
+++ b/doc/sphinx/contributing.rst
@@ -99,9 +99,9 @@ safely removed if you're planning on installing dependencies via ``pip``:
 
 .. code-block:: bash
 
-   pip install --upgrade --user -r requirements.txt
+   pip3 install --upgrade --user -r requirements.txt
 
-Note that some distributions now use ``pip`` for Python3 and ``pip2`` for
+Note that some distributions now use ``pip3`` for Python3 and ``pip2`` for
 Python2.
 
 
--- a/doc/sphinx/contributing.rst
+++ b/doc/sphinx/contributing.rst
@@ -84,22 +84,26 @@ Required Development Tools
    distributed versioning control system git_ and a GitHub account to fork the
    `espressomd/espresso <https://github.com/espressomd/espresso>`_ repository
 
--  To build the sphinx documentation, you will need the Python packages listed
-   in :file:`requirements.txt` in the top-level source directory. To install
-   them, use:
-
-   .. code-block:: bash
-
-      pip install --upgrade --user -r requirements.txt
-
-   Note that some distributions now use ``pip`` for Python3 and ``pip2`` for
-   Python2.
+-  To build the user documentation, you will need Sphinx_.
 
 -  To build the tutorials, you will need LaTeX and Jupyter.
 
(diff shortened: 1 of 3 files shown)


jngrad:
You can check a list of dependencies against your environment with the following script:
```python
import pip
from pip._internal.req.req_file import parse_requirements
from pip._internal.network.session import PipSession
import pkg_resources
from pkg_resources import DistributionNotFound, VersionConflict

def check_dependencies():
    """
    Checks if python dependencies are fullfilled.
    If check passes, return 0. Otherwise, print error and return 1.
    Please note that after the first missing module is found, pkg_resources
    will ignore all other missing packages!
    Source: https://github.com/pypa/pip/issues/2733#issuecomment-257340871
    """
    requirement_file_name = 'requirements.txt'
    dependencies = []
    session = PipSession()
    for req in parse_requirements(requirement_file_name, session=session):
        if req.req is not None:
            dependencies.append(str(req.req))
        else:
            # The req probably refers to a url. Depending on the
            # format req.link.url may be able to be parsed to find the
            # required package.
            pass
    try:
        pkg_resources.working_set.require(dependencies)
    except VersionConflict as e:
        try:
            print("{} was found on your system, "
                  "but {} is required.\n".format(e.dist, e.req))
            return 1
        except AttributeError:
            print(e)
            return 1
    except DistributionNotFound as e:
        print(e)
        return 1
    print('Check was successful')
    return 0

exit(check_dependencies())
```

mkuron:
With this significantly expanded requirements.txt file, we should remove the recommendation to install it from https://github.com/espressomd/espresso/blob/df08343e108f9d53cfda759c3341d6386adbcc6c/doc/sphinx/contributing.rst#required-development-tools. We already removed it from the user documentation a few weeks ago. Nobody wants to install a dozen packages plus dependencies if they don't actually need them.

doc/sphinx/contributing.rst
```diff
@@ -84,22 +84,26 @@ Required Development Tools
...
--  To build the tutorials, you will need LaTeX and Jupyter.
+You can find all Python dependencies of |es| in :file:`requirements.txt` in the
+top-level source directory. Several optional packages for graphics, external
+devices and continuous integration (CI) are not strictly essential and can be
+safely removed if you're planning on installing dependencies via ``pip``:
 
--  To compile the Doxygen code documentation, you will need the tool doxygen_.
+.. code-block:: bash
+
+   pip install --upgrade --user -r requirements.txt
```
mkuron: Change this to `pip3`. That seems to be the default name nowadays.

mkuron (approved):

doc/sphinx/contributing.rst
jngrad: Thanks!

hmenke: `python3 -m pip` is also a possibility.

fweik (approved):

jngrad:
bors r=fweik

PR #3310, opened 2019-11-14
core: Fixed ELC potential difference prefactor + test

fweik:
Description of changes:
 - When using ELC/MMM2d with potential boundary conditions, one
   of the correction factors was over-counted resulting in wrong energies.
 
The other term in the correction in L 348 of elc.cpp also looks suspicious,
but I could not find a straight forward way to test that. All of this code should
only be used with the utmost caution.


RudolfWeeber (approved):
Does it also work for mmm2d?

fweik:
Yes, I fixed both. Now ELC has an independent test and mmm2d is checked against ELC. So for this particular situation there is a test now.

fweik:
bors r=RudolfWeeber

PR #3321, opened 2019-11-17
P3m halo refactor

fweik:
- Factor out almost identical halo comm from p3m and dp3m
- Switch from odd/even communication to MPI_Sendrecv
- Support communicating multiple meshes at once.


jngrad (approved):
LGTM

fweik:
bors r=jngrad

PR #3326, opened 2019-11-18
Update maintainer and CI shell scripts

### LINKED ISSUE #3242: Shell linter

jngrad:
Fixes #3242

Description of changes:
- reduce number of shellcheck warnings from 238 down to 104
   - rules: `SC2002 SC2006 SC2061 SC2063 SC2086 SC2116 SC2148 SC2166 SC2196`
   - 50 of the remaining warnings are false positives from `build_cmake.sh` (`SC2086 SC2154`)
- refactor outdated maintainer shell scripts
- remove deprecated shell syntax and shell commands
- use POSIX alternatives to bashisms when the change is trivial
- quote string variables to guard against whitespace-related issues
- start unit tests before integration tests
- run unit tests for all operating systems (Debian, OpenSUSE, Fedora)

Notes:
- `shellcheck` is unsuitable for CI in its current form (e.g. no rule whitelisting capability)
- the shell scripts now have an homogeneous syntax and few linter warnings, which is desirable if we decide to port them to Python in the future

### DIFF
--- a/maintainer/CI/dox_warnings.sh
+++ b/maintainer/CI/dox_warnings.sh
@@ -69,8 +69,8 @@ grep -Prn '^[\ \t]*(?:\*?[\ \t]+)?[@\\]t?param(?:\[[in, out]+\])?[\ \t]+[a-zA-Z0
 rm -f dox_warnings.log
 
 # process warnings
-${srcdir}/maintainer/CI/dox_warnings.py
-n_warnings=${?}
+"${srcdir}/maintainer/CI/dox_warnings.py"
+n_warnings="${?}"
 
 if [ ! -f dox_warnings.log ]; then
     exit 1
@@ -80,10 +80,10 @@ fi
 cat dox_warnings.log
 
 if [ "${CI}" != "" ]; then
-    "${srcdir}/maintainer/gh_post_docs_warnings.py" doxygen ${n_warnings} dox_warnings.log || exit 1
+    "${srcdir}/maintainer/gh_post_docs_warnings.py" doxygen "${n_warnings}" dox_warnings.log || exit 1
 fi
 
-if [ ${n_warnings} = "0" ]; then
+if [ "${n_warnings}" = "0" ]; then
     echo "Found no warning requiring fixing."
     exit 0
 else
--- a/maintainer/CI/dox_warnings.sh
+++ b/maintainer/CI/dox_warnings.sh
@@ -19,13 +19,13 @@
 
 # This script generates a log of all Doxygen warnings that require fixing.
 
-dox_version=`doxygen --version`
+dox_version=$(doxygen --version)
 if [ "${dox_version}" != "1.8.13" ] && [ "${dox_version}" != "1.8.11" ]; then
     echo "Doxygen version ${dox_version} not supported"
     exit 1
 fi
 
-[ -z "${srcdir}" ] && srcdir=`realpath ..`
+[ -z "${srcdir}" ] && srcdir=$(realpath ..)
 
 # prepare Doxyfile for the XML version
 cp doc/doxygen/Doxyfile doc/doxygen/Doxyfile.bak
--- a/maintainer/CI/dox_warnings.sh
+++ b/maintainer/CI/dox_warnings.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
 # Copyright (C) 2019 The ESPResSo project
 #
 # This file is part of ESPResSo.
--- a/maintainer/CI/dox_warnings.sh
+++ b/maintainer/CI/dox_warnings.sh
@@ -70,7 +70,7 @@ rm -f dox_warnings.log
 
 # process warnings
 ${srcdir}/maintainer/CI/dox_warnings.py
-n_warnings=$?
+n_warnings=${?}
 
 if [ ! -f dox_warnings.log ]; then
     exit 1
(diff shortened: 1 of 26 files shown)


fweik:
Is it possible to run the scripts in the CI in POSIX compatibility mode (e.g. using `sh`, and not `bash`)?
This would be a simple test that they actually are POSIX compatible.

jngrad:
Not really. For example `build_cmake.sh` has 12 bashisms. I've already downgraded scripts that could run in dash with trivial changes. For the less trivial bashisms, I could have a look if you want, but some of the known POSIX equivalents (see [SC2039](https://github.com/koalaman/shellcheck/wiki/SC2039)) don't look very maintainable.

jngrad:
I've removed more bashisms, but several shell scripts still rely on useful bash-specific features:
* variable indirections don't seem to have a POSIX equivalent without `eval`
* shell traps: SIGINT, SIGTERM and SIGQUIT automatically trigger SIGEXIT in bash, while in POSIX one has to write a custom trap for each one of them (e.g. a SIGINT trap must forward SIGINT)
* `$BASH_COMMAND` and `$LINENO` don't exist in POSIX (useful for debug messages)
* arrays only exist natively in bash
* `build_cmake.sh`  downloads and executes a codecov script which is not POSIX-compliant

List of bash scripts:
```sh
head -n 1 $(find maintainer/ testsuite/ -name "*.sh")
```
```
maintainer/CI/build_cmake.sh
maintainer/CI/build_docker.sh
maintainer/CI/doc_warnings.sh
maintainer/benchmarks/runner.sh
maintainer/benchmarks/suite.sh
testsuite/cmake/BashUnitTests.sh
testsuite/cmake/test_install.sh
```

fweik:
I also don't think it is so super important. At the end of the day most of those scripts only run in an environment that we control.

maintainer/CI/dox_warnings.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
```
mkuron, line 1: `/usr/bin/env sh` is useless as `/usr/bin/env` is not guaranteed by POSIX (neither is `/bin/sh`, by the way). I would just use `/bin/sh`.

mkuron (approved):
Looks fine. We should really migrate this mess to Python scripts ASAP though.

maintainer/CI/dox_warnings.sh
jngrad: Here the idea was to be more portable, as the program `env` reads the user `$PATH`. I put this change in the POSIX bucket for convenience, since most scripts modified in that commit had heterogenous shebangs that needed fixing, although it has nothing to do with POSIX. In fact, `sh` might not even point to the `dash` interpreter.

KaiSzuttor (approved):
LGTM, it seems like we develop a DevOps library in bash without knowing it :smile:

jngrad:
bors r=KaiSzuttor

PR #3328, opened 2019-11-19
Python arguments

pdebuyl:
Following the email discussion, @bindgens1 made the changes to adopt a consistent calling convention for the sign of the function distance_vec. In C++, it goes along "get_mi_vector(a, b) returns a-b". The Python interface in system.pyx changed the order of the arguments, which makes the debugging of testcases very hard.

With this pull request, all usage of `espressomd.system.distance_vec` returns the vector "position of first argument - position of second argument", modulo minimal image of course.

The consistency will be very useful to use when we add "velocity_difference" with the Lees-Edwards feature.

Description of changes:
 - Modify the calling convention in system.pyx for the function distance_vec
 - Modify the tests in consequence

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?

### DIFF
--- a/src/python/espressomd/system.pyx
+++ b/src/python/espressomd/system.pyx
@@ -445,7 +445,7 @@ cdef class System:
 
         """
 
-        cdef Vector3d mi_vec = get_mi_vector(make_Vector3d(p2.pos), make_Vector3d(p1.pos), box_geo)
+        cdef Vector3d mi_vec = get_mi_vector(make_Vector3d(p1.pos), make_Vector3d(p2.pos), box_geo)
 
         return make_array_locked(mi_vec)
 
--- a/testsuite/python/dpd.py
+++ b/testsuite/python/dpd.py
@@ -427,7 +427,7 @@ class DPDThermostat(ut.TestCase):
         stress = np.zeros([3, 3])
 
         for pair in pairs:
-            dist = s.distance_vec(pair[0], pair[1])
+            dist = s.distance_vec(pair[1], pair[0])
             if np.linalg.norm(dist) < r_cut:
                 vel_diff = pair[1].v - pair[0].v
                 stress += calc_stress(dist, vel_diff)
--- a/testsuite/python/interactions_bond_angle.py
+++ b/testsuite/python/interactions_bond_angle.py
@@ -108,14 +108,14 @@ class InteractionsAngleBondTest(ut.TestCase):
             for p, sign in [[p1, -1], [p2, +1]]:
                 # Check that force is perpendicular
                 self.assertAlmostEqual(
-                    np.dot(p.f, self.system.distance_vec(p0, p)), 0,
+                    np.dot(p.f, self.system.distance_vec(p, p0)), 0,
                     delta=1E-12, msg="The force is not perpendicular")
                 # Check that force has correct magnitude
                 self.assertAlmostEqual(np.linalg.norm(p.f), np.abs(
                     f_ref) / self.system.distance(p0, p), delta=1E-11)
                 # Check that force decreases the quantity abs(phi - phi0)
                 if np.abs(phi_diff) > 1E-12:  # sign undefined for phi = phi0
-                    force_axis = np.cross(self.system.distance_vec(p, p0), p.f)
+                    force_axis = np.cross(self.system.distance_vec(p0, p), p.f)
                     self.assertEqual(
                         np.sign(phi_diff),
                         np.sign(sign * np.dot(force_axis, self.axis)),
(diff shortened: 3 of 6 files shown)


RudolfWeeber:
Sorry, I think I missed a previous discussion here.

Two points:
* Changing the argument order for Python interface functions is a rather nasty silent interface change. I don't see how we can merge this at present.
* Personally, I find what the distance_vec Python function does the intuitive outcome. I'd assume distance_vec(a,b) to point from a to b. This may be a matter of taste, though.

I agree, that distance_vec and velocity_difference should use the same convention in the Python interface. Since velocity_difference was not yet present, it can be changed without breaking anyone's scripts silently.

I also agree that consistency between Python interface functions and core is desirable, but I don't think this particular case is worth silently breaking user scripts in a hard-to-debug fashion.

I'm sorry that I apparently missed something and did not speak up earlier.

pdebuyl:
Hi @RudolfWeeber 

The email included you, @fweik , @KaiSzuttor , and @jngrad  (and @bindgens1 and me of course).

The reason that we raised it is that we couldn't debug our Lees-Edwards simulations: we were reading our program and the C++ code together. It was only by going to the interface file system.pyx that we realized the mismatch between the C++ code and the python code.

Writing "diff_{ij} = element_i - element_j" is, I believe much more common. In Allen-Tildesley, the notation "r_{ij}" stands for "r_i - r_j", "v_{ij} = v_i - v_j", etc. (Also in the sample programs). Frenkel-Smit uses the same convention.

The code also changed recently (june 2019) without notice in the opposite direction, in https://github.com/espressomd/espresso/pull/2902

RudolfWeeber:
> Writing "diff_{ij} = element_i - element_j" is, I believe much more common. In Allen-Tildesley, the notation "r_{ij}" stands for "r_i - r_j", "v_{ij} = v_i - v_j", etc. (Also in the sample programs). Frenkel-Smit uses the same convention.
Matter of taste in the end. I don't have a strong opinion about what is being used, just about the silent interface change.

> The code also changed recently (june 2019) without notice in the opposite direction, in https://github.com/espressomd/espresso/pull/2902 
Are you sure?
```
@@ -442,12 +435,10 @@ cdef class System(object):
         """Return the distance vector between the particles, respecting periodic boundaries.
 
         """
-        cdef double[3] res, a, b
-        a = p1.pos
-        b = p2.pos
 
-        get_mi_vector(res, b, a)
-        return np.array((res[0], res[1], res[2]))
+        cdef Vector3d mi_vec = get_mi_vector(make_Vector3d(p2.pos), make_Vector3d(p1.pos), box_geo)
+
+        return make_array_locked(mi_vec)
 
```
Looks like unchanged behavior to me. Otherwise that would be a regression.

I understand that the discrepancy in in the sign convention between core and Python interface is annoying.
Still, I maintain that this is not significant enough to risk that users produce wrong results due to silent interface changes.

A silent interface change could be avoided by introducing a new function with a new name and raising an exception explaining the changed sign convention when calling the old function.
Personally I think, it is not worth bothering users for a different sign convention between interface and core, but if more peolpe speak up for a non-silent interface change, I would also not be opposed.

fweik:
My 2 cents: I think the convention is `distance(a, b) = b - a`, e.g. the vector pointing from a to b (c.f. 
[Affine Space](https://en.wikipedia.org/wiki/Affine_space), also notation in [Euclidean Distance](https://en.wikipedia.org/wiki/Euclidean_distance)). `get_mi_vector` is a cryptic name for `distance` and hence should be `b - a`. Also many of these functions have bad names, which does not help. So I think if you want to make it consistent, the order in python should be kept, while the function in the core should be changed and renamed, e.g. to `minimum_image_distance`. I agree with @RudolfWeeber that silent breaking interface change in python is not an option.

jngrad:
In my Oct. 26 response to the email I recommended changing the order of parameters in the core to make them consistent with the python interface, as long as it did not change the interface. Sorry if it was not clear; the examples I gave (angle potentials and tabulated distance) are probably the source of the confusion. What I meant was, changing the order of the function arguments is fine if you also swap the two arguments everywhere in the function body to avoid a sign flip in the return value.

pdebuyl:
Hi all,

Would you be ok with a new function espressomd.system.mi_vector ? This would obviously have to match the convention for the core routine and we could use it for the Lees-Edwards tests with (for us) more clarity.

pdebuyl:
Maybe to add to our motivation: the current convention causes [cognitive dissonance](https://en.wikipedia.org/wiki/Cognitive_dissonance) for us. In debugging Lees-Edwards, we have to keep in mind the C++ code, the Python tests, the various layers in between, and the physics of the problem already. Using counter intuitive and counter common usage argument convention makes it even harder.

This type of issue is now well recognized in the teaching and practice of scientific computing. Greg Wilson is an excellent advocate for these considerations. Here, a talk from EuroSciPy  https://www.youtube.com/watch?v=kmVKGxPlTvc (he also has interesting books and articles on the topic).

RudolfWeeber:
Having two functions in the system class which do the same just with different sign causes more confusion for the code overall than it removes from a Lees-Edwards specific test.
Moreover, in the le_experimental branch on my Github repo, I think, the testing is mostly complete.

I agree with @fweik that the clean solution would be to update the core.
However, there are >100 call sites and the potential for collateral damage is not small.
Not sure, if we are going to do it soon.

I vote to not pursue this topic further at present.

pdebuyl:
My opinion on changing the core is that it should not be changed. The core should remain the stablest part of ESPResSo of course, as the particles' trajectories depend on it.

For the rest, whatever.

KaiSzuttor:
can/should this be closed? @RudolfWeeber

PR #3329, opened 2019-11-19
Fix compiler warning on empty config with GPU

jngrad:


fweik (approved):

fweik:
Warnings are still not errors in CUDA...

mkuron:
> Warnings are still not errors in CUDA...

Correct, `nvcc` does not have a `-Werror` flag.

PR #3331, opened 2019-11-20
Testcase for ELC and MMM2D

reinaual:
Analytical testcase for ELC and MMM2D

### DIFF
--- a/testsuite/python/CMakeLists.txt
+++ b/testsuite/python/CMakeLists.txt
@@ -207,6 +207,7 @@ python_test(FILE linear_momentum.py)
 python_test(FILE linear_momentum_lb.py MAX_NUM_PROC 2 LABELS gpu)
 python_test(FILE mmm1d.py MAX_NUM_PROC 2 LABELS gpu)
 python_test(FILE elc.py MAX_NUM_PROC 2)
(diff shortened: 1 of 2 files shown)


jngrad:
Once the pylint issues are fixed, you can probably squash all commits into a single one.

fweik (changes requested):
I think this a good start. I think you should also test the case where the system is not neutral (e.g. the two charges don't have equal magnitude). There is some interaction with implicit background charge that is present in P3M in that case, which should also be test. And please use a different charge magnitude than 1. in the neutral case. I think you can just add an additional loop around the body of the `scan` function the sets different charges. The modification of the analytical solution should be straight forward.

RudolfWeeber:
Wasn't the non-neutral case the one that didn't work and that we disabled?
IIRC, we so far didn't succeed in fixing it.

fweik:
Well, we'll find out soon. Also the potential difference case should be tested. So far this only tests dielectric contrast.

jngrad:
Offline discussion with @reinaual: no analytical solution for the potential difference case. If a non-analytical approximation can be found, a new PR will be opened. This test can already be shipped in the 4.1.2 release.

fweik (approved):

jngrad:
bors r=fweik

PR #3340, opened 2019-11-26
core: Removed MMM2d

fweik:
In the offline meeting we decided to keep the near-field part of mmm2d. There is no straight forward way to keep the n2 version, also it seems broken. Also MMM2D is available from the scafacos libary,
so this removes all of MMM2D. Take it or leave it.


KaiSzuttor:
ping

RudolfWeeber:
I looked into the Scafacos implementation. It has dielectric contrast. It does, however, not have documentation. 
There are some tests for mmm2d.

RudolfWeeber (approved):

PR #3341, opened 2019-11-26
core: thermostat: Removed heat_up/cool_down mechanism for Langevin ty…

fweik:
…pe thermostats

The Langevin-type thermostats (excluding LB), had broken initial forces.
They did still employ the heat_up/cool_down mechanism which is no longer
correct with the philox rng. This is potentially a serious bug, because it makes
the thermostat silently act with a different temperature than set for at least
parts of the time steps.


KaiSzuttor:
why hast that been caught by any testcase?

fweik:
Because it was not tested.

KaiSzuttor:
Shouldn't the DPD test for the velocity distribution fail if the noise magnitude is wrong?

RudolfWeeber:
You have to trigger the heat_up/cool_down by forcing a recalculation of the initial foces, e.g., via

System.part[0].pos=system.part[0].pos

KaiSzuttor (approved):

PR #3342, opened 2019-11-26
INSTALL: use cmake generics

junghans:


argoyang:
hello,I am not good at espressomd. I have some question about cmake&boost.
My system is Ubuntu16.04. when I cmake,  Errors are as follow:  How can I deal with it? Thank you!
(the boost version 1.67 )
src/python/espressomd/CMakeLists.txt:85 (add_library): Target "espressomd_integrate" links to target "Boost::mpi" but the target   was not found.  Perhaps a find_package() call is missing for an IMPORTED   target, or an ALIAS target is missing?
CMake Error at src/python/espressomd/CMakeLists.txt:85 (add_library):  Target  espressomd_magnetostatics" links to target "Boost::serialization"  but the target was not found.  Perhaps a find_package() call is missing for  an IMPORTED target, or an ALIAS target is missing?

mkuron (approved):

KaiSzuttor (approved):

PR #3347, opened 2019-11-29
Disable P3M benchmark test

### LINKED ISSUE #2924: P3M benchmark randomly fails CI
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/pipelines/7734

jngrad:
Closes #2924

### DIFF
--- a/testsuite/scripts/benchmarks/CMakeLists.txt
+++ b/testsuite/scripts/benchmarks/CMakeLists.txt
@@ -16,7 +16,7 @@ add_custom_target(local_benchmarks
                   COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_SOURCE_DIR}/maintainer/benchmarks ${BENCHMARKS_DIR})
 
 benchmark_test(FILE test_lj.py)
-benchmark_test(FILE test_p3m.py)
+#benchmark_test(FILE test_p3m.py)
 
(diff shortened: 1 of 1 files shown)


fweik:
Hm, if you just comment it out, than it will decay and stop working. I think it would be better to either make it deterministic by fixing the parameters or remove it altogether.

jngrad:
> I think it would be better to either make it deterministic by fixing the parameters

All the information I got from coworkers, from code introspection and from debugging sessions where parameters and random seeds are fixed, are consistent with a non-deterministic P3M tuning algorithm, for a continuous period of 6 months. More specifically, the bisection uses a heuristic that depends on the exact timing of all previous iterations, and these timings are affected by all other threads running on the same CI machine. I don't see what you have in mind here. Didn't you once suggest we should re-implement the P3M tuning in Python with a different algorithm to make it more deterministic?

> it will decay and stop working

True. Let's wait until #3326 is merged, then we can disable the benchmark tests only on the subset of CI jobs where it fails.

fweik:
I meant fixing the P3M parameters instead of tuning them. The tuning as it stands is non-deterministic  because it depends on the (uncontrolled) run time.

jngrad:
That could work on a single machine. Getting the P3M parameters once and re-using them for all other myconfig files would make the benchmark actually useful. But can we come up with P3M parameters that would work on any machine?

PR #3349, opened 2019-12-02
core: random: Factored out v_noise

fweik:
Description of changes:
 - Factored out v_noise function
 - Removed unused code


testsuite/python/dpd.py
```diff
@@ -462,7 +462,7 @@ def test_momentum_conservation(self):
         momentum = np.matmul(s.part[:].v.T, s.part[:].mass)
         for _ in range(10):
             s.integrator.run(25)
-            np.testing.assert_array_less(np.zeros((3, 3)), np.abs(s.part[:].f))
+            np.testing.assert_almost_equal(np.zeros((3,)), np.sum(s.part[:].f))
```
jngrad, line 465: shouldn't it be `np.sum(np.abs(s.part[:].f), axis=1)`?

fweik: Why `abs`? I want to check that the total force is zero.

jngrad: I misunderstood the test logic.

jngrad (approved):
LGTM

PR #3355, opened 2019-12-04
Remove Langevin noise correlation

### LINKED ISSUE #3343: Correlated noise
The langevin thermostat uses maximally correlated noise for the translation and rotation thermalization.

jngrad:
Fixes #3343

Description of changes:
- decoupled translational and rotational Langevin RNG seeds
- add a test for Langevin noise correlation
- cleanup docs

### DIFF
--- a/src/core/random.hpp
+++ b/src/core/random.hpp
@@ -49,6 +49,7 @@ enum class RNGSalt : uint64_t {
   FLUID = 0,
   PARTICLES,
   LANGEVIN,
+  LANGEVIN_ROT,
   SALT_DPD,
   THERMALIZED_BOND
 };
--- a/src/core/random.hpp
+++ b/src/core/random.hpp
@@ -23,9 +23,8 @@
 #define RANDOM_H
 
 /** \file
-
-    A random generator
-*/
+ *  Random number generation using Philox.
+ */
 
 #include "errorhandling.hpp"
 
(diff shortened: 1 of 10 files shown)


jngrad:
@RudolfWeeber I can't get `rotational-diffusion-aniso.py` to pass on my machine without noise correlation, even after restoring the test to its state before your speed-up modifications. The outcome of the test also seems to be very sensitive on the numpy seed. You might want to have a look if CI fails.

bindgens1:
I also had problems with the rotation tests in the modifications that @fweik did in #2957 
More samples for the histogram didn't help solving the problem. However, I have no idea, if the issues are related to one another or if this is an independent issue.

RudolfWeeber:
Please use kT <=1.5
@psci2195, do you know why the test doesn't work for higher temperatures?
Is there an assumption about Peclet numbers in the theory?

jngrad:
@RudolfWeeber the failing `dpd.py` test does not fail on my machine, even within a docker container. The test is also extremely sensitive to the seed of the Langevin thermostat.

fweik:
The failure is due to bad statistics. See #3360, there is no problem. Also this is just thermalized degrees of freedom friction-coupled to other thermalized degrees of freedom, so this should work. This is no different than coupling to a thermalized LB fluid.
I think you can just cherry-pick 05daf01e6760a682ba0d7f42f84bc15d030f7252.
I checked multiple seeds, this is all very robust. The test in principle now also passes with 20 loops,
or with a almost tenfold lower error.

jngrad:
@fweik @KaiSzuttor Thanks!
@RudolfWeeber Ready for review

RudolfWeeber (approved):

jngrad:
bors r=RudolfWeeber

PR #3364, opened 2019-12-06
maintainer: Moved reaction ensemble code out of core

fweik:
The reaction ensemble is pure client code, and should not be in the core. Also this improves the build time for the python modules (at little bit).


jngrad (approved):
pending CI

fweik:
This seems related to https://bugs.llvm.org/show_bug.cgi?id=39191, caused by a forward declaration.

RudolfWeeber:
Please defer until there has been time for offline discussion.

PR #3367, opened 2019-12-08
py: ParticleSlice fixes

### LINKED ISSUE #3363: ParticleSlice forgets particles in member access
looks like in the constructor the index of id_selection is used in masking rather than the value

RudolfWeeber:
* Return correct elements if particle numbering is non-contigeous
* Prevent slicing with negative indices due unclear outcome
* Maintain order created by slicing, e.g.[9:0:-1]
* Require that all particles exist when passing an explicit list of ids
* Maintain order for explicit list of ids (system.part[[3,2,5,1]]
* More rigorous testing

Fixes #3363

### DIFF
--- a/src/python/espressomd/particle_data.pyx
+++ b/src/python/espressomd/particle_data.pyx
@@ -156,7 +156,7 @@ cdef class ParticleHandle:
 
         def __get__(self):
             self.update_particle_data()
-            return make_array_locked(unfolded_position(< Vector3d > self.particle_data.r.p, < Vector3i > self.particle_data.l.i, box_geo.length()))
+            return make_array_locked(unfolded_position( < Vector3d > self.particle_data.r.p, < Vector3i > self.particle_data.l.i, box_geo.length()))
 
     property pos_folded:
         """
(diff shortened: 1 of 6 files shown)


RudolfWeeber:
@KaiSzuttor, based on the offline discussion, I inserted a few comments which hopefully make things more clear

src/python/espressomd/particle_data.pyx
```diff
@@ -1556,14 +1556,43 @@ cdef class _ParticleSliceImpl:
...
-        self.id_selection = id_list[slice_]
-        mask = np.empty(len(self.id_selection), dtype=np.bool)
-        cdef int i
-        for i in range(len(self.id_selection) - 1, -1, -1):
-            mask[i] = particle_exists(i)
-        self.id_selection = self.id_selection[mask]
+        # We distinguish two cases:
+        # * ranges and slices only specify lower and upper bounds for particle
+        #   ids and optionally a step. Gaps in the id range are tolerated.
+        # * Explicit list/tuple/ndarrya containing particle ids. Here
```
KaiSzuttor: typo

src/python/espressomd/particle_data.pyx
```diff
@@ -1556,14 +1556,43 @@ cdef class _ParticleSliceImpl:
...
+            self.id_selection = self.id_selection[mask]
+        elif isinstance(slice_, (list, tuple, np.ndarray)):
+            # Check that all entries are some flavor of integer
+            for id in slice_:
+                if not is_valid_type(id, int):
+                    raise TypeError(
+                        "Particle id must be an integer but got " + str(id))
+                if not particle_exists(id):
+                    raise IndexError("Particle does not exist " + str(id))
+            self.id_selection = np.array(slice_, dtype=int)
```
KaiSzuttor: I think all this code does not belong here. We should put the slice validation into some dedicated place like ```utils```. It should also be tested.

src/python/espressomd/particle_data.pyx
```diff
@@ -1556,14 +1556,43 @@ cdef class _ParticleSliceImpl:
...
+        else:
+            raise TypeError(
+                "ParticleSlice must be initialized with an instance of slice or range, or with a list, tuple, or ndarray of ints, but got " +
```
KaiSzuttor: we should probably use ```"some string {}".format(some_variable)```

src/python/espressomd/particle_data.pyx
```diff
@@ -2026,8 +2048,7 @@ Set quat and scalar dipole moment (dipm) instead.")
         # Ids of the selected particles
         ids = []
         # Did we get a function as argument?
-        if len(args) == 1 and len(kwargs) == 0 and isinstance(
-                args[0], types.FunctionType):
+        if len(args) == 1 and len(kwargs) == 0 and callable(args[0]):
```
KaiSzuttor, line 2064: remark: classes are callable

KaiSzuttor (changes requested):
LGTM besides minor comments

src/python/espressomd/particle_data.pyx
RudolfWeeber: Moved to individaul member functoins in ParticleSlice

src/python/espressomd/particle_data.pyx
RudolfWeeber: Done

src/python/espressomd/particle_data.pyx
RudolfWeeber: Didn't find an other way to handle Cython oddities.

src/python/espressomd/particle_data.pyx
```diff
@@ -1555,15 +1555,57 @@ cdef class _ParticleSliceImpl:
...
+        #   all particles have to exist and the result maintains the order
+        #   specified in the list.
+        if isinstance(slice_, (slice, range)):
+            self.id_selection = self._id_selection_from_slice(slice_)
+        elif isinstance(slice_, (list, tuple, np.ndarray)):
+            self._validate_pid_list(slice_)
+            self.id_selection = np.array(slice_, dtype=int)
+        else:
+            raise TypeError(
+                "ParticleSlice must be initialized with an instance of slice or range, or with a list, tuple, or ndarray of ints, but got {} of type {}".foormat((str(slice_), str(type(slice_)))))
```
KaiSzuttor: typo, is this raise tested?

RudolfWeeber:
> +            raise TypeError(
> +                "ParticleSlice must be initialized with an instance of slice or range, or with a list, tuple, or ndarray of ints, but got {} of type {}".foormat((str(slice_), str(type(slice_)))))
> typo, is this raise tested?
The spell checker in vim doesn't find the typo. Neither do I (but I'm not good at finding typos).
It's not clear which particular illegal type I should test here. The if is more narrow than the cases that would probably work (anything that is iterable and only contains int-likes). I saw no benefit in allowing more, though.
The more realistic errors such as lists containing non-int-likes are checked. [1, n_part/2]. There was at least one such regression in the testsuite which went undetected before this pr.

src/python/espressomd/particle_data.pyx
jngrad: the typo is `.foormat`

KaiSzuttor (approved):

KaiSzuttor:
it seems like kodiak is missing an essential feature, namely updating PRs 😢

fweik:
@KaiSzuttor @jngrad So, kodiak isn't up to the job?

KaiSzuttor:
yes, see https://github.com/chdsbd/kodiak/issues/104

PR #3375, opened 2019-12-11
Update requirements

### LINKED ISSUE #3090: Bump required cmake version to 3.10
Cmkae fails for 3.8 and older.

jngrad:
Increase requirements of CMake, boost, Cython and Python, as discussed in [es meeting 2019-11-26](https://github.com/espressomd/espresso/wiki/Espresso-meeting-2019-11-26).

Closes #3090, Partial fix for #3093

Depends on espressomd/docker#149

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -93,17 +93,6 @@ default:
     - docker
     - linux
 
-min_boost:
-  <<: *global_job_definition
-  stage: build
-  image: gitlab.icp.uni-stuttgart.de:4567/espressomd/docker/ubuntu-python3:min_boost
-  script:
-    - export with_cuda=false myconfig=maxset
-    - bash maintainer/CI/build_cmake.sh
-  tags:
-    - docker
-    - linux
-
 maxset:
   <<: *global_job_definition
   stage: build
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -228,10 +228,10 @@ if(WITH_CUDA)
   endif()
 endif(WITH_CUDA)
 
-find_package(PythonInterp 3.3 REQUIRED)
+find_package(PythonInterp 3.5 REQUIRED)
 
 if(WITH_PYTHON)
-  find_package(Cython 0.23 REQUIRED)
+  find_package(Cython 0.26 REQUIRED)
   execute_process(
     COMMAND ${PYTHON_EXECUTABLE} -c
             "import distutils.sysconfig as cg; print(cg.get_python_inc())"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,12 +18,12 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
-cmake_minimum_required(VERSION 3.4)
+cmake_minimum_required(VERSION 3.10)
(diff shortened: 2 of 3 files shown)


.gitlab-ci.yml
```diff
@@ -93,17 +93,6 @@ default:
     - docker
     - linux
 
-min_boost:
```
KaiSzuttor, line 96: why do you want to remove this test?

jngrad: Here's how I understood our last es meeting:
* we don't test boost 1.55 any more
* boost 1.58 is already tested in `cuda:9.0` and `clang:6.0`
* once Ubuntu 20.04 LTS is released next April, we upgrade boost requirements to 1.65 (for `cuda:9.0` which requires Ubuntu 16.04, we build boost 1.65 from sources)
*  except for building boost from sources, `Dockerfile-min_boost` = `Dockerfile-18.04`, so we can transfer it from master to a new 4.1 branch

fweik: We could think about having one container where all the library versions are pinned to the minimal required versions, so that we don't silently stop testing this when a library version is updated in the distribution. This isn't really complicated, all major package managers let you just fix a version.

jngrad: `cuda:9.0` will essentially fulfill that role. It uses Ubuntu 16.04 and the minimal supported versions of CUDA and CMake. The version numbers of non-graphical Python packages available in the Ubuntu 16.04 repository are used as the minimal versions in `requirements.txt` (fb4862ed894e7). Once Ubuntu 20.04 is out and espresso drops support for CUDA9, we can use the `cuda:10.1` to fulfill that role. We could rename the CI job from `cuda10:maxset` to `ubuntu:min-dependencies` for clarity.

fweik: Whatever you think is appropriate. I'm just saying that it is not as such guaranteed that the package versions in the distributions stay the same, which could lead to a situation where the minimal version of some package is no longer tested.

jngrad: We could come up with an extra check in the CI job that tests the docker image to make sure the versions of dependencies match the minimal requirements.

RudolfWeeber (approved):

RudolfWeeber:
My understanding is that we increase minimum requirements on Boost to 1.65, once we get rid of the Ubuntu 16.04 build. That in turn depends on us being able to give up on cuda 9.
This will be pursued offline.

fweik:
@jngrad did we decide how to deal with the insufficient cmake versions? Unfortunately, I don't remember...

jngrad:
@fweik in [es meeting 2019-11-26](https://github.com/espressomd/espresso/wiki/Espresso-meeting-2019-11-26) we decided to move to CMake from Ubuntu 18.04 and install it via pip in Ubuntu 16.04 docker images. This PR will pass CI in ~2 hours, when the new Ubuntu 16.04 images get deployed.

fweik:
Ah, right. Thank you.

jngrad:
Looks like kodiak won't merge a PR if the last CI pipeline passed while the PR was still in draft mode. Once the PR is marked for review, simply re-run the `status_success` job to make kodiak happy.

PR #3380, opened 2019-12-12
Fix particle-id based observables with non-contigeous particle numbering

### LINKED ISSUE #3362: Some observables return values for wrong particles
because they use the index in the id list rather than the entry.
Working on it.

RudolfWeeber:
Fixes #3362 

* Fixes some pid obervables which were using the index in the list of particle ids rather than the particle id.
* Test pid-based observables with non-contigeous particle ids
* Remove the mass prefactors from ComForce, so that it calculates what is claimed i the docstring.

I did not rename ComForce, since this is for a bugfix release. Later it should become TotalForce.

### DIFF
--- a/src/core/observables/ComForce.hpp
+++ b/src/core/observables/ComForce.hpp
@@ -34,9 +34,9 @@ public:
     for (int i : ids()) {
       if (partCfg[i].p.is_virtual)
         continue;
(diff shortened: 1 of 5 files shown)


jngrad (approved):
LGTM

jngrad:
bors r=jngrad p=1910603306

PR #3386, opened 2019-12-20
ROCm 3.0 support

mkuron:
They dropped `${HIP_ROOT_DIR}/include` from the default include path. Also, I have changed compiler warnings to be consistent with the host compiler.


jngrad (approved):
LGTM

PR #3391, opened 2019-12-27
utils: Test for triangle normal calculation.

fweik:
Description of changes:
 - Document the sign convention for the triangle functions
 - Test normal calculation.


jngrad (approved):
LGTM

PR #3392, opened 2020-01-01
Quaternion formalism comments and docs

### LINKED ISSUE #2964: Some alignment of quaternion formalism with the referenced works

bogdan-tanygin:
Fixes #2964

Description of changes:
 - code minor comment
 - docs minor change
 - bib ref to both

PR Checklist
------------
 - [ ] Tests?
   - [ ] Interface
   - [ ] Core 
 - [ ] Docs?

### DIFF
--- a/doc/doxygen/bibliography.bib
+++ b/doc/doxygen/bibliography.bib
@@ -399,7 +399,10 @@ doi = {10.1016/0021-9991(85)90151-2},
   author={Allen, Michael P and Tildesley, Dominic J},
   year={2017},
   publisher={Oxford university press},
-  url={https://pdfs.semanticscholar.org/7356/504a471e3b0cc5e21c0dd4e00aa5f9cdbdf4.pdf},
+  url={https://global.oup.com/academic/product/computer-simulation-of-liquids-9780198803201},
+  doi={10.1093/oso/9780198803195.001.0001},
+  isbn={9780198803195},
+  edition={2nd},
 }
 
 @Article{swope92a,
--- a/doc/doxygen/bibliography.bib
+++ b/doc/doxygen/bibliography.bib
@@ -394,6 +394,14 @@ year = {1985},
 doi = {10.1016/0021-9991(85)90151-2},
 }
 
+@book{allen2017,
+  title={Computer simulation of liquids},
+  author={Allen, Michael P and Tildesley, Dominic J},
+  year={2017},
+  publisher={Oxford university press},
+  url={https://pdfs.semanticscholar.org/7356/504a471e3b0cc5e21c0dd4e00aa5f9cdbdf4.pdf},
+}
+
 @Article{swope92a,
 author = {Swope, William C. and Ferguson, David M.},
 title = {{Alternative expressions for energies and forces due to angle bending and torsional energy}},
--- a/doc/sphinx/running.rst
+++ b/doc/sphinx/running.rst
@@ -184,7 +184,7 @@ Rotational degrees of freedom and particle anisotropy
 When the feature ``ROTATION`` is compiled in, particles not only have a position, but also an orientation that changes with an angular velocity. A torque on a particle leads to a change in angular velocity depending on the particles rotational inertia. The property :attr:`espressomd.particle_data.ParticleHandle.rinertia` has to be specified as the three eigenvalues of the particles rotational inertia tensor.
 
 The rotational degrees of freedom are also integrated using a velocity Verlet scheme.
-The implementation is based on a quaternion representation of the particle orientation and described in :cite:`omelyan98`.
+The implementation is based on a quaternion representation of the particle orientation and described in :cite:`omelyan98` with quaternion components indexing made according to the formalism :cite:`allen2017`.
    
 When the Langevin thermostat is enabled, the rotational degrees of freedom are also thermalized.
 
--- a/doc/sphinx/running.rst
+++ b/doc/sphinx/running.rst
@@ -184,7 +184,7 @@ Rotational degrees of freedom and particle anisotropy
 When the feature ``ROTATION`` is compiled in, particles not only have a position, but also an orientation that changes with an angular velocity. A torque on a particle leads to a change in angular velocity depending on the particles rotational inertia. The property :attr:`espressomd.particle_data.ParticleHandle.rinertia` has to be specified as the three eigenvalues of the particles rotational inertia tensor.
 
 The rotational degrees of freedom are also integrated using a velocity Verlet scheme.
-The implementation is based on a quaternion representation of the particle orientation and described in :cite:`omelyan98` with quaternion components indexing made according to the formalism :cite:`allen2017`.
+The implementation is based on a quaternion representation of the particle orientation and described in :cite:`omelyan98` with quaternion components indexing made according to the formalism :math:`q = a + b\mathbf{i} + c\mathbf{j} + d\mathbf{k}` :cite:`allen2017`.
    
 When the Langevin thermostat is enabled, the rotational degrees of freedom are also thermalized.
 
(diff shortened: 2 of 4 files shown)


doc/doxygen/bibliography.bib
```diff
@@ -394,6 +394,14 @@ @Article{sonnenschein85a
 doi = {10.1016/0021-9991(85)90151-2},
 }
 
+@book{allen2017,
+  title={Computer simulation of liquids},
+  author={Allen, Michael P and Tildesley, Dominic J},
+  year={2017},
+  publisher={Oxford university press},
+  url={https://pdfs.semanticscholar.org/7356/504a471e3b0cc5e21c0dd4e00aa5f9cdbdf4.pdf},
```
jngrad:
```suggestion
  url={https://global.oup.com/academic/product/computer-simulation-of-liquids-9780198803201},
  doi={10.1093/oso/9780198803195.001.0001},
  isbn={9780198803195},
  edition={2nd},
```

doc/sphinx/zrefs.bib
```diff
@@ -688,6 +688,14 @@ @article{omelyan98
 URL = {https://aip.scitation.org/doi/abs/10.1063/1.168642}
 }
 
+@book{allen2017,
+  title={Computer simulation of liquids},
+  author={Allen, Michael P and Tildesley, Dominic J},
+  year={2017},
+  publisher={Oxford university press},
+  URL={https://pdfs.semanticscholar.org/7356/504a471e3b0cc5e21c0dd4e00aa5f9cdbdf4.pdf},
```
jngrad:
```suggestion
  url={https://global.oup.com/academic/product/computer-simulation-of-liquids-9780198803201},
  doi={10.1093/oso/9780198803195.001.0001},
  isbn={9780198803195},
  edition={2nd},
```

doc/sphinx/running.rst
```diff
@@ -184,7 +184,7 @@ Rotational degrees of freedom and particle anisotropy
 When the feature ``ROTATION`` is compiled in, particles not only have a position, but also an orientation that changes with an angular velocity. A torque on a particle leads to a change in angular velocity depending on the particles rotational inertia. The property :attr:`espressomd.particle_data.ParticleHandle.rinertia` has to be specified as the three eigenvalues of the particles rotational inertia tensor.
 
 The rotational degrees of freedom are also integrated using a velocity Verlet scheme.
-The implementation is based on a quaternion representation of the particle orientation and described in :cite:`omelyan98`.
+The implementation is based on a quaternion representation of the particle orientation and described in :cite:`omelyan98` with quaternion components indexing made according to the formalism :cite:`allen2017`.
```
jngrad: We should probably have the formalism written explicitly at least once in the user guide:
```suggestion
The implementation is based on a quaternion representation of the particle orientation and described in :cite:`omelyan98` with quaternion components indexing made according to the formalism :math:`q = a + b\mathbf{i} + c\mathbf{j} + d\mathbf{k}` :cite:`allen2017`.
```

jngrad (changes requested):
The failed CI jobs have been restarted, it was a runner issue on our side. Sorry for the inconvenience.

The first 10 merge commits don't seem to add history and make the git graph more difficult to read. What would you think of rebasing the branch to have ee4232f0b4cc after the last commit in the python branch?

Beside the proposed changes, this PR looks good to me.

bogdan-tanygin:
> The first 10 merge commits don't seem to add history and make the git graph more difficult to read. What would you think of rebasing the branch to have ee4232f after the last commit in the python branch?

in general, I agree. However, these merge commits are distributed over the whole 2019th year. Hence, they do not confuse the history graph so much (they are not a continuous sequence) and, also, I'm not sure how to rebase this safely (I've tried and have some suspicious feeling, maybe you could advise the safe commands). Nevertheless, I agree, next time I will not create new branches from my main branch, rather I will checkout directly from the Es. base. Also, I will merge updates using the --ff-only in case of conflicts absence. Hence, this will remove such commits further.

jngrad:
> Hence, they do not confuse the history graph so much

The issue with merge commits is they prevent me from cherry-picking your branch into the 4.1.3 branch. When this happens, I need to manually inspect the graph to see which commits belong to your branch vs. which commits come from elsewhere to craft a valid sequence of `git cherry-pick` commands.

> I've tried and have some suspicious feeling, maybe you could advise the safe commands

The canonical way is to use the 3-commit `rebase --onto` feature: new base, old base, tip of the branch. The annoying part is that this command puts the result in a detached HEAD. Here's how to handle it:
```sh
git checkout quat-formalism-comments
# rebase commit range
git rebase --onto upstream/python ee4232f0b4c^ HEAD
# we are now in a detached HEAD, we'll store it in a new branch
git branch quat-formalism-comments-rebased
git checkout quat-formalism-comments-rebased
# optional: squash the 3 peer-review commits
git reset --soft HEAD~2
git commit --amend -m 'Apply code review feedback

Co-Authored-By: Jean-Noël Grad <jgrad@icp.uni-stuttgart.de>'
# push force
git push -u -f origin quat-formalism-comments-rebased:quat-formalism-comments
```
This can also be achieved with a cherry-picking with git v1.7+:
```sh
git checkout -b quat-formalism-comments-rebased upstream/python
git cherry-pick ee4232f0b4c^..quat-formalism-comments
# optional: squash the 3 peer-review commits
git reset --soft HEAD~2
git commit --amend -m 'Apply code review feedback

Co-Authored-By: Jean-Noël Grad <jgrad@icp.uni-stuttgart.de>'
# push force
git push -u -f origin quat-formalism-comments-rebased:quat-formalism-comments
```

I can do it for you if you want.

The squashing makes the code review history shorter. GitHub provides the [`Add suggestion to batch`](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/incorporating-feedback-in-your-pull-request) option to apply all code review suggestions in a single commit. This also avoids triggering the CI pipeline multiple times.

jngrad (approved):

bogdan-tanygin:
Hi @jngrad , thanks for the best practices sharing! It's useful.

PR #3394, opened 2020-01-02
WIP: Build on Windows

RudolfWeeber:
There are still linking issues with boost.

So far:
* MSVC compiler switch handling
* Some shell escape coorections in CMake for Ninja (which Visual Studio uses for building). They are probably also needed for Ninja on Linux
* Removal of some posix specifics
* Minor C++ adjustments. The MSVC compiler is slightly less tolerant about ambiguities than Clang and G++. Also, there are some differences in the libraries. E.g. math constants have to be activated separately.


RudolfWeeber:
Supporting MSVC allows one to sue the vcpkg package manger, which easily provideds boost-mpi etc.
Testing will follow via Github actions. The build can be run from the command line.
For now, I opened the PR to see if there are regressions on Linux.

CMakeLists.txt
```diff
@@ -333,7 +333,7 @@ endif(WITH_VALGRIND_INSTRUMENTATION)
 # MPI
 #
 
-find_package(MPI 3.0 REQUIRED)
+find_package(MPI 2.0 REQUIRED)
```
mkuron, line 336: we do actually use some MPI 3 functions. Perhaps Microsoft MPI has partial MPI 3 support, otherwise you would be seeing compile errors?

RudolfWeeber:
According to the comiit messages, we require MPI 3.0 for const-correctness reasons.

 

According to https://github.com/microsoft/Microsoft-MPI they implement a subset of the 3.1 standard. We’ll probably need MS-specific handling in CMake.

src/script_interface/accumulators/Correlator.hpp
```diff
@@ -11,8 +11,8 @@
  * (at your option) any later version.
  *
  * ESPResSo is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
```
jngrad, line 15: lines 14 and 15 in the license agreement have been accidentally exchanged

fweik:
I think we should add a CI test with MSVC, e.g. using appveyor, before moving forward with this. This is especially important here as, to my knowledge, very few developers use Windows/MSVC.

RudolfWeeber:
I’ll provide the test using github actions, once the Windows build actually works. There are still some issues with undefined symbols.

I opend the PR now, to check for regressions.

Once the tests are there, we probably can get a binary distributable of Espresso for Windwows, as well.

KaiSzuttor:
any news here?

KaiSzuttor:
@RudolfWeeber do you still work on this?

fweik:
We decided offline to drop native windows support because suppport for WSL2 sufficient.

Issue #3215, opened 2019-09-26
Use AST parser in the tutorial test logic


jngrad:
Replace as many of the regex patterns in [`importlib_wrapper.py`](https://github.com/espressomd/espresso/blob/python/testsuite/scripts/importlib_wrapper.py) by AST expressions.

PR #3406, opened 2020-01-08
samples: make drude sample deterministic.

KaiSzuttor:

### DIFF
--- a/samples/drude_bmimpf6.py
+++ b/samples/drude_bmimpf6.py
@@ -62,6 +62,7 @@ args = parser.parse_args()
 
 print("\nArguments:", args)
 
+np.random.seed(42)
 # NUM PARTICLES AND BOX
 n_ionpairs = 100
 n_part = n_ionpairs * 2


fweik (approved):

jngrad (commented):
I'm pretty sure you need to add the `random_seeds=False` argument in the drude sample test, e.g.:
```python
    "@SAMPLES_DIR@/drude_bmimpf6.py", gpu=gpu, random_seeds=False,
    n_max_steps=1000, n_timing_steps=10, n_int_cycles=5, **kwargs)
```
Also, did you run this sample test in a loop on a coyote machine to check if it failed after a few hundred runs? The issue comes from the P3M tuning algorithm, which is not deterministic.

fweik:
@jngrad some of the failures are from particle placement, which is fixed by this. So I think it is at least an improvement, even if it is not a complete fix.

KaiSzuttor:
that's why this PR does not close the issue

jngrad:
There is still the issue of numpy seeds being removed from samples in CI. Have a look at the contents of `build/testsuite/scripts/samples/local_samples/drude_bmimpf6_cpu_processed.py`

KaiSzuttor:
hm... at least 2/4 core devs did not know. There is an issue with the truck factor here.

jngrad:
At the end of the [sample tests wiki section](https://github.com/espressomd/espresso/wiki/Testing#tutorials-samples-benchmarks), I can change

> Given the stochastic nature of the scripts (NumPy and ESPResSo RNG seeds are assigned randomly)

to

> Given the stochastic nature of the scripts (NumPy and ESPResSo RNG seeds will be re-assigned randomly, unless specified otherwise)

I think the idea was to discourage the use of magic values in the seeds.

We'll have the opportunity to update that wiki page once I'm done with #3215 (hopefully this week).

PR #3414, opened 2020-01-14
doc: improve electrophoresis sample.

KaiSzuttor:
- added observable version for persistence length analysis and corrected python analysis
- improved sampling

old plot shows wrong raw data:

persistence_length_old

new plot:

persistence_length

### DIFF
--- a/samples/electrophoresis.py
+++ b/samples/electrophoresis.py
@@ -17,39 +17,46 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 """
-Simulate electrophoresis of a linear polymer using the P3M solver.
+Simulate electrophoresis of a linear polymer using the P3M electrostatics solver.
 """
+import logging
+
+import matplotlib.pyplot as plt
+import numpy as np
+import scipy.optimize
+
 import espressomd
+import espressomd.observables
+import espressomd.polymer
+from espressomd import electrostatics, interactions
+
+logging.basicConfig(level=logging.INFO)
+
 
 required_features = ["P3M", "EXTERNAL_FORCES", "WCA"]
 espressomd.assert_features(required_features)
 
-from espressomd import interactions
-from espressomd import electrostatics
-import numpy as np
+N_SAMPLES = 1000
+N_INT_STEPS = 100
+E_FIELD = 1.0
+N_MONOMERS = 20
+N_IONS = 100
+WARM_STEPS = 20
+WARM_N_TIMES = 20
+MIN_DIST = 0.9
 
-# System parameters
-#############################################################
 
 system = espressomd.System(box_l=3 * [100.0])
 
-# Seed
-#############################################################
-system.set_random_state_PRNG()
-#system.seed = system.cell_system.get_state()['n_nodes'] * [1234]
-np.random.seed(seed=system.seed)
-
 system.time_step = 0.01
 system.cell_system.skin = 0.4
-system.periodicity = [True, True, True]
-# system.cell_system.set_n_square(use_verlet_lists=False)
 
-# Non-bonded interactions
+# non-bonded interactions
 ###############################################################
 # WCA between monomers
 system.non_bonded_inter[0, 0].wca.set_params(epsilon=1, sigma=1)
 
-# WCA counterions - polymer
+# WCA counter-ions - polymer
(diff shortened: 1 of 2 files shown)


samples/electrophoresis.py
```diff
@@ -59,214 +64,183 @@
...
 
-
-# system.part[:-1].add_bond((harmonic, np.arange(n_monomers)[1:]))
-# system.part[1:-1].add_bond((harmonicangle, np.arange(n_monomers)[:-2],
-# np.arange(n_monomers)[2:]))
-
-# Particle creation with loops:
-for i in range(n_monomers):
+for i in range(N_MONOMERS):
     if i > 0:
```
jngrad: you could drop the conditional with `for i in range(1, N_MONOMERS):`

testsuite/scripts/samples/test_electrophoresis.py
```diff
@@ -20,29 +20,29 @@
...
+        self.assertTrue(27. < sample.opt[0] < 30.,
+                        "length = {:.1f}".format(sample.opt[0]))
+        self.assertTrue(
+            32. < sample.opt_obs[0] < 36., "length = {:.1f}".format(sample.opt_obs[0]))
 
     def test_mobility(self):
-        value = sample.mu
-        self.assertLess(abs(value), 0.45, "mobility = {:.2f}".format(value))
+        self.assertTrue(abs(sample.mu - 10.2) < 0.1,
+                        "mobility = {:.2f}".format(sample.mu))
```
jngrad: The error message
```
  File "testsuite/scripts/samples/test_electrophoresis.py", line 38, in test_mobility
    "mobility = {:.2f}".format(sample.mu))
AssertionError: False is not true : mobility = 10.26
```
is not very helpful, you may consider instead:
```python
self.assertAlmostEqual(sample.mu, 10.2, delta=0.1)
```
which outputs
```
  File "testsuite/scripts/samples/test_electrophoresis.py", line 37, in test_mobility
    self.assertAlmostEqual(sample.mu, 10.2, delta=0.1)
AssertionError: 10.341812051847997 != 10.2 within 0.1 delta
```

testsuite/scripts/samples/test_electrophoresis.py
```diff
@@ -20,29 +20,29 @@
...
 class Sample(ut.TestCase):
     system = sample.system
 
-    @ut.skipIf(tuple(map(int, np.__version__.split("."))) < (1, 10),
-               "not supported for numpy < v1.10")
     def test_persistence_length(self):
-        value = np.mean(sample.fit[0])
-        self.assertTrue(5. < value < 110., "length = {:.0f}".format(value))
+        self.assertAlmostEqual(sample.persistence_length, 28., delta=2,
+                               msg="persistence length of {:.1f} not correct".format(sample.persistence_length_obs))
```
jngrad:
```suggestion
                               msg="persistence length of {:.1f} not correct".format(sample.persistence_length))
```
I don't think the `msg` argument was designed with this kind of code duplication in mind, which is prone to subtle copy-paste errors that will actually hurt maintainers instead of guiding them. The default error message of unittest assertions is already clear enough, showing both the function name (`test_persistence_length`) and the faulty variable if the assertion is a one-liner. I would vote for removing custom messages that only format the variable names. I'm not even sure it can be automated via `inspect` in a reliable way (see https://stackoverflow.com/a/2749881, https://stackoverflow.com/a/2749857).

jngrad (changes requested):

testsuite/scripts/samples/test_electrophoresis.py
KaiSzuttor: I don't have a strong opinion on that  but I also think that we should not over-engineer one-liners.

jngrad (approved):
pending CI

jngrad:
I'm not following: did kodiak just merge `python` into a PR whose last commit did not pass CI?

KaiSzuttor:
it did not wait for the CI to pass because this PR is the only one in the queue

jngrad:
I thought kodiak would not merge `python` if the GitHub status said CI failed.

KaiSzuttor:
``` bash
# if there are running status checks on a PR when it's up for merge, don't
# wait for those to finish before updating the branch
optimistic_updates = false # default: true
```
https://kodiakhq.com/docs/detailed-setup

jngrad:
Ah right, thanks!

PR #3419, opened 2020-01-18
Fix broken python tests and refactor assertions

jngrad:


fweik (approved):

jngrad:
We can't merge it yet, the modified dpd test is timing out on arm64. I'll have a look. Probably just a matter of removing one of the tested `kT` values.

PR #3427, opened 2020-01-21
Fix Boost tests

### LINKED ISSUE #3426: Fix Boost tests
[`BOOST_CHECK_CLOSE`](https://www.boost.org/doc/libs/1_66_0/libs/test/doc/html/boost_test/utf_reference/testing_tool_ref/assertion_boost_level_close.html) takes a percentage as tolerance. Epsilon values must be multiplied by 100.

In this example, the first line is incorrect:
https://github.com/espressomd/espresso/blob/d553bd52da3a74dc00d177e3ebb9330ff5d4b273/src/utils/tests/int_pow_test.cpp#L43-L44

In this example, the line is correct:
https://github.com/espressomd/espresso/blob/d553bd52da3a74dc00d177e3ebb9330ff5d4b273/src/utils/tests/interpolation_test.cpp#L254

This affects both 4.1 and 4.2.

jngrad:
Fixes #3426

### DIFF
--- a/src/core/unit_tests/grid_test.cpp
+++ b/src/core/unit_tests/grid_test.cpp
@@ -170,9 +170,9 @@ BOOST_AUTO_TEST_CASE(regular_decomposition_test) {
     auto const result = regular_decomposition(box, {0, 0, 0}, node_grid);
     auto const local_box_l = result.length();
 
-    BOOST_CHECK_CLOSE(box_l[0], local_box_l[0] * node_grid[0], eps);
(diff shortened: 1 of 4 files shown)


fweik (approved):

jngrad:
@junghans I've tried cherry-picking this PR locally on top of 4.1.2 and it merged cleanly. You should be able to use it directly in 4.1.2 builds, without an if condition on `i586` (the tests were wrong for all archs).

junghans:
Seems to work! `collision_detection` still fails reliably.

PR #3434, opened 2020-01-23
P3M interpolation cleanup

fweik:
Description of changes:
 - Removed interpolation function duplication from P3M/DP3M
 - Reduced redundancy from cached interpolation weights
 - Fixed syntax of p3m benchmark


RudolfWeeber:
@fweik, this PR is marked WIP. Would you still like me to review this now?

fweik:
No, its not complete yet, sorry. I'll let you know.

src/core/electrostatics_magnetostatics/p3m_interpolation.hpp
```diff
@@ -0,0 +1,181 @@
+#ifndef ESPRESSO_P3M_INTERPOLATION_HPP
```
jngrad, line 21: missing copyright header

src/core/electrostatics_magnetostatics/p3m.cpp
```diff
@@ -95,22 +95,13 @@ p3m_data_struct p3m;
...
 static bool p3m_sanity_checks_system(const Utils::Vector3i &grid);
 
 /** Checks for correctness for charges in P3M of the cao_cut,
  *  necessary when the box length changes
  */
 static bool p3m_sanity_checks_boxl();
 
-/** Interpolate the P-th order charge assignment function from
- *  @cite hockney88a 5-189 (or 8-61). The following charge fractions
- *  are also tabulated in @cite deserno98a @cite deserno98b.
```
jngrad, line 110: I think these references are actually useful and should be moved to the docstring of
https://github.com/espressomd/espresso/blob/7f90057b87cb5b676a29fb42dbb7897157c04b95/src/utils/include/utils/math/bspline.hpp#L29-L30
This other function should be documented as well, looks like it's the first derivative:
https://github.com/espressomd/espresso/blob/7f90057b87cb5b676a29fb42dbb7897157c04b95/src/utils/include/utils/math/bspline.hpp#L189

src/python/espressomd/electrostatics.pxd
```diff
@@ -153,27 +135,26 @@ IF ELECTROSTATICS:
 
             return p3m_set_params(r_cut, mesh, cao, alpha, accuracy)
 
-        cdef inline python_p3m_set_tune_params(p_r_cut, p_mesh, p_cao, p_alpha, p_accuracy, p_n_interpol):
+        cdef inline python_p3m_set_tune_params(p_r_cut, p_mesh, p_cao, p_alpha, p_accuracy):
             # cdef inline python_p3m_set_tune_params():
```
jngrad: here and in
https://github.com/espressomd/espresso/blob/7f90057b87cb5b676a29fb42dbb7897157c04b95/src/python/espressomd/magnetostatics.pyx#L233
do these comments serve any purpose?

jngrad (changes requested):
skimmed through the diff, looks good overall, only have a few remarks

src/python/espressomd/electrostatics.pxd
fweik: No idea, probably not.

src/core/electrostatics_magnetostatics/p3m.cpp
fweik: If we want to keep them they should go on `p3m_interpolate`, as they are p3m-specific citations, while bsplines are not, and they are used in different contexts...

jngrad: If the coefficients in `bspline()` haven't been derived specifically for charge interpolation, then you're right, we don't need to cite the P3M paper (deserno98a) from which they come from. I'm asking this because I couldn't find the same tabulated values in other documents describing the B-splines recursion formula.

fweik: They are the centered ones (base interval `[-0.5, 0.5]` ), the formula from wikipedia is for base inveral `[0, 1] `. I think those base functions should be cleaned up anyway, but I think that is for another PR.

jngrad (approved):

PR #3436, opened 2020-01-23
maintainer: Removed guards for external features from generated confi…

fweik:
…g header.

Description of changes:
 - Drop define guards for `external` features from generated `config-features.hpp`.
   This is not needed anymore because the `myconfig.hpp` provided by the user is 
   validated before. This required a hard coded list of external features is now removed.


RudolfWeeber:
I'd prefer an MSVC-capable solution. The undefs can probably generated prgramatically as well. Will take a look next week.

PR #3441, opened 2020-01-28
utility: added inertia tensor calculation at the python level.

KaiSzuttor:


src/python/espressomd/inertia_tensor.py
```diff
@@ -0,0 +1,98 @@
+import numpy as np
+
+
+def inertia_tensor(positions, masses):
+    """
+    Calculate the moment of inertia tensor for given point masses
```
fweik: Isn't that thing called just `inertia tensor` ? `moment of inertia tensor` sounds odd, because the moment of inertia is a vector. C.f. `§32 The inertia tensor` from [Scripture](http://renaissance.ucsd.edu/courses/mae207/mech.pdf).

KaiSzuttor: I adopted the nomenclature from [wikipedia](https://en.wikipedia.org/wiki/Moment_of_inertia#Inertia_tensor) but maybe you are right

requirements.txt
```diff
@@ -27,3 +27,4 @@ pycodestyle==2.3.1
 pylint>=2.2.2
 astroid>=2.1.0
 isort>=4.3.4
+setuptools>=41.2.0
```
jngrad:
```suggestion
setuptools>=20.7.0
```
Ubuntu 19.10 only provides 41.1.0. The `SpecifierSet` feature has been available for 5 years. I just tried with the Ubuntu 16.04 version of setuptools (20.7.0) and it worked fine.

src/python/espressomd/inertia_tensor.py
fweik: I think `intertia tensor` is just short for `moment of intertia tensor`, I think what you did is fine, it was late...

testsuite/python/unittest_decorators.py
```diff
@@ -50,7 +51,19 @@ def skipIfMissingModules(*args):
...
+
+
+def skipIfUnmetModuleVersionRequirement(module, version_requirement):
+    """Unittest skipIf decorator for unmet module version requirement."""
+    try:
+        _module = importlib.import_module(module)
+    except ImportError:
+        return skipIfMissingModules(module)
+    if not setuptools.version.pkg_resources.packaging.specifiers.SpecifierSet(
+            version_requirement).contains(_module.__version__):
```
jngrad, line 66: This will fail on modules that don't provide a `__version__` variable, e.g. `ast`, but I can't see a workaround. Adding an extra  guard `assert hasattr(_module, '__version__')` wouldn't help, as it generates the same error message. We (core team) just need to be aware of that loophole when reviewing PRs that include this fixture.

requirements.txt
KaiSzuttor: okay

testsuite/python/unittest_decorators.py
KaiSzuttor: i think we can live with that

src/python/espressomd/inertia_tensor.py
```diff
@@ -0,0 +1,98 @@
...
+    -----
+    See wikipedia_.
+
+    .. _wikipedia: https://en.wikipedia.org/wiki/Moment_of_inertia#Inertia_tensor
+
+    """
+    inertia_tensor = np.zeros((3, 3))
+    for ind, p in enumerate(positions):
+        mass = masses[ind]
+        inertia_tensor[0, 0] += mass * (p[1]**2.0 + p[2]**2.0)
```
fweik: Would't it be less error prone and verbose to write this as tensor product?

```python
inertia_tensor += mass * (np.dot(p, p) * np.eye(3) - np.outer(p, p))
```

KaiSzuttor: :+1:

samples/rigid_body.py
```diff
@@ -19,42 +19,36 @@
...
+import enum
+import math
 
 import numpy as np
 
+import espressomd
+required_features = ["VIRTUAL_SITES_RELATIVE", "MASS", "ROTATIONAL_INERTIA"]
+espressomd.assert_features(required_features)
+import espressomd.virtual_sites
+import espressomd.inertia_tensor
```
fweik: I think this does not exist anymore.

fweik (approved):

PR #3443, opened 2020-01-30
Bond cutoff

### LINKED ISSUE #3442: Not all bond type considered for cutoff calculation
The calculation of the required cutoff for bonded interactions does not consider all bond
types that should be taken into account. The cutoff is calculate in `double recalc_maximal_cutoff_bonded()` in `bonded_interaction_data.cpp:26`, that calculation is
at least missing the `Harmonic_dumbbell_bond` and the `Quartic_bond`.

fweik:
Fixes #3442

Description of changes:
 - Define and consider cutoff for all bond types.


jngrad (approved):
LGTM

PR #3444, opened 2020-02-03
Refactor thermostats: RNG, callbacks and documentation

### LINKED ISSUE #3119: Switch NPT to Philox

jngrad:
Description of changes:
- switch NPT thermostat RNG from Mersenne twister to Philox (fixes #3119)
- add extra RNG correlation checks in unit tests using boost accumulators
- make thermostats object-oriented and generate their MPI callbacks with macros

### DIFF
--- a/src/core/thermostat.hpp
+++ b/src/core/thermostat.hpp
@@ -114,9 +114,14 @@ public:
     }
     pref_noise_rotation = sigma(temperature, time_step, gamma_rotation);
   }
-  /** Calculate the noise standard deviation. */
+  /** Calculate the noise prefactor.
+   *  Evaluates the quantity @f$ \sqrt{2 k_B T \gamma / dt} / \sigma_\eta @f$
+   *  with @f$ \sigma_\eta @f$ the standard deviation of the random uniform
+   *  process @f$ \eta(t) @f$.
+   */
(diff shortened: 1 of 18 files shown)


src/core/thermostat.hpp
```diff
@@ -94,8 +91,13 @@ extern bool thermo_virtual;
  * parameter structs
  ************************************************/
 
+struct BaseThermostat {
```
fweik, line 94: Can you please explain why the base class is needed? As far as I can tell the inheritance is not used, and in general (mutable) data members in base classes are a bad idea (due to the non-locality of mutation).

jngrad: I originally thought about moving most of the thermostat RNG MPI callbacks into the base struct as getters and setters, and only have a few MPI callbacks (not macro-generated) taking as argument the thermostat switch to select from which global object to call the setter/getter, but these plans didn't materialize. I wouldn't mind reverting that inheritance; it's probably too early to look into that yet.

If by non-locality of mutation you mean the ability of Brownian to change the value of the Langevin counter, I'm not sure how we could run into that situation, as having a `BaseThermostat` method that takes a ref to another `BaseThermostat` object is not useful.

fweik (approved):

PR #3446, opened 2020-02-04
maintainer: Handle generation of handling of external features in con…

fweik:
…fig generation automatically


fweik:
On a compiler on which the myconfig checking does not work (e.g. MSVC), definition of `external` features (e.g. such that are set by the build system) in the myconfig are ignored.  The old checks were useless (`#undef FEATURE` directly followed by `#if defined(FEATURE)`) and were removed. I think this is the best we can do while having the external features available in the myconfig.

RudolfWeeber (approved):
I guess something better could only be done using two separate compile units. Too much work.
Macros suck. Maybe we could transitoin to const bools which live in namespaces.
Then it would also not be possible to declare non-existing features.

Probably better work on struct of arrays and make more of that feature stuff go away.

PR #3448, opened 2020-02-04
Cleanup analyze module

jngrad:
Description of changes:
- simplify assertions
- update documentation


fweik (approved):

KaiSzuttor:
are you sure its worth refactoring the analyze module if most of it should go into the observables framework?

jngrad:
@KaiSzuttor I was supposed to move some of the features from the analysis module to observable 3 weeks ago, but the documentation, some assertions and multiple tests were wrong. This caused a massive delay and my work on observables now collides with @fweik's work. The tests were fixed in #3419 and the docs/assertions were fixed in this PR. Now @fweik and I can work independently on moving the analysis module to observables.

PR #3449, opened 2020-02-05
Update array_properties.py

biermanncarl:
In the ArrayPropertyTest class, the lbf actor setup was moved to the test_lb() method, which was the only test using the actor. The actor is now removed after the test. This fixes an issue in test_partial_periodic() when changing the system's periodicity. The error which occured can be seen in event.cpp, function on_cell_structure_change(). This, however, didn't cause the test to fail because the error was not handled by Python. It did cause the test to fail during development of the Stokesian Dynamics functionality, when a handle_errors() was added into globals.pyx.

This replaces my two previos change suggestions on the file. Sorry for the confusion.

### DIFF
--- a/testsuite/python/array_properties.py
+++ b/testsuite/python/array_properties.py
@@ -119,6 +119,12 @@ class ArrayPropertyTest(ut.TestCase):
         cpy = np.copy(v)
         self.assertTrue(cpy.flags.writeable)
 
+    def setUp(self):
+        self.system.box_l = [12.0, 12.0, 12.0]
+
+    def tearDown(self):
+        self.system.actors.clear()
+
     def test_common(self):
 
         # Check for exception for various operators
@@ -275,8 +281,6 @@ class ArrayPropertyTest(ut.TestCase):
         self.locked_operators(lbf[0, 0, 0].stress)
         self.locked_operators(lbf[0, 0, 0].stress_neq)
         self.locked_operators(lbf[0, 0, 0].population)
-
-        self.system.actors.clear()
 
     @utx.skipIfMissingFeatures(["LANGEVIN_PER_PARTICLE",
                                 "PARTICLE_ANISOTROPY"])
--- a/testsuite/python/array_properties.py
+++ b/testsuite/python/array_properties.py
@@ -268,14 +268,14 @@ class ArrayPropertyTest(ut.TestCase):
     def test_lb(self):
         lbf = lb.LBFluid(agrid=0.5, dens=1, visc=1, tau=0.01)
         self.system.actors.add(lbf)
-    
+
         # Check for exception for various operators
         # LB
         self.locked_operators(lbf[0, 0, 0].velocity)
         self.locked_operators(lbf[0, 0, 0].stress)
         self.locked_operators(lbf[0, 0, 0].stress_neq)
         self.locked_operators(lbf[0, 0, 0].population)
-        
+
         self.system.actors.clear()
 
     @utx.skipIfMissingFeatures(["LANGEVIN_PER_PARTICLE",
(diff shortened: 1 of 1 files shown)


testsuite/python/array_properties.py
```diff
@@ -268,12 +266,17 @@ def test_rot_aniso(self):
...
+        lbf = lb.LBFluid(agrid=0.5, dens=1, visc=1, tau=0.01)
+        self.system.actors.add(lbf)
+
         # Check for exception for various operators
         # LB
-        self.locked_operators(self.lbf[0, 0, 0].velocity)
-        self.locked_operators(self.lbf[0, 0, 0].stress)
-        self.locked_operators(self.lbf[0, 0, 0].stress_neq)
-        self.locked_operators(self.lbf[0, 0, 0].population)
+        self.locked_operators(lbf[0, 0, 0].velocity)
```
KaiSzuttor, line 280: why is it prohibited to set a velocity?

testsuite/python/array_properties.py
```diff
@@ -268,12 +266,17 @@ def test_rot_aniso(self):
...
+        self.locked_operators(lbf[0, 0, 0].stress)
+        self.locked_operators(lbf[0, 0, 0].stress_neq)
+        self.locked_operators(lbf[0, 0, 0].population)
+
+        self.system.actors.clear()
```
KaiSzuttor: in general this should go into the special function ```tearDown```

testsuite/python/array_properties.py
KaiSzuttor: and why does this test pass at all? There should only be an error raised if the shape is wrong

https://github.com/espressomd/espresso/blob/b89263db5db6e4084118fc4c54ba0f9068bcb9e5/src/python/espressomd/lb.pyx#L387-L400

testsuite/python/array_properties.py
biermanncarl: The actors.clear() is necessary there because otherwise it would interfere with test_partial_periodic(). The latter test raises a C++ error as long as an LB actor is active.

testsuite/python/array_properties.py
biermanncarl: Sorry I do not understand the question. What do you mean by "prohibited to set a velocity"? Where does such an error occur?

testsuite/python/array_properties.py
biermanncarl: I suppose the test_lb() test should go into its own testcase then, because it does not share the same setup with the other tests?

testsuite/python/array_properties.py
biermanncarl: I think I partly understand your question now. But I am not the author of these lines of code, I just adapted them so they would run with the new setup. I cannot answer any questions on why they are there.

testsuite/python/array_properties.py
biermanncarl: Discussed this with Jean-Noel. Agreed. Will do that.

testsuite/python/array_properties.py
KaiSzuttor: I think I misunderstood the test... Probably the test should have some doc-strings what is being tested but you are right that this is not your business.

testsuite/python/array_properties.py
KaiSzuttor: Maybe to clarify: the ```tearDown``` method is called automatically after each function that matches ```test*```. See [unittest documentation](https://docs.python.org/3.8/library/unittest.html#unittest.TestCase.tearDown).

biermanncarl: Yep, I first wrongly assumed it would only be called at the very end of all tests.

KaiSzuttor:
superseded by #3452

PR #3451, opened 2020-02-05
WIP: cmake: Switched from deprecated FindPythonInterp to FindPython3 and i…

fweik:
…mported targets for python libs

Description of changes:
 - Switch python detection from FindPythonIntterp to FindPython3 because the former is
   deprecated from cmake 3.12 and the latter provides imported targets.


fweik:
I guess we can't have that with bionic...

PR #3461, opened 2020-02-06
Refactor thermostats: RNG and code duplication

### LINKED ISSUE #3460: Rotation test not properly guarded
```
File "/home/weeber/es-builds/wa/testsuite/python/rotation.py", line 46, in test_raise_if_improper 
matrix = self.rotation.as_matrix() AttributeError: 'Rotation' object has no attribute 'as_matrix'
```
with SciPy version 1.3.3

jngrad:
Description of changes:
- make thermostats fully object-oriented
- convert DPD and thermalized bond to `BaseThermostat`
- centralize their interface in `thermostat.hpp`
- refactor the random number generators using templates
- guard scipy in the testsuite (fixes #3460)


src/core/random.hpp
```diff
@@ -101,38 +101,31 @@ Utils::Vector<uint64_t, 4> philox_4_uint64s(uint64_t counter, int key1,
...
- * sequence.
+ * @tparam salt RNG salt
+ * @tparam N    Size of the noise vector
  *
+ * @return Vector of uniform random numbers.
  */
-template <RNGSalt salt>
-Utils::Vector3d v_noise(uint64_t counter, int key1, int key2 = 0) {
+template <RNGSalt salt, size_t N = 3,
+          typename std::enable_if<(N > 1) and (N <= 4), int>::type = 0>
```
fweik: `std::enable_if_t`. Also I prefer `class = std::enable_if_t...` because I find' the version without the extra `int` less noisy, but I guess this is a matter of taste.

src/core/random.hpp
```diff
@@ -101,38 +101,31 @@ Utils::Vector<uint64_t, 4> philox_4_uint64s(uint64_t counter, int key1,
...
-  auto const noise = philox_4_uint64s<salt>(counter, key1, key2);
+template <RNGSalt salt, size_t N,
+          typename std::enable_if<N == 1, int>::type = 0>
+auto noise_uniform(uint64_t counter, int key1, int key2 = 0) {
 
-  using Utils::uniform;
-  return Utils::Vector3d{uniform(noise[0]), uniform(noise[1]),
-                         uniform(noise[2])} -
-         Utils::Vector3d::broadcast(0.5);
+  auto const integers = philox_4_uint64s<salt>(counter, key1, key2);
```
fweik, line 123: Wouldn't it be sufficient to use a `Philox2x64` here?

src/core/random.hpp
```diff
@@ -148,30 +141,45 @@ Utils::Vector3d v_noise(uint64_t counter, int key1, int key2 = 0) {
...
-  return {sqrt(-2.0 * log(u[0])) * cos(two_pi * u[1]),
-          sqrt(-2.0 * log(u[0])) * sin(two_pi * u[1]),
-          sqrt(-2.0 * log(u[2])) * cos(two_pi * u[3])};
+  // Box-Muller transform code adapted from
+  // https://en.wikipedia.org/wiki/Box%E2%80%93Muller_transform
+  // optimizations: the modulo is cached (logarithms are expensive), the
+  // sin/cos are evaluated simultaneously by gcc or separately by Clang
+  Utils::VectorXd<N> noise{};
+  constexpr double two_pi = 2.0 * Utils::pi();
+  auto modulo = sqrt(-2.0 * log(u[0]));
```
fweik: Please don't reuse variables like this. Make `modulo` and and `angle` const, and introduce new variables in the inner scope. You can get this by nesting the if's like

```c++
if(N > 2) {
  auto const module = ...
 ...
 if(N > 3) {
 ....
}
}
```
because `(N > 2)` is a special case of `(N > 3)`.

src/core/random.hpp
```diff
@@ -101,38 +101,31 @@ Utils::Vector<uint64_t, 4> philox_4_uint64s(uint64_t counter, int key1,
...
+  auto const integers = philox_4_uint64s<salt>(counter, key1, key2);
+  Utils::VectorXd<N> noise{};
+  std::transform(integers.begin(), integers.begin() + N, noise.begin(),
+                 [](size_t value) { return Utils::uniform(value) - 0.5; });
+  return noise;
+}
 
-  auto const noise = philox_4_uint64s<salt>(counter, key1, key2);
+template <RNGSalt salt, size_t N,
+          typename std::enable_if<N == 1, int>::type = 0>
```
fweik: `std::enable_if_t`

src/core/random.hpp
```diff
@@ -148,30 +141,45 @@ Utils::Vector3d v_noise(uint64_t counter, int key1, int key2 = 0) {
...
  */
-template <RNGSalt salt>
-inline Utils::Vector3d v_noise_g(uint64_t counter, int key1, int key2 = 0) {
-
-  auto const noise = philox_4_uint64s<salt>(counter, key1, key2);
-  using Utils::uniform;
-  Utils::Vector4d u{uniform(noise[0]), uniform(noise[1]), uniform(noise[2]),
-                    uniform(noise[3])};
+template <RNGSalt salt, size_t N = 3,
+          typename std::enable_if<(N >= 1) and (N <= 4), int>::type = 0>
```
fweik: `std::enable_if_t`

src/core/thermostat.hpp
```diff
@@ -92,6 +92,24 @@ extern bool thermo_virtual;
...
+      throw "The RNG is not initialized";
+    }
+    return rng_counter->value();
+  }
+  /** Is the RNG counter initialized */
+  bool rng_is_initialized() const { return rng_counter != nullptr; }
+
+private:
   /** RNG counter. */
   std::unique_ptr<Utils::Counter<uint64_t>> rng_counter;
```
fweik: There is no need anymore to store this on the heap right? A `boost::optional<uint64_t>` would too I think.

fweik (changes requested):
LGTM except for some minor changes.

RudolfWeeber:
```
> +private:
>    /** RNG counter. */
>    std::unique_ptr<Utils::Counter<uint64_t>> rng_counter;
``` 
> There is no need anymore to store this on the heap right? A `boost::optional<uint64_t>` would too I think.
IIRC, there were performance issues with optioals in the context of bonded interactions. Would those apply here or was that related to frequent creation of an optional?

fweik:
No, doesn't matter here, it's one static optional per thermostat. For an int an optional is also basically an pair<int, bool>...

src/core/random.hpp
jngrad: `Philox2x64` can only take 1 particle id for the key:
```
src/core/random.hpp:95:18: error: too many initializers for ‘const key_type {aka const r123array1x64}’
   const key_type k{id1, id2};
```
Although the function that generates 1 random uniform value is only called in an NPT function that has no second particle (hence the second key is always 0), we could in the future have a need to generate 1 random value from two particle ids. We would also introduce ambiguity in the function call, since the function generating N=2-4 random values has an optional argument.

src/core/random.hpp
jngrad: thanks for the suggestion

src/core/random.hpp
fweik: It takes 64 bits as key, e.g. 2 * 32 bit or 2 int.

fweik (approved):

PR #3463, opened 2020-02-07
Cone shape

KaiSzuttor:
Implementation of a cone with finite thickness.
This shape should replace the hollow cone (for which the
implementation is dubious).
shape-conical_frustum

conical_frustum


src/core/shapes/ConicalFrustum.cpp
```diff
@@ -0,0 +1,105 @@
...
+      }
+    } else {
+      // inside
+      if (m_direction == 1) {
+        dist *= -1.0;
+      }
+    }
+  } else {
+    vec = Utils::Vector3d{};
+    dist = -m_thickness / 2.0;
```
KaiSzuttor: this needs to respect ```m_direction```

src/core/shapes/ConicalFrustum.cpp
```diff
@@ -0,0 +1,108 @@
...
+ */
+#include <boost/algorithm/clamp.hpp>
+
+#include "ConicalFrustum.hpp"
+#include <utils/Vector.hpp>
+#include <utils/math/coordinate_transformation.hpp>
+
+/*
+ *
+   -------->r1
```
fweik: I think you should move this beautiful ascii art to the docstring of the class. Also you should add a high-level docstring to the class which describes the shape (e.g. like the python doc string).

pkreissl (commented):
I don't agree with calling your shape `ConicalFrustum`. I think a conical frustum has sharp edges, plus its surface includes the base and the cut. I can't think of a proper name though at the moment.

fweik:
@pkreissl do you have a proposal for a better name? I understand your point, but I can't come up with a better name. I also did not like the original _hollow cone_, because it is not just a cone. In this case I'd say it's close enough so that a user would not be surprise by what the shape actually is...

KaiSzuttor:
i think we won't find a compact name that correctly describes the shape. Therefore I looked for the sweet spot compact vs. exact.

pkreissl:
i am also having a hard time, finding an appropriate name. Thinking of "real world" funnels, maybe something with `Funnel` might be an option? I would not be too happy with just `Funnel`, though, as I think the common mathematical expectation would then be a first funnel surface (see, e.g., [MathWorld](url), [WolframAlpha](https://www.wolframalpha.com/input/?i=first+funnel+surface)), for which the _z_-component is based on the natural logarithm.
Maybe `StraightFunnel` would be an option?

KaiSzuttor:
> Maybe StraightFunnel would be an option?

this would be a cylinder for me...

pkreissl:
`ConicalFunnel`? I don't know, maybe something along the lines.

KaiSzuttor:
What about ```HollowConicalFrustum```? That would not clash with the name of any preexisting geometric shape and the details (like the rounded corners) to such shapes have to be looked up by the user via the documentation anyhow...

pkreissl:
For me that would be okay. I was just a bit worried about the _non-hollow_ aspect of your shape name, which really surprised me (and probably would also have been unexpected for potential users).

KaiSzuttor:
yeah, we don't want the shape to be a [ChronicalFrustrum](https://en.wiktionary.org/wiki/frustrum)

KaiSzuttor:
p.s.: name is fixed now. Not gonna change again

jngrad:
I've rolled back to an earlier version of `cuda:10.1` until I can fix #3476.

fweik (approved):

PR #3467, opened 2020-02-10
Update shapes code

jngrad:
Description of changes:
- remove an incorrect ifdef guard
- modernize old code


jngrad:
This is an easy PR, code review can be done by @jrfinkbeiner or @biermanncarl.

For parts of this PR, it might be helpful to use a less granular `git diff`:
```sh
git fetch origin pull/3467/head:pr-3467
git checkout pr-3467
git diff --color-words HEAD^ HEAD
```

src/core/shapes/Stomatocyte.cpp
```diff
@@ -33,41 +33,28 @@ void Stomatocyte::calculate_dist(const Utils::Vector3d &pos, double &dist,
...
-  closest_pos = m_position + mu * m_orientation;
+  auto const closest_pos = m_position + mu * m_orientation;
 
   // So the shortest distance to the line is
 
-  Utils::Vector2d dist_2D;
-  dist_2D[0] = (closest_pos - pos).norm();
-  dist_2D[1] = mu * m_orientation.norm();
+  auto const dist_2D =
+      Utils::Vector2d{{(closest_pos - pos).norm(), mu * m_orientation.norm()}};
```
fweik: Is there a specific reason why you are using two pairs of braces for initialization? `Vector2d{0., 1.}` is totally equivalent, with the difference that `Array<double, 2>{0., 1.}` would work two (using aggregate init). See also [the discussion in the core guidlines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es23-prefer-the--initializer-syntax) on initializer synatax.

jngrad: More of a habit rather than a specific reason. This avoids ambiguous code in situations where pointers are involved, because a ctor call with two pointers can be interpreted as begin/end iterators:
```c++
#include<vector>
int main() {
    int *a = nullptr, *b = nullptr;
    std::vector<int> vec{a, b};  // segfault, no compiler warning
    std::vector<int> vec{{a, b}};  // compiler warning: type mismatch
}
```
This is also more coherent with what you would write for e.g.
```c++
struct A {int a[2]; int b;};
struct A object{{1,2}, 3};
```
i.e. the outer curly braces call the ctor, the inner curly braces delimit the first argument to the ctor.

I took this habit without really looking into style guides. It also solves an issue that is rather unlikely to occur. Should I refrain from using this syntax from now on?

fweik: Well you made me ask myself and then you why there are additional braces. Also the single braces are easier teachable, because they work on all types on classes. I don't think the difference between a constructor, aggregate initialization and an initializer list is sth I'd teach to beginners. Instead I'd teach the uniform syntax that works almost always and does the right thing except in rare occasions. (But there is much debate on this area, see e.g. https://www.youtube.com/watch?v=7DTlWPgX6zs, also I think the rules changed again slightly with C++17...).

jngrad: I actually took this habit after watching two different CppCon short talks where the iterator issue forced speakers to use double braces in their codebase. I haven't watched the talk you just posted yet. Should I revert the double braces initialization?

src/core/shapes/HollowCone.cpp
```diff
@@ -29,50 +29,49 @@ using namespace std;
...
-      yp, zp, xpp, ypp, sin_xy, cos_xy, normal_x_3D, normal_y_3D, normal_z_3D,
-      normal_3D_x, normal_3D_y, normal_3D_z;
 
   // Set the dimensions of the hollow cone
 
-  r0 = m_inner_radius;
-  r1 = m_outer_radius;
-  w = m_width;
-  alpha = m_opening_angle;
+  auto const r0 = m_inner_radius;
```
KaiSzuttor: if they can be const there is no need to copy them at all...

src/core/shapes/HollowCone.cpp
```diff
@@ -29,50 +29,49 @@ using namespace std;
...
+  auto const r1 = m_outer_radius;
+  auto const w = m_width;
+  auto const alpha = m_opening_angle;
 
   // Set the point for which we want to know the distance
 
-  Utils::Vector3d point_3D = pos;
+  Utils::Vector3d const point_3D = pos;
```
KaiSzuttor: same here

src/core/shapes/HollowCone.cpp
```diff
@@ -101,40 +100,39 @@ void HollowCone::calculate_dist(const Utils::Vector3d &pos, double &dist,
...
-          time4 * w * w + time4 * time4 * w * w -
-          (x_2D - r0) * (-1.0 + 2.0 * time4) * w * cos(alpha) +
-          y_2D * (-1.0 + 2.0 * time4) * w * sin(alpha) + r1 * r1 -
-          2.0 * y_2D * r1 * cos(alpha) +
-          (-2.0 * x_2D * r1 + 2.0 * r0 * r1) * sin(alpha);
-  mdst3 = sqrt(mdst3);
-
-  double distlist[4] = {mdst0, mdst1, mdst2, mdst3};
+
+  auto const mdst1 =
```
KaiSzuttor: please don't invest any time in this code... there is an open PR which might make this code go

KaiSzuttor (commented):
this PR should wait until PR #3463 is reviewed

src/core/shapes/Stomatocyte.cpp
jngrad: actually I can drop the double braces, they don't always help with pointers:
```c++
#include<vector>
#include<string>
int main () {
    std::vector<std::string> vec{{"1", "2"}};     // segfault: iterators
    std::vector<std::string> vec{"1", "2"};       // ok
    std::vector<std::string> vec{{"1"}, {"2"}};   // ok, "correct" way?
    std::vector<std::string> vec{{{"1"}, {"2"}}}; // ok
    return vec.size();
}
```

pstaerk (approved):

src/core/shapes/HollowCone.cpp
jngrad: Once you've merged your PR, I'll merge it here to delete the `HollowCone`.

jngrad: done, `HollowCone` is gone

KaiSzuttor (approved):

PR #3482, opened 2020-02-13
Remove ESPResSo system seed mechanism

jngrad:
All thermostats are now Philox-based. The Mersenne Twister system seed infrastructure (global variables, MPI callbacks, Python bindings) is no longer needed.


KaiSzuttor (approved):

jngrad:
`espressomd.polymer.linear_polymer_positions` has become flaky. I can reproduce the error with `ctest --repeat-until-fail 1000  --output-on-failure -R polymer_linear` (edit: with `seed = random.randint(0,1000)` to change the seed, otherwise the test always passes). If I print the Mersenne Twister seed from a failed test and re-use it in the test script, the test fails every time. The only change to `polymer.cpp` is that its Mersenne Twister now discards the first 1 million values to remove correlation.

src/core/polymer.cpp
```diff
@@ -126,9 +127,9 @@ draw_polymer_positions(PartCfg &partCfg, int const n_polymers,
   std::vector<std::vector<Utils::Vector3d>> positions(
       n_polymers, std::vector<Utils::Vector3d>(beads_per_chain));
 
-  auto rng = [mt = std::mt19937{static_cast<unsigned>(seed)},
-              dist = std::uniform_real_distribution<double>(
-                  0.0, 1.0)]() mutable { return dist(mt); };
+  auto mt = Random::mt19937(static_cast<unsigned>(seed));
```
fweik: Is there a reason why you widened the scope of `mt`?

fweik:
The polymer generation can fail, so it should not have a random seed. Why would you expect this to work?

src/core/polymer.cpp
jngrad: Nice catch, I originally moved `mt` out to discard the first 1 million values, but then I made a wrapper in `random.hpp` to avoid the situation where MT would used without warmup. I'll put it back in the capturing group.

jngrad:
Why would the fixed seed in the test script generate correct positions in all machines except `osx`? If the MT rng produces random values that are architecture-dependent, fixing the seed is useless.

fweik:
>  If the MT rng produces random values that are architecture-dependent

is that so?

fweik:
> The 10000th consecutive invocation of a default-constructed std::mt19937 is required to produce the value 4123659995.

> The 10000th consecutive invocation of a default-constructed std::mt19937_64 is required to produce the value 9981545732273789042

jngrad:
I wasn't sure, that's why I used "if". I'm looking through `polymer.cpp` to see if something is not deterministic. The OpenGL visualizer is also currently broken on my machine, so I can't visually check the algorithm output.

PR #3484, opened 2020-02-14
Fix polymer generation

### LINKED ISSUE #3483: Polymer generation does not obey minimum image convention

fweik:
Fixes #3483.

Description of changes:
 - Obey minimum image convention
 - To not falsely pre-populate all positions with the zero vector
 - Refactored generation loop to make it more redable
 - Check for constraint violations between starting positions

### DIFF
--- a/src/core/polymer.cpp
+++ b/src/core/polymer.cpp
@@ -148,15 +148,16 @@ draw_polymer_positions(PartCfg &partCfg, int const n_polymers,
     if (m == 0) {
       return (p < start_positions.size()) ? start_positions[p]
                                           : random_position(rng);
-    } else if (not use_bond_angle or m < 2) {
-      return positions[p][m - 1] + bond_length * random_unit_vector(rng);
-    } else {
-      auto const last_vec = positions[p][m - 1] - positions[p][m - 2];
-      return positions[p][m - 1] +
-             Utils::vec_rotate(
-                 vector_product(last_vec, random_unit_vector(rng)), bond_angle,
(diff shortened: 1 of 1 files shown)


src/core/polymer.cpp
```diff
@@ -123,91 +119,90 @@ draw_polymer_positions(PartCfg &partCfg, int const n_polymers,
...
+    if (is_valid_pos(start_positions[p])) {
+      positions[p].push_back(start_positions[p]);
     } else {
-      positions[p][0] = start_positions[p];
+      throw std::runtime_error("Invalid start positions.");
     }
   }
 
-  // create remaining monomers' positions
+  /* Draw a monomer position, obeying angel and starting position
```
jngrad:
```suggestion
  /* Draw a monomer position, obeying angle and starting position
```

src/core/polymer.cpp
```diff
@@ -123,91 +119,90 @@ draw_polymer_positions(PartCfg &partCfg, int const n_polymers,
...
+   * constraints where apropriate. */
```
jngrad:
```suggestion
   * constraints where appropriate. */
```

jngrad (changes requested):

jngrad:
This is a lot better! The `polymer_linear.py` test doesn't fail anymore, even with a different seed at each execution. This isn't the first time that a bug is ignored by setting a magic seed in the test script...

jngrad (approved):

PR #3485, opened 2020-02-14
coldet: Allow binding to unrelated virtual sites.

fweik:
Currently virtual sites are completely excluded from collisions, which is changed by this PR.
It is sufficient to only exclude virtual sites of the type that are placed by collisions. With this it
is e.g. possible to collide with the surface of a raspberry. (And ghosts need not know anymore
if they are virtual, which is the actual reason for this.)

### DIFF
--- a/src/core/collision.hpp
+++ b/src/core/collision.hpp
@@ -131,9 +131,15 @@ inline void detect_collision(Particle const &p1, Particle const &p2,
     if (!glue_to_surface_criterion(p1, p2))
       return;
 
-  // Ignore virtual particles
-  if ((p1.p.is_virtual) || (p2.p.is_virtual))
-    return;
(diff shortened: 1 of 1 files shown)


RudolfWeeber:
I'm not convinced, this is safe. For the virtual sites based modes like bind at point of collision, do not work when binding to another virtual site (such as the beads of a raspberry). This is, because the update order of virtual sites is not guaranteed. Only the bind centers method would be safe to use with virtual sites as collision partners.

So, if no vs based collision scheme is active, the restriction can be removed, but for the other cases, it is not clear to me how to make them safe without having the virtual info on the ghosts.

fweik:
I see. The "virtual site based binding" seems to me like a hack. The fact that integration details of ghosts have to be exposed seem just plain wrong to me.

PR #3491, opened 2020-02-17
Remove infinite loop in linear polymer

jngrad:
Description of changes:
- fix infinite loop in `draw_polymer_positions()` (regression from #3484)
- remove fixed seed in the `polymer_linear.py` test that used to hide a bug in the previous implementation of `draw_polymer_positions()`

### DIFF
--- a/src/core/polymer.cpp
+++ b/src/core/polymer.cpp
@@ -177,6 +177,7 @@ draw_polymer_positions(PartCfg &partCfg, int const n_polymers,
   // create remaining monomers' positions by backtracking.
   for (int p = 0; p < n_polymers; ++p) {
     for (int attempts_poly = 0; attempts_poly < max_tries; attempts_poly++) {
+      int rejections = 0;
       while (positions[p].size() < beads_per_chain) {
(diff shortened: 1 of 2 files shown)


src/core/polymer.cpp
```diff
@@ -186,6 +187,11 @@ draw_polymer_positions(PartCfg &partCfg, int const n_polymers,
         } else if (not positions[p].empty()) {
           /* Go back one position and try again */
           positions[p].pop_back();
+          rejections++;
```
fweik, line 190: Can you please explain why this is needed? Couldn't you just increase `attempts_poly`?

fweik: Ah sorry, when the first one has fixed position. But there has to be a more elegant solution than this...

jngrad: this while loop doesn't increment `attempts_poly`, therefore in high density systems, you can get stuck in an infinite cycle of deleting the previous bead and adding a new one at the same position

fweik (approved):

PR #2701, opened 2019-04-10
Interface Walberla LB

RudolfWeeber:
Current status:
* All LB features except pressure tensor are in
* Pressure tensor can be merged as soon as corresponding PR in PbmPy and Walberla are merged
* Electrokinetics is WIP and tracked in a separate ticket
* Documentation is not yet adapted
* GPu support and support for switching accuracy to single precision on the CPu is not yet done.

The current plan is to merge this after the release of 4.2

PR #3493, opened 2020-02-17
Shape union

KaiSzuttor:
A union of shapes.

### DIFF
--- a/src/python/espressomd/shapes.py
+++ b/src/python/espressomd/shapes.py
@@ -14,7 +14,10 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
-from .script_interface import ScriptInterfaceHelper, script_interface_register
+
+import collections.abc
+
+from .script_interface import ScriptInterfaceHelper, script_interface_register, ScriptObjectRegistry
 from .utils import requires_experimental_features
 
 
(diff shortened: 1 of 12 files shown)


src/core/shapes/ShapeUnion.hpp
```diff
@@ -0,0 +1,72 @@
...
+  }
+  void calculate_dist(Utils::Vector3d const &pos, double &dist,
+                      Utils::Vector3d &vec) const {
+    auto dist_compare = [&pos](std::pair<double, Utils::Vector3d> const &res,
+                               std::shared_ptr<Shapes::Shape> const &s) {
+      double d;
+      Utils::Vector3d vec;
+      (*s).calculate_dist(pos, d, vec);
+      if (d < 0.0)
+        throw std::runtime_error(
```
fweik: This makes the union unusable as an LB boundary.

jngrad (commented):
Looks good, but it would be nice to have a minimal documentation for this feature. For example, the `calculate_dist()` method could use a docstring like `/** Calculate the distance to the closest shape */`. In the Sphinx docs, file `constraints.rst`, one could make section 11.2 a subsection of section 11.1 (and make 11.2.1 to 11.2.4 subsubsections of 11.2, but keep 11.2.5-11.2.6 as subsections of 11.1), and have a new section 11.3 to describe the union.

fweik:
@jngrad this should have a formal definition of what is calculated (which is not hard), but first this has to be actually feature complete ^^

jngrad:
@fweik of course. I saw your comment when writing my review. I was a little confused at first as to what was calculated, and thought the union was a variant, until I read through the boost accumulator code ^^

src/python/espressomd/shapes.py
```diff
@@ -266,3 +269,73 @@ class HollowConicalFrustum(Shape, ScriptInterfaceHelper):
     .. image:: figures/conical_frustum.png
     """
     _so_name = "Shapes::HollowConicalFrustum"
+
+
+@script_interface_register
+class ShapeUnion(Shape, ScriptObjectRegistry):
```
fweik: I think this should be called `Union`, it's clear from the context that this is about `Shapes`.

fweik (approved):
LGTM. The change on the bondary logic need to be applied to #2701 (Walberla).

KaiSzuttor:
@fweik I've simplified the LB boundary initialization a bit via boost.

PR #3504, opened 2020-02-18
WIP Collision detection rework

### LINKED ISSUE #3296: Requirements for reworked collision detection
This is my understanding of a dicsussion during the Espresso offline meeting

The feature creates bonds between colliding particles. 
It can additionally creates "non-sliding contacts" by creating virtual sites related to the colliding particles and bonds between those virtual sites.

There is an other "mode", in which, when two particles collide, a third particle in the vicinity is searched. If one is present, angle bonds are created to preserve the relative location of the three particles. 
This mode currently has only one customer, namely the soot particle agglomeration work by @hirschsn.

I'll focus on the 'local' scheme, which deals only with the colliding particles.
The criterion for triggering a collision should include
* a pair of particle types 
* a cut-off distance
* a probability (via a random number)
* question: Are more advanced criteria such as non-bonded pair energy needed? If so, how would one deal with potentials that have a long range contribution?
* Question: are more advanced criteria like distance-dependent probabilities needed?

Actions taken on collision:
* creation of a pair bond between the colliding particles
* type conversion of the colliding particles (to 'inactivate' reaction sites)
* creation of virtual sites relative to the colliding particles and a bond between them. The position of the virtual sites on the line connecting the colliding particles can be chosen. The type of the virtual particles can be chosen.
* safely interupt the integration loop to allow users to do extra model-specific stuff

Design:
* a two-stage process needs to be used. During the pair traversal, potential collisions are queued. Later, the queue is processed and actions are taken
* the code for performing the collision actions needs to be outside the collision detection code. 
* the current collision handling code is too complicated
* The challange is managing the particle ids of newly created particles. The collision detection needs to know these, to place bonds between them
  The most complicated collision action might be this:
    * paritcles p1 and p2 have collided. There types are t1 and t2
    * create a bond of type t_c between p1 and p2
    * create a virtual sites v1 relative to p1. It's position depends on the positoins of p1 and p2 and an offset coming from the collision detection parameters
    * create a virtual site v2 relative to p2 analogously
    * add a bond of type t_v between the virtual sites
    * change type of particle p1 to t1_n
    * change type of particle p2 to t2_n
* I'm under the impression that only two collision modes will remain
  * the two particle one, in which each individual collision action can be activated as needed
  * the three-particle angle bond one

Potential users, please comment on whether that covers your models' requirements.

itischler:
will fix #3296

Description of changes:
 -  reworking collision detection to not use different modes anymore, which will make it more flexible to use.


src/core/collision.hpp
```diff
@@ -42,54 +42,46 @@
...
-  // Square of distance at which particle are bound
-  double distance2;
-
-  /// bond type used between centers of colliding particles
-  int bond_centers;
-  /// bond type used between virtual sites
-  int bond_vs;
-  /// particle type for virtual sites created on collision
-  int vs_particle_type;
+  int active;
```
jngrad, line 55: Shouldn't this `int active` flag be a `bool`? Also, its docstring now says `/// collision handling mode, a combination of constants COLLISION_MODE_*`, which doesn't seem to be the case. This flag is also read by two conditionals, but its value is never changed outside of the zero-initialization in the class ctor.

src/core/collision.hpp
```diff
@@ -109,14 +101,31 @@ bool validate_collision_parameters();
...
+    return (count_multiples == 2 ||
+            std::count(collision_params.particle_type.begin(),
+                       collision_params.particle_type.end(), p2.p.type) > 0);
+  return 0;
+}
+
+inline bool collision_detection_criterion(Particle const &p1,
+                                          Particle const &p2) {
+  if (collision_params.rate > 0)
+    return d_random() < collision_params.rate * time_step;
```
jngrad, line 122: Function `d_random()` was removed in 0abe004dc28. You can either use a Philox counter (ex: [`BaseThermostat`](https://github.com/espressomd/espresso/blob/af75344bed302efeab0910234fad4cb4e956ae77/src/core/thermostat.hpp#L95-L121)), or a Mersenne Twister (ex: [`ParticleCache_test`](https://github.com/espressomd/espresso/blob/af75344bed302efeab0910234fad4cb4e956ae77/src/core/unit_tests/ParticleCache_test.cpp#L67-L76)), and pass their random integers to a function that generates a real number between 0 and 1 (ex: [`Utils::uniform()`](https://github.com/espressomd/espresso/blob/af75344bed302efeab0910234fad4cb4e956ae77/src/utils/include/utils/uniform.hpp#L36-L41)). In both cases, you'll need to store the RNG state somewhere.

fweik:
I think the whole construction would be much easier if there where a set of parameter struct for every type pair, much like for the non-bonded interactions. Also the cutoff needs to be properly calculated, and respected in the global cutoff calculation instead of throwing an error if the cutoff requirement is not fulfilled. The current approach is needlessly complicated, and the implementation with the vectors has considerable performance bugs.

KaiSzuttor:
is anything planed or ongoing here?

fweik:
In my opinion this functionality can not added in the core at the moment without increasing the technical debt. This is what triggered all the core refactoring that I eventually failed to deliver in time. I still think that the (MD) core code needs to be brought into better
shape before this can be implemented, and I think this requires some strategic planning on what actually the requirements are and how methods can be hosted by the core without introducing massive interdependencies. It is just one of the things where
change can not be introduced without making things worse because of the bad state the code is in.

RudolfWeeber:
Planning for collision detection and bond breakage and the core refactoring which is a pre-requisite should be discussed at an Espresso meeting.

KaiSzuttor:
> Planning for collision detection and bond breakage and the core refactoring which is a pre-requisite should be discussed at an Espresso meeting.

that might be true... I just wanted to have an update because we have a few PRs that seem to be stuck or not being worked on anymore

jngrad:
Closing in favor of #4483.

PR #3505, opened 2020-02-18
Remove max_seen_particle global from python interface

fweik:
Description of changes:
 - Added particle_data interface functions to get particle ids
   and maximal id
 - Removed global usage

The rest is pending on the collision detection refactoring, and other stuff...


jngrad (approved):
LGTM, pending CI

PR #3509, opened 2020-02-19
Fix electrophoresis sample test

### LINKED ISSUE #3415: Increase error bounds for electrophoresis sample
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/pipelines/10735

jngrad:
Fixes #3415


KaiSzuttor:
just checked where the difference in the observable value vs direct calculation comes from... it's just the sampling stride which is 1 vs 100 time-steps.

KaiSzuttor (approved):

PR #3511, opened 2020-02-19
Fix bond drawing bugs in the OpenGL visualizer

### LINKED ISSUE #3497: Undefined variable exception in the OpenGL visualizer with periodic images
A variable `dim` is accessed when drawing bonds in periodic images of the unit cell:
https://github.com/espressomd/espresso/blob/f9ae0e88f4b62b63696330553ab1aaff9e7b2cfb/src/python/espressomd/visualization_opengl.pyx#L1225-L1226
https://github.com/espressomd/espresso/blob/f9ae0e88f4b62b63696330553ab1aaff9e7b2cfb/src/python/espressomd/visualization_opengl.pyx#L1261-L1264

However the variable is never declared, so when this part of the code is executed, a `NameError` is raised, preventing the image from being drawn, but without a segfault. This happens only when the system has bonds and the visualizer is instantiated with periodic image display (e.g. with argument `periodic_images=[1,1,1]`). The regression was introduced by 0a22eeb94ee in 2016.

The Cython compiler did not complain because `undeclared-variable` compiler warnings are silenced when a wildcard import is used (discovered by #3496):
https://github.com/espressomd/espresso/blob/f9ae0e88f4b62b63696330553ab1aaff9e7b2cfb/src/python/espressomd/visualization_opengl.pyx#L20

The bond length comparison is also wrong, as it assumes a cubic box. The code could also be made more readable.

jngrad:
Fixes #3497

Description of changes:
- restore the periodic image functionality for systems with bonds (fixes the `NameError` exception)
- correctly calculate the orientation of bonds cut by the faces of the simulation box (fix connectivity issues at the interface between the unit cell and periodic images)

### DIFF
--- a/src/python/espressomd/visualization_opengl.pyx
+++ b/src/python/espressomd/visualization_opengl.pyx
@@ -413,6 +413,8 @@ class openGLLive:
 
         # HAS PERIODIC IMAGES
         self.has_images = any(i != 0 for i in self.specs['periodic_images'])
+        self.image_vectors = [
+            list(range(-i, i + 1)) for i in self.specs['periodic_images']]
 
         # INITS
         self.system = system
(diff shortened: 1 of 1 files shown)


jngrad:
@christophlohrmann please check the bond connectivity between the unit cell and periodic images by adding `periodic_images=[0,1,0]` to instantiations of the OpenGL visualizer in the samples.

fweik (approved):
LGTM, can be merged if @christophlohrmann approves.

christophlohrmann (approved):
Bonds in polymer-type simulations look correct now. Also periodic images are displayed nicely. => Visual inspection test passed.

PR #3512, opened 2020-02-19
Layered

fweik:
Description of changes:
 - Removed layered topology

This only needed to exist for MMM2D, which was since removed. The same
functionality can be achieved by the regular domain composition using the
`fully_connected` flags.


src/python/espressomd/globals.pxd
```diff
@@ -65,6 +65,13 @@ cdef extern from "nonbonded_interactions/nonbonded_interaction_data.hpp":
     double recalc_maximal_cutoff_bonded()
     double recalc_maximal_cutoff_nonbonded()
 
+cdef extern from "cells.hpp":
+    ctypedef struct CellStructure:
+        int type
+        bool use_verlet_list
+
+    CellStructure cell_structure
```
jngrad: I don't understand this diff. This include from "cells.hpp" shouldn't be here, because the parent merge commit 341e43c42a894b410c from your merge commit 337f46f02c is removing it. Could it be that you accidentally used your local branch here when resolving a merge conflict?

fweik: There was a conflict that I resolved by hand, I may very well have gotten it wrong.

fweik: Yeah, I picked the wrong side, give me a second.

jngrad: ok I was just confused because GitHub won't show diff lines in a merge commit (even in the long format) when a conflict was resolved by using only one side of the history (because there is no change compared to that history).

jngrad (approved):
LGTM, pending CI

PR #3533, opened 2020-02-21
Fix memory leak in OpenGL visualizer

### LINKED ISSUE #3530: Online visualization memory leak
There is a memory leak when running a simulation with the OpenGL visualizer. The leak rate is around 12 MiB/min for a polymer (`visualization_bonded.py`) and 0.3 MiB/min for a Lennard-Jones liquid (`visualization_ljliquid.py`). It does not change when doubling or halving the number of particles in the system. The leak persists even if the simulation freezes, e.g. when the for loop in the integration thread stops. Not sure if the bug comes from the espresso code or from the OpenGL driver. Issue reported by the mailing list. Confirmed on the current python branch (cd66da057d) and 4.1.2 release.

Here is a shell script to measure memory usage of a simulation as a function of time. The monitoring loop stops when the visualizer window is closed by the user.
```bash
./pypresso ../samples/visualization_bonded.py --opengl &
pid=${!}
while [ $? == 0 ]; do
  sleep 10
  ps -p ${pid} -o etimes,pmem,vsz --no-headers >> ram.tsv
done
```
Data processing with matplotlib:
```python
import numpy as np
from scipy.signal import savgol_filter
from psutil import virtual_memory

# three columns: time in seconds, physical memory usage in percentage with 0.1%
# resolution (requires a linear Savitzky-Golay filter), virtual memory usage in kiB
data = np.loadtxt('ram.tsv')
time = data[:,0] / 60.  # runtime in minutes
pmem = data[:,1] * virtual_memory().total / 100. /1e6  # physical memory in MiB
vmem = data[:,2] / 1e3  # virtual memory in MiB

print('memory leak rate: {:.1f} MiB / min'.format(
    (pmem[-1] - pmem[0]) / (time[-1] - time[0])))

plt.plot(time, savgol_filter(pmem - pmem[0], 61, 1), label='Physical memory')
plt.plot(time, vmem - vmem[0], label='Virtual memory')
plt.xlabel('Time (min)')
plt.ylabel('Memory usage (MiB)')
plt.title('Memory leak after OpenGL initialization')
plt.legend()
plt.show()
```
The polymer has a linear trend, the LJ liquid has steps:
ram-bonds-visu
ram-lj-visu

jngrad:
Fixes #3530

Description of changes:
- free memory of quadric objects; they caused a 12 MiB/min memory leak in simulations containing bonds and a 3 MiB/min memory leak in simulations containing cylindrical shapes (`Cylinder`, `SpheroCylinder`, `SimplePore`, `Slitpore`)
* fix an issue in the `visualization_constraints.py` sample that prevented a `Slitpore` from being added to the list of constraints

### DIFF
--- a/samples/visualization_constraints.py
+++ b/samples/visualization_constraints.py
@@ -30,7 +30,7 @@ group = parser.add_mutually_exclusive_group()
 group.add_argument("--wall", action="store_const", dest="shape", const="Wall",
                    default="Wall")
 for shape in ("Sphere", "Ellipsoid", "Cylinder", "SpheroCylinder",
-              "Stomatocyte", "SimplePore", "SlitPore", "HollowConicalFrustum"):
+              "Stomatocyte", "SimplePore", "Slitpore", "HollowConicalFrustum"):
     group.add_argument("--" + shape.lower(), action="store_const",
                        dest="shape", const=shape)
 args = parser.parse_args()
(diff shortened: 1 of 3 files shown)


jngrad:
@christophlohrmann please check a few samples, e.g. the following are directly affected by this PR:
```sh
./pypresso ../samples/visualization_bonded.py --opengl
./pypresso ../samples/visualization_constraints.py --simplepore
./pypresso ../samples/visualization_constraints.py --slitpore
./pypresso ../samples/visualization_constraints.py --cylinder
./pypresso ../samples/visualization_constraints.py --spherocylinder
```

christophlohrmann (commented):
The shapes look like they are supposed to. Also memory doesn't blow up after longer runtime.
Still, in the samples, particles are on both sides of the constraints and also can hop across the constraints

jngrad:
I've just tried to rewrite the particle placement loop to make some shapes impossible to cross, but it didn't work for the slitpore, where particles spawn too close to each other due to the confined geometry.

Particles crossing constraints is most likely a side-effect of the force capping mechanism, which prevents the force from becoming infinite when crossing a wall or surface. Writing custom particle placement code for each constraint and validating the simulation against crashes for each constraint in CI seems too much work for such a simple sample, and will make the sample a lot more difficult to re-use by novice users.

KaiSzuttor:
you should use energy minimization instead of force capping

jngrad:
We use steepest descent to minimize the energy, and then the Langevin thermostat with a high gamma. The force capping is there to prevent the simulation from crashing.

fweik (approved):

PR #3538, opened 2020-02-24
Opengl rasterization for union

KaiSzuttor:


fweik (approved):

src/core/shapes/Union.hpp
```diff
@@ -66,6 +67,14 @@ class Union : public Shape {
                         dist_compare);
   }
 
+  bool is_inside(Utils::Vector3d const &pos) const override {
```
KaiSzuttor, line 71: @fweik this override is necessary if the union shape is used in a LB boundary.

src/core/shapes/Union.hpp
```diff
@@ -66,6 +67,14 @@ class Union : public Shape {
...
+    return boost::accumulate(
```
fweik: I think you should use `any_of`, which will short-cut and makes the intent clearer.

fweik (changes requested):

src/core/shapes/Union.hpp
KaiSzuttor: you probably mean ```all_of```

KaiSzuttor:
i will provide a test for LB boundary using the union shape which will make codecov happy

fweik (approved):

PR #3543, opened 2020-02-25
maintainer: moved shapes out of core

KaiSzuttor:


src/core/CMakeLists.txt
```diff
@@ -1,149 +1,148 @@
 set(EspressoCore_SRC
-  accumulators.cpp
```
fweik, line 2: Is there any specific reason for the indentation change? I feel this will just lead to merge conflicts for no good reason.

KaiSzuttor: besides beautification, no

KaiSzuttor: but I think we should think about using ```cmake-format```

src/script_interface/CMakeLists.txt
```diff
@@ -12,30 +12,38 @@ set(EspressoScriptInterface_SRC
...
-    "${CMAKE_CURRENT_SOURCE_DIR}/ScriptInterfaceBase.cpp"
-)
+    "${CMAKE_CURRENT_SOURCE_DIR}/ScriptInterfaceBase.cpp")
 
 if(H5MD)
-    list(APPEND EspressoScriptInterface_SRC
-        "${CMAKE_CURRENT_SOURCE_DIR}/h5md/initialize.cpp"
-        "${CMAKE_CURRENT_SOURCE_DIR}/h5md/h5md.cpp"
-        )
+  list(APPEND EspressoScriptInterface_SRC
```
fweik: I think the mordern way to do this is using [target_sources](https://cmake.org/cmake/help/latest/command/target_sources.html), which allows updating from subdirectories. I like that better because this allows to handle all source files from the CMakeLists.txt in directory they sit in. Would also be an idea for `core`.

src/core/CMakeLists.txt
KaiSzuttor: I can revert the formatting changes if needed

fweik: If a single correct formatting can be enforced by a tool that would be perfect. I think I looked at `cmake-format` once, but the results with the default config were not so good...

KaiSzuttor: that should be possible. It probably has the same issues as ```autopep8```, *i.e.* the exact formatting depends on the version of the tool

src/core/CMakeLists.txt
```diff
@@ -155,19 +154,22 @@ target_include_directories(EspressoCore PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
...
+    EspressoCore PUBLIC "$<$<NOT:$<CXX_COMPILER_ID:Clang>>:-ftest-coverage>")
+  if(NOT CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
     target_link_libraries(EspressoCore PUBLIC gcov)
   endif()
 endif()
 
-target_link_libraries(EspressoCore PUBLIC EspressoConfig utils)
-target_link_libraries(EspressoCore PUBLIC Boost::serialization Boost::mpi MPI::MPI_CXX Random123)
-target_link_libraries(EspressoCore PRIVATE Profiler)
+target_link_libraries(EspressoCore PUBLIC EspressoConfig utils Boost::serialization Boost::mpi MPI::MPI_CXX Random123 shapes PRIVATE Profiler)
```
fweik, line 173: Do all of them have to be `PUBLIC`? E.g. iirc `Random123` is not used on the interface of core.

KaiSzuttor: I have checked...

* Random123: https://github.com/espressomd/espresso/blob/5aab7f7ec5012ee1247de667772f140720829c81/src/core/random.hpp#L29

will check the others

fweik (approved):

jngrad (approved):

src/core/CMakeLists.txt
```diff
@@ -181,11 +183,15 @@ endif(SCAFACOS)
 
 # Subdirectories
 add_subdirectory(io)
-
+if(H5MD)
+target_link_libraries(EspressoCore PUBLIC ${HDF5_LIBRARIES} Boost::filesystem h5xx)
+target_sources(EspressoCore PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/io/writer/h5md_core.cpp)
+target_include_directories(EspressoCore SYSTEM PUBLIC ${HDF5_INCLUDE_DIRS} ${CMAKE_CURRRENT_SOURCE_DIR}/io/writer)
+target_compile_definitions(EspressoCore PUBLIC H5XX_USE_MPI)
+endif(H5MD)
```
jngrad, line 191: This change actually fixes a bug in ESPResSo 4.1.4. In particular, the `${HDF5_LIBRARIES}` needs to be linked against `EspressoCore`, otherwise it is not possible to link `ScriptInterface` unit tests in recent versions of openSUSE due to missing hdf5 symbols. For example, on openSUSE Leap 15.4 x86_64:
```
/usr/lib64/gcc/x86_64-suse-linux/7/../../../../x86_64-suse-linux/bin/ld:
    /home/abuild/rpmbuild/BUILD/espresso/build/src/core/io/writer/H5mdCore.so:
    undefined reference to `H5Pset_fapl_mpio'
collect2: error: ld returned 1 exit status
make[3]: *** [src/core/unit_tests/CMakeFiles/AutoParameters_test.dir/build.make:112:
              src/core/unit_tests/AutoParameters_test] Error 1
```

PR #3544, opened 2020-02-25
Removed max_seen_particle global

fweik:
Removed global and corresponding state tracking, this information can be obtained from
the local particle index.


jngrad (approved):
LGTM, pending CI

PR #3545, opened 2020-02-26
Modernize GPU particle comm

fweik:
Fixes #

Description of changes:
 - Replaced manual memory management by vectors
 - Two special allocators to make this possible
 - Buffer size is adapted on-demand, so the state tracking
   code could be simplified
 - General cleanup, removal of unused stuff

### DIFF
--- a/src/core/CudaDeviceAllocator.hpp
+++ b/src/core/CudaDeviceAllocator.hpp
@@ -1,0 +1,52 @@
+#ifndef ESPRESSO_CUDA_DEVICE_ALLOCATOR_HPP
+#define ESPRESSO_CUDA_DEVICE_ALLOCATOR_HPP
+
+#include <utils/device_qualifier.hpp>
+
+#ifndef DEVICE_COMPILER
+#error "This header can only be used with a CUDA-capable compiler."
+#endif
+
+#include <thrust/device_free.h>
+#include <thrust/device_malloc.h>
+#include <thrust/device_ptr.h>
+#include <thrust/device_reference.h>
+
+/**
+ * @brief Allocator that uses CUDA to allocate GPU memory.
+ *
+ * This allocator uses thrust::malloc and thrust::free to
+ * manage device memory. To be able to have static containers
+ * for device memory, exceptions in deallocations are suppressed.
+ * This is needed because static containers may be destroyed after
+ * the CUDA system was teared down, leading to device_free to fail.
+ *
+ * @tparam T Type to allocate memory for.
+ */
+template <class T> struct CudaDeviceAllocator {
+  using value_type = T;
+  using pointer = thrust::device_ptr<T>;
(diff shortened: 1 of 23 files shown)


src/core/CudaDeviceAllocator.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+  CudaDeviceAllocator() noexcept = default;
+  template <class U>
+  explicit CudaDeviceAllocator(const CudaDeviceAllocator<U> &) {}
+  template <class U> bool operator==(const CudaDeviceAllocator<U> &) const {
+    return true;
+  }
+  template <class U> bool operator!=(const CudaDeviceAllocator<U> &) const {
+    return false;
+  }
```
KaiSzuttor, line 41: can you please explain why this is necessary?

src/core/cuda_common_cuda.cu
```diff
@@ -98,145 +83,25 @@ void _cuda_check_errors(const dim3 &block, const dim3 &grid,
...
-  if (part_index < global_part_vars_device->number_of_particles) {
-    particle_forces_device[3 * part_index + 0] = 0.0f;
-    particle_forces_device[3 * part_index + 1] = 0.0f;
-    particle_forces_device[3 * part_index + 2] = 0.0f;
-#ifdef ROTATION
-    particle_torques_device[3 * part_index + 0] = 0.0f;
-    particle_torques_device[3 * part_index + 1] = 0.0f;
-    particle_torques_device[3 * part_index + 2] = 0.0f;
-#endif
+template <class T> void resize_or_replace(device_vector<T> &vec, size_t n) {
```
KaiSzuttor, line 101: please add a docstring

KaiSzuttor (changes requested):
LGTM besides some missing doc

src/core/cuda_common_cuda.cu
fweik: The docstring would be very similar to the function signature, are you sure
it is required for an internal function?

src/core/CudaDeviceAllocator.hpp
fweik: Otherwise it's not an `Allocator`: https://en.cppreference.com/w/cpp/named_req/Allocator

KaiSzuttor: > returns true only if the storage allocated by the allocator a1 can be deallocated through a2.

so this is always fulfilled?

fweik: Yes, because they all call `cudaFree`.

src/core/cuda_common_cuda.cu
KaiSzuttor: I had the need of existence for this function in mind which should be documented as you have done by now.

KaiSzuttor (approved):

Issue #3550, opened 2020-02-28
HollowConicalFrustum's orientation does not align with its `axis` parameter


pkreissl:
Currently, `HollowConicialFrustum` is visualized via rasterization of its surface via distance calculation &ndash; consequently, its visualization is a direct representation of the respective constraint applied in the core (meaning, this can't just be a visualization bug).

Running the sample script `espressomd/samples/visualization_constraints.py --hollowconicalfrustum`, one observes the expected visualization:
![visualization_constraints_wrong](https://user-images.githubusercontent.com/12623404/75537533-947a1d80-5a17-11ea-9081-0a6b03b72b69.png)
However, there is no change to the system when changing the `axis` parameter of the constraint, e.g., from preset `axis=[0.0, 0.0, 1.0]` to  `axis=[0.0, 1.0, 1.0]` (line 107 in the script).
For clarification, see the following figure, where the grey solid shape shows what one would expect here.
![visualization_constraints_expected](https://user-images.githubusercontent.com/12623404/75537520-917f2d00-5a17-11ea-8b9b-09a141c78a22.png)

pkreissl:
@KaiSzuttor , maybe you want to have a look.

PR #3551, opened 2020-02-28
Improved visualization

pkreissl:
Description of changes:
 - Add visualization of `HollowConicalFrustum` (relies on OpenGL Extrusion library).
 - Improve visualization of `SimplePore` when OpenGL Extrusion library is present.
 - Add capability to handle `Union`s in the visualizer.

The required OpenGL Extrusion library can typically be installed using the standard package manager. It is called `libgle3-dev` on Ubuntu, `libgle` on Fedora.

This PR might depend on issue #3550 &ndash; I implemented the correct visualization, which at the moment does not match what happens in the core ~bur~ but rather what would be expected.


src/python/espressomd/visualization_opengl.pyx
```diff
@@ -783,19 +789,47 @@ class openGLLive:
...
         coll_shape_obj = collections.defaultdict(list)
-        for c in self.system.constraints:
-            if type(c) == espressomd.constraints.ShapeBasedConstraint:
-                t = c.get_parameter('particle_type')
-                s = c.get_parameter('shape')
-                n = s.name()
-                if n in primitive_shapes:
-                    coll_shape_obj[n].append([s, t])
+        for constraint in self.system.constraints:
+            if type(constraint) == espressomd.constraints.ShapeBasedConstraint:
```
KaiSzuttor: ```isinstance(constraint, espressomd.constraints.ShapeBasedConstraint)```

src/python/espressomd/visualization_opengl.pyx
```diff
@@ -783,19 +789,47 @@ class openGLLive:
...
+                ptype = constraint.get_parameter('particle_type')
```
KaiSzuttor, line 828: pep8  says: ```particle_type``` or ```p_type``` or ```part_type```

src/python/espressomd/visualization_opengl.pyx
```diff
@@ -783,19 +789,47 @@ class openGLLive:
         # Collect shapes and interaction type (for coloring) from constraints
         primitive_shapes = [
             'Shapes::Wall', 'Shapes::Cylinder', 'Shapes::Ellipsoid',
-            'Shapes::SimplePore', 'Shapes::Slitpore', 'Shapes::Sphere',
-            'Shapes::SpheroCylinder']
+            'Shapes::HollowConicalFrustum', 'Shapes::SimplePore',
+            'Shapes::Slitpore', 'Shapes::Sphere', 'Shapes::SpheroCylinder']
+
+        def get_shapes_from_union(union):
```
KaiSzuttor: if you want to hide the special case of the union, you could differentiate between the union and the others within this function. You could also probably get rid of ```update_shapes_collection``` and fail late. Just try drawing the ```HollowConicalFrustum``` with GLE and if you fail fall back to ```Shapes::Misc```.

KaiSzuttor (changes requested):
Looks good, maybe we can get rid of some special handling

src/python/espressomd/visualization_opengl.pyx
pkreissl: As discussed offline, pep8 will be part of a general refactor of the visualizer.

KaiSzuttor (approved):

PR #3559, opened 2020-03-03
Rotation params

### LINKED ISSUE #3550: HollowConicalFrustum's orientation does not align with its `axis` parameter

KaiSzuttor:
Fixes #3550

Description of changes:
 - fixing utils function ```rotation_params``` that returned wrong results if the input vector's norm was not unity

### DIFF
--- a/src/utils/tests/vec_rotate_test.cpp
+++ b/src/utils/tests/vec_rotate_test.cpp
@@ -19,7 +19,8 @@
 #define BOOST_TEST_DYN_LINK
 #include <boost/test/unit_test.hpp>
 
-#include "utils/math/vec_rotate.hpp"
+#include <utils/constants.hpp>
+#include <utils/math/vec_rotate.hpp>
 using Utils::vec_rotate;
 
 #include <cmath>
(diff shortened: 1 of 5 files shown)


fweik:
LGTM when you made it work with the minimal required version of boost.

src/utils/tests/vec_rotate_test.cpp
```diff
@@ -45,3 +46,15 @@ BOOST_AUTO_TEST_CASE(rotation) {
...
 }
+
+BOOST_AUTO_TEST_CASE(rotation_params) {
+  Utils::Vector3d v1 = {1.0, 0.0, 0.0};
+  Utils::Vector3d v2 = {1.0, 1.0, 0.0};
+
+  double angle;
+  Utils::Vector3d rotation_axis;
+  std::tie(angle, rotation_axis) = Utils::rotation_params(v1, v2);
+  BOOST_TEST(angle == M_PI_4, boost::test_tools::tolerance(1e-7));
```
fweik: `M_PI_4` is a gnu-ism.

src/utils/tests/vec_rotate_test.cpp
```diff
@@ -23,6 +23,7 @@
 using Utils::vec_rotate;
 
 #include <cmath>
+#include <iostream>
```
jngrad: leftover from `std::cout`?

fweik (approved):

PR #3560, opened 2020-03-03
Reshape output of Observables

### LINKED ISSUE #2233: Proper return type for Observables needed
Observables return floating point tensors, but have no
way of communicating the dimensions of the data (just
the size). We can't use `boost::multi_array` for that because
there the rank is static, but we need dynamic rank for the
Observables.

jngrad:
Fixes #2233

Description of changes:
- observables now return reshaped numpy arrays
- observables return type is documented in the docstrings


src/python/espressomd/observables.py
```diff
@@ -14,15 +14,19 @@
...
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+import numpy as np
 from .script_interface import ScriptInterfaceHelper, script_interface_register
 
 
 @script_interface_register
 class Observable(ScriptInterfaceHelper):
     _so_name = "Observables::Observable"
-    _so_bind_methods = ("calculate", "n_values")
+    _so_bind_methods = ("calculate_flat_array", "shape")
```
fweik: I don't like `calculate_flat_array`, what was wrong with `calculate`? `calculate_flat_array` suggests that it can also calculate something else, which is not the case.

jngrad: Should I rename it `_calculate`? We need a different name to avoid a name collision.

fweik: With what would it collide?

jngrad: It wouldn't. I just tried it, but for some reason you can't have a method starting with an underscore in the `_so_bind_methods` attribute of a `script_interface_register`.

jngrad: nvm, forgot to update the two other scriptinterface files

src/python/espressomd/observables.py
```diff
@@ -14,15 +14,19 @@
...
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+import numpy as np
 from .script_interface import ScriptInterfaceHelper, script_interface_register
 
 
 @script_interface_register
 class Observable(ScriptInterfaceHelper):
     _so_name = "Observables::Observable"
-    _so_bind_methods = ("calculate", "n_values")
+    _so_bind_methods = ("_calculate", "shape")
```
KaiSzuttor: why do you rename it? We should stick to our naming conventions.

jngrad: just tried it, but
```python
def calculate(self):
    return np.array(self.call_method("calculate")).reshape(self.shape())
```
is overshadowed by the `self.calculate()` method created by the `ScriptInterfaceHelper`.

fweik (approved):

RudolfWeeber:
On Tue, Mar 03, 2020 at 05:45:41AM -0800, Jean-Noël Grad wrote:
> 
> 
> 
> >  from .script_interface import ScriptInterfaceHelper, script_interface_register
>  
>  
>  @script_interface_register
>  class Observable(ScriptInterfaceHelper):
>      _so_name = "Observables::Observable"
> -    _so_bind_methods = ("calculate", "n_values")
> +    _so_bind_methods = ("_calculate", "shape")
> 
> just tried it, but
> ```python
> def calculate(self):
>     return np.array(self.call_method("calculate")).reshape(self.shape())
> ```
> is overshadowed by the `self.calculate()` method created by the `ScriptInterfaceHelper`.
Just to be sure:
You can ommit the calculate in bind_methods. This is just a short way of manually implementing
```
def calculate(...):
    self.call_method("calculate", {})
```

KaiSzuttor (approved):

jngrad:
@RudolfWeeber This works, thanks! We discussed the naming convention internally and chose to use the underscore prefix because it allows users to add the underscore prefix in their old 4.1 scripts to get the old behavior back, without having to rework the numpy reshape logic (for example the LB stress tensor requires more than just a reshape operation). This info would only be mentioned in the release notes, and the `_calculate` method would then be removed in espresso 4.3 or 5.0 using some strategy (for example yours). Do you think we should use your solution right now for 4.2?

RudolfWeeber:
On Tue, Mar 03, 2020 at 06:32:46AM -0800, Jean-Noël Grad wrote:
> @RudolfWeeber This works, thanks! We discussed the naming convention internally and chose to use the underscore prefix because it allows users to add the underscore prefix in their old 4.1 scripts to get the old behavior back, without having to rework the numpy reshape logic (for example the LB stress tensor requires more than just a reshape operation). This info would only be mentioned in the release notes, and the `_calculate` method would then be removed in espresso 4.3 or 5.0 using some strategy (for example yours). Do you think we should use your solution right now for 4.2?
Not sure, I understand.
In my opiniion, in 4.2., observables should return the correct shape when using .calculate().
That probably includes the stress tensor. For total stress, it is anyway not guaranteed that it is symmetric, and for lb stress, the format is confusing and error-prone. I'd change that toa a 3x3 matrix as well.
I don't care, whether or not the old behavior stays available as _calculate(), although I'd personally not do it, because people have to adapt their scripts anyway. Then, adapting to the final (physically reasonable) return shape would be best, imo.

jngrad:
> and for lb stress, the format is confusing and error-prone. I'd change that toa a 3x3 matrix as well.

The LB stress tensor is now returned as a symmetric matrix instead of a flattened lower triangle.

> I'd personally not do it, because people have to adapt their scripts anyway.

I'll remove the binding to the old method.

PR #3564, opened 2020-03-04
Virtual Sites Cleanup

fweik:
This is some intermediate cleanup of virtual sites. Biggest change
is that the velocities are now always updated. This has actually very
little performance impact, and is less error prone. 

Description of changes:
 - Simplified interface of `VirtualSites`
 - Some cleanup of the pressure calc
 - Removed `VirtualSites::have_velocity()` mechanism


jngrad (approved):
LGTM, pending CI

jngrad:
```diff
- * @return Flat vector with elements of @ref m.
+ * @return Flat vector with elements of @p m.
```
Just for the explanation of the doxygen warning: argument parameters cannot be referenced with a link, so we format them in teletype font with `@p`.

PR #3568, opened 2020-03-09
Fix deprecation warning from collections.abc

jngrad:
Fix a deprecation warning that will break espresso in Python 3.9.


fweik (approved):

jngrad:
@junghans I've seen Fedora is testing all packages against early pre-releases of Python 3.9. You might need to add that patch to the espresso 4.1.2 specfile once the MPICH issue in [Bugzilla 1793563](https://bugzilla.redhat.com/show_bug.cgi?id=1793563) is resolved.

junghans:
@jngrad thanks!

PR #3570, opened 2020-03-10
Parametric weight function for DPD....

bindgens1:
... with a parabolic weight function

Description of changes:
- New optional parameter "k" in `DPDInteraction` for the exponent for the weight function

### DIFF
--- a/doc/sphinx/inter_non-bonded.rst
+++ b/doc/sphinx/inter_non-bonded.rst
@@ -529,12 +529,13 @@ The parameters can be set via::
     system.non_bonded_inter[type1, type2].dpd.set_params(**kwargs)
 
 This command defines an interaction between particles of the types ``type1`` and ``type2``
-that contains velocity-dependent friction and noise according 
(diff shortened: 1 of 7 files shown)


fweik:
~~In principle the `type` parameter is obsolete now, no? `type = 0` is just the same as `k = 0`.~~
Ah sorry, this is of course nonsens.

bindgens1:
yeah 1- x^0 is zero and the weight function is totally gone.

Couldn't find a way to simplify it even more.

fweik (changes requested):
Ok, this looks good to me, but you should add a bit more documentation. In the main docs of the
DPD interaction (6.1.14. DPD interaction in `doc/sphinx/inter_non-bonded.rst`) you should
include the exponent factor into the formulas for the weight function and maybe add a literature
reference to the paper that describes the physical effect of changing the parameter.

fweik:
LGTM, thank you for your contribution. I'll take it from here.

fweik (approved):

PR #3571, opened 2020-03-10
Partial ROCm 3.1 support

mkuron:
Make CMake changes. Needs `ln -s /opt/rocm/bin/hcc* /opt/rocm/hip/bin/` to work around bug in hipcc linker wrapper. Build then succeeds, but anything that passes GPU pointers between compilation units (most notably, EK and LB) fails and all tests deadlock in the AMD HSA shutdown procedure.


CMakeLists.txt
```diff
@@ -187,13 +187,13 @@ if(WITH_CUDA)
       message(STATUS "Found HIP compiler: ${HIP_HIPCC_EXECUTABLE}")
       set(CUDA 1)
       set(HIP 1)
-      list(APPEND HIP_HCC_FLAGS "-I${HIP_ROOT_DIR}/include -Wno-c99-designator -Wno-macro-redefined -Wno-duplicate-decl-specifier -std=c++14")
+      list(APPEND HIP_HCC_FLAGS "-I${HIP_ROOT_DIR}/include -I${HIP_ROOT_DIR}/../include -Wno-c99-designator -Wno-macro-redefined -Wno-duplicate-decl-specifier -std=c++14")
```
jngrad: `HIP_HCC_FLAGS` expands to `"-I/opt/rocm/include -I/opt/rocm/../include -Wno-c99..."` on ROCm 3.0, but `/opt/rocm/../include` doesn't exist. Is there a way in CMake to find the path of the parent ROCm folder? This would allow us to write `"-I${HIP_ROOT_DIR}/include -I${ROCM_ROOT_DIR}/include -Wno-c99..."` and save us some headache when debugging this CMake logic.

mkuron: Unfortunately there is not. So maybe we should merge your patch instead of this pull request that contains an explicit version check.

jngrad: We could do like in [pytorch/pytorch:torch/utils/cpp_extension.py#L53-L73](https://github.com/pytorch/pytorch/blob/d30fa4837e10505d172e0cc7a6849b1a07aeb573/torch/utils/cpp_extension.py#L53-L73), e.g.
```cmake
      string(REGEX REPLACE "^(.+)/bin/hipcc$" "\\1" ROCM_HOME ${HIP_HIPCC_EXECUTABLE})
      list(APPEND HIP_HCC_FLAGS "-I${HIP_ROOT_DIR}/include -I${ROCM_HOME}/include -Wno-c99-designator -Wno-macro-redefined -Wno-duplicate-decl-specifier -std=c++14")
```
But this seems superfluous given that we already hardcode the path to `/opt/rocm`:
https://github.com/espressomd/espresso/blob/a4596e3e07e3c948a52217a32ced0e962f0bf165/CMakeLists.txt#L182

Why not directly do the following?
```cmake
set(ROCM_HOME "/opt/rocm")
list(APPEND CMAKE_MODULE_PATH "${ROCM_HOME}/hip/cmake") 
# ...
list(APPEND HIP_HCC_FLAGS "-I${HIP_ROOT_DIR}/include -I${ROCM_HOME}/include -Wno-c99-designator -Wno-macro-redefined -Wno-duplicate-decl-specifier -std=c++14")
```

KaiSzuttor:
I think in general we should try to keep these details out of the main ```CMakeLists.txt```. Otherwise this file will be unreadable at some point. So whatever the solution is, please put it in a separate cmake module in the ```cmake``` directory of the project root.

CMakeLists.txt
mkuron: `ROCM_HOME` sounds like a sensible solution.

mkuron:
@KaiSzuttor, it makes no sense to move this stuff elsewhere because the usual place to do all the library detection is in the main CMakeLists.txt. CUDA detection takes up just as much space.

KaiSzuttor:
> it makes no sense to move this stuff elsewhere because the usual place to do all the library detection is in the main CMakeLists.txt. CUDA detection takes up just as much space.

Can you please explain this statement? Just because there is mess elsewhere in the file it does not mean that we should not clean up and extend the mess at another place.

jngrad:
> So whatever the solution is, please put it in a separate cmake module in the `cmake` directory of the project root.

I'm currently trying to refactor the CMake logic for detecting CUDA. We are using the  `FindCUDA` code that has been [deprecated in CMake 3.10](https://cmake.org/cmake/help/v3.10/module/FindCUDA.html) and doesn't fully support CUDA 10.0+. For example, the cublas error that has been blocking the Stokesian Dynamics PR for weeks is resolved by simply using the CMake native CUDA support. As part of the refactor, I thought about moving some of the logic to dedicated `*.cmake` files in the espresso `/cmake` folder, e.g. the definitions of `add_gpu_library` and some library path detection and variable creation logic. However we'll probably have to keep a minimal if/else structure in the main `CMakeLists.txt` file for the 3 CUDA compilers we detect.

KaiSzuttor:
> it makes no sense to move this stuff elsewhere because the usual place to do all the library detection is in the main CMakeLists.txt

no, it happens in the ```Find*.cmake``` files

mkuron:
> no, it happens in the Find*.cmake files

Setting compiler flags, which is what we are currently discussing, shouldn't happen inside FindHIP.cmake. The choice of compiler flags is Espresso-specific. Find*.cmake files are supposed to be provided either by CMake or by the respective library to deal with things common to all use cases.

KaiSzuttor:
Compiler flags should be set on targets not globally, so they neither belong into the main cmake file nor into FindHIP.cmake

KaiSzuttor:
[An introduction to modern cmake](https://cliutils.gitlab.io/modern-cmake/chapters/basics/structure.html):
> You often want a cmake folder, with all of your helper modules. This is where your Find*.cmake files go.

jngrad:
Please don't invest any more time on this PR, I'm including it in my CMake refactoring PR. It's taking more time than anticipated because the `FindCython.cmake` is also broken.

PR #3445, opened 2020-02-04
WIP: Stokesian Dynamics

biermanncarl:
Fixes #3241

Issue #2977, opened 2019-07-08
Fast math breaks some tests


bindgens1:
I have two failing tests on our cluster with the current development branch:

1)

```espresso/src/core/unit_tests/grid_test.cpp(79)```
+ 
```espresso/src/core/unit_tests/grid_test.cpp(82)```

Fails with:  
```error: in "get_mi_coord_test": absolute value of std::abs(get_mi_coord(a, b, box_l, true) - (a - b) - box_l){4.4408920985006262e-16} exceeds 2.2204460492503131e-16```
+ 
```error: in "get_mi_coord_test": absolute value of std::abs(get_mi_coord(b, a, box_l, true) - (b - a) + box_l){4.4408920985006262e-16} exceeds 2.2204460492503131e-16```

This seems to be the same issue twice

2) 
```
Test  #44: rotational_inertia ......................***Failed

FF
======================================================================
FAIL: test_energy_and_momentum_conservation (__main__.RotationalInertia)
----------------------------------------------------------------------
Traceback (most recent call last):
/espresso/build/testsuite/python/rotational_inertia.py", line 154, in test_energy_and_momentum_conservation
    self.assertAlmostEqual(self.energy(p), E0, places=3)
AssertionError: 14.85059589306411 != 14.85 within 3 places

======================================================================
FAIL: test_stability (__main__.RotationalInertia)
----------------------------------------------------------------------
Traceback (most recent call last):
/espresso/build/testsuite/python/rotational_inertia.py", line 84, in test_stability
    i, k, self.L_0_lab[k], self.L_lab[k]))
AssertionError: 0.004388219520887526 not less than or equal to 0.004 : Inertial motion around stable axis J1: Deviation in angular momentum is too large. Step 2, coordinate 2, expected -0.7955, got -0.7998882195208875

----------------------------------------------------------------------
Ran 2 tests in 0.096s

FAILED (failures=2)

```

I compiled Espresso with the intel compiler. Are these known issues that occured before?
In case the intel compiler is no longer supported, I will try again with gcc the next days.

**Comment by mkuron** [comment]
The Intel compiler is supported and we last ran CI tests with version 2018 Update 4 about one week ago. I've just triggered a build on the current master to make sure that it still works: https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/135095.

Which version are you using? Did you enable any non-standard optimizations? What CPU are you running on?

**Comment by bindgens1** [comment]
I didn't specify any special compiler flags.

```
[...]

PR #3574, opened 2020-03-10
Refactor CMake package inclusion

jngrad:
Description of changes:
- move logic to import packages from `CMakeLists.txt` to dedicated helper files `cmake/Find<package>.cmake` for `find_package()`
- enforce the Cython version requested in `CMakeLists.txt`
- CMake now fails if `WITH_CUDA` is set to true but no CUDA-capable compiler is found
- CMake now fails if `WITH_CLANG_TIDY` is set to true but Clang-tidy is not found or its version doesn't match the Clang compiler version
- drop deprecated `FindCUDA` in favor of native CUDA support in CMake 3.10 (required for  #3445)
- add partial support for ROCm 3.1 (closes #3571, required for espressomd/docker#156)


cmake/FindCUDACompiler.cmake
```diff
@@ -0,0 +1,126 @@
...
+        get_filename_component(NVCC_EXECUTABLE_DIRNAME "${CUDA_NVCC_EXECUTABLE}" DIRECTORY)
+        get_filename_component(NVCC_EXECUTABLE_DIRNAME "${NVCC_EXECUTABLE_DIRNAME}" DIRECTORY)
+        message(WARNING "Your nvcc (${CUDA_NVCC_EXECUTABLE}) does not appear to match your CUDA libraries (in ${CUDA_TOOLKIT_ROOT_DIR}). While Espresso will still compile, you might get unexpected crashes. Please point CUDA_TOOLKIT_ROOT_DIR to your CUDA toolkit path, e.g. by adding -DCUDA_TOOLKIT_ROOT_DIR=${NVCC_EXECUTABLE_DIRNAME} to your cmake command.")
+      endif()
+      set(CUDA 1)
+      set(CUDA_COMPILER_EXE ${CUDA_NVCC_EXECUTABLE})
+
+      set(CUDA_LINK_LIBRARIES_KEYWORD PUBLIC)
+
+      set(CUDA_NVCC_FLAGS_DEBUG "${CUDA_NVCC_FLAGS_DEBUG} -g")
```
KaiSzuttor, line 100: we should probably introduce a pseudo interface library where we add compile flags and later link against, *e.g.*:
```
add_library(gpu_interface INTERFACE)
target_compile_options(gpu_interface INTERFACE "$<$<CONFIG:DEBUG>:-g>")
```

jngrad:
@KaiSzuttor here is a draft of the next PR where we split the CUDA compiler detection mechanism in separate files: [jngrad/espresso:refactor-cmake-with_gpu](https://github.com/jngrad/espresso/commits/refactor-cmake-with_gpu)

maintainer/CI/build_cmake.sh
```diff
@@ -130,7 +130,10 @@ if [ "${with_coverage}" = true ]; then
 fi
 
 cmake_params="-DCMAKE_BUILD_TYPE=${build_type} -DWARNINGS_ARE_ERRORS=ON -DTEST_NP:INT=${check_procs} ${cmake_params}"
-cmake_params="${cmake_params} -DCMAKE_CXX_FLAGS=${cxx_flags} -DCUDA_NVCC_FLAGS=${nvcc_flags}"
+cmake_params="${cmake_params} -DCMAKE_CXX_FLAGS=${cxx_flags}"
```
KaiSzuttor, line 133: why do we manually set cxx flags here?

KaiSzuttor (commented):
that improves readability a lot! thanks. Could you please look at my comments, maybe we can get rid of some global flags.

cmake/FindCUDACompiler.cmake
jngrad: we definitely need to do that, but I would prefer to look at it in the next PR, where these flags have been cleaned up and where troubleshooting the CMake logic for different GPU compilers is a lot easier thanks to the new CMake flag

maintainer/CI/build_cmake.sh
jngrad: Maybe we can replace that logic with the corresponding `-DCMAKE_BUILD_TYPE=Debug`. This is again overlapping with the next PR, and might also depend on #2977:
https://github.com/espressomd/espresso/blob/be99119c8113b1f2a40ca51d5b07b738bf981eee/.gitlab-ci.yml#L407

KaiSzuttor: all right, let's do it in the other PR

KaiSzuttor (approved):

KaiSzuttor:
> @KaiSzuttor here is a draft of the next PR where we split the CUDA compiler detection mechanism in separate files: jngrad/espresso:refactor-cmake-with_gpu

just open a draft PR, LGTM

PR #3575, opened 2020-03-11
Visualizer cleanup

pkreissl:
On the way of improving the OpenGL visualizer code &hellip;

Description of changes:
 - Put (almost) all stuff concerning the visualization of shapes in a `Shape` class and respective child classes (`Sphere(Shape)` etc.)
 - Use Rodrigues' rotation formula in calculation of `Camera` rotation &ndash; gets rid of unnecessary complexity, `Quaternion`.
 - In general: attempt to simplify, make code more readable, and use variable names according to pep8.

A first step into the right direction, I hope &hellip;


src/python/espressomd/visualization_opengl.pyx
```diff
@@ -345,8 +345,7 @@ class openGLLive:
         # OVERWRITE WITH USER PROPERTIES
         for key in kwargs:
             if key not in self.specs:
-                raise ValueError(
-                    key + ' is not a valid visualization property')
+                raise ValueError(f'{key} is no valid visualization property')
```
jngrad, line 344: f-strings are only available starting from Python 3.6 ([PEP 498](https://www.python.org/dev/peps/pep-0498/)). I think we should bump espresso to python 3.6 ASAP, python 3.5 was dropped from NEP 29 two months ago ([NEP 29 support table](https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table)).

jngrad: Bumping python will conflict with the currently opened PR on CMake refactor... Since no test fails here, we can probably keep these f-strings and bump python in a few days in a separate PR, where we could also bump cython and python modules accordingly.

jngrad:
Thanks for this refactor, it's a lot more readable! I had a few refactoring ideas on my todo list and will annotate them in the code review, if you didn't already fix them.

@christophlohrmann could you please review this PR by running visualization samples again?

src/python/espressomd/visualization_opengl.pyx
```diff
@@ -2016,6 +1878,519 @@ class openGLLive:
...
+        quadric = OpenGL.GLU.gluNewQuadric()
+        # basic position and orientation
+        OpenGL.GL.glTranslate(self.center[0], self.center[1], self.center[2])
+        ax, rx, ry = rotation_helper(self.axis)
+        OpenGL.GL.glRotatef(ax, rx, ry, 0.0)
+
+        if bool(OpenGL.GLE.gleSpiral):
+            self._draw_using_gle()
+        else:
+            self._draw_using_primitives(quadric)
```
jngrad: You can move the quadric creation and deletion within the body of this `else:` clause. There is no need to waste memory allocation+deallocation on the heap for a large object that is not used.

src/python/espressomd/visualization_opengl.pyx
```diff
@@ -2016,6 +1878,519 @@ class openGLLive:
...
+        if bool(OpenGL.GLE.gleSpiral):
+            self._draw_using_gle()
+        else:
+            super().draw()
+
+    # if available, use the GL Extrusion library
+    def _draw_using_gle(self):
+        set_solid_material(self.color, self.material)
+        OpenGL.GL.glPushMatrix()
+        quadric = OpenGL.GLU.gluNewQuadric()
```
jngrad: This quadric object is not used, you can remove its creation and deletion.

src/python/espressomd/visualization_opengl.pyx
jngrad: Actually, you can move the quadric creation/deletion directly in the body of `self._draw_using_primitives()`.

jngrad (changes requested):

src/python/espressomd/visualization_opengl.pyx
```diff
@@ -1490,47 +1356,47 @@ class openGLLive:
...
-                k, (maxlen - len(k)) * ' ', v)
-            self._draw_text(
-                10, self.specs['window_size'][1] - 10 - 15 * y, txt, self.text_color)
+        for k, v in particle_dict.items():
+            txt = f"{k} {(max_len - len(k)) * ' '} {v}"
+            self._draw_text(10, self.specs['window_size'][1] - 10 - 15 * y,
+                            txt, self.text_color)
             y += 1
 
     # CALLED ION WINDOW POSITION/SIZE CHANGE
```
jngrad:
```suggestion
    # CALLED ON WINDOW POSITION/SIZE CHANGE
```
Also, the "etc." latin abbreviation can only be followed by a single dot in https://github.com/espressomd/espresso/blob/afd52ecc2af6f4924b9f65b1ad7e0b4c633ae3cd/src/python/espressomd/visualization_opengl.pyx#L1258
```diff
-    # COLORS, MATERIALS ETC..
+    # COLORS, MATERIALS, ETC.
```

jngrad (changes requested):

jngrad:
A more general thought that goes beyond the scope of that PR: could this refactoring enable us to split the `visualization_opengl.pyx` file into multiple ones? This Cython file generates a 100'000-line long C++ file that takes a while to compile. This significantly slows down debugging of well-hidden errors (my recent PR on the OpenGL memory leak was not enjoyable) and probably exposes more import statements that necessary in the main Cython file.

src/python/espressomd/visualization_opengl.pyx
KaiSzuttor: the raise of the ```ValueError``` is probably untested.

christophlohrmann (changes requested):
From running the samples:
- line 1270 in ``visualization_opengl.pyx`` : ``maxq`` -> ``max_q``
- lines 84 and 87 in ``samples/visualization_interactive.py`` : ``keyboardManager`` -> ``keyboard_manager``
- in general: all nontrivial, rasterized constraints run very slow now

KaiSzuttor:
> in general: all nontrivial, rasterized constraints run very slow now

@pkreissl could you please check

src/python/espressomd/visualization_opengl.pyx
jngrad: the visualizer is not tested. Our CI has python 3.5 only in the clang-6 container, but it cannot go away until we deal with walberla to rely on clang-9 instead of clang-6.

KaiSzuttor:
you don't necessarily have to push each commit separately :smile:  think of the climate

pkreissl:
Alright, the speed issue should be gone now. The problem was that shape recalculated the rasterization every time, when in fact it should just have drawn the same positions as seen from a potentially different viewpoint (Camera).

christophlohrmann:
speedup confirmed

samples/visualization_interactive.py
```diff
@@ -81,10 +81,10 @@ def decreaseTemp():
 
 
 # Register button callbacks
-visualizer.keyboardManager.register_button(
+visualizer.keyboard_manager.register_button(
```
jngrad, line 84: please do `grep -rF keyboardManager samples/ doc/`

KaiSzuttor:
LB visualization is also a lot slower

jngrad (changes requested):

jngrad:
@pkreissl why isn't `maintainer/CI/fix_style.sh` working on your machine?

pkreissl:
> @pkreissl why isn't `maintainer/CI/fix_style.sh` working on your machine?

Probably it is. However, I am currently working on my home machine, on which I can't run the script due to missing `clang-format_6.0` (the version installed here is `9.0.1`) &hellip;

jngrad:
Wouldn't it be less frustrating for you to remove the `clang-format` stuff and fix the python issues at once, rather than struggle against the CI pipeline? The CI pipeline will always win.

src/python/espressomd/visualization_opengl.py
```diff
@@ -2356,7 +2354,9 @@ def draw(self):
 
 # OPENGL DRAW WRAPPERS
 
-def set_solid_material(color, material=[0.6, 1.0, 0.1, 0.4, 1.0]):
+def set_solid_material(color, material=None):
```
jngrad: Why this change? This pylint warning is fixed by transforming the list into a tuple.

pkreissl: My bad, a rather obvious solution &hellip;

KaiSzuttor:
please use ssh. Otherwise this is an endless game CI vs @pkreissl

KaiSzuttor:
Did you find out why LB visualization is much slower?

pkreissl:
> Wouldn't it be less frustrating for you to remove the `clang-format` stuff and fix the python issues at once, rather than struggle against the CI pipeline? The CI pipeline will always win.

It most certainly would &hellip; Problem is, that just removing the `clang-format` stuff does not really help as also my `autopep8` and  `pycodestyle` have more recent versions than required (`1.4.3` and `2.5.0`).

pkreissl:
> Did you find out why LB visualization is much slower?

Offline discussion: actually LB visualization itself seems to be as fast as before. However, start-up time of the visualization script `visulalization_poiseuille.py` has increased. I noticed similar behaviour for other visualization scripts (like `visulalization_constraints.py`).
Will have another look at that.

KaiSzuttor:
> It most certainly would … Problem is, that just removing the clang-format stuff does not really help as also my autopep8 and pycodestyle have more recent versions than required (1.4.3 and 2.5.0).

docker / virtualenv

src/python/espressomd/visualization_opengl.py
```diff
@@ -32,8 +32,7 @@
 from .particle_data import ParticleHandle
 
 
-class openGLLive:
-
+class openGLLive(object):
```
KaiSzuttor: not needed

src/python/espressomd/visualization_opengl.py
```diff
@@ -1836,7 +1840,7 @@ def _init_opengl(self):
 
 # END OF MAIN CLASS
 
-class Shape:
+class Shape(object):
```
KaiSzuttor: inheritance not needed

KaiSzuttor (commented):
btw. wouldn't it be a great performance benefit if it would be possible to ignore particles of given type completely?

pkreissl:
> Your pull request does not meet our code style rules. Pylint summary:
> 
> ```
> ************* Module visualization_opengl
> src/python/espressomd/visualization_opengl.py:1435:31: W0613: Unused argument 'data' (unused-argument)
> src/python/espressomd/visualization_opengl.py:1508:38: W0613: Unused argument 'pos_old' (unused-argument)
> src/python/espressomd/visualization_opengl.py:1517:35: W0613: Unused argument 'pos' (unused-argument)
> src/python/espressomd/visualization_opengl.py:1517:40: W0613: Unused argument 'pos_old' (unused-argument)
> ```
> 
> You can generate these warnings with `maintainer/CI/fix_style.sh`. This is the same command that I have executed to generate the log above.

@KaiSzuttor , @jngrad , i am pretty sure that these warnings are false positives. The OpenGL callbacks pass values to the respective functions called, even if they are not used, there. Any ideas? Or how do I silence `pylint`?

@KaiSzuttor , could you please check whether the speed problem is fixed now? Turns out rasterizing a shape in `__init__` (even if visualization never uses rasterization) is not the best idea &hellip; ;)

jngrad:
@pkreissl I think your last commit 27c58fd accidentally re-introduced bad code. The pylint warnings are true positives: the flagged arguments are never used.

KaiSzuttor:
i mean look at the date

jngrad:
@pkreissl Nevermind, if it's a callback with a mandatory signature, you need to silence the pylint warning by adding `# pylint: disable=unused-argument` above the function

src/python/espressomd/visualization_opengl.py
pkreissl: True, as we are using Python 3 &ndash; followed the suggestion of pycharm not realizing that I still had it configured for Python 2, where explicit inheritance was necessary to prevent use of old-style classes. Removed it again.

KaiSzuttor:
```
./pypresso ../samples/visualization_constraints.py --stomatocyte

[...]

  File "_ctypes/callbacks.c", line 234, in 'calling callback function'
  File "/tikhome/kai/espresso/build/src/python/espressomd/visualization_opengl.py", line 1369, in display
    self._display_all()
  File "/tikhome/kai/espresso/build/src/python/espressomd/visualization_opengl.py", line 1270, in _display_all
    self._draw_system()
  File "/tikhome/kai/espresso/build/src/python/espressomd/visualization_opengl.py", line 901, in _draw_system
    self._draw_constraints()
  File "/tikhome/kai/espresso/build/src/python/espressomd/visualization_opengl.py", line 954, in _draw_constraints
    shape.draw()
  File "/tikhome/kai/espresso/build/src/python/espressomd/visualization_opengl.py", line 1875, in draw
    for point in self.rasterized_surface_points:
TypeError: 'NoneType' object is not iterable
```

christophlohrmann:
``visualization_poiseuille.py`` still takes forever to start

pkreissl:
> `visualization_poiseuille.py` still takes forever to start

Actually, I don't think so... at least for me everything works with fast start-up...

KaiSzuttor:
can we put this into numbers?

christophlohrmann (approved):

jngrad:
The slowdown was most likely caused by the presence of `visualization_opengl.so` and `visualization_opengl.py` in the build folder (CMake doesn't erase old build files), causing Python to load the last one in alphabetical order, aka the outdated .so file.

jngrad (approved):

KaiSzuttor:
in my alphabet “p” comes before “s” ^^

jngrad:
that's what I said^^ .so comes last, therefore Python overwrites the .py path by the .so path. You can try it locally by creating two empty files with the same name but with .so and .py extension and import it: the .so one will be loaded.

PR #3578, opened 2020-03-12
Fix returned array shape for accumulator

### LINKED ISSUE #3577: observable accumulators do not return correct shape
E.g.: ```TimeSeries``` returns an array of shape ```(N_samples, flat_observable_shape)``` instead of ```(N_samples, observable_shape)```

KaiSzuttor:
Fixes part of #3577

Description of changes:
 - Test for correct shape
 - Return correct shape from observable accumulators

### DIFF
--- a/testsuite/python/accumulator.py
+++ b/testsuite/python/accumulator.py
@@ -29,6 +29,9 @@ import espressomd.observables
 import espressomd.accumulators
 
 
+N_PART = 4
+
+
 class AccumulatorTest(ut.TestCase):
 
     """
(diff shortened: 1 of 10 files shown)


testsuite/python/accumulator.py
```diff
@@ -29,6 +29,9 @@
 import espressomd.accumulators
 
 
+N = 4
```
jngrad: I would prefer `N_PART` for consistency with other tests, and because it's simpler to grep a word than a single character.

jngrad (dismissed):
LGTM, pending the inheritance fix

KaiSzuttor:
just talked to @fweik. He prefers to implement a core function that returns the shape. I will have a look.

fweik:
I don't think this should be done on the python level. Determining the shape require knowledge of the core implementation, and should therefor be done there.

fweik (approved):

PR #3582, opened 2020-03-12
Refactor CMake logic for the CUDA compiler

jngrad:
Description of changes:
- the CUDA compiler guessing mechanism was removed
- the `WITH_CUDA` option is now opt-in
- for non-standard CUDA compilers (Clang, HIP), the user has to provide an extra string option `WITH_CUDA_COMPILER` (any of "nvcc", "clang", "hip")
- if the requested CUDA compiler is not found or is too old, CMake will fail
- the minimal CUDA compiler version numbers are defined in the top-level `CMakeLists.txt`
- the convoluted (and incorrect) CXX optimization flag deduction mechanism in `build_cmake.sh` was replaced with dedicated build types (`Coverage` and `RelWithAssert`)


CMakeLists.txt
```diff
@@ -19,7 +19,8 @@
 #
 
 cmake_minimum_required(VERSION 3.10)
-if(${CMAKE_VERSION} VERSION_GREATER_EQUAL 3.13)
+message(STATUS "The CMake identification is ${CMAKE_VERSION}")
```
KaiSzuttor: why not just "CMake version: ${CMAKE_VERSION}"

CMakeLists.txt
```diff
@@ -286,7 +287,24 @@ set(CMAKE_CXX_EXTENSIONS OFF)
...
+  set(WITH_CUDA_COMPILER "nvcc" CACHE STRING "Select the CUDA compiler, can be 'nvcc' (default), 'hip', or 'clang'")
+  if (WITH_CUDA_COMPILER STREQUAL "nvcc")
+    find_package(CUDA ${CMAKE_CUDA_STANDARD} REQUIRED)
+    find_package(CUDACompilerNVCC ${CMAKE_CUDA_STANDARD} REQUIRED)
+  elseif (WITH_CUDA_COMPILER STREQUAL "hip")
+    set(ROCM_HOME "/opt/rocm" CACHE FILEPATH "Path to AMD ROCm")
+    list(APPEND CMAKE_MODULE_PATH "${ROCM_HOME}/hip/cmake")
+    find_package(HIP 1.5.18494 MODULE REQUIRED)
+    set(HCC_PATH "${HIP_ROOT_DIR}")
+    find_package(HIP 1.5.18494 MODULE REQUIRED)
```
KaiSzuttor: why do you have to call ```find_package(HIP``` two times?

jngrad: it's explained in 9ac0c09e7

cmake/FindCUDACompilerNVCC.cmake
```diff
@@ -0,0 +1,48 @@
...
+  message(WARNING "Your nvcc (${CUDA_NVCC_EXECUTABLE}) does not appear to match your CUDA libraries (in ${CUDA_TOOLKIT_ROOT_DIR}). While ESPResSo will still compile, you might get unexpected crashes. Please point CUDA_TOOLKIT_ROOT_DIR to your CUDA toolkit path, e.g. by adding -DCUDA_TOOLKIT_ROOT_DIR='${NVCC_EXECUTABLE_DIRNAME}' to your cmake command.")
+endif()
+
+set(CUDA_LINK_LIBRARIES_KEYWORD PUBLIC)
+
+SET(CUDA_PROPAGATE_HOST_FLAGS OFF)
+
+set(CUDA_NVCC_FLAGS_DEBUG "${CUDA_NVCC_FLAGS_DEBUG} -g")
+set(CUDA_NVCC_FLAGS_RELEASE "${CUDA_NVCC_FLAGS_RELEASE} -O3 -DNDEBUG")
+set(CUDA_NVCC_FLAGS_MINSIZEREL "${CUDA_NVCC_FLAGS_MINSIZEREL} -Os -DNDEBUG")
```
jngrad: several issues here:
* `-Os` is not a valid optimization level: `nvcc fatal   : 's': expected a number`
* there seems to be a difference between `-O2`, `-Xptxas -O2`, `-Xcompiler -O2` ([StackOverflow](https://stackoverflow.com/a/43845984))

CMakeLists.txt
jngrad: The next line in the CMake log is `-- The C compiler identification is GNU 7.5.0`, but I wouldn't mind using your formulation.

jngrad: formulation was changed

CMakeLists.txt
jngrad: I've added an extra CMake comment on top of the git commit explanation.

cmake/FindCUDACompilerNVCC.cmake
jngrad: progress:
* CUDA compiler flags are actually quite complicated. Also, device and host compiler optimization flags don't mix well.
* We don't need to touch the host and device flags in `build_cmake.sh`. The logic is wrong anyway, and it is partially overridden by CMake in Debug builds.

I'm currently removing a lot of useless code in `build_cmake.sh`, but getting the CMake logic right will take me a few more days, I'm afraid. After that, we'll be able to experiment with CMake pseudo-interfaces for GPU targets.

jngrad: I wasted two man hours investigating a Clang compiler error, because when we do
https://github.com/espressomd/espresso/blob/d67ab4aefd263997cbec82015052a19f48b48385/src/core/CMakeLists.txt#L147
we forward CUDA compiler flags to regular .cpp files:
https://github.com/espressomd/espresso/blob/d67ab4aefd263997cbec82015052a19f48b48385/CMakeLists.txt#L165
even though a few lines later there's a mechanism to set those flags to .cu files only:
https://github.com/espressomd/espresso/blob/d67ab4aefd263997cbec82015052a19f48b48385/CMakeLists.txt#L172

jngrad: I wasted one man hour because neither `CMakeLists.txt` nor `build_cmake.sh` throw an error when using the incompatible combination of CMake flags `-DCUDA_NVCC_EXECUTABLE=$(which clang++) -DWITH_COVERAGE=ON`.

jngrad: My last PR introduced a regression: the boost version depends on the CUDA version, so CUDA must be loaded before boost...

jngrad: Why do we override C++ flags only for release builds and not all other types of build???
https://github.com/espressomd/espresso/blob/f291cf448f30b28488830e65556436659870bbf7/src/core/CMakeLists.txt#L78-L79

jngrad: Why `-01` specifically? And why only for release builds?
https://github.com/espressomd/espresso/blob/f291cf448f30b28488830e65556436659870bbf7/CMakeLists.txt#L347-L348

mkuron: FindCUDA didn‘t really have a concept of debug vs. release, so we manually set these flags to match the host compiler and get debuggable or optimized CUDA code as requested.

jngrad: > FindCUDA didn‘t really have a concept of debug vs. release

doesn't it since 3.0? [`CUDA_NVCC_FLAGS_<CONFIG>`](https://cmake.org/cmake/help/v3.0/module/FindCUDA.html?highlight=CUDA_NVCC_FLAGS_%3CCONFIG%3E)

HIP also does: [`HIP_<COMPILER>_FLAGS_<CONFIG>`](https://github.com/ROCm-Developer-Tools/HIP/blob/821c60a3d98f5f4d6cda9440c1452f73edb93b5a/cmake/FindHIP.cmake#L15-L19)

mkuron: I believe that didn't set the flags for both host and device, but that may no longer be true. The `-g -O1` for `WITH_ASAN` is a compromise between good sanitizability and reasonable performance. In Debug mode, you have `-O0 -g` anyway.

jngrad:
Some jobs are taking longer to compile espresso, yet the test time remains the same. This could be due to caching. One exception is the tutorial job, where the tutorial and sample tests take almost twice longer to run (on the same `coyote11`). On my machine, removing `CUDA_NVCC_FLAGS_DEBUG` didn't change the samples test time, so I'm not sure what the reason might be.

The changes are best visualized by removing the first two commits in the global diff. I'll now work on writing generator expressions in targets instead of writing flags in global variables, and try using target interfaces to reduce code duplication.

jngrad:
I can now confirm that the only difference between `python` and 61e02dc8c10 is that `-0g` was added to all C++ files and to the CUDA files processed by Clang. This might explain why compilation takes longer, however the Clang CI job should take less time to run the tests thanks to `-Og`, but now it takes 60% more time to run them. Same thing for tutorials, they should run faster.

There is no difference in line coverage between `-Og` and `-O0` for maxset (unit + integration tests).

KaiSzuttor:
any news here?

jngrad:
I finally found out the slowdown was caused by an optimization flag `-Og` being injected last in the list of compiler flags, overriding the `-O3` flag (the last optimization flag takes precedence). I now have to figure out which part of the CMake logic is responsible for that.

jngrad:
Looks like the C++ flags were always set to `-O3` unless coverage is enabled... The original `build_cmake.sh` is definitely broken.

jngrad:
oh come on, `build_cmake.sh` doesn't print the last third of the `$cmake_params` string, I wasted another 2 man hours fixing the wrong compiler flags

jngrad:
hmm... `set(EspressoCuda_SRC` contains a mix of .cpp and .cu files. We cannot use the solution
```cmake
add_library(gpu_interface INTERFACE)
target_compile_options(gpu_interface INTERFACE "$<$<CONFIG:DEBUG>:-g>")
target_link_libraries(${ARGV0} PUBLIC gpu_interface)
```
because CUDA-specific flags then apply to .cpp files as well, throwing a compiler error. We also can't use `$<$<COMPILE_LANGUAGE:CXX>:-g>` because it's not portable according to the [docs](https://cmake.org/cmake/help/v3.10/manual/cmake-generator-expressions.7.html#manual:cmake-generator-expressions(7)):
> This generator expression has limited use because it is not possible to use it with the Visual Studio generators. Portable buildsystems would not use this expression, and would create separate libraries for each source file language instead:
> ```cmake
> add_library(myapp_c foo.c)
> add_library(myapp_cxx bar.cpp)
> target_compile_options(myapp_cxx PUBLIC -fno-exceptions)
> add_executable(myapp main.cpp)
> target_link_libraries(myapp myapp_c myapp_cxx)
> ```

jngrad:
Splitting the `${GPU_TARGET_NAME} = EspressoCore` into two targets with different compiler flags doesn't seem to work even with `PUBLIC`: properties of `EspressoCore` are not inherited by `EspressoCore_cxx` and `EspressoCore_cu`:
```cmake
  add_library(${GPU_TARGET_NAME}_cxx ${${GPU_TARGET_NAME}_sources_cxx})
  add_library(${GPU_TARGET_NAME}_cu  ${${GPU_TARGET_NAME}_sources_cu})
  target_link_libraries(${GPU_TARGET_NAME}_cu PUBLIC gpu_interface)
  add_library(${GPU_TARGET_NAME})
  set_target_properties(${GPU_TARGET_NAME} PROPERTIES LINKER_LANGUAGE "CXX")
  target_link_libraries(${GPU_TARGET_NAME} PUBLIC ${GPU_TARGET_NAME}_cxx ${GPU_TARGET_NAME}_cu)
  target_link_libraries(${GPU_TARGET_NAME} PRIVATE ${CUDA_LIBRARY} ${CUDART_LIBRARY})
```
```
You have called ADD_LIBRARY for library EspressoCore without any source files. This typically indicates a problem with your CMakeLists.txt file
[  1%] Building CXX object src/core/CMakeFiles/EspressoCore_cxx.dir/cells.cpp.o
In file included from /work/jgrad/espresso-fork-PR/src/core/cells.cpp:27:
In file included from /work/jgrad/espresso-fork-PR/src/core/cells.hpp:41:
In file included from /work/jgrad/espresso-fork-PR/src/core/CellStructure.hpp:25:
In file included from /work/jgrad/espresso-fork-PR/src/core/Cell.hpp:22:
/work/jgrad/espresso-fork-PR/src/core/Particle.hpp:22:10: fatal error: 'config.hpp' file not found
#include "config.hpp"
         ^~~~~~~~~~~~
[  3%] Building CXX object src/core/CMakeFiles/EspressoCore_cu.dir/actor/DipolarBarnesHut_cuda.cu.o
/work/jgrad/espresso-fork-PR/src/core/actor/DipolarBarnesHut_cuda.cu:24:10: fatal error: 'cuda_wrapper.hpp' file not found
#include "cuda_wrapper.hpp"
         ^~~~~~~~~~~~~~~~~~
```

jngrad:
Oh right, `EspressoConfig` is `PRIVATE`. That explains everything. In fact, we can decouple the CUDA code from the C++ code using `EspressoCore` for C++ and `EspressoCore_cu` for CUDA, and only expose the bare minimum of libraries and header files to `EspressoCore_cu`:
```cmake
target_link_libraries(EspressoCore_cu PRIVATE EspressoConfig shapes)
target_include_directories(EspressoCore_cu PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
target_link_libraries(EspressoCore PRIVATE EspressoCore_cu)
```
This actually compiles on Clang and passes the Python tests.

jngrad:
Let me amend the previous statement: we can't separate CUDA from C++ sources files using two targets. If `EspressoCore_cu` is:
- a shared library (58bdbc8d): there's a relink issue: can't install ESPResSo outside the build dir
  ```
  /usr/bin/python3: Relink `/lib/x86_64-linux-gnu/libsystemd.so.0' with `/lib/x86_64-linux-gnu/librt.so.1' for IFUNC symbol `clock_gettime'
  /usr/bin/python3: Relink `/lib/x86_64-linux-gnu/libudev.so.1' with `/lib/x86_64-linux-gnu/librt.so.1' for IFUNC symbol `clock_gettime'
  ```
- a static library (71937be1):
   - Clang fails if `-fPIC` isn't placed judiciously in the list of compiler flags for `EspressoCore`
     ```
     /usr/bin/ld: libEspressoCuda.a(DipolarBarnesHut_cuda.cu.o): relocation R_X86_64_32 against symbol `_Z20initializationKernelv' can not be used when making a shared object; recompile with -fPIC
     ```
   - in addition, there is a linking issue: can't install ESPResSo outside the build dir
     ```
     ImportError: libcudart.so.8.0: cannot open shared object file: No such file or directory
     ```
- an object library (bc89a3e73): can't link `EspressoConfig shapes` nor inherit them: compilation fails

The non-portable `$<COMPILE_LANGUAGE:language>` evaluates to something non-binary: both `$<COMPILE_LANGUAGE:CUDA>` and `$<NOT:$<COMPILE_LANGUAGE:CXX>>` evaluate to false on CUDA files.

I love CMake.

@KaiSzuttor Do you have any suggestion? I can't think of another way to introduce a GPU `INTERFACE` library that would only affect the .cu files and not the .cpp files in target `EspressoCore`.

jngrad:
@KaiSzuttor Found the solution: installing thrust 1.9.5 to fix the `va_printf` issue and doing `export LD_LIBRARY_PATH=/work/jgrad/cuda-10.0-thrust195/lib64:$LD_LIBRARY_PATH` to get shared libraries to find the correct CUDA libraries. I'll resume work on this PR.

CMakeLists.txt
```diff
@@ -402,18 +393,21 @@ if(WITH_ASAN AND WITH_MSAN)
...
-      "${CMAKE_CXX_FLAGS} -fsanitize=memory -fno-omit-frame-pointer")
+  list(APPEND cxx_interface_flags -fsanitize=memory -fno-omit-frame-pointer)
 endif()
 if(WITH_UBSAN)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=undefined")
+  list(APPEND cxx_interface_flags -fsanitize=undefined)
 endif()
 
+add_library(cxx_interface INTERFACE)
+target_compile_options(cxx_interface INTERFACE ${cxx_interface_flags})
```
fweik: out of currriosity, why do you keep the list in `cxx_interface_flags` instead of directly using `target_compile_options`?

jngrad: We actually need to pass these flags to both the compiler and linker, otherwise we get unresolved symbols (e.g. ASAN in [clang:6.0 logfile](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/223148)).

jngrad: In the line below, I pass the same list to [`target_link_libraries()`](https://cmake.org/cmake/help/v3.12/command/target_link_libraries.html?highlight=target_link#id2), which defines linker flags. CMake 3.13 introduced [`target_link_options()`](https://cmake.org/cmake/help/v3.13/command/target_link_options.html) to pass only linker flags, in an effort to make the intent clearer thanks to the name similarity with `target_compile_options()`.

fweik: Only the `-fsanitize=` no? (c.f.  https://gitlab.icp.uni-stuttgart.de/fweik/mdx/-/blob/master/CMakeLists.txt) Otherwise passing compiler options to the linker doesn't do anything, and is  just confusing. Also you still wouldn't need the list, right?

jngrad: Thanks for the hint! I wasn't sure if `-fsanitize=...` alone was sufficient, but it actually was.

jngrad:
The `osx-cuda` CI job is failing because of unresolved symbols in library `EspressoCuda`. These symbols are defined in `EspressoCore`. For example:

```c++
// lb_inertialess_tracers_cuda_interface.hpp
extern IBM_CUDA_ParticleDataInput *IBM_ParticleDataInput_host;
// lb_inertialess_tracers_cuda_interface.cpp
IBM_CUDA_ParticleDataInput *IBM_ParticleDataInput_host = nullptr;
// lb_inertialess_tracers_cuda.cu
    if (IBM_ParticleDataInput_host != NULL) {
```
Did I introduce a design flaw by creating a shared object `EspressoCuda.so` with the cuda code and `EspressoCore.so` for the C++ code? None of the other GPU build seems to mind. If it's not an issue, then we can ignore the CI failure: `osx-cuda` was removed in #3652.

The `intel:19` CI failure is just a regression in the image that wasn't caught in docker CI.

CMakeLists.txt
fweik: Yeah, some of the sanitizers have a runtime component, which needs to be linked...

fweik:
@jngrad IIRC there are cyclic dependencies between the CUDA and the core code...

jngrad:
@KaiSzuttor The PR is ready from my side. Let me now if you have additional changes to suggest. Otherwise, I'll rebase commits 58bdbc8dc6854 to 7025e65f to make the git history less chaotic, and resolve the merge conflict. The slowdown issues mentioned 2 weeks ago have been resolved and the `EspressoCuda` and `EspressoCore` are now a single target again to solve the cyclic dependency.

CMakeLists.txt
```diff
@@ -19,36 +19,47 @@
...
+if(POLICY CMP0025)
+  # make CXX_COMPILER_ID return "AppleClang" instead of "Clang" for Apple Clang
+  cmake_policy(SET CMP0025 NEW)
+endif()
 include(FeatureSummary)
 include(GNUInstallDirs)
 project(ESPResSo)
 include(cmake/FindPythonModule.cmake)
-if(CMAKE_VERSION VERSION_GREATER_EQUAL "3.12")
+include(cmake/option_enum.cmake)
```
KaiSzuttor: if you move line 54 further up, you can omit specifying the whole path. Also you can omit the ```.cmake``` part.

CMakeLists.txt
```diff
@@ -130,6 +141,43 @@ endforeach(func_name __PRETTY_FUNCTION__ __FUNCTION__)
...
+    list(APPEND CMAKE_MODULE_PATH "${ROCM_HOME}/hip/cmake")
+    find_package(HIP 1.5.18494 MODULE REQUIRED)
+    # patch HCC_PATH environment variable and reload HIP
+    if(HIP_VERSION VERSION_LESS "3.1")
+      set(HCC_PATH "${HIP_ROOT_DIR}")
+    else()
+      set(HCC_PATH "${ROCM_HOME}/hcc")
+    endif()
+    find_package(HIP 1.5.18494 MODULE REQUIRED)
+    find_package(CUDACompilerHIP 6.0 REQUIRED)
```
KaiSzuttor: can't we hide this in a module?

CMakeLists.txt
```diff
@@ -130,6 +141,43 @@ endforeach(func_name __PRETTY_FUNCTION__ __FUNCTION__)
...
+  elseif(WITH_CUDA_COMPILER STREQUAL "clang")
+    if(WITH_COVERAGE)
+      message(
+        FATAL_ERROR
+          "The combination of options '-DWITH_CUDA_COMPILER=${WITH_CUDA_COMPILER} -DWITH_COVERAGE=${WITH_COVERAGE}' is not supported."
```
KaiSzuttor: I think having the name for the option hardcoded in the error string is not maintainable

CMakeLists.txt
```diff
@@ -130,6 +141,43 @@ endforeach(func_name __PRETTY_FUNCTION__ __FUNCTION__)
...
+      )
+    endif()
+    set(MINIMAL_CUDA_VERSION 8.0)
+    find_package(CUDACompilerClang 6.0 REQUIRED)
+  else()
+    message(
+      FATAL_ERROR
+        "Cannot process option -DWITH_CUDA_COMPILER='${WITH_CUDA_COMPILER}'")
```
KaiSzuttor: see above

CMakeLists.txt
```diff
@@ -283,70 +331,56 @@ set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/${PYTHON_INSTDIR}/espressomd")
...
+    -Wextra
+    $<$<BOOL:${WARNINGS_ARE_ERRORS}>:-Werror>
+    # disable warnings from -Wextra
+    -Wno-sign-compare
+    -Wno-unused-function
+    -Wno-unused-variable
+    -Wno-unused-parameter
+    -Wno-missing-braces
+    $<$<AND:$<CXX_COMPILER_ID:GNU>,$<STREQUAL:${CMAKE_SYSTEM_PROCESSOR},arm>>:-Wno-psabi>
+    $<$<AND:$<NOT:$<CXX_COMPILER_ID:Intel>>,$<NOT:$<AND:$<CXX_COMPILER_ID:GNU>,$<VERSION_LESS:$<CXX_COMPILER_VERSION>,7.0.0>>>>:-Wno-implicit-fallthrough>
```
KaiSzuttor: hm... this is really hard to read.

KaiSzuttor (dismissed):
Looks good to me besides the comments. Thanks @jngrad that looks much cleaner now.

CMakeLists.txt
jngrad: just these two, or the following ones too?

CMakeLists.txt
jngrad: yes, I'll move it to `cmake/FindCUDACompilerHIP.cmake`

mkuron: Where does that `6.0` version number come from? Is that the Clang version of hipcc? I don't think it makes sense to require a minimum version there because the HIP version already implies a certain minimum Clang version.

jngrad: It's the Clang version. I put 6.0 here since this is the oldest Clang version we're testing. Do you know where I could find a table showing the HIP <=> Clang version mapping?

CMakeLists.txt
jngrad: I'm reverting the ones with more than 1 AND/OR, here and in other CMake files.

CMakeLists.txt
jngrad: fixed

CMakeLists.txt
jngrad: I've removed that Clang check in be61c3a1e.

fweik (approved):

PR #3584, opened 2020-03-16
virtual sites: Missing import

christophlohrmann:
Description of changes:
- added missing import in ``VirtualSitesRelative.cpp``


KaiSzuttor:
I think the includes are not totally correct in a lot of places. Does this missing import lead to a failing build in some circumstance?

christophlohrmann:
Yes. For my 'leaner' swimmer-config

KaiSzuttor (approved):

jngrad:
Once [iwyu](https://include-what-you-use.org/) gets a stable release for Clang 10, we can have give it a try. The stable release for Clang 6 was not working great on espresso code due to some boost headers being incorrectly flagged as circular imports.

PR #3586, opened 2020-03-18
Refactor Analysis module: distance functions

jngrad:
Description of changes:
- remove `min_dist2()`
- move logic from `dist_to()` into `system.distance_vec()`
- remove unused code
- fix a regression in `PidProfileObservable` introduced by #3560
- fix a regression in the OpenGL visualizer introduced by #3575


jngrad:
@RudolfWeeber I'm splitting the work in multiple PRs. I'm converting the analysis functions `rdf`, `linear_momentum` and `angular_momentum` as observables in another branch. It's gonna be a substantial extension of the Observable framework. The current framework doesn't scale well, e.g. the `PidProfileObservable` assumes a 3D profile (to get a 1D profile, one has to duplicate the 3D profile class), and has a lot of code duplication with `CylindricalProfile` for no good reason (from what I understand, it's only to get class attributes named `min_r`/`min_phi` instead of `min_x`/`min_y`). A solution of the style `template <size_t dimension=3> class ProfileObservable` leads to compiler errors because the templated classes have the same symbol. The long-term solution here is probably to have a templated class with structs as template argument, e.g. `Cartesian1D`, `Cartesian3D`, `Cylindrical3D`, `Radial1D`, etc. As an illustration, the RDF requires the creation of either a `PtypeProfileObservable1d -> {PtypeObservable, ProfileObservable1d}` class or a `RadialProfile` class.

src/python/espressomd/system.pxd
```diff
@@ -34,3 +36,13 @@ cdef bool skin_set
 cdef extern from "particle_data.hpp":
     int init_type_map(int type) except +
     int number_of_particles_with_type(int type) except +
+
+cdef extern from "PartCfg.hpp":
```
KaiSzuttor: why did you put the particle stuff here?

src/python/espressomd/system.pyx
```diff
@@ -359,6 +360,63 @@ cdef class System:
 
         return make_array_locked(mi_vec)
 
+    def dist_to(self, id=None, pos=None):
```
KaiSzuttor: the name of this function is very inexpressive. Should this function actually exist? If you have a function that gives you the distance between two positions, say ```dist(pos1, pos2)```, you can just to ```min_dist = np.min([dist(pos1, p.pos) for p in system.part])```. So if it's a not often used one-liner in python, we probably don't need a function for it?

KaiSzuttor (commented):
Maybe we can get rid of ```distto```

KaiSzuttor:
Concerning the histogram observables: this is probably a bit more involved because currently it also has the issue that histogram information cannot be retrieved. Also it may be worth to first refactor the observable framework and afterwards think about interaction with the histograms?

src/python/espressomd/system.pxd
jngrad: `dist_to` takes a `partCfg()` as first argument, so we have to declare it in the .pxd file

KaiSzuttor: but why does it belong to ```system```?

jngrad: it doesn't, it's only a side-effect of the underlying core function; we could remove that partCfg dependency by writing an overloaded `dist_to` function in the core

src/python/espressomd/system.pyx
jngrad: in my notes of the meeting, I wrote "dist_to? should be with system.distance", but maybe I didn't catch what you said about removing it

src/python/espressomd/system.pxd
KaiSzuttor: as mentioned below, I think we don't need the function at all... I think we should just have a free function that does the minimum image convention for us taking a position and the box as an input a.k.a. ```get_mi_vector```

src/python/espressomd/system.pyx
KaiSzuttor: yeah sorry, it just got clear to me looking at the code

jngrad:
I haven't looked into histograms, not sure how much of if it will overlap with my work on the `ProfileObservable` classes. I'm not really convinced by the current framework, for example `ProfileObservable` inherits from `Observable` but `CylindricalProfile` doesn't. It's unclear whether `SphericalProfile` should derive from `Observable` or not. Also, the way `CylindricalProfile` is currently used, it would simplify the code to have it inherit from `Observable`, but I'm not sure it's the right direction. Having a template class `ProfileObservable<coord_system>` inheriting from `Cartesian`/`Cylindrical`/`Spherical` would be less confusing.

src/python/espressomd/system.pyx
jngrad: would  you like me to change the `system.distance_vec()` to take either particles or 3d vectors, so that we can use your one-liner?

RudolfWeeber:
```

>@jngrad commented on this pull request.

>

  _____  

>In src/python/espressomd/system.pyx <https://github.com/espressomd/espresso/pull/3586#discussion_r395094258> :

>> @@ -359,6 +360,63 @@ cdef class System:
> 
>         return make_array_locked(mi_vec)
> 
>+    def dist_to(self, id=None, pos=None):

>would you like me to change the system.distance_vec() to take either particles or 3d vectors, so that we can use your one-liner?

Yes. There should be only system.distance and system.distance_vec. Both should take either a 3-array or a ParticleHandle as either of their arguments.

```

src/python/espressomd/system.pyx
```diff
@@ -345,77 +345,50 @@ cdef class System:
         return self.box_l[0] * self.box_l[1] * self.box_l[2]
 
     def distance(self, p1, p2):
-        """Return the scalar distance between the particles, respecting periodic boundaries.
+        """Return the scalar distance between particles or between a particle
+        and a point, respecting periodic boundaries.
```
KaiSzuttor: or between two points

src/python/espressomd/system.pyx
```diff
@@ -345,77 +345,50 @@ cdef class System:
...
+        Parameters
+        ----------
+        p1 : :class:`~espressomd.particle_data.ParticleHandle` or (3,) array of :obj:`float`
+            First particle or position.
+        p2 : :class:`~espressomd.particle_data.ParticleHandle` or (3,) array of :obj:`float`
+            Second particle or position.
 
         """
         res = self.distance_vec(p1, p2)
         return np.sqrt(res[0]**2 + res[1]**2 + res[2]**2)
```
KaiSzuttor: ```np.linalg.norm```

testsuite/python/analyze_distance.py
```diff
@@ -95,22 +95,35 @@ def test_nbhood(self):
...
+            self.assertAlmostEqual(
+                np.min([self.system.distance([i, i, i], p.pos)
+                        for p in parts]),
+                self.dist_to_pos([i, i, i]), delta=1e-10)
+            self.assertEqual(
+                np.min([self.system.distance([i, i, i], p.pos)
+                        for p in parts]),
+                np.min([self.system.distance(p.pos, [i, i, i]) for p in parts]))
+
+    def test_distance_to_id(self):
```
KaiSzuttor: you probably mean ```test_distance_to_part```

RudolfWeeber (approved):

PR #3588, opened 2020-03-20
Update CI images

jngrad:
Description of changes:
- replace `intel:18` with `intel:19`
- remove `s390x`

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -357,11 +357,6 @@ ubuntu:i386:
 ubuntu:ppc64le:
   <<: *arch_definition
 
-ubuntu:s390x:
-  <<: *arch_definition
-  variables:
-    LD_BIND_NOW: none
-
 ### Builds with OS X
 
 osx:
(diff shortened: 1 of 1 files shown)


mkuron:
Do we have ppc64 (not ppc64le) testing on Fedora? The reason we tested on s390x was that it's Ubuntu's only big-endian platform and @junghans's builds always discovered endianness bugs in every release.

mkuron:
While you're at it, you can also drop the 32-bit architectures. Ubuntu 20.04 is only weeks away from being released, and that was how long we wanted to support them.

junghans:
Fedora is still building 32-bit packages, but I can always exclude these architectures.

RudolfWeeber:
>Fedora is still building 32-bit packages, but I can always exclude these architectures.

IMO, they can stay active, as long as they work, but no more time should be invested, once they break. Since we’re taking 32bit out of our CI, this will happen sooner or later.

junghans:
Plus you guys get some testing through CoPR as well.

junghans (approved):

fweik (approved):

PR #3593, opened 2020-03-25
Fix LBGPU velocity interpolation close to boundary

RudolfWeeber:
Fixes #

Description of changes:
LB GPu velocity interpolation used incorrect velocities at boundary nodes. This lead to incorrect resutls for interpolations closer than one lattice constant to the boundary.
This adds a missing unit conversion 1/lattice_speed for boundary nodes, which fixes the issue.

Tests for interpolation at the boundary position and at the mid-point between the boundary and the next lattice site.

This should go into 4.1.3


testsuite/python/lb_interpolation.py
```diff
@@ -23,11 +23,11 @@
 import espressomd.shapes
 import espressomd.lb
 
-AGRID = 1.5
+AGRID = 1.5
```
jngrad: there is an extra whitespace
```suggestion
AGRID = 1.5
```

testsuite/python/lb_interpolation.py
```diff
@@ -72,9 +72,21 @@ def test_interpolated_velocity(self):
...
         self.system.integrator.run(250)
-        # Shear plane for boundary 1
-        # for pos in itertools.product((AGRID,), np.arange(0.5 * AGRID, BOX_L, AGRID), np.arange(0.5 * AGRID, BOX_L, AGRID)):
-        #     np.testing.assert_almost_equal(self.lbf.get_interpolated_velocity(pos)[2], 0.0)
+        # Check interpolated vel at upper boundary. The node position is at 
+        # box_l[0]-agrid/2.
+        np.testing.assert_allclose(self.lbf.get_interpolated_velocity(
+            [self.system.box_l[0] - AGRID / 2, 0, 0]), np.array([0, 0, V_BOUNDARY]))
+
+        # Cehck interpolated velocity involving boundary and neighboring node
```
jngrad: typo
```suggestion
        # Check interpolated velocity involving boundary and neighboring node
```

testsuite/python/lb_interpolation.py
```diff
@@ -72,9 +72,21 @@ def test_interpolated_velocity(self):
         """
         self.set_boundaries([0.0, 0.0, V_BOUNDARY])
         self.system.integrator.run(250)
-        # Shear plane for boundary 1
-        # for pos in itertools.product((AGRID,), np.arange(0.5 * AGRID, BOX_L, AGRID), np.arange(0.5 * AGRID, BOX_L, AGRID)):
-        #     np.testing.assert_almost_equal(self.lbf.get_interpolated_velocity(pos)[2], 0.0)
+        # Check interpolated vel at upper boundary. The node position is at 
+        # box_l[0]-agrid/2.
+        np.testing.assert_allclose(self.lbf.get_interpolated_velocity(
```
jngrad: do we need to guard the result of `get_interpolated_velocity` with `np.copy()`?

jngrad (changes requested):

testsuite/python/lb_interpolation.py
RudolfWeeber: We let the CI decide.

jngrad:
Did you figure out the source of the error `Lattice velocity exceeds the Mach number limit in function void lb_boundary_mach_check()` in failed CI jobs?

RudolfWeeber:
I don't understand the larger error for the clang:6.0 and cuda 9 images. On my machine, (and apparently on the cuda 10 container), the accuracy is much better. Atol=1E-10  worked locally.

I'm increasing the tolerance for now, because the result is still much better than the one with the missing prefactor without this PR. But I don't understand why that should be needed.

RudolfWeeber:
Accuracy reduction was caused by an unintended parameter change (TAU) which I reverted now. still don't understand the platform-dependence, now.

jngrad:
Seems to be working now. Could you please revert the `print/sys.stdout` changes and squash the PR down to a single commit? This will make backporting go smoother.

RudolfWeeber:
I added the print stuff on purpose, just like in the LB test. Since the error message output from the Mach limit test cannot be supressed, the output of the test is confusing without the new print.

 

W.r. to squashing I already eliminated the merge commit. couldn’t you cherry-pick the commit range with a current git. The frequent context switches for revisiting PRs impedes my science work too much, currently.

jngrad (approved):

KaiSzuttor:
is the boundary velocity stored in MD units? The correct fix would be to store it in LB units and not adding this unit conversion in the LB code. This should be reverted.

RudolfWeeber:
>is the boundary velocity stored in MD units? The correct fix would be to store it in LB units and not adding this unit conversion in the LB code. This should be reverted.

Apparantly it is in LBGPU. I find the boundary code to be rather ‘write-only’ and the LBGPU is to be replaced anyway, so I did not make an attempt at cleaning things up.

 

For Walberla, we use the same units everywhere, obviously.

KaiSzuttor:
all right

Issue #3594, opened 2020-03-25
Performance regression when creating particles


RudolfWeeber:
The following script changed from ~0.8 to 50 secs with recent changes in the Python branch
```
import espressomd
import numpy as np

s=espressomd.System(box_l=[10,10,10])
s.part.add(pos=np.zeros((100000,3)))
```

It is my impression that this is caused in/around the particle node lookup, particularly, when the particle node cache is rebuilt.

In my actual simulations, I need 1^6 particles, so the particle creation would take nearly ten minutes...
It gets worse, when bonds are involved, because then, even more particle node lookups happen.

@fweik, could you take a look, please?
I can probably work around it temporarily as long as my simulations are on a single node, but probably this will come up in more situations.

RudolfWeeber:
With full optimization, perf reports
60% in cell_structure::add_local_particle and 30% in get_maximal_particle_id.

fweik:
From a quick glance the reason is that the python code gets the maximal id after every added particle,
which causes a lot of communication. This is actually not needed, for a slice it only needs to be calculated once, and then all the subsequent ids can be filled in.

RudolfWeeber:
Profile with -Og: 
```
# Event count (approx.): 881886834380
#
# Overhead  Command  Shared Object                                      Symbol                                                                                                                                                                                                                                                                                                                                                                                                                               
# ........  .......  .................................................  .....................................................................................................................................................................................................................................................................................................................................................................................................................................
#
    17.06%  python3  EspressoCore.so                                    [.] CellStructure::update_particle_index
     7.59%  python3  _init.so                                           [.] std::__detail::_Hash_node_value_base<std::pair<int const, int> >::_M_v
     7.50%  python3  EspressoCore.so                                    [.] CellStructure::update_particle_index
     5.75%  python3  EspressoCore.so                                    [.] get_maximal_particle_id()::$_6::operator()
     5.13%  python3  EspressoCore.so                                    [.] std::pair<int, int>::pair<int const, int, true>
     4.98%  python3  _init.so                                           [.] std::max<int>
     4.40%  python3  _init.so                                           [.] __gnu_cxx::__aligned_buffer<std::pair<int const, int> >::_M_addr
     4.22%  python3  ScriptInterface.so                                 [.] std::__detail::_Node_iterator_base<std::pair<int const, int>, false>::_M_incr
[...]

PR #3595, opened 2020-03-25
Improve performance of particle updates

fweik:
Fixes maybe #3594.

Description of changes:
- Remove an expensive check from `get_particle_node`
- Reduce number of particle index update when adding particles


RudolfWeeber (approved):
Works. Thanks. I'll open the PR for the more efficient id generation in a minute.

PR #3598, opened 2020-03-26
Py: Generate ids for bulk particle creation more efficiently

RudolfWeeber:
Description of changes:
When adding many particles without specifying the ids
```
system.part.add(pos=...)
```
generate all new particle ids at once. This avoids n-1 calls to get_maximal_particle_id() which gets expensive for big systems (100k particles).

### DIFF
--- a/src/python/espressomd/particle_data.pyx
+++ b/src/python/espressomd/particle_data.pyx
@@ -154,7 +154,7 @@ cdef class ParticleHandle:
 
         def __get__(self):
             self.update_particle_data()
-            return make_array_locked(unfolded_position( < Vector3d > self.particle_data.r.p, < Vector3i > self.particle_data.l.i, box_geo.length()))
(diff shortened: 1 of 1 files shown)


src/python/espressomd/particle_data.pyx
```diff
@@ -1862,15 +1862,21 @@ Set quat and scalar dipole moment (dipm) instead.")
             raise ValueError(
                 "When adding several particles at once, all lists of attributes have to have the same size")
 
-        # Place new particles and collect ids
-        ids = []
+        # If particle ids haven't been provided, use free ones
+        # beyond the highest existing one
+        if not "id" in Ps:
+            first_id = get_maximal_particle_id() + 1
+            Ps["id"] = [i for i in range(first_id, first_id + n_parts)]
```
fweik: The list comprehension is needed because it needs to be a list?

RudolfWeeber:
I assumed a list was needed, but actually that's not so. Everything down stream seams to be able to cope with a range object\

fweik:
You can actually pass a `range` from the interface...

fweik (approved):

PR #3601, opened 2020-03-28
Back PartCfg by the particle fetch cache

### LINKED ISSUE #3600: Performance regression when evaluating PidObservable
Evaluating ParticlePositions for 40k particles takes 40 sec with a current ES, but <0.1s with an older version.
The profile suggests that most of the time is spent in building the particle cache.

Test script:
```
import espressomd
import espressomd.observables

import numpy as np
from time import time

n=40000
s=espressomd.System(box_l=[10,10,10])
s.part.add(id=range(n),pos=np.random.random((n,3))*s.box_l)
obs=espressomd.observables.ParticlePositions(ids=s.part[:].id)
tick=time()
x=obs.calculate()
tock=time()
print(tock-tick)
```

fweik:
Fixes #3600.

Description of changes:
- Remove custom particle gathering code in PartCfg
- Fetch particles for PartCfg via the fetch cache
- Switch PidObservable from PartCfg to the fetch cache
- Switch cluster analysis from PartCfg to the fetch cache
- Switch statistics_chain from PartCfg to the fetch cache

This is much cleaner and way faster than the old stuff.


RudolfWeeber (approved):
Thanks a lot! With this quick solution, I don't have to manually cherry-pick together an Espresso version for my simulations.

Issue #3421, opened 2020-01-19
NEP 29: new release schedule for scientific python packages


jngrad:
The [NEP 29 - Recommend Python and Numpy version support as a community policy standard](https://numpy.org/neps/nep-0029-deprecation_policy.html) proposes a new release schedule for `numpy`. Major python packages like `matplotlib`, `jupyter` and `scipy` have already adopted NEP 29, which means that most python dependencies of espresso will be affected either directly or indirectly. The proposed release cycle is 42 months, which is slightly shorter than the 48 months time window for espresso (we support the last two Ubuntu LTS).

This has consequences on our own release schedule. There is a handy [support table](https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table) and [drop schedule](https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule) to keep track of when to update our [requirements.txt](https://github.com/espressomd/espresso/blob/python/requirements.txt). We'll have to be more strict about version pinning for the few python packages that cannot be installed from Ubuntu repositories, to avoid e.g. #3420. We might have to drop support for Ubuntu 16.04 LTS in the planned espresso 4.1.3 release (due several months after Ubuntu 20.04 LTS is out), and tweak our docker [.gitlab-ci.yml](https://github.com/espressomd/docker/blob/master/.gitlab-ci.yml) to test Ubuntu 16.04 images against the 4.1.2 tag instead of the 4.1 branch.

RudolfWeeber:
>The  <https://numpy.org/neps/nep-0029-deprecation_policy.html> NEP 29 - Recommend Python and Numpy version support as a community policy standard proposes a new release schedule for numpy. Major python >packages like matplotlib, jupyter and scipy have already adopted NEP 29, which means that most python dependencies of espresso will be affected >either directly or indirectly. The proposed release cycle is 42 months, which is slightly shorter than the 48 months time window for espresso (we support the >last two Ubuntu LTS).

I think we don’t support 16.04 for any support schedule reasons. We have it because we are stuck with it for ICP internal reasons.

 

[...]

PR #3606, opened 2020-03-30
Adding pre-commit tool config plus cmake-format

### LINKED ISSUE #3591: Add and enforce CMake formatting rules
We should have a .cmake-format file, and enforce it in the CI. We might also consider checking
if there are source files that are not mentioned in the build system and warn about that (not sure about this one).

KaiSzuttor:
Fixes #3591

Depends on https://github.com/espressomd/docker/pull/158

Description of changes:
- pulled out tool version checking to separate files
- added configuration file for [pre-commit](https://pre-commit.com)
- added cmake-format configuration and style fix


maintainer/format/autopep8.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
```
jngrad, line 19: here and in the other new .sh files: should we copy-paste the license header from `maintainer/CI/fix_style.sh`?

requirements.txt
```diff
@@ -28,3 +28,4 @@ pylint>=2.2.2
 astroid>=2.1.0
 isort>=4.3.4
 setuptools>=20.7.0
+pre-commit>=2.2.0
```
jngrad, line 31: should we pin `cmake-format==0.6.9` just like we pinned `autopep8` and `pycodestyle`?

maintainer/format/autopep8.sh
KaiSzuttor: yes probably

maintainer/format/autopep8.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
```
jngrad: Do we really need Bash? The original file was converted pure Dash ([`984b061`](https://github.com/espressomd/espresso/commit/984b061645a80f5e184eabf077a43ffddb0c26b8#diff-78719c369a7782258dc4e7743964d748)) to improve interoperability and prevent the re-introduction of bashisms.

doc/tutorials/01-lennard_jones/CMakeLists.txt
```diff
@@ -1,8 +1,12 @@
...
-nb_export(TARGET tutorial_01 SUFFIX "" FILE "01-lennard_jones.ipynb" HTML_RUN VAR_SUBST)
+nb_export(
+  TARGET
+  tutorial_01
+  SUFFIX
+  ""
+  FILE
+  "01-lennard_jones.ipynb"
+  HTML_RUN
+  VAR_SUBST)
```
jngrad: The majority of the changes looks very good, but in tutorials, the changes look inconsistent: the `configure_tutorial_target` command is nicely condensed while the `nb_export` command is split over too many lines; can't we have it condensed over two lines only? This would be below the 80 chars limit:
```cmake
nb_export(TARGET tutorial_04 SUFFIX "1"
          FILE "04-lattice_boltzmann_part1.ipynb" HTML_RUN)
```

src/core/unit_tests/CMakeLists.txt
```diff
@@ -4,45 +4,93 @@
 
 # This file is part of ESPResSo.
 
-# ESPResSo is free software: you can redistribute it and/or modify
-# it under the terms of the GNU General Public License as published by
-# the Free Software Foundation, either version 3 of the License, or
-# (at your option) any later version.
+# ESPResSo is free software: you can redistribute it and/or modify it under the
```
jngrad: This formatting should not happen; please revert this change and add the missing `#` symbol in empty lines of the license header.

src/core/unit_tests/CMakeLists.txt
```diff
@@ -4,45 +4,93 @@
...
-unit_test(NAME AutoParameters_test SRC AutoParameters_test.cpp DEPENDS ScriptInterface)
-unit_test(NAME AutoParameter_test SRC AutoParameter_test.cpp DEPENDS ScriptInterface)
+unit_test(NAME ScriptInterface_test SRC ScriptInterface_test.cpp DEPENDS
+          ScriptInterface)
+unit_test(
+  NAME
+  MpiCallbacks_test
+  SRC
+  MpiCallbacks_test.cpp
+  DEPENDS
```
jngrad: One word by line is overzealous.

src/shapes/unit_tests/CMakeLists.txt
```diff
@@ -1,5 +1,33 @@
 include(unit_test)
-unit_test(NAME Wall_test SRC Wall_test.cpp DEPENDS shapes utils)
-unit_test(NAME HollowConicalFrustum_test SRC HollowConicalFrustum_test.cpp DEPENDS shapes utils)
-unit_test(NAME Union_test SRC Union_test.cpp DEPENDS shapes utils)
-unit_test(NAME Ellipsoid_test SRC Ellipsoid_test.cpp DEPENDS shapes utils)
+unit_test(
+  NAME
+  Wall_test
+  SRC
```
jngrad: overzealous

src/utils/tests/CMakeLists.txt
```diff
@@ -3,9 +3,17 @@ include(unit_test)
...
 unit_test(NAME Vector_test SRC Vector_test.cpp DEPENDS utils)
 unit_test(NAME Factory_test SRC Factory_test.cpp DEPENDS utils)
-unit_test(NAME NumeratedContainer_test SRC NumeratedContainer_test.cpp DEPENDS utils)
+unit_test(NAME NumeratedContainer_test SRC NumeratedContainer_test.cpp DEPENDS
+          utils)
 unit_test(NAME make_function_test SRC make_function_test.cpp DEPENDS utils)
-unit_test(NAME List_test SRC List_test.cpp DEPENDS Boost::serialization utils)
+unit_test(
+  NAME
+  List_test
```
jngrad: overzealous

testsuite/python/CMakeLists.txt
```diff
@@ -130,7 +165,15 @@ python_test(FILE analyze_energy.py MAX_NUM_PROC 2)
 python_test(FILE analyze_mass_related.py MAX_NUM_PROC 4)
 python_test(FILE rdf.py MAX_NUM_PROC 1)
 python_test(FILE coulomb_mixed_periodicity.py MAX_NUM_PROC 4 LABELS long)
-python_test(FILE coulomb_cloud_wall_duplicated.py MAX_NUM_PROC 4 LABELS gpu LABELS long)
+python_test(
+  FILE
+  coulomb_cloud_wall_duplicated.py
+  MAX_NUM_PROC
```
jngrad: overzealous

jngrad (changes requested):

doc/tutorials/01-lennard_jones/CMakeLists.txt
jngrad: all the other tutorials are still affected

KaiSzuttor: I think you looked at outdated files.

doc/tutorials/04-lattice_boltzmann/CMakeLists.txt
```diff
@@ -1,24 +1,37 @@
...
+          HTML_RUN)
+nb_export(TARGET tutorial_04 SUFFIX "2" FILE "04-lattice_boltzmann_part2.ipynb"
+          HTML_RUN)
+nb_export(TARGET tutorial_04 SUFFIX "3" FILE "04-lattice_boltzmann_part3.ipynb"
+          HTML_RUN)
+nb_export(
+  TARGET
+  tutorial_04
+  SUFFIX
+  "4"
```
jngrad, line 31: overzealous

doc/tutorials/05-raspberry_electrophoresis/CMakeLists.txt
```diff
@@ -1,10 +1,15 @@
...
-  TARGET tutorial_05
-  DEPENDS
+  TARGET tutorial_05 DEPENDS 05-raspberry_electrophoresis.ipynb
+  figures/raspberry_snapshot.png posVsTime_sample.dat)
+
+nb_export(
+  TARGET
+  tutorial_05
+  SUFFIX
+  ""
```
jngrad, line 9: overzealous

jngrad (changes requested):

doc/tutorials/04-lattice_boltzmann/CMakeLists.txt
KaiSzuttor: that is a matter of taste. If you have a long list it might be more readable to have a column.
Im not sure if there is a global sweet-spot and we have to live with something like that here.
If I further increase the value for the maximum elements you will get more of unreadable 
horizontal lists.

fweik:
Just my two cents: I'm not sure that adding `pre-commit` is worth it over just calling the shell scripts. It's just one more thing to know. @jngrad if you decide to keep it, make sure that @KaiSzuttor properly documents it in the CI documentation.

jngrad:
@fweik once this is merged, we won't be able to call the shell scripts anymore, because the `find` commands that detect the file extensions were replaced by `pre-commit`. Everyone will have to install `pre-commit` to run the style scripts. I don't really mind, because I don't use `fix_style.sh` anyway; I wrote my own implementation of `pre-commit` almost two years ago (`git-style`) and will keep using it because it has more features.

fweik:
@jngrad yes, the alternative is to keep the find expressions.

KaiSzuttor:
Btw. you already have to pip install the autopep8 package, the pycodestyle package and the cmake-format package due to the pinned versions. You will most probably do that via ```pip install -r requirements.txt```. If so, you will automagically not notice the change at all.

jngrad (approved):
ok from my side

KaiSzuttor:
interesting
``` bash
24   File "/home/espresso/.local/bin/pre-commit", line 7, in <module>
25     from pre_commit.main import main
26   File "/home/espresso/.local/lib/python3.5/site-packages/pre_commit/main.py", line 189
27     version=f'%(prog)s {C.VERSION}',
28                                   ^
29 SyntaxError: invalid syntax
30 Passed style check
```

jngrad:
@KaiSzuttor We shouldn't delay #3421 any longer and start moving towards Python 3.6 every time our workflow breaks. espressomd/docker#160 moves the style check packages to the Ubuntu 20.04 image, it'll only be a matter of updating the yml file here to use Ubuntu 20.04 image instead of Ubuntu 16.04.

KaiSzuttor:
but, do you understand where the error comes from?

maintainer/CI/fix_style.sh
```diff
@@ -18,36 +18,19 @@
...
-    echo "${AUTOPEP8} is $(${AUTOPEP8} --version 2>&1)"
+if ! hash pre-commit 2>/dev/null; then
+    echo "pre-commit command not found."
     exit 2
 fi
 
-find . \( -name '*.hpp' -o -name '*.cpp' -o -name '*.cu' -o -name '*.cuh' \) -not -path './libs/*' | xargs -r -n 5 -P 8 "${CLANGFORMAT}" -i -style=file || exit 3
-find . \( -name '*.py' -o -name '*.pyx' -o -name '*.pxd' \) -not -path './libs/*' | xargs -r -n 5 -P 8 "${AUTOPEP8}" --ignore=E266,W291,W293 --in-place --aggressive || exit 3
-find . -type f -executable ! -name '*.sh' ! -name '*.py' ! -name '*.sh.in' ! -name pypresso.cmakein -not -path './.git/*' | xargs -r -n 5 -P 8 chmod -x || exit 3
+pre-commit run --all-files
```
jngrad, line 33: I think dash and bash return the error code of the last command, so if this line fails due to a Python3.6-specific f-string (Ubuntu 16.04 uses Python3.5), the shell script will keep going to the next instructions and return 0, unless we add the following:
```suggestion
pre-commit run --all-files || exit 1
```

jngrad: Actually no, don't do that! `pre-commit` returns 1 when it touched a file.

jngrad: Is there an older version of `pre-commit` compatible with Python 3.5 that we can use?

KaiSzuttor:
I just realised that the issue is also that the packaged pip silently installs packages that may not be compatible with the python version. If you upgrade pip and try to install pre-commit in the container, you actually get:
``` bash
ERROR: Package 'pre-commit' requires a different Python: 3.5.2 not in '>=3.6.1'
```
Installing an older version would not help because a dependency of pre-commit needs 3.6.

mkuron:
> > Is there an older version of pre-commit compatible with Python 3.5 that we can use?

> Installing an older version would not help because a dependency of pre-commit needs 3.6.

You could explicitly install specific (python 3.5-compatible) versions of the dependencies to circumvent that problem. Of course, that means adding all dependencies to the Dockerfile. pip is a terrible package manager.

RudolfWeeber:
I wonder whether this PR takes too much work from too many people.

 

Personally, I just run JN’s script before committing and that’s that.

 

Do we really want to follow through with this?

 

Or is my impression wrong, that this hasn’t converged to the point of just needing a review?

KaiSzuttor:
> Or is my impression wrong, that this hasn’t converged to the point of just needing a review?

my impression is that sometimes we comment on PRs without having looked at the changes

jngrad:
> I wonder whether this PR takes too much work from too many people.

There wouldn't be so much friction if ICP machines had Clang-9 installed.

> Do we really want to follow through with this?

The `pre-commit` tool removes some complexity from the shell scripts by taking care of the path detection. It's unclear to me how many people will use the commit hook feature. I most likely won't, because I rely a lot on rebase operations and don't want them to be interrupted, nor do I want to work on two clones, one with the hook and one without.

It's also unclear to me why `pre-commit` works so fast. On my previous ICP machine that had twice as less RAM the `fix_style.sh` script would use several GB of RAM and freeze the OS for several minutes (Rudolf and Christoph also experience performance issues). I'd prefer to have `pre-commit` in ESPResSo rather than my tool in ESPResSo, because my tool is untested and quite complex.

KaiSzuttor:
the issues came solely from the default pip version, I fixed it in https://github.com/espressomd/docker/pull/162

KaiSzuttor:
closed for #3622

PR #3607, opened 2020-03-30
Core: Fix LB profile observable

### LINKED ISSUE #3603: LB Profile observables: incorrect interpolation positoins
Can be tested by placing particles at the nodes of the lb, and then comparing
the particle velocities (and FluxDensityProfile) against LbVelocityProfile:

```
import numpy as np
import sys

import espressomd
from espressomd.lb import LBFluidGPU
import espressomd.observables

size_in_agrid = [12,14,17]
agrid = 0.7

box_size = agrid * np.array(size_in_agrid)
dt=0.01


system = espressomd.System(box_l=box_size)
lbf = LBFluidGPU(dens=1,agrid=agrid, tau=dt,visc=1,kT=0)
system.actors.add(lbf)

expected=np.zeros(size_in_agrid+[3])

for i in range(size_in_agrid[0]):
  for j in range(size_in_agrid[1]):
    for k in range(size_in_agrid[2]):
      p=system.part.add(pos=agrid *(np.array((i,j,k))+.5),v=2*np.array((i,j,k)))
      lbf[i,j,k].velocity=p.v
      expected[i,j,k,:]=p.v


lb_vel_profile=espressomd.observables.LBVelocityProfile(
        n_x_bins=size_in_agrid[0],
        n_y_bins=size_in_agrid[1],
        n_z_bins=size_in_agrid[2],
        min_x=0,
        min_y=0,
        min_z=0,
        max_x=system.box_l[0],
        max_y=system.box_l[1],
        max_z=system.box_l[2],
        sampling_offset_x=agrid/2,
        sampling_offset_y=agrid/2,
        sampling_offset_z=agrid/2,
        sampling_delta_x=agrid,
        sampling_delta_y=agrid,
        sampling_delta_z=agrid,
        allow_empty_bins=True)
particle_flux_dens=espressomd.observables.FluxDensityProfile(ids=system.part[:].id,
        n_x_bins=size_in_agrid[0],
        n_y_bins=size_in_agrid[1],
        n_z_bins=size_in_agrid[2],
        min_x=0,
        min_y=0,
        min_z=0,
        max_x=system.box_l[0],
        max_y=system.box_l[1],
        max_z=system.box_l[2])
np.testing.assert_allclose(particle_flux_dens.calculate()*agrid**3,expected)
#np.testing.assert_allclose(lb_vel_profile.calculate(),expected)
print(lb_vel_profile.calculate())
```

RudolfWeeber:
Fixes #3603 

* respect min- and max_x,y,z rather than using 0 and box_size
* remove off-by-one error for the number of sampling positions for certain choices of box size and sampling_delta or offset

This is probably the least invasive way to fix the issue.
LB and particle profile observables are consistent, when using the same min- and max_x,y,z and a sampling_offset of agrid/2.
Alternatively, the sampling_delta parameter could be removed, but then, min_x would have to be set to agrid/2 and max_x to box_l +agrid/2. That's probably more confusing.

### DIFF
--- a/src/core/observables/LBProfileObservable.hpp
+++ b/src/core/observables/LBProfileObservable.hpp
@@ -57,19 +57,19 @@ public:
     sampling_positions.clear();
     if (sampling_delta_x == 0 or sampling_delta_y == 0 or sampling_delta_z == 0)
       throw std::runtime_error("Parameter delta_x/y/z must not be zero!");
-    const auto n_samples_x = static_cast<size_t>(std::rint(
-        (box_geo.length()[0] - sampling_offset_x) / sampling_delta_x));
-    const auto n_samples_y = static_cast<size_t>(std::rint(
-        (box_geo.length()[1] - sampling_offset_y) / sampling_delta_y));
-    const auto n_samples_z = static_cast<size_t>(std::rint(
-        (box_geo.length()[2] - sampling_offset_z) / sampling_delta_z));
+    const auto n_samples_x =
+        static_cast<size_t>(std::rint((max_x - min_x) / sampling_delta_x));
+    const auto n_samples_y =
+        static_cast<size_t>(std::rint((max_y - min_y) / sampling_delta_y));
+    const auto n_samples_z =
+        static_cast<size_t>(std::rint((max_z - min_z) / sampling_delta_z));
     for (size_t x = 0; x < n_samples_x; ++x) {
       for (size_t y = 0; y < n_samples_y; ++y) {
         for (size_t z = 0; z < n_samples_z; ++z) {
-          sampling_positions.push_back(
-              Utils::Vector3d{{sampling_offset_x + x * sampling_delta_x,
-                               sampling_offset_y + y * sampling_delta_y,
-                               sampling_offset_z + z * sampling_delta_z}});
+          sampling_positions.push_back(Utils::Vector3d{
+              {min_x + sampling_offset_x + x * sampling_delta_x,
+               min_y + sampling_offset_y + y * sampling_delta_y,
+               min_z + sampling_offset_z + z * sampling_delta_z}});
         }
       }
     }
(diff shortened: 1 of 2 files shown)


src/core/observables/LBProfileObservable.hpp
```diff
@@ -57,19 +57,19 @@ class LBProfileObservable : public ProfileObservable {
...
-    const auto n_samples_y = static_cast<size_t>(std::rint(
-        (box_geo.length()[1] - sampling_offset_y) / sampling_delta_y));
-    const auto n_samples_z = static_cast<size_t>(std::rint(
-        (box_geo.length()[2] - sampling_offset_z) / sampling_delta_z));
+    const auto n_samples_x =
+        static_cast<size_t>(std::rint((max_x - min_x) / sampling_delta_x));
+    const auto n_samples_y =
+        static_cast<size_t>(std::rint((max_y - min_y) / sampling_delta_y));
+    const auto n_samples_z =
+        static_cast<size_t>(std::rint((max_z - min_z) / sampling_delta_z));
```
jngrad, line 65: I don't understand the logic here. Let's take the observable `LBVelocityProfile` as an example. We sample the LB velocity every `sampling_delta_*`, but the histogram stores the result in an array of dimensions `n_*_bins`. This is why you had zeroes on the last dimension of each coordinate: there is no guarantee that the allocated histogram has the same dimensions as the ones calculated here. You could even get into a segfault if you don't correctly guess the values and give too small `n_*_bins` sizes. Am I missing something here?

KaiSzuttor: the histogram is independent of the LB fluid sampling. There might be the case of no hits for a bin. That's why there is a flag ```allow_empty_bins```.

jngrad: alright, thanks!

KaiSzuttor (approved):

Issue #2832, opened 2019-05-13
HTTP status codes


jngrad:
At the moment scripts posting to GitHub PRs ([maintainer/gh_post_style_patch.py](https://github.com/espressomd/espresso/blob/python/maintainer/gh_post_style_patch.py), [maintainer/gh_post_docs_warnings.py](https://github.com/espressomd/espresso/blob/python/maintainer/gh_post_docs_warnings.py)) do not check for HTTP status codes. When the status code isn't 200, `comments.json()` returns a dict of the HTTP status instead of a list of key/value pairs, leading to cryptic messages ([logfile](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/112798)) because iterating a dict generates strings:
```
Traceback (most recent call last):
  File "/builds/espressomd/espresso/maintainer/gh_post_docs_warnings.py", line 43, in <module>
    if comment['user']['login'] == 'espresso-ci' and prefix + '_warnings.sh' in comment['body']:
TypeError: string indices must be integers
```
The python `request` module has a section on [checking status codes](https://2.python-requests.org/en/master/user/quickstart/#response-status-codes). We should implement a retry mechanism with delay for `requests.get()`, but for `requests.delete()` and `requests.post()` I'm not sure. Currently both Python scripts run through a bash script without `set -e`, so they fail silently.

KaiSzuttor:
maybe we should have a look into github api wrappers, *e.g.* [PyGithub](https://github.com/PyGithub/PyGithub)

KaiSzuttor:
this has not been an issue for quite some time

jngrad:
> this has not been an issue for quite some time

True. Furthermore, the scripts now halt when an HTTP error code is found, instead of printing a cryptic error message (5636ee03cda5).

PR #3613, opened 2020-03-31
Update espresso-ci scripts

### LINKED ISSUE #3612: Update espresso-ci bot with new GET syntax
GitHub has deprecated the GET syntax we use for espresso-ci ([Deprecating API authentication through query parameters](https://developer.github.com/changes/2020-02-10-deprecating-auth-through-query-param/)). We currently get a monthly email reminder, then starting September 30th the GitHub API will randomly fail to incentivize the migration to the new GET syntax.

jngrad:
Description of changes:
- use the new GET syntax for the GitHub API (fixes #3612)
- print a clear error message and exit Python with a non-zero return value when the GitHub API returns an HTTP error code (partial fix for #2832)

### DIFF
--- a/maintainer/gh_post_docs_warnings.py
+++ b/maintainer/gh_post_docs_warnings.py
@@ -28,7 +28,8 @@ if not os.environ['CI_COMMIT_REF_NAME'].startswith('PR-'):
 
 PR = os.environ['CI_COMMIT_REF_NAME'][3:]
 URL = 'https://api.github.com/repos/espressomd/espresso/issues/' + \
-      PR + '/comments?access_token=' + os.environ['GITHUB_TOKEN']
+      PR + '/comments'
+HEADERS = {'Authorization': 'token ' + os.environ['GITHUB_TOKEN']}
 SIZELIMIT = 5000
 
 doc_type, has_warnings, filepath_warnings = sys.argv[-3:]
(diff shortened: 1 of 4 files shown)


src/core/statistics.hpp
```diff
@@ -59,7 +59,7 @@ int get_n_part_conf();
 double mindist(PartCfg &, IntList const &set1, IntList const &set2);
 
 /** Find all particles within a given radius @p r_catch around a position.
- *  @param partCfg    @copybrief PartCfg
+ *  @param partCfg    @copybrieef PartCfg
```
fweik: typo

KaiSzuttor (approved):
LGTM modulo tmp changes

jngrad:
I should have seen that coming: we need a chained if-else to test all possible names for clang-format and we need to use `clang-format-9` instead of 6. The difference is minimal (only 2 files) but it's a circular change with version 6, i.e. once formatted with version 9, formatting it with the version 6 installed at the ICP will revert the change... So we're stuck with the Ubuntu 16.04 image until the ICP migrates to Ubuntu 20.04, which might take us months.

KaiSzuttor:
I already approved

fweik:
Yes, but there were new changes after, did you look at them?

fweik:
Also, since when is the approval not removed if new changes are pushed?

jngrad:
> Also, since when is the approval not removed if new changes are pushed?

I realized that behavior several weeks ago and found it strange too, but did not question it. The "marked this pull request as ready for review 11 hours ago" message it posted automatically when removing the draft status of a PR. I did not add the `automerge` label immediately because we had to wait for the PR to complete to make sure we could trigger the webhook for a successful CI pipeline.

fweik:
It's a setting in the branch protection rules. Not sure if this conflicts with kodiak's merge ~~request~~ commits, if so it has to be deactivated again.

PR #3615, opened 2020-03-31
Proper memory management for ParticleList

fweik:
Description of changes:
- Proper serialization for `Particle` and sub-types.
- Proper memory ressource management for `ParticleList`
- Removed manual construction/desctruntion of Particles
- Fixed leak of bond memory under some circustances
- No more placement new!


src/core/Particle.hpp
```diff
@@ -170,6 +171,57 @@ struct ParticleProperties {
 #ifdef ENGINE
   ParticleParametersSwimming swim;
 #endif
+
+  template <class Archive> void serialize(Archive &ar, long int /* version */) {
+    ar &identity;
+    ar &mol_id;
```
jngrad, line 177: Just for clarification: before this PR the sub-structs are serialized as byte streams, and we now use the standard way of serializing members? It's not immediately obvious to me what is the benefit here, given that we now have to write every member twice with the correct ifdef logic. Does it have anything to do with being able to reconstruct a Particle with pointers members in the future? Or, did you forget to delete `serialization/Particle.hpp`? In that case, this makes sense.

fweik: The old code looked like this:

```c++
ar >> make_array(reinterpret_cast<char *>(&p), sizeof(Particle));
new (&(p.bl)) IntList(p.bl.size());
ar >> p.bl;
```

Doing the placment new over existing particles can leak (because their
destructors are not run), and to replace that, everything has to be done
properly. It's unfortunate that the members have to be repeated, but I
think we just have to live with that for now (because of C++'s lack of
introspection). One could add type annotations (e.g. for 
[boost.hanna](https://boostorg.github.io/hana/index.html#tutorial-introspection-adapting)), but that's out of scope here.

> Or, did you forget to delete serialization/Particle.hpp?

Jup, thanks.

fweik: In principle we could just bitwise serialize the sub-structs, but that's also not optimal: If a
non-trivial member is added later, this will cause totally non-obvious errors.

jngrad: I was confused because this new serialization strategy came on top of the previous one, but if you delete the old one, it all makes sense.

jngrad (approved):

PR #3617, opened 2020-04-01
Check_resort: return RESORT_GLOBAL as default

### LINKED ISSUE #3616: topology_check_resort inconsistent type
Introduced in 7f4caf7, `topology_check_resort` returns `true` as default. This was used in `cells_update_ghosts` to decide if resorting is necessary or not. Later in 4e468e9, the function was changed to return an `unsigned` and is currently defined as follows:
```c++
unsigned topology_check_resort(int cs, unsigned local_resort) {
  switch (cs) {
  case CELL_STRUCTURE_DOMDEC:
  case CELL_STRUCTURE_NSQUARE:
    return boost::mpi::all_reduce(comm_cart, local_resort,
                                  std::bit_or<unsigned>());
  default:
    return true; // <----
  }
}
```
Results should semantically be values of `enum Cells::Resort`.

Now I don't directly do a PR because I don't know what you @fweik intend to be returned here: `Cells::Resort::LOCAL` or `Cells::Resort::GLOBAL`. In my opinion it should be `GLOBAL`. Because if this case would ever happen, we need to return a value `v` s.t. `v >= local_resort` on all processes. And this we can guarantee only for `GLOBAL`.

hirschsn:
Fixes #3616 

Description of changes:
- Return GLOBAL as a default.

(Edit: Sorry, wrong issue referenced.)


fweik (approved):
Thanks!

PR #3619, opened 2020-04-01
maintainer: use new images.

KaiSzuttor:
Fixes #

Description of changes:
-

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,7 +1,4 @@
-variables: 
-  TAG: f6612ba27629a89f8c3a592740803eec777388ce
-
-image: docker.pkg.github.com/espressomd/docker/ubuntu-18.04:${TAG}
+image: docker.pkg.github.com/espressomd/docker/ubuntu-18.04:f6612ba27629a89f8c3a592740803eec777388ce
 
 stages:
   - permission
@@ -52,7 +49,7 @@ status_pending:
 style:
   <<: *global_job_definition
   stage: prepare
-  image: docker.pkg.github.com/espressomd/docker/clang:${TAG}
+  image: docker.pkg.github.com/espressomd/docker/clang:f6612ba27629a89f8c3a592740803eec777388ce
   dependencies: []
   before_script:
     - git submodule deinit .
@@ -72,7 +69,7 @@ style:
 style_doxygen:
   <<: *global_job_definition
   stage: prepare
-  image: docker.pkg.github.com/espressomd/docker/ubuntu-cuda-10.1:${TAG}
+  image: docker.pkg.github.com/espressomd/docker/ubuntu-cuda-10.1:f6612ba27629a89f8c3a592740803eec777388ce
   dependencies: []
   script:
     - mkdir build
@@ -121,7 +118,7 @@ no_rotation:
 debian:10:
   <<: *global_job_definition
   stage: build
-  image: docker.pkg.github.com/espressomd/docker/debian:${TAG}
+  image: docker.pkg.github.com/espressomd/docker/debian:f6612ba27629a89f8c3a592740803eec777388ce
   script:
     - export with_cuda=false
     - export myconfig=maxset make_check_python=false
@@ -133,7 +130,7 @@ debian:10:
 opensuse:15.1:
   <<: *global_job_definition
   stage: build
-  image: docker.pkg.github.com/espressomd/docker/opensuse:${TAG}
+  image: docker.pkg.github.com/espressomd/docker/opensuse:f6612ba27629a89f8c3a592740803eec777388ce
   script:
     - export with_cuda=false myconfig=maxset make_check_python=false
     - bash maintainer/CI/build_cmake.sh
@@ -144,7 +141,7 @@ opensuse:15.1:
 centos:7:
   <<: *global_job_definition
   stage: build
-  image: docker.pkg.github.com/espressomd/docker/centos:${TAG}
+  image: docker.pkg.github.com/espressomd/docker/centos:f6612ba27629a89f8c3a592740803eec777388ce
   script:
     - export with_cuda=false myconfig=maxset make_check_python=true
     - bash maintainer/CI/build_cmake.sh
@@ -155,7 +152,7 @@ centos:7:
 fedora:
   <<: *global_job_definition
   stage: build
-  image: docker.pkg.github.com/espressomd/docker/fedora:${TAG}
+  image: docker.pkg.github.com/espressomd/docker/fedora:f6612ba27629a89f8c3a592740803eec777388ce
   script:
     - export with_cuda=false myconfig=maxset make_check_python=false
     - bash maintainer/CI/build_cmake.sh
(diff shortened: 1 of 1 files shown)


fweik:
Do you known https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/ci/lint ?

KaiSzuttor:
yeah by now :)

fweik:
~~Are you sure that the dash is allowed in the image name?~~ I can't even read regex...

fweik:
https://stackoverflow.com/questions/42642561/docker-restrictions-regarding-naming-container#42642992

KaiSzuttor:
on my local machine I can do 
``` bash
docker pull docker.pkg.github.com/espressomd/docker/ubuntu-18.04:f6612ba27629a89f8c3a592740803eec777388ce
```

KaiSzuttor:
firewall

fweik:
Can't you push to the gitlab repo? That would be better anyway, then it's local to the CI runners

KaiSzuttor:
according to @mkuron that is not possible again bc of firewall

KaiSzuttor:
https://github.community/t5/GitHub-Actions/docker-pull-from-public-GitHub-Package-Registry-fail-with-quot/m-p/32888#M1294

fweik:
Still better to have it local no?

KaiSzuttor:
if the pull-time/build-time ratio is large, yes

fweik:
The reference you posted seems to imply that you just have to include an auth token?

KaiSzuttor:
i have to authenticate on gitlab, yes

RudolfWeeber:
There is some merit in having the images on a public registry, though. Other people can use them as well, then.

KaiSzuttor:
whats the problem with that?

fweik:
> There is some **merit** in having the images on a public registry, though.

jngrad:
He meant it's an advantage. I think we should also have the dockerfiles in the espresso repo as well, such that everything is self-contained.

fweik:
Well its not really public, though. You need auth...

RudolfWeeber:
Yes. I meant, having the images publicly accessible would be an advantage.

I'm confused. Is it really the case that one can't make the docker images on github readable for everyone without authentication?

KaiSzuttor:
hopefully, this is only a temporary restriction

fweik:
@RudolfWeeber yes, as stated in the [link that Kai posted](https://github.community/t5/GitHub-Actions/docker-pull-from-public-GitHub-Package-Registry-fail-with-quot/m-p/32888#M1294).

RudolfWeeber:
OK. Then we can store them at the ICP just as well.

 

Why do they do that, though?

KaiSzuttor:
> OK. Then we can store them at the ICP just as well.

That's not possible due to firewall restrictions.

KaiSzuttor:
We could temporarily use dockerhub...

fweik:
I'm not sure that I am still following? What is the problem now?

KaiSzuttor:
i tried to authenticate on gitlab but somehow could not manage to get that going. We also cannot push to gitlab due to the firewall. So we could temporarily also push to dockerhub and pull from there.

fweik:
I supposed you followed https://docs.gitlab.com/ce/ci/docker/using_docker_images.html#define-an-image-from-a-private-container-registry and updated the runner config?

KaiSzuttor:
yes and no, I tried a job env variable, now trying runner config

KaiSzuttor:
no luck either

KaiSzuttor:
works...

KaiSzuttor:
for the record: you have to create an access token for a user, say ```espresso-ci``` with read packages rights. Then do ```echo -n 'espresso-ci:<TOKEN>' | base64```. The resulting string can then be used as the ```auth``` string as shown in the [gitlab documentation](https://docs.gitlab.com/ee/ci/docker/using_docker_images.html#using-statically-defined-credentials)

KaiSzuttor:
@jngrad did I remove the latex stuff too early? You said it's only needed for 4.1

fweik:
> for the record: you have to create an access token for a user, say `espresso-ci` with read packages rights. Then do `echo -n 'espresso-ci:<TOKEN>' | base64`. The resulting string can then be used as the `auth` string as shown in the [gitlab documentation](https://docs.gitlab.com/ee/ci/docker/using_docker_images.html#using-statically-defined-credentials)

Please add this to the CI docs

KaiSzuttor:
@jngrad could please add that to the CI doc and remove outdated stuff?

jngrad:
> did I remove the latex stuff too early? You said it's only needed for 4.1

Sorry about that, didn't realize doxygen was using BibTeX. I'll have to update the release notes accordingly. `texlive-base` should be enough and only takes 70 MB.

> could please add that to the CI doc and remove outdated stuff?

What do you mean by outdated stuff? I'm lost here.
- Are we leaving the GitHub docker images on the espressomd/docker repo? Why is that? Is it because GitHub Actions trigger the docker pipeline for every commit, not just the commits where the dockerfiles changed?
- What will happen to 4.1.3, do we also apply these changes? This means we drop support for intel and emulated archs in a bugfix release.
- I thought this was a proof-of-concept. If we merge this PR and the docker PR, we need to move the base commit of the docker PR to a new branch called `4.1+4.2` and accept that we won't be able to regenerate their images anymore, only re-deploy those that are still in the container registry (because we can't use `git clone -b <commit>` during the build stage). Currently opened PRs like waLBerla and Stokesian will need to merge `python` to be able to get the new images.

KaiSzuttor:
* > Are we leaving the GitHub docker images on the espressomd/docker repo? Why is that? Is it because GitHub Actions trigger the docker pipeline for every commit, not just the commits where the dockerfiles changed?

The docker images reside on github for now because we cannot push them to gitlab due to a firewall rule that we cannot easily change.

* > What will happen to 4.1.3, do we also apply these changes? This means we drop support for intel and emulated archs in a bugfix release.

We used the intel and emulated archs builds in order to find bugs. There are no users for emulated archs plus we don't drop support for intel, we just don't test for it.

* > I thought this was a proof-of-concept. If we merge this PR and the docker PR, we need to move the base commit of the docker PR to a new branch called 4.1+4.2 and accept that we won't be able to regenerate their images anymore, only re-deploy those that are still in the container registry (because we can't use git clone -b <commit> during the build stage). Currently opened PRs like waLBerla and Stokesian will need to merge python to be able to get the new images.

Sorry, I don't get what you want to say here.

RudolfWeeber:
Concerning 4.1.3: I'd not worry about loosing testing of exotic archs. For Intel I don't have a firm opinion. We could just compile it once manually at the ICP.
On the other hand, I wouldn't be willing to invest in Intel support if it didn't work anymore.

With regards to open PRs: they need to merge Python from time to time anyway, andd if it's currently a bad moment, the CI-related commit can probably be cherry-picked without major conflicts.

We can discuss it once more in a meeting, if need be, but once we are sure, the new stuff is stable, I'd personally not delay much longer.

jngrad:
> > *  Are we leaving the GitHub docker images on the espressomd/docker repo? Why is that? Is it because GitHub Actions trigger the docker pipeline for every commit, not just the commits where the dockerfiles changed?
> 
> 
> The docker images reside on github for now because we cannot push them to gitlab due to a firewall rule that we cannot easily change.

I get the issue about the firewall, but why do we need to store the dockerfiles on espressomd/docker?

> Sorry, I don't get what you want to say here.

I had plans to untangle docker images for the 4.1 release and 4.2 release and push them to branches `4.1` and `4.2`. I can't do that if if 4.1.3 uses the new docker images. 4.1.3 will have to use a mix of the new docker images from GitHub and old images from our registry.

RudolfWeeber:
I think the docker files should be in a separate repo so the build and CI stuff is separate.

We don't have to keep our copy of espressomd/docker on gitlab, anymore.

jngrad:
@KaiSzuttor would you like to squash the temporary commits before merging?

fweik:
I think we ant to merge #3622

PR #3621, opened 2020-04-01
Modernize Cell and ParticleList

### LINKED ISSUE #3539: ParticleList:: switch to boost::vector

fweik:
Fixes #3539.

Depends on #3618.

Description of changes:
- More STL-like interface for `ParticleList`
- Replaced manual management by vector
- Avoid unneeded resizes of ghost cells during resort
- Made particles a member of `Cell` instead of a base


src/utils/include/utils/Bag.hpp
```diff
@@ -0,0 +1,160 @@
...
+
+    m_storage.pop_back();
+
+    return it;
+  }
+
+  /**
+   * @brief Swap two Bags.
+   *
+   * Efficiently swap to bags by swapping
```
jngrad, line 150:
```suggestion
   * Efficiently swap two bags by swapping
```

src/utils/include/utils/Bag.hpp
```diff
@@ -0,0 +1,160 @@
...
+ * @brief Bag of elements.
+ *
+ * A bag is a container in which the elements do not have
+ * a fixed order. It can be considered an unordered variant
+ * of vector, and implements the Container named requirement
+ * specified in C++11.
+ *
+ * Elements in the container do not have a stable position and
+ * removing elements can change the order of the other elements.
+ * The loser contract (compared to a vector) allows removing any
```
jngrad, line 20:
```suggestion
 * The looser contract (compared to a vector) allows removing any
```

jngrad (changes requested):
Overall LGTM

jngrad (dismissed):
ok from my side

RudolfWeeber:
How do cells shrink again, once many particles have been in them and left.

fweik:
They don't, because there is no obvious place to do it. Do you think that is needed, and if so, why?

KaiSzuttor (dismissed):
LGTM

fweik:
@jngrad can you please approve again, I did a rebase. Maybe it is slightly annoying if the reviews are dismissed on change ^^

jngrad (approved):

RudolfWeeber:
I'm not sure about the never shrinking. 
Was that also the case for the old implementation?

Issues might arise for systems, where a really dense region migrates through space.
Say: A single polymer chain and a small number of much larger colloids. The colloid-colloid interaction would force a large cell size.

We probably have to try that out.

Otherwise, the PR looks good.

fweik:
The old implementation did shrink the cells. I think performance wise it's better to just not shrink it, that does not only avoid allocations, but also index updates for the whole cell. Memory is usually not
a limiting factor in MD simulations. Also I'd say if you have all of your particles in one cell, running out of memory is not your main problem :-) Let's see if this comes up, if so we can add an option to reclaim memory on resort or something.

PR #3623, opened 2020-04-02
ROCm 3.3 support

### LINKED ISSUE #3587: CI build failed for merged PR
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/pipelines/11478

### LINKED ISSUE #3620: ROCm: Memory access fault by GPU node
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/pipelines/11601

mkuron:
The `ln -s /opt/rocm/bin/hcc* /opt/rocm/hip/bin/` issue has been worked around by properly setting `HCC_PATH` on the CMake side.
The shutdown issue has been worked around by replacing interrupts with polling (suggested at https://github.com/ROCm-Developer-Tools/roctracer/issues/22#issuecomment-572814465). Something is wrong with the destruction order in our code, but I cannot easily identify what. It's not the missing `cudaDestoryStream` though.

Fixes #3620 (according to `ctest -R save_checkpoint_lb.cpu-p3m.cpu-lj-therm.lb_1 --repeat-until-fail 1000`).
Fixes #3587 (according to `ctest -R ek_charged_plate --repeat-until-fail 100`).

**TODO**
- https://github.com/espressomd/docker/blob/master/docker/rocm-python3/Dockerfile-latest needs to be updated to ROCm 3.3 once this pull request is merged.


mkuron:
No idea what's wrong with the ek_fluctuations test. Philox works, LB thermalization statistics are correct, EK thermalization statistics are incorrect. The responsible function does not look suspicious in any way:
https://github.com/espressomd/espresso/blob/8813c4cff7ea6cc8c4b848f4a9531d81babb27ff/src/core/grid_based_algorithms/electrokinetics_cuda.cu#L1346-L1414
Is anyone familiar with that test?

jngrad:
Note: while tinkering with `/opt/rocm/bin/hipcc` compiler flags in #3582, I get `ek_fluctuations` to fail reproducibly when passing flags `-O3`, `-Og` or `-g` to the compiler for .cu files.

cmake/FindCUDACompiler.cmake
```diff
@@ -64,14 +64,19 @@ else()
...
+    endif()
+
     find_package(HIP MODULE)
     message(STATUS "Found HIP compiler: ${HIP_HIPCC_EXECUTABLE}")
     set(CUDA 1)
     set(HIP 1)
     set(CUDA_COMPILER_EXE ${HIP_HIPCC_EXECUTABLE})
     list(APPEND HIP_HCC_FLAGS "-I${HIP_ROOT_DIR}/include -I${ROCM_HOME}/include -Wno-c99-designator -Wno-macro-redefined -Wno-duplicate-decl-specifier -std=c++${CMAKE_CXX_STANDARD}")
-    list(APPEND HIP_HCC_FLAGS "-pedantic -Wall -Wextra -Wno-sign-compare -Wno-unused-function -Wno-unused-variable -Wno-unused-parameter -Wno-missing-braces -Wno-gnu-anonymous-struct -Wno-nested-anon-types -Wno-gnu-zero-variadic-macro-arguments")
+    list(APPEND HIP_HCC_FLAGS "-pedantic -Wall -Wextra -Wno-sign-compare -Wno-unused-function -Wno-unused-variable -Wno-unused-parameter -Wno-missing-braces -Wno-gnu-anonymous-struct -Wno-nested-anon-types -Wno-gnu-zero-variadic-macro-arguments -Wno-deprecated-copy")
```
jngrad: `error: unknown warning option '-Wno-deprecated-copy'; did you mean '-Wno-deprecated'? [-Werror,-Wunknown-warning-option]`
if it's a ROCm3.3-specific flag, we should add a version guard

mkuron:
> I get ek_fluctuations to fail reproducibly when passing flags `-O3`, `-Og` or `-g` to the compiler for .cu files.

Interesting. I'll bisect optimization flags tomorrow to find out what is causing it.

**EDIT**: `-fsave-optimization-record` helped me discover that `-fno-inline-functions` works around the issue. I guess there is an unsafe inline somewhere.

mkuron:
Fixed now. That `nullptr` hack seems to have weird interactions with inlining in recent LLVM versions.

jngrad:
`ek_fluctuations` is now fixed, thanks!

What should we do now with regards to ROCm support? We aren't the only ones affected by the disastrous ROCm versioning strategy. Only supporting ROCm 3.3 might be an issue for users who pinned an earlier ROCm version on their system to stabilize their environment. Should we test both ROCm 3.0 and 3.3 in CI? We could test one on a weekly schedule to limit workload. Only testing 3.3 in CI runs the risk of running into a regression for 3.0.

jngrad (approved):

mkuron:
> Only supporting ROCm 3.3 might be an issue for users who pinned an earlier ROCm version on their system to stabilize their environment.

This has become less of an issue now that you can install multiple ROCm versions side-by-side.

The changes in this merge request certainly won't break compatibility with 3.0. Testing 3.0 only wouldn't be sufficient either as someone might be pinning an even older version. Testing every release since 2.0 isn't an option either. That means we can only support older versions on a best-effort basis and simply guard any problematic changes with version checks. Looking back at my ROCm compatibility patches since v2.0, they primarily deal with CMake issues. Silent breakage like the ek_fluctuations test is a rather rare thing.

PR #3625, opened 2020-04-02
Local particles

fweik:
The goal here is to make it transparent which call sites actually need access to
the cells, and which just need to iterate the particles. There are a quite a lot of
changes her, but most of them are just textual replacements, so should be fast
to check. The final goal here is to be able to hide the cells from client code of
the particle storage. From my side #2899 can be closed with this.

Description of changes:
- Add functions {local_particles, ghost_particles} to CellStructure to directly get the local/ghost particle ranges
- Replaced all occurrences of particles access thru the {local, ghost} cell ranges by
  direct calls to the aforementioned functions
- Replaced CellPList by the now functionally equivalent Utils::Span

### DIFF
--- a/src/core/CellStructure.hpp
+++ b/src/core/CellStructure.hpp
@@ -53,17 +53,19 @@ enum Resort : unsigned {
 };
 }
 
+namespace Cells {
+inline ParticleRange particles(Utils::Span<Cell *> cells) {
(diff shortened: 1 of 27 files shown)


jngrad (approved):
LGTM

PR #3628, opened 2020-04-03
utils: Made List behave correctly on self-assignment.

fweik:
We can't add a test, because explicit self-assignment causes a warning with clang.


jngrad (approved):

fweik:
I don't think this needs to be ported back to 4.1

PR #3636, opened 2020-04-06
WIP: Sanitizer build with Ubuntu 20.04

RudolfWeeber:


KaiSzuttor:
you can try locally by:
```docker run -it docker.pkg.github.com/espressomd/docker/ubuntu-20.04:a1192b35590a1a474c55fe1e9a1e6c25758454ea bash```

**Comment by jngrad** [comment]
```
2382 /builds/espressomd/espresso/build/src/python/espressomd/collision_detection.cpp:10435:5: error: 'tp_print' is deprecated [-Werror,-Wdeprecated-declarations]
2383     0,
2384     ^
2385 /usr/include/python3.8/cpython/object.h:260:5: note: 'tp_print' has been explicitly marked deprecated here
2386     Py_DEPRECATED(3.8) int (*tp_print)(PyObject *, FILE *, int);
2387     ^
2388 /usr/include/python3.8/pyport.h:515:54: note: expanded from macro 'Py_DEPRECATED'
2389 #define Py_DEPRECATED(VERSION_UNUSED) __attribute__((__deprecated__))
```
To resolve that issue, it seems we need to either downgrade Cython 0.29.14 to 0.29.13, or add the compiler flag `-Wno-deprecated-declaration` (https://github.com/pandas-dev/pandas/issues/30609#issuecomment-605373212)

**Comment by RudolfWeeber** [comment]
>docker run -it docker.pkg.github.com/espressomd/docker/ubuntu-20.04:a1192b35590a1a474c55fe1e9a1e6c25758454ea bash

I’m aware of that. The deprecation stuff didn’t happen, locally, for whatever reason.

**Comment by RudolfWeeber** [comment]
Boost MPI again, it seems:
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/221196

**Comment by fweik** [comment]
Yup, the offending line is `/usr/include/boost/mpi/detail/binary_buffer_iprimitive.hpp:44`, which calls `front()` on `std::vector`, which is undefined if the vector is empty.

**Comment by fweik** [comment]
It's also trivial to fix, which doesn't help because we don't control the code.

**Comment by RudolfWeeber** [comment]
>It's also trivial to fix, which doesn't help because we don't control the code.

Suppression file it is, then.

**Comment by fweik** [comment]
I don't know. The sanitizer says this is a nullpointer dereference, so this might also segfault

**Comment by fweik** [comment]
Ah, wait a second. The buffer is passed into the class, maybe it can be fixed from the outside...

**Comment by fweik** [comment]
No, it can't. This was also introduced quite recently. We should really ship our own version of boost mpi.

File: .gitlab-ci.yml [hunk 1]
```diff
@@ -102,6 +102,17 @@ no_rotation:
     - docker
     - linux
 
+ubuntu-20.04-sanitizer:
+  <<: *global_job_definition
+  image: docker.pkg.github.com/espressomd/docker/ubuntu-20.04:a1192b35590a1a474c55fe1e9a1e6c25758454ea
+  stage: build
+  script:
+    - export myconfig=maxset with_cuda=false with_coverage=false with_static_analysis=true with_asan=true with_ubsan=true test_timeout=900 ASAN_OPTIONS="allocator_may_return_null=1" OMPI_CC=clang-9 OMPI_CXX=clang++-9 CC=clang-9 CXX=clang++-9 UBSAN_OPTIONS=suppressions=/home/espresso/espresso/maintainer/CI/ubsan.supp
```
jngrad:
```suggestion
    - export myconfig=maxset with_cuda=false with_coverage=false with_static_analysis=true with_asan=true with_ubsan=true test_timeout=900 ASAN_OPTIONS="allocator_may_return_null=1" OMPI_CC=clang-9 OMPI_CXX=clang++-9 CC=clang-9 CXX=clang++-9 UBSAN_OPTIONS=suppressions=${CI_PROJECT_DIR}/maintainer/CI/ubsan.supp
```

jngrad (changes requested):

.gitlab-ci.yml
```diff
@@ -102,6 +102,17 @@ no_rotation:
     - docker
     - linux
 
+ubuntu-20.04-sanitizer:
+  <<: *global_job_definition
+  image: docker.pkg.github.com/espressomd/docker/ubuntu-20.04:a1192b35590a1a474c55fe1e9a1e6c25758454ea
+  stage: build
+  script:
+    - export myconfig=maxset with_cuda=false with_coverage=false with_static_analysis=true with_asan=true with_ubsan=true test_timeout=900 ASAN_OPTIONS="allocator_may_return_null=1" OMPI_CC=clang-9 OMPI_CXX=clang++-9 CC=clang-9 CXX=clang++-9 UBSAN_OPTIONS=suppressions=${CI_PROJECT_DIR}/maintainer/CI/ubsan.supp
```
KaiSzuttor:
```suggestion
   variables:
     myconfig: 'maxset'
     with_cuda: 'false'
     with_coverage: 'false'
     with_static_analysis: 'true'
     with_asan: 'true'
     with_ubsan: 'true'
     test_timeout: '900'
     ASAN_OPTIONS: 'allocator_may_return_null=1'
     OMPI_CC: 'clang-9'
     OMPI_CXX: 'clang++-9'
     CC: 'clang-9'
     CXX: 'clang++-9'
     UBSAN_OPTIONS: 'suppressions=${CI_PROJECT_DIR}/maintainer/CI/ubsan.supp'
  script:
```

KaiSzuttor (commented):
use env vars

jngrad:
@RudolfWeeber This PR looks ready to me. What is still missing?
@KaiSzuttor We should delay your proposed env var changes to avoid conflicts with #3582, or only apply it to the Ubuntu 20.04 job.

src/core/cuda_wrapper.hpp
```diff
@@ -22,6 +22,7 @@
 #if defined(__CUDACC__)
 
 #include <cuda.h>
+#include "clang_thrust_workaround.cuh"
```
jngrad, line 24: The file `clang_thrust_workaround.cuh` seems to be missing. Also, I just found out the issue was fixed in thrust 1.9.5 (I'm writing a report for them)
```suggestion
#if THRUST_VERSION < 100905
#include "clang_thrust_workaround.cuh"
#endif
```

jngrad: Here is my investigation: https://github.com/thrust/thrust/issues/1032#issuecomment-611617047; if CUDA 10.1 is shipped with thrust 1.9.5, we should install it on ICP machines and drop support for the combination Clang+CUDA10.0 is ESPResSo.

jngrad: @RudolfWeeber sorry, I forgot that editing a message doesn't send an email. In my previous message, I meant to say that you probably forgot to commit file `clang_thrust_workaround.cuh`.

fweik:
For the record: I think the _workaround_ is a bad idea.

jngrad:
I'm also not sure if it can be done like this from the ESPResSo side. When I tinkered with that solution, I found the only way to get it to work was to apply it directly to the thrust code, but maybe I'm wrong.

RudolfWeeber:
It looks like the workaround is already present in the OSX container...

Let's make a decision about how to proceed in the ES meeting.

jngrad:
offline core team discussion: if Clang 8.0 can build with thrust 1.9.3 keep it, otherwise add thrust 1.9.5 as a submodule.

RudolfWeeber:
Core team discussion yielded that @jngrad will take over this PR.

* Sanitizers work with Clang9.
* At the ES meeting, it was decided to use Clang 8 and the latest GCC versions in the image and at the ICP rather than patching system headers to allow compilation with Cuda and more recent compilers
* The thrust workaround mentioned above might have to be removed again when building with Clang-8. If it is still needed, we decided to ship a new enough thrust with Esprseso as git submodule for the time being.

I'll post Docker changes to the Ubuntu 20 container separately.

PR #3637, opened 2020-04-07
Bond Refactor

### LINKED ISSUE #3084: Separate bond iteration, particle resolution and kernel
Bonds need ghost particle access and synchronization, and as such need to be handled by
the cell system, probably within `short_range_loop` or similar. To this end the iteration over
the bonds, the resolving of the id's to particles and the kernel execution needs to be separated.
The should be separated kernels that take a parameter struct and a list of particle pointers as
input, and then e.g. calculate the force or apply some other kernel.

fweik:
Fixes #3084.

This is obviously only one step in the large solution, but I think this PR is already
too big. Still I feel like this is a major improvement. The next step is to integrate
the bond handling with the short range loop, and move that into `CellStructure`,
so that the implementation details can be hidden. But one step at a time.

Description of changes:
- Proper data structure for bond lists
- Proper iterator for bonds
- Separate concerns between bond handling and bond parameters
- Replaced manual marshaling of bonds by serialization in ghost comm and mpiio
- Reorganization of bond kernels


src/core/BondList.hpp
```diff
@@ -0,0 +1,235 @@
...
+ * This holds the bonds of a particle.
+ *
+ * Implementation notes:
+ * Internally the bond list is represented as a sequence of
+ * integers. For each bond, first the ids of the bond partners
+ * are stored (which are positive numbers), then the negative
+ * of the value of the id of the bond plus one (which also is positive) is
+ * stored. This way we can use the sign bit of the int as a delimiter,
+ * every time a negative number is encountered one bond ends, and
+ * a new one starts. This mechanism allows is to efficiently store
```
jngrad:
```suggestion
 * a new one starts. This mechanism allows us to efficiently store
```

src/core/BondList.hpp
```diff
@@ -0,0 +1,235 @@
...
+   */
+  void insert(BondView const &bond) {
+    boost::copy(bond.partner_ids(), std::back_inserter(m_storage));
+    assert(bond.bond_id() >= 0);
+    m_storage.push_back(-(bond.bond_id() + 1));
+  }
+
+  /**
+   * @brief Erase a bond from the list.
+   * @param pos Iterator iterator pointing to elementn to erase.
```
jngrad:
```suggestion
   * @param pos Iterator pointing to element to erase.
```

src/core/BondList.hpp
```diff
@@ -0,0 +1,235 @@
...
+   * @return iterator pointing one past the erased element.
+   */
+  const_iterator erase(const_iterator pos) {
+    return Iterator{m_storage.erase(pos.m_it, std::next(find_end(pos.m_it)))};
+  }
+
+  /**
+   * @brief Number of bonds.
+   * @return The numebr of bonds int the list.
```
jngrad:
```suggestion
   * @return The number of bonds in the list.
```

src/core/unit_tests/BondList_test.cpp
```diff
@@ -0,0 +1,205 @@
...
+    BOOST_CHECK(not(BondView{id, partners} != BondView{id, partners_same}));
+    BOOST_CHECK(
+        not(BondView{id, partners} == BondView{id, partners_different}));
+    BOOST_CHECK(not(BondView{id, partners} == BondView{id + 1, partners_same}));
+  }
+}
+
+BOOST_AUTO_TEST_CASE(default_ctor) {
+  /* BondList can be default constructed and a default constructed BondList ist
+   * is empty. */
```
jngrad:
```suggestion
  /* BondList can be default constructed and a default constructed BondList
   * is empty. */
```

src/core/rattle.cpp
```diff
@@ -92,62 +88,73 @@ void save_old_pos(const ParticleRange &particles,
...
+                                          Utils::Span<Particle *> partners) {
+      auto const &iaparams = bonded_ia_params[bond_id];
+
+      if (iaparams.type == BONDED_IA_RIGID_BOND) {
+        auto const corrected = add_pos_corr_vec(iaparams, p1, *partners[0]);
+        if (corrected)
+          *repeat_ += 1;
+      }
+
+      /* Rigid bonds can not break */
```
jngrad:
```suggestion
      /* Rigid bonds cannot break */
```

src/core/BondList.hpp
```diff
@@ -0,0 +1,235 @@
...
+  const_iterator begin() const { return Iterator{m_storage.begin()}; }
+  /**
+   * @brief Iterator past the end of the range of bonds in the list.
+   */
+  const_iterator end() const { return Iterator{m_storage.end()}; }
+
+  /**
+   * @brief Add a bond to the list.
+   *
+   * @param bond to add.
```
jngrad:
```suggestion
   * @param bond Bond to add.
```

src/core/BondList.hpp
```diff
@@ -0,0 +1,235 @@
...
+
+  friend void swap(BondList &lhs, BondList &rhs) {
+    using std::swap;
+
+    swap(lhs.m_storage, rhs.m_storage);
+  }
+};
+
+/**
+ * @brief Check if the if a specific bond in a bond list.
```
jngrad:
```suggestion
 * @brief Check if there is a specific bond in a bond list.
```

src/core/BondList.hpp
```diff
@@ -0,0 +1,235 @@
...
+ *
+ * @param bonds List of bonds to check
+ * @param partner_id Id of the bond partner
+ * @param bond_id Id of the bond parameters.
+ * @return True iff the is a bond to the specified id of the specified type.
```
jngrad:
```suggestion
 * @return True iff there is a bond to the specified id of the specified type.
```

jngrad (changes requested):
That's a lot more readable, thanks! There are a few issues in the docs.

jngrad (dismissed):

fweik:
@jngrad when we bump the boost version, we should really grep for `BOOST_VERSION`

jngrad:
@fweik I've added the grep pattern in the open issue that tracks dependency version bumps in 4.2

fweik:
The issue picked up by the UB sanitizer is a bug in boost.container, that has since been fixed. Especially its fixed in version 1.66. Is there a schedule for bumping the boost version? Otherwise
I would just replace the boost vector by a stl vector, which would add 8 bytes per particle...

jngrad:
> Especially its fixed in version 1.66.

That's unfortunate, because Ubuntu 18.04 ships with `libboost-container1.65.1`, and `updates`/`backports` has nothing. Under these conditions, we'll most likely bump to 1.65. This can be done once `clang:6.0` is decommissioned (plan is in espressomd/docker#165).

fweik:
It's fixed from 1.58, so no problem. For the old versions we can just fall back to a stl vector.

jngrad (dismissed):

fweik:
To actually make the boost vector smaller, an even newer boost version is needed. I'll deal with this tomorrow together with the exclusions. This only affects the cache efficiency anyway...

jngrad (approved):

PR #3642, opened 2020-04-08
Update CI images

jngrad:
Description of changes:
- update image tags for non-ubuntu CI jobs
- update name of image for Travis-CI
- add manual jobs for `latest` images
- update docs deployment

PR #3648, opened 2020-04-09
CMake: use python -m sphinx instead of sphinx-build et al.

mkuron:
This helps in situations where the python3 version of `sphinx-build` is not in $PATH, but the python2 version is.

Also apply the same trick to Cython, where we already used it as a fallback.

Problem reported by @biermanncarl and discussed with @KaiSzuttor.

### DIFF
--- a/cmake/FindSphinx.cmake
+++ b/cmake/FindSphinx.cmake
@@ -1,25 +1,13 @@
 include(FindPackageHandleStandardArgs)
 
-find_program(
-  SPHINX_EXECUTABLE
-  NAMES sphinx-build
-        sphinx-build-${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}
-  HINTS $ENV{SPHINX_DIR} PATHS /opt/local/ /usr/local/
-                               $ENV{HOME}/Library/Python/2.7/ PATH_SUFFIXES bin
-  DOC "Sphinx documentation generator.")
+set(SPHINX_EXECUTABLE ${PYTHON_EXECUTABLE} -m sphinx)
+set(SPHINX_API_DOC_EXE ${PYTHON_EXECUTABLE} -m sphinx.apidoc)
 
-find_program(
-  SPHINX_API_DOC_EXE
-  NAMES sphinx-apidoc
-        sphinx-apidoc-${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR} HINTS
-  PATH_SUFFIXES bin PATHS /opt/local/ /usr/local/ $ENV{HOME}/Library/Python/2.7/
-  DOC "Sphinx api-doc executable.")
+execute_process(
+  COMMAND ${SPHINX_EXECUTABLE} --version OUTPUT_VARIABLE QUERY_VERSION_OUT
+  ERROR_VARIABLE QUERY_VERSION_ERR RESULT_VARIABLE QUERY_VERSION_RESULT)
 
-if(Sphinx_FIND_VERSION)
-  execute_process(
-    COMMAND "${SPHINX_EXECUTABLE}" --version OUTPUT_VARIABLE QUERY_VERSION_OUT
-    ERROR_VARIABLE QUERY_VERSION_ERR RESULT_VARIABLE QUERY_VERSION_RESULT)
-
+if(NOT QUERY_VERSION_RESULT)
   # Sphinx switched at some point from returning ther version on stdout to
   # printing it at stderr. Since we do not know ther version yet, we use stdout
   # if it matches a version regex, or stderr otherwise.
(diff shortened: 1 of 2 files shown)


cmake/FindSphinx.cmake
```diff
@@ -29,20 +17,19 @@ if(Sphinx_FIND_VERSION)
...
-  set(SPHINX_VERSION_COMPATIBLE TRUE)
-  # Blacklist broken version
-  if("${SPHINX_VERSION}" VERSION_EQUAL "2.1.0")
-    message(WARNING "Sphinx version 2.1.0 is not compatible.")
-    set(SPHINX_VERSION_COMPATIBLE FALSE)
-  endif()
+set(SPHINX_VERSION_COMPATIBLE TRUE)
+# Blacklist broken version
+if("${SPHINX_VERSION}" VERSION_EQUAL "2.1.0")
+  message(WARNING "Sphinx version 2.1.0 is not compatible.")
```
jngrad: to avoid conflicts with the Sphinx PR in #3642, we could either integrate this commit in #3642, or apply this patch:
```suggestion
if("${SPHINX_VERSION}" VERSION_EQUAL "2.1.0" OR "${SPHINX_VERSION}" VERSION_EQUAL "3.0.0")
  message(WARNING "Sphinx version ${SPHINX_VERSION} is not compatible.")
```

KaiSzuttor: that is a trivial merge, right?

KaiSzuttor (dismissed):

cmake/FindSphinx.cmake
jngrad: yes, but it's an annoying backport in 4.1.3 because merge commits break `git cherry-pick`

cmake/FindSphinx.cmake
```diff
@@ -1,25 +1,13 @@
...
 
-find_program(
-  SPHINX_EXECUTABLE
-  NAMES sphinx-build
-        sphinx-build-${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}
-  HINTS $ENV{SPHINX_DIR} PATHS /opt/local/ /usr/local/
-                               $ENV{HOME}/Library/Python/2.7/ PATH_SUFFIXES bin
-  DOC "Sphinx documentation generator.")
+set(SPHINX_EXECUTABLE ${PYTHON_EXECUTABLE} -m sphinx)
+set(SPHINX_API_DOC_EXE ${PYTHON_EXECUTABLE} -m sphinx.apidoc)
```
jngrad: This change and the one in the Cython file generate the following CMake log:
```
-- Found PythonInterp: /usr/bin/python3 (found suitable version "3.6.9", minimum required is "3.5") 
-- Found Cython: /usr/bin/python3;-m;cython (found suitable version "0.26.1", minimum required is "0.26") 
-- Found PythonHeaders: /usr/include/python3.6m  
-- Found NumPy: /tikhome/jgrad/.local/lib/python3.6/site-packages/numpy/core/include (found version "1.17.3") 
-- Found Sphinx: /usr/bin/python3;-m;sphinx (found suitable version "2.0.1", minimum required is "1.6.6") 
```
Is there a way to get the absolute path of Sphinx and Cython instead of a list cast to a string (`/usr/bin/python3;-m;sphinx`)?

jngrad: if not, could we instead quote the value of `SPHINX_EXECUTABLE`, `SPHINX_API_DOC_EXE`, `CYTHON_EXECUTABLE` to prevent the cast to string that adds semicolons?

mkuron: These semicolons are only added when printing the "Found XXX" message. The list is correctly unpacked when calling the program. If we convert the list to a string inside FindSphinx/FindCython, CMake later tries to call that whole string as the executable and obviously fails.

> Is there a way to get the absolute path of Sphinx and Cython

No. In fact, there might not even be one.

KaiSzuttor: let's first get in functional and then pretty, CI is failing...

mkuron:
Okay, seems to be working now.

cmake/FindSphinx.cmake
KaiSzuttor: @jngrad what's the issue with the semicolons?

KaiSzuttor:
I think this is also not needed... If you'd call the wrong executable on the command line because your ```PATH``` is not set correctly I do not expect Espresso to be able to guess where the "correct" executable can be found.

mkuron:
Sometimes there is no correct executable, so modifying `$PATH` gets you nowhere. See e.g. https://packages.ubuntu.com/bionic/all/python3-sphinx/filelist: it does not contain a `sphinx-build`.

KaiSzuttor:
it does, how should it work if now

KaiSzuttor:
second last line

mkuron:
> how should it work if not

It does not work on a freshly-installed Ubuntu 18.04. You cannot build documentation there without this patch. I tried it.

> second last line

Seriously, who adds `/usr/share/sphinx/scripts/python3` to their `$PATH` manually?

mkuron:
Also, this patch actually simplifies finding the tool because the process doesn't need to worry about paths.

PR #3661, opened 2020-04-14
fix_style: try python3 -m pre_commit if pre-commit not found

mkuron:
Same trick as in #3648. @biermanncarl reported being unable to use fix-style.sh because the pip binary directory is not in `$PATH` by default.

### DIFF
--- a/maintainer/CI/fix_style.sh
+++ b/maintainer/CI/fix_style.sh
@@ -26,8 +26,13 @@ if ! git diff-index --quiet HEAD -- && [ "${1}" != "-f" ]; then
 fi
 
 if ! hash pre-commit 2>/dev/null; then
-    echo "pre-commit command not found."
-    exit 2
+    python3 -m pre_commit 2>&1 >/dev/null
+    if [ "$?" = "0" ]; then
+        alias pre-commit="python3 -m pre_commit"
+    else
+        echo "pre-commit command not found."
+        exit 2
+    fi
 fi
 
 pre-commit run --all-files


KaiSzuttor:
I'm not sure if we want to fix the general issue of ```---user``` flag pip-installed executables not being in the ```PATH``` in our shell script. This is not related to the logic of our shell scripts at all and does not belong there, we should keep them simple.

mkuron:
As it turns out, pylint and cmake-format have the same problem. @biermanncarl is currently working on applying the same trick there. I think this should be fixed in the scripts because it makes them easier to use. The pip binary directory is usually not in `$PATH` by default, so why not make our scripts aware of that fact.

jngrad:
I also think it should be the user's responsibility to deal with the `$PATH` when installing packages via pip. Yet, I remember once reading a piece arguing for the generalization of `python3 -m package`, to fix more than just the `$PATH` issue (e.g. which python interpreter is used), but cannot find the source anymore. Is there a recommendation? I could only find [PEP 547](https://www.python.org/dev/peps/pep-0547/).

mkuron:
> Is there a recommendation?

I don't think there is an official recommendation from the Python developers, but it does not have any downsides as far as I can tell. And it has the obvious upside of not depending on anything but the python interpreter and its module search path.

KaiSzuttor:
well, it changes observable behavior and I'm not sure if it's needed.

mkuron:
It only changes behavior in cases where behavior was previously essentially undefined. If you only had these tools installed in one place or both your highest-precedence `$PATH` entry and your highest-precedence Python module search path entry led to the same version of the respective tool, you will still get that exact same behavior.

KaiSzuttor:
I think to make it clearer we should not have this as a fallback but rather use it as the only way to call the python scripts and document that we have decided to do so... or only use executables that can also be directly found on the command line. If we decide to use ```python -m``` calls, we should probably use the isolated mode ```-I``` according to [this post](https://stackoverflow.com/questions/22241420/execution-of-python-code-with-m-option-or-not). Supporting both possible ways to find the scripts is probably unnecessarily complex logic in the scripts. @jngrad what's your opinion?

maintainer/CI/fix_style.sh
```diff
@@ -26,8 +26,13 @@ if ! git diff-index --quiet HEAD -- && [ "${1}" != "-f" ]; then
 fi
 
 if ! hash pre-commit 2>/dev/null; then
-    echo "pre-commit command not found."
-    exit 2
+    python3 -m pre_commit 2>&1 >/dev/null
```
KaiSzuttor, line 29: can we rely on interpreter name being ```python3```?

mkuron: Ideally, we would pull that from CMake, but the maintainer scripts are independent of the build system. However, there is a PEP that guarantees names like `python` for Python 2 and `python3` for Python 3 and that `python` is never Python 3. Most notably, this guarantees you get the same Python interpreter that you get when you run `pip3`.

jngrad:
The `-I` flag should be avoided: it will cause python to skip `~/.local/lib/python3.X/site-packages/` and ignores the `$PYTHONPATH` variable. It's used when you want to access python packages installed on the system by the admin.

I agree with @KaiSzuttor though: providing the `python3 -m` method might make troubleshooting more difficult for users. If they forgot to set the `$PATH`, they have bigger issues to worry about than espresso scripts not working. Having sphinx and pre-commit work while other Python packages don't might make it more difficult for us to figure out what's wrong, as users will report that sphinx and pre-commit are in their `$PATH` because they were found, while in reality they were found by `python3 -m`.

KaiSzuttor:
My point was rather that we should only keep one version... either we rely on finding the script in ```PATH``` *or* in ```PYTHONPATH```. I think it would also be fine to always use the ```-m``` variant if we document it.

mkuron:
Okay, that sounds like a plan. I'll ask @biermanncarl to take over this pull request to implement it.

PR #3667, opened 2020-04-16
Generic observables

KaiSzuttor:
- Added `src/particle_observables` for generic single particle observables
 - Use new observables lib in the core.


KaiSzuttor:
```non-portable path to file '<observables/Observable.hpp>'``` this is an issue because we now have two include paths that only differ by case. Shouldn't we avoid case-sensitive naming at all?

**Comment by mkuron** [comment]
> Shouldn't we avoid case-sensitive naming at all?

Yes, otherwise you can't build Espresso on Windows or macOS anymore.

File: src/core/observables/PidObservable.hpp [hunk 1]
```diff
@@ -69,6 +69,30 @@ template <class T> struct shape_impl<std::vector<T>> {
...
+  for (auto const &elem : vec) {
+    flatten(elem, out);
+  }
+}
+
+inline std::vector<double> flatten(Utils::Vector3d const &vec) {
+  return vec.as_vector();
+}
+
+template <class T> std::vector<double> flatten(std::vector<T> const &vec) {
```
fweik: This catches also a `std::vector<double>`, and then won't work. It solve this with SFINAE, a base case that just adds the value, and a specialization that loops and recurses. That would then also work for say `std::vector<Vector<Vector3>>`, e.g. a matrix.

fweik:
@KaiSzuttor the problem is that espresso does not have namespaces and proper unique corresponding include paths. With such generic names as `observables`, this is bond to have
collisions...

KaiSzuttor:
you are right, but I remember that we had issues with the case of filenames already (case-(in)sensitive filesystem)... google styleguide proposes all lowercase.

src/core/observables/PidObservable.hpp
KaiSzuttor: why wouldn't it work?

fweik: Ah maybe it would

fweik:
We do not follow the google style guide.

src/utils/include/utils/flatten.hpp
```diff
@@ -17,16 +17,29 @@ struct flatten_impl {
...
+template <class It> using value_type_t = typename it_value_type<It>::value_type;
 
 template <class T, class OutputIterator>
 struct flatten_impl<T, OutputIterator,
                     std::enable_if_t<std::is_convertible<
                         T, value_type_t<OutputIterator>>::value>> {
   static OutputIterator apply(T const &v, OutputIterator out) {
     *out = v;
-    return std::next(out);
+    return ++out;
```
fweik, line 44: How is this better than `next`?

KaiSzuttor:
@RudolfWeeber @fweik any suggestions for working around the naming conflict ```src/core/observables```  <-> ```libs/observables```.

fweik:
Yes, rename one of them.

fweik:
Do you plan to keep the observables as a submodule? Wouldn't it be easier to just add the code again to the Espresso repository?

KaiSzuttor:
yeah when it converged we can add it

src/core/observables/ParticleTraits.hpp
```diff
@@ -0,0 +1,39 @@
+#ifndef OBSERVABLES_PARTICLE_TRAITS
+#define OBSERVABLES_PARTICLE_TRAITS
+
+#include "Particle.hpp"
+#include "config.hpp"
+
+namespace GenObs {
+template <> struct traits<const ::Particle *> {
```
fweik: I don't think this should be cv qualified... In general I  find that specialization for a pointer type awkward. If you happen to not have a pointer, than ... you add an additional specialization?

src/core/observables/ParticleTraits.hpp
```diff
@@ -0,0 +1,39 @@
...
+  using Particle = ::Particle;
+
+  auto position(Particle const *const p) const { return p->r.p; }
```
fweik: If you deal with pointers, you have to check if they are null...

src/core/observables/PidObservable.hpp
```diff
@@ -49,5 +53,42 @@ class PidObservable : virtual public Observable {
   std::vector<int> const &ids() const { return m_ids; }
 };
 
-} // Namespace Observables
+namespace detail {
+template <class T> struct shape_impl;
+
+template <> struct shape_impl<double> {
+  static std::vector<size_t> eval(size_t /* n_part */) { return {1}; }
```
fweik, line 69: This size of the return type is statically known, so you could return an `array` instead of a `vector`

src/core/observables/PidObservable.hpp
```diff
@@ -49,5 +53,42 @@ class PidObservable : virtual public Observable {
...
+  using PidObservable::PidObservable;
+  std::vector<size_t> shape() const override {
+    using std::declval;
+
+    return detail::shape_impl<decltype(declval<ObsType>()(
+        declval<Utils::Span<const Particle *const>>()))>::eval(ids().size());
+  }
+
+  std::vector<double>
+  evaluate(Utils::Span<const Particle *const> particles) const override {
```
fweik: I think you should change the signature to taking a span of `std::reference_wrapper`. `reference_wrapper` is sometime a bit awkward to work with, but I think here it fits the bill, and it will make the downstream code clearer. With this you can push the null check up to the caller...

libs/genobs/include/genobs/observable.hpp
```diff
@@ -0,0 +1,42 @@
...
+ * not have state.
+ *
+ * @tparam Left left operand of the product.
+ * @tparam Right right operand of the product.
+ */
+template <class Left, class Right> struct Product : Left, Right {
+  Product(Left left = {}, Right right = {})
+      : Left(std::move(left)), Right(std::move(right)) {}
+
+  template <class Particle, class Traits = traits<Particle>>
```
fweik: I think you should add an indirection here, e.g. introduce a new meta function, maybe called `default_traits` that selects the traits class. Maybe

```c++
template<class Particle>
using default_traits = std::decay_t<traits<Particle>>;
```
in the traits header and then

```c++
template <class Particle, class Traits = default_traits<Particle>>
```
here

fweik (changes requested):
see comments

src/core/observables/PidObservable.hpp
KaiSzuttor: yeah but shape is an overridden function and the size of the array is not the same among the observables

src/core/observables/PidObservable.hpp
```diff
@@ -43,7 +43,7 @@ class PidObservable : virtual public Observable {
   std::vector<int> m_ids;
 
   virtual std::vector<double>
-  evaluate(Utils::Span<const Particle *const> particles) const = 0;
+  evaluate(Utils::Span<std::reference_wrapper<Particle>> particles) const = 0;
```
fweik: That should be `...<const Particle>` right?

fweik:
I'm not sure about the `std::reference_wrapper`, all the `get()` a little bit off-putting. What do you think?

KaiSzuttor:
maybe we should just use the traits everywhere? I also just realized that genobs should reside in src and not libs and that the tests are currently not run plus the given espresso test infrastructure should be used.

fweik:
Ok, maybe it's not even a bad thing if it's slightly inconvenient to use the particles directly. The physical structure needs to be discussed, but only after you have moved the code back to espresso. Then we also have to talk names...

src/core/observables/Current.hpp
```diff
@@ -32,16 +32,13 @@ class Current : public PidObservable {
...
     for (auto p : particles) {
-      double charge = p.get().p.q;
-      res[0] += charge * p.get().m.v[0];
-      res[1] += charge * p.get().m.v[1];
-      res[2] += charge * p.get().m.v[2];
+      current += (traits::charge(p) * traits::velocity(p));
     };
 #endif
-    return res;
+    return (Utils::hadamard_division(current, box_geo.length())).as_vector();
```
fweik: I don't think that this is a good solution, because you don't know what the correct volume is. If you imagine particles confined to a cylinder across the box (e.g. a 'wire'), with this definition the current would change if you change the empty box around your wire. I think the solution is probably to just not call it current, but find a more correct name.

KaiSzuttor: changed it back, you are right

fweik:
@KaiSzuttor since I was heavily involved in the design, you should probably get a review from somebody else.

src/core/observables/ParticleTraits.hpp
```diff
@@ -0,0 +1,37 @@
+#ifndef OBSERVABLES_PARTICLE_TRAITS
```
jngrad, line 19: missing copyright header

src/genobs/include/genobs/algorithms.hpp
```diff
@@ -0,0 +1,75 @@
+#ifndef ALGORITHMS_HPP
```
jngrad: missing copyright header

src/genobs/include/genobs/observable.hpp
```diff
@@ -0,0 +1,41 @@
+#ifndef INCLUDE_OBSERVABLES_OBSERVABLE_HPP
```
jngrad: missing copyright header

src/genobs/include/genobs/properties.hpp
```diff
@@ -0,0 +1,69 @@
+#ifndef OBSERVABLES_PROPERTIES_HPP
```
jngrad: missing copyright header

src/genobs/tests/algorithms.cpp
```diff
@@ -0,0 +1,49 @@
+#define BOOST_TEST_MODULE algorithms_test
```
jngrad: missing copyright header

src/genobs/tests/mock.hpp
```diff
@@ -0,0 +1,33 @@
+#ifndef TESTS_MOCK_HPP
```
jngrad: missing copyright header

src/genobs/tests/observables.cpp
```diff
@@ -0,0 +1,70 @@
+#define BOOST_TEST_MODULE observables_test
```
jngrad: missing copyright header

src/genobs/tests/properties.cpp
```diff
@@ -0,0 +1,24 @@
+#define BOOST_TEST_MODULE traits_test
```
jngrad: missing copyright header

src/utils/include/utils/flatten.hpp
```diff
@@ -0,0 +1,50 @@
+#ifndef ESPRESSO_UTILS_FLATTEN_HPP
```
jngrad, line 19: missing copyright header

src/genobs/include/genobs/observable.hpp
```diff
@@ -0,0 +1,41 @@
...
+};
+
+using Momentum = Product<Mass, Velocity>;
+using AverageMomentum = Average<Momentum>;
+using CenterOfMassPosition = WeightedAverage<Position, Mass>;
+using CenterOfMassVelocity = WeightedAverage<Velocity, Mass>;
+using TotalForce = Sum<Force>;
+using Positions = Map<Position>;
+using Velocities = Map<Velocity>;
+using Forces = Map<Force>;
```
jngrad: Should we write down a Doxygen page to give an overview of how to construct new observables in the core and how to integrate them with accumulators, correlators, and the ScriptInterface? What I have in mind is something similar to  http://espressomd.org/html/dox/script_interface.html

jngrad (dismissed):

src/genobs/include/genobs/observable.hpp
KaiSzuttor: the new observable library does not interfere with the ScriptInterface, we should probably document the ```ParticleObservable``` class. The naming is probably also not the best yet.

fweik: In principle you should replace `PidObservables`, then you could reuse the name.

src/core/observables/ParticleTraits.hpp
KaiSzuttor: is this a PR review criterium? This is checked before every release... actually we have tooling for that, so if we want to have it as a PR check, we could do that together with the other formatting checks.

jngrad: We should consider enforcing it in CI for the subset of C++/CUDA files. Tracking down the year range after the fact is tedious (the first year depends on whether the file was copied from another file, or restored from a reverted commit) and unnecessarily slows down the release workflow.

fweik:
@RudolfWeeber did you have time to look at this?

src/core/observables/BondAngles.hpp
```diff
@@ -36,13 +36,15 @@ class BondAngles : public PidObservable {
   using PidObservable::PidObservable;
 
   std::vector<double>
-  evaluate(Utils::Span<const Particle *const> particles) const override {
+  evaluate(Utils::Span<std::reference_wrapper<const Particle>> particles,
+           const GenObs::traits<Particle> &traits) const override {
```
RudolfWeeber: Maybe declare a ParticleRefSpan for Utils::Span<std::reference_wrapper<const Particle>> ?

src/core/observables/ParticleTraits.hpp
```diff
@@ -0,0 +1,55 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+#ifndef OBSERVABLES_PARTICLE_TRAITS
+#define OBSERVABLES_PARTICLE_TRAITS
+
+#include "Particle.hpp"
+#include "config.hpp"
+
+namespace GenObs {
+template <> struct traits<Particle> {
```
RudolfWeeber, line 31: Some sort of documentation is needed here, i.e., the purpose of the traits class and what we are using it for

src/core/observables/ParticleTraits.hpp
```diff
@@ -0,0 +1,55 @@
...
+  auto position(Particle const &p) const { return p.r.p; }
+  auto velocity(Particle const &p) const { return p.m.v; }
+  auto mass(Particle const &p) const {
+#ifdef VIRTUAL_SITES
+    if (p.p.is_virtual)
+      return decltype(p.p.mass){};
```
RudolfWeeber, line 38: Pls add comment explaining why this is done

src/core/observables/ParticleTraits.hpp
```diff
@@ -0,0 +1,55 @@
...
+#endif
+    return p.p.mass;
+  }
+  auto charge(Particle const &p) const { return p.p.q; }
+  auto force(Particle const &p) const {
+#ifdef VIRTUAL_SITES
+    if (p.p.is_virtual)
+      return decltype(p.f.f){};
```
RudolfWeeber: VS comment here as well

src/core/observables/PidObservable.cpp
```diff
@@ -49,10 +50,10 @@ std::vector<double> PidObservable::operator()() const {
     offset += this_size;
   }
 
-  std::vector<const Particle *> particles_ptrs(particles.size());
-  boost::transform(particles, particles_ptrs.begin(),
-                   [](auto const &p) { return std::addressof(p); });
-
-  return this->evaluate(particles_ptrs);
+  std::vector<std::reference_wrapper<const Particle>> particles_ptrs(
```
RudolfWeeber: Name seems to be out of date?

src/core/observables/PidObservable.hpp
```diff
@@ -49,5 +54,44 @@ class PidObservable : virtual public Observable {
   std::vector<int> const &ids() const { return m_ids; }
 };
 
-} // Namespace Observables
+namespace detail {
+template <class T> struct shape_impl;
```
RudolfWeeber, line 66: Please add a comment explaining the shape finding mechanism

src/genobs/include/genobs/algorithms.hpp
```diff
@@ -0,0 +1,93 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+#ifndef ALGORITHMS_HPP
+#define ALGORITHMS_HPP
+#include <algorithm>
+#include <numeric>
+#include <utility>
+#include <vector>
+
+namespace GenObs {
```
RudolfWeeber: Please add a file-level doxygen on what's in here and per-algorithm doxygen stating the requirements on the template params etc.

src/genobs/include/genobs/properties.hpp
```diff
@@ -0,0 +1,87 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+#ifndef OBSERVABLES_PROPERTIES_HPP
+#define OBSERVABLES_PROPERTIES_HPP
+
+namespace GenObs {
```
RudolfWeeber: Please add file-level doxygen on what's found here

src/genobs/include/genobs/properties.hpp
```diff
@@ -0,0 +1,87 @@
...
+template <class DoF> struct traits;
```
RudolfWeeber: What's DoF?

src/genobs/include/genobs/properties.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+namespace detail {
+template <class T> struct decay { using type = typename std::decay_t<T>; };
```
RudolfWeeber: What's this for?

RudolfWeeber (changes requested):
All good, except in-code documentation is needed. Please see individual comments.

fweik:
I don't like the name `genobs`. Shouldn't this just be called particle observables or similar?

src/genobs/include/genobs/properties.hpp
KaiSzuttor: type deduction for the traits... so you don't have to match pointer or ref to ```type``` because it will decay to ```type```

fweik: the `typename` is not needed I think

src/genobs/include/genobs/properties.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+template <class U> struct decay<std::reference_wrapper<U>> {
+  using type = typename std::decay_t<U>;
```
fweik: is `typename` needed?

src/genobs/include/genobs/properties.hpp
KaiSzuttor: @fweik?

fweik: A common abbreviation :-)

KaiSzuttor:
> I don't like the name genobs. Shouldn't this just be called particle observables or similar?

Naming is hard, renaming too ;)

fweik:
https://www.youtube.com/watch?v=EeGw32e5Em4

src/core/observables/PidObservable.hpp
```diff
@@ -33,6 +33,8 @@
 
 namespace Observables {
 
+using PartRefSpan = Utils::Span<std::reference_wrapper<const Particle>>;
```
fweik: Wouldn't a name like `ParticleReferenceRange` or `ConstParticleReferenceRange` that tells something about the concept be more useful, it would also be more in line with`ParticleRange` from core.

KaiSzuttor:
why does the integrator test fail? https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/230703
Can't see where I touch any related code.

fweik:
Timeout could be a technical problem. Did you check locally (with more than one core)?

KaiSzuttor:
ok CI is green... @RudolfWeeber you can have a look again

src/core/observables/ParticleTraits.hpp
```diff
@@ -0,0 +1,62 @@
...
+    if (p.p.is_virtual)
+      return decltype(p.p.mass){};
+#endif
+    return p.p.mass;
+  }
+  auto charge(Particle const &p) const { return p.p.q; }
+  auto force(Particle const &p) const {
+#ifdef VIRTUAL_SITES
+    if (p.p.is_virtual)
+      // we exclude virtual particles since their force does not have a meaning
```
RudolfWeeber: IMO, the force has a meaning. I'd suggest:
Forces on virtual sites are transferred to the base particle, where they contribute to the equation of motion. They are excluded on virtual sites to avoid double-counting.

RudolfWeeber:
Doc is good except the comment on VS forces. Once that's done, we can go ahead with this.

RudolfWeeber (dismissed):

src/core/observables/ParticleTraits.hpp
KaiSzuttor: Could you please explain the difference between mass and force concerning the meaning in the context of virtual sites? It's not clear to me why one of those should have a meaning for a virtual site and the other not. I see that there is a fundamental difference because on of them is changed by the integrator and the other not but still to write a useful comment I'd need your input.

src/core/observables/ParticleTraits.hpp
```diff
@@ -0,0 +1,62 @@
...
+  auto mass(Particle const &p) const {
+#ifdef VIRTUAL_SITES
+    // we exclude virtual particles since their mass does not have a meaning
+    if (p.p.is_virtual)
+      return decltype(p.p.mass){};
+#endif
+    return p.p.mass;
+  }
+  auto charge(Particle const &p) const { return p.p.q; }
+  auto force(Particle const &p) const {
```
fweik: I don't think that the force should be here, because it is not a property of a single particle but of a set of particles.

fweik (changes requested):
See inline comment on `ParticleTraits::force`

src/core/observables/ParticleTraits.hpp
KaiSzuttor: how is the force not a property of a single particle?

fweik: Because forces stem from interactions with other particles. There is another, more technical, difference, the forces depend on the system parameters while all the other properties do not.
It'd be nice if this was encoded in the type...

KaiSzuttor: > Because forces stem from interactions with other particles

Like velocities may.

fweik: I don't agree with this. From a given state point you can tell the velocity by only using the state of that particle, which is not true for forces. More fundamentally a particles does not 'have' a force,
that is a property of the system (e.g. of the Hamiltonian), otherwise physics would be pretty boring.

fweik: To put it differently, you can chose the velocity (and position for that matter) for any particle independently, but given a state of the system and a Hamiltonian, there is only one 'correct' set
of forces for particles, and in general this set depends on the set of the positions and velocities of all particles.

KaiSzuttor: That is even more complicated because we usually have a Hamiltonian where forces can depend on velocities... So what do you want to propose now?

fweik: That does not change anything. A force particle trait is not composable, because it can not be calculated per particle. That there is something related to forces stored on the particles is an
implementation detail, and leaking it leads to all kind of problems.

fweik: (Btw., I don't think that we can have a "Hamiltonian where forces can depend on velocities")

src/genobs/include/genobs/observable.hpp
fweik: I think before we can properly document this we have to clarify some concepts here with regards to thee `PidObservable`, to me the use of `PidObservable` for the bonds seems
slightly abusive, because it does not generalize to other particle filters (e.g. dihedrals between subsequent particlles of type 5 does not make sense). But this is out of scope. It would be nice to have a dockstring on `ParticleObservable` thou.

fweik (approved):

PR #3668, opened 2020-04-16
pitchfork

KaiSzuttor:
pitchfork library layout for observables

see: https://api.csswg.org/bikeshed/?force=1&url=https://raw.githubusercontent.com/vector-of-bool/pitchfork/develop/data/spec.bs#src


mkuron:
Does this layout have any benefit if there is only a single consumer of the library? Do you intend to connect some other MD engine to it? If not, it only makes the paths even longer by adding an extra "src/core" or "include/core".

jngrad:
The namespace change makes the observable code so much more readable, thanks! The directory substructure mirrors the namespace hierarchy, which makes the code more intuitive. I think the maintainability benefit outweighs the issue of filepath lengths.

jngrad (dismissed):

fweik:
I'm not sure if there is benefit in a partial transformation.

fweik (changes requested):
The namespace should be called `Core` according to the [naming convention](https://github.com/espressomd/espresso/wiki/Naming-conventions-and-folder-structure). Also then should also rename the cmake target to match the namespace and directory name.

fweik:
We can discuss if we want to change the convention for namespaces, but it should be consistent.

RudolfWeeber (changes requested):
I have mixed feelings about this.
* I understand that introducing the core namespace makes things simpler in conjunction with the script interface. But am I allowed to write `using namespace core` into core .cpp files, then? 
If yes, the namespaces can be introdcued from my side.

* The "core "namespace should be capitalized as the other namespaces, or all namespaces need to be converted to lower case. Personally I prefer uppercase. 

* I find the structure inversion (include/core/) confusing. 

* If we are changing structure at all, it needs to be done for the entire core at once. Otherwise, it's super-confusing.

KaiSzuttor:
> The "core "namespace should be capitalized as the other namespaces, or all namespaces need to be converted to lower case. Personally I prefer uppercase.

The reason for having it lowercase is the direct correspondence to the path (which should be all lower-case). What's the reason for your uppercase preference?

> I find the structure inversion (include/core/) confusing

We already have this structure for utils and is the standard layout if we want to follow the above mentioned layout.

KaiSzuttor:
> I find the structure inversion (include/core/) confusing

Also this gives you the benefit of expressive include paths ala ```#include "core/something.hpp"```.

RudolfWeeber:
>The reason for having it lowercase is the direct correspondence to the path (which should be all lower-case). What's the reason for your uppercase preference?

I prefer uppercase, for better distinction from variables. But either all uppercase or all lowercase for all namespaces in the project. I see no benefit in matching the case to the case of the directory.

 

 

>>I find the structure inversion (include/core/) confusing

>We already have this structure for utils and is the standard layout if we want to follow the above mentioned layout.

I don’t think we should follow the pitchfork specification (standard is too strong a word, I think). I was too slow to interfere with the utils library, but I don’t think it should be applied further.

I neither like the inversion nor the long path names.

 

 

As I said, I’d be fine with a Core namespace, as long as the capitalization of namespaces throughout the project is consistent.

RudolfWeeber:
>gives you the benefit of expressive include paths ala #include "core/something.hpp".

This is already the case. See stuff under scriptinterface/observables/

KaiSzuttor:
> I don’t think we should follow the pitchfork specification (standard is too strong a word, I think). I was too slow to interfere with the utils library, but I don’t think it should be applied further.

Could you then please propose another specification for the source code layout?

RudolfWeeber:
>>I don’t think we should follow the pitchfork specification (standard is too strong a word, I think). I was too slow to interfere with the utils library, >but I don’t think it should be applied further.

>Could you then please propose another specification for the source code layout?

Header files stored with the source files, subdirectories for the logical components. I.e., mostly, what’s there for the core at present.

I think your last PR restructuring the core provided the necessary clarity. 

 

I think the Core namespace might be worth pursuing, but it would have to be done throughout the core, then.

fweik:
@RudolfWeeber I agree that the partial application just adds to confusion, as I wrote above. 

The benefit of this layout to be is twofold:  It allows the separation of exported from internal headers (either `include` or `src`), and it gives "namespacing" to the header filenames, e.g. the names in different components can not collide.

Due to the way the lookup for includes works a structure like pitchfork is required, as long we don't have modules. Contrary to what @mkuron feared above the paths do not enter your actual code except for the consumers of the component,
but the `include` directory is exported to the consumers via the cmake target, the `src` and `src/core` is added as a private include directory within the target. E.g. the include statements within the component don't have to change, for consumers, the component name has to be added, which is kind of the point.

I agree that having some of the headers split from the source files is not optimal, but in C++ current compilation model this is the only way physical encapsulation can be achieved. I also don't think
that it is particularly confusing as long as it is consistent, especially because this is somewhat the essence of C++ library design, and fairly standard.

KaiSzuttor:
> But am I allowed to write using namespace core into core .cpp files, then?

Why would you need that? @RudolfWeeber

KaiSzuttor:
this layout is also present in ```src/shapes```

KaiSzuttor:
there seems to be a need for discussion first

PR #3673, opened 2020-04-20
Update documentation and clarify ambiguities

### LINKED ISSUE #3633: Charge unit in espresso
While going through the page on electrostatics, I realized that the prefactor should be KT*lB/(eo**2) instead of kT*lB as suggested by the package. Most example programs have mentioned the prefactor to be 2 and the charges to be integers.  How do these charges connect to the electronic charge?

### LINKED ISSUE #3656: H5md feature unavailable
How does one go about fixing the unavailability of H5md? My espressomd.io.writer seems to have the vtf feature available in it. So, I am not sure where the error is coming from. A look at the source code for espressomd.io.writer.h5md suggested that 'H5md' is not part of the features of code_info. Any way out of this? 

    h5_file=h5md.H5md(filename="systeminfo.h5", write_pos=True, write_vel=False, write_force=False, write_species=True, write_mass=False, write_charge=True, write_ordered=True)
  File "/home/dpo854/mycode/espresso/install/lib/python3.4/site-packages/espressomd/io/writer/h5md.py", line 29, in __init__
    raise RuntimeError("H5md not available.")
RuntimeError: H5md not available.

jngrad:
Description of changes:
- remove CUDA installation instructions on OSX (follow-up to #3652)
- update links to dockerfiles for non-Ubuntu OSes
- correct the Coulomb prefactor formula (fixes #3633)
- in electrostatics scripts, replace magic values by their full expressions
- document the difference between serial and parallel versions of `libhdf5-dev` (fixes #3656)
- improve documentation of the script that shows the effect of MPI and `node_grid` on domain decomposition (see mailing list email "dividing simulation box using MPI" from 2020-04-20)


doc/sphinx/electrostatics.rst
```diff
@@ -4,19 +4,21 @@ Electrostatics
 ==============
 
 The Coulomb (or electrostatic) interaction is defined as
-follows. For a pair of particles at distance :math:`r` with charges
-:math:`q_1` and :math:`q_2`, the interaction is given by
+follows. For a pair of particles at distance :math:`r` with formal charges
+:math:`z_1` and :math:`z_2`, the electric potential energy is given by
```
fweik: I don't think that `z` is a good symbol here, as it is typically used for a related but different quantity. Also I don't think `formal charge` is a common term. What is wrong with `q`?

doc/sphinx/electrostatics.rst
```diff
@@ -4,19 +4,21 @@ Electrostatics
...
 
-.. math:: U_C(r)=C \cdot \frac{q_1 q_2}{r}.
+.. math:: U_C(r)=C \cdot \frac{z_1 z_2}{r}.
 
 where
 
 .. math::
-   C=\frac{1}{4\pi \varepsilon_0 \varepsilon_r}
+   C=\frac{e^2}{4\pi \varepsilon_0 \varepsilon_r}
```
fweik: I don't believe this to be correct. I don't think that any of this is related to the elementary charge (which is a constant of nature, one absolute amount of charge, whose numerical value depends on the unit system). I rather think that the charges are measured in whatever the charge units are.

fweik (changes requested):

doc/sphinx/electrostatics.rst
jngrad: * isn't `z_i` the standard notation for the charge number of ions?
* the Coulomb prefactor was missing the quantity `e^2` to be consistent with the definition of the Bjerrum length in the next paragraph

This change was discussed in https://github.com/espressomd/espresso/issues/3633#issuecomment-609725762

fweik: >  isn't z_i the standard notation for the charge number of ions?

Yes, but in the formula its charges.

> the Coulomb prefactor was missing the quantity e^2 to be consistent with the definition of the Bjerrum length in the next paragraph

The Coulomb prefactor is not related to e^2 in any way.

fweik: Also it wasn't 'discussed' in the cited comment, Jonas just claimed something :-)

doc/sphinx/electrostatics.rst
jngrad: We're missing an `e^2` somewhere to make these equations consistent. The issue here is about definition. You probably think of the Coulomb prefactor `C` as the Coulomb constant `k_e = 1/ (4 * pi * e0)`, which is also what I intuitively assumed the first time I read the text, but the text actually defines `C` with regards to the Bjerrum length.

fweik: Well the interaction energy in espresso is `C * q_1 * q_2 / r`, independently of what the text defines...

jngrad: We could rewrite the text, starting from the Coulomb equation `U_C = k_e * q1 * q2 / r` with `k_e = 1/ (4 * pi * e0)` and then introduce the ESPResSo convention where `U_C = C * z1 * z2 / r` with `C` defined from the Bjerrum length. Most ESPResSo users prefer to put an integer charge (aka charge number) on their coarse-grained particles and delegate the actual units to the prefactor `C`. This would clarify the ESPResSo convention while remaining in line with physics textbooks.

fweik: I find it odd to push a unit system here, frankly. And charge numbers are not a concept that exists in electrodynamics. If you particles happen to not be ions, `e` is by no means a natural charge unit.

jngrad: Then we either write `C = l_B * k_BT / e^2`, or we drop the sentence about the Bjerrum length.

fweik: I would do the former but this is also not a hill I'm willing to die on. I just find the idea that quantities have specific natural unit odd and mislead in general, and confusing in Espresso in particular because it does not have a fix unit system.

samples/visualization_cellsystem.py
```diff
@@ -16,7 +16,12 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 Visualize the system cells and MPI domains. Run ESPResSo in parallel
-to color particles by node.
+to color particles by node. With OpenMPI, this can be achieved using
+``mpiexec -n 4 ./pypresso ../samples/visualization_cellsystem.py``.
+Set property ``system.cell_system.node_grid = [n_x, n_y, n_z]``
+(with ``n_x * n_y * n_z`` equal to the number of MPI threads) to
```
fweik: `MPI threads` shoudl be `MPI processes` or `MPI ranks`.

fweik (changes requested):

doc/sphinx/electrostatics.rst
jngrad: > `e` is by no means a natural charge unit

This is a valid argument, how about introducing z_i as "normalized charges" `z_i = q_i / e` instead of formal charges?

fweik: I'm not clear still on why they have to be introduced at all.

fweik: Why is there da "definition" for `C` at all, when right after it says that is a settable parameter?
So I'd do the following. Drop the definition of `C` as such, then two examples which `C` to use for a relative permittivity, and for a specific Bjerrum length?

fweik:
Sorry if this was a little bit confused, I think I can give a better explanation. Espresso implements an interaction of the form `C * q_1 * q_2 / r`, whose interaction strength is given by `C`. For a specific
value of `C` you can interpret this a electrostatics, e.g. if you derive this from a permittivity, it might
be the effective electrostatics in a medium. The Bjerrum length is just an odd/convenient way to state `\epsilon_r`, if you assume vacuum electrostatics of our universe (which Espresso does not force you to). `C` has units of [E] / ([L] * [Q]^2) so if you change the charge unit, the numbers for `C` and the charges changes, while the product stays the same. The energy is independent of the charge unit, as it should be.

Of course you can always move a charge factor from the `q` to `C`, but since you might be in some coarse grained world in your simulation, which does not deal in ions, but e.g. in dust particles, there
is no concept of a charge number and it is artificial to introduce it.

doc/sphinx/electrostatics.rst
jngrad: It would be simpler to just rewrite `C=l_B k_B T / e^2` to fix the inconsistency. As you said above and in https://github.com/espressomd/espresso/pull/3673#issuecomment-616797265, the user can move `q` around for convenience in simulations with ions, without changing the energy.

jngrad:
> The energy is independent of the charge unit, as it should be.

Yes, the user just needs to be careful with dimensional analysis when setting up `C` and the charges. The tutorials and samples tweak the value of `C` to actually have an energy in kJ/mol, such that `system.analysis.energy()` gives an energy with that unit. As long as the same units are used in the WCA potential and thermostat, it should be fine.

fweik (approved):
Sorry for the many words, I feel we could have gotten there quicke if I'd be more to the point ^^ LGTM now

doc/sphinx/electrostatics.rst
jonaslandsgesell: ;-) You can do the math and check. Without the adaptions the unit of energy is not correct - at least if you stick with typical conventions that q_i is a charge that can be measured in multiples of elementary charges. If you choose another convention then this may not have to be the case as you point out below. However, in such a case it would be nice to read which convention you follow instead.

PR #3677, opened 2020-04-21
All_compare test: Expect "true" on single process

hirschsn:
The all_compare test checks if all_compare correctly returns false if supplied with different values. This, however, only holds true if the test is called with more than one processes.


fweik (approved):

fweik:
Thanks!

PR #3678, opened 2020-04-21
LBCPU: Fix rounding issue for grid size calculation

RudolfWeeber:
Description of changes:
Replaces ceil by round in lbcpu n*agrid ~= box_l calculation + a test in which the old behavior resulted in a false positive


fweik:
The `Lattice` class is a gift that just keeps on giving...

fweik (dismissed):

fweik (approved):

Issue #3611, opened 2020-03-31
CI Cuda Jobs


RudolfWeeber:
This is the suggested list of GPU CI jobs to run in the future:
* Ubuntu 20.04 with Ubuntu-supplied Cuda 10.1
* Ubuntu 18.04 with Ubuntu-supplied Cuda 9.1
* NVidia supplied Ubuntu with Cuda-latest (currently 10.2)
* OSX with Cuda 10.2
* Ubuntu 20.04 with Rocm (3, I suppose?)

This ticket only pertains to GPU jobs. Everything else CI/docker is a different matter

mkuron:
We can drop CUDA on OSX. Apple hasn't shipped Macs with Nvidia GPUs since the Kepler generation and CUDA 10.2 is the last version to support Kepler.

jngrad:
Here's the status:
* Ubuntu 20.04 with Ubuntu-supplied Cuda 10.1 -> done, we can drop our `cuda10.1` image
* Ubuntu 18.04 with Ubuntu-supplied Cuda 9.1 -> done, we can drop our `cuda9.2` image
* Ubuntu 20.04 with ROCm -> base image `dev-ubuntu-20.04` not yet available
* Nvidia-supplied Ubuntu with Cuda-latest -> base image `nvidia/cuda:10.2-devel-ubuntu20.04` not yet available
* OSX with Cuda 10.2 -> removed
* Ubuntu 16.04 with Cuda 9 and Clang 6 -> removed

jngrad:
Concerning the NVidia-supplied Ubuntu image, we run into the following CMake issue when looking for CUDA package cublas with `find_library(CUDA_CUBLAS_LIBRARIES NAMES cublas PATHS ...)`:
```
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
CUDA_cublas_device_LIBRARY (ADVANCED)
    linked by target "sd_gpu" in directory /home/espresso/espresso/libs/stokesian_dynamics/src

-- Configuring incomplete, errors occurred!
```
[...]

PR #3684, opened 2020-04-24
Move Clang-Tidy and sanitizers to Clang 9 in Ubuntu 20.04

### LINKED ISSUE #3627: Move clang-tidy, ASAN, UBSAN to clang-9
Results so far:
Build with Clang9 on the ICP machines:
* Clang++ cannot be used for the Cuda code, unless we are prepared to patch thrust. The code makes use of `va_printf` which is not defined (https://github.com/thrust/thrust/issues/1032). Alternatively, we could include the workaround discussed in the isseu in Espresso. Personally, I'd just use Nvcc.
* The following clang-tidy checks get triggered due to a more modern clang-tidy:

    * bugprone-exception-escape,
    * bugprone-macro-parentheses,
    * bugprone-narrowing-conversions,
    * bugprone-unhandled-self-assignment,
    * modernize-avoid-c-arrays,
  
  As per the discussion on the Espresso meeting, I'll disable them. If we want, we can address (part of) them separately.
* The ASAN needs ASAN_OPTIONS="allocator_may_return_null=1" to avoid an error from an OpenMPI library

### LINKED ISSUE #3650: Clang++ 9.0 and boost::optional rebind semantics in CUDA code
I get a compiler error on `src/core/grid_based_algorithms/lbgpu_cuda.cu` with Clang++-9.0, boost 1.71 and CUDA 10.1 from `nvidia-cuda-toolkit` on Ubuntu 20.04. The two offending functions are:
https://github.com/espressomd/espresso/blob/48ba8d16152ca44e09cb2da2a03b2e4d99665a4b/src/core/grid_based_algorithms/lbgpu_cuda.cu#L3042-L3048

According to the [boost::optional docs](https://www.boost.org/doc/libs/1_71_0/libs/optional/doc/html/boost_optional/rebinding_semantics_for_assignment_of_optional_references.html), we should be able to rebind an optional using the assignment operator, even if that optional is already initialized. The following MWE compiles just fine as a .cpp file, but fails as a .cu file:
```c++
#include <boost/optional.hpp>
template <typename T>
class A {
public:
  A(T &a) : x(a) {}
  T x;
};
boost::optional<A<int>> rng_counter_coupling_gpu;

void foo(int a) {
  rng_counter_coupling_gpu = A<int>(a);
}
```
```
$> clang++-9 -pedantic -fPIC -std=c++14 -c mwe.cpp
$> clang++-9 -pedantic -fPIC -std=c++14 --cuda-gpu-arch=sm_52 -c mwe.cu
mwe.cu:11:28: error: use of overloaded operator '=' is ambiguous (with operand types 'boost::optional<A<int> >' and 'A<int>')
  rng_counter_coupling_gpu = A<int>(a);
  ~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~
/usr/include/boost/optional/optional.hpp:993:5: note: candidate function [with Expr = A<int>]
    operator= ( Expr&& expr )
    ^
/usr/include/boost/optional/optional.hpp:1064:5: note: candidate function [with T_ = A<int>]
    operator= ( T_&& val )
    ^
/usr/include/boost/optional/optional.hpp:1047:15: note: candidate function
    optional& operator= ( optional && ) = default;
              ^
/usr/include/boost/optional/optional.hpp:1035:15: note: candidate function
    optional& operator= ( optional const& rhs ) = default;
              ^
1 error generated when compiling for sm_52.
```
It's not reproducible on the ICP machines with `clang++-9 -pedantic -fPIC -std=c++14 --cuda-gpu-arch=sm_52 --cuda-path=/usr/local/cuda-10.0 -c mwe.cu`. I had to start `docker run --user espresso --runtime=nvidia espressomd/docker-ubuntu-20.04:01cf23487c9eeec758daf425c9947d8089ed9da0 bash` and install `nvidia-cuda-toolkit`.

Making `foo()` check the initialization status of the optional doesn't help:
```c++
void foo(int a) {
  if(rng_counter_coupling_gpu) {
    *rng_counter_coupling_gpu = A<int>(a);
  } else {
    rng_counter_coupling_gpu = A<int>(a);
  }
}
```
Making `foo()` always derefence is wrong, but compilation of the .cu file finally succeeds:
```c++
void foo(int a) {
  *rng_counter_coupling_gpu = A<int>(a);
}
```
It's wrong because dereferencing an un-initialized boost optional will cause an assertion to throw, and this is indeed what happens if you do that in a .cpp file with `int main() { foo(1); }`. Yet ESPResSo compiles with this incorrect workaround in `lbgpu_cuda.cu`, and all the tests pass.

What am I missing here?

jngrad:
Fixes #3627, fixes #3650, closes #3636

Description of changes:
- run Clang-Tidy, ASAN, UBSAN in the Ubuntu 20.04 image
- re-enable Clang-Tidy on CUDA code
- fix Clang-Tidy 9 warnings
- replace `boost::optional` by custom code in CUDA sources
- remove unused code


jngrad:
@RudolfWeeber According to the Clang-Tidy 9 job [logfile](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/226004), it took 81 min to run, of which around 18 min were spent pulling the image from GitHub. Hopefully, as we increase usage of that image, it will remain in the runners cache and take as long to run as the Clang-Tidy 6 image. To that effect, I will in an upcoming PR replace the `cuda10.1` image by the Ubuntu 20.04 image (#3611).

From my side, we can now remove the `clang:6.0` job and switch the `empty` job from Clang 6 to Clang 9. After this PR is merged, I'll further simplify CI images by removing the Clang 6 image (it's still required at the moment for clang-format in the `style` job, I overlooked that packages in the Ubuntu 20.04 image) and merging the `cuda10.1` and Ubuntu 20.04 images together. There will be other considerations when doing that change, so let's limit this PR to the Clang-Tidy change.

You can now run static analysis on an ICP machine. There is a slight discrepancy in the version number, Ubuntu 20.04 has the bugfixed version 9.0.1 while ICP machines have 9.0.0. There are `boost::mpi` warnings, but they won't cause a compiler error and will disappear once we migrate ICP machines to Ubuntu 20.04, which has a more recent Boost version where the issue was fixed. You'll still need to tell Clang where to find the CUDA library, since ICP machines have multiple ones installed and Clang selects the first one available, aka CUDA 8.0. The CUDA 10.0 we have still has thrust 1.9.3 which has the `va_printf` bug, so you'll need to either copy the library somewhere and change thrust to 1.9.5, or install CUDA 10.1 in `/usr/local`; this inconvenience will also be resolved when we migrate to Ubuntu 20.04. Here is the command line I use to run Clang-Tidy locally without docker:
```sh
CC=clang-9 CXX=clang++-9 cmake .. -DWITH_CUDA=ON -DWITH_CUDA_COMPILER=clang \
  -DWITH_CLANG_TIDY=ON -DCMAKE_CXX_FLAGS=--cuda-path=/work/jgrad/cuda-10.0-thrust195
```

jngrad:
Nevermind, `clang-format-9` is already in the Ubuntu 20.04 image. We can drop the `clang:6.0` image in this PR.

src/core/BondList.hpp
```diff
@@ -171,7 +171,8 @@ class BondList {
 
     return *this;
   }
-  BondList &operator=(BondList &&rhs) noexcept {
+  // NOLINTNEXTLINE(bugprone-exception-escape)
+  BondList &operator=(BondList &&rhs) {
```
fweik: This should probably be fixed, because it has performances effects if `BondList` is nothrow-move-assignable... Do you remember what the problem here was?

src/core/BondList.hpp
```diff
@@ -224,6 +225,7 @@ class BondList {
    */
   bool empty() const { return m_storage.empty(); }
 
+  // NOLINTNEXTLINE(bugprone-exception-escape)
```
fweik, line 228: Same here as in L175

src/core/BondList.hpp
jngrad: I accidentally removed the `noexcept`, good that you spotted that.

jngrad:
```
/home/espresso/espresso/src/core/BondList.hpp:174:13: error: an
exception may be thrown in function 'operator=' which should not
throw exceptions [bugprone-exception-escape,-warnings-as-errors]
  BondList &operator=(BondList &&rhs) noexcept {
            ^
/home/espresso/espresso/src/core/BondList.hpp:227:15: error: an
exception may be thrown in function 'swap' which should not throw
exceptions [bugprone-exception-escape,-warnings-as-errors]
  friend void swap(BondList &lhs, BondList &rhs) {
              ^
```

fweik: Ah, probably because the respective operations in the boost container can throw. I think since there is no performance benefit from them at the moment we could just switch the backend to `std::vector`. I'm a bit uncomfortable about all the linter exceptions...

src/utils/include/utils/mpi/gatherv.hpp
```diff
@@ -61,6 +61,7 @@ void gatherv_impl(const boost::mpi::communicator &comm, const T *in_values,
 
     /* not in-place */
     if (in_values != out_values) {
+      // NOLINTNEXTLINE(clang-analyzer-core.NullDereference)
```
fweik: This can be easily fiixed: `if(in_value && out_value && in_values != out_values)`

src/core/electrostatics_magnetostatics/scafacos.cpp
```diff
@@ -230,7 +229,7 @@ double long_range_energy() {
              std::inner_product(particles.charges.begin(),
                                 particles.charges.end(),
                                 particles.potentials.begin(), 0.0);
-    } else {
+    } else { // NOLINT(readability-else-after-return)
```
fweik: I think this can be easily fixed remove the `else {}`

src/core/electrostatics_magnetostatics/scafacos.cpp
```diff
@@ -269,26 +268,26 @@ void tune_r_cut() {
...
+  /* Run bisection */
   while (std::fabs(r_min - r_max) > tune_limit) {
     const double dr = 0.5 * (r_max - r_min);
     const double t_mid = time_r_cut(r_min + dr);
     t_min = time_r_cut(r_min);
     t_max = time_r_cut(r_max);
 
     if (t_min <= 0.0) {
-      r_min += tune_limit;
+      r_min += tune_limit; // NOLINT(clang-analyzer-deadcode.DeadStores)
```
fweik: I think this _is_ a dead store, so it can be removed

src/core/electrostatics_magnetostatics/scafacos.cpp
```diff
@@ -269,26 +268,26 @@ void tune_r_cut() {
...
       break;
     }
 
     if (t_max <= 0.0) {
-      r_min -= tune_limit;
+      r_min -= tune_limit; // NOLINT(clang-analyzer-deadcode.DeadStores)
```
fweik: I think this _is_ a dead store, so it can be removed

src/core/grid_based_algorithms/OptionalCounter.hpp
```diff
@@ -0,0 +1,64 @@
...
+ * ESPResSo is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
```
fweik: We are a header guard code base, also `#pragma once` has it's own set of problems...

src/core/grid_based_algorithms/OptionalCounter.hpp
```diff
@@ -0,0 +1,64 @@
...
+ *  and CUDA 10.1 (see espressomd/espresso#3650).
+ */
+class OptionalCounter {
+private:
+  Utils::Counter<uint64_t> m_counter;
+  bool m_initialized;
+
+public:
+  OptionalCounter() : m_counter{}, m_initialized(false) {}
+  OptionalCounter(Utils::Counter<uint64_t> &counter)
```
fweik: `const &`

src/core/BondList.hpp
jngrad: Me too, but I only added these `NOLINT` in places where I couldn't find a quick fix. Hopefully we can address them in the future.

src/core/grid_based_algorithms/OptionalCounter.hpp
```diff
@@ -0,0 +1,64 @@
...
+      : m_counter(counter), m_initialized(true) {}
+  OptionalCounter &operator=(Utils::Counter<uint64_t> &counter) {
```
fweik: `const &` or take by value, in which case you can drop the self-assignment check

src/core/grid_based_algorithms/OptionalCounter.hpp
```diff
@@ -0,0 +1,64 @@
...
+    if (&m_counter != &counter) {
+      m_counter = counter;
+      m_initialized = true;
+    }
+    return *this;
+  }
+  OptionalCounter &operator=(Utils::Counter<uint64_t> &&counter) {
```
fweik: This has no benefit over the copy version and can be dropped.

src/core/grid_based_algorithms/lbgpu_cuda.cu
```diff
@@ -1173,9 +1173,11 @@ calc_values_in_LB_units(LB_nodes_gpu n_a, Utils::Array<float, 19> &mode,
   } else {
     d_p_v[print_index].rho = 0.0f;
 
+    // NOLINTNEXTLINE(modernize-loop-convert)
```
fweik: `boost::fill(d_p_v[print_index].v, 0.f)` ?

src/core/grid_based_algorithms/lbgpu_cuda.cu
```diff
@@ -1173,9 +1173,11 @@ calc_values_in_LB_units(LB_nodes_gpu n_a, Utils::Array<float, 19> &mode,
...
     for (int i = 0; i < 3; i++)
       d_p_v[print_index].v[i] = 0.0f;
 
+    // NOLINTNEXTLINE(modernize-loop-convert)
```
fweik: See above

.clang-tidy
```diff
@@ -36,7 +36,7 @@ Checks: |
    readability-string-compare,
    readability-uniqueptr-delete-release,
    readability-function-size'
-WarningsAsErrors: '*'
+WarningsAsErrors: '*,-clang-analyzer-core.StackAddrEscapeBase,-clang-analyzer-optin.mpi.MPI-Checker'
```
fweik, line 39: So the conclusion was that there is no smaller gun than excluding `core.StackAddrEscapeBase`? This is really unfortunate because this is almost always a bug, as it is in the place where the warning is caused in boost mpi...

src/core/BondList.hpp
fweik: Make `noexcept`

fweik (changes requested):
That was a lot of work, thank you! I looked over the changes, and most of them is fairly uncontroversial.
I've marked a few place where I think a linter exception can be avoided by a reasonably simple change. In general I don't think we should use that many exceptions, but linter exceptions should be a last resort in places where there is a false positive or the broken code is beyond our control e.g. in a library. (the latter means that we are knowingly shipping broken code, which is what a linter exception means in such a case).

.clang-tidy
mkuron: This should already be fixed by the Boost version shipped in Ubuntu 20.04.

src/core/electrostatics_magnetostatics/scafacos.cpp
jngrad: Obviously, but could it be a bug, e.g. someone accidentally adding a `break;`? I looked into the history, but it has been like this for several years.

.clang-tidy
jngrad: Yes it's unfortunate, but Clang-Tidy offers no alternative for this warning. A conflicting priority here is that we need to be able to compile the code with the dev environment of the ICP machines, where we have the old Boost version (until we migrate to Ubuntu 20.08).

src/core/electrostatics_magnetostatics/scafacos.cpp
fweik: No I think it's fine, the `r_{min, max}` are only used internally for the bisection, which is terminated by the `break`. Adding the linter exception makes it look as if this was there on purpose, which will make somebody in the future spend time figuring out why this jean-noel dude in 2020 added an linter exception to keep a dead store :-)

.clang-tidy
fweik: Ok, @mkuron it's not sufficient that this is fixed on ubuntu 20.04, it needs to be fixed in all supported boost versions, which is not the case.

mkuron: > need to be able to compile the code with the dev environment of the ICP machines

Do you depend on warnings-as-errors there? As a stop-gap fix, you might just patch the header in /usr/include/boost locally for the time being. Frank already set up a test install of Ubuntu 20.04 today and it's looking pretty good, so upgrading the ICP is probably more a matter of weeks rather than months.

> it needs to be fixed in all supported boost versions, which is not the case

Does it? End users don't do clang-tidy. Since we require a specific version of clang-tidy anyway (whatever version is installed in our Docker container at any given time is likely the only version that gives zero warnings for the code base at the same time), it doesn't matter if we need a certain Boost version to get zero warnings. Other versions of clang-tidy usually give small but non-zero numbers of warnings. That may even already be the case for other versions of boost.

jngrad: I agree this warning provides valuable information but we will have to wait for ICP computers to migrate to Ubuntu 20.04 before we can revert this change.

fweik (approved):
LGTM

.clang-tidy
fweik: If you stilll want to have a discussion about this, please open a ticket. I think we should go forward with this PR now.

PR #3690, opened 2020-04-27
Removed remaining globals from ghosts

fweik:
Description of changes:
- Removed dependence of the ghost comm on global state
- Made ghosts operate on `ParticlelList` instead of `Cell`, which is all they need to know.
- Removed some left-over management functions that are no longer needed because cells/ghost communications are proper types now.


jngrad (approved):
LGTM

PR #3696, opened 2020-04-30
Removed communication and geometry globals from domain decomposition

fweik:
If I wouldn't know better I'd say that this code was specifically crafted to be hard to refactor.

Description of changes:
- Removed geometry globals from dd
- Removed communication globals from dd
- Modified dd so that the periodicity does not change the communication pattern
- Treat periodicity change as geometry change
- Moved `get_mi_*` stuff to `BoxGeometry`

### DIFF
--- a/src/core/event.cpp
+++ b/src/core/event.cpp
@@ -230,7 +230,7 @@ void on_coulomb_change() {
 void on_short_range_ia_change() {
   invalidate_obs();
 
-  cells_on_geometry_change(0);
+  cells_on_geometry_change(false);
 
   recalc_forces = true;
 }
@@ -256,7 +256,7 @@ void on_boxl_change() {
   grid_changed_box_l(box_geo);
   /* Electrostatics cutoffs mostly depend on the system size,
      therefore recalculate them. */
-  cells_on_geometry_change(0);
+  cells_on_geometry_change(false);
 
 /* Now give methods a chance to react to the change in box length */
 #ifdef ELECTROSTATICS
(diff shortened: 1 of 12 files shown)


src/core/event.cpp
```diff
@@ -325,8 +325,7 @@ void on_parameter_change(int field) {
     break;
   case FIELD_NODEGRID:
     grid_changed_n_nodes();
-    cells_on_geometry_change(CELL_FLAG_GRIDCHANGED);
-    break;
+    cells_re_init(CELL_STRUCTURE_CURRENT, cell_structure.min_range);
   case FIELD_MINNUMCELLS:
   case FIELD_MAXNUMCELLS:
     cells_re_init(CELL_STRUCTURE_CURRENT, cell_structure.min_range);
```
jngrad, line 327: is this fallthrough intentional? `cells_re_init()` is now called twice with the same arguments if case `FIELD_NODEGRID` matches. Maybe you meant to delete L328?

src/utils/include/utils/mpi/cart_comm.hpp
```diff
@@ -121,6 +121,38 @@ cart_neighbors(const boost::mpi::communicator &comm) {
 
   return ret;
 }
+
+/**
+ * @brief Information about a cartesian communicator.
+ *
+ * Memebers coorespond to the output arguments of
```
jngrad:
```suggestion
 * Members correspond to the output arguments of
```

jngrad (changes requested):
There are several parts of the ghost communication code where I'll trust you :)

src/core/event.cpp
fweik: Ah yes, good catch.

fweik:
It's clear to me that this is a little bit of a leap of faith, but it has to be done to get that under control. We can then subsequently find out what the actual pre- and postconditions are and add tests. Also all of this has fairly good coverage by the integration tests.

jngrad (approved):

PR #3698, opened 2020-05-01
Factored out atom decomposition

fweik:
Description of changes:
- Made atom decomposition (that is the proper name for the n-squared cell system) 
  self-contained


jngrad (approved):
LGTM

PR #3699, opened 2020-05-01
Code cleanup

jngrad:
Description of changes:
- simplify handling of UBSAN and ASAN environment variables in CI and CMake
- better explain GPU-related CMake options and variables
- cleanup header includes
- cleanup documentation
- delete unused code, specfile and old UBSAN suppression file


jngrad:
* several commits need to be cherry-picked in 4.1.3
* @fweik there should be no overlap with your atom decomposition PR

src/utils/include/utils/print.hpp
```diff
@@ -1,42 +0,0 @@
...
-
-/**
- * @brief Python style print function.
- */
-template <typename T> void print(T v) { std::cout << v << '\n'; }
-
-/**
- * @brief Python style print function.
- */
-template <typename Arg, typename... Args> void print(Arg v, Args... args) {
```
fweik, line 35: Can I keep that one? I use that for debugging...

src/core/actor/DipolarBarnesHut.hpp
```diff
@@ -30,6 +30,9 @@
 #include "errorhandling.hpp"
 #include "grid.hpp"
 
+#include <cstdio>
```
fweik: I think the method should not directly write output in the first place. Since `errorhandling` is already included, I think it would be better to just replace the printing by a runtime error. `iostream` is a _very_ large header, so reexporting has a negative impact on compile time. Same for direct sum...

fweik (dismissed):
LGTM, I made two comments, where you can decided if you want to address them, but you can also merge as is.

src/utils/include/utils/print.hpp
jngrad: I can restore it if you need it, but if the file is no longer part of the git history, you could also keep an untracked copy. It'll not be added if you use `git add -u src` and rebase operations will ignore it. I have quite a few such untracked files lying around in my clone and never had an issue with them. You could add a commit hook just to be sure it's not getting accidentally added.

src/core/actor/DipolarBarnesHut.hpp
jngrad: I'll have a look.

src/utils/include/utils/print.hpp
jngrad: @fweik you marked it as resolved because you'll use the untracked copy?

fweik (approved):

PR #3701, opened 2020-05-01
Removed CellSystem.max_num_cells and CellSystem.min_num_cells

fweik:
After offline discussion.

Description of changes:
- Removed mutable DD cell number limiters


jngrad (approved):
looks good

PR #3706, opened 2020-05-06
Refactor RDF as an observable

jngrad:
Folllow-up to #3599

Description of changes:
- replaced the RDF analysis tool by an RDF observable
- removed the particle position storage mechanism from `statistics.cpp`
- documented observable `TimeSeries`


jngrad:
The RDF observable is kind of a stand-alone class. If we decide to port `calc_part_distribution` to an observable (which is very similar to RDF) in the future, we might create an intermediate observable class that takes two lists of particle ids, but for now I'm somewhat ok with this architecture.

The time series documentation needs to be backported in 4.1.3.

src/core/observables/RDF.cpp
```diff
@@ -0,0 +1,84 @@
...
+    for (auto jt = mixed_flag ? particles2.begin() : std::next(it),
+              jend = mixed_flag ? particles2.end() : particles1.end();
+         jt != jend; ++jt) {
+      auto const p1 = *it, p2 = *jt;
+      auto const dist = get_mi_vector(p1->r.p, p2->r.p, box_geo).norm();
+      if (dist > min_r && dist < max_r) {
+        auto const ind =
+            static_cast<int>(std::floor((dist - min_r) * inv_bin_width));
+        res[ind]++;
+      }
```
KaiSzuttor: I think we once said that we want to store particle pairs that should be considered. I think the creation of the pairs should not be part of the observable here. What do you think?

src/python/espressomd/observables.py
```diff
@@ -883,3 +883,42 @@ class CylindricalLBVelocityProfile(ProfileObservable):
 
     """
     _so_name = "Observables::CylindricalLBVelocityProfile"
+
+
+@script_interface_register
+class RDF(Observable):
+
+    """Calculates a radial distribution function.
+    The result is normalized by the spherical bin shell, the total number
```
KaiSzuttor: also known as bulk concentration

src/core/observables/RDF.cpp
jngrad: It didn't cross my mind. I've tinkered with the RDF code, but it's not immediately obvious to me how to delegate pair creation to another function. If we store a `std::vector<std::pair<int,int>> particle_pairs` in the RDF class, we'll have to copy either `N*M` or `N*N/2` particle objects instead of `N+M` resp. `N` particle objects. If we delegate particle pair creation to the `operator()()` method, then we only move the logic to another member and add an extra overhead (3x slowdown in f71695c0f3c). Do you have a better strategy in mind?

fweik: I don't think explicitly passing the pairs is feasible (as you already found out). More abstractly speaking, what the RDF is calculating is, given two ranges, an folding (or reduction) over the cartesian product of two ranges plus an reduction kernel. The first one is a general thing that can be reused (we already have a special case of this, `Utils::for_each_pair`, which is the special case where both ranges are the same), and the kernel is calculating the distance, and updating a histogram. So in the end I think for a proper design we have to do the same separation between kernels and accumulation for pairs as we did for the single particle observables, but perhaps with slightly different algorithms (?).

jngrad: I've just created an overload for `Utils::for_each_pair()` that takes two ranges in a052ee8, and `RDF::evaluate()` now runs 5% faster in the mixed case, 1% slower in the unmixed case. Could still be improved, e.g. the overload could decay to the original if both range iterators are the same.

fweik: I think we should go forward with this, this is a good starting point for latter refactorings. I have some comments on your implementation of `for_each_pair`, but we can discuss that in the actual review. (There would be different possibilities for the design, e.g. you could implement a meta-range that runs over the cartesian products, but I think this is clearer and more "STL-like" so potentially less surprising). I think if we properly dispatch to the single rang e case in the implementation of `for_each_pair` we should be able to obtain at least the same performance as the old code...

jngrad: Alright, the overload was a bad idea due to template collision, had to find a different name. Concerning performance, the timing actually changes whenever the project is recompiled, so I can't really get something consistent, it's more or less within 1% compared to the hardcoded pair loop. I don't think it matters that much, the hardcoded pair loop implementation was already 10% and 25% faster in the mixed and unmixed case compared to the original `statistics.cpp` implementation.

fweik: I agree, let's not worry too much about the performance. The algorithm can later be optimized in isolation...

src/utils/include/utils/for_each_pair.hpp
```diff
@@ -42,15 +42,49 @@ void for_each_pair(ForwardIterator first, ForwardIterator last, BinaryOp op) {
...
+void for_each_cartesian_pair(ForwardIterator first1, ForwardIterator last1,
+                             ForwardIterator first2, ForwardIterator last2,
+                             BinaryOp op) {
+  if (first1 == first2 and last1 == last2) {
+    for_each_pair(first1, last1, op);
+    return;
+  }
+  while (first1 != last1) {
+    for (auto it = first2; it != last2; ++it) {
+      if (*it != *first1) {
```
fweik: I'm not sure that comparison by value is what we want here. This limits the usability to value types that are comparable without need, and the behavior is different from `for_each` (which does `it != first`, e.g. shallow comparison), which is surprising. I think in the general algorithm the pairs should not be filtered, this can be implemented in the op. One could also implement a `for_each_pair_if` on if wanted.

jngrad: I was unsure about that solution too. If we remove that conditional, then we cannot decay this function to `for_each_pair`, since that one removes diagonal elements. In fact, we probably don't want to decay in any case, because `for_each_pair` also removes the symmetric pairs.

src/python/espressomd/observables.py
jngrad: done

src/utils/include/utils/for_each_pair.hpp
jngrad: I've coded a `for_each_cartesian_pair_if` and removed the decay to `for_each_pair`.

src/utils/tests/for_each_pair_test.cpp
```diff
@@ -74,3 +76,107 @@ BOOST_AUTO_TEST_CASE(range) {
...
+
+  auto pair_inserter = [&pairs](int a, int b) { pairs.emplace_back(a, b); };
+  auto binary_comparison = [](int a, int b) { return a != b; };
+
+  /* Collect pairs */
+  for_each_cartesian_pair_if(vec1, vec2, pair_inserter, binary_comparison);
+
+  /* Check the result */
+  check_cartesian_pairs(n_values1, n_values2, pairs, true);
+}
```
KaiSzuttor, line 132: this test contains a lot of code duplication, it may be worth it to extract reusable parts into a testclass or function

fweik:
@KaiSzuttor did you have time to look at this?

doc/tutorials/02-charged_system/scripts/nacl.py
```diff
@@ -113,6 +115,25 @@ def combination_rule_sigma(rule, sig1, sig2):
...
     system.integrator.run(100)
 
+print("\n--->Analysis setup")
+# Calculate the averaged rdfs
+rdf_bins = 100
+r_min = 0.0
+r_max = system.box_l[0] / 2.0
+ptypes = system.part[:].type
+pids_anion = system.part[:].id[numpy.nonzero(ptypes == types["Anion"])]
+pids_cation = system.part[:].id[numpy.nonzero(ptypes == types["Cation"])]
```
KaiSzuttor: i think you should use ```system.part.select``` here

doc/tutorials/02-charged_system/scripts/nacl_units.py
```diff
@@ -126,6 +128,30 @@ def combination_rule_sigma(rule, sig1, sig2):
...
     system.integrator.run(100)
 
+print("\n--->Analysis setup")
+# Calculate the averaged rdfs
+rdf_bins = 500
+r_min = 0.0
+r_max = system.box_l[0] / 2.0
+ptypes = system.part[:].type
+pids_Cl = system.part[:].id[numpy.nonzero(ptypes == types["Cl"])]
+pids_Na = system.part[:].id[numpy.nonzero(ptypes == types["Na"])]
```
KaiSzuttor: ```system.part.select```

doc/tutorials/02-charged_system/scripts/nacl_units.py
```diff
@@ -136,41 +162,15 @@ def combination_rule_sigma(rule, sig1, sig2):
...
 # Write out the data
-rdf_fp = open('rdf.data', 'w')
-for i in range(rdf_bins):
-    rdf_fp.write("%1.5e %1.5e %1.5e %1.5e\n" %
-                 (r[i], rdf_00[i], rdf_01[i], rdf_11[i]))
-rdf_fp.close()
+with open('rdf.data', 'w') as rdf_fp:
+    for i in range(rdf_bins):
+        rdf_fp.write("%1.5e %1.5e %1.5e %1.5e\n" %
+                     (r[i], rdf_00[i], rdf_01[i], rdf_11[i]))
```
KaiSzuttor:
``` python
np.savetxt('rdf.data', np.c_[r, rdf_00, rdf_01, rdf_11])
```

samples/drude_bmimpf6.py
```diff
@@ -347,19 +349,44 @@ def combination_rule_sigma(rule, sig1, sig2):
...
-    file_traj = open(args.path + "traj.xyz", "w")
     n_parts_tot = len(system.part)
 
+    # RDFs
+    rdf_bins = 100
+    r_min = 0.0
+    r_max = system.box_l[0] / 2.0
+    ptypes = system.part[:].type
+    pids_pf6 = system.part[:].id[np.nonzero(ptypes == types["PF6"])]
+    pids_bmim = system.part[:].id[np.nonzero(ptypes == types["BMIM_COM"])]
```
KaiSzuttor: ```system.part.select```

samples/drude_bmimpf6.py
```diff
@@ -371,41 +398,18 @@ def combination_rule_sigma(rule, sig1, sig2):
...
+    rdf_00 = acc_00.get_mean()
+    rdf_11 = acc_11.get_mean()
+    rdf_01 = acc_01.get_mean()
+    r = obs_01.bin_centers()
 
-    file_traj.close()
+    with open(args.path + "rdf.dat", 'w') as rdf_fp:
+        for i in range(rdf_bins):
+            rdf_fp.write("%1.5e %1.5e %1.5e %1.5e\n" %
+                         (r[i], rdf_00[i], rdf_11[i], rdf_01[i]))
```
KaiSzuttor:
``` python
np.savetxt(args.path + "rdf.dat", np.c_[r, rdf_00, rdf_11, rdf_01])
```

KaiSzuttor (changes requested):
LGTM except for some small tweaks.

src/utils/tests/for_each_pair_test.cpp
jngrad: yes, we can replace a lot of code duplication using a fixture

doc/tutorials/02-charged_system/scripts/nacl.py
jngrad: thanks, I forgot about that feature!

KaiSzuttor (approved):

PR #3708, opened 2020-05-09
Replace macros in Shapes::Rhomboid

### LINKED ISSUE #3630: Replace macros in Shapes::Rhomboid
Shape::Rhomboid contains macros that mix calculation with control flow which cause problems with clang-tidy and are a bad idea on general grounds. Those should be replaced.
When this is fixed, the clang-tidy check `bugprone-macro-parentheses` can be enabled.

jngrad:
Fixes #3630

Description of changes:
- replaced macros by lambdas in `Rhomboid::calculate_dist()`
- re-enabled Clang-Tidy check `bugprone-macro-parentheses`

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -6,7 +6,6 @@ Checks: |
    -clang-analyzer-optin.mpi.MPI-Checker,
    -clang-analyzer-security.FloatLoopCounter,
    bugprone-*,
-   -bugprone-macro-parentheses,
    clang-analyzer-alpha.*,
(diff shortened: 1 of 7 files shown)


jngrad:
w.r.t performance, the calculation is 1% faster in the core, although the error bar is so large it's most likely just as fast as the macro implementation.

fweik (dismissed):
Great work, I did expect this to be fixed right away!

fweik (approved):

PR #3712, opened 2020-05-13
Refactor Observable_stat framework

jngrad:
Description of changes:
- convert `Observable_stat` structs to self-contained classes
   - remove manual memory management when resizing `Observable_stat` objects
   - replace hazardous raw pointers by `Utils::Span`
   - hide observable implementation details
      - accumulation of the flat array and MPI reduction are done by the class
      - finding the contribution of specific bonded/non-bonded IAs is done by the class
- dependency inversions
   - replace manually-curated `Observable_stat` list in `invalidate_obs()` by a self-updating callback list
   - decouple observables from the pressure/energy/coulomb/dipolar frameworks
   - remove most `statistics.hpp` includes (it's now only visible to 9 source files)
- analysis `energy()` now returns the lower triangle of the non-bonded IAs, to be consistent with `pressure()` and `stress_tensor()`
- improve Doxygen documentation of the pressure/energy/coulomb/dipolar frameworks

### DIFF
--- a/src/core/Observable_stat.hpp
+++ b/src/core/Observable_stat.hpp
@@ -19,6 +19,7 @@
 #ifndef ESPRESSO_OBSERVABLE_STAT_HPP
 #define ESPRESSO_OBSERVABLE_STAT_HPP
 
+#include <utility>
 #include <vector>
 
 struct Observable_stat {
@@ -66,11 +67,26 @@ struct Observable_stat {
     for (size_t i = chunk_size; i < data.size(); ++i)
       data[i] /= volume;
   }
+
+  /** Get contribution from a bonded interaction */
+  double *bonded_ia(int bond_id) { return bonded + (chunk_size * bond_id); }
+
+  /** Get contribution from a non-bonded interaction */
+  double *nonbonded_ia(int type1, int type2) {
+    extern int max_seen_particle_type;
+    if (type1 > type2) {
+      using std::swap;
+      swap(type1, type2);
+    }
+    return non_bonded +
+           chunk_size *
+               (((2 * max_seen_particle_type - 1 - type1) * type1) / 2 + type2);
+  }
 };
 
 /** Structure used only in the pressure and stress tensor calculation to
-   distinguish
-    non-bonded intra- and inter- molecular contributions. */
+ *  distinguish non-bonded intra- and inter- molecular contributions.
+ */
 struct Observable_stat_non_bonded {
   /** Array for observables on each node. */
   std::vector<double> data_nb;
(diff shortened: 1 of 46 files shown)


jngrad:
@fweik @RudolfWeeber the old implementation with raw pointers gave me quite some trouble. The coulomb/dipolar contributions were overflowing a zero into the virtual site and external field contributions when no coulomb/dipolar method was active, because some places in the coulomb/dipolar calculation code forgot to guard against `n_coulomb==0` or `n_dipolar==0` and were overflowing a single zero into the virtual site or external field. I also encountered overflows when switching between P3M and ELC (this increases the number of coulomb contributions without increasing the size of the array), although it surfaced halfway through the refactor and I can't tell by reading the original pointer implementation if this was an existing bug or a regression from my side. In any case, we didn't run into buffer overflow crashes because the virtual site and external field offered just enough padding to prevent out-of-bounds access. As long as the contributions from coulomb, dipolar, virtual site and external field were written to the array in that order, the overflowing values were eventually replaced by the correct values. The code is now properly guarded and ASAN doesn't complain.

This PR achieved most of the work needed to convert `Observable_stat` structs into genuine `Observable` classes. It's essentially a matter of either moving the logic from `analysis.pyx` to `observables.py` (and converting it to a Cython file), or registering the `Observable_stat` classes in the ScriptInterface framework. I haven't done that, because we should discuss it first. There are a few open questions from my side:
* all dipolar and most coulomb methods don't contribute to the pressure
* intra- and inter-molecular non-bonded energies are commented out in the python interface and inexistent in the core (edit: removed in 5693cc8)
* the `StressTensor` observable seems to short-circuit the ideal gas component of NpT (edit: most likely a mistake, it's been present since the start in aa60df48, removed in 17349d1)

If moving to genuine `Observable` classes requires too much work, we could defer it to a future PR. This PR already provides significant improvements in terms of code clarity and pointer safety.

fweik:
I'll have a look this afternoon. I'm not convinced yet that we'll keep all the functionality, we'll see...

fweik:
@jngrad Do I understand it correctly that the globals are only needed to avoid recalculation of the observables?

jngrad:
offline discussion: By the looks of it, these globals are write-only from the point of view of the core. They are read only from the ScriptInterface and the StressTensor observable. There is no need to cache these values. We could just split up the different contributions into separate observables, and remove all the MPI callbacks triggered by system changes. This can be done in a separate PR.

src/core/Observable_stat.hpp
```diff
@@ -19,59 +19,220 @@
...
+protected:
+  /** Get contribution from a non-bonded interaction */
+  Utils::Span<double> non_bonded_contribution(double *base_pointer, int type1,
+                                              int type2) const {
+    extern int max_seen_particle_type;
+    if (type1 > type2) {
+      using std::swap;
+      swap(type1, type2);
+    }
+    int offset = ((2 * max_seen_particle_type - 1 - type1) * type1) / 2 + type2;
```
fweik: This is basically what `Utils::upper_triangular` does I think?

jngrad: indeed, thanks!

fweik (approved):
LGTM, clearly an improvement!

PR #3717, opened 2020-05-15
core: Reinit longrange stuff on manual cutoff or skin change

fweik:
Description of changes:
- (d)P3M depends on the local box size and skin (because this sets he bounding box for local particles), so it has to be reinitialized if they change.

@jngrad this is also a candidate for back porting.


fweik:
Unfortunately I couldn't think of a reliable way to test this in a robust fash, as it depends on loads of other circumstances...

jngrad (approved):

PR #3721, opened 2020-05-16
Removed some code duplication from ELC

fweik:
Description of changes:
- Per particle functions for dielectric and self-forces and energies


jngrad (approved):
nice, thanks!

Issue #3718, opened 2020-05-15
Create StressTensor/Pressure/Energy observables


jngrad:
Follow-up to #3712.

Next step is to convert the `Observable_stat` globals into `Observable` classes. Here is the plan:
* `StressTensorKinetic`: NxN matrix
* `StressTensorVirtualSites`: NxN matrix
* `StressTensorExternalFields`: NxN matrix
* `StressTensorCoulomb`: NxN matrix, there doesn't seem to be a need for 2 or 3 matrices
* `StressTensorDipolar`: skip it: not implemented
* for interactions, if we want to preserve the way they have always been output:
   * `StressTensorBonded`: NxN matrix
   * `StressTensorNonBonded`: NxN matrix
   * `StressTensorBondedContributions`: ZxZxNxN with Z the number of bonded IAs
   * `StressTensorNonBondedContributions`: ZxZxNxN? if Z is `max_seen_particle_type`, we'll have a large matrix with reaction ensemble algorithms (the core base class `ReactionAlgorithm` sets the non-interacting type to 100)
   * `StressTensorNonBondedIntra`: skip it: just get the diagonal terms of `StressTensorNonBondedContributions`
   * `StressTensorNonBondedInter`: skip it, just get the off-diagonal terms of `StressTensorNonBondedContributions`
* simpler alternative for interactions:
   * `StressTensorBonded(bonds=[bond1, bond2] or ['FENE', 'AngleCosine'] or 'all')`: NxN matrix
   * `StressTensorNonBonded(types1=[0, 1] or 'all', types2=[3, 5] or 'all', interactions=['wca', 'hertzian'] or 'all')`: NxN matrix

Same strategy for `Pressure*` and `Energy*` observables. `EnergyDipolar` will be implemented.

That's going to be a lot of files. Should we create a new subfolder `src/core/observables/impl` to move those and all the already existing `final` classes, so that `src/core/observables` only contains `virtual` base classes, e.g. `PidObservable`, `ProfileObservable`, `PidProfileObservable`, etc.?

jngrad:
offline discussion: rename `StressTensor` to `PressureTensor`, create `Energy` and `Pressure` for the total energy and total scalar pressure, remove state tracking from the global observables, leave the analysis module as it is now until a better solution is found (it's helpful to print all contributions when debugging simulations).

jonaslandsgesell:
>the core base class ReactionAlgorithm sets the non-interacting type to 100

[...]

PR #3723, opened 2020-05-18
Create observables based on Observable_stat and remove state tracking

### LINKED ISSUE #2200: stress tensor should be renamed to pressure tensor
since it has the sign convention of a pressure tensor

jngrad:
Fixes #2200, partial fix for #3718

Description of changes:
- removed state tracking from `Observable_stat`
- created `Observables` classes `PressureTensor`, `Pressure`, `Energy`
- renamed Stress Tensor to Pressure Tensor everywhere


src/core/Observable_stat.cpp
```diff
@@ -32,28 +32,15 @@
 
 extern boost::mpi::communicator comm_cart;
```
fweik: Sorry I missed that in the review of the last PR. Please don't use extern definitions, this makes tool usage impossible and can lead to very obscure errors, e.g. if the original variable is renamed or it's type is changed. Also the dependency (on `communication.hpp`) should be explicitly stated.

src/core/Observable_stat.cpp
```diff
@@ -32,28 +32,15 @@
...
-}
-
 void Observable_stat::resize() {
   // number of chunks for different interaction types
-  auto const n_coulomb =
-      m_pressure_obs ? Coulomb::pressure_n() : Coulomb::energy_n();
-  auto const n_dipolar =
-      m_pressure_obs ? Dipole::pressure_n() : Dipole::energy_n();
-  size_t n_vs = 0;
+  size_t n_coulomb = 0, n_dipolar = 0, n_vs = 0;
```
fweik: I'd tend to make the contributions exist in any case, to reduce the noise level. Also those can be const either way.

src/core/pressure_inline.hpp
```diff
@@ -201,8 +196,10 @@ inline void add_kinetic_virials(Particle const &p1, bool v_comp) {
 
   /* kinetic energy */
   if (v_comp) {
-    obs_scalar_pressure.local.kinetic[0] +=
-        (p1.m.v - (p1.f.f * (0.5 * time_step / p1.p.mass))).norm2() * p1.p.mass;
+    // velocity at half the time step: v(t + dt / 2) =
```
fweik, line 204: Since this is purely additive, and only needed for NpT (IIRC), shouldn't that be added in the NpT calculation? The `time_step` in the observable calculation is really alien to me, and most users of this function would not be concerned with this.

fweik (changes requested):
LGTM except for minor changes. Maybe @RudolfWeeber can have a look on the interfaces changes, also somebody else needs to sign off on the stress tensor name change ;-)

src/core/pressure_inline.hpp
jngrad: Not sure if I understand what you mean. The velocity is subtracted, then squared before being accumulated.

Also, the NpT scalar pressure diverges from the pressure tensor a few line below, since we use uncorrected velocities. This will have to change once we remove the scalar pressure observable in favor of the trace of the pressure tensor.

src/core/Observable_stat.cpp
jngrad: That's an oversight from my part, this extern is unnecessary with the proper include. Thanks!

src/core/pressure_inline.hpp
fweik: Sorry, I meant that if the logic for the calculation should be reused with multiple velocities it should be a free function that takes the velocity as an argument, which can then be called in the pressure calc with the particle velocity, and in NpT with the corrected velocity. The oddness of the function stems from the fact that it does two things (obtain the velocity, and then do the calculation), which hampers the reusability of the latter. But if this drops out anyway, we can leave it as is.

jngrad: I just tried it in 399ef52805d, but it fails the `PressureTensor` test because velocity compensation is done by default by `PressureTensor`, even when NpT isn't active. We could add a parameter `v_comp` to `PressureTensor`, in which case `Pressure` and `Energy` should also get it, although `Energy` will not use it (it doesn't have the logic to compensate the linear velocity and rotational velocity). I'd suggest to leave things as they are for now; once we've got a plan to create all the observables, we'll calculate the scalar pressure as the trace of the pressure tensor and this `v_comp` variable will be unused. This will break the NpT pressure calculation, but honestly the NpT pressure/energy calculation is already questionable (the NpT PV=NRT sample was working correctly according to Michael 2 years ago, but it was broken when I and Ingo used it 1 year ago).

src/core/Observable_stat.cpp
jngrad: done

fweik (dismissed):

jngrad:
@RudolfWeeber would you like to have a look? The API change is limited to renaming the `StessTensor` observable to `PressureTensor`. Methods in `system.analysis` did not change name and return the same dict objects as before.

RudolfWeeber:
Yes, I’ll review the PR.

RudolfWeeber:
If we use "pressure tensor"now, that name should be used also in the espressomd.analysis module. Otherwise, it'd be a bit confusing.

Where does the energy of the Homogeneous magnetic field go? If I understand the Pr correctly, this is no longer separable from the dipole-dipole energy? That's probably not ideal.

jngrad:
> Where does the energy of the Homogeneous magnetic field go? If I understand the Pr correctly, this is no longer separable from the dipole-dipole energy? That's probably not ideal.

Indeed, this PR merges all dipolar contributions into a single field. Electrostatic and magnetostatic methods were storing their various contributions into unlabeled slots, which only works if contributors know exactly what they are doing when they write `obs_energy.local.dipolar[0] += energy;`. The python interface used to give the different contributions with the integer index of the field, which can only be interpreted by those familiar with the C++ code.

Looking again at the diff, it seems the dipolar energies are written to the fields following a pattern, e.g. magnetic field-dipole and dipole-dipole interactions in `[0]`, dipolar solver in `[1]`, corrections in `[2]`. If we revert back to the old behavior, we should at least document this somewhere, as it wasn't obvious to me.

> If we use "pressure tensor"now, that name should be used also in the espressomd.analysis module. Otherwise, it'd be a bit confusing.

I would prefer not to modify this part of the `espressomd.analysis` module as it would break everyone's scripts. The long-term plan is to replace this module with observables anyway, so there's no need for developers and users to spend effort in adapting code and scripts to the new convention in analysis, only to have to spend time again adapting scripts to observables once we remove the feature from the analysis module.

At the moment, the PressureTensor/Pressure/Energy observable framework is still work in progress. What about the following: once we're done with the observables infrastructure, we'll let ICP members know they should consider updating their scripts to use observables while the analysis module remains for a grace period, as we did for steepest descent (that grace period would end before the release deadline). For as long as the observables framework remains a WIP, I'd be fine with naming `PressureTensor` back to `StressTensor`.

RudolfWeeber:
>Indeed, this PR merges all dipolar contributions into a single field. Electrostatic and magnetostatic methods were storing their various contributions into >unlabeled slots, which only works if contributors know exactly what they are doing when they write obs_energy.local.dipolar[0] += energy;. The >python interface used to give the different contributions with the integer index of the field, which can only be interpreted by those familiar with the C++ code.

>Looking again at the diff, it seems the dipolar energies are written to the fields following a pattern, e.g. magnetic field-dipole and dipole-dipole interactions in >[0], dipolar solver in [1], corrections in [2]. If we revert back to the old behavior, we should at least document this somewhere, as it wasn't obvious to me.

I’d keep the individual ones, but as you suggest, change the key names in the return dict of analsys.energy()

 

 

>If we use "pressure tensor"now, that name should be used also in the espressomd.analysis module. Otherwise, it'd be a bit confusing.

I would >prefer not to modify this part of the espressomd.analysis module as it would break everyone's scripts. The long-term plan is to replace this module >with observables anyway, so there's no need for developers and users to spend effort in adapting code and scripts to the new convention in analysis, only to have >to spend time again adapting scripts to observables once we remove the feature from the analysis module.

>At the moment, the PressureTensor/Pressure/Energy observable framework is still work in progress. What about the following: once we're done with the >observables infrastructure, we'll let ICP members know they should consider updating their scripts to use observables while the analysis module remains for a >grace period, as we did for steepest descent (that grace period would end before the release deadline). For as long as the observables framework remains a WIP, >I'd be fine with naming PressureTensor back to StressTensor.

I think using the same name for the same thing is preferable, even if users have to change it in their script. We had a few non-silent api changes in previous minor versions as well, e.g., with random seeds and with LB setup. 

 

There is that issue that people seem to use the opposite sign convention for stress, i.e., positive number means that the system wants to contract.

 

So, I vote for renaming to pressure tensor everywhere. 

Than it should also be clear that pressure and pressure tensor use the same sign convention.

jngrad:
> So, I vote for renaming to pressure tensor everywhere.

Done everywhere in c5a19f4d2e. I've also made the change in exported functions of the LB code. There were a few comments in the LB code and LB tutorials to remind developers that LB stress was in fact LB pressure.

> I’d keep the individual ones, but as you suggest, change the key names in the return dict of analsys.energy()

I've removed the commits that merged dipole slots into a single field. Same for electrostatics. These slots are now documented in 072180ff.

The rest of the commits have been slightly touched to resolve conflicts from the removal.

fweik:
@RudolfWeeber can you please check if your points have been addressed?

RudolfWeeber (approved):

fweik:
For what its worth, I think it is a mistake to keep the separate long-rang and short-range contributions from the dipolar and electrstatics solvers exposed. Using them separately is almost always a mistake,
we should not encourage depending on details of solvers.

PR #3725, opened 2020-05-22
Fix Variant bug in pack_pair()

### LINKED ISSUE #3711: Variant_test fails with boost 1.73.0
The unit test `Variant_test` fails with the most recent version of boost 1.73.0 in the `unpack_pair` test. The `const std::vector<Variant> &v` passed to `unpack_pair` (Variant.hpp:79) contains garbage. :)

The reason is some fancy template gone wrong in the script interface. Compiler warning:
```
[...]
/home/hirschsn/esp/espresso-add/src/script_interface/constraints/initialize.cpp:67:79:   required from here
/home/hirschsn/esp/espresso-add/src/script_interface/Variant.hpp:75:29: warning: narrowing conversion of ‘(double)pair.std::pair<int, double>::second’ from ‘double’ to ‘boost::enable_if_c<true, bool>::type’ {aka ‘bool’} [-Wnarrowing]
   75 |   return {{pair.first, pair.second}};
      |                        ~~~~~^~~~~~
```

Latest ESPResSo python branch.
Edit: Forgot to mention: Of course, with boost 1.72.0 everything works fine.

jngrad:
Fixes #3711

### DIFF
--- a/src/script_interface/Variant.hpp
+++ b/src/script_interface/Variant.hpp
@@ -72,7 +72,7 @@ inline bool is_none(Variant const &v) { return is_type<None>(v); }
 
 template <typename T, typename U>
 std::vector<Variant> pack_pair(const std::pair<T, U> &pair) {
-  return {{pair.first, pair.second}};
+  return {pair.first, pair.second};
 }
 
 template <typename T, typename U>


fweik (approved):

jngrad:
@junghans You'll most likely need this patch to build espressomd 4.1.2 on the upcoming Fedora 33. It merges cleanly (backported in 4.1.3 as f1286e4941).

PR #3731, opened 2020-05-25
missing coulomb prefactors in ELCIC

### LINKED ISSUE #3722: ELC self-forces do not obey Coulomb prefactor.
The elc self interaction is add directly to the particles in `ELC_P3M_self_forces`, but from what I can tell it does not consider the Coulomb prefactor.

reinaual:
Fixes #3722


fweik:
Ah thank you. I missed that the ELC energy is added on a different path.

fweik (approved):

PR #3732, opened 2020-05-26
Update installation instruction for mac

KaiSzuttor:
as reported by C. Holm


fweik:
@mkuron can you maybe have a look at the command change, I don't understand that ports stuff...

mkuron:
Cannot confirm this. The documentation was correct as-is. Maybe he needs to update his MacPorts. I believe that was changed a few months ago.
```
$ port installed openmpi*
  openmpi-default @4.0.1_1 (active)
$ port select --list mpi
Available versions for mpi:
	none
	openmpi-mp (active)
```

KaiSzuttor:
Even if the documentation is correct in some cases, it obviously does not cover all macport environments.

fweik:
I'd rather have it not reference fortran if not needed...

mkuron (approved):
Okay, he is right. I seem to have had an OpenMPI package without Fortran bindings.

```
$ port install openmpi-default +gcc10
$ port select --list mpi
Available versions for mpi:
	none
	openmpi-mp-fortran (active)
```

The `+gcc10` variant is the default, so Fortran bindings will be installed by default.

fweik:
So the updated version is correct, yes?

fweik (approved):

PR #3745, opened 2020-05-31
Modernized Readme file

jngrad:
Follow-up to the [2020-05-26 ESPResSo meeting](https://github.com/espressomd/espresso/wiki/Espresso-meeting-2020-05-26#user-survey)

Description of changes:
- explain how to get released versions
- invite users to subscribe to the mailing list
- give clear instructions on how to cite ESPResSo
- for contributors, briefly explain our release workflow


fweik (dismissed):

fweik:
This also reflects my understanding of the API guarantees that we want to make.

Readme.md
```diff
@@ -47,45 +47,62 @@ simulations.  It is mainly developed at the Institute for
...
+upcoming releases. You'll also get notifications on bugfix releases.
 
 ## License
 
 Copyright (C) 2010-2019 The ESPResSo project
-Copyright (C) 2002,2003,2004,2005,2006,2007,2008,2009,2010 
-Max-Planck-Institute for Polymer Research, Theory Group
-  
+
+Copyright (C) 2002-2010 Max-Planck-Institute for Polymer Research, Theory Group
```
KaiSzuttor, line 92: I think we should just refer to the (non-existing) License file that could then also serve as the `header template`.

RudolfWeeber:
Can the mailing list info be moved up?  I don't think most people will read that far. If possible, the mailing list info should be immediately after the brief description of ES's purpose.

Readme.md
jngrad: I would be fine with shortening this section of the Readme to the following text, which links to the [license](https://github.com/espressomd/espresso/blob/python/COPYING) and acknowledges both the MPI and ESPResSo organizations:
```
Copyright (C) 2010-2019 The ESPResSo project
Copyright (C) 2002-2010 Max-Planck-Institute for Polymer Research, Theory Group
ESPResSo is free software: you can redistribute it and/or modify it under
the terms of the GNU General Public License as published by the Free
Software Foundation, either version 3 of the License, or (at your option)
any later version. A copy of the license is provided in [COPYING](COPYING).
```
The issue with providing instead a link to the [header template](https://github.com/espressomd/espresso/blob/python/maintainer/header_template.txt) is that the MPI line is missing.

jngrad:
@RudolfWeeber Placing the mailing list inside the Installation section of the Readme felt natural to me, as new users would typically look first for the installation instructions. By putting it after the ES description, i.e. before the documentation and installation instructions, I'm afraid new users might just skip it. I can move up the mailing list a little bit, between the installation and citation section above, but if you insist I'll place it after the ES description, as you proposed.

Readme.md
KaiSzuttor: Why is the MPI part missing in the header template? I think we should move the header template to ```LICENSE``` in the root and add the MPI part.

jngrad: The license header can only contain years where the file received non-trivial contributions, therefore the MPI cannot appear for files created from scratch after 2010. Having this header template puts us in a grey area, since source files committed without a license header in the first commit where they appear are automatically licensed in a way that depends on the country(ies) of origin of the contributor, if I recall correctly. Adding the GPL after the fact during releases is not really permitted without the agreement of all people who committed non-trivial things in that file since its creation. That's why Google and Fedora have annoying bots that reject your contributions until you sign a legal document that allows them to relicense your contributions as GNU GPL if you forgot to put it in the first commit.

If I'm not mistaken, `COPYING` is the standard place to put the license in GNU GPL projects, while `LICENSE` is the standard place to put the license in Apache projects.

RudolfWeeber:
JN, please proceed as you see fit with this PR.

RudolfWeeber (approved):

Issue #3713, opened 2020-05-14
Segmentation fault with Espresso 4.1.1 using p3m, Gay-Berne potential, Lennard-Jones potential and virtual sites


Kischy:
I have a very weird segmentation fault in one of my simulations. The curios thing is that I do multiple simulations in exactly the same manner and the only thing I change between different simulations is the position of a virtual particle, that carries a charge. For one particular charge position the simulation fails with a segmentation fault, all the others work just fine. The point at which it fails is different in every run, but it fails consistently. I also tested it on a different machine and using a different version of espresso, but the error persists. I can give more information if needed.

Here is an example of the error message:
```
[dhcp-184:16942] *** Process received signal ***
[dhcp-184:16942] Signal: Segmentation fault (11)
[dhcp-184:16942] Signal code: Address not mapped (1)
[dhcp-184:16942] Failing at address: 0xfffffffc0271d940
[dhcp-184:16942] [ 0] /lib/x86_64-linux-gnu/libc.so.6(+0x3ef20)[0x7f8f7abf6f20]
[dhcp-184:16942] [ 1] /home/chaege/espresso-4_1_1_ILC/espresso/build/src/core/EspressoCore.so(void p3m_do_assign_charge<7>(double, Utils::Vector<double, 3ul>&, int)+0x21c)[0x7f8f757211bc]
[dhcp-184:16942] [ 2] /home/chaege/espresso-4_1_1_ILC/espresso/build/src/core/EspressoCore.so(p3m_charge_assign(ParticleRange const&)+0x759)[0x7f8f75713ee9]
[dhcp-184:16942] [ 3] /home/chaege/espresso-4_1_1_ILC/espresso/build/src/core/EspressoCore.so(Coulomb::calc_long_range_force(ParticleRange const&)+0x35)[0x7f8f75725745]
[dhcp-184:16942] [ 4] /home/chaege/espresso-4_1_1_ILC/espresso/build/src/core/EspressoCore.so(force_calc(CellStructure&)+0x168)[0x7f8f75662248]
[dhcp-184:16942] [ 5] /home/chaege/espresso-4_1_1_ILC/espresso/build/src/core/EspressoCore.so(integrate_vv(int, int)+0x163)[0x7f8f756786f3]
[dhcp-184:16942] [ 6] /home/chaege/espresso-4_1_1_ILC/espresso/build/src/core/EspressoCore.so(+0xed469)[0x7f8f7562d469]
[...]

PR #3748, opened 2020-06-05
Quaternion formalism in the core and range check

jngrad:
Partial fix for #3713

Description of changes:
- document quaternion formalism in the core
- add an assertion to catch invalid square roots
- explain that force capping doesn't affect torques
- use vector operations

### DIFF
--- a/doc/doxygen/bibliography.bib
+++ b/doc/doxygen/bibliography.bib
@@ -443,14 +443,35 @@ doi = {10.1016/0021-9991(85)90151-2},
 }
 
 @book{allen2017,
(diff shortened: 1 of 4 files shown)


RudolfWeeber:
LGTM. Except that steepest descent does not cap the forces/torques but the maximum displacement in position and orientation.

RudolfWeeber (dismissed):

jngrad:
> Except that steepest descent does not cap the forces/torques

good point, I'll correct the sentence

RudolfWeeber (approved):

PR #3751, opened 2020-06-08
Store unit system in H5MD file

jkrajniak:
Add option to define and store the unit system used in the simulation according to the H5MD module units defined here: https://nongnu.org/h5md/modules/units.html#unit-string

Description of changes:
 - add new argument to the class `h5md.H5md`: `unit_system`
 - enable H5MD module `units` if the unit system is set


jngrad:
Several CI images are missing the `dataclasses` python module. I'll have a look this week.

junghans (dismissed):
Great job.

jngrad (dismissed):
Looks good to me, thanks!

jngrad:
@KaiSzuttor would you like to sign off on these changes? This PR makes use of [`dataclasses`](https://docs.python.org/3.7/library/dataclasses.html), which offer the same safety checks as a `NamedTuple` while providing default values. This is part of the Python Standard Library in python3.7. There is a backport for python3.6, so we can update our Ubuntu 18.04 CI images accordingly (jngrad/docker@f674421b3), as well as the macOS image.

If we don't want to add this `dataclasses` requirement, we could rewrite the `UnitSystem` class as a regular class, without the benefit of argument name validation. I wouldn't mind adding this requirement: since python3.6 will be dropped from NEP 29 on [June 23 2020](https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule), people still using python3.6 will have more to worry about than just installing a backported package via pip.

.pre-commit-config.yaml
```diff
@@ -18,7 +18,7 @@ repos:
         always_run: false
         files: '.*\.(py|pyx|pxd)'
         exclude: '\.pylintrc|.*.\.py\.in'
-        args: ["--ignore=E266,W291,W293", "--in-place", "--aggressive"]
+        args: ["--ignore=E266,E701,W291,W293", "--in-place", "--aggressive"]
```
KaiSzuttor, line 21: why is this needed?

doc/sphinx/io.rst
```diff
@@ -164,7 +166,10 @@ to the :class:`~espressomd.io.writer.h5md.H5md` class. Currently available:
     - ``write_mass``: particle masses
 
     - ``write_ordered``: if particles should be written ordered according to their
-      id (implies serial write).
+      id (implies serial write)
+
+    - ``unit_system``: optionally, physical units for time, mass, length and
```
KaiSzuttor, line 170: what about other SI base units that are defined in the [H5MD documentation](https://nongnu.org/h5md/modules/units.html)?

KaiSzuttor (commented):
In general I think we should not write our own unit system code. So what may make sense in the future is to use [Pint](https://pint.readthedocs.io/en/0.12/) and derive the strings that need to be written to the file automatically.
Looks good to me for now. Thanks for the contribution!

KaiSzuttor:
Concerning the dataclasses, I think it's not worth the dependency overall because the respective constructor would not be a lot of additional code and the other provided functions like ```__repr__``` and the comparison functions are not needed.

.pre-commit-config.yaml
jngrad: E701 causes dataclasses members to be formatted on a new line:
```diff
-        force: str = dataclasses.field(
-            init=False, default='')
+        force:
+            str = dataclasses.field(
+                init=False, default='')
```
This change causes a syntax error:
```
1559   File "/builds/espressomd/espresso/build/src/python/espressomd/io/writer/h5md.py", line 36
1560     force:
1561          ^
1562 SyntaxError: invalid syntax
```
E701 cannot be disabled on a per-line basis. I tried using line continuation symbols `\` to work around it, without success. I think we should permanently disable it, even if we remove dataclasses, because it might cause us trouble again in the future, and troubleshooting this error took me a while.

doc/sphinx/io.rst
jngrad: The other SI units (current, temperature, moles, candela) do not seem to apply for espresso particles. The per-particle temperature is more of a trick for tailored Langevin dynamics. Same thing for derived units. Torque could have been relevant, but isn't defined.

src/python/espressomd/io/writer/h5md.py
```diff
@@ -42,19 +40,25 @@ class UnitSystem:
...
-        mass: str
-        time: str
-        length: str
-        charge: str
-        force: str = dataclasses.field(init=False, default='')
-        velocity: str = dataclasses.field(init=False, default='')
-
-        def __post_init__(self):
+
+        def __init__(self, **kwargs):
```
KaiSzuttor, line 44: you could still use type annotations if you wanted to...

KaiSzuttor (approved):

src/python/espressomd/io/writer/h5md.py
jngrad: there's no need to re-implement dataclasses :) providing a different type already throws a meaningful error message: `RuntimeError: Provided argument of type int is not convertible to std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >`

PR #3755, opened 2020-06-15
Deduce type of global fields

fweik:
Description of changes:
- Deduce the type of globals exposed thru the 'global' mechanism.


jngrad (approved):
nice, thanks!

PR #3756, opened 2020-06-15
Remove compensation flag for kinetic pressure contribution

fweik:
Description of changes:
- Fixed kinetic pressure compensation in pressure observable
- Do not depend on NpT instantanuous pressure in pressure observable calc
- Removed no unneeded `v_comp` flag for compensated kinetic pressure


jngrad (approved):
LGTM

PR #3757, opened 2020-06-15
CMake GPU dependencies inversion

jngrad:
Create a `find_gpu_library()` helper function to delegate GPU library inclusion in parts of the CMake tree that need them. This avoids updating `FindCUDACompiler*.cmake` files whenever a new GPU dependency is required.

Description of changes:
- fix a bug in `option_if_available()` that failed CI in the Intel build
- fix variable `CUDA_LIBRARY` (was an empty string on Clang)
- mark essential CUDA libraries as `REQUIRED`
- remove an incorrect include `-Isrc/core/SYSTEM`
- fix a compiler error in builds without feature `ROTATION`
- give `CMAKE_CXX_FLAGS` higher precedence

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -417,6 +417,9 @@ endif()
 
 target_link_libraries(cxx_interface INTERFACE coverage_interface)
 
+# Give CMAKE_CXX_FLAGS higher precedence
+target_compile_options(cxx_interface INTERFACE ${CMAKE_CXX_FLAGS})
+
 #
 # Testing
 # ##############################################################################
--- a/cmake/FindCUDACompilerClang.cmake
+++ b/cmake/FindCUDACompilerClang.cmake
@@ -77,10 +77,8 @@ function(find_gpu_library)
   endif()
 endfunction(find_gpu_library)
 
-find_library(
-  CUDART_LIBRARY NAMES cudart PATHS ${CUDA_DIR}/lib64 ${CUDA_DIR}/lib
-                                    /usr/local/nvidia/lib
-                                    /usr/lib/x86_64-linux-gnu NO_DEFAULT_PATH)
+find_gpu_library(VARNAME CUDA_LIBRARY NAMES cuda REQUIRED)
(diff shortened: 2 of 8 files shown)


jngrad:
To be more specific, this PR allows us to do that: 3726d5462

fweik (dismissed):
That is a neat solution

jngrad:
The Intel compiler fails in CI ([logfile](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/235183)) on this line:
https://github.com/espressomd/espresso/blob/7c8a3c30b0dc5710af58347df77f593ef1bbaf39/src/core/global.cpp#L122-L124
where `std::underlying_type_t<ActiveLB>` is `unsigned int`. Replacing the cast by `reinterpret_cast<std::size_t*>()` compiles just fine though, so it might be an issue with `std::underlying_type_t`. This workaround works:
```diff
-enum class ActiveLB { NONE, CPU, GPU };
+enum class ActiveLB: size_t { NONE, CPU, GPU };
```

fweik:
No, it's not an issue with `std::underlying_type_t`, the intel compiler just chose a different type to represent the enum. This means it that it was UB until now :-) You should fix the type, as you suggested, but maybe to int, which is what clang and gcc chose..

fweik:
Just to clarify, doing a `reinterpret_cast` to pointer to a type which is not the type of the object, and then actually accessing the object thru the pointer is not allowed

jngrad:
There is another issue, `CMAKE_CXX_FLAGS` has lower priority that `CMAKE_CXX_FLAGS_${BUILD}`, that's an issue for the Intel job. It can be mitigated using a temporary variable: https://gitlab.freedesktop.org/poppler/poppler/commit/d03f5ab76b0d0dee199b65fc801236a38830b5fc

jngrad:
> No, it's not an issue with `std::underlying_type_t`

I read the compiler error too fast and did not realize it said `unsigned int`, not `unsigned long int` :smile:

fweik (approved):

PR #3759, opened 2020-06-16
Pitchfork layout: rename CMake targets

### LINKED ISSUE #3624: ESPResSo installation is broken (.so and .py name collision)
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/219704
```
   File "/builds/espressomd/espresso/build/testsuite/python/ek_eof_one_species_base.py", line 26, in <module>
1714     import espressomd.shapes
1715 ImportError: dynamic module does not define module export function (PyInit_shapes)
```
We currently install the compiled Cython files (.so), compiled libraries (.so) and Python files (.py) in the same folder, which is just asking for trouble. #3543 introduced a name collision between `shapes.py` and `shapes.so`. The waLBerla integration will pull in even more shared objects, increasing the likelihood of name collisions.

* We cannot rename `shapes.so` to `libshapes.so` because CMake already adds a `lib` prefix depending on the OS ([`add_library()` docs](https://cmake.org/cmake/help/v3.10/command/add_library.html#normal-libraries))
* Renaming the target to `EspressoShapes` leads to the following error:
  ```
  /home/espresso/espresso/src/core/grid_based_algorithms/lbboundaries/LBBoundary.hpp:25:10: fatal error: shapes/NoWhere.hpp: No such file or directory
   #include <shapes/NoWhere.hpp>
            ^~~~~~~~~~~~~~~~~~~~
  compilation terminated.
  ```
* Installing the target in a subfolder (e.g. `site-packages/espressomd/src/shapes.so`) leads to an import error:
  ```
  >>> import espressomd.shapes
    File "<stdin>", line 1, in <module>
    File "/tmp/espresso-unit-tests/lib/python3.6/site-packages/espressomd/__init__.py", line 23, in <module>
      from . import _init
    File "_init.pyx", line 20, in init espressomd._init
  ImportError: shapes.so: cannot open shared object file: No such file or directory
  ```

jngrad:
Fixes #3624

Adding the `Espresso` prefix to targets in `/src` with "common" names will help against collisions, e.g. `EspressoShapes`, `EspressoUtils`, `EspressoParticleObservables`. This is also coherent with the naming of other targets in that directory, e.g. `EspressoCore`, `EspressoConfig`.

### DIFF
--- a/cmake/FindCUDACompilerNVCC.cmake
+++ b/cmake/FindCUDACompilerNVCC.cmake
@@ -78,7 +78,7 @@ function(add_gpu_library)
   set_property(TARGET ${GPU_TARGET_NAME} PROPERTY CUDA_SEPARABLE_COMPILATION ON)
   target_link_libraries(${GPU_TARGET_NAME} PRIVATE
     ${CUDA_LIBRARY} ${CUDART_LIBRARY} ${CUDA_CUFFT_LIBRARIES}
-    utils Boost::serialization Boost::mpi)
+    EspressoUtils Boost::serialization Boost::mpi)
 endfunction()
 
 include(FindPackageHandleStandardArgs)
--- a/cmake/FindCUDACompilerNVCC.cmake
+++ b/cmake/FindCUDACompilerNVCC.cmake
@@ -77,8 +77,7 @@ function(add_gpu_library)
   set(GPU_TARGET_NAME ${ARGV0})
   set_property(TARGET ${GPU_TARGET_NAME} PROPERTY CUDA_SEPARABLE_COMPILATION ON)
   target_link_libraries(${GPU_TARGET_NAME} PRIVATE
-    ${CUDA_LIBRARY} ${CUDART_LIBRARY} ${CUDA_CUFFT_LIBRARIES}
-    EspressoUtils Boost::serialization Boost::mpi)
+    ${CUDA_LIBRARY} ${CUDART_LIBRARY} ${CUDA_CUFFT_LIBRARIES})
 endfunction()
 
 include(FindPackageHandleStandardArgs)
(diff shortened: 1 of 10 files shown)


fweik:
What is the rational behind adding the prefix? If the goal is to make the name unique, shouldn't the target name be derived from the path/namespace to ensure it is unique? (in pitchfork spirit). If you want to add another `Observable` target, you have to come up with another unique prefix? Maybe I'm also not getting what the idea is...

jngrad:
At the moment, when building espresso, generated .so and .py files are neatly organized in subfolders that reproduce the pitchfork layout. However when we install espresso, this layout is flattened. We end up with files with the same name but different extensions, preventing python from importing them. We could also run into files overwriting each other, e.g. if `shapes.py` becomes `shapes.pyx`.

Adding another `Observable` target is easy: if you store sources in e.g. `/src/histogram_observables`, name the target `EspressoHistogramObservables`. For a more foolproof solution, we could look into what waLBerla does with target names generated from filepaths, or look into not flattening the pitchfork layout. However right now, my goal is to make CI green. CI job `installation` has been red since March 2, because manual jobs that fail don't count as a failure. This is preventing us from moving forward with the Stokesian Dynamics and waLBerla PRs, as we need to deploy external shared libraries in the espresso folder, but we cannot test it since the `shapes` target is broken.

fweik:
Okay, I guess to me the names feel like those are targets to be exported (which they are not). If you are happy with this it's fine. I guess the cython/python nameing collision should eventually be fixed by putting the pure bindings into an extra module which would also make it much more comprehensible..

fweik (approved):

jngrad:
> If you are happy with this it's fine.

For me it's more of a quick fix to remove a deadlock in CI. I would be happier with a long-term solution, but the lack of activity in #3624 shows it's not a priority for us at the moment.

PR #3763, opened 2020-06-18
Fix steepest descent interface

jngrad:
Description of changes:
- prevent visualizers from running when steepest descent is active by raising an exception (otherwise the visualizer freezes)
- additionally, only in 4.1.3: fix incorrect assertions in the steepest descent interface (0f5a3e84aa)

### DIFF
--- a/src/python/espressomd/visualization_mayavi.pyx
+++ b/src/python/espressomd/visualization_mayavi.pyx
@@ -21,6 +21,7 @@ from libcpp cimport bool
 from libcpp.vector cimport vector
 from .particle_data import ParticleHandle
 from .particle_data cimport *
+from .integrate import SteepestDescent
 from .interactions import NonBondedInteractions
 from .interactions cimport BONDED_IA_DIHEDRAL, BONDED_IA_TABULATED_DIHEDRAL
 from .interactions cimport IA_parameters, get_ia_param, bonded_ia_params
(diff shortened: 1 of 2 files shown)


fweik:
Why doesn't the visualizer work with steepest descent? Adding extra conditions for specifics integrators in unrelated code probably isn't a good idea...

jngrad:
I haven't looked at the details, but I guess the visualizer thread will run the integrator with `run(steps=1)`, which means the steepest descent will converge immediately and return true in `integrator_step_1()`, which exits the integration loop before we do the ghost update and second force calculation. Also, the main integration loop will skip many callbacks (LB, virtual sites, collision detection in 4.1).

jngrad:
I'm also not happy with introducing a dependency on an integrator in the visualizers. I originally thought about doing `str(type(self.system.integrator.get_state())) == 'SteepestDescent'`, but this has the potential to silently break if we change the name of the class. The proper way to address this would be to either rewrite steepest descent as an integrator (and think about what it means to minimize a LB fluid in a deterministic way), or create a `Minimization` framework that wouldn't use the main integration loop (or would use with its own set of callbacks instead of using the integrator callbacks), and then the visualizers can gracefully fail when there is no integrator active.

jngrad:
offline discussion: it's unclear what the real source of the issue is. Closing.

PR #3765, opened 2020-06-19
Faster tests

RudolfWeeber:
Accelerate tests.

### DIFF
--- a/testsuite/python/correlation.py
+++ b/testsuite/python/correlation.py
@@ -55,9 +55,9 @@ class CorrelatorTest(ut.TestCase):
             obs1=obs, tau_lin=10, tau_max=2, delta_N=1,
             corr_operation="square_distance_componentwise")
 
-        s.integrator.run(100)
+        s.integrator.run(1000)
         s.auto_update_accumulators.add(acc)
-        s.integrator.run(2000)
+        s.integrator.run(1000)
 
         corr = acc.result()
 
@@ -80,12 +80,11 @@ class CorrelatorTest(ut.TestCase):
 
         obs = espressomd.observables.ParticleVelocities(ids=(0,))
         acc = espressomd.accumulators.Correlator(
-            obs1=obs, tau_lin=10, tau_max=10.0, delta_N=1,
+            obs1=obs, tau_lin=10, tau_max=2, delta_N=1,
             corr_operation="tensor_product")
 
-        s.integrator.run(1000)
         s.auto_update_accumulators.add(acc)
-        s.integrator.run(20000)
+        s.integrator.run(1000)
 
         corr = acc.result()
 
@@ -105,12 +104,11 @@ class CorrelatorTest(ut.TestCase):
 
         obs = espressomd.observables.ParticleVelocities(ids=(0,))
         acc = espressomd.accumulators.Correlator(
-            obs1=obs, tau_lin=10, tau_max=10.0, delta_N=1,
+            obs1=obs, tau_lin=10, tau_max=2, delta_N=1,
             corr_operation="componentwise_product")
 
-        s.integrator.run(1000)
         s.auto_update_accumulators.add(acc)
-        s.integrator.run(20000)
+        s.integrator.run(1000)
 
         corr = acc.result()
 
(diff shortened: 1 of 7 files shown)


jngrad:
note: `testsuite/python/wang_landau_reaction_ensemble.py` has always been slow (there was a ticket 2 years ago), but it has become really slow recently. I'm not sure at which point it started, at least a couple of weeks according to the GitLab CI logs. More specifically, the coverage builds in CI jobs `default` and `maxset` can take anywhere from 100s to 260s. Sometimes we reach the 300s mark and the test times out. Maybe the convergence criterion could be relaxed.

testsuite/python/correlation.py
```diff
@@ -50,13 +50,13 @@ def test_square_distance_componentwise(self):
         s = self.system
         s.part.add(id=0, pos=(0, 0, 0), v=(1, 2, 3))
 
-        O = espressomd.observables.ParticlePositions(ids=(0,))
-        C2 = espressomd.accumulators.Correlator(
-            obs1=O, tau_lin=10, tau_max=2, delta_N=1,
+        obs = espressomd.observables.ParticlePositions(ids=(0,))
+        acc = espressomd.accumulators.Correlator(
+            obs1=obs, tau_lin=10, tau_max=10.0, delta_N=1,
```
jngrad: I accidentally reverted the `tau_max` value during the merge. However by setting `tau_max=2` in 7864a096 you moved from the multiple tau correlator to the linear tau correlator (since  `tau_lin < tau_max`), which according to the docs should be slower in this case because `tau_max = 200 * dt`. Was this change intentional?

RudolfWeeber:
My understanding is:

`Tau_lin *delta_n` is an integer and counts the number of time steps for which the correlation is evaluated at the finest resolution (delta_n).

`tau_max` is a float and measured in simulation time.

So, as long as 

`tau_max >time_step  *delta_n *tau_lin`, at least two hierarchy levels are used, and the level aggregation code is tested.

Also, the correlator always fills a hierarchy level that has been started.

 

The smallest tau_max that sattisifies the inequality should test the full code. Chosing a bigger one just means more calculations.

So, as far as I understand, `tau_max=2.0` should do the trick.

jngrad:
> `tau_max >time_step  *delta_n *tau_lin`, at least two hierarchy levels are used, and the level aggregation code is tested.

Then we should reflect that in the documentation, because it seems outdated:
https://github.com/espressomd/espresso/blob/9cc0cd17a54ca56b56c5c99ad2c1758be2b67658/src/python/espressomd/accumulators.py#L200-L201

testsuite/python/correlation.py
jngrad: The merge was fixed.

testsuite/python/correlation.py
```diff
@@ -52,12 +52,12 @@ def test_square_distance_componentwise(self):
 
         obs = espressomd.observables.ParticlePositions(ids=(0,))
         acc = espressomd.accumulators.Correlator(
-            obs1=obs, tau_lin=10, tau_max=10.0, delta_N=1,
+            obs1=obs, tau_lin=10, tau_max=2, delta_N=1,
```
jngrad, line 55: during the rebase, you accidentally discarded the merge conflict resolutions; the `tau_max=2` change must be applied to the new test cases at lines 83, 108 and 133 (but not 189, because this test doesn't run the integrator and the custom `tau_max` value there is important)

RudolfWeeber:
The correlation test also contained "warmup" integrations which are not needed, since the system is deterministic and set up in its final state.
Also, the sampling integrations were shortened.

jngrad:
This PR saves around 3-4 min on `clang-sanitizer` on coyote11. Summary of the touched tests, runtime in seconds:
| test                | before | after |
| :---                |   ---: |  ---: |
| thermalized_bond    |  13.20 |  8.23 |
| reaction_ensemble   |  58.60 | 17.22 |
| dpd                 | 114.52 | 51.67 |
| langevin_thermostat | 144.34 | 64.73 |
| brownian_dynamics   | 171.88 | 17.54 |
| correlation         |  57.26 |  5.93 |
| integrator_npt      | 111.41 | 32.83 |

(from [baseline logfile](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/238481) vs. [PR logfile](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/238654))

Taking the sum of each column, the gain in runtime is 8 min. The 220s gain we see in the logfile could be due to the GPU bottleneck: GPU tests run in serial, which can cause 1 core to be idle when all non-GPU tests have completed.

jngrad (approved):
LGTM, thanks!

PR #3766, opened 2020-06-22
Simplify Observable_stat

fweik:
Description of changes:
- Increase code reuse for Observable_stat
- Added inter/intra contributions for non-bonded energy
- Removed unused non-bonded and bonded fields from `CUDA_energy`

### DIFF
--- a/src/core/electrostatics_magnetostatics/dipole.hpp
+++ b/src/core/electrostatics_magnetostatics/dipole.hpp
@@ -68,35 +68,6 @@ struct Dipole_parameters {
 extern Dipole_parameters dipole;
 
 namespace Dipole {
-/** Number of electrostatic contributions to the system pressure calculation. */
-constexpr size_t pressure_n() { return 0; }
-
-/** Number of electrostatic contributions to the system energy calculation.
- *  - slot 0: energies from particle pairs and magnetic field constraints
- *  - slot 1: energies from magnetostatics solvers
- *  - slot 2: energy corrections
- */
-inline size_t energy_n() {
-  switch (dipole.method) {
-  case DIPOLAR_NONE:
-    return 1; // because there may be an external magnetic field
-  case DIPOLAR_P3M:
-  case DIPOLAR_ALL_WITH_ALL_AND_NO_REPLICA:
-  case DIPOLAR_DS:
-  case DIPOLAR_DS_GPU:
-#ifdef DIPOLAR_BARNES_HUT
-  case DIPOLAR_BH_GPU:
-#endif
-  case DIPOLAR_SCAFACOS:
-    return 2;
-  case DIPOLAR_MDLC_P3M:
-  case DIPOLAR_MDLC_DS:
-    return 3;
-  default:
-    return 0;
-  }
-}
-
 void calc_pressure_long_range();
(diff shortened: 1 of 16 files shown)


src/core/electrostatics_magnetostatics/dipole.hpp
```diff
@@ -68,35 +68,6 @@ struct Dipole_parameters {
 extern Dipole_parameters dipole;
 
 namespace Dipole {
-/** Number of electrostatic contributions to the system pressure calculation. */
-constexpr size_t pressure_n() { return 0; }
-
-/** Number of electrostatic contributions to the system energy calculation.
- *  - slot 0: energies from particle pairs and magnetic field constraints
- *  - slot 1: energies from magnetostatics solvers
- *  - slot 2: energy corrections
```
jngrad, line 77: By removing this dependency, we also lose the attached documentation. Based on our offlline discussion last week, losing this docstring should not be a problem in the final state of the observable framework. Until then, we just have to make sure no one accidentally does the following:
```diff
--- a/src/core/Observable_stat.cpp
--- b/src/core/Observable_stat.cpp
@@ -31,1 +31,6 @@
+ #ifdef DIPOLES
  auto constexpr n_dipolar = 3;
+ #else
+ auto constexpr n_dipolar = 0;  // bug: a constraint actually writes to the
+                                // obs, even when DIPOLES is not compiled in
+ #endif
```

fweik:
I'm not sure that I will work on this further. The CUDA part ist pretty messed up.

jngrad:
I avoided this CUDA compiler error by moving the MPI logic into the .cpp file, at the cost of using the `comm_cart` global.

fweik:
Yes, which apparently slipped my fingers in code review...

src/core/electrostatics_magnetostatics/dipole.hpp
fweik: I'm not sure I understand your point.

jngrad: These docstrings were introduced at the request of @RudolfWeeber in the last `Observable_stat` refactor. When doing the code review for this PR, I was about to propose adding feature guards on the `n_coulomb` and `n_dipolar` variables to be consistent with the feature guard on `n_vs`. By reading the docstrings you removed, I realized you didn't add feature guards on `n_dipolar` because of a side effect from a magnetostatics constraint.

fweik: I see. To me this does not seem right, e.g. the type should not contain comments about how clients use it. Also wouldn't that just cause a compiler error exactly at the place of use, which is better information?

fweik:
The CUDA stuff was solvable by moving the reduction to a free function. This is obviously not finished yet, but I think this state is worth merging.

src/core/electrostatics_magnetostatics/dipole.hpp
jngrad: right, the Spans do range checking. I was reasoning on the old implementation for some reason.

jngrad (approved):
LGTM, thanks!

Issue #2628, opened 2019-03-24
Remove global variables


jngrad:
Global variables make code difficult to read and maintain. Tracking their origin is an annoyance, and name collision with local variables is always a risk. Should we consider prefixing global variables with `g_`, just like we prefix class member variables with `m_` and enum values with their enum name? Or enclosing them in namespaces, like enum values from enum classes?

Here is one special example, hiding in the global scope of some LB code:
https://github.com/espressomd/espresso/blob/b25c0148bde7908af926e6360198ac7532c6688a/src/core/grid_based_algorithms/lbgpu.cpp#L115

Documentation is also affected by this situation, because Doxygen doesn't have a reliable scope resolution mechanism (Doxygen incorrectly states this variable is referenced by 380 functions).

fweik:
The one you mention is also a linking hazard. In general most should also be static.

jngrad:
Other example: there is a global variable `energy`:
https://github.com/espressomd/espresso/blob/182020aa5d0c09c4283b3b55db7439038e98a7a2/src/core/energy.hpp#L38
which can easily be shadowed in functions calculating energies, forcing the use of alternative variable names, like `_energy`, `eng`, `ret`.

Removing globals is also an option: e416e7e528bf3da75, d47a73138a, 34b8821dd3f682, fab58ca37ae, 6b56534

jngrad:
Progress report:
* thermostat/integrator global variables were moved into structs
* global variables associated with the Mersenne Twister were removed (0abe004dc)
* many obsolete globals were removed from the Python interface (1cb64f280403, 80a79079674e, 2ba1480cf)
* most accesses to `local_cells` have been removed (#2899) and the last occurrences are now being accessed through a function (#3501)

jngrad:
Progress report:
* removed CUDA-related globals: a07a124590ad4eae540 955ad2a71f9583 1d6c7ee1258012e81 fe8ddaade293 436ce0733902 823b2f296898fa0a bb02c861755503f8 235480bc1743dcf fd3f7fd495
* removed cellsystem-related globals: 6a3e36457f3b 4bab12f4801a6 9a11ffcac7c0c 3ef73b7518d39ea98 84690b23060a04f2 c829dd965c6fc1301
* removed electrostatics-related globals: 05f5a066e652e2fbd2ef3eb12b70980fff70bce2, 54ea3b5573c458dcbea581e0364c6e3eebf6a297
[...]

PR #3770, opened 2020-06-23
Remove Observable_stat external linkage

jngrad:
Description of changes:
- re-use code in the cython interface to extract data from `Observable_stat` objects
- remove `obs_scalar_pressure` global variable
   - write-only variable in the core -> can be calculated from `obs_pressure_tensor` in cython
   - simplify pressure kernels
   - less MPI communication
- rename `obs_pressure_tensor` to `obs_pressure` for consistency with `obs_energy`
- remove external linkage for  `obs_pressure` and `obs_energy`
   - these globals are no longer visible outside of their translation unit
   - use a const reference getter for the script interface
   - partial fix for #2628
- narrow includes of `Observable_stat.hpp`
- add ELC/DLC energy corrections to the P3M/DP3M/DDS solver energy slots
   - the `analyze.energy()`, `analyze.pressure()`, `analyze.pressure_tensor()` functions now only reports 2 slots for Coulomb and dipolar contributions: particles contribution and electrostatics/magnetostatics solver contribution

### DIFF
--- a/src/core/cuda_common_cuda.cu
+++ b/src/core/cuda_common_cuda.cu
@@ -22,7 +22,6 @@
 
 #include "debug.hpp"
 
-#include "Observable_stat.hpp"
 #include "ParticleRange.hpp"
 #include "cuda_init.hpp"
 #include "cuda_interface.hpp"
@@ -221,15 +220,12 @@ void clear_energy_on_GPU() {
   cuda_safe_mem(cudaMemset(energy_device, 0, sizeof(CUDA_energy)));
 }
 
-void copy_energy_from_GPU(Observable_stat &obs_energy) {
+CUDA_energy copy_energy_from_GPU() {
   if (!global_part_vars_host.communication_enabled)
-    return;
+    return {};
   cuda_safe_mem(cudaMemcpy(&energy_host, energy_device, sizeof(CUDA_energy),
                            cudaMemcpyDeviceToHost));
-  if (!obs_energy.coulomb.empty())
-    obs_energy.coulomb[1] += energy_host.coulomb;
-  if (obs_energy.dipolar.size() >= 2)
-    obs_energy.dipolar[1] += energy_host.dipolar;
+  return energy_host;
 }
 
 void _cuda_safe_mem(cudaError_t CU_err, const char *file, unsigned int line) {
--- a/src/core/cuda_common_cuda.cu
+++ b/src/core/cuda_common_cuda.cu
@@ -221,8 +221,7 @@ void clear_energy_on_GPU() {
   cuda_safe_mem(cudaMemset(energy_device, 0, sizeof(CUDA_energy)));
 }
 
-void copy_energy_from_GPU() {
-  extern Observable_stat obs_energy;
+void copy_energy_from_GPU(Observable_stat &obs_energy) {
   if (!global_part_vars_host.communication_enabled)
     return;
   cuda_safe_mem(cudaMemcpy(&energy_host, energy_device, sizeof(CUDA_energy),
--- a/src/core/cuda_common_cuda.cu
+++ b/src/core/cuda_common_cuda.cu
@@ -227,7 +227,7 @@ void copy_energy_from_GPU(Observable_stat &obs_energy) {
   cuda_safe_mem(cudaMemcpy(&energy_host, energy_device, sizeof(CUDA_energy),
                            cudaMemcpyDeviceToHost));
   if (!obs_energy.coulomb.empty())
-    obs_energy.coulomb[0] += energy_host.coulomb;
+    obs_energy.coulomb[1] += energy_host.coulomb;
   if (obs_energy.dipolar.size() >= 2)
     obs_energy.dipolar[1] += energy_host.dipolar;
 }
--- a/src/core/electrostatics_magnetostatics/coulomb.cpp
+++ b/src/core/electrostatics_magnetostatics/coulomb.cpp
@@ -42,8 +42,7 @@
 Coulomb_parameters coulomb;
 
 namespace Coulomb {
-void calc_pressure_long_range(Observable_stat &virials,
-                              Observable_stat &p_tensor,
+void calc_pressure_long_range(Observable_stat &p_tensor,
                               const ParticleRange &particles) {
   switch (coulomb.method) {
 #ifdef P3M
@@ -60,7 +59,6 @@ void calc_pressure_long_range(Observable_stat &virials,
     p3m_charge_assign(particles);
     auto const p3m_p_tensor = p3m_calc_kspace_pressure_tensor();
     std::copy_n(p3m_p_tensor.data(), 9, p_tensor.coulomb.begin() + 9);
-    virials.coulomb[1] = p3m_p_tensor[0] + p3m_p_tensor[4] + p3m_p_tensor[8];
 
     break;
   }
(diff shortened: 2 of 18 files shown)


src/python/espressomd/analyze.pyx
```diff
@@ -32,12 +33,22 @@ from .utils import array_locked, is_valid_type
...
-    """Transform a scalar Observable_stat struct to a python dict.
+cdef _Observable_stat_to_dict(Observable_stat obs, int size,
+                              cbool calc_sp):
+    """Transform an Observable_stat struct to a python dict.
+
+    Parameters
+    ----------
+    obs :
+        Core observable.
+    size : :obj:`int`, \{1, 9\}
```
fweik: This is already contained in `obs` right?

src/core/cuda_common_cuda.cu
```diff
@@ -221,8 +221,7 @@ void clear_energy_on_GPU() {
   cuda_safe_mem(cudaMemset(energy_device, 0, sizeof(CUDA_energy)));
 }
 
-void copy_energy_from_GPU() {
-  extern Observable_stat obs_energy;
+void copy_energy_from_GPU(Observable_stat &obs_energy) {
```
fweik: I think it would be easier to return the gpu energy struct from the function, that way, the cuda interface code can be completely decoupled from the obssttat code, and avoid an output parameter.

src/core/electrostatics_magnetostatics/coulomb.cpp
```diff
@@ -42,8 +42,7 @@
 Coulomb_parameters coulomb;
 
 namespace Coulomb {
-void calc_pressure_long_range(Observable_stat &virials,
-                              Observable_stat &p_tensor,
+void calc_pressure_long_range(Observable_stat &p_tensor,
```
fweik: Also here I think you can just return the value (or multiple values, if you want to keep the elc accounting, I have dropped that).

src/core/energy.cpp
```diff
@@ -65,13 +67,14 @@ void energy_calc(const double time) {
   on_observable_calc();
 
   for (auto const &p : cell_structure.local_particles()) {
-    add_kinetic_energy(p);
+    add_kinetic_energy(p, obs_energy);
```
fweik: Here also just return the value

fweik (changes requested):
Clear improvement. I think some of the output parameters should be avoided in the cases where that is simple, see inline comments. This helps to reduce coupling.

src/core/electrostatics_magnetostatics/coulomb.cpp
jngrad: It would be convenient if we could also drop the separate counting of the dipolar correction slot. @RudolfWeeber what do you think about adding the DLC energy correction from `add_mdlc_energy_corrections()` directly into the slot where the DP3M/DS energy is stored? This mechanism for storing ELC and DLC corrections in a separate slot of `Observable_stat` looks like a debug feature to me.

fweik: I agree, but you can also just return an vector for now and we sort this out later.

jngrad: Well that would be a `Vector2d` because we cannot overwrite the first slot, which comes from particles. This means adding a for loop with an offset or manually assigning the two values. If we remove the ELC slot, it makes sense to remove the DLC too, to make the API change consistent. IIRC Rudolf only needed two slots.

src/python/espressomd/analyze.pyx
jngrad: yes, I just need to write a getter for the private member `m_chunk_size`

src/core/cuda_common_cuda.cu
jngrad: good idea

src/core/electrostatics_magnetostatics/coulomb.cpp
fweik: I'd return an array with 3 values, because there are three slots. Otherwise the concerns are not separated because the caller has to know which slot is which/which one is set

src/python/espressomd/analyze.pyx
jngrad: Why do we store Coulomb energies in the particle slot? Shouldn't it be stored in the electrostatics solver slot, as it is done for dipolar energies?
https://github.com/espressomd/espresso/blob/35e18f7167f95f75057384993e0ef771bc061b12/src/core/cuda_interface.cpp#L204-L207

src/core/electrostatics_magnetostatics/coulomb.cpp
jngrad: This means we need a `for` loop to add the `Vector3d` values instead of doing a copy.

src/python/espressomd/analyze.pyx
fweik: I happen to know that it works if you change that 0 to 1. I think it's also more correct, because it is an aggregated term.

src/core/electrostatics_magnetostatics/coulomb.cpp
fweik: `boost::copy(calc_pressure_long_range(...), obs_pressure.coulomb)` ?

jngrad: wouldn't that overwrite the first slot of `obs_energy.coulomb`, which comes from the short-range loop?

fweik: Ah I see what you mean... Would be easier if `Observable_stat` would have a proper value type :D

src/python/espressomd/analyze.pyx
jngrad: I'll do the change for consistency with our slot convention, even if it changes nothing to the actual results (P3M energy is not implemented on GPU).

src/core/electrostatics_magnetostatics/coulomb.hpp
```diff
@@ -60,8 +60,7 @@ struct Coulomb_parameters {
 extern Coulomb_parameters coulomb;
 
 namespace Coulomb {
-void calc_pressure_long_range(Observable_stat &virials,
-                              Observable_stat &p_tensor,
+void calc_pressure_long_range(Observable_stat &p_tensor,
```
fweik: You should be able to remove the include for `Observable_stat` now

src/core/electrostatics_magnetostatics/dipole.hpp
```diff
@@ -80,8 +80,7 @@ void init();
 
 void calc_long_range_force(const ParticleRange &particles);
 
-void calc_energy_long_range(Observable_stat &energy,
-                            const ParticleRange &particles);
+double calc_energy_long_range(const ParticleRange &particles);
```
fweik, line 81: Also I think in this file the include can be removed.

fweik (changes requested):
I think you can remove some more includes so that we also have physical decoupling

src/core/electrostatics_magnetostatics/coulomb.hpp
jngrad: ah right, forgot about the pressure observable...

fweik: This should work the same way as energy, and then there should be no mention of Obsstat anymore...

fweik (approved):
LGTM

PR #3776, opened 2020-06-26
Reaction ensemble sample with complex reaction and analytical solution

jonaslandsgesell:
@jngrad could be interesting for you if you plan to refactor some code in the reaction ensemble module


jonaslandsgesell:
Hey Jean-Noel, instead of guarding sympy I use scipy.fsolve in the other PR now: #3778, I think this is better and we can close this PR here.

PR #3779, opened 2020-06-26
Reaction ensemble maintainance 

jngrad:
Description of changes:
- modernize code
- remove unused functions
- unit test helper functions


src/core/reaction_ensemble.hpp
```diff
@@ -389,9 +389,64 @@ class WidomInsertion : public ReactionAlgorithm {
...
+  if (counter_allowed_entries) {
+    return static_cast<double>(result) / counter_allowed_entries;
+  }
+  return 0.0;
+}
+
+/**
+ * Checks whether a number is in an array.
+ */
+inline bool is_in_vector(int value, std::vector<int> const &rng) {
```
fweik: That is `Utils::contains`

src/core/reaction_ensemble.hpp
```diff
@@ -389,9 +389,64 @@ class WidomInsertion : public ReactionAlgorithm {
...
+  return boost::range::find(rng, value) != rng.end();
+}
+
+/**
+ * Finds the minimum non negative value in the provided double array and returns
+ * this value.
+ */
+inline double find_minimum_non_negative_value(std::vector<double> const &rng) {
```
fweik, line 430: This should probably be formulated as an accumulate. This implementation is UB if `rng` is empty, which needs to be checked.

fweik (changes requested):

fweik (dismissed):
LGTM

jngrad:
coyote 11 with 3 CI jobs running in parallel:
```
45/148 Test #101: wang_landau_reaction_ensemble .............................   Passed  286.36 sec
```
The test is still going to time out even on the most recent machines under a small load. We have to remove it from the coverage.

fweik:
The coverage matters because it has different optimizer flags?

jngrad:
This plays a role. Although in `cuda9-maxset`, the test is often 3 times faster. Maybe the GPU tests running at the same time as Wang-Landau take some load off the CPU?

jngrad:
here are the relevant optimization flags:
```
cuda10-maxset
-O3 -g -fPIC   -pthread [-Wall and friends...] 
default
-Og -g -fPIC   -pthread [-Wall and friends...] -g --coverage -fprofile-arcs -ftest-coverage
cuda9-maxset
-Og -g -fPIC   -pthread [-Wall and friends...] -g --coverage -fprofile-arcs -ftest-coverage
```
The custom test timeout values in `.gitlab-ci.yml` don't seem to have been updated in a while. For example, `cuda9-maxset` has 900s timeout instead of the default 300s, while its tests take at most 60s.

fweik (approved):

PR #3747, opened 2020-06-03
Document BoxGeometry-related functions

jngrad:
Fixes #3697

Description of changes:
- document `get_mi_coord()`, `get_mi_vector()` and related functions
- add unit test for `fold_position()`

PR #3780, opened 2020-06-26
Topology

fweik:
The goal of this PR is to have better encapsulation for the cell systems (now dubbed `particle decompositions`).
To this end, an interface `ParticleDecomposition` is introduced, and the two currently implemented cell systems
are introduced as implementations of this interface. The cell system globals are removed and ownership for the
cell sytems now rests with `CellStructure`, which also maintains the local particle index. To be able to split the
index bookkeeping from the particle decompositions, the latter report back the change in the particle content
of the node on resort.
To make this change a bit easier, the decomposition is always reset when one of the parameters affecting it
is changed (interaction rang, node grid, box size, ...). This has some minor performance impact for NpT and
when setting up many interactions (many changes to the range), the latter can be fixed later, because range
changes should be considered in an lazy fashing only when actually calculating the pairs.


fweik:
@jngrad Could you maybe give me some feedback where additional documentation is needed? Also if there is not too much opposition, I'd like to squash the commits, I don't think the history is interesting enough to warrant the work needed to make
this a comprehensible story. Most commits were just undo points...

jngrad:
I'm fine with you rebasing this PR.

src/core/CellStructure.hpp
```diff
@@ -297,10 +282,21 @@ struct CellStructure {
    */
   void remove_all_particles();
 
-private:
-  /** One of @ref Cells::Resort, announces the level of resort needed.
+  /**
+   * @brief Get the underlyind particle decomposition.
```
jngrad:
```suggestion
   * @brief Get the underlying particle decomposition.
```

src/core/CellStructure.hpp
```diff
@@ -377,6 +386,69 @@ struct CellStructure {
...
+  void set_atom_decomposition(boost::mpi::communicator const &comm,
+                              BoxGeometry const &box);
+
+  /**
+   * @brief Set the particle decomposition to
+   *        DomainDecomposition.
+   *
+   *        @param comm Cartesian communicator to use.
+   *        @param box Box Geometry
+   *        @param local_geo Geoemtry of the local box.
```
jngrad:
```suggestion
   *        @param local_geo Geometry of the local box.
```

src/core/DomainDecomposition.hpp
```diff
@@ -0,0 +1,208 @@
...
+   *  @brief Calculate cell grid dimensions, cell sizes and number of cells.
+   *
+   *  Calculates the cell grid, based on the local box size and the range.
+   *  If the number of cells is larger than max_num_cells,
+   *  it increases max_range until the number of cells is
+   *  smaller or equal max_num_cells. It sets:
+   *  cell_grid,
+   *  ghost_cell_grid,
+   *  cell_size, and
+   *  inv_cell_size.
```
jngrad, line 179: suggestion: here and in other places in this file, it could be beneficial to prefix global variable names by `@ref`. This is not just for formatting the doxygen html output, but also for CI: if these globals are removed in the future, out-of-date docstrings will be flagged in the doxygen style job.

src/core/ParticleDecomposition.hpp
```diff
@@ -0,0 +1,107 @@
...
+ * near-by can be calculated. Related to this it provides descriptions
+ * of the ghost communications by which particles can be synchronized that
+ * are not owned locally, but interact with local particles.
+ */
+class ParticleDecomposition {
+public:
+  /**
+   * @brief Resort particles.
+   *
+   * After calling this function, every particle is in it's home cell.
```
jngrad:
```suggestion
   * After calling this function, every particle is in its home cell.
```

src/core/ParticleDecomposition.hpp
```diff
@@ -0,0 +1,107 @@
+#ifndef ESPRESSO_PARTICLE_DECOMPOSITION_HPP
```
jngrad, line 19: missing license header

src/core/cells.hpp
```diff
@@ -137,4 +123,12 @@ void check_resort_particles();
  */
 Cell *find_current_cell(const Particle &p);
 
+/**
+ * @brief Return a pointer to the global DomainDecomposition.
+ *
+ * @return Pointer to the decomposition if it is settt and is
```
jngrad:
```suggestion
 * @return Pointer to the decomposition if it is set and is
```

src/core/DomainDecomposition.hpp
```diff
@@ -0,0 +1,208 @@
...
+   *
+   *  @param range Required interacting range. All pairs closer
+   *         than this distance are found.
+   */
+  void create_cell_grid(double range);
+
+  /** Init cell interactions for cell system domain decomposition.
+   * initializes the interacting neighbor cell list of a cell The
+   * created list of interacting neighbor cells is used by the Verlet
+   * algorithm.
```
jngrad:
```suggestion
   *  Initializes the interacting neighbor cell list of a cell.
   *  This list of interacting neighbor cells is used by the Verlet
   * algorithm.
```

jngrad (changes requested):
Here is the result of a quick look through the docs. Please make sure the improvements from #3747 are correctly recovered when you merge python into this branch; several functions whose docstrings were touched in in #3747 have been moved to new files.

fweik:
Thank you, I'll look into your suggestions.

src/core/DomainDecomposition.hpp
fweik: Those are not globals but class members. I had to remove the `@ref` for class members, because doxygen seems to not understand the scope.

jngrad: I should have paid more attention to the code :) I think for class members one needs to do something like `@ref class_name::member_name "member_name"`, but it's unreadable and prone to typos that won't get caught in the CI style job, so I avoid to create such refs unless I really need to have a link in the html output.

src/core/DomainDecomposition.hpp
```diff
@@ -0,0 +1,208 @@
...
+ *  \image html  linked_cells.gif "Linked cells structure"
+ *
+ *  2D representation of a linked cell grid: cell_grid =
+ *  {4,4}, ghost_cell_grid = {6,6}
+ *
+ * Each cell has 3^D neighbor cells (For cell 14 they are
+ * marked). Since we deal with pair forces, it is sufficient to
+ * calculate only half of the interactions (Newtons law: action =
+ * reaction). We have chosen the upper half e.g. all neighbor cells with
+ * a higher linear index (For cell 14 they are marked in light
```
jngrad:
```suggestion
 * Each cell has 3^D neighbor cells. Since we deal with pair forces, it is
 * sufficient to calculate only half of the interactions (Newton's law:
 * action = reaction). We have chosen the upper half e.g. all neighbor
 * cells with a higher linear index (for cell 14 they are marked in light
```
remove typo and text duplication

fweik:
@RudolfWeeber can you please have a preliminary look at this?

jngrad (changes requested):
one last regression missing from #3747

fweik:
Thanks, @jngrad, missed that one.

src/core/DomainDecomposition.cpp
```diff
@@ -0,0 +1,582 @@
...
+  auto const cart_info = Utils::Mpi::cart_get<3>(m_comm);
+
+  int i, n_local_cells, new_cells;
+  double cell_range[3];
+
+  /* initialize */
+  cell_range[0] = cell_range[1] = cell_range[2] = range;
+
+  /* Min num cells can not be smaller than calc_processor_min_num_cells. */
+  int min_num_cells = calc_processor_min_num_cells();
```
RudolfWeeber, line 283: doesn't this have to be multiplied by the number of mpi ranks?

src/core/DomainDecomposition.cpp
```diff
@@ -0,0 +1,582 @@
...
+     since this also ensures that the cell size is at most half the box
+     length. However, if there is only one processor for a direction, there
+     have to be at least two cells for this direction. */
+  return boost::accumulate(Utils::Mpi::cart_get<3>(m_comm).dims, 1,
+                           [](int n_cells, int grid) {
+                             return (grid == 1) ? 2 * n_cells : n_cells;
+                           });
+}
+
+void DomainDecomposition::create_cell_grid(double range) {
```
RudolfWeeber, line 273: It might be worthwhile to pull out the algorithm which calculates the local cell grid as a function of the comm cart shape and the ia range, such that it can be unit tested.

RudolfWeeber (commented):
This is a big improvement in clarity for the particle storage.

src/core/DomainDecomposition.cpp
fweik: I think this is per node, I'll check and add a comment

src/core/DomainDecomposition.cpp
fweik: I agree that this should be tested, but to do this we first have to find out what the expected behavior is :-) I'm not sure I understand all of the logic...

RudolfWeeber:
As discussed elswhere, I timed the cell system update due to interaction creation.
For a 100k particle system, creating 100 bonded ia takes some 4 secs. 
Since this affects a rather narrow class of simulations and can be worked around by swapping the order of bond and particle creation, we can choose to live with this temporarily, but the lazy cell system update needs to be pursued, I think.

RudolfWeeber (approved):

fweik:
Ok, I'll see that I include the lazy parameter update into the other PR with priority. I think this is the better solution anyway, because it is one less thing that can be messed up.

src/core/DomainDecomposition.cpp
fweik: All of this is per node, from my understanding the code is correct.

RudolfWeeber:
@jfrad, is this resolved? If so, this Pr could be merged, I think.

RudolfWeeber:
@jngrad. Mistyped in previous comment.

jngrad (approved):

PR #3781, opened 2020-06-27
Observables and correlators/accumulators maintainance

jngrad:
Description of changes:
- fix broken tensor_product correlator (only returned zeros in 4.1)
- add more correlators and accumulators test cases
- test script interface getters/setters for observables, correlators and accumulators
- make code coverage more reproducible

### DIFF
--- a/src/core/accumulators/Correlator.cpp
+++ b/src/core/accumulators/Correlator.cpp
@@ -65,7 +65,7 @@ std::vector<double> compress_discard2(std::vector<double> const &A1,
 
 std::vector<double> scalar_product(std::vector<double> const &A,
                                    std::vector<double> const &B,
-                                   Utils::Vector3d) {
+                                   Utils::Vector3d const &) {
   if (A.size() != B.size()) {
     throw std::runtime_error(
         "Error in scalar product: The vector sizes do not match");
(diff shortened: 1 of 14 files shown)


src/core/accumulators/Correlator.cpp
```diff
@@ -91,13 +91,13 @@ std::vector<double> componentwise_product(std::vector<double> const &A,
...
                                    std::vector<double> const &B,
-                                   Utils::Vector3d) {
+                                   Utils::Vector3d const &) {
   std::vector<double> C(A.size() * B.size());
   auto C_it = C.begin();
 
-  for (auto A_it = A.begin(); A_it != A.begin(); ++A_it) {
-    for (double B_it : B) {
-      *(C_it++) = *A_it * B_it;
+  for (double a : A) {
```
fweik, line 98: Is there a specific reason why you can't just say `boost::transform(A, B, C.begin(), std::multiplies<>{})`?
(In this case a would even consider using an `std::back_inserter` instead of explicitly resizing the vector,
and doing `boost::transform(A, B, std::back_inserter(C), std::multiplies<>{})`)

fweik: Ah sorry, it really is the tensor product.  It's too late.

fweik (approved):

PR #3784, opened 2020-06-30
ScaFaCoS maintenance

jngrad:
Description of changes:
- fix a bug in the ScaFaCoS interface when calling the `get_params()` method of a ScaFaCos-based actor
- add a test for the ScaFaCos `get_params()` method
- remove unused ScaFaCoS function
- document and test FCS weights attribute
- improve testsuite readability

### DIFF
--- a/src/python/espressomd/scafacos.pyx
+++ b/src/python/espressomd/scafacos.pyx
@@ -45,13 +45,13 @@ IF SCAFACOS == 1:
             self._set_params_in_es_core()
 
         def valid_keys(self):
-            tmp = ["method_name", "method_params", "prefactor"]
+            tmp = self.required_keys().copy()
             if not self.dipolar:
-                tmp.append("check_neutrality")
+                tmp.add("check_neutrality")
             return tmp
 
         def required_keys(self):
-            return "method_name", "method_params", "prefactor"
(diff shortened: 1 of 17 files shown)


src/python/espressomd/scafacos.pyx
```diff
@@ -45,13 +45,13 @@ IF SCAFACOS == 1:
             self._set_params_in_es_core()
 
         def valid_keys(self):
-            tmp = ["method_name", "method_params", "prefactor"]
+            tmp = self.required_keys().copy()
```
fweik, line 48: Couldn't you do `return self.required_keys().copy().add("check_neutrality")`?

fweik: I mean, wrapped in an `if` of course

jngrad: I just tried it, but the `add()` method of `set` returns None.

jngrad: and we also can't do `return self.required_keys().copy() + {"check_neutrality"}`

fweik: Not so fluent :-)

fweik (approved):

PR #3785, opened 2020-07-01
h5md: cleanup.

KaiSzuttor:
this is more an exploration for future work on the h5md stuff to find out what needs to be implemented in ```h5xx``` to remove the low-level hdf5 API usage and to simplify further extensions to the code.

### DIFF
--- a/doc/sphinx/io.rst
+++ b/doc/sphinx/io.rst
@@ -142,33 +142,20 @@ respective hdf5-file. This may, for example, look like:
     from espressomd.io.writer import h5md
     system = espressomd.System(box_l=[100.0, 100.0, 100.0])
     # ... add particles here
-    h5 = h5md.H5md(filename="trajectory.h5", write_pos=True, write_vel=True)
+    h5 = h5md.H5md(filename="trajectory.h5")
 
-If a file with the given filename exists and has a valid H5MD structures,
+An optional argument to the constructor of :class:`espressomd.io.writer.h5md.H5md` is
+an instance of :class:`espressomd.io.writer.h5md.UnitSystem` which encapsulates 
+physical units for time, mass, length and electrical charge.	
+
+If a file with the given filename exists and has a valid H5MD structure,
 it will be backed up to a file with suffix ".bak". This backup file will be
 deleted when the new file is closed at the end of the simulation with
 ``h5.close()``.
 
-The current implementation allows to write the following properties: positions,
-velocities, forces, species (|es| types), and masses of the particles. In order
-to write any property, you have to set the respective boolean flag as an option
-to the :class:`~espressomd.io.writer.h5md.H5md` class. Currently available:
+The current implementation allways writes the following properties: positions,
+velocities, forces, species (|es| types), charges, and masses of the particles.
 
-    - ``write_pos``: particle positions
-
-    - ``write_vel``: particle velocities
-
-    - ``write_force``: particle forces
-
-    - ``write_species``: particle types
-
-    - ``write_mass``: particle masses
-
-    - ``write_ordered``: if particles should be written ordered according to their
-      id (implies serial write)
-
-    - ``unit_system``: optionally, physical units for time, mass, length and
-      electrical charge.
 
 In simulations with varying numbers of particles (MC or reactions), the
 size of the dataset will be adapted if the maximum number of particles
--- a/doc/sphinx/io.rst
+++ b/doc/sphinx/io.rst
@@ -142,20 +142,19 @@ respective hdf5-file. This may, for example, look like:
     from espressomd.io.writer import h5md
     system = espressomd.System(box_l=[100.0, 100.0, 100.0])
     # ... add particles here
-    h5 = h5md.H5md(filename="trajectory.h5")
+    h5 = h5md.H5md(file_path="trajectory.h5")
 
 An optional argument to the constructor of :class:`espressomd.io.writer.h5md.H5md` is
 an instance of :class:`espressomd.io.writer.h5md.UnitSystem` which encapsulates 
 physical units for time, mass, length and electrical charge.	
 
-If a file with the given filename exists and has a valid H5MD structure,
+If a file at the given filepath exists and has a valid H5MD structure,
 it will be backed up to a file with suffix ".bak". This backup file will be
 deleted when the new file is closed at the end of the simulation with
 ``h5.close()``.
 
 The current implementation allways writes the following properties: positions,
 velocities, forces, species (|es| types), charges, and masses of the particles.
-
 
 In simulations with varying numbers of particles (MC or reactions), the
 size of the dataset will be adapted if the maximum number of particles
--- a/doc/sphinx/io.rst
+++ b/doc/sphinx/io.rst
@@ -153,7 +153,7 @@ it will be backed up to a file with suffix ".bak". This backup file will be
 deleted when the new file is closed at the end of the simulation with
 ``h5.close()``.
 
-The current implementation allways writes the following properties: positions,
+The current implementation always writes the following properties: positions,
 velocities, forces, species (|es| types), charges, and masses of the particles.
 
 In simulations with varying numbers of particles (MC or reactions), the
@@ -163,11 +163,11 @@ be written to the trajectory for the id. If you have a parallel
 simulation, please keep in mind that the sequence of particles in general
 changes from timestep to timestep. Therefore you have to always use the
 dataset for the ids to track which position/velocity/force/type/mass
-entry belongs to which particle. To write data to the hdf5 file, simply
-call the H5md object :meth:`~espressomd.io.writer.h5md.H5md.write` method without any arguments.
+entry belongs to which particle. To write data to the HDF5 file, simply
+call the method :meth:`~espressomd.io.writer.h5md.H5md.write` without any arguments.
 
-After the last write call, you have to call the
-:meth:`~espressomd.io.writer.h5md.H5md.close` method to remove
+After the last write, you have to call
+:meth:`~espressomd.io.writer.h5md.H5md.close` to remove
 the backup file, close the datasets, etc.
 
 H5MD files can be read and modified with the python module h5py (for
--- a/src/core/io/writer/h5md_core.cpp
+++ b/src/core/io/writer/h5md_core.cpp
@@ -37,18 +37,16 @@ using Vector2hs = Utils::Vector<hsize_t, 2>;
 using Vector3hs = Utils::Vector<hsize_t, 3>;
 
 static void backup_file(const std::string &from, const std::string &to) {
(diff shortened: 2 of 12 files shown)


src/core/io/writer/h5md_core.cpp
```diff
@@ -48,56 +49,42 @@ static void backup_file(const std::string &from, const std::string &to) {
...
-  auto world = boost::mpi::communicator();
-
-  if (m_write_ordered)
-    m_hdf5_comm = world.split(world.rank(), 0);
-  else
-    m_hdf5_comm = world;
-
-  if (m_write_ordered && this_node != 0)
-    return;
+  m_hdf5_comm = boost::mpi::communicator();
```
fweik: Since this is fixed now you can initialize it in the constructor, should also probably be a passed in, so that the client can chose which nodes participate...

fweik:
Did we decide to do the API break with the sorting? @jngrad @RudolfWeeber any opinion on that? I find' it hard to estimate what this will do to existing analysis code...

fweik:
I think it is ok, because it is breaking (because the keyword was removed).

KaiSzuttor:
@RudolfWeeber could you please have a look if the code is ok now?

src/core/io/writer/h5md_core.cpp
```diff
@@ -47,562 +52,340 @@ static void backup_file(const std::string &from, const std::string &to) {
...
   m_absolute_script_path = boost::filesystem::canonical(script_path);
-  init_filestructure();
-  bool file_exists = boost::filesystem::exists(m_filename);
+  bool file_exists = boost::filesystem::exists(m_file_path);
   bool backup_file_exists = boost::filesystem::exists(m_backup_filename);
   /* Perform a barrier synchronization. Otherwise one process might already
    * create the file while another still checks for its existence. */
-  if (!m_write_ordered)
-    MPI_Barrier(m_hdf5_comm);
+  MPI_Barrier(m_comm);
```
fweik: If you use the boost mpi functions (`m_comm.barrier()`) mpi errors are checked

src/core/io/writer/h5md_core.cpp
```diff
@@ -47,562 +52,340 @@ static void backup_file(const std::string &from, const std::string &to) {
...
-    n_part = partCfg.size();
-  } else if (m_write_ordered && this_node != 0)
-    return;
-  else if (!m_write_ordered) {
-    num_particles_to_be_written = particles.size();
-    n_part = boost::mpi::all_reduce(m_hdf5_comm, num_particles_to_be_written,
-                                    std::plus<int>());
-  }
+void File::create_new_file(const std::string &filename, BoxGeometry &geometry) {
+  if (this_node == 0)
```
fweik: The `this_node` global can be dropped (`m_comm.rank()`), please do also remove the corresponding header to check that you catch all places.

src/core/io/writer/h5md_core.cpp
```diff
@@ -47,562 +52,340 @@ static void backup_file(const std::string &from, const std::string &to) {
...
-    num_particles_to_be_written = partCfg.size();
-    n_part = partCfg.size();
-  } else if (m_write_ordered && this_node != 0)
-    return;
-  else if (!m_write_ordered) {
-    num_particles_to_be_written = particles.size();
-    n_part = boost::mpi::all_reduce(m_hdf5_comm, num_particles_to_be_written,
-                                    std::plus<int>());
-  }
+void File::create_new_file(const std::string &filename, BoxGeometry &geometry) {
```
fweik: `geometry` seems unused

src/core/io/writer/h5md_core.cpp
```diff
@@ -47,562 +52,340 @@ static void backup_file(const std::string &from, const std::string &to) {
...
+  if (this_node == 0)
+    write_script(filename, m_absolute_script_path);
+  MPI_Barrier(m_comm);
```
fweik: `m_comm.barrier()`

src/core/io/writer/h5md_core.cpp
```diff
@@ -47,562 +52,340 @@ static void backup_file(const std::string &from, const std::string &to) {
...
-
-    for (auto &current_particle : particles) {
-      fill_arrays_for_h5md_write_with_particle_property(
-          particle_index, id, typ, mass, pos, image, vel, f, charge,
-          current_particle, write_dat, bond);
-      particle_index++;
-    }
-  }
+void File::close() {
+  if (this_node == 0)
```
fweik: `m_comm.rank()`

src/core/io/writer/h5md_core.cpp
```diff
@@ -47,562 +52,340 @@ static void backup_file(const std::string &from, const std::string &to) {
...
-  H5Sselect_hyperslab(ds, H5S_SELECT_SET, offset, nullptr, count, nullptr);
-  /* Create a temporary dataspace. */
-  hid_t ds_new = H5Screate_simple(rank, count, maxdims.data());
-  /* Finally write the data to the dataset. */
-  H5Dwrite(dataset.hid(), dataset.get_type(), ds_new, ds, H5P_DEFAULT,
-           data.origin());
-  H5Sclose(ds_new);
-  H5Sclose(ds);
-}
+void File::write(const ParticleRange &particles, double time, int step) {
```
fweik: You also have to take the box geometry as a parameter here, then drop the `box_geo` global and the corresponding header include.

src/core/io/writer/h5md_core.cpp
```diff
@@ -47,562 +52,340 @@ static void backup_file(const std::string &from, const std::string &to) {
...
-  /* Clean up. */
-  H5Dclose(dset);
-  H5Sclose(space);
-  H5Tclose(dtype);
-  H5Fclose(file_id);
+  int const n_part_local = particles.size();
+  // calculate count and offset
+  int prefix = 0;
+  // calculate prefix for write of the current process
+  MPI_Exscan(&n_part_local, &prefix, 1, MPI_INT, MPI_SUM, m_comm);
```
fweik: I think `exscan` is not available from boost.mpi, but it provides a macro for error checking, like so
```c++
BOOST_MPI_CHECK_RESULT(MPI_Exscan, (&n_part_local, &prefix, 1, MPI_INT, MPI_SUM, m_comm));
```

src/core/io/writer/h5md_core.cpp
```diff
@@ -47,562 +52,340 @@ static void backup_file(const std::string &from, const std::string &to) {
...
-void File::Flush() {
-  if (m_write_ordered) {
-    if (this_node == 0)
-      H5Fflush(m_h5md_file.hid(), H5F_SCOPE_GLOBAL);
-  } else
-    H5Fflush(m_h5md_file.hid(), H5F_SCOPE_GLOBAL);
+  int n_bonds_local = bond.shape()[1];
+  int n_bonds_total = n_bonds_local;
+  int prefix_bonds = 0;
+  MPI_Exscan(&n_bonds_local, &prefix_bonds, 1, MPI_INT, MPI_SUM, m_comm);
```
fweik: Check for MPI errors

src/core/io/writer/h5md_core.cpp
```diff
@@ -47,562 +52,340 @@ static void backup_file(const std::string &from, const std::string &to) {
...
+  MPI_Allreduce(&n_bonds_local, &n_bonds_total, 1, MPI_INT, MPI_SUM, m_comm);
```
fweik: Please use boost.mpi, it has an overload which returns the accumulated value, so it can be const like so
```c++
auto const n_bond_total = boost::mpi::all_reduce(...);
```

src/core/io/writer/h5md_core.hpp
```diff
@@ -26,186 +26,185 @@
 #include <boost/mpi/communicator.hpp>
 #include <h5xx/h5xx.hpp>
 
+#include <BoxGeometry.hpp>
 #include <algorithm>
 #include <string>
 #include <unordered_map>
+#include <utility>
 
 #include "PartCfg.hpp"
```
fweik: I think this is no longer needed.

src/core/io/writer/h5md_core.hpp
```diff
@@ -26,186 +26,185 @@
...
+   */
+  void write(const ParticleRange &particles, double time, int step);
+
   /**
-   * @brief Method to force flush to h5md file.
+   * @brief Retrieve the path to the hdf5 file.
+   * @return The path as a string.
    */
-  void Flush();
+  std::string &file_path() { return m_file_path; };
```
fweik: Shouldn't this be a `const &`, or is the client supposed to change the path after the file has been created?

src/core/io/writer/h5md_core.hpp
```diff
@@ -26,186 +26,185 @@
...
 
-private:
-  boost::mpi::communicator m_hdf5_comm;
-  bool m_already_wrote_bonds = false;
+  /**
+   * @brief Retrieve the path to the simulation script.
+   * @return The path as a string.
+   */
+  std::string &script_path() { return m_script_path; };
```
fweik, line 102: `const &`?

fweik (changes requested):
Overall this is a huge improvement in readability. I think the design is good, but I think some changes in details could improve the code further.

doc/sphinx/io.rst
```diff
@@ -142,33 +142,19 @@ respective hdf5-file. This may, for example, look like:
...
-    - ``write_species``: particle types
-
-    - ``write_mass``: particle masses
-
-    - ``write_ordered``: if particles should be written ordered according to their
-      id (implies serial write)
-
-    - ``unit_system``: optionally, physical units for time, mass, length and
-      electrical charge.
+The current implementation allways writes the following properties: positions,
```
jngrad:
```suggestion
The current implementation always writes the following properties: positions,
```

testsuite/python/h5md.py
```diff
@@ -92,157 +127,99 @@ def test_metadata(self):
...
+            self.assertEqual(	
+                self.py_file['particles/atoms/charge/value'].attrs['unit'], b'e')	
+        if espressomd.has_features(['MASS']):	
+            self.assertEqual(	
+                self.py_file['particles/atoms/mass/value'].attrs['unit'], b'u')	
+        self.assertEqual(	
+            self.py_file['particles/atoms/force/value'].attrs['unit'],	
+            b'm u ps-2')	
+        self.assertEqual(	
+            self.py_file['particles/atoms/velocity/value'].attrs['unit'],
```
jngrad: trailing whitespaces

KaiSzuttor: that comes from automatic formatting. I don't know why this happened.

fweik (approved):
LGTM

PR #3789, opened 2020-07-03
add description for VS Code setup

biermanncarl:
As per request of @RudolfWeeber I added a description on how to use ESPResSo inside Visual Studio Code.


fweik:
I think we should maybe make it clear that this is only relevant if the user _wants_ to use VS Code. I think this could maybe be a bit too specific for installation docs, @jngrad what do you think?

KaiSzuttor:
it's also prone to get incorrect pretty fast

jngrad:
> I think we should maybe make it clear that this is only relevant if the user _wants_ to use VS Code. I think this could maybe be a bit too specific for installation docs, @jngrad what do you think?

I would say the section [Development](https://github.com/espressomd/espresso/wiki/Development) in our wiki would be a better fit than the user guide. The user guide actually links to the wiki in the [Contributing](http://espressomd.org/html/doc/contributing.html#contributing-your-own-code) page. The wiki could provide a subsection on setting up IDEs, e.g.:
```markdown
# Configuring the IDE
## VS Code
[here the content of this PR]
```

> it's also prone to get incorrect pretty fast

True. Moving this VS Code contribution to the wiki would allow anyone in the organization to change the text without the need for code review, which makes it easy to update the text when the IDE changes its interface. Mentioning for which version of VS Code this was tested would also help in finding information that requires updating.

RudolfWeeber:
@jngrad, the information in the doc is not intended for developers but for end users writing and debugging simulation scripts.

I'll update the text to make that clear. I'll try to make  it more general, using VS Code as an example.

I think many users aren't aware that writing simulation scripts in a Python IDE is easily possible and helpful.

fweik:
@RudolfWeeber I also agree that this is useful end-user documentation, we could maybe consider adding hints for other common IDEs such as [spyder](https://www.spyder-ide.org/). I think computational scientists often don't use the best tools for their job ^^

biermanncarl:
Thanks for the suggestions. I incorporated them. I wondered if we should add a remark like "The steps necessary to get |es| running in a different IDE should be very similar to the ones shown for *VS Code*."

fweik:
@biermanncarl thank you, these can be merged from my side. @jngrad what do you think?

KaiSzuttor:
I think the crucial information here is to replace the python interpreter used by the IDE. Setting it to the pypresso path is probably sufficient for all IDEs. So similar to Carls suggestion, I'd add this information.

jngrad (commented):
once Kai's comment is addressed, this will be ready to merge from my side

fweik (approved):

jngrad (approved):

PR #3794, opened 2020-07-07
Script interface Backend Replacement

fweik:
## Description of changes
- Rename all the things
- Make `Variant` (shared) own contained objects
- Introduce separate packing format for sending objects via MPI/serializing
- Make manager and context concepts explicit
- Moved script interface unit tests into script interface
- Added more tests
- Removed unused `Utils::make_function`
- Remvoed unused constructors of `ScriptInterface::AutoParameter`


src/script_interface/Context.hpp
```diff
@@ -0,0 +1,121 @@
...
+class Context : public std::enable_shared_from_this<Context> {
+public:
+  /**
+   * @brief Call method on remote instances
+   *
+   * @param o Internal identified of the instance
+   * @param name Name of the method to call
+   * @param arguments Arguments to the call
+   */
+  virtual void nofity_call_method(const ObjectHandle *self,
```
jngrad: here and elsewhere: should this method read `notify_call_method()`?

src/script_interface/Context.hpp
```diff
@@ -0,0 +1,121 @@
...
+  virtual void notify_set_parameter(const ObjectHandle *self,
+                                    std::string const &name,
+                                    Variant const &value) = 0;
+
+  /**
+   * @brief Delete remote instances
+   *
+   * @param o Internal identified of the instance
+   */
+  virtual void nofity_delete_handle(const ObjectHandle *self) = 0;
```
jngrad: here and elsewhere: should this method read `notify_delete_handle()`?

fweik:
@jngrad I don't understand the sphinx error, can you help me out?

src/python/espressomd/script_interface.pyx
```diff
@@ -89,38 +101,32 @@ cdef class PScriptInterface:
...
-        return oid
+
+        self.sip = sip
 
     def call_method(self, method, **kwargs):
         """
         Call a method of the core class.
-
         Parameters
         ----------
```
jngrad: Subtle regression: this newline was required to create two paragraphs, one for the description and one for the "Parameters" title. Without it, the `Parameters` word is appended to the previous paragraph and the dashes trigger an error.

fweik: Thank you. The error message is misleading then, it complains about the indentation...

jngrad: Yeah that's not a very helpful error message. I thought it came from a Sphinx extension, but it's reproducible with vanilla Sphinx syntax. Looking through the Sphinx issue tracker, the same error message can be raised in different scenarios involving a missing newline character. An open-source project felt the need to document this error message directly [in their user manual](https://openmdao.org/releases/0.0.9/docs/quick-ref/sphinx.html?highlight=unexpected%20indentation#some-basics).

src/script_interface/Context.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+#include "ObjectHandle.hpp"
+#include "Variant.hpp"
+
+#include <boost/utility/string_ref.hpp>
+
+#include <memory>
+
+namespace ScriptInterface {
+class Context : public std::enable_shared_from_this<Context> {
```
KaiSzuttor, line 47: can you explain why ```std::enable_shared_from_this``` is needed here?

src/script_interface/Context.hpp
```diff
@@ -0,0 +1,90 @@
...
+public:
+  /**
+   * @brief Call method on remote instances
+   *
+   * @param self Internal identified of the instance
```
KaiSzuttor: typo

src/script_interface/auto_parameters/AutoParameter.hpp
```diff
@@ -158,35 +139,36 @@ struct AutoParameter {
    */
   template <typename F, typename G,
             /* Try to guess the type from the return type of the getter */
-            typename R = typename decltype(
-                Utils::make_function(std::declval<G>()))::result_type>
+            typename R = decltype(std::declval<G>()())>
```
KaiSzuttor: where is this used?

src/script_interface/auto_parameters/AutoParameter.hpp
```diff
@@ -158,35 +139,36 @@ struct AutoParameter {
...
 
   /**
    * @brief Read-only parameter with a user-provided getter.
    * @overload
    */
   template <typename G,
             /* Try to guess the type from the return type of the getter */
-            typename R = typename decltype(
-                Utils::make_function(std::declval<G>()))::result_type>
+            typename R = decltype(std::declval<G>()())>
```
KaiSzuttor: where is this used?

src/script_interface/Context.hpp
fweik: Do you know what it does?

KaiSzuttor: yes

fweik: Ok, it's needed here to that all the `ObjectHandle`s in one context share ownership of the same context. E.g. if you make a new handle in a context, it should share ownership of that context.

KaiSzuttor: Okay maybe you should add a docstring to the class explaining what the role of a ```Context``` is here

src/script_interface/ObjectManager.hpp
```diff
@@ -0,0 +1,97 @@
...
+
+#include "MpiCallbacks.hpp"
+
+#include <utils/Factory.hpp>
+
+#include <memory>
+#include <stdexcept>
+
+namespace ScriptInterface {
+class ObjectManager {
```
RudolfWeeber: Please document, what this class is for

src/script_interface/PackedVariant.hpp
```diff
@@ -0,0 +1,153 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+#ifndef SCRIPT_INTERFACE_PACKED_VARIANT_HPP
+#define SCRIPT_INTERFACE_PACKED_VARIANT_HPP
+
```
RudolfWeeber: Maybe add a file level docstring explaining the high level purpose of the functions.

RudolfWeeber (dismissed):
Didn't read everything, yet.

fweik:
I'd really appreciate it if I could get some feedback on the design.

RudolfWeeber:
I’m not totally clear about the role of ObjectManager yet.

Is its purpose to hide the concept of contexts from the client?

 

Why is there serialization code in the object manager. Is that just to save the creation policy along with the state of the object?

fweik:
When restoring from serialization, a context is needed to create the objects in, this is why it is on the contex. Now that you mention it, maybe it would be better as a free function, let me try.

The `ObjectManager` (which is in dire need of a better name) is to deal with the multiple contexts that exist, which is needed e.g. for serialization, because for every object, it needs to be stored in which kind of context it belongs.

fweik:
I've added  some docs and move the serialization code to `ObjectHandle`

src/script_interface/Context.hpp
jngrad:
```suggestion
   * @param self Internal identifier of the instance
```

src/script_interface/Context.hpp
```diff
@@ -0,0 +1,94 @@
...
+                                    std::string const &name,
+                                    Variant const &value) = 0;
+
+  /**
+   * @brief Get a new reference counted instance of a script interface by
+   * name.
+   *
+   * Objects created thru a Context get shared ownership of that context,
+   * e.g. the lifetime of the context is at least as long as the objects
+   * created by it. Therefor the object can always assume that the context
```
jngrad:
```suggestion
   * created by it. Therefore the object can always assume that the context
```

src/script_interface/packed_variant.hpp
```diff
@@ -0,0 +1,180 @@
...
+  return reinterpret_cast<ObjectId>(p);
+}
+
+/**
+ * @brief Variant value with references replaced by ids.
+ */
+using PackedVariant = boost::make_recursive_variant<
+    None, bool, int, double, std::string, std::vector<int>, std::vector<double>,
+    ObjectId, std::vector<boost::recursive_variant_>, Utils::Vector2d,
+    Utils::Vector3d, Utils::Vector4d>::type;
```
jngrad: I didn't get what you meant by replacing references by ids. This sounds like a description of `PackVisitor`. To me `PackedVariant` looks like a plain list of allowed types for a variant.

src/script_interface/Exception.hpp
```diff
@@ -16,13 +16,22 @@
...
+#include <stdexcept>
+#include <string>
 
-extern LangevinThermostat langevin;
-extern BrownianThermostat brownian;
-extern IsotropicNptThermostat npt_iso;
+namespace ScriptInterface {
+struct Exception : public std::exception {
+  explicit Exception(const char *msg) : message(msg) {}
+  explicit Exception(std::string msg) : message(std::move(msg)) {}
```
jngrad, line 28: What is the purpose of this custom exception class? It has a generic name, yet seems to be only used in `AutoParameters` for read-only and unknown parameters. Can it be safely used elsewhere in the script interface?

src/script_interface/GlobalContext.cpp
```diff
@@ -0,0 +1,104 @@
...
+namespace ScriptInterface {
+void GlobalContext::make_handle(ObjectId id, const std::string &name,
+                                const PackedMap &parameters) {
+  try {
+    ObjectRef so = m_node_local_context->make_shared(
+        name, unpack(parameters, m_local_objects));
+
+    m_local_objects.emplace(std::make_pair(id, std::move(so)));
+  } catch (Exception const &) {
+  }
```
jngrad, line 49: Here and below, why do we skip the `Exception` type?

src/script_interface/packed_variant.hpp
```diff
@@ -0,0 +1,180 @@
...
+
+/**
+ * @brief Id for object.
+ *
+ * This assigns every ObjectHandle a unique id.
+ */
+inline ObjectId object_id(const ObjectHandle *p) {
+  static_assert(sizeof(const ObjectHandle *) <= sizeof(ObjectId), "");
+  return reinterpret_cast<ObjectId>(p);
+}
```
jngrad: Using the object address as a unique id is safe in an MPI environment because we only serialize objects on the head node, right? And we don't have issues with polymorphism?

src/script_interface/packed_variant.hpp
fweik: Currently, `ScriptInterface::Variant` can only hold object ids, with this, it can hold shared pointers to objects. This is more covenient, because the `Variant` keeps the pointee alive and no special treatment is needed for them. The packed version is needed because for sending over mpi and for serialization we can't store pointers. In this case the pointers are replaced by ids and the client has to somehow deal with the shared ownership, and provide pointers again when converting back.

src/script_interface/Exception.hpp
fweik: This is so that we can catch all exceptions from the script interface but not others.

src/script_interface/GlobalContext.cpp
fweik: Because exceptions are handled on the head node. Only exceptions from the script interface are caught, because we know that they also occur on the head node.

src/script_interface/packed_variant.hpp
fweik: This code is only used on the head node as a hash function, so this should be fine. The polymorphism doesn't matter because the address as a `ObjectHandle` is unique within every object. It will not be at the start of the object for derived types, but this does not matter for this purpose.

src/script_interface/GlobalContext.cpp
fweik: If the mechanism with the context callbacks is not clear, it may be helpful to follow the call path for, say, `ObejctHandle::set_parameter`, to see what happens where. Basically the interface functions in `ObjecHandle`
call hooks in the context, and the context can then decide if it wants to react to the events. The `GlobalContext`
broadcasts the values to all nodes and then, e.g., sets the parameters on the local copies of the object, to keep
the copies on all nodes synchronous. The code you are citing here only runs on the remove copies, so the
exceptions that are catched are those of the remote local copies, which are redundant.

src/script_interface/packed_variant.hpp
jngrad: ok I see now what you meant (`PackedVariant` contains an `ObjectId` while `Variant` contais an `ObjectRef`), but it's not intuitive and very easy to overlook that difference. We need to write this down in some stand-alone doxygen file. For example, If we want to add support for a new type like `Vector3i` we need to touch both `PackedVariant` and `Variant`. Since `ObjectId` is a typedef to `size_t`, does this prevent us from adding `size_t` in the variants due to type collision?

fweik: What I really would like to have is some notion of native types (as opposed to object handles), which are the same in the packed and unpacked version. But because all the variant stuff is used all over the client code at the moment it is too cumbersome to change that. I also don't think that new types should be added to the list of native types, that is not
a scalable solution. I'm not a fan of technical writing in some doxygen  file, that will just get out of sync with the code,
at the end it can't be avoided to read the code anyway to understand how the library code works.

fweik: Also the `PackedVariant` is some internal technical detail, even if clients roll their own `ObjectHandle` implementation (which they shouldn't), they don't have to know about this.

jngrad: > I'm not a fan of technical writing in some doxygen file, that will just get out of sync with the code

Doxygen is designed to detect when such technical files get out of sync with the code. You can even move the docstrings out of the code and into stand-alone `.dox` files if you want to, as long as you provide the function/class/variable name, it will generate valid documentation. This concern you have is only the result of contributors writing stand-alone doxygen files *without* adding `@ref` in front of every function name, which means there is no way for them to detect desynchronized docs by reading the doxygen warnings and no way for CI to complain about it. I actually had to create a dictionary of function names for my espresso spellchecker because the parser rarely finds a `@ref` in front of non-english words.

> at the end it can't be avoided to read the code anyway to understand how the library code works

We don't want to be in a situation where we have to read the library code to understand how it works. It took me months to create a RDF observable because I had to understand the entire script interface. This was a tremendous waste of my time, because everything I learned will become useless once we merge this PR.

fweik: Well I just think that formal languages do a way better job in describing software behavior than English does.

fweik: > It took me months to create a RDF observable because I had to understand the entire script interface. 

Frankly I don't think this is true, others managed to add classes without reading all the code. I agree that the client side should be simplified thou, but I don't this will happen.

src/script_interface/ObjectManager.hpp
jngrad: docstrings were added in 5d0eeba99a42

jngrad (approved):
ok from my side

fweik:
Thank you for adding the docs ;-)

PR #3801, opened 2020-07-15
mda: extend MDAnalysis plugin by bond information.

KaiSzuttor:


KaiSzuttor:
there is an issue for bonds with more than two partners to be resolved

src/python/espressomd/MDA_ESP/__init__.py
```diff
@@ -164,6 +178,7 @@ def parse(self):
                  Tempfactors(np.zeros(natoms)),
                  ICodes(np.array([' '], dtype=object)),
                  Charges(np.array(charges)),
+                 Bonds(bonds)
```
fweik: Should the angles and dihedrals be here too?

jngrad: yes, it's already fixed in the branch, but GitHub is not updating it (see 03a07713ae)

fweik (approved):
Great that there is a test now.

jngrad (approved):

PR #3803, opened 2020-07-17
Python coverage

KaiSzuttor:
Maybe a starting point for getting python coverage.

### DIFF
--- a/src/python/espressomd/CMakeLists.txt
+++ b/src/python/espressomd/CMakeLists.txt
@@ -60,6 +60,9 @@ list(APPEND cython_SRC ${CMAKE_CURRENT_BINARY_DIR}/code_info.pyx)
 list(REMOVE_DUPLICATES cython_SRC)
 
 add_library(pyx_interface INTERFACE)
+if(WITH_COVERAGE)
+  target_compile_options(pyx_interface INTERFACE -DCYTHON_TRACE)
+endif(WITH_COVERAGE)
 if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
   target_compile_options(
     pyx_interface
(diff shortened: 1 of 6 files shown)


KaiSzuttor:
depends on https://github.com/espressomd/docker/pull/185

jngrad:
That looks like `/tikhome/jgrad/Documents/COVERAGE-KEEP/patch.diff`, where the `-DCYTHON_TRACE` flag adds coverage support for Cython files (requires `# cython: linetrace=True` at the beginning of every Cython file, see `/tikhome/jgrad/Documents/COVERAGE-KEEP/instructions-Python.sh`).

The last time I used it in parallel mode, random line numbers appeared in the coverage report that exceeded the number of lines in the respective python scripts, forcing me to run the Python tests on a single MPI rank. I also had to post-process the coverage reports to remove false positives in the Cython coverage.

It has been a year or so, maybe the python package fixed these issue since then. I'll try it again.

src/python/espressomd/CMakeLists.txt
```diff
@@ -101,9 +104,9 @@ foreach(cython_file ${cython_SRC})
       OUTPUT ${outputpath}
       COMMAND
         ${CYTHON_EXECUTABLE} $<$<BOOL:${WARNINGS_ARE_ERRORS}>:--warning-errors>
-        -3 --cplus --directive embedsignature=True --directive binding=True -I
-        ${CMAKE_CURRENT_SOURCE_DIR} -I ${CMAKE_CURRENT_BINARY_DIR}
-        ${cython_file} -o ${outputpath}
+        -3 --cplus --directive embedsignature=True --directive binding=True
+        --directive linetrace=True -I ${CMAKE_CURRENT_SOURCE_DIR} -I
```
jngrad, line 108: Do we want `--directive linetrace=True` in all builds? This add a substantial amount of code that is not relevant to non-coverage builds.

KaiSzuttor:
memo: ```coverage``` executable is named ```python3-coverage``` if installed with the package manager in ubuntu

fweik:
Potentially you can use `$PYTHON_EXECUTABLE -m converage` and this will work whatever, but I have not checked...

src/python/espressomd/CMakeLists.txt
KaiSzuttor: no, it's just not "if"-ed yet

jngrad:
On my workstation this PR generates an empty coverage file. Here's my version: c66c16319. There are a few issues with it, e.g. the .pyx file paths sometimes contain `/testsuite/` or `/build/` by mistake. Also for codecov we probably need to convert the absolute paths into relative paths.

jngrad:
Currently this python package can only report hit lines, not missed lines, so the coverage is always 100%. There's a way to get line misses by parsing the generated .cpp files (see the result in `/work/jgrad/espresso-fork-tag401/build-cov/coverage_checkpoint_html4/index.html`), but this feature requires a little bit of Python code and relies on Cython output whose format is probably not part of the API. There is a PR open since 2017 to get the Cython AST parser in the API, but it's still not on the agenda. Getting line misses for Python code requires running the AST parser, too.

The coverage reports for Cython source code also mixes real lines with `tree fragment` code blocks that do not belong to the .pyx file, which causes the coverage to report random line hits in license headers and the following few lines of actual code.

KaiSzuttor:
there are testcases for the coverage feature in the cython repository, so we should be able to get it working

KaiSzuttor:
giving up here for now

PR #3809, opened 2020-07-22
maintainer: coverage for integration tests.

KaiSzuttor:


jngrad:
LGTM, although the coverage is not 100% trustworthy. For example, the `save_checkpoint.py` test has no coverage for LB, even though the corresponding code is actually run on my workstation. Do we really need the 6000 lines of gcov output in the CI log?

.codecov.yml
```diff
@@ -50,3 +48,6 @@ comment:
 
 ignore:
   - "libs"
+fixes:
+  - "testsuite/python/save_checkpoint*::testsuite/python/save_checkpoint.py"
+  - "testsuite/python/test_checkpoint*::testsuite/python/test_checkpoint.py"
```
jngrad, line 50: Could it be that, for Python coverage (in contrast to C++ coverage), this operation actually causes the last `save_checkpoint*` file (in alphabetic or timestamp order) to be used for the coverage report?  I'm trying to figure out why some parts of the checkpoint tests appear with no coverage.

jngrad: The `.coverage` file from the cuda9 artifact contains line coverage for the part that are red in codecov.

KaiSzuttor: yeah it seems like it overwrites previous reports

jngrad (dismissed):

.codecov.yml
jngrad: offline discussion: let's discuss plans for a more maintainable checkpointing test framework on Monday meeting, that would solve this issue as a side-effect

KaiSzuttor:
@jngrad could you please approve again, I just disabled the not so useful codecov comments

jngrad (approved):

PR #3817, opened 2020-07-27
test: electrophoresis sampel increase tolerance

### LINKED ISSUE #3816: CI build failed for merged PR
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/pipelines/12630

RudolfWeeber:
Fixes #3816 

Description of changes:
Increase tolerance for polymer persistence length to mitigate flaky test.

### DIFF
--- a/testsuite/scripts/samples/test_electrophoresis.py
+++ b/testsuite/scripts/samples/test_electrophoresis.py
@@ -30,8 +30,8 @@ class Sample(ut.TestCase):
     def test_persistence_length(self):
         # These two values differ due to undersampling, they converge
         # to the same value around N_SAMPLES=1000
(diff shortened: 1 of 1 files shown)


KaiSzuttor:
shouldn't we make the test deterministic instead of increasing the tolerance?

RudolfWeeber:
Not trivial. Coulomb tuning is involved. The outcome of that is somewhat random.

One would have to figure out how to inject tuning parameters without bloating the sample script so much that it gets confusing to new users.

 

I don’t have time to take that up right now.

KaiSzuttor:
okay, but at which delta do we stop?

RudolfWeeber:
>okay, but at which delta do we stop?

Cannot be answered in the abstract.

KaiSzuttor:
fixed in #3821

PR #3827, opened 2020-07-27
maintainer: perform mac jobs on github.

KaiSzuttor:
Use github actions to run tests on a mac os platform. We probably can rely on prebuilt binaries for the dependencies using homebrew:
>If available, bottled binaries will be used by default except under the following conditions:
  Options were passed to the install command, i.e. brew install <formula> will use a bottled version of the formula, but brew 
    install --enable-bar <formula> will trigger a source build.
  The --build-from-source option is invoked.
  The machine is not running a supported version of macOS as all bottled builds are generated only for supported macOS 
  versions.
  Homebrew is installed to a prefix other than the standard /usr/local (although some bottles support this).


.github/workflows/build.yml
```diff
@@ -0,0 +1,18 @@
...
+
+jobs:
+  tests:
+    runs-on: macos-latest
+    steps:
+    - uses: actions/checkout@master
+    - run: |
+       brew install boost boost-mpi fftw
+       brew install hdf5-mpi
+       pip3 install numpy cython
```
jngrad: here we would need `vtk`, `scipy`, `h5py`

jngrad (approved):

PR #3830, opened 2020-07-28
Fix CMake warning, update macOS installation instructions

jngrad:
Description of changes:
- fix `GNUInstallDirs`-related warning generated in CMake 3.18
- update macOS installation instructions for command line tools
- increase timeout on macOS runner from GitHub Workflows

### DIFF
--- a/doc/sphinx/installation.rst
+++ b/doc/sphinx/installation.rst
@@ -184,16 +184,26 @@ using the following commands:
     sudo rm -r ~/anaconda[23]
     sudo rm -r /Library/Python
 
-If you want to install MacPorts, download the installer package
-appropriate for your Mac OS X version from
-https://www.macports.org/install.php and install it.
-
-If you want to install Homebrew, use the following commands.
+Install the command line tools (gcc, git, make, etc.) with either:
 
 .. code-block:: bash
 
     sudo xcode-select --install
+
+or:
+
+.. code-block:: bash
+
     sudo xcodebuild -license accept
+
+If you want to install MacPorts, download the installer package
+appropriate for your Mac OS X version from
+https://www.macports.org/install.php and install it.
+
+If you want to install Homebrew, use the following command:
+
+.. code-block:: bash
+
     /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
 
 Installing packages using MacPorts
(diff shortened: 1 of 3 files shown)


doc/sphinx/installation.rst
```diff
@@ -184,16 +184,26 @@ using the following commands:
     sudo rm -r ~/anaconda[23]
     sudo rm -r /Library/Python
 
+Install the command line tools (gcc, git, make, etc.) with either:
```
KaiSzuttor: I think we do not need to document this because it's already contained in the macports documentation https://www.macports.org/install.php

mkuron: Actually we do because we also need them for the Homebrew instructions. Also, we only tell people to download MacPorts, not to follow the install guide there.

Note that Xcode does not include gcc, so this line should mention either `clang` or the C++ compiler.

doc/sphinx/installation.rst
```diff
@@ -184,16 +184,26 @@ using the following commands:
...
+
+.. code-block:: bash
+
+    sudo xcode-select --install
+
+or:
+
+.. code-block:: bash
+
+    sudo xcodebuild -license accept
```
mkuron: `sudo xcodebuild -license accept` does not install anything. If Xcode has been downloaded from the Mac App Store but never started, it accepts its license. Please change to something like "If you have already installed Xcode from the Mac App Store, run `sudo xcodebuild -license accept`. Else install the command line tools using `sudo xcode-select --install`."

mkuron (changes requested):

KaiSzuttor (approved):

doc/sphinx/installation.rst
jngrad: In 15a0799 I've replaced everything with links to the official instructions for MacPorts and Homebrew.

KaiSzuttor: you commented on an outdated diff

mkuron:
Do you really think it's a good idea to leave people on their own in installing Homebrew or MacPorts? Not everyone knows how to do that and their own install instructions can be intimidating to novice users.

jngrad:
@mkuron I was not able to follow the MacPorts installation instructions in our user guide, the terminal threw an error. Even after you explained to me how the second command was only relevant to Xcode installed from the App Store, I still misunderstood what the command really did. The official MacPorts installation instructions explain why `xcodebuild -license` is only relevant to Xcode 4+. I agree it's quite verbose, but linking to it guarantees our macOS instructions are always up-to-date.

PR #3838, opened 2020-07-30
SD interface cleanup

fweik:
Description of changes:
- Removed unused data copy and communication from SD particle buffer
- Removed dependency on globals
- Reduced code duplication

@RudolfWeeber it is somewhat surprising to me that you signed off on this...


jngrad (approved):
LGTM

PR #3841, opened 2020-07-30
P3M generic influence function

fweik:
Fixes #

Description of changes:
- Implement P3M influence function that works for all powers of the differential operator
- Use that for P3M force and energy influence function calculation
- *Should* also work for dp3m... *should*


fweik:
@jngrad is it true that it is now possible to reference an other member variable in a docstring of a member function? This seems like a surprising limitation of doxygen, I'm sure I'm doing it wrong somehow.

jngrad:
> @jngrad is it true that it is now possible to reference an other member variable in a docstring of a member function?

True, but you're trying to reference private methods, which are currently not processed by doxygen:
https://github.com/espressomd/espresso/blob/22c789a31bae96e66a08eee9b071e710e6f2b125/doc/doxygen/Doxyfile.in#L438-L442

The default value of `NO` is helpful for C++-only libraries, where private members do not need to be shown to the public because they are not part of the API. But we have Python as the interface, where we already hide private members with a leading underscore. I wouldn't mind turning this doxygen field to `YES` given that our doxygen docs are primarily intended to developers.

fweik:
I see. I think in this special case I'll probably go for a different design anyway, but thanks.

fweik:
Sorry for the noise, should be ready for review now.

jngrad (dismissed):
LGTM

jngrad (approved):

PR #3831, opened 2020-07-28
Test: add test for lb pressure tensor acf against viscosity

RudolfWeeber:
Fixes #

Description of changes:
-

PR #3847, opened 2020-08-03
Fix regression in CPU LB thermalization.

### LINKED ISSUE #3772: Thermalized CPU LB produces unphysical currents
When using the thermalized CPU LB with boundaries that create corners, unphysical currents occur. See below an example. With the CPU implementation, there are currents in y and z direction without any external forces or velocity boundary conditions. For the GPU implementation, this does not happen.

```python
from espressomd import System, lb, lbboundaries, shapes
import numpy as np

impl = 'CPU' #'GPU'

box_l = [40,15,40]
LB_params = {'agrid':1,
             'tau':1,
             'dens':1,
             'visc':0.2,
             'kT':1e-8,
             'seed':41}
padding = 4
system = System(box_l = box_l)
system.cell_system.skin = 1
system.time_step = LB_params['tau']

if impl == 'GPU':
    lbf = lb.LBFluidGPU(**LB_params)
elif impl == 'CPU':
    lbf = lb.LBFluid(**LB_params)
    
system.actors.add(lbf)

top_wall = shapes.Wall(dist = -(box_l[2]-padding), normal = [0,0,-1])
bottom_wall = shapes.Wall(dist = padding, normal = [0,0,1])
left_wall = shapes.Wall(dist = padding, normal = [1,0,0])
right_wall = shapes.Wall(dist = -(box_l[0]-padding), normal = [-1,0,0])
for wall in [top_wall,bottom_wall, left_wall,right_wall]:
    noslip_wall = lbboundaries.LBBoundary(shape = wall)
    system.lbboundaries.add(noslip_wall)

system.integrator.run(100)

#lbf.print_vtk_velocity(f'./velocity_{impl}.vtk')  

probe_xs = np.linspace(0,box_l[0],num=40)
vels_over_time = []
n_samples = 100
for _ in range(n_samples):
    probe_vels = np.array([lbf.get_interpolated_velocity([x,0,box_l[2]/2.]) for x in probe_xs])
    vels_over_time.append(np.mean(probe_vels,axis = 0))
    system.integrator.run(10)
vels_over_time = np.array(vels_over_time)
mean_vel = np.mean(vels_over_time, axis = 0)
std_vel = np.std(vels_over_time, axis=0)
SEM_vel = std_vel/np.sqrt(n_samples)

print(f'velocity = {mean_vel} +- {SEM_vel}')
print(np.all(np.abs(mean_vel)<std_vel))
#I know we should compare to the standard error of the mean (divide by sqrt(n_samples)) but that takes forever to produce True on GPU. 
                          
#fig = plt.figure()
#for direction in [0,1,2]:
    #plt.plot(probe_xs, probe_vels[:,direction], label = f'v_{direction}')
#plt.xlabel('x')
#plt.legend()
#plt.show()
```

### LINKED ISSUE #3804: Thermalization of CPU LB broken
IMHO broken for CPU. Minimal example:
```python
import argparse

parser.add_argument("--use_gpu", default=False, action="store_true")
parser.add_argument("--kt", type=float, default=0)
args = parser.parse_args()

import espressomd
from espressomd.observables import LBFluidPressureTensor

system = espressomd.System(box_l=[20, 20, 20])
system.time_step = 0.01
system.cell_system.skin = 0.4
if args.use_gpu:
    lbf = espressomd.lb.LBFluidGPU(agrid=1.0, dens=1.9, visc=1.0, tau=0.01, kT=args.kt, seed=7)
else:
    lbf = espressomd.lb.LBFluid(agrid=1.0, dens=1.9, visc=1.0, tau=0.01, kT=args.kt, seed=7)
system.actors.add(lbf)
lb_pressure = LBFluidPressureTensor()
for i in range(2):  # change as you like
    system.integrator.run(1000)  # same here, problem persists
    print(lb_pressure.calculate())
```
Testing via `run_minimal.sh`:
```bash
#!/bin/bash

echo "CPU, kt=0:"
./pypresso minimal.py
echo "GPU, kt=0:"
./pypresso minimal.py --use_gpu
echo "CPU, kt=1:"
./pypresso minimal.py --kt 1
echo "GPU, kt=1:"
./pypresso minimal.py --use_gpu --kt 1
```
yields:
```
$ ./run_minimal.sh 
CPU, kt=0:
[[6333.33325386    0.            0.        ]
 [   0.         6333.33325386    0.        ]
 [   0.            0.         6333.33325386]]
[[6333.33325386    0.            0.        ]
 [   0.         6333.33325386    0.        ]
 [   0.            0.         6333.33325386]]
GPU, kt=0:
WARNING: More than one GPU detected, please note ESPResSo uses device 0 by default regardless of usage or capability. The GPU to be used can be modified by setting System.cuda_init_handle.device.
[[6333.33353698    0.            0.        ]
 [   0.         6333.33353698    0.        ]
 [   0.            0.         6333.33353698]]
[[6333.33353698    0.            0.        ]
 [   0.         6333.33353698    0.        ]
 [   0.            0.         6333.33353698]]
CPU, kt=1:
[[6.39378045e+03 1.90899139e+00 1.95662446e+00]
 [1.90899139e+00 6.38988958e+03 1.94024253e+00]
 [1.95662446e+00 1.94024253e+00 6.38844944e+03]]
[[6.39384017e+03 1.91773972e+00 2.03000427e+00]
 [1.91773972e+00 6.39001967e+03 1.98502662e+00]
 [2.03000427e+00 1.98502662e+00 6.38863714e+03]]
GPU, kt=1:
WARNING: More than one GPU detected, please note ESPResSo uses device 0 by default regardless of usage or capability. The GPU to be used can be modified by setting System.cuda_init_handle.device.
[[ 6.33442971e+03 -2.38058610e-02 -6.62311537e-02]
 [-2.38058610e-02  6.33435239e+03  4.31334905e-02]
 [-6.62311537e-02  4.31334905e-02  6.33439990e+03]]
[[ 6.33472237e+03 -5.51996531e-02 -2.82031257e-02]
 [-5.51996531e-02  6.33468426e+03  1.38701318e-02]
 [-2.82031257e-02  1.38701318e-02  6.33475568e+03]]
```
`kT=0` case is fine, but with thermalization, CPU produces off-dagonal pressure values of order $10^0$, instead of GPU, order $10^{-2}$ which seems more reasonable, I think.

pkreissl:
Fixes #3804 , fixes #3772

The issue was actually a regression that happened with the switch to philox in commit 
[f3cc4ba](https://github.com/espressomd/espresso/pull/2392/commits/f3cc4ba918b80ba86b7558bd76c3f9b16d961e38). Random numbers formerly part of the interval (-0.5,0.5] were replaced by random numers in (0,1].
With this fix the `lb_pressure_tensor_acf.py` runs successfully for both CPU and GPU. However, as @KaiSzuttor correctly mentioned in PR #3831 the CPU part of the test takes a while to execute (on my machine, single core the whole test takes 136 s). I could try to make that faster which, however, would require tweaking the tolerance limits `tol_node` and `tol_global`. @RudolfWeeber , what was your reasoning behind the chosen limits? Or are they semi-arbitrary choices?

Further, this PR corrects the comparison of the off-diagonal elements `avg_ij` vs. `avg_ji` in the test.


jngrad:
This looks like a candidate for a 4.1.4 release...

mkuron:
What are the physical consequences of this error? Are all our thermalized fluid simulations of the past 1.5 years wrong?

KaiSzuttor:
the CPU version has probably not been used by many

KaiSzuttor:
looking at this function the issue is the non trivial RNG code. There is a level of abstraction missing.

RudolfWeeber:
@pkreissl, can you please check whether this also solves #3772.
I picked the tolerance limits empirically. Please keep the limits for the gpu tests and run the cpu test shorter with wider margins.
The test may be faster for the same accuracy for a bigger fluid at less steps (due to ghost comm).

This definitely warrants a 4.1.4 release.

KaiSzuttor:
> @pkreissl, can you please check whether this also solves #3772.
> I picked the tolerance limits empirically. Please keep the limits for the gpu tests and run the cpu test shorter with wider margins.
> The test may be faster for the same accuracy for a bigger fluid at less steps (due to ghost comm).
> 
> This definitely warrants a 4.1.4 release.

while I see that it might be tempting to think that a test with empirical or arbitrary tolerance is better than no test, we should invest some time in investigating if the tolerance limit can be deduced from some known facts abou the algorithm. A "false green" can harm when it comes to test driven development. Also with the upcoming change to Walberla we should be confident about the tolerance in our testsuite.

RudolfWeeber:
For <sigma_ij> =0, that should be possible.

Since <sigma_ij>^2 can probably be found in Ulf’s thesis and one can assume that a Gaussian distribution is present.

 

 

For the ACF, the situation is too complicated, as a numerical integration and a least square fit is involved.

The semi-empirical tolerance was 10% deviation from prescribed viscosity. Based on the work  that Patrick and I did on evaluating viscosity from stress acf, that’s probably as good as it gets.

 

 

More generally, we might have to introduce some long-running tests that run in a separate CI task.

Running some of them 10x as long as we do now is the only way to be absolutely sure that they converge to the correct value.

 

The need for that became apparent with the Walberla branch.

E.g., slight errors in ghost layer handling caused momentum changes that only became apparent in long simulations.

The error was of a nature I hadn’t considered when writing unit tests.

KaiSzuttor:
> Since <sigma_ij>^2 can probably be found in Ulf’s thesis 

I thought @pkreissl already checked and was not sure if he found it or not

pkreissl:
> > Since <sigma_ij>^2 can probably be found in Ulf’s thesis
> 
> I thought @pkreissl already checked and was not sure if he found it or not

The respective correlations for D3Q19 are given in eq. (4.51) of Ulfs dissertation -- however, they depend on the eigenvalues. I guess, those can be obtained via eqs. (4.29)/(4.30). But as I already mentioned, I have not yet had in-depth contact with the document (and its notation), so I am not yet overly confident, here. Comments from people with higher LB expertise are welcome (@mkuron , @KaiSzuttor ?).

KaiSzuttor:
if I'd know what we are looking for I'd let you know :smiley:

pkreissl:
> if I'd know what we are looking for I'd let you know smiley

The expected variance of the stress tensor element combinations in terms of LB input parameters... :)

mkuron:
My knowledge about thermalization is contained in https://i10git.cs.fau.de/pycodegen/lbmpy/-/blob/master/lbmpy/fluctuatinglb.py, which was directly translated form Ulf's thesis.

KaiSzuttor:
> My knowledge about thermalization is contained in https://i10git.cs.fau.de/pycodegen/lbmpy/-/blob/master/lbmpy/fluctuatinglb.py, which was directly translated form Ulf's thesis.

where are the unit tests. That's the interesting part...

KaiSzuttor:
as far as I can see the unit tests in walberla also have "empirical" accuracy limits. So this does not help. @mkuron could you please have a look at the highly cited thesis and check if we can extract values for the accuracy limits?

KaiSzuttor:
Correct me if I'm wrong:
according to [Dünweg et al.](https://journals.aps.org/pre/pdf/10.1103/PhysRevE.76.036704?casa_token=Ig9myMlFpq8AAAAA%3AzLXWfwZAKtvgT3CfdzURW0PoO5qNLEW5BCaSKVY_iwgMpGVDenQoUrnPDG1WrfcJTOfssBREUsWmK9c) Eq. 62 the local stress should fluctuate with the following covariance matrix:

- for diagonal elements: 
   image 

- for off-diagonal elements:
   image

jngrad:
@pkreissl if there is really no way to write a faster/non-statistical test, 722bc52c6f shows how to disable it from coverage

RudolfWeeber:
In LbmPy, stresss tensor fluctuagions can be matched against the expressions 
immediately above the cited one:

```

    res_pressure = dh.gather_array("pressure").reshape((-1,3,3))
    
    c_s = np.sqrt(1/3) # speed of sound
    
    # average of pressure tensor
    p_av_expected =np.diag([rho_0*c_s**2]*3) 
    np.testing.assert_allclose(np.mean(res_pressure,axis=0), p_av_expected, atol=c_s**2/500)
    
    # Pressure tensor fluctuations
    agrid = 1
    dt = 1
    dim = 3

    mu = temperature *dt**2 /(c_s**2 * agrid**(dim+2))

    # relaxation rates are dfined differently in Walberla
    gamma_shear = 1- omega_shear
    gamma_bulk = 1 - omega_bulk
    
    p_var_diag_expected = 6 *mu *rho_0 *c_s**4 * (1-gamma_bulk**2)
    p_var_off_diag_expected = (1-gamma_shear**2) * mu * rho_0 * c_s**4

    p_var_expected = np.zeros((3,3))
    for k in range(3): 
        for l in range(3):
            if k==l: p_var_expected[k,l] = p_var_diag_expected
            else: p_var_expected[k,l] = p_var_off_diag_expected
    np.testing.assert_allclose(np.var(res_pressure,axis=0),p_var_expected,rtol=0.03)
```
The same should be possible in Espresso. There, the relaxation rates match
the definition in the paper, afaik.

KaiSzuttor:
```p_var_diag_expected = 6 *mu *rho_0 *c_s**4 * (1-gamma_bulk**2)``` this differs from the variance I read and cited above. Can you please check?

**Comment by pkreissl** [comment]
> `p_var_diag_expected = 6 *mu *rho_0 *c_s**4 * (1-gamma_bulk**2)` this differs from the variance I read and cited above. Can you please check?

with p_var_diag_expected == <R_{alpha,alpha},R_{beta,beta}>, fluctuations in the trace, this is actually exactly eq. (61) in the *Phys. Rev. E* you cited...

**Comment by KaiSzuttor** [comment]
ah I was not able to parse the tensor of rank 4 in my head and missed Eq. 61. You are probably right.

**Comment by KaiSzuttor** [comment]
so we should at least add this test to the walberla PR

**Comment by pkreissl** [comment]
Looking into that right now...

**Comment by RudolfWeeber** [comment]
Turns out that the test in lbmpy only works for the relaxation rates I chose randomly. Lottery win...

In fact, the stress tensor fluctuations <sigma_{ij}^2> do not depend on the relaxation rates at all.
Re-reading in the paper, I get the impression that the equations refer to the fluctuations ADDED to the stress modes in the thermalization step, not to the RESULTING stress from combining thermalization and relaxation.

In fact, the expression in the paper is used in LbmPy when adding the fluctuations in fluctuation_amplitude_from_temperature()
https://i10git.cs.fau.de/pycodegen/lbmpy/-/blob/master/lbmpy/fluctuatinglb.py

**Comment by mkuron** [comment]
> the equations refer to the fluctuations ADDED to the stress modes in the thermalization step

Correct, these are the equations for $\sigma^r$. He also gives some equations without superscript (3.13, 4.48, 4.51) that appear to relate to the superscripted ones via a relaxation rate-dependent factor.

**Comment by RudolfWeeber** [comment]
* Please keep the average stress test for the CPU. That can run with much fewer steps.
You'll probably have to move the sampling to a separate function which gets called upon test setup. Then you can split the average test and the acf test into separate funcitons. The acf test then goes into the GPU class.

In MD units, `c_s**2 * factor` is not a suitable tolerance for the off-diagonal elements of the pressure tensor, since c_s**2 ~=20k in MD units.

**Comment by pkreissl** [comment]
With re-enabled acf test for GPU the test takes ~27 s on the local machine. I excluded the test from the builds with coverage, still `clang-sanitizer` CI-job times out after 900 s. `osx` also fails, but with the `domain_decomposition` test which should not be influenced by this PR.

**Comment by jngrad** [comment]
The `osx` job will disappear once kodiak merges this PR into the python branch. It doesn't seem to have failed in the latest commit. Regarding the sanitizer job, you could selectively disable the test with jngrad/espresso@1c16301ab508c2, but this is not a clean solution. The proposal for a long-term solution for these statistical tests (#3883) hasn't received any feedback from the core team yet.

File: testsuite/python/lb_pressure_tensor.py [hunk 1]
```diff
@@ -32,7 +32,7 @@
 KT = 0.8
 
 
-class TestLBPressureACF:
+class TestLBPressureTensor:
```
KaiSzuttor, line 35: i think this testcase needs a lot more documentation

pkreissl:
So, I've added references to the relevant equations in scientific publications.
For the equilibrium stess, we have eq. (19) in [Ladd et. al, 2001](https://doi.org/10.1023/A:1010414013942):

  \mathbf{\Pi}_\mathrm{eq} = \rho c_\mathrm{s}^2\mathbf{1} + \rho \mathbf{u}\mathbf{u}

For me, that would mean:

\mathbf{\Pi}_\mathrm{eq} = \rho c_\mathrm{s}^2\mathbf{1} + (2)/(V) (1)/(2) m u^2 = \rho c_\mathrm{s}^2\mathbf{1} + 2 k_\mathrm{B}T / V

In the test, however, the stress matches the above expression **without** the factor 2 in the \mathbf{u}\mathbf{u}-term (with factor it would be off)!
Am I missing something, here, or do we have another bug in the core?

testsuite/python/lb_pressure_tensor.py
pkreissl: After adding the references to equations, which parts of the test are not clear to you? I think, as is, there are sufficient comments that do in combination with naming of fuctions and variables explain what happens. However, if there are specific places that are not clear yet, I am happy to add more comments.

pkreissl:
Ok from an ideal gas, we know that on average the kinetic energy is

(1)/(2) m u^2 = (3)/(2) k_\mathrm{B} T,

and by equipartition that every dof has (1)/(2) k_\mathrm{B} T, which resolves the "issue" above.

jngrad:
@pkreissl should it be documented in the LB code? Doxygen supports LaTeX, see e.g. [bonded_interactions.dox](https://github.com/espressomd/espresso/blob/f760050199b94226b9127b309bfa7db569507605/src/core/bonded_interactions/bonded_interactions.dox#L254-L285) for the syntax.

pkreissl:
> @pkreissl should it be documented in the LB code? Doxygen supports LaTeX, see e.g. [bonded_interactions.dox](https://github.com/espressomd/espresso/blob/f760050199b94226b9127b309bfa7db569507605/src/core/bonded_interactions/bonded_interactions.dox#L254-L285) for the syntax.

You mean the expected stress tensor value? Can do that, but I don't think that this directly occurs in the core. There just the thermalization fluctuations are added to the modes and the tensor evaluated. If you still find that useful, sure, I can add it as a comment to `lb_calc_pressure_tensor()`.

jngrad:
> You mean the expected stress tensor value? Can do that, but I don't think that this directly occurs in the core.

nevermind, I mixed up the python test we looked into this morning and the source code :) maybe it would be helpful to write a short comment about the kinetic energy formula and equipartition, so that future developers know why there is no factor 2 in there

RudolfWeeber (approved):

PR #3856, opened 2020-08-09
Modernize code

jngrad:
Fix multiple issues detected by lgtm tooling.

Description of changes:
- python: catch specialized exceptions instead of `BaseException` (lgtm [rule 6780080](https://lgtm.com/rules/6780080/))
- core: add include guards and modernize variable declarations (C++ Core Guidelines [SF.8](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-guards), [ES.12](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-reuse), [ES.22](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-init), [ES.25](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-const), [ES.49](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-casts-named))

### DIFF
--- a/samples/save_checkpoint.py
+++ b/samples/save_checkpoint.py
@@ -41,7 +41,7 @@ if not checkpoint.checkpoint_signals:
     checkpoint.register_signal(signal.SIGINT)
 
 # test for user data
-myvar = "some script variable"  # lgtm[py/multiple-definition]
+myvar = "some script variable"
 checkpoint.register("myvar")
 myvar = "updated value"  # demo of how the register function works
 
--- a/samples/save_checkpoint.py
+++ b/samples/save_checkpoint.py
@@ -41,7 +41,7 @@ if not checkpoint.checkpoint_signals:
     checkpoint.register_signal(signal.SIGINT)
 
 # test for user data
-myvar = "some script variable"
+myvar = "some script variable"  # lgtm[py/multiple-definition]
 checkpoint.register("myvar")
 myvar = "updated value"  # demo of how the register function works
 
--- a/src/core/electrostatics_magnetostatics/elc.cpp
+++ b/src/core/electrostatics_magnetostatics/elc.cpp
@@ -1460,7 +1460,7 @@ void ELC_P3M_restore_p3m_sums(ParticleRange const &particles) {
 
   MPI_Allreduce(node_sums, tot_sums, 3, MPI_DOUBLE, MPI_SUM, comm_cart);
 
-  p3m.sum_qpart = static_cast<int>(std::round(tot_sums[0]));
+  p3m.sum_qpart = (int)(tot_sums[0] + 0.1);
   p3m.sum_q2 = tot_sums[1];
   p3m.square_sum_q = Utils::sqr(tot_sums[2]);
 }
--- a/src/core/electrostatics_magnetostatics/elc.cpp
+++ b/src/core/electrostatics_magnetostatics/elc.cpp
@@ -217,15 +217,14 @@ void distribute(int size) {
  *  See @cite yeh99a.
  */
 static void add_dipole_force(const ParticleRange &particles) {
-  double pref = coulomb.prefactor * 4 * M_PI * ux * uy * uz;
-  int size = 3;
+  double const pref = coulomb.prefactor * 4 * M_PI * ux * uy * uz;
+  int const size = 3;
 
   auto local_particles = particles;
 
   /* for nonneutral systems, this shift gives the background contribution
      (rsp. for this shift, the DM of the background is zero) */
-  double shift = 0.5 * box_geo.length()[2];
-  double field_tot = 0;
+  double const shift = 0.5 * box_geo.length()[2];
 
   // collect moments
 
@@ -258,7 +257,7 @@ static void add_dipole_force(const ParticleRange &particles) {
   distribute(size);
 
   // Yeh + Berkowitz dipole term @cite yeh99a
-  field_tot = gblcblk[0];
+  double field_tot = gblcblk[0];
 
   // Const. potential contribution
   if (elc_params.const_pot) {
@@ -281,11 +280,11 @@ static void add_dipole_force(const ParticleRange &particles) {
  *  See @cite yeh99a.
  */
 static double dipole_energy(const ParticleRange &particles) {
-  double pref = coulomb.prefactor * 2 * M_PI * ux * uy * uz;
-  int size = 7;
+  double const pref = coulomb.prefactor * 2 * M_PI * ux * uy * uz;
+  int const size = 7;
   /* for nonneutral systems, this shift gives the background contribution
      (rsp. for this shift, the DM of the background is zero) */
-  double shift = 0.5 * box_geo.length()[2];
+  double const shift = 0.5 * box_geo.length()[2];
 
   // collect moments
 
@@ -354,18 +353,18 @@ static double dipole_energy(const ParticleRange &particles) {
 /*****************************************************************/
 
 inline double image_sum_b(double q, double z) {
-  double shift = 0.5 * box_geo.length()[2];
-  double fac = elc_params.delta_mid_top * elc_params.delta_mid_bot;
-  double image_sum =
+  double const shift = 0.5 * box_geo.length()[2];
+  double const fac = elc_params.delta_mid_top * elc_params.delta_mid_bot;
+  double const image_sum =
       (q / (1.0 - fac) * (z - 2.0 * fac * box_geo.length()[2] / (1.0 - fac))) -
       q * shift / (1 - fac);
   return image_sum;
 }
 
 inline double image_sum_t(double q, double z) {
-  double shift = 0.5 * box_geo.length()[2];
-  double fac = elc_params.delta_mid_top * elc_params.delta_mid_bot;
-  double image_sum =
+  double const shift = 0.5 * box_geo.length()[2];
+  double const fac = elc_params.delta_mid_top * elc_params.delta_mid_bot;
+  double const image_sum =
       (q / (1.0 - fac) * (z + 2.0 * fac * box_geo.length()[2] / (1.0 - fac))) -
       q * shift / (1 - fac);
   return image_sum;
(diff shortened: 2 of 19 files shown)


samples/save_checkpoint.py
```diff
@@ -41,7 +41,7 @@
     checkpoint.register_signal(signal.SIGINT)
 
 # test for user data
-myvar = "some script variable"
+myvar = "some script variable"  # lgtm[py/multiple-definition]
```
KaiSzuttor: I think we should not include any comments for tooling in samples. Especially LGTM is currently more an experimental tool, where it's not sure whether we will use it in the future.

src/core/electrostatics_magnetostatics/elc.cpp
```diff
@@ -1464,7 +1460,7 @@ void ELC_P3M_restore_p3m_sums(ParticleRange const &particles) {
 
   MPI_Allreduce(node_sums, tot_sums, 3, MPI_DOUBLE, MPI_SUM, comm_cart);
 
-  p3m.sum_qpart = (int)(tot_sums[0] + 0.1);
+  p3m.sum_qpart = static_cast<int>(tot_sums[0] + 0.1);
```
KaiSzuttor: does somebody know what this hardcoded ```0.1``` means?

src/core/electrostatics_magnetostatics/fft.cpp
```diff
@@ -277,25 +277,21 @@ int calc_send_block(const int *pos1, const int *grid1, const int *pos2,
...
-  int m_out_offset;
-
-  m_in_offset = element * (dim[2] - size[2]);
-  s_in_offset = element * (dim[2] * (dim[1] - size[1]));
-  m_out_offset = (element * size[0]) - element;
-  li_in = element * (start[2] + dim[2] * (start[1] + dim[1] * start[0]));
+  auto const m_out_offset = (element * size[0]) - element;
+  /* linear index of in grid, linear index of out grid */
+  int li_in = element * (start[2] + dim[2] * (start[1] + dim[1] * start[0]));
+  int li_out = 0;
```
KaiSzuttor: you can declare this variable in the loop as ```const``` the compiler will optimise.

src/core/electrostatics_magnetostatics/fft.cpp
```diff
@@ -329,28 +325,22 @@ void pack_block_permute1(double const *const in, double *const out,
...
-  s_out_offset = (element * size[0] * size[1]) - element;
-  li_in = element * (start[2] + dim[2] * (start[1] + dim[1] * start[0]));
-
-  for (s = 0; s < size[0]; s++) { /* mid changing out */
-    m_out_start = element * (s * size[1]);
-    for (m = 0; m < size[1]; m++) { /* fast changing out */
+  auto const s_out_offset = (element * size[0] * size[1]) - element;
+  /* linear index of in grid, linear index of out grid */
+  int li_in = element * (start[2] + dim[2] * (start[1] + dim[1] * start[0]));
+  int li_out = 0;
```
KaiSzuttor: same here

src/core/global.cpp
```diff
@@ -186,21 +186,22 @@ std::size_t hash_value(Datafield const &field) {
 void common_bcast_parameter(int i) {
   switch (fields.at(i).type) {
   case Datafield::Type::INT:
-    MPI_Bcast((int *)fields.at(i).data, fields.at(i).dimension, MPI_INT, 0,
-              comm_cart);
+    MPI_Bcast(static_cast<int *>(fields.at(i).data), fields.at(i).dimension,
```
KaiSzuttor: ```boost::mpi::broadcast```

src/core/global.cpp
```diff
@@ -186,21 +186,22 @@ std::size_t hash_value(Datafield const &field) {
...
+              MPI_INT, 0, comm_cart);
     break;
   case Datafield::Type::UNSIGNED_LONG:
-    MPI_Bcast((unsigned long *)fields.at(i).data, fields.at(i).dimension,
-              MPI_UNSIGNED_LONG, 0, comm_cart);
+    MPI_Bcast(static_cast<unsigned long *>(fields.at(i).data),
```
KaiSzuttor: ```boost::mpi::broadcast```

src/core/global.cpp
```diff
@@ -186,21 +186,22 @@ std::size_t hash_value(Datafield const &field) {
...
+              fields.at(i).dimension, MPI_UNSIGNED_LONG, 0, comm_cart);
     break;
   case Datafield::Type::BOOL:
     static_assert(sizeof(bool) == sizeof(char),
                   "bool datatype does not have the expected size");
-    MPI_Bcast((char *)fields.at(i).data, 1, MPI_CHAR, 0, comm_cart);
+    MPI_Bcast(static_cast<char *>(fields.at(i).data), 1, MPI_CHAR, 0,
```
KaiSzuttor: ```boost::mpi::broadcast```

src/core/global.cpp
```diff
@@ -186,21 +186,22 @@ std::size_t hash_value(Datafield const &field) {
...
+              comm_cart);
     break;
   case Datafield::Type::DOUBLE:
-    MPI_Bcast((double *)fields.at(i).data, fields.at(i).dimension, MPI_DOUBLE,
-              0, comm_cart);
+    MPI_Bcast(static_cast<double *>(fields.at(i).data), fields.at(i).dimension,
```
KaiSzuttor: ```boost::mpi::broadcast```

src/python/espressomd/reaction_ensemble.pyx
```diff
@@ -202,7 +202,7 @@ cdef class ReactionAlgorithm:
         for k in self._valid_keys_add():
             try:
                 self._params[k] = kwargs[k]
-            except BaseException:
+            except Exception:
```
KaiSzuttor: this code looks suspicious

src/core/electrostatics_magnetostatics/elc.cpp
reinaual: Afaik this is the poor mans version of `ceiling` the `double` value. The `double` datatype is used in the first place to `MPI_Allreduce` all tracked sums with one call.

jngrad: This seems a good candidate for `std::round()`, as it stores the number of particles as a `double`.

src/core/electrostatics_magnetostatics/fft.cpp
jngrad: right, I'll declare it in the loop, however it is not `const` as we increment it in the innermost loop

src/python/espressomd/reaction_ensemble.pyx
jngrad: true, it can be rewritten as a conditional

src/core/electrostatics_magnetostatics/elc.cpp
```diff
@@ -1460,7 +1460,7 @@ void ELC_P3M_restore_p3m_sums(ParticleRange const &particles) {
 
   MPI_Allreduce(node_sums, tot_sums, 3, MPI_DOUBLE, MPI_SUM, comm_cart);
 
-  p3m.sum_qpart = static_cast<int>(tot_sums[0] + 0.1);
+  p3m.sum_qpart = static_cast<int>(std::round(tot_sums[0]));
```
KaiSzuttor: sure about this? the ```+0.1``` probably was added to produce ```std::ceil```?

src/core/electrostatics_magnetostatics/fft.cpp
```diff
@@ -283,12 +283,12 @@ void pack_block_permute1(double const *const in, double *const out,
   auto const s_in_offset = element * (dim[2] * (dim[1] - size[1]));
   /* offset for mid changing indices of output grid */
   auto const m_out_offset = (element * size[0]) - element;
-  /* linear index of in grid, linear index of out grid */
+  /* linear index of in grid */
```
KaiSzuttor, line 286: is this docstring missing a word?

src/core/electrostatics_magnetostatics/fft.cpp
```diff
@@ -283,12 +283,12 @@ void pack_block_permute1(double const *const in, double *const out,
...
   int li_in = element * (start[2] + dim[2] * (start[1] + dim[1] * start[0]));
-  int li_out = 0;
 
   for (int s = 0; s < size[0]; s++) { /* fast changing out */
-    li_out = element * s;
+    /* linear index of out grid */
```
KaiSzuttor, line 290: same here

src/core/electrostatics_magnetostatics/fft.cpp
```diff
@@ -331,14 +331,14 @@ void pack_block_permute2(double const *const in, double *const out,
   auto const s_in_offset = element * (dim[2] * (dim[1] - size[1]));
   /* offset for slow changing index of output grid */
   auto const s_out_offset = (element * size[0] * size[1]) - element;
-  /* linear index of in grid, linear index of out grid */
+  /* linear index of in grid */
```
KaiSzuttor, line 334: same here

src/core/electrostatics_magnetostatics/elc.cpp
jngrad: that's precisely why I didn't touch the rounding operation in the first commit :smiley:

We can argue about the intent of the original developer, to me `std::round()` makes sense because in `(int) (x + 0.1)`, the `0.1` will be larger than machine precision for a double (for small x values, obviously), so the expression is effectively `X + 0.1` (with `X` the "real" value), and the static cast to int will lead to a narrowing conversion, so we get X. Since `sum_qpart` is supposed to contain the integer number of particles that contribute to the sum of charges, this checks out. Using `std::ceil` would make the code sensitive to machine precision, e.g. if the double stores X.0000000001 it will become X+1. This would be a silent bug, because the tests ignore the value of `sum_qpart` (incrementing its value changes nothing to the ELC tests outcome).

This `(int) (x + 0.1)` logic is quite fragile, but it's used in the P3M and DP3M code as well, so I'll revert this change to the original version.

src/core/electrostatics_magnetostatics/fft.cpp
jngrad: that's the name of the argument `double const *const in`

src/core/electrostatics_magnetostatics/fft.cpp
jngrad: that's the name of the argument `double const *const out`

src/core/electrostatics_magnetostatics/fft.cpp
KaiSzuttor: okay, i think this comment is not readable and should go, but it's not important

KaiSzuttor (approved):

PR #3859, opened 2020-08-14
Fix CMake and CUDA compiler warnings

### LINKED ISSUE #3797: Warnings in CUDA builds
There are warnings in the CUDA build:

```
espresso/src/utils/include/utils/Span.hpp(88): warning: calling a __host__ function from a __host__ __device__ function is not allowed
          detected during instantiation of "Utils::Span<T>::Span(const C &) [with T=const double, C=std::vector<double, std::allocator<double>>, <unnamed>=std::vector<double, std::allocator<double>>, <unnamed>=std::vector<double, std::allocator<double>>]" 
espresso/src/core/electrostatics_magnetostatics/mmm-common.hpp(49): here
```
which should be fixed. Also why are warnings are not errors for CUDA?

jngrad:
Description of changes:
- fix several issues in the CMake logic
- fix function visibility issues in CUDA code (fixes #3797)
- document the CPU and GPU implementations of `specfunc`


fweik:
Why did you rename `mmm-common.cpp`?

jngrad:
> Why did you rename `mmm-common.cpp`?

I did not touch `mmm-common.cpp`. If you meant `mmm-common.hpp`, I had to move some of its content into a new file `mmm-base.hpp` that can be safely included in `mmm-common_cuda.hpp`. The CUDA compiler warning #3797 comes from the visibility of the CPU implementation of `evaluateAsTaylorSeriesAt()` in the CUDA code, which despite the different signature (2 vs. 3 arguments) caused nvcc to attempt instantiating the CPU function, even though it didn't need it. Furthermore the CUDA code doesn't need to know about the CPU implementation `specfunc.hpp`.

fweik:
Ah sorry, I meant the header of course. `mmm-base.hpp` is pretty nondescript, if not misleading...

jngrad:
> Ah sorry, I meant the header of course. `mmm-base.hpp` is pretty nondescript, if not misleading...

Do you have any suggestion? The file contains so little it's hard to come up with a good name. Maybe `mmm-polygamma.hpp`?

fweik:
Also shouldn't the exported variables and functions be in a cpp file with the same name as the header?

fweik:
I have a feeling that the stuff in `mmm-base.hpp` should be with the spec functions, if I understand the logic correctly. I'd just duplicate the code here, since the specfunction stuff is already duplicated.

fweik:
Ceterum censeo... :-)

jngrad:
> Also shouldn't the exported variables and functions be in a cpp file with the same name as the header?

Yes.

> I have a feeling that the stuff in `mmm-base.hpp` should be with the spec functions, if I understand the logic correctly. I'd just duplicate the code here, since the specfunction stuff is already duplicated.

I would prefer not to duplicate any more code. I only realized today that `specfunc` and `mmm-common` were duplicated for GPU. Storing these duplicates in `/src/core/actor/` instead of `/src/core/electrostatics_magnetostatics/` seems like an anti-pattern to me.

I'll try to limit changes to `specfunc` here to make it easier to backport this PR to the 4.1 release. If we really need to refactor the code via duplication, we can discuss it again in a future PR that I'm working on to remove unused specfunc functions.

fweik:
Well not a hill I'm willing to die on, but for the record I do think that duplication is the best solution given the circumstance.

src/core/electrostatics_magnetostatics/mmm-modpsi.hpp
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+#ifndef MMM_BASE_H
+#define MMM_BASE_H
+
+#include <vector>
+
+/** \name Math constants */
+/*@{*/
+#define C_2PI (2 * M_PI)
```
fweik, line 34: This can be inlined.

src/core/electrostatics_magnetostatics/mmm-modpsi.hpp
```diff
@@ -0,0 +1,45 @@
...
+#define C_GAMMA (0.57721566490153286060651209008)
```
fweik, line 35: `constexpr double ...`

src/core/electrostatics_magnetostatics/mmm-modpsi.hpp
```diff
@@ -0,0 +1,45 @@
...
+/*@}*/
+
+/** Table of the Taylor expansions of the modified polygamma functions */
+extern std::vector<std::vector<double>> modPsi;
+extern int n_modPsi;
```
fweik, line 40: This is just `modPsi.size()` right?

fweik (changes requested):

fweik:
Otherwise this looks good to me, except for the inline comments.

jngrad:
This refactoring is planned, see [jngrad/espresso:specfunc-refactor](https://github.com/jngrad/espresso/tree/specfunc-refactor). The goal of this PR is to fix compiler warnings.

fweik (approved):

PR #3862, opened 2020-08-16
cmake: turn ctest  in cache variable

### LINKED ISSUE #3861: ninja: error: build.ninja:2172: bad $-escape (literal $ must be written as $$)
On Gentoo, I am getting the following error build with ninja:
```
ninja: error: build.ninja:2172: bad $-escape (literal $ must be written as $$)
```

The line in question is:
```
2169 # Custom command for src/CMakeFiles/check_unit_tests
2170
2171 build src/CMakeFiles/check_unit_tests: CUSTOM_COMMAND
2172  COMMAND = cd /var/tmp/portage/sci-physics/espresso-4.1.3/work/espresso-4.1.3_build/src && /usr/bin/ctest --timeout 300 $(ARGS) --output-on-failure
```
and I guess `$(ARGS)` is the issue. This might have been added by me in #2252! However, otherwise I don't see a way to exclude some of the flaky tests.

junghans:
Fixes #3861

Description of changes:
- remove Makefile variable expansion commands from custom targets to improve compatibility with other build systems like `ninja`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -485,7 +485,9 @@ if(WITH_TESTS)
   if(WITH_PYTHON)
     add_subdirectory(testsuite)
   endif(WITH_PYTHON)
-  set(CTEST_ARGS "" CACHE STRING "Extra arguments to give to ctest calls (separated by semicolons)")
+  set(CTEST_ARGS ""
+      CACHE STRING
+            "Extra arguments to give to ctest calls (separated by semicolons)")
 endif(WITH_TESTS)
 
 if(WITH_BENCHMARKS)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -485,7 +485,7 @@ if(WITH_TESTS)
   if(WITH_PYTHON)
     add_subdirectory(testsuite)
   endif(WITH_PYTHON)
-  set(CTEST_ARGS "" CACHE STRING "Extra arguments to give to ctest calls")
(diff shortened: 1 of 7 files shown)


jngrad (approved):

jngrad:
Note sure what this macOS error is about ([logfile](https://github.com/junghans/espresso/runs/1003905310)), but the same test passed in the macOS job in this PR.
```
26/155 Test  #40: magnetostaticInteractions .................................***Failed    0.05 sec
--------------------------------------------------------------------------
A call to mkdir was unable to create the desired directory:

  Directory: /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/ompi.Mac-1597853893506.501/pid.5289
  Error:     Invalid argument

Please check to ensure you have adequate permissions to perform
the desired operation.
--------------------------------------------------------------------------
[Mac-1597853893506.local:05289] [[63771,0],0] ORTE_ERROR_LOG: Error in file util/session_dir.c at line 107
[Mac-1597853893506.local:05289] [[63771,0],0] ORTE_ERROR_LOG: Error in file util/session_dir.c at line 346
--------------------------------------------------------------------------
It looks like orte_init failed for some reason; your parallel process is
likely to abort.  There are many reasons that a parallel process can
fail during orte_init; some of which are due to configuration or
environment problems.  This failure appears to be an internal failure;
here's some additional information (which may only be relevant to an
Open MPI developer):

  orte_session_dir failed
  --> Returned value Error (-1) instead of ORTE_SUCCESS
--------------------------------------------------------------------------
```

PR #3863, opened 2020-08-16
FFT maintenance

jngrad:
Description of changes:
- cleanup documentation
- modernize variable declarations (C++ Core Guidelines [ES.22](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-init), [ES.25](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-const), [ES.49](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-casts-named))

### DIFF
--- a/src/core/electrostatics_magnetostatics/fft.cpp
+++ b/src/core/electrostatics_magnetostatics/fft.cpp
@@ -42,13 +42,8 @@ using Utils::get_linear_index;
 
 #include <utils/Span.hpp>
 
-/************************************************
- * DEFINES
- ************************************************/
-
 /** @name MPI tags for FFT communication */
 /*@{*/
-/** Tag for communication in fft_init() */
 /** Tag for communication in forw_grid_comm() */
 #define REQ_FFT_FORW 301
 /** Tag for communication in back_grid_comm() */
@@ -60,16 +55,15 @@ namespace {
  *  communicate to each other, when you change the node grid.
  *  Changing the domain decomposition requires communication. This
  *  function finds (hopefully) the best way to do this. As input it
- *  needs the two grids (grid1, grid2) and a linear list (node_list1)
- *  with the node identities for grid1. The linear list (node_list2)
+ *  needs the two grids (@p grid1, @p grid2) and a linear list (@p node_list1)
+ *  with the node identities for @p grid1. The linear list (@p node_list2)
  *  for the second grid is calculated. For the communication group of
- *  the calling node it calculates a list (group) with the node
- *  identities and the positions (pos1, pos2) of that nodes in grid1
- *  and grid2. The return value is the size of the communication
+ *  the calling node it calculates a list (@c group) with the node
+ *  identities and the positions (@p my_pos, @p pos) of that nodes in @p grid1
+ *  and @p grid2. The return value is the size of the communication
  *  group. It gives -1 if the two grids do not fit to each other
- *  (grid1 and grid2 have to be component wise multiples of each
- *  other. see e.g. \ref calc_2d_grid in \ref grid.cpp for how to do
- *  this.).
+ *  (@p grid1 and @p grid2 have to be component-wise multiples of each
+ *  other, see e.g. \ref calc_2d_grid for how to do this).
  *
  *  \param[in]  grid1       The node grid you start with.
  *  \param[in]  grid2       The node grid you want to have.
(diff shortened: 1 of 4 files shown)


src/core/electrostatics_magnetostatics/fft.cpp
```diff
@@ -190,17 +184,13 @@ int calc_local_mesh(const int *n_pos, const int *n_grid, const int *mesh,
...
     /* correct round off errors */
-    if ((mesh[i] / (double)n_grid[i]) * (n_pos[i] + 1) - mesh_off[i] - last[i] <
-        1.0e-15)
+    if (ratio * (n_pos[i] + 1) - mesh_off[i] - last[i] < 1.0e-15)
       last[i]--;
-    if (1.0 + (mesh[i] / (double)n_grid[i]) * n_pos[i] - mesh_off[i] -
-            start[i] <
-        1.0e-15)
+    if (1.0 + ratio * n_pos[i] - mesh_off[i] - start[i] < 1.0e-15)
       start[i]--;
```
jngrad, line 194: These "round off" checks look suspicious to me, as the `ceil` and `floor` should have dealt with precision issues already... According to the [coverage report](https://codecov.io/gh/espressomd/espresso/src/17291e1b2f5f63b4d0bead214b67b87a4e1f1068/src/core/electrostatics_magnetostatics/fft.cpp#L188), the `if` bodies are never visited (i.e. `last[i]--` and `start[i]--`).

src/core/electrostatics_magnetostatics/fft.cpp
```diff
@@ -482,8 +472,8 @@ void calc_2d_grid(int n, int grid[3]) {
 } // namespace
 
 int fft_init(const Utils::Vector3i &ca_mesh_dim, int const *ca_mesh_margin,
-             int *global_mesh_dim, double *global_mesh_off, int *ks_pnum,
-             fft_data_struct &fft, const Utils::Vector3i &grid,
+             int const *global_mesh_dim, double const *global_mesh_off,
+             int &ks_pnum, fft_data_struct &fft, const Utils::Vector3i &grid,
              const boost::mpi::communicator &comm) {
```
jngrad, line 477: I tried writing a C++ unit test for the FFT, but `MPI_Cart_coords()` fails with return value `MPI_ERR_TOPOLOGY`, and since this error code is not checked, the unit test continues and initializes an FFT plan with incorrect values that lead to integer overflow. I tried with
```c++
int main(int argc, char **argv) {
  auto mpi_env = mpi_init();
  Communication::init(mpi_env);
  return boost::unit_test::unit_test_main(init_unit_test, argc, argv);
}
```

and

```c++
int main(int argc, char **argv) {
  boost::mpi::environment mpi_env(argc, argv);
  return boost::unit_test::unit_test_main(init_unit_test, argc, argv);
}
```

without success.

This uncaught error code issue is unrelated to the issue with OpenMPI+UCX.

fweik: I'd say this is a logic error, and MPI is trying to tell you that the topology you asked for can not be realized with the number of ranks. Is there any reason why you are using `Communication::init` and not just the lower level functions. I'd say calling into the core already partially defeats the purpose of the test.

fweik: Btw. I *really* think the code should be moved out of the core before adding tests. We have to check that it is possible to run this standalone, otherwise we have to fix it. Logically, the fft code can node depend on core code. You don't have to know about espresso to do fft's ...

src/core/electrostatics_magnetostatics/fft.cpp
fweik: I don't know, I don't think its worth while to think about this.

fweik (approved):

src/core/electrostatics_magnetostatics/fft.cpp
fweik: More to your point: You should use the cartesian utility functions from utils/mpi, they have error check, and throw is there is a problem.

jngrad: Thanks, this sorts out the MPI error. The test draft is in 68b2bf081f7, I will not be working on it any further. I don't understand what the extra "padding" is for (i.e. the extension of buf_size, not the margin), I get a different FFT output depending on the number of MPI ranks, the fftw3 on a 3D grid of cosines returns 0 instead of the the fundamental modes, and ASAN warns about direct and indirect memory leaks.

PR #3871, opened 2020-08-21
Fix CMake logic for imported targets

jngrad:
A few issues with our build system have been revealed by the waLBerla branch. See the section on imported targets in [cmake-developer(7)](https://cmake.org/cmake/help/v3.18/manual/cmake-developer.7.html?highlight=namespaced#a-sample-find-module) for more details on namespaces. Boost dependencies were set up so as to include non-existent directories.


src/utils/CMakeLists.txt
```diff
@@ -1,8 +1,7 @@
 add_library(EspressoUtils INTERFACE)
 target_include_directories(
-  EspressoUtils
-  INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
-            $<INSTALL_INTERFACE:include> Boost::serialization Boost::mpi)
+  EspressoUtils INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
```
fweik, line 3: Logically (and physically) `EspressoUtils` does depend on `Boost::mpi`, no?

jngrad: What is the right way to encode that dependency? Right now, `INTERFACE Boost::serialization Boost::mpi` only adds `-I/work/jgrad/espresso-fork-PR/src/utils/Boost::serialization -I/work/jgrad/espresso-fork-PR/src/utils/Boost::mpi ` to the compiler options. We could add `${Boost_INCLUDE_DIRS}`, but this is just a string and not a dependency, and on my workstation it gets stripped from the compiler options anyway (probably because it's `/usr/include`, which is redundant).

fweik: That should be `target_link_libraries` I think. Clearly the code that was there was not correct, but I think if there is a way to properly declare the dependencies, it should be done. It was may (maybe naiive) understanding that this should be possible with imported targets. Again, not an expert, but it would be my expectation that I only have to `target_link_libraries(<target> PRIVATE EspressoUtils`) if I wanted to use the library.

jngrad: I'm also not an expert. Right now `src/utils/tests/CMakeLists.txt` contains commands to add `Boost::serialization` and/or `Boost::mpi` in `target_link_libraries` as needed to compile tests that depend on `EspressoUtils`. Now we could add `target_link_libraries( EspressoUtils INTERFACE Boost::serialization Boost::mpi)` to automatically forward boost to targets that depend on `EspressoUtils`, I just tried it and it seems to work. To my surprise `Boost::mpi` doesn't seem to include `mpi.h`, which is good because CUDA code depends on `EspressoUtils` yet cannot include `mpi.h`.

fweik: Yeah arguably this is a bug in FindBoost. From the previous reasoning MPI should also be included right?

jngrad: I guess so. If there is ever a fix, it shouldn't be any problem for us. My concern about `mpi.h` was unfounded, because it's only an issue if we actually write `#include <mpi.h>` in the source code. It's perfectly fine to link a CUDA target against `MPI::MPI_CXX`.

fweik: Yes, I agree. Otherwise, the mpi stuff would have to go into an extra target to avoid spreading the dependency. If the consumer of the target has to know about the dependencies, that just feels wrong to me ^^

cmake/FindFFTW3.cmake
```diff
@@ -38,7 +38,7 @@ find_package_handle_standard_args(FFTW3 DEFAULT_MSG FFTW3_LIBRARIES
 
 mark_as_advanced(FFTW3_LIBRARIES FFTW3_INCLUDE_DIR)
 
-if(FFTW3_FOUND)
+if(FFTW3_FOUND AND NOT TARGET FFTW3::FFTW3)
```
KaiSzuttor, line 41: why do we need the check for the target?

fweik: It is an error to define the target if it already exists. this is common practice I think. Also walberla defines the target too IIRC.

jngrad: because walberla also imports the FFTW3 target, causing a collision:
```
CMake Error at cmake/FindFFTW3.cmake:42 (add_library):
  add_library cannot create imported target "FFTW3::FFTW3" because another
  target with the same name already exists.
Call Stack (most recent call first):
  build-seraue-release/_deps/walberla-src/CMakeLists.txt:963 (find_package)


CMake Error at cmake/FindFFTW3.cmake:44 (target_link_libraries):
  Cannot specify link libraries for target "FFTW3::FFTW3" which is not built
  by this project.
Call Stack (most recent call first):
  build-seraue-release/_deps/walberla-src/CMakeLists.txt:963 (find_package)
```

[cmake-developer(7)](https://cmake.org/cmake/help/v3.18/manual/cmake-developer.7.html?highlight=namespaced#a-sample-find-module)  says:
> When providing imported targets, these should be namespaced (hence the `Foo::` prefix); CMake will recognize that values passed to `target_link_libraries()` that contain `::` in their name are supposed to be imported targets (rather than just library names), and will produce appropriate diagnostic messages if that target does not exist (see policy CMP0028).
> ```cmake
> if(Foo_FOUND AND NOT TARGET Foo::Foo)
>   add_library(Foo::Foo UNKNOWN IMPORTED)
>   ...
> endif()
> ```

This is documented in the commit message 93d30a8e3d.

KaiSzuttor: sorry, I'm new to meaningful commit messages :smile:

KaiSzuttor (dismissed):

fweik:
I thought MPI should also be added to the target? Did you read the conversation @KaiSzuttor ?

fweik:
I though' we'd add `MPI::MPI_CXX` to the public dependencies of `EspressoUtils`?

jngrad:
I'm a little lost about what we're trying to achieve here. Do we want to explicitly spell out the dependencies on every target, in order to make the Boost libraries "`PRIVATE`" in `EspressoUtils` (I'm putting quotes here because Utils is a header only lib, so it can only be `INTERFACE` ) and make the unit tests compile even if we remove e.g. `Boost::serialization` in the future from `EspressoUtils`? Or do we want to have Boost libraries and MPI as public dependencies of `EspressoUtils` as originally planned, so that unit tests and `EspressoCore` automatically inherit from them?

fweik:
> do we want to have Boost libraries and MPI as public dependencies of EspressoUtils as originally planned, so that unit tests and EspressoCore automatically inherit from them?

This, independently, but it is not related how the target is used. They have to be public because they are used at the interface of `EspressoUtils`, at least that is my understanding of best practices.

fweik (approved):

PR #3787, opened 2020-07-02
WIP: short range and cells refactor

fweik:
Depends on #3780.

Exposition only for now.

Description of changes:
-

PR #3874, opened 2020-08-25
Refactor short range loop and cells

jngrad:
Rewrite of #3787

Description of changes:
- simplify short range loop
   - remove `for_each_pair.hpp ` and `verlet_ia.hpp` helper functions
   - delegate distance calculation and cell iteration to the `CellStructure` class
- hide implementation details of `CellStructure`
   - make `local_cells()` private
   - encapsulate debug functions
   - remove global variables `n_verlet_updates` and `rebuild_verletlist`


jngrad:
There is a regression that causes a significant performance drop in Debug builds with coverage enabled (8x slowdown). The tests that are timing out (`mass-and-rinertia_per_particle` and `rotational-diffusion-aniso`) are statistical tests, whose runtime vary a lot between two runs, which makes bisection a lot more difficult. Furthermore the `rotational-diffusion-aniso` test had the wrong fixture and never executed `test_case_10` for Brownian Dynamics. The test passes on the python branch but not in this PR:
```
1/155 Test #59: rotational-diffusion-aniso .......***Failed   25.26 sec
FAIL: test_case_10 (__main__.RotDiffAniso)
  File "/ssd/jgrad/espresso-fork-PR/build-seraue-release/testsuite/python/rotational-diffusion-aniso.py", line 383, in test_case_10
    self.check_rot_diffusion(n)
  File "/ssd/jgrad/espresso-fork-PR/build-seraue-release/testsuite/python/rotational-diffusion-aniso.py", line 327, in check_rot_diffusion
    'diffusion is too large: {1}'.format(j, dcosjj_dev[j]))
AssertionError: 1.0 not less than or equal to 0.2 : Relative deviation dcosjj_dev[0] in a rotational diffusion is too large: 1.0

Ran 3 tests in 19.858s

FAILED (failures=1)
```

fweik:
I'd say it is probably worthwhile to find better tests than the one timing out. Debug-build performance is a problem in general in C++, and to put time into this to accommodate tests that are not up to par in the first place feels not like the right way to go to me.

jngrad:
I think the regression was in 1bf4f041b, which accidentally removed the `INACTIVE_CUTOFF` check.

src/core/CellStructure.hpp
```diff
@@ -33,6 +33,7 @@
 #include "algorithm/link_cell.hpp"
 #include "bond_error.hpp"
 #include "ghosts.hpp"
+#include "nonbonded_interactions/nonbonded_interaction_data.hpp"
```
fweik: This is not needed (you ended up not using the constant). It also drags in globals, so should not be here in any case.

src/core/CellStructure.hpp
```diff
@@ -516,7 +517,8 @@ struct CellStructure {
    * @param pair_kernel Kernel to apply
    */
   template <class PairKernel> void non_bonded_loop(PairKernel pair_kernel) {
-    link_cell(pair_kernel);
+    if (maximal_cutoff() > 0.)
```
fweik: This should not include the cutoff zero case, e.g. for maximal_cutoff() == 0. the short range loop should be executed, it can only be skipped if the cutoff is smaller zero.

src/core/CellStructure.hpp
jngrad: I need this include for the prototype of `maximal_cutoff()`

src/core/CellStructure.hpp
jngrad: I copied this code from
https://github.com/espressomd/espresso/blob/64cf40060bacbe9043caa0d9e1a8fd3e0643019c/src/core/integrate.cpp#L511-L515
should I include 0 in `interaction_range()` too?

src/core/CellStructure.hpp
fweik: But that operates on globals?

src/core/CellStructure.hpp
fweik: Maybe it is ok, I don't remember why this is. Logically it should be ok to exclude the zero case I think.

src/core/CellStructure.hpp
fweik: So please don't do this. This is slightly shocking to me.

jngrad: The function iterates over all interactions data structures. I could do a forward declaration of the `maximal_cutoff()` prototype to avoid including the complete header file.

src/core/CellStructure.hpp
jngrad: It was probably done to avoid division by zero warnings in static analysis.

src/core/CellStructure.hpp
fweik: This is worse. If you need to know the maximal cutoff, it should be passed in right? You should also check if you can move the decision if the function should be called up the call stack.

RudolfWeeber:
@jngrad, could you please run the LJ benchmark on this, e.g., because of the use of boost::optional in the hot loop.
I did not yet read everything. Will take 1-2 more days.

RudolfWeeber:
Why is there a default argument to distance_cutoff in the short_range-loop() function, and why is it 1?

jngrad:
> Why is there a default argument to distance_cutoff in the short_range-loop() function, and why is it 1?

The default argument in the short range loop has a positive non-zero value such that the pair kernel is always executed, except when you provide a custom cutoff value as the third argument. If that cutoff value is negative (i.e. there are no non-bonded interactions), the pair kernel isn't executed. We do not have an immediate use for it anymore, now that we always re-calculate the non-bonded cutoff to fix time outs in CI.

> could you please run the LJ benchmark on this, e.g., because of the use of boost::optional in the hot loop.

|  commit |    label |  empty | default | maxset |
| ------- | -------- | ------ | ------- | ------ |
| 986714b | base commit |  56.4% |  100.0% | 101.5% |
| 66c7a94 | final state |  58.7% |  102.0% | 102.9% |

RudolfWeeber (approved):

PR #3878, opened 2020-08-27
Bump Python and CMake versions

jngrad:
Description of changes:
- bump CMake version requirement from 3.10 to 3.11 to get full support for imported targets
    * fixes a cryptic error message when compiling ESPResSo with FFTW3 using CMake 3.10
- bump Python version requirement from 3.5 to 3.6 (NEP 29, see drop table in #3421)
    * enables f-strings
- fix flaw in `espressomd.utils.check_type_or_throw_except()` that lead to subtle errors when arrays had the wrong size
- add tests for functions in `espressomd.utils`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -166,7 +166,7 @@ if(WITH_CUDA)
   endif()
 endif(WITH_CUDA)
 
-find_package(PythonInterp 3.5 REQUIRED)
+find_package(PythonInterp 3.6 REQUIRED)
 
 if(WITH_PYTHON)
   find_package(Cython 0.26 REQUIRED)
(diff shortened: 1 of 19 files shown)


KaiSzuttor (approved):
Do we have any documentation for the python requirements?

jngrad:
> Do we have any documentation for the python requirements?

Not that I know of. If we did, how would we make sure the docs are kept in sync with CMake? For python packages, the requirements are easily extracted (https://github.com/espressomd/espresso/network/dependencies), but for Python/CMake/CUDA/Boost versions this is more tricky, as we have conditional requirements between these components that depend on the user environment. We could state that a NEP29-compliant Python environment is recommended to build the project (we actually lag 6 months behind NEP29, which gives our users more flexibility), but we only test the latest versions of Python packages in CI indirectly via bug reports from the 4.1 build on Fedora Koji.

jngrad:
Thanks @schlaicha for reporting the CMake 3.10 issue to the core team.

Issue #3877, opened 2020-08-26
Division by zero and buffer overflow in 4.1 DP3M


jngrad:
MWE:
```python
import espressomd
import espressomd.magnetostatics
system = espressomd.System(box_l=[10., 10., 10.])
system.part.add(pos=[[0, 0, 0], [.5, .5, .5]],
                rotation=2 * [(1, 1, 1)], dip=2 * [(1, 0, 0)])
system.time_step = 0.01
solver = espressomd.magnetostatics.DipolarP3M(prefactor=2, accuracy=1e-2)
system.actors.add(solver)
```

With 4.1.3 and Clang 6 in docker:
```sh
docker run --user espresso --runtime=nvidia -it gitlab.icp.uni-stuttgart.de:4567/espressomd/docker/clang-python3:6.0 bash
git clone --depth=30 --recursive -b 4.1 https://github.com/espressomd/espresso.git
cd espresso
git checkout 098adfe32
export with_ccache=true make_check_python=false make_check_unit_tests=false build_procs=$(nproc)
export myconfig=maxset with_coverage=false with_static_analysis=false with_asan=true with_ubsan=true
bash maintainer/CI/build_cmake.sh
cd build
vi mwe.py # here copy MWE
./pypresso mwe.py
```
```
/home/espresso/espresso/src/core/electrostatics_magnetostatics/p3m-dipolar.cpp:1322:16: runtime error: division by zero
    #0 0x7f8aeaa3afbf in dp3m_calc_influence_function_force() /home/espresso/espresso/src/core/electrostatics_magnetostatics/p3m-dipolar.cpp:1322:16
    #1 0x7f8aeaa35828 in dp3m_scaleby_box_l() /home/espresso/espresso/src/core/electrostatics_magnetostatics/p3m-dipolar.cpp:2436:3
    #2 0x7f8aeaa758c1 in dp3m_mcr_time(int, int, double, double) /home/espresso/espresso/src/core/electrostatics_magnetostatics/p3m-dipolar.cpp:1551:3
    #3 0x7f8aeaa74acf in dp3m_mc_time(char**, int, int, double, double, double*, double*, double*) /home/espresso/espresso/src/core/electrostatics_magnetostatics/p3m-dipolar.cpp:1656:14
[...]

PR #3879, opened 2020-08-28
Remove duplicated functions in DP3M code

jngrad:
Part of the effort to reduce divergences between P3M and DP3M code. Follow-up to #3841.

Description of changes:
- group common members of the P3M and DP3M code in a base struct
- replace the 4 duplicated versions of `calc_meshift()` by a single one that works for all edge cases (unit tested)
- create an analogue of the P3M `G_opt<S,m>()` function for DP3M to reduce code duplication
- fix `heap-buffer-overflow` bug caused by a mesh size synchronization issue between P3M and FFT data structs
- fix `DipolarP3M` actor: remove unused `inter` and `inter2` parameters, fix range check on `mesh` that caused checkpointing to fail
- add a python test for the non-metallic case of DP3M
- fix unittest classes for P3M/DP3M tests (the list of actors must be cleared during teardown to avoid side-effects)
- modernize variable declarations (C++ Core Guidelines [ES.22](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-init), [ES.25](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-const), [ES.49](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-casts-named), [F.20](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f20-for-out-output-values-prefer-return-values-to-output-parameters))

### DIFF
--- a/src/core/electrostatics_magnetostatics/dp3m_influence_function.hpp
+++ b/src/core/electrostatics_magnetostatics/dp3m_influence_function.hpp
@@ -25,6 +25,7 @@
 
 #include <utils/Vector.hpp>
 #include <utils/constants.hpp>
+#include <utils/index.hpp>
 #include <utils/math/int_pow.hpp>
 #include <utils/math/sinc.hpp>
 #include <utils/math/sqr.hpp>
@@ -32,6 +33,8 @@
 #include <boost/range/numeric.hpp>
 
 #include <cmath>
+
+#if defined(DP3M)
 
 /** Calculate the aliasing sums for the optimal influence function.
  *
@@ -77,7 +80,8 @@ double G_opt_dipolar(P3MParameters const &params, Utils::Vector3i const &shift,
       }
     }
   }
-  return numerator / (int_pow<S>(d_op.norm2()) * Utils::sqr(denominator));
+  return numerator / (int_pow<S>(static_cast<double>(d_op.norm2())) *
+                      Utils::sqr(denominator));
 }
 
 /**
@@ -213,4 +217,5 @@ double grid_influence_function_self_energy(P3MParameters const &params,
   return energy;
 }
 
+#endif
 #endif
--- a/src/core/electrostatics_magnetostatics/dp3m_influence_function.hpp
+++ b/src/core/electrostatics_magnetostatics/dp3m_influence_function.hpp
@@ -144,4 +144,73 @@ std::vector<double> grid_influence_function(P3MParameters const &params,
   return g;
 }
 
+double G_opt_dipolar_self_energy(P3MParameters const &params,
+                                 Utils::Vector3i const &shift) {
+  using Utils::sinc;
+  double u_sum = 0.0;
+  constexpr int limit = P3M_BRILLOUIN + 5;
+
+  auto const f1 = 1.0 / static_cast<double>(params.mesh[0]);
+
+  for (double mx = -limit; mx <= limit; mx++) {
+    auto const nmx = shift[0] + params.mesh[0] * mx;
+    auto const sx = std::pow(sinc(f1 * nmx), 2.0 * params.cao);
+    for (double my = -limit; my <= limit; my++) {
+      auto const nmy = shift[1] + params.mesh[0] * my;
+      auto const sy = sx * std::pow(sinc(f1 * nmy), 2.0 * params.cao);
+      for (double mz = -limit; mz <= limit; mz++) {
+        auto const nmz = shift[2] + params.mesh[0] * mz;
+        auto const sz = sy * std::pow(sinc(f1 * nmz), 2.0 * params.cao);
+        u_sum += sz;
+      }
+    }
+  }
+  return u_sum;
+}
+
+/**
+ * @brief Calculate self-energy of the influence function.
+ *
+ * @param params DP3M parameters
+ * @param n_start Lower left corner of the grid
+ * @param n_end Upper right corner of the grid.
+ * @param g Energies on the grid.
+ * @return Total self-energy.
+ */
+double grid_influence_function_self_energy(P3MParameters const &params,
+                                           Utils::Vector3i const &n_start,
+                                           Utils::Vector3i const &n_end,
+                                           std::vector<double> const &g) {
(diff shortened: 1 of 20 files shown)


jngrad:
Without the workaround in 0a66cf98ef, I get the following ASAN report in Clang 9:
```
==3811888==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x604000046374 at pc 0x14eb7923a6b9 bp 0x7fff96e7a070 sp 0x7fff96e7a068
READ of size 4 at 0x604000046374 thread T0
    #0 0x14eb7923a6b8 in std::vector<double, std::allocator<double> > grid_influence_function<3ul>(P3MParameters const&, Utils::Vector<int, 3ul> const&, Utils::Vector<int, 3ul> const&, Utils::Vector<double, 3ul> const&) /work/jgrad/espresso-fork-PR/src/core/electrostatics_magnetostatics/dp3m_influence_function.hpp:139:46
    #1 0x14eb79214da6 in dp3m_calc_influence_function_force() /work/jgrad/espresso-fork-PR/src/core/electrostatics_magnetostatics/p3m-dipolar.cpp:722:18
    #2 0x14eb79214a4f in dp3m_scaleby_box_l() /work/jgrad/espresso-fork-PR/src/core/electrostatics_magnetostatics/p3m-dipolar.cpp:1475:3
    #3 0x14eb792212c2 in dp3m_adaptive_tune(char**) /work/jgrad/espresso-fork-PR/src/core/electrostatics_magnetostatics/p3m-dipolar.cpp:1213:3
   ... cython frames ...

0x604000046374 is located 0 bytes to the right of 36-byte region [0x604000046350,0x604000046374)
allocated by thread T0 here:
    #0 0x14eb8361431d in operator new(unsigned long) /build/llvm-toolchain-9-uSl4bC/llvm-toolchain-9-9/projects/compiler-rt/lib/asan/asan_new_delete.cc:99:3
    #1 0x14eb7cb176b5 in std::_Vector_base<int, std::allocator<int> >::_M_create_storage(unsigned long) /usr/bin/../lib/gcc/x86_64-linux-gnu/7.5.0/../../../../include/c++/7.5.0/bits/stl_vector.h:187:33
    #2 0x14eb7cb17357 in std::_Vector_base<int, std::allocator<int> >::_Vector_base(unsigned long, std::allocator<int> const&) /usr/bin/../lib/gcc/x86_64-linux-gnu/7.5.0/../../../../include/c++/7.5.0/bits/stl_vector.h:138:9
    #3 0x14eb7cb186a9 in std::vector<int, std::allocator<int> >::vector(unsigned long, std::allocator<int> const&) /usr/bin/../lib/gcc/x86_64-linux-gnu/7.5.0/../../../../include/c++/7.5.0/bits/stl_vector.h:284:9
    #4 0x14eb791fdc9e in detail::calc_meshift(std::array<int, 3ul> const&, bool) /work/jgrad/espresso-fork-PR/src/core/electrostatics_magnetostatics/p3m-common.hpp:204:14
    #5 0x14eb79238df2 in std::vector<double, std::allocator<double> > grid_influence_function<3ul>(P3MParameters const&, Utils::Vector<int, 3ul> const&, Utils::Vector<int, 3ul> const&, Utils::Vector<double, 3ul> const&) /work/jgrad/espresso-fork-PR/src/core/electrostatics_magnetostatics/dp3m_influence_function.hpp:122:7
    #6 0x14eb79214da6 in dp3m_calc_influence_function_force() /work/jgrad/espresso-fork-PR/src/core/electrostatics_magnetostatics/p3m-dipolar.cpp:722:18
    #7 0x14eb79214a4f in dp3m_scaleby_box_l() /work/jgrad/espresso-fork-PR/src/core/electrostatics_magnetostatics/p3m-dipolar.cpp:1475:3
    #8 0x14eb792212c2 in dp3m_adaptive_tune(char**) /work/jgrad/espresso-fork-PR/src/core/electrostatics_magnetostatics/p3m-dipolar.cpp:1213:3
   ... cython frames ...
```
This is caused by the value of `n[2]` being equal to the size of `shifts[0]`, causing an access out of bounds. There is a discrepancy between the FFT mesh size `dp3m.fft.plan[3].new_mesh` and the DP3M one in `dp3m.params.mesh` that doesn't happen in the P3M code. Could be the same issue as in #3877.

RudolfWeeber:
lgtm. But where do the magic numbers in the test originate from?

jngrad:
I ran the solver once with high accuracy and extracted the tuned parameters. The test `magnetostaticInteractions.py` creates a DipolarP3M actor with these tuned parameters and skips tuning to make the outcome deterministic. You can then scale the reference energies by the prefactor and box volume. This is the same approach that is used in `electrostaticInteractions.py`.

While writing that test I found out the magnetostatic force between two dipoles didn't change with `epsilon`. Is this expected?

The FFT issue still needs to be addressed. It could very well be that a single line is missing somewhere to update the mesh, but the FFT plan creation logic in DP3M is not very readable.

jngrad:
The `heap-buffer-overflow` should be fixed by 49ad51e75d. This solution is based on the P3M code. We can't add `assert` statements in the `(d)p3m_calc_influence_function_*()` function to catch this bug, because we often get FFT plans where one of the 3 values in vector `Vector3i end = Vector3i{dp3m.fft.plan[3].start} + Vector3i{dp3m.fft.plan[3].new_mesh}` exceeds the size of the vectors calculated by `detail::calc_meshift()`. This doesn't cause buffer overflows because it just so happens that the value exceeding the buffer size is never read in the normal flow of the program.

I didn't run the benchmarks because my workstation is extremely slow, and not just for file I/O (Xorg issue). If I understand the code correctly, the functions touched in this PR are only executed during initialization, tuning and parameters change, so it's unlikely the performance of simulations will be affected.

Using a non-cubic box should be possible at some point, but this will require finding a new FFT library to replace the existing custom code. I already looked into FFTW3's 3D grid FFT recently to write a test (68b2bf081), but couldn't get it to work.

RudolfWeeber (approved):

PR #3881, opened 2020-08-31
Tutorial 1 - Lennard-Jones

### LINKED ISSUE #3818: Tutorial 1: LJ
* the overlap removal should be done using steepest descent

* the odd indexing
```
for i in range(1,n+1)
```
in the sampling loop needs to be changed to zero-based

* The tutorial assumes without justification that the energy time series is uncorrelated. 
At the very least, an acf of the data needs to be plotted so the auto-corrleaiton time can be read  visually. It needs to be stated that the effective sample count in the error calculation is for a minimum sampling distance of two auto-correlation times.
I don't think, we need to do the full underlying math in the first tutorial.

* The concept of observables and accumulators needs to be explained in 1-2 sentenes.

schlaicha:
Refactor and improve the LJ tutorial.

Fixes #3818

Description of changes:
- now use to zero-based indexing in the loop


doc/tutorials/01-lennard_jones/01-lennard_jones.ipynb
```diff
@@ -424,14 +424,14 @@
     "instantaneous_temperature = np.zeros(sampling_iterations)\n",
     "etotal = np.zeros(sampling_iterations)\n",
     "\n",
-    "for i in range(1, sampling_iterations + 1):\n",
+    "for i in range(0, sampling_iterations):\n",
```
KaiSzuttor: starting from ```0``` is the default and it is good style to omit the default value here.

schlaicha: Fixed in a571818

RudolfWeeber:
We will not be able to support alternative solutions. It would require running the test several times.

The one without the loop is better practice (more declarative), so let's keep that.

RudolfWeeber:
I added learning objectives to the tutor's notes

schlaicha:
> We will not be able to support alternative solutions. It would require running the test several times.
> 
> The one without the loop is better practice (more declarative), so let's keep that.

Agreed and pushed in 7562149

schlaicha:
@jngrad I accidentally deleted my comments in the started review discussion. Only thing left from there is a possibility to check all the links in the notebooks using the CMake infrastructure.

doc/tutorials/01-lennard_jones/01-lennard_jones.ipynb
```diff
@@ -170,13 +170,49 @@
...
+    "solution2": "hidden",
+    "solution2_first": true
+   },
+   "source": [
+    "**Exercise:**\n",
+    "\n",
+    "* Create an instance of an espresso system and store it in a variable called <tt>system</tt>;\n",
+    "  use <tt>BOX_L<tt> as box length.\n",
+    "\n",
+    "See [ESPResSo documentation](http://espressomd.org/html/doc/system_setup.html) and [module documentaion](http://espressomd.org/html/doc/espressomd.html#module-espressomd.system)."
```
jngrad:
```suggestion
    "See [ESPResSo documentation](http://espressomd.org/html/doc/system_setup.html) and [module documentation](http://espressomd.org/html/doc/espressomd.html#module-espressomd.system)."
```

doc/tutorials/01-lennard_jones/01-lennard_jones.ipynb
```diff
@@ -170,13 +170,49 @@
...
+   "cell_type": "markdown",
+   "metadata": {
+    "solution2": "hidden",
+    "solution2_first": true
+   },
+   "source": [
+    "**Exercise:**\n",
+    "\n",
+    "* Create an instance of an espresso system and store it in a variable called <tt>system</tt>;\n",
+    "  use <tt>BOX_L<tt> as box length.\n",
```
jngrad:
```suggestion
    "  use <tt>BOX_L</tt> as box length.\n",
```
took me a while to figure out which line caused half the tutorial to be in monospace font :)

doc/tutorials/01-lennard_jones/01-lennard_jones.ipynb
```diff
@@ -336,42 +456,69 @@
...
-   "metadata": {},
+   "metadata": {
+    "solution2": "hidden",
+    "solution2_first": true
+   },
    "source": [
-    "### Integrating equations of motion and taking measurements\n",
+    "**Exercise:**\n",
+    "\n",
+    "* Use [<tt>espressomd.minimize_energy</tt>](http://espressomd.org/html/doc/espressomd.html#module-espressomd.minimize_energy) to make sure that there are no forces > <tt>F_MAX<tt>.\n",
```
jngrad:
```suggestion
    "* Use [<tt>espressomd.minimize_energy</tt>](http://espressomd.org/html/doc/espressomd.html#module-espressomd.minimize_energy) to make sure that there are no forces > <tt>F_MAX</tt>.\n",
```

doc/tutorials/01-lennard_jones/01-lennard_jones.ipynb
```diff
@@ -314,20 +405,49 @@
...
+   "metadata": {
+    "solution2": "hidden",
+    "solution2_first": true
+   },
+   "source": [
+    "**Exercise:**\n",
+    "\n",
+    "* Setup a Lennard-Jones interaction with $\\epsilon=$<textt>LJ_EPS</textt> and $\\sigma=$<textt>LJ_SIG</textt> which are cut and shifted to zero at $r_\\mathrm{c}=$<textt>LJ_CUT</textt>$\\times\\sigma$.\n",
+    "\n",
+    "  *Hint:* Use the <textt>[espressomd.system](http://espressomd.org/html/doc/espressomd.html#module-espressomd.system).non_bonded_inter</textt> handle object to instantiate a [espressomd.interactions.LennardJonesInteraction](http://espressomd.org/html/doc/espressomd.html#module-espressomd.interactions)."
```
jngrad, line 400: `<textt>` is not a valid tag and prevents the creation of the link to the user guide

doc/tutorials/01-lennard_jones/NotesForTutor.md
```diff
@@ -0,0 +1,34 @@
...
+After the tutorial, students should be able to 
+
+* explain
+    * role of Lennard-Jones potential
+    * role of radial distribution function
+    * connection between mean squared displacement and diffusion coefficient
+    * how to use an auto correlation function to estimate correlation times and how that affects error estimation
+
+
+## Learning objectives (Espresso)
```
jngrad:
```suggestion
## Learning objectives (ESPResSo)
```

doc/tutorials/01-lennard_jones/01-lennard_jones.ipynb
```diff
@@ -336,42 +456,69 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "solution2": "hidden"
+   },
+   "outputs": [],
+   "source": [
+    "from espressomd.minimize_energy import steepest_descent\n",
```
jngrad: `espressomd.minimize_energy.steepest_descent()` is [deprecated](https://github.com/espressomd/espresso/wiki/es-420-release-notes#changed-and-removed-functionality) in ESPResSo 4.2 in favor of `system.integrator.set_steepest_descent()`, which takes the same arguments. You should also activate the Langevin thermostat after steepest descent. Declaring it before will prevent steepest descent from following the gradient. Here is a MWE to reproduce the bug:
```python
import espressomd
system = espressomd.System(box_l=[1.0, 1.0, 1.0])
system.cell_system.set_domain_decomposition(use_verlet_lists=True)
system.cell_system.skin = 0.2
system.time_step = 0.01
system.non_bonded_inter[0, 0].wca.set_params(epsilon=1, sigma=0.1)
system.part.add(pos=[0, 0, 0])
system.part.add(pos=[0.1, 0, 0])
#system.thermostat.set_langevin(kT=2, gamma=2, seed=42)  # uncomment this line to get the bug
system.integrator.set_steepest_descent(f_max=0, gamma=1e-3, max_displacement=0.001)
system.integrator.run(1)
print(system.part[:].pos)
```
without Langevin, particles move away from each other along the gradient line:
```
[[-0.001  0.     0.   ]
 [ 0.101  0.     0.   ]]
```
with Langevin, random forces are added:
```
[[-0.001       0.001       0.001     ]
 [ 0.101       0.00060591 -0.001     ]]
```
This silent bug will generate an error message once #3880 is merged.

jngrad (changes requested):
Thanks for writing links to the user guide, this will be very helpful to new users!

Other points to address, that are not part of the diff:
* The MSD correlator explanation is no longer accurate. The first two columns don't store the lag time and sample size anymore, instead one has to call `msd_corr.lag_times()` explicitly to get the lag times. This regression was introduced in a previous PR.
* The introduction paragraph needs some cleanup. The core is no longer written in the C language.

doc/tutorials/01-lennard_jones/01-lennard_jones.ipynb
schlaicha: Sorry, I only looked at the following code cells which rendered fine due to cancellation of errors :-/

jngrad:
@schlaicha please use batch commits, we have limited CI resources :)

schlaicha:
Sorry, was too nice to just click on the `commit change` button. Is there a way to batch commit on the github UI or do I need to manually do this?

jngrad:
click "Add suggestion to batch" ([GitHub docs](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/incorporating-feedback-in-your-pull-request))

doc/tutorials/01-lennard_jones/01-lennard_jones.ipynb
schlaicha: Thanks for noticing. Fxed in 4ba0765.

doc/tutorials/01-lennard_jones/01-lennard_jones.ipynb
schlaicha: Overlooked that one. Fixed in 4ba0765.

schlaicha:
> Other points to address, that are not part of the diff:
> 
>     * The MSD correlator explanation is no longer accurate. The first two columns don't store the lag time and sample size anymore, instead one has to call `msd_corr.lag_times()` explicitly to get the lag times. This regression was introduced in a previous PR.
> 
>     * The introduction paragraph needs some cleanup. The core is no longer written in the C language.

Thanks, I will work on these points.

doc/tutorials/01-lennard_jones/01-lennard_jones.ipynb
```diff
@@ -336,42 +405,142 @@
...
+    "    prev_energy = system.analysis.energy()['total']\n",
+    "    system.integrator.run(EMSTEP)\n",
+    "    maxforce = np.max(np.abs(system.part[:].f))\n",
+    "    relforce = np.abs((maxforce-prev_maxforce)/prev_maxforce)\n",
+    "    energy = system.analysis.energy()['total']\n",
+    "    relener = np.abs((energy-prev_energy)/prev_energy)\n",
+    "    print(\"minimization step: {:4.0f}\\tmax. rel. force change:{:+3.3e}\\trel. energy change:{:+3.3e}\".format((i+1)*EMSTEP,relforce, relener))\n",
+    "    if relforce < F_TOL or relener < E_TOL:\n",
+    "        break\n",
+    "    i += 1"
```
jngrad: @schlaicha @christophlohrmann should we mention somewhere in your tutorials that this construction is for learning purposes and shouldn't always be copy-pasted in a production script? This construction is not wrong, and the steepest descent integrator was designed to be used in that way (it's referred to as a custom-criterion steepest descent in the user guide), but it won't always work out-of-the-box in a different simulation script:
* the gradient descent stops when the system evolves slowly, which could lead to early exit in dense LJ fluids
* the convergence criterion is based on forces measured by their components while the integrator uses their norm: `np.max(np.abs(system.part[:].f))` vs. `np.max(np.linalg.norm(system.part[:].f, axis=1))`
* the criterion for early exit is based on the maximal force on all particles, whereas the integrator actually works on a subset of all particles

The last point can lead to silent bugs when re-using that steepest descent logic in scripts that set up raspberries or walls of particles. Since fixed particles are skipped by the integrator, their forces won't change, and their forces are typically far larger than any other force in the system, so the derivative of the maximal force is 0. Here is a MWE with a raspberry:
```python
import espressomd
import numpy as np

system = espressomd.System(box_l=[40.] * 3)
system.time_step = 0.01
system.cell_system.skin = 0.3

system.non_bonded_inter[1, 0].wca.set_params(epsilon=48, sigma=3)
system.non_bonded_inter[1, 1].wca.set_params(epsilon=1, sigma=1)
harmonic_bond = espressomd.interactions.HarmonicBond(k=3000., r_0=3)
system.bonded_inter.add(harmonic_bond)

np.random.seed(0)
center = system.box_l / 2 
system.part.add(id=0, pos=center, type=0, q=-40, fix=3*[1], rotation=3*[1])
for i in range(1, 114):
    colSurfPos = np.random.randn(3)
    colSurfPos = colSurfPos / np.linalg.norm(colSurfPos) * 3 + center
    system.part.add(id=i, pos=colSurfPos, type=1)
    system.part[i].add_bond((harmonic_bond, 0))

F_MAX = 500
system.integrator.set_steepest_descent(f_max=F_MAX, gamma=1.0, max_displacement=0.01)
n = system.integrator.run(10000)
print(f'convergence reached after {n} steps')
print(f'maximal force is {np.max(np.linalg.norm(system.part[:].f, axis=1)):.1f}')
print(f'maximal force excluding fixed particle is {np.max(np.linalg.norm(system.part[1:].f, axis=1)):.1f} < {F_MAX}')
```
output:
```
convergence reached after 46 steps
maximal force is 1989.8
maximal force excluding fixed particle is 464.9 < 500
```

On a side-note, the `F_MAX` variable name looks confusing to me, as it suggests a maximal force when in reality you're using the integrator magic value `0`. I originally thought you had three convergence criteria: maximal force, derivative of the maximal force, derivative of the energy. Maybe `F_MAX_DISABLED = 0` would be more expressive, or using `f_max=0` directly in the parameter list?

schlaicha: Sure, I agree that there is *no* generic method to equilibrate your system. However, this statement also applies to the force capping and the steepest descent with a maximal force. I'll add a note for the tutors and maybe one sentence in the notebook.

> the criterion for early exit is based on the maximal force on all particles, whereas the integrator actually works on a subset of all particles

Actually that would be great from a didactic point to mention and maybe include in one of the later tutorials. Is there an easy way to know the subset of particles that the integrator actually uses?

> the convergence criterion is based on forces measured by their components while the integrator uses their norm

Agreed, let's change this in the tutorials.

> using f_max=0 directly in the parameter list

Sounds good to me!

jngrad: Yes, there is no generic method, and force capping is also concerned... I just felt like we should mention it once, otherwise people might copy-paste that block into their simulations. Optimization strategies can be difficult to construct because they depend a lot on the properties of the system. 

\<digression\>While researching optimization strategies and their pitfalls a few years ago, I found a paper that uses an optimization algorithm to find the optimal parameters of another optimization algorithm ([doi:10.1016/0377-2217(94)00350-5](https://doi.org/10.1016/0377-2217(94)00350-5)). It ended up on a slide in one of my group seminars :smiley: \</digression\>

> Actually that would be great from a didactic point to mention and maybe include in one of the later tutorials. Is there an easy way to know the subset of particles that the integrator actually uses?

When particles are fixed in one or more dimensions, the components of the force along the fixed axes are not propagated by integrators. It's not specific to steepest descent. You would have to use `system.part[:].fix` as a mask, e.g. `np.max(np.linalg.norm(system.part[:].f * np.logical_not(system.part[:].fix), axis=1))`, to get the forces propagated by the integrator.

There are also pitfalls with virtual sites, which transfer their force to the tracked particles but only get synchronized at the beginning of each integration step, which causes them to be desynchronized after the last integration step:
```python
import espressomd
import numpy as np
system = espressomd.System(box_l=[40.] * 3)
system.time_step = 0.01
system.cell_system.skin = 0.3
system.non_bonded_inter[1, 1].wca.set_params(epsilon=1, sigma=1)
center = system.box_l / 2 
p1 = system.part.add(id=0, pos=[0, 0, 0], type=1)
p2 = system.part.add(id=1, pos=[0, 0, 0.1], type=1)
p2.vs_auto_relate_to(p1.id)
system.integrator.set_steepest_descent(f_max=800, gamma=1.0, max_displacement=0.01)
n = system.integrator.run(100)
print(f'convergence reached after {n} steps')
print(f'maximal force is {np.max(np.linalg.norm(system.part[:].f, axis=1)):.1f}')
print('system.integrator.run(0)')
system.integrator.run(0)
print(f'maximal force is {np.max(np.linalg.norm(system.part[:].f, axis=1)):.1f}')
```
output:
```
convergence reached after 70 steps
maximal force is 758.7
system.integrator.run(0)
maximal force is 638.0
```
The first value is the one that was propagated by the integrator. The second value represents the true force with the current system state.

schlaicha: Should be resolved with 
0ed1854 and 
5104cdf, also see #3891.

schlaicha:
@jngrad The comparison of the RDF that you included in the end in 
cc1e27b (and which for now I put into the solution) looks really bad for me even for 10x longer simulations. Is this really what we want to show?

schlaicha:
Still to do:

>  The MSD correlator explanation is no longer accurate. The first two columns don't store the lag time and sample size anymore, instead one has to call msd_corr.lag_times() explicitly to get the lag times. This regression was introduced in a previous PR.
    The introduction paragraph needs some cleanup. The core is no longer written in the C language.

testsuite/scripts/tutorials/test_01-lennard_jones.py
```diff
@@ -28,9 +28,13 @@
 class Tutorial(ut.TestCase):
     system = tutorial.system
 
-    def test_rdf_curve(self):
-        self.assertGreater(
-            np.corrcoef(tutorial.rdf, tutorial.theo_rdf)[1, 0], 0.985)
+    def test_rdf(self):
+        rms = np.sqrt(((tutorial.rdf - tutorial.theo_rdf)**2).mean()) / len(tutorial.r)
```
jngrad: * why divide by the length of the data series outside the square root?
* would it be more helpful to compare the data series directly? e.g. with `np.testing.assert_allclose(tutorial.rdf, tutorial.theo_rdf, rtol=0, atol=5e-4)`, in case of failure the test would format the two data series and give you a summary of the statistics (number of mismatched data points, max relative difference, max absolute difference). This would however make the test a lot more sensitive to outliers, whereas the RMS averages out artifacts.

testsuite/scripts/tutorials/test_01-lennard_jones.py
```diff
@@ -28,9 +28,13 @@
...
+        self.assertLess(rms, 5e-4)
+
+    def test_potential_energy(self):
+        # Test that the potential energy/particle agrees with the value from Verlet, Phys. Rev. 1967
+        self.assertLess(np.abs(tutorial.sim_energy + 5.38)/5.38,0.01)
```
jngrad: If I'm reading this correctly, we test `|x + 5.38| < 0.0538`, solving for `x` we get `-5.4338 < x < -5.3262`. This could be expressed as
```python
ref_energy = -5.38
self.assertAlmostEqual(tutorial.sim_energy, ref_energy, delta=ref_energy / 100)
```
or as
```python
self.assertAlmostEqual(tutorial.sim_energy, -5.38, 1)
```
because `place=1` is 0.97% if I'm not mistaken.

testsuite/scripts/tutorials/test_01-lennard_jones.py
schlaicha: The idea was that the rms obviously depends on the number of data points, to get an independent measure one could calculate the RMS per data point. However, I agree usnig `np.testing.assert_allclose(tutorial.rdf, tutorial.theo_rdf, rtol=0, atol=5e-4)` is much more elegant.

doc/tutorials/01-lennard_jones/01-lennard_jones.ipynb
```diff
@@ -393,98 +441,103 @@
...
+    "* Use <tt>[espressomd.integrate](http://espressomd.org/html/doc/espressomd.html#module-espressomd.minimize_energy).set_steepest_descent</tt> to relax the initial configuration.\n",
+    "  Use a maximal displacement to <tt>MAX_DISPLACEMENT</tt>.\n",
+    "  A damping constant <tt>gamma = DAMPING</tt> usually is a good choice.\n",
+    "* Use the relative change of the system total energy and the maximal force as a convergence criterion.\n",
+    "  See the [<tt>espressomd.analyze</tt>](http://espressomd.org/html/doc/espressomd.html#module-espressomd.analyze) documentation on how to obtain energies and the [<tt>espressomd.particle_data</tt> module](http://espressomd.org/html/doc/espressomd.html#module-espressomd.particle_data) for the forces.\n",
+    "  The steepest descent has converged if the relative force change < <tt>F_TOL</tt>\n",
+    "* Break the minimization loop after a maximal number of <tt>MAX_STEPS</tt> steps or if convergence is achieved.\n",
+    "  Check for convergence every <tt>EMSTEP</tt> steps.\n",
+    "  \n",
+    "***Hint:*** To obtain the initial fores one has to initialize integrator using <tt>integ_steps=0</tt>."
```
RudolfWeeber: I think `system.integrator.run(0)` has to be mentioned explicitly. I wouldn't have understood that otherwise, I think. 

Could you please make an attempt to simplify the loop? A single criterion such as absolute force should be enough, IMO.
There only is a point in counting the number of times the loop is executed, if the non-convergence case is actually treated.

doc/tutorials/01-lennard_jones/01-lennard_jones.ipynb
```diff
@@ -393,98 +441,103 @@
...
+    "\n",
+    "i = 0\n",
+    "while i < MAX_STEPS//EMSTEP:\n",
+    "    prev_maxforce = maxforce\n",
+    "    prev_energy = energy\n",
+    "    system.integrator.run(EMSTEP)\n",
+    "    maxforce = np.max(np.linalg.norm(system.part[:].f, axis=1))\n",
+    "    relforce = np.abs((maxforce-prev_maxforce)/prev_maxforce)\n",
+    "    energy = system.analysis.energy()['total']\n",
+    "    relener = np.abs((energy-prev_energy)/prev_energy)\n",
```
RudolfWeeber: please adhere to pep8 naming convention. I.e., underscores to separate wores

doc/tutorials/01-lennard_jones/01-lennard_jones.ipynb
```diff
@@ -393,98 +441,103 @@
...
+    "\n",
+    "Simulations can be carried out in different thermodynamic ensembles such as NVE (particle __N__umber, __V__olume, __E__nergy), NVT (particle __N__umber, __V__olume, __T__emperature) or NPT-isotropic (particle __N__umber, __P__ressure, __T__emperature).\n",
+    "\n",
+    "The NVE ensemble is simulated without a thermostat. A previously enabled thermostat can be switched off as follows:\n",
+    "\n",
+    "``` python\n",
+    "system.thermostat.turn_off()\n",
+    "```\n",
+    "\n",
+    "The NVT and NPT ensembles require a thermostat. In this tutorial, we use the Langevin thermostat.\n",
```
RudolfWeeber: NpT (small letters for intensive variables)

doc/tutorials/01-lennard_jones/01-lennard_jones.ipynb
```diff
@@ -493,180 +546,374 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "code_folding": [],
+    "solution2": "hidden"
+   },
+   "source": [
+    "```python\n",
+    "etotal = np.zeros(SAMPLING_ITERATIONS)\n",
```
RudolfWeeber: pep8 naming: e_total

doc/tutorials/01-lennard_jones/01-lennard_jones.ipynb
```diff
@@ -493,180 +546,374 @@
...
+    "ekin = np.zeros(SAMPLING_ITERATIONS)\n",
+    "time = np.zeros(SAMPLING_ITERATIONS)\n",
     "\n",
-    "\\begin{equation}\n",
-    "    SE      = \\sqrt{\\frac{\\sigma^2}{N}},\n",
-    "\\end{equation}\n",
+    "from espressomd.observables import ParticlePositions, RDF\n",
```
RudolfWeeber: Is ParticlePositions still needed with the msd removed?

doc/tutorials/01-lennard_jones/01-lennard_jones.ipynb
```diff
@@ -206,56 +246,64 @@
...
+    "solution2": "hidden",
+    "solution2_first": true
+   },
    "source": [
-    "system.thermostat.turn_off()"
+    "<!-- **Exercise:** -->\n",
+    "\n",
+    "* Create <tt>N_PART</tt> particles at random positions.\n",
+    "\n",
+    "  Use [system.part.add()](link).\n",
```
RudolfWeeber: pls fill in link

RudolfWeeber (changes requested):
For the most part, minor fixes. If the overlap removal would become somewhat simpler, that would be great for a beginner's tutorial.

doc/tutorials/01-lennard_jones/01-lennard_jones.ipynb
schlaicha: I'll implement the requested changes suggested below, only two comments on the minimization loop:
* I agree that this reads complicated. But the problem is that in steepest descent we minimize energies whereas the MD relies on forces only, so for me the general question is how to define a reasonable convergence criterion? I agree that this reads slightly complicated but only on the python side (unrelated to ESPResSo). At the same time this follows the practice of other MD packages (Gromacs, LAMMPS, ...). If there is any favour on which criterion is better in the ESPResSo community I'm happy to change accordingly.
* the non-convergence case should actually be treated.
I strongly believe that the adapted steepest descent with above convergence criteria is something we should offer to the user in the mid-term, which would also greatly simplify the minimization in the code (see discussion in #3898).

RudolfWeeber:
>I agree that this reads complicated. But the problem is that in steepest descent we minimize energies whereas the MD relies on forces only, 

I’m confused. The steepest descent code in espresso (to my knowledge) is force based. It moves particles by a certain amount along the force vectors acting on them.

It terminates, when the largest force observed dudring the last step is less than f_max.

See https://github.com/espressomd/espresso/blob/python/src/core/integrators/steepest_descent.cpp

 

Isn’t that exactly what’s needed?

 

Maybe I’m missing something and we should discuss it off-line

schlaicha:
> I’m confused. The steepest descent code in espresso (to my knowledge) is force based. It moves particles by a certain amount along the force vectors acting on them.

I'm sorry, of course you're right. On a Sunday evening I missed that grad x E is equal to the force. And indeed my argumentation is exactly opposite for MD/MC.  

> Isn’t that exactly what’s needed?

Yes, only that in general I would not know what value fmax is good enough for a specific system / timestep. My suggestion was to have a relative convergence criterion directly in (which would simplify the typical simulation script/tutorial). But maybe that is also not needed.

> Could you please make an attempt to simplify the loop? A single criterion such as absolute force should be enough, IMO.

So removing the energy convergence criterion is simple enough? I would not see how to make it much simpler otherwise...

RudolfWeeber:
Yes, a force only criterion is what I’d use.

 

The totally clean way to do it would be comparing acceleration `F/m` to `1/dt`, That is, what determines integrator stability, independent of the system.

But let’s sort that out an other time.

doc/tutorials/01-lennard_jones/01-lennard_jones.ipynb
schlaicha: The non-convergence case **is** actually treated and I added more explanations to the `system.integrator.run(0)` command.

RudolfWeeber (dismissed):

doc/tutorials/01-lennard_jones/NotesForTutor.md
```diff
@@ -0,0 +1,44 @@
...
+* Non-bonded interactions are defined between pairs of particle types.
+  Mention that type=0 is implied when creating particles without specifying one.
+* Explain Lennard-Jones cutoff and shift - we don't shift here to make energies comparable to historic simulations (Verlet at al.).
+* Explain reduced units - here: LJ units!
+* Explain steepest descent: In general overlap removal can
+  be done in multiple ways (warmup with force capping, 
+  small time step, etc.).
+  Additionally the steepest descent algorithm can get trapped in local minima and the convergence criterion is system-dependent.
+* Mention that accumulators are updated auomatically.
+* Statistical independent measurements for t > correlation time.
```
jngrad:
```suggestion
* Statistically independent measurements for t > correlation time.
```

doc/tutorials/01-lennard_jones/NotesForTutor.md
```diff
@@ -0,0 +1,44 @@
...
+      * LJ agrees well with experiments for noble gases!
+    * how to use an auto correlation function to estimate correlation times and how that affects error estimation
+
+
+## Learning objectives (ESPResSo)
+
+After the tutorial, students should be able to 
+
+* instance a simulation object and set basic parameters like time step
+* add particles to a simulation,
```
jngrad:
```suggestion
* add particles to a simulation
```

doc/tutorials/01-lennard_jones/NotesForTutor.md
```diff
@@ -0,0 +1,44 @@
...
+      * relation to structure factor
+      * LJ agrees well with experiments for noble gases!
+    * how to use an auto correlation function to estimate correlation times and how that affects error estimation
+
+
+## Learning objectives (ESPResSo)
+
+After the tutorial, students should be able to 
+
+* instance a simulation object and set basic parameters like time step
```
jngrad:
```suggestion
* instance a system object and set basic parameters like the time step
```

doc/tutorials/01-lennard_jones/NotesForTutor.md
```diff
@@ -0,0 +1,44 @@
+# Hints for Tutors: Tutorial 01 - Lennard Jones
+
+## Learning objectives (physics):
```
jngrad:
```suggestion
## Learning objectives (physics)
```

doc/tutorials/01-lennard_jones/01-lennard_jones.ipynb
```diff
@@ -393,98 +439,98 @@
...
-   "metadata": {},
+   "metadata": {
+    "solution2": "hidden",
+    "solution2_first": true
+   },
    "source": [
-    "We now use the plotting library <tt>matplotlib</tt> available in Python to visualize the measurements."
+    "**Exercise:**\n",
+    "\n",
+    "* Use <tt>[espressomd.integrate](http://espressomd.org/html/doc/espressomd.html#module-espressomd.minimize_energy).set_steepest_descent</tt> to relax the initial configuration.\n",
```
jngrad:
```suggestion
    "* Use [<tt>espressomd.integrate.set_steepest_descent</tt>](http://espressomd.org/html/doc/running.html#steepest-descent) to relax the initial configuration.\n",
```
broken link (`minimize_energy` was removed), not caught in CI because the link was invalid Markdown syntax

doc/tutorials/01-lennard_jones/01-lennard_jones.ipynb
```diff
@@ -336,55 +381,56 @@
...
-   "metadata": {},
+   "metadata": {
+    "solution2": "hidden",
+    "solution2_first": true
+   },
    "source": [
-    "### Integrating equations of motion and taking measurements\n",
+    "<!-- **Exercise:** -->\n",
+    "\n",
+    "* Setup a Lennard-Jones interaction with $\\epsilon=$<textt>LJ_EPS</textt> and $\\sigma=$<textt>LJ_SIG</textt> which are cut and shifted to zero at $r_\\mathrm{c}=$<textt>LJ_CUT</textt>$\\times\\sigma$.\n",
```
jngrad:
```suggestion
    "* Setup a Lennard-Jones interaction with $\\epsilon=$<tt>LJ_EPS</tt> and $\\sigma=$<tt>LJ_SIG</tt> which are cut and shifted to zero at $r_\\mathrm{c}=$<tt>LJ_CUT</tt>$\\times\\sigma$.\n",
```
invalid Markdown syntax

doc/tutorials/01-lennard_jones/01-lennard_jones.ipynb
```diff
@@ -493,180 +539,374 @@
...
-    "### Simple Error Estimation on Time Series Data\n",
-    "\n",
-    "A simple way to estimate the error of an observable is to use the standard error of the mean (SE) for $N$\n",
-    "_uncorrelated_ samples:\n",
+    "**Exercise:**\n",
+    "\n",
+    "**Part I:**\n",
+    "* Set up the main integration loop.\n",
+    "  Take <tt>SAMPLING_ITERATIONS</tt> measurements every <tt>SAMPLING_INTERVAL</tt> integration steps.\n",
+    "* Monitor the potential and kinetic energies using the analysis method <tt>[system.analysis](http://espressomd.org/html/doc/espressomd.html#module-espressomd.analyze).energy()</tt>.\n",
```
jngrad:
```suggestion
    "* Monitor the potential and kinetic energies using the analysis method [<tt>system.analysis.energy()</tt>](http://espressomd.org/html/doc/espressomd.html#module-espressomd.analyze).\n",
```

doc/tutorials/01-lennard_jones/01-lennard_jones.ipynb
```diff
@@ -493,180 +539,374 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Binary Lennard-Jones Liquid\n",
+    "\n",
+    "A two-component Lennard-Jones liquid can be simulated by placing particles of two types (0 and 1) into the system. Depending on the Lennard-Jones parameters, the two components either mix or separate.\n",
+    "\n",
+    "1. Modify the code such that half of the particles are of <tt>type=1</tt>. Type 0 is implied for the remaining particles.\n",
+    "2. Specify Lennard-Jones interactions between type 0 particles with other type 0 particles, type 1 particles with other type 1 particles, and type 0 particles with type 1 particles (set parameters for <tt>system.non_bonded_inter[i,j].lennard_jones</tt> where <tt>{i,j}</tt> can be <tt>{0,0}</tt>, <tt>{1,1}</tt>, and <tt>{0,1}</tt>. Use the same Lennard-Jones parameters for interactions within a component, but use a different <tt>lj_cut_mixed</tt> parameter for the cutoff of the Lennard-Jones interaction between particles of type 0 and particles of type 1. Set this parameter to $2^{\\frac16}\\sigma$ to get de-mixing or to $2.5\\sigma$ to get mixing between the two components.\n",
```
jngrad:
```suggestion
    "2. Specify Lennard-Jones interactions between type 0 particles with other type 0 particles, type 1 particles with other type 1 particles, and type 0 particles with type 1 particles (set parameters for <tt>system.non_bonded_inter[i,j].lennard_jones</tt> where <tt>{i,j}</tt> can be <tt>{0,0}</tt>, <tt>{1,1}</tt>, and <tt>{0,1}</tt>. Use the same Lennard-Jones parameters for interactions within a component, but use a different <tt>lj_cut_mixed</tt> parameter for the cutoff of the Lennard-Jones interaction between particles of type 0 and particles of type 1. Set this parameter to $2^{\\frac{1}{6}}\\sigma$ to get de-mixing or to $2.5\\sigma$ to get mixing between the two components.\n",
```

jngrad (changes requested):

doc/tutorials/01-lennard_jones/01-lennard_jones.ipynb
jngrad: no color highlighting because line is too long: the frac16 at the end was missing braces

doc/tutorials/01-lennard_jones/01-lennard_jones.ipynb
```diff
@@ -493,180 +539,373 @@
...
    "cell_type": "markdown",
-   "metadata": {},
+   "metadata": {
+    "solution2": "hidden"
+   },
    "source": [
-    "## Comparison with the literature\n",
-    "\n",
-    "Empirical radial distribution functions have been determined for pure fluids <a href='#[5]'>[5]</a>, mixtures <a href='#[6]'>[6]</a> and confined fluids <a href='#[7]'>[7]</a>. We will compare our distribution $g(r)$ to the theoretical distribution $g(r^*, \\rho^*, T^*)$ of a pure fluid <a href='#[5]'>[5]</a>."
+    "From the roughly exponential decay of the correlation function it dawns that only every second sample should be used to calculate statistically independent averages."
```
jngrad, line 785: My lecture on error estimation advises against this practice because it can lead to an overestimation of the standard error of the mean. I think it is better practice to carry out binning analysis, or better, autocorrelation function integration.

jngrad (approved):

PR #3884, opened 2020-09-03
Use global counter in thermostats

### LINKED ISSUE #3573: Unit tests thermostats_test and random_test are very slow
in debug builds. This is due to the use of heap allocations and std::functions in the how loop.
With a better implementation these tests can be considerably sped up, which would lead to
a smoother workflow.
We should also discuss if statistical tests on the thermostats are needed or a good idea, or
if would be better if a mock RNG could be passed to then, and then checked if it is used correctly.

Also I'm not convinced that we should test the statistics of the RNG, I think it should rather be
tested that it implements philox correctly, e.g. checking that it produces the correct sequence
for a small set of parameters.

For reference, on my AMD Ryzen Threadripper 1920X, the tests take:
```
thermostats_test .................   Passed  191.93 sec
random_test ......................   Passed  147.47 sec
```

### LINKED ISSUE #3585: Thermostat noise for different seeds is correlated
Currently the thermostat noise for different seeds is correlated, the seed just shifts the sequence.
Currently the seed is used to offset the counter, but it should be used in the key to get independent
sequences. Also there should only be one counter...

### LINKED ISSUE #3840: Noise behavior of stokesian dynamics and other thermostats differ
SD seems to use a counter based noise source, as do the other thermostats. The counter value is derived from the simulation
time by calculating `sim_time / time_step`, which means that if the users resets the time, the noise is repeated. This differs from
the other thermostats, where the noise is derive from a monotonous counter which can not be reset. Also the SD version leads
to correlated noise if the time step is changed.

jngrad:
Description of changes:
- Remove RNG correlation stemming from seed offsets (fixes #3585)
    - seeds are now used as keys
    - a global, monotonically increasing counter is used in the thermostats
    - the only way to reset this counter is to create a new `System`
- Remove RNG correlation stemming from resetting `sim_time` or `time_step` during simulations with SD (fixes #3840)
    - the SD thermostat now uses the same RNG interface as other thermostats
- Accelerate RNG unit tests (fixes #3573)
    - they now take 2 sec to run in coverage and sanitizer builds in CI
- Split thermostats from integrators
    - better separation of concerns
    - globals from `integrate.hpp` are passed to thermostats as function parameters


KaiSzuttor:
wouldn't it be a good idea to have a global that monotonically increases with time. It may be useful also outside of the thermostat context at some point?

jngrad:
> wouldn't it be a good idea to have a global that monotonically increases with time. It may be useful also outside of the thermostat context at some point?

Yes, this is how the global counter is implemented. But we cannot rewind it by resetting the system time. It could be made a property of the integrator, or a property of the system to be able to re-use it elsewhere.

jngrad:
Using the same global counter in all thermostats solves the correlation issue but creates new problems. Some thermostats have specific needs: DPD needs extra incrementation when a DPD particle interacts with a constraint, and LB is not incremented when the integrator is steepest descent. Using a global counter incremented by the integrator means `thermostat.cpp` must include `integrator.hpp` and thermostat users such as `src/core/constraints/ShapeBasedConstraint.cpp` must include both `integrator.hpp` and `thermostat.hpp`. Having one counter per thermostat removes these issues.

jngrad:
New attempt without the global counter in #3888.

KaiSzuttor:
IIRC, the idea of the counter based RNG is that by knowing the value of the counter and two particle ids we determine one random number (in fact we could get 4). If we need more than one random number per particle pair and integration, the whole logic brakes down, right?

KaiSzuttor:
> LB is not incremented when the integrator is steepest descent

what do you mean by that?

KaiSzuttor:
> DPD needs extra incrementation when a DPD particle interacts with a constraint

why?

KaiSzuttor:
> Using a global counter incremented by the integrator means thermostat.cpp must include integrator.hpp

I think the counter state should be passed as an argument to all methods that need to know about it

src/core/constraints/ShapeBasedConstraint.cpp
```diff
@@ -74,10 +74,8 @@ ParticleForce ShapeBasedConstraint::force(Particle const &p,
                                           dist, &torque1, &torque2);
 #ifdef DPD
       if (thermo_switch & THERMO_DPD) {
-        force1 +=
-            dpd_pair_force(p, part_rep, ia_params, dist_vec, dist, dist * dist);
-        // Additional use of DPD here requires counter increase
-        dpd_rng_counter_increment();
```
KaiSzuttor, line 80: this counter has never been incremented twice... the control flow is in this if branch *or* in the one below

src/core/integrate.hpp
```diff
@@ -144,3 +151,9 @@ int integrate_set_npt_isotropic(double ext_pressure, double piston,
                                 bool zdir_rescale, bool cubic_box);
 
 #endif
+
+extern Utils::Counter<uint64_t> integrator_counter;
```
KaiSzuttor, line 155: those declarations should come before the ```#endif``` of the header guard

jngrad:
> IIRC, the idea of the counter based RNG is that by knowing the value of the counter and two particle ids we determine one random number (in fact we could get 4). If we need more than one random number per particle pair and integration, the whole logic brakes down, right?

True for consumers of the thermostats. However, the thermostats functions can generate more than 1 random number per tuple `(counter, seed, pid1, pid2)` using the salt, because they are the one to call the RNG functions. This is used e.g. in NpT with `NPTISO0_HALF_STEP1` and  `NPTISO0_HALF_STEP2`. This cannot be used in DPD because the salt has to be known at compile time, while the DPD thermostat needs to generate an arbitrary number of random numbers per unique tuple, depending on the number of shape-based constraints in the system.

On a side-note, adding a new salt means shifting all the following salts in the `RNGSalt` enum, which is a silent change: scripts from one espresso minor version will generate different random numbers with the next espresso minor version where salts are shifted.

---

> > LB is not incremented when the integrator is steepest descent
> 
> what do you mean by that?

The function that increments the LB counters (CPU and GPU) is not called when using steepest descent:
https://github.com/espressomd/espresso/blob/2fbea6e9671f6e18b092d8ec67f4f8b4776ed887/src/core/integrate.cpp#L267-L268
With a global counter, the LB RNG sequence will be incremented anyway, just like the other thermostats. Looking again at the code, this should not pose an issue.

---

> > DPD needs extra incrementation when a DPD particle interacts with a constraint
> 
> why?

Looking at the code, I would say it prevents correlation in the noise when you have more than one shape-based constraint. The shape-based constraint has a particle representation `Particle ShapeBasedConstraint::part_rep` which is default-constructed, i.e. its particle id is always `-1`, presumably to ensure real particles are numbered continuously.

This makes DPD quite different, because its counter gets out of sync with the number of integration steps, and gets incremented by `system.integrator.run(0)` while other thermostats don't. This is also why we have this logic at line 191 of the checkpoint test:

https://github.com/espressomd/espresso/blob/2fbea6e9671f6e18b092d8ec67f4f8b4776ed887/testsuite/python/test_checkpoint.py#L187-L191

---

> > Using a global counter incremented by the integrator means thermostat.cpp must include integrator.hpp
> 
> I think the counter state should be passed as an argument to all methods that need to know about it

I tried in 486612d83f but this means adding an extra argument to all classes in the constraint infrastructure just for DPD. Or adding a getter function to get the value of the global counter variable (to avoid exposing it via `extern`), in which case you need to include `integrate.hpp`.

KaiSzuttor:
> This cannot be used in DPD because the salt has to be known at compile time, while the DPD thermostat needs to generate an arbitrary number of random numbers per unique tuple, depending on the number of shape-based constraints in the system.

here the id of the particle representation of the constraint has to be used

KaiSzuttor:
> Looking at the code, I would say it prevents correlation in the noise when you have more than one shape-based constraint. The shape-based constraint has a particle representation Particle ShapeBasedConstraint::part_rep which is default-constructed, i.e. its particle id is always -1, presumably to ensure real particles are numbered continuously.

Then this needs to change.

KaiSzuttor:
> This cannot be used in DPD because the salt has to be known at compile time, while the DPD thermostat needs to generate an arbitrary number of random numbers per unique tuple, depending on the number of shape-based constraints in the system.

This would also not be the right approach. Actually, the interface for the random numbers should be such that it cannot be used the wrong way... Then we would see that we neither have the correct RNG interface for the NPT case nor have we solved the special case of constraint interaction via DPD (which should be handled just like any other particle particle interaction). The real issue here is that the particle can be default initialized which leads to the issue that you don't know which information of the particle representation in the constraint can actually be used and which not.

KaiSzuttor:
> With a global counter, the LB RNG sequence will be incremented anyway, just like the other thermostats. Looking again at the code, this should not pose an issue.

isn't the correct behavior to not propagate the system's state at all in case of steepest descent since there is no notion of time here? not sure about that

KaiSzuttor:
> I tried in 486612d but this means adding an extra argument to all classes in the constraint infrastructure just for DPD. Or adding a getter function to get the value of the global counter variable (to avoid exposing it via extern), in which case you need to include integrate.hpp.

What's wrong which the additional argument? Following your reasoning we should have more global variables.

KaiSzuttor:
> On a side-note, adding a new salt means shifting all the following salts in the RNGSalt enum, which is a silent change: scripts from one espresso minor version will generate different random numbers with the next espresso minor version where salts are shifted.

Not an issue if you add the salt at the end?

KaiSzuttor:
offline discussion: the issue with the DPD interaction on ShapeBasedConstraint has it's origin in the fact that the particle representation in the Constraint is not a real particle and thus has not a meaningful particle id (which is in fact a part of the counter based RNG we are using). One possible solution could be to actually create the particle and only store the pointer to the particle in the constraint. The interaction of this solution with other parts of the code, however, is not clear.

src/core/random.hpp
```diff
@@ -97,24 +98,28 @@ auto philox_4_uint64s(uint64_t counter, int key1, int key2 = 0) {
...
+ * @param seed seed for random number generation
+ * @param key1 key for random number generation
+ * @param key2 key for random number generation
  *
  * @return Vector of uniform random numbers.
  */
 template <RNGSalt salt, size_t N = 3,
           std::enable_if_t<(N > 1) and (N <= 4), int> = 0>
-auto noise_uniform(uint64_t counter, int key1, int key2 = 0) {
+auto noise_uniform(uint64_t counter, uint32_t seed, int key1, int key2 = 0) {
```
KaiSzuttor, line 110: i think what is missing is an abstraction for retrieving the random numbers specifically for the case of particle ids

jngrad:
The DPD interaction currently prevents the use of a single counter, we'll need to find a long-term solution based on e.g. uuids (see #3892). Until then, the correlation bugs can be fixed by #3888, where thermostats still have their own counter.

PR #3886, opened 2020-09-07
WIP: tutorial 2 - charged systems

christophlohrmann:
Description of changes:
- New tutorial for charged rod counterion condensation for the Summer School


doc/tutorials/13-charged_rod/13-charged_rod.ipynb
```diff
@@ -0,0 +1,270 @@
+{
```
KaiSzuttor: <p>... of the theory from the tutorial on charged systems [... link to tutorial ... ]</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/espressomd/espresso/pull/3886/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

KaiSzuttor: <p>i'm not sure if we want to keep to the PEP8 convention, but if so, all those variables need to be CAPITALIZED</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/espressomd/espresso/pull/3886/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

KaiSzuttor: <p>run run</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/espressomd/espresso/pull/3886/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

KaiSzuttor: <p>until the system is equilibrated</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/espressomd/espresso/pull/3886/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

doc/tutorials/13-charged_rod/13-charged_rod.ipynb
```diff
@@ -0,0 +1,270 @@
...
+   "source": [
+    "force_cap=1\n",
+    "system.force_cap = force_cap\n",
+    "\n",
+    "for i in range(4000):\n",
+    "    system.integrator.run(200)\n",
+    "    dist=system.analysis.min_dist(p1=[ion_type],p2=[ion_type, rod_type])\n",
+    "    print( f\"Warm up step {i}: min_dist={dist} force_cap={force_cap}\")\n",
+    "    force_cap = force_cap + 1.\n",
+    "    system.force_cap = force_cap\n",
```
jngrad: is this manual warmup necessary? tutorial 02 part 1 shows how to use the steepest descent algorithm to equilibrate the system

schlaicha: We should go for best practice in all tutorials and have a steepest descent with energy/force convergence criterion, see #3881

jngrad: just realized tutorial 02 part 1 uses the `minimize_energy` module, which is deprecated...

christophlohrmann: according to @schlaicha this tutorial 13 will replace tutorial 02 part 1

doc/tutorials/13-charged_rod/13-charged_rod.ipynb
```diff
@@ -0,0 +1,536 @@
...
+    "def remove_overlap(system, sd_params):   \n",
+    "    #remove overlap by steepest descent\n",
+    "    # Set up steepest descent integration\n",
+    "    system.integrator.set_steepest_descent(f_max=sd_params['f_max'],\n",
+    "                                           gamma=sd_params['damping'],\n",
+    "                                           max_displacement=sd_params['max_displacement'])\n",
+    "    \n",
+    "    # Initialize integrator to obtain initial forces\n",
+    "    system.integrator.run(0)\n",
+    "    maxforce = np.max(np.linalg.norm(system.part[:].f))\n",
```
jngrad: Here and in the loop below, `maxforce` contains the square root of the sum of squared forces. Did you mean to calculate the maximal force?
```suggestion
    "    maxforce = np.max(np.linalg.norm(system.part[:].f, axis=1))\n",
```

doc/tutorials/13-charged_rod/13-charged_rod.ipynb
```diff
@@ -0,0 +1,536 @@
...
+    "    energy = system.analysis.energy()['total']\n",
+    "    \n",
+    "    i = 0\n",
+    "    while i < sd_params['max_steps']//sd_params['emstep']:\n",
+    "        prev_maxforce = maxforce\n",
+    "        prev_energy = energy\n",
+    "        print(prev_energy)\n",
+    "        system.integrator.run(sd_params['emstep'])\n",
+    "        maxforce = np.max(np.linalg.norm(system.part[:].f))\n",
```
jngrad:
```suggestion
    "        maxforce = np.max(np.linalg.norm(system.part[:].f, axis=1))\n",
```

RudolfWeeber:
What is still missing, here?

jngrad:
@RudolfWeeber I think there was an issue with the P3M tuning algorithm in non-cubic boxes under specific conditions.

christophlohrmann:
Part 1 is ready for review I would say. I will add the test and tutor's notes today. Part 2 has still a lot of issues and will deal with a different system than in the current state (dilute salt between capacitors rather than ionic liquid)

doc/tutorials/02-charged_system/02-charged_system-1.ipynb
```diff
@@ -39,57 +58,162 @@
...
-    "n_part = 200\n",
-    "n_ionpairs = n_part / 2\n",
-    "density = 0.5\n",
-    "time_step = 0.01\n",
-    "temp = 1.0\n",
-    "gamma = 1.0\n",
-    "l_bjerrum = 7.0\n",
+    "**Hints:**\n",
+    "* The WCA potential uses the same parameters as the Lennard-Jones potential, but the cutoff and shift are calculated automatically\n",
+    "* Use the Lorentz-Berthelot combining rule to determine the ``sigma`` parameter of the interaction between the rod particles and the ions"
```
RudolfWeeber: Pls mention the expressions for the combination rule

doc/tutorials/02-charged_system/02-charged_system-1.ipynb
```diff
@@ -39,57 +58,162 @@
...
+    "\n",
+    "    # calculate charge of the single rod beads\n",
+    "    rod_length = system.box_l[2]\n",
+    "    total_rod_charge = rod_charge_dens*rod_length\n",
+    "    rod_charge_per_bead = total_rod_charge/N_rod_beads\n",
+    "\n",
+    "    # number of counterions\n",
+    "    N_ions = int(total_rod_charge/ion_valency)\n",
+    "\n",
+    "    for idx in range(N_rod_beads):\n",
```
RudolfWeeber: Ideally, pls avoid the loops:
```
rod_positions =np.arange(...) * np.array((1,0,0))
rod_particles = system.part.add(pos=rod_positions,...)

counter_ions = system.part.add(pos=np.random.random(...))
```
Of course, keeping the handles only makes sense, if they are later needed to refer to the groups of particles.

doc/tutorials/02-charged_system/02-charged_system-1.ipynb
```diff
@@ -98,21 +222,53 @@
...
    "outputs": [],
    "source": [
-    "# Setup System\n",
-    "box_l = (n_part / density)**(1. / 3.)\n",
-    "system = System(box_l=[box_l, box_l, box_l])\n",
-    "system.periodicity = [True, True, True]\n",
-    "system.time_step = time_step\n",
-    "system.cell_system.skin = 0.3"
+    "p3m_params = {'prefactor':kT*bjerrum_length,\n",
+    "              'accuracy':1e-3}\n",
```
RudolfWeeber: Maybe a word on how to pick the accuracy? What would be reasonable in production simulations?

doc/tutorials/02-charged_system/02-charged_system-1.ipynb
```diff
@@ -98,21 +222,53 @@
...
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "We now fill this simulation box with particles at random positions, using type and charge from our dictionaries.\n",
-    "Using the length of the particle list <tt>system.part</tt> for the id, we make sure that our particles are numbered consecutively.\n",
-    "The particle type is used to link non-bonded interactions to a certain group of particles."
+    "Before we can start the simulation, we need to remove the overlap between particles to avoid large forces which would crash the simulation. For this, we use the steepest descent integrator with a relative convergence criterium for forces and energies."
```
RudolfWeeber: The overlap removal needs to happen before the P3M is turned on. Otherwise, the steepest descent could move charges on top each other.

doc/tutorials/02-charged_system/02-charged_system-1.ipynb
```diff
@@ -147,41 +300,119 @@
...
+    "* Write a function to integrate the system and return the total energy and the radial distance of the ions from the rod\n",
+    "* Use the signature ``integrate_calc_observables(system, N_frames, steps_per_frame, ion_types)``\n",
+    "* ``N_frames`` is the total number of samples taken from the system\n",
+    "* ``steps_per_frame`` is the the number of integration steps between consecutive frames\n",
+    "* ``ion_types`` is a list of types for which the radial distances should be calculated. For the moment we only have counterions, but later we will also add additional salt ions for which we would like to calculate the density\n",
+    "* return a list of the total system energy as well as a dictionary of the radial positions sorted by ion type, e.g. ``radial_distances[counterion_type] = [1.1,4.5,2.7,...]``\n",
+    "\n",
+    "**Hints:**\n",
+    "* We are only interested in the distance from the rod, so the $x_3$ coordinate is irrelevant\n",
+    "* Use ``system.part.select(type=...)`` to get only the particles of a specific type\n",
```
RudolfWeeber: If the handle to the slice is kept when creating the particles, it can be passed in here rather than the type, unless two counter ion types come up later in the tutorial?

doc/tutorials/02-charged_system/02-charged_system-1.ipynb
```diff
@@ -147,41 +300,119 @@
...
+    "\n",
+    "        print(f'progress={frame_idx/N_frames*100:.0f}%', end='\\r')\n",
+    "\n",
+    "        system.integrator.run(steps_per_frame)\n",
+    "\n",
+    "        energies.append(system.analysis.energy()['total'])\n",
+    "\n",
+    "        for ion_type, ions in particles_by_type.items():\n",
+    "            for ion in ions:\n",
+    "                radial_distances[ion_type].append(np.linalg.norm(ion.pos_folded[0:2]-system_center[0:2]))\n",
```
RudolfWeeber: Why don't you use the CylindricalDensityProfile observable? If there is a reason to do the calculation by hand, plese mention in the hints that pos_folded should be used, and why.

doc/tutorials/02-charged_system/02-charged_system-1.ipynb
```diff
@@ -380,68 +680,129 @@
...
-    "# Plot the distribution functions\n",
-    "plt.figure(figsize=(10, 6), dpi=80)\n",
-    "plt.plot(r[:], rdf_00[:], label='Na$-$Na')\n",
-    "plt.plot(r[:], rdf_01[:], label='Na$-$Cl')\n",
-    "plt.xlabel('$r$', fontsize=20)\n",
-    "plt.ylabel('$g(r)$', fontsize=20)\n",
-    "plt.legend(fontsize=20)\n",
-    "plt.show()"
+    "def add_salt(system, anion_params, cation_params):        \n",
+    "    for _ in range(anion_params['number']):\n",
```
RudolfWeeber: Ideally, create the particles in a single call and keep a reference to the resulting slice.

doc/tutorials/02-charged_system/02-charged_system-1.ipynb
```diff
@@ -380,68 +680,129 @@
...
+    "                'number':500}\n",
+    "cation_params = {'type':4,\n",
+    "                 'valency':2,\n",
+    "                 'number':500}\n",
+    "rod_length = 21\n",
+    "#TODO\n",
+    "steps_per_frame = 100\n",
+    "N_frames = 500\n",
+    "\n",
+    "total_anion_charge = -anion_params['number']*anion_params['valency']\n",
```
RudolfWeeber: If the slice is kept on particle creation, their charge can be queried directly rather than be assumed
`np.sum(an_ions.q)`

doc/tutorials/Readme.rst
```diff
@@ -16,6 +16,7 @@ physical systems. Currently, the following tutorials are available:
 * :file:`08-visualization`: Using the online visualizers of ESPResSo
 * :file:`11-ferrofluid`: Modelling a monolayer ferrofluid system
 * :file:`12-constant_pH`: Modelling the titration of a weak acid using the constant pH method
+* :file:`13-charged_rod`: Modelling counterion condensation on a charged rod
```
RudolfWeeber: Numbers mismatch. But do we really want to delete the crystal stuff?

RudolfWeeber (changes requested):
I think it is a nice tutorial. I'd have aded it as a new one rather than replace the crystal stuff, though.
For minor points in how the script is written, please see the comments.
It would be nice to have a few sentences interpretation after the plots, i.e., where PB theory holds and where it does not.
It would also be good to add a tutor's notes files with learning objectives in terms of physics and Espresso and points the tutor might mention along the way.
Please see the one contained in the tutorial 1 PR.

For the school, the first part is probably enough. Epending on how much more work the second part needs, it might be useful to split this pr.

christophlohrmann:
@RudolfWeeber Ok I will make the requested changes and reduce the PR to only touch part 1. As for the replacement of the crystal, @schlaicha can probably argue for that

doc/tutorials/02-charged_system/02-charged_system-1.ipynb
christophlohrmann: would not the p3m exactly prevent same-sign charges to be too close to each other before the real simulation starts?

RudolfWeeber:
>would not the p3m exactly prevent same-sign charges to be too close to each other before the real simulation starts?

 

Not sure, the real space part does have a divergence at r=0, doesn’t it? With dipoles you can definitely crash the system by doing steepest descent with an active p3m.

That aside, doing it without p3m is faster.

christophlohrmann:
The wca diverges as well. And in the tutorial the overlap removal is way faster than the "production" runs anyway

doc/tutorials/Readme.rst
schlaicha: A few comments on the crystal tutorial:
* Minor point(?): for @christophlohrmann and myself it was not clear what the physical motivation of such a simulation would be: For nanocapacitors/elechtrochemistry you need a constant charge or constant E-field whereas we employ a constant D-Field and the cystallization itself is not so fascinating IMHO.
* More important: The tutorial focuses a lot on physical constants/units and how to convert them to ESPResSo, but there are a lot of inconsistencies that we could not resolve. For example, the mass density of the ionic liquid was chosen to match experimental numbers, but in the script the electrode mass was included??
* Major point: Througout the tutorial we consider an ionic liquid close to its coexistence with the crystal, i.e. correlation times explode which is not really useful for an interactive tutorial (I agree though that it's something we should/could have somewhere [where?]). The dilute electrolyte is much nicer in this respect.
Again, I'm not insisting on removing the crystal tutorial, but to make it convincing IMO there is much more work needed than in the present PR for the dilute electrolyte (which also offers physical questions/interpretations in much shorter simulation times).

schlaicha:
> would not the p3m exactly prevent same-sign charges to be too close to each other before the real simulation starts?
> Not sure, the real space part does have a divergence at r=0, doesn’t it? With dipoles you can definitely crash the system by doing steepest descent with an active p3m. That aside, doing it without p3m is faster.

The steepest descent searches for local minima and thus might end up indeed in an unfavourable state for the simulation; however, having electrostatics turned on/off is only a minor change to this problem: Indeed you could have a staring configuration where the combination of electrostatic and non-electrostatic forces can trap you in a local minimum that will lead to forces that make your simulation crash, but on the other hand in general by minimizing without electrostatics and then turning it on afterwards you might also get a strong electrostatic force. My favour is minimizing with all interactions present but this might well be a matter of taste.

In the end the minimization is so fast anyway that the argument of having P3M in there or not should be a minor detail of the simulation.

Issue #3844, opened 2020-08-03
Test rotational dynamics against numerically integrated Euler eqn


RudolfWeeber:

jngrad:
More details in #3713. The paper is Omelyan 1998 ([10.1063/1.168642](https://doi.org/10.1063/1.168642)).

jngrad:
I tried again to substitute the indices from the paper using the quaternion formalism in ESPResSo:
```c++
Qd[0] = 0.5 * (-p.r.quat[3] * p.m.omega[0] - p.r.quat[0] * p.m.omega[1] + p.r.quat[2] * p.m.omega[2]);
Qd[1] = 0.5 * (p.r.quat[0] * p.m.omega[0] - p.r.quat[3] * p.m.omega[1] - p.r.quat[1] * p.m.omega[2]);
Qd[2] = 0.5 * (p.r.quat[1] * p.m.omega[0] + p.r.quat[2] * p.m.omega[1] + p.r.quat[0] * p.m.omega[2]);
Qd[3] = 0.5 * (-p.r.quat[2] * p.m.omega[0] + p.r.quat[1] * p.m.omega[1] - p.r.quat[3] * p.m.omega[2]);

Qdd[0] = 0.5 * (-p.r.quat[3] * Wd[0] - p.r.quat[0] * Wd[1] + p.r.quat[2] * Wd[2]) - p.r.quat[1] * S1;
Qdd[1] = 0.5 * (p.r.quat[0] * Wd[0] - p.r.quat[3] * Wd[1] - p.r.quat[1] * Wd[2]) - p.r.quat[2] * S1;
Qdd[2] = 0.5 * (p.r.quat[1] * Wd[0] + p.r.quat[2] * Wd[1] + p.r.quat[0] * Wd[2]) - p.r.quat[3] * S1;
Qdd[3] = 0.5 * (-p.r.quat[2] * Wd[0] + p.r.quat[1] * Wd[1] - p.r.quat[3] * Wd[2]) - p.r.quat[0] * S1;
```
but these yield incorrect angular velocities when plugged into `src/core/rotation.cpp`. I also worked out the A matrix backwards from the source code, but it doesn't look like the one from Omelyan 1998. I'm not sure where the matrix comes from.

[...]

Issue #1917, opened 2018-03-19
Brownian Dynamics: anisotropic moment of inertia support


bogdan-tanygin:
Non-linear components of the rotational Euler's EOM have not been taken into account (by myself) within the PR #1842. Probably, we (me) need to make some calculations and prepare new PR with corresponding capabilities (in case if a closed-form solution exists there at all). Meanwhile, I've added the documentation to the #1842 with a corresponding limitation description.

bogdan-tanygin:
This could be a fundamental limitation of the BD in case of an unstable stationary terminal drag motion possibility under some external torque applied. Similarly like the Dzhanibekov effect but for the viscous drag. I'll check, maybe someone already knows an answer.

PR #3887, opened 2020-09-07
Rotational code maintenance

### LINKED ISSUE #3101: GB should return only one torque
Currently GB returns one force, but two independent torques. This is not needed,
as the opposing torque can be calculated (as is the opposing force). It should instead
return a `ParticleForce`.

jngrad:
Description of changes:
- return a `ParticleForce` from pair force kernels (fixes #3101)
- add test cases for rotational dynamics (partial fix for #3844)
    - check trivial torque-driven and torque-free cases


jngrad:
@RudolfWeeber a few notes:
* I'm not exactly sure why the torque-free case drifts with error function `x * sin(x)` (details in https://github.com/espressomd/espresso/issues/3844#issuecomment-688516319), so I used the envelope of the error function in the test tolerance (line `tol = 1e-6 + 1.1e-8 * j`).
* I tried to write a torque-free case with rotations on 2 axes, but couldn't reproduce the trajectory even after re-implementing the `omega[0] = time_step * (torque[0] + omega[1] * omega[2] * (rinertia[1] - rinertia[2])) / rinertia[0]` logic in Python with the same number of iterations as in the core.
* The quaternion function doesn't seem to be called by the brownian and stokesian integrators. Is it intentional?

jngrad:
> * The quaternion function doesn't seem to be called by the brownian and stokesian integrators. Is it intentional?

Already discussed in #1917 for BD. The documentation has the following disclaimer, which checks out in the torque-driven cases:

> Note: the rotational Brownian dynamics implementation is compatible with particles which have the isotropic moment of inertia tensor only.

src/core/forces_inline.hpp
```diff
@@ -192,15 +192,14 @@ inline Utils::Vector3d calc_non_bonded_pair_force_parts(
...
-    auto const forces =
-        gb_pair_force(p1.r.calc_director(), p2.r.calc_director(), ia_params, d,
-                      dist, torque1, torque2);
-    force += std::get<0>(forces);
+    auto const pf = gb_pair_force(p1.r.calc_director(), p2.r.calc_director(),
+                                  ia_params, d, dist, torque1);
+    force += pf.f;
     if (torque1) {
-      *torque1 += std::get<1>(forces);
+      *torque1 += pf.torque;
```
fweik: I think there is a low hanging fruit here for considerable improvement: I think it would be better to return a `ParticleForce` from this function, instead of the sep. output parameters for the torques, and calculate the opposing
torque at the call site (as it is done for the force). If this is combined with a function `ParticleForce opposing_force(distance, f)` or similar, this can be made more ideomatic and there is no mention of torques anywhere is rotation is disabled. Also that function can be reused in other places, where this logic is also need.
If that isn't clear, the opposing force calculation is nothing specific for gay bern, it's just a consequence of of
momentum conservation, and this is the same for all central forces that produce toruqe. This is just a consequence
of Newtons law.

src/core/forces_inline.hpp
```diff
@@ -192,36 +193,23 @@ inline Utils::Vector3d calc_non_bonded_pair_force_parts(
...
-inline Utils::Vector3d calc_non_bonded_pair_force(Particle const &p1,
-                                                  Particle const &p2,
-                                                  Utils::Vector3d const &d,
-                                                  double dist) {
-  IA_parameters const &ia_params = *get_ia_param(p1.p.type, p2.p.type);
-  return calc_non_bonded_pair_force(p1, p2, ia_params, d, dist);
+inline Utils::Vector3d calc_opposing_torque(ParticleForce const &pf,
+                                            Utils::Vector3d const &d) {
+#ifdef ROTATION
+  // guard against numerical instability in the vector product
```
fweik: Can you explain what numerical instability you mean? Isn't the vector product just a sum of products some of the components would just vanish if there are zero-components in the torque?

jngrad: The torque isn't part of the vector product, so we cannot rely on vanishing components. If the first torque is a null vector, the second torque has to be a null vector too, because the force is collinear to the distance vector, but due to precision errors there is sometimes a residual with magnitude 1e-16 in the calculated counter-torque. This causes collision detection to crash:
```
 13/158 Test  #93: collision_detection .......................................***Failed   18.87 sec
f=[4.986014e+02   4.449108e+02 -5.791557e+02] d=[ 5.224425e-02  4.661846e-02 -6.068486e-02] t2=[-0.000000e+00 -3.552714e-15 -0.000000e+00]
f=[-1.246381e+02 -2.226122e+02 -1.175642e+02] d=[-4.399725e-02 -7.858208e-02 -4.150015e-02] t2=[-1.776357e-15 -0.000000e+00 -0.000000e+00]
f=[5.373112e+00  -5.428668e+00 -7.974942e+00] d=[ 5.381213e-02 -5.436853e-02 -7.986966e-02] t2=[-0.000000e+00 -5.551115e-17 -0.000000e+00]
terminate called after throwing an instance of 'std::runtime_error'
  what():  misplaced particle with id 200
[seraue:2333958] *** Process received signal ***
[seraue:2333958] Signal: Aborted (6)
[seraue:2333958] Signal code:  (-6)
...
[seraue:2333958] [ 7] /work/jgrad/espresso/build/src/core/EspressoCore.so(CellStructure::check_particle_sorting()+0x167)[0x14a3878f4307]
[seraue:2333958] [ 8] /work/jgrad/espresso/build/src/core/EspressoCore.so(cells_update_ghosts(unsigned int)+0xc1)[0x14a38783b4b1]
[seraue:2333958] [ 9] /work/jgrad/espresso/build/src/core/EspressoCore.so(handle_collisions()+0x819)[0x14a387845989]
[seraue:2333958] [10] /work/jgrad/espresso/build/src/core/EspressoCore.so(integrate(int, int)+0x1c3)[0x14a387892a63]
```

fweik: Ok I see. I'd say this is more a problem with the coldet test, and should be fixed there.

jngrad: The comment line is misleading, I'll change it.

src/core/forces_inline.hpp
jngrad: done

RudolfWeeber (approved):

PR #3888, opened 2020-09-08
Refactor thermostat infrastructure

### LINKED ISSUE #3573: Unit tests thermostats_test and random_test are very slow

### LINKED ISSUE #3585: Thermostat noise for different seeds is correlated

### LINKED ISSUE #3840: Noise behavior of stokesian dynamics and other thermostats differ

jngrad:
Description of changes:
- Remove RNG correlation stemming from seed offsets (fixes #3585)
    - seeds are now used as keys
    - a monotonically increasing counter is used in each thermostat
    - the only way to reset these counters is to create a new `System`
- Remove RNG correlation stemming from resetting `sim_time` or `time_step` during simulations with SD (fixes #3840)
    - the SD thermostat now uses the same RNG interface as other thermostats
- Accelerate RNG unit tests (fixes #3573)
    - they now take 2 seconds to run in coverage and sanitizer builds in CI
- Separate thermostats from integrators
    - better separation of concerns


KaiSzuttor:
I think we already discussed that we do not want to have one counter per thermostat. Let's please first discuss and then implement.

jngrad:
You can view this PR draft as a subset of the first one. It's unclear to me how the DPD specifics can be resolved and how much time it will take. This PR solves all correlations issues; moving from the state in this PR to the state with a global counter should be straightforward.

doc/sphinx/system_setup.rst
```diff
@@ -210,6 +210,15 @@ the temperature of a thermostat, you actually do not define the
 temperature, but the value of the thermal energy :math:`k_B T` in the
 current unit system (see the discussion on units, Section :ref:`On units`).
 
+All thermostats have a ``seed`` argument that controls the state of the random
+number generator (Philox Counter-based RNG). This seed is required on first
+activation of a thermostat, unless stated otherwise. It can be omitted in
+subsequent calls of the method that activates the same thermostat. The random
+sequence also depends on a monotonically increasing global counter that is
```
KaiSzuttor: that is not true for this PR

doc/sphinx/system_setup.rst
```diff
@@ -210,6 +210,15 @@ the temperature of a thermostat, you actually do not define the
...
+incremented after each integration step. It is the user's responsibility to
+decide whether the thermostats should be deterministic (by using a fixed seed)
```
KaiSzuttor: that's confusing... the thermostat is always deterministic for a given seed.

src/core/constraints/ShapeBasedConstraint.cpp
```diff
@@ -77,7 +78,7 @@ ParticleForce ShapeBasedConstraint::force(Particle const &p,
         force1 +=
             dpd_pair_force(p, part_rep, ia_params, dist_vec, dist, dist * dist);
         // Additional use of DPD here requires counter increase
-        dpd_rng_counter_increment();
+        dpd.rng_increment();
```
KaiSzuttor, line 81: shouldn't we increment the counter where the random number is used? actually, we can increment the counter automatically in the function that provides the random number.

src/core/integrators/velocity_verlet_npt.cpp
```diff
@@ -69,10 +69,10 @@ void velocity_verlet_npt_finalize_p_inst() {
     }
   }
 
-  double p_tmp = 0.0;
-  MPI_Reduce(&nptiso.p_inst, &p_tmp, 1, MPI_DOUBLE, MPI_SUM, 0, comm_cart);
+  double p_sum = 0.0;
+  MPI_Reduce(&nptiso.p_inst, &p_sum, 1, MPI_DOUBLE, MPI_SUM, 0, comm_cart);
```
KaiSzuttor: please use boost mpi

src/python/espressomd/interactions.pyx
```diff
@@ -1700,6 +1700,15 @@ cdef class BondedInteraction:
     def __reduce__(self):
         return (self.__class__, (self._bond_id,))
 
+    def __getstate__(self):
+        # for most bonds, we only need to pickle arguments passed to the
+        # constructor, yet some bond classes need to pass additional data
+        return (self.params,)
+
+    def __setstate__(self, params):
+        # parameters are already set in the core by the contructor
```
KaiSzuttor: that comment is wrong.... the pickle interactions is in the ```__reduce__``` function, right? https://docs.python.org/3/library/pickle.html#object.__reduce__

src/python/espressomd/interactions.pyx
```diff
@@ -1700,6 +1700,15 @@ cdef class BondedInteraction:
     def __reduce__(self):
         return (self.__class__, (self._bond_id,))
 
+    def __getstate__(self):
+        # for most bonds, we only need to pickle arguments passed to the
```
KaiSzuttor: this function should not be needed. please adapt the existing ```__reduce__``` function if necessary.

src/python/espressomd/interactions.pyx
```diff
@@ -2091,6 +2100,15 @@ class ThermalizedBond(BondedInteraction):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
+    def __getstate__(self):
+        return (self.params, thermalized_bond.rng_counter())
+
+    def __setstate__(self, params):
+        # parameters are already set in the core by the contructor, we only
```
KaiSzuttor: we should not manage the pickle interaction in several places

KaiSzuttor (changes requested):

RudolfWeeber:
For every case possible, the counter increment should be in one central place called from the integrate function, shouldn’t it?

That would already give us some of the benefit (and the basic structure) for the global counter.

For now, dpd with constraints has to be the exception.

doc/sphinx/system_setup.rst
jngrad: good catch

doc/sphinx/system_setup.rst
jngrad: I find it confusing too, it's probably better to simply remove that last sentence

src/core/constraints/ShapeBasedConstraint.cpp
jngrad: Don't we want to use the same counter state for all calls to RNGs within the same integration step? Otherwise the trajectory depends on the order particles are traversed.

On a side-note, I think this function actually belongs to `forces.cpp` or similar. The constraint shouldn't need to know anything about thermostats and non-bonded interactions, and should simply pass a reference to its particle representation.

src/python/espressomd/interactions.pyx
jngrad: didn't realize `__reduce__` was used in pickle interactions, I'll revert these changes and use `__reduce__` directly

jngrad:
> For every case possible, the counter increment should be in one central place called from the integrate function, shouldn’t it? That would already give us some of the benefit (and the basic structure) for the global counter.

Yes, that's what is achieved by this PR: the integration loop in `integrate.cpp` calls `void philox_counter_increment()` which is defined in `thermostat.cpp`, where all currently active thermostats get their counter incremented.

src/python/espressomd/interactions.pyx
jngrad: changes reverted, counter is now checkpointed by `__reduce__` in 9c454a62aea

testsuite/python/test_checkpoint.py
```diff
@@ -277,30 +282,30 @@ def test_integrator_SDM_gpu(self):
...
+        reference = {'r_0': 0.0, 'k': 1.0, 'r_cut': 0.0}
+        self.assertEqual(state, reference)
         state = system.part[1].bonds[0][0].params
-        reference = {'r_0': 0.0, 'k': 1.0}
-        for key in reference:
-            self.assertAlmostEqual(state[key], reference[key], delta=1E-10)
+        self.assertEqual(state, reference)
         if 'THERM.LB' not in modes:
-            state = system.part[1].bonds[1][0].params
+            state = system.part[1].bonds[1][0]._get_params_from_es_core()
```
KaiSzuttor: you should not rely on a private function here...

KaiSzuttor (changes requested):
LGTM besides comment

testsuite/python/test_checkpoint.py
jngrad: agreed

KaiSzuttor (approved):

Issue #3889, opened 2020-09-09
p3m tuning unintuitive


christophlohrmann:
Below example shows that calling the ``tune()`` function of a previously tuned p3m will not tune all parameters again. To get an actual retuning of the parameters, ``tune()`` has to be given the original default values of p3m (except ``r_cut``, that needs special treatment)

Expected/intuitive behaviour would be a  tuning of all parameters by default and not tuning the parameters that are explicitly passed.

```python
from espressomd import System, electrostatics
import numpy as np

system = System(box_l = 3*[10])
system.time_step = 0.01
system.cell_system.skin = 0.4
for _ in range(25):
    system.part.add(pos=np.random.random(3) * system.box_l, q=1)
    system.part.add(pos=np.random.random(3) * system.box_l, q=-1)

p3m = electrostatics.P3M(prefactor = 1, accuracy = 1e-3)
system.actors.add(p3m)

print('\n\n\n')
p3m.tune()
#observe: mesh, r_cut, cao are fixed

def retune_p3m(p3m):
    p3m_default_params = p3m.default_params()
    p3m.tune(mesh = p3m_default_params['mesh'],
             cao = p3m_default_params['cao'],
             r_cut = 0,
             alpha = p3m_default_params['alpha'])

print('\n\n\n')
retune_p3m(p3m) 
```

PR #3890, opened 2020-09-09
core: p3m: reset r_cut when default value is provided in tuning

reinaual:
This fixes the behavior of `p3m.tune()` if the `r_cut`-default value is provided (mentioned in #3889).

### DIFF
--- a/src/core/electrostatics_magnetostatics/p3m.cpp
+++ b/src/core/electrostatics_magnetostatics/p3m.cpp
@@ -192,6 +192,9 @@ void p3m_set_tune_params(double r_cut, const int mesh[3], int cao, double alpha,
   if (r_cut >= 0) {
     p3m.params.r_cut = r_cut;
     p3m.params.r_cut_iL = r_cut * (1. / box_geo.length()[0]);
+  } else if (r_cut == -1.0) {
+    p3m.params.r_cut = 0;
+    p3m.params.r_cut_iL = 0;
   }
 
   if (mesh[0] >= 0) {


reinaual:
With the current interface the retuning of all parameters in P3M should be achievable by providing the default values:

```
import espressomd

system = espressomd.System(box_l=[1, 1, 1])

system.time_step = 0.01

system.part.add(pos=[0, 0, 0], q=1)
system.part.add(pos=[0.5, 0, 0], q=-1)


p3m = espressomd.electrostatics.P3M(prefactor=1,
                                    accuracy=1e-3)

system.actors.add(p3m)

p3m_default_params = p3m.default_params()

p3m.tune(mesh = p3m_default_params['mesh'],
         cao = p3m_default_params['cao'],
         alpha = p3m_default_params['alpha'],
         r_cut = p3m_default_params['r_cut'])
```

without this the output of `p3m.tune(...)` looks like this

```
P3M tune parameters: Accuracy goal = 9.94074e-04 prefactor = 1.00000e+00
System: box_l = 1.00000e+00 # charged part = 2 Sum[q_i^2] = 2.00000e+00
fixed r_cut_iL 0.494141
mesh cao r_cut_iL     alpha_L      err          rs_err     ks_err     time [ms]
2    7   cao too large for this mesh
```
and changes to
```
P3M tune parameters: Accuracy goal = 9.94074e-04 prefactor = 1.00000e+00
System: box_l = 1.00000e+00 # charged part = 2 Sum[q_i^2] = 2.00000e+00
mesh cao r_cut_iL     alpha_L      err          rs_err     ks_err     time [ms]
2    7   cao too large for this mesh
2    7   cao too large for this mesh

```

KaiSzuttor:
Documentation states

> All parameters passed to the method [tune()] are fixed in the tuning routine.

Maybe I misunderstand this but this means, that the current behavior is not what the doc says and the behavior in this PR is also not what the doc says.

What I'd expect from the tuning function is the following behavior:

- If I want to fix a parameter, i provide it to ```tune``` as a kwarg
- If I don't provide any parameter, I expect the tuning of all parameters

@reinaual what do you think?

reinaual:
I admit that i did not read the documentation, and what you wrote is exactly what @christophlohrmann mentioned in #3889. The `tune()` function does not what is documented, with this PR this is not fixed, only the behavior of the `r_cut` parameter is matched to the behavior of the others.
If you decided that we want the behavior which is documented for the `tune()` function, i can have a look at this. I just did not want to change the behavior of this function without being told to do so.

KaiSzuttor:
@reinaual yeah please do so :smile:, should be fixed

PR #3896, opened 2020-09-15
Fix Stokesian Dynamics build system

jngrad:
Description of changes:
- fix an incorrect conditional that lead CMake to fetch Stokesian Dynamics even though its dependencies were not found


christophlohrmann (approved):
Thanks for the fix

KaiSzuttor (approved):

PR #3902, opened 2020-09-17
Check tutorial URLs pointing to the user guide

jngrad:
Check that tutorials are always in sync with the Sphinx documentation.


jngrad:
The check works as expected in CI: with broken URLs in the tutorials (9e928e127), the pipeline fails ([logfile](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/252745)).

RudolfWeeber (approved):

PR #3903, opened 2020-09-19
WIP: Gibbs ensemble scripts

RudolfWeeber:
This refactors the Gibbs ensemble scripts.
The final state, to my understanding, should be the following:
There is a server script, which, via socket communication, controls two clients. 
Each client represents one of the two systems of the Gibbs ensemble.

The server knows nothing about the details of the simulation system. 
It controls the Gibbs Monte Carlo protocol and issues commands to the clients for
* particle insertion
* particle removal
* move a single particle
* volume change
* confirm/revert last operation
It also receives updated energies from the clients.

A client base class implements these operations for a standard Espressso system.
Users can then derive from that for their individual cases. For the LJ case which is currently present, the only thing which needs to be overridden is the energy, where tail and shift corrections have to be added.


RudolfWeeber:
So far, this contains disentangling and readability work. 
* Box indices have been replaced by references to the box in most cases
* Energy corrections are moved to the client side as they are system-dependent
* Readability

@schlaicha

jonaslandsgesell:
I think that the Gibbs ensemble implementation would be more helpful if (in each box) MD would be used in the step "MOVE_PART" to propose new states instead of a single MC move -- and then accept/reject the new state (after running MD) according to the metropolis criterion.  This way also polymeric systems could be simulated.

RudolfWeeber:
>I think that the Gibbs ensemble implementation would be more helpful if (in each box) MD would be used in the step "MOVE_PART" to propose new states >instead of a single MC move -- and then accept/reject the new state (after running MD) according to the metropolis criterion. This way also polymeric systems >could be simulated.

I agree. For the tutorial, I’m going to leave the MC variant, though.

schlaicha:
> I think that the Gibbs ensemble implementation would be more helpful if (in each box) MD would be used in the step "MOVE_PART" to propose new states >instead of a single MC move -- and then accept/reject the new state (after running MD) according to the metropolis criterion. This way also polymeric systems >could be simulated.
> I agree. For the tutorial, I’m going to leave the MC variant, though.

In general sampling the configurational space via MD steps yields "more physical" configurations (whatever that means), yet sometimes this is exactly what you want to avoid when you do MC (like e.g. if there is an energy barrier, you can do collective moves etc.). I.e. we should leave the interface as flexible as possible and let the user decide, together with a few typical samples/tutorials. I discussed with @RudolfWeeber that in a mid-termed project we could enhance ESPResSo's MC capabilities.

samples/gibbs_ensemble/gibbs_ensemble_socket.py
```diff
@@ -53,7 +53,7 @@
 parser.add_argument('-C', '--client-script', nargs=1)
 parser.add_argument('-n', '--number-of-particles', type=int, nargs=1)
 parser.add_argument('-l', '--box-length', type=float, nargs=1)
-parser.add_argument('-T', '--temperature', type=float, nargs=1)
+parser.add_argument('-T', '--temperature', type=float, nargs=1, required=True)
```
KaiSzuttor: it's kind of odd to declare an optional argument and make it required, why not use a positional argument?

samples/gibbs_ensemble/gibbs.py
```diff
@@ -0,0 +1,139 @@
+import socket
+import struct
+import pickle
+import numpy as np
+
+
+class MessageId:
```
KaiSzuttor: this should use enumerations which are a language feature since 3.4:

``` python
import enum

@enum.unique
class MessageId(enum.Enum):
    START = 0
```

samples/gibbs_ensemble/gibbs.py
```diff
@@ -0,0 +1,139 @@
...
+    CHANGE_VOLUME = 3
+    CHANGE_VOLUME_REVERT = 31
+    EXCHANGE_PART_ADD = 4
+    EXCHANGE_PART_ADD_REVERT = 41
+    EXCHANGE_PART_REMOVE = 5
+    EXCHANGE_PART_REMOVE_REVERT = 51
+    ENERGY = 6
+
+
+class Client():
```
KaiSzuttor: these parentheses can be removed

samples/gibbs_ensemble/gibbs.py
```diff
@@ -0,0 +1,139 @@
...
+    '''
+    Base class for a Gibbs ensemble client. Each client runs a single
+    system, executes requested changes and returns energies.
+    '''
+    _system = None
```
KaiSzuttor: this should not be a class variable, right? each instance of the client class should have it's own instance of a system.

samples/gibbs_ensemble/gibbs.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def __init__(self, system):
+        self._system = system
+
+    def random_particle(self):
+        id = np.random.choice(self._system.part[:].id)
+        return self._system.part[id]
```
KaiSzuttor, line 35: why not ```return np.random.choice(self._system.part)```

samples/gibbs_ensemble/gibbs.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def move_particle(self):
+        '''Moves a particle to a new random position. The old position is saved in _old_part_position.'''
+        p = self.random_particle()
+        self._last_modified_pid = p.id
+        self._saved_pos = p.pos
```
KaiSzuttor, line 41: don't you have to copy here?

samples/gibbs_ensemble/gibbs.py
```diff
@@ -0,0 +1,139 @@
...
+    CHANGE_VOLUME_REVERT = 31
+    EXCHANGE_PART_ADD = 4
+    EXCHANGE_PART_ADD_REVERT = 41
+    EXCHANGE_PART_REMOVE = 5
+    EXCHANGE_PART_REMOVE_REVERT = 51
+    ENERGY = 6
+
+
+class Client():
+    '''
```
KaiSzuttor, line 25: i think we should consistently use ```"""```

RudolfWeeber:
A cool. I gave up un enums I last checked, because there were two competing enum packages…

RudolfWeeber:
>it's kind of odd to declare an optional argument and make it required, why not use a positional argument?

I don’t use positional arguments, because the order can be mistaken easily. That’s what I also preach in my “managing simulation data” lecture.

samples/gibbs_ensemble/gibbs.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def backup_particle(self, p):
+        """This implementation only saves position"""
+        self._saved_particle_state = {'id': p.id, 'pos': p.pos}
+
+    def remove_particle(self):
+        '''Removes a random particle. The old position and index are saved'''
+        p = self.random_particle()
+        self.backup_particle(p)
+        p.remove()
```
KaiSzuttor: if we wanted to show best practices here, we should probably not use a lot of function with side-effects and consider changing it to a more functional style approach with free functions that return stuff (and can be tested)

samples/gibbs_ensemble/gibbs.py
```diff
@@ -0,0 +1,139 @@
...
+            elif msg[0] == MessageId.EXCHANGE_PART_REMOVE:
+                self.remove_particle()
+                self._send_data(pickle.dumps(
+                    [MessageId.ENERGY, self.energy()]))
+            elif msg[0] == MessageId.MOVE_PART_REVERT:
+                self.move_particle_revert()
+            elif msg[0] == MessageId.EXCHANGE_PART_ADD_REVERT:
+                self.remove_last_added_particle()
+            elif msg[0] == MessageId.EXCHANGE_PART_REMOVE_REVERT:
+                self.remove_particle_revert()
```
KaiSzuttor: this whole block is probably something like a ```MessageHandler``` that can react on messages of type ```MessageId``` and is more or less just a dict from ```MessageId``` to a function.

samples/gibbs_ensemble/gibbs_ensemble_client.py
```diff
@@ -0,0 +1,104 @@
...
+LJ_EPSILON = 1.0
+LJ_SIGMA = 1.0
+LJ_CUTOFF = 2.5
+LJ_SHIFT = 0
+
+
+parser = argparse.ArgumentParser()
+parser.add_argument('-s', '--seed', type=int)
+parser.add_argument('-H', '--host', type=str, default='localhost')
+parser.add_argument('-p', '--port', type=int, required=True)
```
KaiSzuttor: same here, make it a positional argument

samples/gibbs_ensemble/gibbs_ensemble_client.py
```diff
@@ -0,0 +1,104 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""
+Client part of the Gibbs ensemble simulation. This script handles the
+simulation boxes and communicates the energies to the host. The Monte-Carlo
+part of the simulation is done by the :file:`gibbs_ensemble_socket.py` script.
+"""
+
+import espressomd
```
KaiSzuttor: it's a matter of taste, but good practice is to first import external modules and after a blank the local ones

KaiSzuttor: there's a tool called ```isort``` for that: https://github.com/pycqa/isort/

samples/gibbs_ensemble/gibbs_ensemble_client.py
```diff
@@ -0,0 +1,104 @@
...
+    return 8.0 / 3.0 * np.pi * density * lj_epsilon * \
+        lj_sigma**3 * (1.0 / 3.0 * np.power(lj_cutoff / lj_sigma, -9) -
+                       np.power(lj_cutoff / lj_sigma, -3))
+
+
+def calc_shift_correction(density, lj_epsilon, lj_cutoff, lj_shift):
+    '''
+    Calculates the shift correction to the energies of the box.
+    '''
+    # difference in the potential integrated from 0 to cutoff distance
```
KaiSzuttor: this comment should go into the docstring of the function

samples/gibbs_ensemble/gibbs_ensemble_client.py
```diff
@@ -0,0 +1,104 @@
...
+        return super().energy() + \
+            calc_tail_correction(self.density(), self._lj_epsilon, self._lj_sigma, self._lj_cutoff) + \
+            calc_shift_correction(self.density(), self._lj_epsilon,
+                                  self._lj_cutoff, self._lj_shift)
+
+
+
+# init system
+# The intial system has to be big enough to fit twice the
+# maximuminteraction range
```
KaiSzuttor: space missing after maximum

KaiSzuttor (changes requested):
in general i think we want to use the [multiprocessing module](https://stackoverflow.com/questions/43861164/passing-data-between-separately-running-python-scripts) for the communication instead of plain sockets *in the future*. Besides my comments it looks okay to me.

KaiSzuttor:
> it's kind of odd to declare an optional argument and make it required, why not use a positional argument?
> I don’t use positional arguments, because the order can be mistaken easily. That’s what I also preach in my “managing simulation data” lecture.

If you have so many positional arguments that you fear to mix them up, you are probably doing something wrong. Here, for examples it seems you only have one required argument which should also be positional just as with any tool you use in your console.

RudolfWeeber:
I agree about using some higher-level communication infrastrucutre, but will not have time to do it in this iteration.

Before this becoms an Espresso module rather than a sample/tutorial, move classes should be introduced as well. I.e. move classse should hold the detail of a specific move and the steps to apply and revert it. It might be discussed whether acceptance criteria should be member functions or free functions (which would be ugly, because one cannot overload w/r to argument type in Python). Also not in this iteration, though.

samples/gibbs_ensemble/gibbs.py
RudolfWeeber: No. Instances of ArrayLocked contain a copy of the data. They are no longer tied to the underlying particle. Only write access to elements are blocked. This variable is only assigned and read.

samples/gibbs_ensemble/gibbs.py
RudolfWeeber: I don't really want to do such things manually. I personally use """, whereas the previous author used '''. I don't really find it important. If autopep8 can be convinced to do this automatically, I'm fine with adding that rule

samples/gibbs_ensemble/gibbs.py
RudolfWeeber: I implemented this. Tutors will have to do a bit more explaining when teching Python beginners, though.

samples/gibbs_ensemble/gibbs.py
RudolfWeeber: Doc states it requires an 1d array, which ParticleList is not.

samples/gibbs_ensemble/gibbs.py
KaiSzuttor: it just about consistency... just thought that python code in the samples is much more prominent to users than code in the python interface that is not read directly too often. Search and replace is also trivial in this case but if you think that code style is not that important in the samples you can leave it as is.

RudolfWeeber:
This is as far as I want to take the refactoring of the Gibbs ensemble scripts.

The task I actually intended to do was makeing a tutorial in which this is used to calculate critical temperatures...

Let's see how that goes...

samples/gibbs_ensemble/gibbs.py
KaiSzuttor: > For consistency, always use """triple double quotes""" around docstrings

[pep8](https://www.python.org/dev/peps/pep-0257/#id15)

samples/gibbs_ensemble/gibbs_ensemble_socket.py
```diff
@@ -443,10 +442,11 @@ def check_exchange_particle(source_box, dest_box, inner_potential):
         # timing
         tock = time.time()
 
-        print("step %d, densities %.3f %.3f, volumes %.2f, %.2f %.f ms / move" % 
-              (i, densities[0][-1], densities[1][-1], 
-               boxes[0].box_l**3, boxes[1].box_l**3,
-               (tock - tick) * 1000 / REPORT_INTERVAL))
+        if len(densities[0]):
+            print("step %d, densities %.3f %.3f, volumes %.2f, %.2f %.f ms / move" %
```
KaiSzuttor, line 446: this snippet gets a lot more readable if you use format strings ala ```print(f'step {i})```. I also think that in general it's good practice to *not* directly print stuff to stdout but instead use the logging infrastructure of python.

KaiSzuttor (changes requested):

samples/gibbs_ensemble/aggregate.py
```diff
@@ -0,0 +1,59 @@
+import numpy as np
+import sys
+import pickle
+import scipy.optimize
+import matplotlib.pyplot as plt
+
```
KaiSzuttor, line 6: docstrings are missing

samples/gibbs_ensemble/aggregate.py
```diff
@@ -0,0 +1,59 @@
...
+    data = pickle.load(open(f, 'rb'))
+
+    kT = data['args'].temperature
+    densities = np.array(data['densities'])
+    indices = data['list_indices']
+
+    # First index to consider
+    start = np.where(np.array(indices) > WARMUP_LENGTH)[0][0]
+
+    # Determin which is teh dilute phase from the last density recorded
```
KaiSzuttor, line 23: please use a spellchecker for the comments

samples/gibbs_ensemble/gibbs.py
KaiSzuttor: https://docs.python.org/3/library/random.html#random.choice

samples/gibbs_ensemble/lj_gibbs_client.py
```diff
@@ -0,0 +1,105 @@
...
+LJ_EPSILON = 1.0
+LJ_SIGMA = 1.0
+LJ_CUTOFF = 2.5
+LJ_SHIFT = 0
+
+
+parser = argparse.ArgumentParser()
+parser.add_argument('-s', '--seed', type=int)
+parser.add_argument('-H', '--host', type=str, default='localhost')
+parser.add_argument('-p', '--port', type=int, required=True)
```
KaiSzuttor, line 45: for a single required argument it should be avoided to mark it as optional. If you provide proper help messages with the ```help``` keyword argument to ```add_argument``` the meaning of the positional argument is not ambiguous.

RudolfWeeber:
>for a single required argument it should be avoided to mark it as optional. If you provide proper help messages with the help keyword argument to >add_argument the meaning of the positional argument is not ambiguous.

 

Can we please agree to disagree about positional arguments. I find their use in simulations scripts error-prone as I mentioned before.

RudolfWeeber:
>> +    def random_particle(self):
>+        id = np.random.choice(self._system.part[:].id)
>+        return self._system.part[id]

>https://docs.python.org/3/library/random.html#random.choice

 

I am aware of that function but I don’t trust it in this case, because from the docs it is not apparent to me, wheter the implementation relies on [] (or __getitem__()) to retrieve individual elements from the popoluation. If it does, or will in the future,  this would give wrong results, because system.part[] behaves somewhat like a dictionary.

So if len(system.part) is N that does not mean that all valid entries are accessible via [0] .. [N-1].

 

Safety fist.

KaiSzuttor:
> for a single required argument it should be avoided to mark it as optional. If you provide proper help messages with the help keyword argument to >add_argument the meaning of the positional argument is not ambiguous.
> Can we please agree to disagree about positional arguments. I find their use in simulations scripts error-prone as I mentioned before.

I still cannot follow your reasoning. Correct me if I'm wrong: you are saying that the standard way of calling any command line script should not be done here because it is error prone? What makes our sample scripts so special? And how can you use a single positional argument incorrectly?

KaiSzuttor:
> > + def random_particle(self): + id = np.random.choice(self._system.part[:].id) + return self._system.part[id] https://docs.python.org/3/library/random.html#random.choice
> I am aware of that function but I don’t trust it in this case, because from the docs it is not apparent to me, wheter the implementation relies on [] (or __getitem__()) to retrieve individual elements from the popoluation. If it does, or will in the future, this would give wrong results, because system.part[] behaves somewhat like a dictionary. So if len(system.part) is N that does not mean that all valid entries are accessible via [0] .. [N-1]. Safety fist.

so you are saying that our ```ParticleList``` does not behave like a python sequence? That's at least a surprising interface, right?

RudolfWeeber:
The most important issue with this is that it doesn't reproduce the scaling of
(rho_dens-rho_dilute) \sim (t-t_c)^beta with beta=0.32 
reported in the litrature.

RudolfWeeber:
Yes. It should probably be renamed to something not containing `list`.

KaiSzuttor:
> Yes. It should probably be renamed to something not containing `list`.

or maybe make it behave like a sequence?

KaiSzuttor:
it's probably a good idea to have some discussion about the particle container in the interface. the current one seems to have some surprising features.

RudolfWeeber:
>>it's probably a good idea to have some discussion about the particle container in the interface. the current one seems to have some surprising features.

Better think of it like a dictionary with some index slicing capabilities.

 

We had this discussion already. The existing behaviour was deemed the least supririnsg, since the argument of the [] always matches the particle id.

KaiSzuttor:
> >it's probably a good idea to have some discussion about the particle container in the interface. the current one seems to have some surprising features.
> Better think of it like a dictionary with some index slicing capabilities. We had this discussion already. The existing behaviour was deemed the least supririnsg, since the argument of the [] always matches the particle id.

fair enough. i still think there is room for improvement, *e.g.* by totally disentangling container index and particle id. It's not clear to me why they should be connected. However, that would mean to introduce a breaking change to a lot of user scripts.

itischler:
> The most important issue with this is that it doesn't reproduce the scaling of
> (rho_dens-rho_dilute) \sim (t-t_c)^beta with beta=0.32
> reported in the litrature.

From the first look at it it seems like the system is not very well equilibrated. The equilibration can be sped up by adjusting the probability of the different attempt types. was there any reason to set them to have equal probability?

jonaslandsgesell:
> From the first look at it it seems like the system is not very well equilibrated. The equilibration can be sped up by adjusting the probability of the different attempt types. was there any reason to set them to have equal probability?

Running MD for the displacement moves and then accepting rejecting the resulting conformations with the MC criterion would help a lot in speeding up and would avoid inventing hand taylored MC moves for every system (think of polymers etc).

RudolfWeeber:
>From the first look at it it seems like the system is not very well equilibrated. The equilibration can be sped up by adjusting the probability of the different >attempt types. was there any reason to set them to have equal probability?

 

That’s what The Book suggested. We didn’t have more success with other ratios, though. Also, the number of cycles was increased considerably, so even if the ratio is not optimal, a reasonable result would be expected.

 

The results aren’t just noisy, the scaling is incorrect (fitted via aggregate.py).

RudolfWeeber:
Jonas, I agree, but there is no point making it more complex until it is correct.

jonaslandsgesell:
> Jonas, I agree, but there is no point making it more complex until it is correct.

Applying the hybrid MD/MC scheme would also help making it correct and avoid convergence issues in MC due to long correlations.

itischler:
> That’s what The Book suggested. We didn’t have more success with other ratios, though.

Ok. I mean it should not matter as long as detailed balance is in place.

> Also, the number of cycles was increased considerably, so even if the ratio is not optimal, a reasonable result would be expected. The results aren’t just noisy, the scaling is incorrect (fitted via aggregate.py).

It is not that the data is noisy, but looking at the densities, they still drift apart after those 300000 steps (at least for the simulations i looked at). Which would lead to a smaller density difference when averaging begins at 100000 steps.

KaiSzuttor:
> Jonas, I agree, but there is no point making it more complex until it is correct.

I don't quite get the strategy here. How do you plan to find the issue? Is the only test for correctness a statistical test?

PR #3914, opened 2020-09-27
tutorial 2: charged rod

christophlohrmann:
Description of changes:
- replace crystal tuorial with charged rod tutorial


christophlohrmann:
Sorry I don't know what I did to kill the previous pull request.

I changed according to @RudolfWeeber's requests, but the radial profile is still manual.
I will look how to do that with an observable

schlaicha:
Adding my comments from the closed PR #3886 

A few comments on the crystal tutorial:

* Minor point(?): for @christophlohrmann and myself it was not clear what the physical motivation of such a simulation would be: For nanocapacitors/elechtrochemistry you need a constant charge or constant E-field whereas we employ a constant D-Field and the cystallization itself is not so fascinating IMHO.
* More important: The tutorial focuses a lot on physical constants/units and how to convert them to ESPResSo, but there are a lot of inconsistencies that we could not resolve. For example, the mass density of the ionic liquid was chosen to match experimental numbers, but in the script the electrode mass was included??
* Major point: Througout the tutorial we consider an ionic liquid close to its coexistence with the crystal, i.e. correlation times explode which is not really useful for an interactive tutorial (I agree though that it's something we should/could have somewhere [where?]). The dilute electrolyte is much nicer in this respect.
Again, I'm not insisting on removing the crystal tutorial, but to make it convincing IMO there is much more work needed than in the present PR for the dilute electrolyte (which also offers physical questions/interpretations in much shorter simulation times).

> would not the p3m exactly prevent same-sign charges to be too close to each other before the real simulation starts?
> Not sure, the real space part does have a divergence at r=0, doesn’t it? With dipoles you can definitely crash the system by doing steepest descent with an active p3m. That aside, doing it without p3m is faster.

The steepest descent searches for local minima and thus might end up indeed in an unfavourable state for the simulation; however, having electrostatics turned on/off is only a minor change to this problem: Indeed you could have a staring configuration where the combination of electrostatic and non-electrostatic forces can trap you in a local minimum that will lead to forces that make your simulation crash, but on the other hand in general by minimizing without electrostatics and then turning it on afterwards you might also get a strong electrostatic force. My favour is minimizing with all interactions present but this might well be a matter of taste.

In the end the minimization is so fast anyway that the argument of having P3M in there or not should be a minor detail of the simulation.

christophlohrmann:
The current state still contains the ugly workaround for #3920

christophlohrmann:
I have a version with a parameter set that does (most of the time) not trigger the k-space error. Caveat: Runs much longer and sometimes (~1 out of 5 times) the previous tuning leads to a k-space cutoff that still triggers the error.

RudolfWeeber:
Let’s try to get the tuning sorted out.

doc/tutorials/02-charged_system/02-charged_system-1.ipynb
```diff
@@ -4,33 +4,52 @@
    "cell_type": "markdown",
```
KaiSzuttor, line 4: <p>please use ALL_CAPITAL letters for constants</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/espressomd/espresso/pull/3914/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

KaiSzuttor, line 4: <p>same here</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/espressomd/espresso/pull/3914/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

RudolfWeeber (approved):

PR #3917, opened 2020-09-28
Tutorial 12 - constant-pH

jrfinkbeiner:
* restructured setup
* added exercises
* created tutor notes


doc/tutorials/12-constant_pH/12-constant_pH.ipynb
```diff
@@ -210,152 +372,308 @@
...
     "polymers = polymer.linear_polymer_positions(n_polymers=1,\n",
-    "                                            beads_per_chain=N_acid,\n",
+    "                                            beads_per_chain=N_ACID,\n",
     "                                            bond_length=0.9, seed=23)\n",
-    "for polymer in polymers:\n",
-    "    for index, position in enumerate(polymer):\n",
-    "        id = len(system.part)\n",
-    "        system.part.add(id=id, pos=position, type=TYPE_A, q=q_A)\n",
+    "# add the polymer particles composed of ionizable acid groups, initially in the ionized state\n",
+    "for local_polymer in polymers:\n",
```
RudolfWeeber: What's a `local_polymer`. 
Preferrably use `for polymer in polymers:`

For bond creation:
```
prev_particle = None
for pos in polymer:
    p = system.part.add(pos=pos)
    if prev_particle:
        p.add_bond((H,prev_particle))
    prev_particle = p[0]
```

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
```diff
@@ -1,5 +1,12 @@
 {
```
RudolfWeeber, line 1: Please mention a rationale for the p3m accuracy. Would 1E-3 be acceptable in production simulations? If not, pleae mention that.

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
```diff
@@ -210,152 +372,308 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "solution2": "hidden"
+   },
+   "source": [
+    "```python\n",
+    "def equilibrate_pH():\n",
+    "    RE.reaction(reaction_steps=20 * N_ACID + 1)\n",
```
RudolfWeeber, line 592: why +1?

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
```diff
@@ -210,152 +372,308 @@
...
    "source": [
-    "# the reference data from Henderson-Hasselbalch equation\n",
-    "def ideal_alpha(pH, pK):\n",
-    "    return 1. / (1 + 10**(pK - pH))\n",
+    "Since the system can now be equilibrated, the integration/sampling loop can be written.\n",
     "\n",
+    "**Exercise:**\n",
     "\n",
-    "# empty lists as placeholders for collecting data\n",
-    "numAs_at_each_pH = []  # number of A- species observed at each sample\n",
```
RudolfWeeber, line 339: Please use pep8 naming with underscores between wores for numAs

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
```diff
@@ -364,7 +682,7 @@
    "source": [
     "## Results\n",
     "\n",
-    "Finally we plot our results and compare them to the analytical results obtained from the Henderson-Hasselbalch equation.\n",
```
RudolfWeeber, line 367: Is there a literature ref for the eqn somewhere? If not, please add it.

RudolfWeeber (changes requested):
This is a very nice tutorial. All the steps are explained well.

Please see minor comments inline.

In general, the naming is somewhat inconsistent.
In theory, in pep8, constatns should be ALL_CAPS, variables and functions should be lower_case, with underscoes as word separators. Please check, whether that can be adapted with a reasonable effort. (Some IDEs lik vs code (install the Python extension)) have a "rename symbol" function. Since there a quite a lot of symbols, please leave it, if the effort is unreasonable.

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
jrfinkbeiner: There is a name conflict with the imported module: 
`from espressomd import electrostatics, polymer, reaction_ensemble`

For consistency, I would either change all imports or change the name of the local iteration variable. 

I'll change the bond creation.

KaiSzuttor: it's good practice to use the namespacing of python, *.i.e.* instead of:
``` python
```
you should do:
``` python
```
 that way you can use the module via ```espressomd.electrostatics```, same goes for polymer and reaction ensemble.

RudolfWeeber:
>it's good practice to use the namespacing of python, .i.e. instead of:

>from espressomd import electrostatics

>you should do:

>import espressomd.electrostatics

>that way you can use the module via espressomd.electrostatics, same goes for polymer and reaction ensemble.

I don’t necessarily agree.

Especially if something is used more than once (like reaction ensemble here), I think an extra `espressomd.` in each call hurts readability.

For stuff that’s used just once, I agree with what Kai wrote.

Some degree of personal taste is probably involved.

KaiSzuttor:
> it's good practice to use the namespacing of python, .i.e. instead of: from espressomd import electrostatics you should do: import espressomd.electrostatics that way you can use the module via espressomd.electrostatics, same goes for polymer and reaction ensemble.
> I don’t necessarily agree. Especially if something is used more than once (like reaction ensemble here), I think an extra `espressomd.` in each call hurts readability. For stuff that’s used just once, I agree with what Kai wrote. Some degree of personal taste is probably involved.

In general one should avoid importing individual symbols because it will lead to clashes like here. If you realise that you repeat yourself writing the enclosing module or package 100 times you can still change it. I think going from the default of only importing the package or whole module instead of specific symbols will result in a cleaner code than the doing it the other way around.

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
christophlohrmann, line 1: <p><code>system.integrator.set_steepest_descent(</code>   will does not work with the thermostat added before</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/espressomd/espresso/pull/3917/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='27'/>

reinaual: this has already been fixed but not yet published due to @jrfinkbeiner participating in a meeting

christophlohrmann:
In the tests, you should probably also import the tutorial with ``USE_WCA = True`` and ``USE_ELECTROSTATICS = True``. And then just run a few integration steps to see if the program finishes

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
jngrad: The `USE_WCA` and `USE_ELECTROSTATICS` globals mean we have 3 tutorials in one jupyter notebook, but we have only one CI test.

RudolfWeeber:
For now, please put the configuration (in terms of USE_WCA and friends) into the tutorial test that you whish to do next week.
Once that is done, please let me know. Unless someone raises important concerns, this PR should be merged by tonight.

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
jrfinkbeiner: I think this should be the correct paper:

> Hasselbalch, K. A. (1917). "Die Berechnung der Wasserstoffzahl des Blutes aus der freien und gebundenen Kohlensäure desselben, und die Sauerstoffbindung des Blutes als Funktion der Wasserstoffzahl"

But there is no accessible online version of it.

jonaslandsgesell: To be honest, a link to wikipedia is much more helpful: https://en.wikipedia.org/wiki/Henderson%E2%80%93Hasselbalch_equation you can place links to websites in ipython notebooks. I would also propose to do this in other tutorials.

RudolfWeeber (approved):

doc/tutorials/12-constant_pH/12-constant_pH.ipynb
jonaslandsgesell: For documentation: the acceptance probability for an ideal system (Epot =0) is always exactly 1 in the const pH ensemble for pH=pK. Therefore the Markov Chain after an even amount of steps will always only have even particle numbers and therefore the observed particle numbers are always even. In order to observe all particle numbers fluctuations we perform an odd amount of steps.

PR #3925, opened 2020-09-30
P3M: reordered sanity checks in initialization

### LINKED ISSUE #3920: p3m retuning uses initial state from previous tunings
Reducing the box size and retuning p3m is in some cases not possible because the k-space cutoff is kept from the first tuning. Interestingly, it works when one just tries again.
MNWE:
```python
import numpy as np
from espressomd import System, electrostatics

def place_parts(system, n_part,q):
    for _ in range(n_part//2):
        system.part.add(pos = np.random.random(3)*system.box_l,
                        q = q)
        system.part.add(pos = np.random.random(3)*system.box_l,
                        q = -q)

n_part = 10
q = 1.

box_l = np.array(3*[100])
system = System(box_l = box_l)
system.time_step = 0.01
system.cell_system.skin = 0.4

p3m = electrostatics.P3M(prefactor = 1,
                         accuracy = 1e-2)
place_parts(system, n_part, q)
system.actors.add(p3m)

system.part.clear()
system.box_l = 3*[5]
place_parts(system, n_part, q)

try:
    p3m.tune()
except:
    p3m.tune()
    print('success the second time')
```

reinaual:
Fixes #3920 

Description of changes:
- introduced `p3m_sanity_checks_parameters`
- reordered sanity checks in `p3m_init` to update `cao_cut` before it gets checked

### DIFF
--- a/src/core/electrostatics_magnetostatics/p3m.cpp
+++ b/src/core/electrostatics_magnetostatics/p3m.cpp
@@ -80,6 +80,10 @@ static bool p3m_sanity_checks_system(const Utils::Vector3i &grid);
  */
 static bool p3m_sanity_checks_boxl();
 
+/** Checks that necessary parameters are set
+ */
+static bool p3m_sanity_checks_parameters();
+
 /** Calculate the optimal influence function of @cite hockney88a.
  *  (optimised for force calculations)
  *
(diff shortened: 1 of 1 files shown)


KaiSzuttor (approved):
this whole logic is hard to read, i had to check the source code to find out that ```true``` actually means that something went wrong during the validation. This seems to be a refactor candidate for the near future since it's kind of a workhorse of espresso.

PR #3938, opened 2020-10-06
Config: Include WCA in default myconfig.hpp

RudolfWeeber:


jngrad:
@RudolfWeeber WCA was already defined in this file at line 79. The recent failures in CI come from the empty config file, iirc.

RudolfWeeber:
Ah, OK, I got confused.

PR #3960, opened 2020-10-20
New shape: Quarterpipe

christophlohrmann:
Description of changes:
- introduce a new shape to smooth out the intersection between perpendicular surfaces


KaiSzuttor:
Are you sure that this is needed since the addition of Union shapes?

christophlohrmann:
I am pretty sure I cannot combine a ``rhomboid(direction = 1)`` with a ``cylinder(direction = -1)`` to achieve the desired effect. What I would get is a cylinder pipe with a box sticking into it.

doc/sphinx/constraints.rst
```diff
@@ -450,6 +451,26 @@ Note: in the OpenGL visualizer, if the OpenGL Extrusion library is not available
...
 
 
+SmoothCorner
+""""""""""""""""""""
+
+:class:`espressomd.shapes.SmoothCorner`
+
+A smoothed concave corner, which can be used to connect two walls or rhomboids in a smooth way.
+The shape is constructed by removing a quarter cylinder of radius ``radius``
+from a cuboid with side length ``radius``. It can be created via ::
```
KaiSzuttor: i think you need to add a sketch to make this clear

christophlohrmann:
I realize that the ``direction`` parameter is not a good name, since it is used with a different meaning in other shapes. Any suggestions?

jngrad:
@christophlohrmann you'll have to merge the python branch into the PR to pass CI, all 3 runners are failing (https://github.com/espressomd/espresso/issues/3895#issuecomment-715242963).

KaiSzuttor:
@RudolfWeeber

src/shapes/src/SmoothCorner.cpp
```diff
@@ -0,0 +1,138 @@
...
+    vec = -dist * m_direction;
+  } else {
+    // go to cylindrical coordinates (r,delta_phi) where m_axis is the z-axis
+    // and  m_direction defines the delta_phi = 0 axis
+    auto x = pos_projected * m_direction;
+    auto y = pos_projected * Utils::vector_product(m_axis, m_direction);
+    auto pos_delta_phi = std::atan2(y, x);
+    while (pos_delta_phi > 2 * Utils::pi()) {
+      pos_delta_phi -= Utils::pi();
+    }
```
reinaual: Why do we need to do this? According to [std::atan2](https://en.cppreference.com/w/cpp/numeric/math/atan2) the return value should be between [-pi, pi] if nothing goes wrong. If it is necessary, is [std::fmod](https://en.cppreference.com/w/cpp/numeric/math/fmod) what you are looking for?

src/shapes/src/SmoothCorner.cpp
```diff
@@ -0,0 +1,138 @@
...
+  auto upper_lower = rel_z > 0 ? 1 : -1;
+  auto dist_top_bottom = Utils::abs(rel_z) - 0.5 * m_height;
+
+  // if inside the shape, check if the bottom or top wall are closer
+  if (dist < 0. and dist_top_bottom < 0) {
+    if (Utils::abs(dist_top_bottom) < Utils::abs(dist)) {
+      dist = dist_top_bottom;
+      vec = upper_lower * dist * m_axis;
+    }
+  }
```
reinaual: This expression can be evaluated by two comparisons
```suggestion
  if (dist_top_bottom < 0 and dist_top_bottom > dist) {
    dist = dist_top_bottom;
    vec = upper_lower * dist * m_axis;
  }
```

src/shapes/src/SmoothCorner.cpp
christophlohrmann: Thanks for pointing that out, I thought the result would be in [0,2 pi]. I will remove the loop

RudolfWeeber:
If I understand the description correctly:
* this should be called edge rather than corner, or even QuaterPipe or Quatercylinder
* The `direction` parameter might be called `long_axis`
* the `axis` parameter `bisector` (English for "Wineklhalbierende", according to Google)

RudolfWeeber:
Is a review of the actual code still needed, or did someone do that already?

christophlohrmann:
``QuaterPipe`` would be my favourite. I still like ``axis`` because it has the same meaning as for the ``Cylinder`` shape. ``bisector`` seems indeed be better than direction

reinaual:
> Is a review of the actual code still needed, or did someone do that already?

I have not reviewed it, i just pointed out the things which i found by just briefly looking at the code.

src/shapes/src/SmoothCorner.cpp
```diff
@@ -0,0 +1,138 @@
...
+    auto to_point_1 =
+        m_radius * vec_rotate(m_axis, -0.5 * m_delta_phi, m_direction);
+    auto to_point_2 =
+        m_radius * vec_rotate(m_axis, 0.5 * m_delta_phi, m_direction);
+
+    if (pos_delta_phi <= -0.5 * m_delta_phi) {
+      vec = pos_projected - to_point_1;
+      dist = vec.norm();
+    } else {
+      if (pos_delta_phi >= 0.5 * m_delta_phi) {
```
reinaual: i guess this can be brought to the outer layer with an `else if`

KaiSzuttor:
> > Is a review of the actual code still needed, or did someone do that already?
> 
> 
> 
> I have not reviewed it, i just pointed out the things which i found by just briefly looking at the code.

i think you just qualified for a full review 👍🏻😜

src/shapes/src/SmoothCorner.cpp
```diff
@@ -0,0 +1,138 @@
...
+ *
+ * ESPResSo is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+#include <boost/algorithm/clamp.hpp>
```
reinaual: I see where this is coming from but you don't need this.
```suggestion
```

doc/sphinx/constraints.rst
```diff
@@ -450,6 +451,36 @@ Note: in the OpenGL visualizer, if the OpenGL Extrusion library is not available
...
+A cross section with an explanation of the parameters is shown in the figure below.
+The three-dimensional shape is an extrusion of this profile.
+
+.. figure:: figures/quarterpipe.png
+   :alt: Cross section of the quarterpipe shape.
+   :height: 4.00000cm
+
+``center`` defines the  center of the cylinder. 
+``axis`` is the axis of the cylinder,
+orthogonal to the curve of the pipe.
```
reinaual: this linebreak looks odd to me because the line below is even longer
```suggestion
``axis`` is the axis of the cylinder, orthogonal to the curve of the pipe.
```

src/shapes/src/Quarterpipe.cpp
```diff
@@ -0,0 +1,146 @@
...
+          dist = in_zone_7 ? dist_plane_1 : dist_plane_2;
+          vec = in_zone_7 ? vec_plane_1 : vec_plane_2;
+        }
+      }
+    }
+  } // 2D end, now to the 3rd dimension
+
+  auto upper_lower = rel_z > 0 ? 1 : -1;
+  // signed distance to the closer top/bottom wall
+  auto dist_cover = Utils::abs(rel_z) - 0.5 * m_height;
```
reinaual: you could replace the `Utils:abs` call which you have also done in all cases below
```suggestion
  auto dist_cover = upper_lower * rel_z - 0.5 * m_height;
```

src/shapes/src/Quarterpipe.cpp
```diff
@@ -0,0 +1,146 @@
...
+          auto corner_point = std::sqrt(2.) * m_radius * m_orientation;
+          vec = pos_projected - corner_point;
+          dist = vec.norm();
+        } else if (dist_plane_1 < 0. and dist_plane_2 < 0.) {
+          // Zone 4, 5 or 6
+
+          // First, find which side is closer
+          // Distinguishes zones 4 and 5 from 4 and 6
+          bool in_zone_4_or_5 =
+              Utils::abs(dist_plane_1) < Utils::abs(dist_plane_2);
```
reinaual: we already know the sign from the `else if` above...
```suggestion
          bool in_zone_4_or_5 = dist_plane_1 > dist_plane_2;
```

src/shapes/src/Quarterpipe.cpp
```diff
@@ -0,0 +1,146 @@
...
+    if (pos_delta_phi <= -0.5 * m_delta_phi) {
+      // Zone 1
+      vec = pos_projected - to_point_1;
+      dist = vec.norm();
+    } else if (pos_delta_phi >= 0.5 * m_delta_phi) {
+      // Zone 2
+      vec = pos_projected - to_point_2;
+      dist = vec.norm();
+    } else {
+      if (pos_norm < m_radius) {
```
reinaual: this can also be brought one layer out as an `else if`

KaiSzuttor (changes requested):
i don't understand the sketch at all :D

src/shapes/src/Quarterpipe.cpp
```diff
@@ -0,0 +1,142 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <shapes/Quarterpipe.hpp>
+#include <utils/Vector.hpp>
```
reinaual, line 21:
```suggestion
#include <utils/Vector.hpp>
#include <utils/constants.hpp>
```

src/shapes/src/Quarterpipe.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+      // Zone 4, 6, 8 or 9
+      auto dist_plane_2 =
+          (pos_projected * to_point_2 / to_point_2.norm()) - to_point_2.norm();
+      auto vec_plane_2 = dist_plane_2 * to_point_2 / to_point_2.norm();
+
+      // Distinguish the Zones
+      if (dist_plane_1 >= 0. and dist_plane_2 >= 0.) {
+        // Zone 9
+        auto corner_point = std::sqrt(2.) * m_radius * m_orientation;
```
reinaual:
```suggestion
        auto corner_point = Utils::sqrt_2() * m_radius * m_orientation;
```

samples/visualization_constraints.py
```diff
@@ -107,6 +107,14 @@
                                       direction=1, radius=15, tube_radius=6),
         particle_type=0, penetrable=True)
 
+elif args.shape == "Quarterpipe":
+    system.constraints.add(
+        shape=espressomd.shapes.Quarterpipe(center=[25, 15, 15], axis=[1, 0, 0],
+                                            orientation=[0, 1, 1], radius=25,
+                                            height=30),
+        particle_type=0, penetrable=True)
```
jngrad, line 115: * I'm not sure it is a good idea to have two sharp edges in a shape interacting with a LJ liquid; wouldn't it be safer to have e.g. 4 quaterpipes to form a pipe, or a quaterpipe with its X,Y,Z edge lengths equal to the box lengths? 
* the rasterization doesn't draw dots on the shape top and bottom surfaces, giving the impression the shape is hollow.

christophlohrmann: the shape becomes unsafe when side walls touch the periodic boundaries. Then Particles can enter the interaction region from the other side and pick up a lot of potential energy that propels them through the box. I will change the ``height`` though to hide the top/bottom

RudolfWeeber:
What is the status? Is there a conclusion about the sketch?

KaiSzuttor:
> What is the status? Is there a conclusion about the sketch?

review in progress

KaiSzuttor:
decided to extend the existing HollowConicalFrustum by a orientation and central angle argument.

RudolfWeeber:
*	`decided to extend the existing HollowConicalFrustum by a orientation and central angle argument.`

How can a Kegelstumpf be made a Quaterpipe?

KaiSzuttor:
the cone with 0 conicity is a finite cylinder. if you further allow to specify the central angle you end up with a quarter pipe

RudolfWeeber:
OK. That such a creative re-use is possible would have to be made clear in the documentation including an image, I think.

The name of the shape doesn’t really suggest it. In theory, a new name would also be possible, since the shape was not in a release yet.

KaiSzuttor:
it's just a segment of the existing shape, thats not too far fetched

RudolfWeeber:
See #4014

KaiSzuttor:
this will be implemented by recycling the ```HollowConicalFrustum``` shape

Issue #3019, opened 2019-07-25
Bond_parameters should be a variant


fweik:
This is currently a tagged union, which requires manual construction and destruction when changing the active type (the union contains non-trivial types). This requires boiler plate code for switching the active type and is error prone. This can be made automatic by turning the union
into a `boost::variant`.

fweik:
Is anybody working on this?

jngrad:
I'm not aware of anyone working on this.

jngrad:
I gave it a try yesterday, but the refactoring effort is quite significant because the kernels are written as free functions. Today I tried to break it down into smaller tasks, and came up with the following plan:

* make bond parameter structs serializable (easy)
    * bond communication can then be simplified substantially
* move the force/energy kernels into the corresponding structs (difficult)
    * dependency inversion required in `bonded_interaction_data.hpp`: the kernels are currently separated from the corresponding structs so that the short range loop has access to the cutoff method but not to the kernels (which bring in too many dependencies)
* make kernel signatures more homogeneous
    * pass positions instead of particles in `IBM_*_CalcForce()`
    * not sure what to do about `bonded_coulomb_pair_{force,energy}()` and `calc_oif_local()`
* create visitor patterns for energy and force kernels

To illustrate this strategy, a small C++ prototype is attached ([variant.cpp](https://github.com/espressomd/espresso/files/5754852/variant.cpp.txt)). The key part is this line:
```c++
// calculate the bonded energy for a distance vector `vec` and bond `bond_id`
auto energy = boost::apply_visitor([&vec](auto const &ia){ return ia.energy(vec); },
                                   bonded_ia_params[bond_id]);
```

The difficulty here is that the number of arguments in the kernels is different for pair, angle and dihedral bonds, and further vary for OIF and coulomb kernels, yet the visitor pattern requires a homogeneous function signature. To work around that, we could define in every bond pair an `energy(v1, v2) {return 0;}` overload and in every bond angle an `energy(v) {return 0;}` overload (see [variant_angle.cpp](https://github.com/espressomd/espresso/files/5754854/variant_angle.cpp.txt)), but this is not a clean solution, and we definitively don't want the OIF and Coulomb kernel signatures to bleed into all the other bonds.

Notes for the compiler explorer: the macro `INLINE_KERNEL` is commented out to make the function calls stand out in the generated assembly. If you use Clang 11 instead of GCC 10.2, make sure to enable Boost 1.71+ to get less boilerplate instructions.

[...]

PR #3961, opened 2020-10-20
Fix tabulated bond memory leaks

### LINKED ISSUE #3906: Memory leaks in tabulated bonds
The head node callback `mpi_bcast_ia_params()`and its worker node callback `mpi_bcast_ia_params_slave()` are mostly identical and could be merged into a single callback, with a conditional on the MPI rank to manage manual memory allocation on worker nodes. This callback could then be moved from `communication.cpp` to e.g. `src/core/bonded_interactions/bonded_interaction_data.cpp`.

Right now the worker implementation allocates a pointer with `new` and overwrites the old pointer without a proper `delete`, leading to a memory leak every time a tabulated bond is updated, when running with 2 MPI ranks or more. Here is the relevant code:

https://github.com/espressomd/espresso/blob/f760050199b94226b9127b309bfa7db569507605/src/core/communication.cpp#L336-L345

Adding `delete` at L344 before changing the pointer doesn't solve the issue, because the MPI communication at L336 already overwrote the address of the allocated memory on the worker node with the address to the allocated memory on the head node.

Solving this requires adding a `delete` before the communication at L336, although it would be better to write a proper serialization method with `boost::serialization` to avoid communicating pointer members. The serialization can also take care of packing the vectors. In fact, there is already a `TabulatedPotential::serialization()` method, so it's just a matter of calling it from the bonded interactions code.

Here is a MWE to trigger the memory leak:
```python
import numpy as np
import espressomd
from espressomd.interactions import TabulatedDistance
import time

system.time_step = 0.01
system.cell_system.skin = 0.4

a = np.arange(4000000)
print('array size: {:.0f} MB'.format(a.size * a.itemsize / 1024**2))

print('start allocation')
system.bonded_inter.add(TabulatedDistance(min=0.1, max=2, energy=a, force=a))
for i in range(10):
    system.bonded_inter[0] = TabulatedDistance(min=0.1, max=2, energy=a, force=a)
    time.sleep(0.1)

print('stop allocation')
time.sleep(5)
```
```sh
mpiexec -n 2 ./pypresso mwe.py
```

Takes 1400 MB of RAM on 2fbea6e967. After adding a `delete`, takes "only" 900 MB of RAM, which is still larger than the expected 200 MB, so there might be another memory leak... possibly somewhere around the `TabulatedPotential` code, because the non-bonded tabulated interaction also consumes a lot of RAM when its parameters are updated in a loop. This second memory leak also has a surprising garbage collection behavior: memory seems to be allocated for a new container before the previous container was deallocated, which shows up in `htop` as RAM bouncing back and forth between two values after each loop iteration. That's why there is this `sleep(5)` statement at the end of the MWE, to get the final 900 MB of RAM consumption (peak RAM consumption is 1050 MB).

jngrad:
Fixes #3906

Description of changes:
- fix dangling pointers in tabulated bonds
- reduce code duplication in callbacks


src/core/communication.cpp
```diff
@@ -335,14 +335,15 @@ void mpi_bcast_ia_params_slave(int i, int j) {
...
+    }
     MPI_Bcast(&(bonded_ia_params[i]), sizeof(Bonded_ia_parameters), MPI_BYTE, 0,
               comm_cart);
     /* For tabulated potentials we have to send the tables extra */
     if (is_tabulated_bond(bonded_ia_params[i].type)) {
-      auto *tab_pot = new TabulatedPotential();
-      boost::mpi::broadcast(comm_cart, *tab_pot, 0);
-
-      bonded_ia_params[i].p.tab.pot = tab_pot;
+      bonded_ia_params[i].p.tab.pot = new TabulatedPotential();
```
KaiSzuttor: haven't looked into the code, but can't we use smart pointers?

jngrad: I don't think so: we overwrite the memory block of the bond object on worker nodes with a bytestring of the same object from the head node, so the counter from a `shared_ptr` would be overwritten. As for `unique_ptr`, in the line you highlighted, we do an assignment, so the smart pointer goes out of scope and gets garbage-collected... but it contains a random memory address (because of the bytestring deserialization a few lines above), which is a recipe for segmentation faults.

KaiSzuttor: i'm not sure i'm following... but it sounds like this is a place for larger scale refactoring which is out of scope of this pull request?

jngrad: the correct way would be to write boost serialization methods, which can handle pointer members out of the box, e.g.
```c++
struct Tabulated_bond_parameters {
  int type;
  TabulatedPotential *pot;

private:
  friend boost::serialization::access;
  template <typename Archive>
  void serialize(Archive &ar, long int /* version */) {
    ar &type;
    ar &pot;
  }
};
```
but there is no guarantee that we are overwriting a bond type with the same bond type, so we would have to add destructors too... Not sure if this is worth the effort, if our long-term goal is to replace this interaction communication infrastructure with a python setter.

jngrad: Also, a much more promising refactoring goal is breaking the dependency on `MpiCallbacks.hpp` from consumers of MPI callbacks ([fix-3906...fix-3907-p1](https://github.com/jngrad/espresso/compare/fix-3906...fix-3907-p1)), which as a side effect is removing global variables and removing prototypes of MPI worker functions from header files.

KaiSzuttor: > Also, a much more promising refactoring goal is breaking the dependency on `MpiCallbacks.hpp` from consumers of MPI callbacks ([fix-3906...fix-3907-p1](https://github.com/jngrad/espresso/compare/fix-3906...fix-3907-p1)), which as a side effect is removing global variables and removing prototypes of MPI worker functions from header files.

Isn't that totally orthogonal to the memory leak issue?

KaiSzuttor: > the correct way would be to write boost serialization methods, which can handle pointer members out of the box, e.g.
> 
> ```c++
> struct Tabulated_bond_parameters {
>   int type;
>   TabulatedPotential *pot;
> 
> private:
>   friend boost::serialization::access;
>   template <typename Archive>
>   void serialize(Archive &ar, long int /* version */) {
>     ar &type;
>     ar &pot;
>   }
> };
> ```
> 
> but there is no guarantee that we are overwriting a bond type with the same bond type, so we would have to add destructors too... Not sure if this is worth the effort, if our long-term goal is to replace this interaction communication infrastructure with a python setter.

sorry, if that has been discussed and i don't remember it anymore... Do we have some design plans for the interaction stuff already?

RudolfWeeber:
>sorry, if that has been discussed and i don't remember it anymore... Do we have some design plans for the interaction stuff already?

My understanding is that

*	Actors, bonded and non-bonded interactions should be used via the script-interface, which handles the broadcasting of parameters, so the comm infrastructure for it in the core goes away
*	The BondedInteractionParams type should be a variant rather than a union
*	Since the communication is handled by the script interface, bonds are set in the core on the node level by assignment. As long as the strucs holding the parameters for the individual bodns either don’t contain manual memory management or provide a destructor, it should not be possible to loose memory anymore.

KaiSzuttor:
so think we should now merge this but also stop refactoring the existing infrastructure if the plan is to move most of the parts to the sciptinterface.

src/core/bonded_interactions/bonded_tab.cpp
```diff
@@ -37,6 +37,11 @@ int tabulated_bonded_set_params(int bond_type,
   make_bond_type_exist(bond_type);
 
   /* set types */
+  if (bonded_ia_params[bond_type].type == BONDED_IA_TABULATED_DISTANCE or
+      bonded_ia_params[bond_type].type == BONDED_IA_TABULATED_ANGLE or
+      bonded_ia_params[bond_type].type == BONDED_IA_TABULATED_DIHEDRAL) {
+    delete bonded_ia_params[bond_type].p.tab.pot;
```
KaiSzuttor, line 43: cant't you just check for ```nullptr```?

KaiSzuttor:
``` c++
if (bonded_ia_params[bond_type].p.tab.pot) delete bonded_ia_params[bond_type].p.tab.pot;
```

fweik:
IMHO, #3019 should be fixed before this. This correct way to solve this is to have the tabulated bond manage it's own memory (via unique ptr or similar), but this is not possible as long as the bond parameters is a `union`, because you can only have trivial types in a union. This is why I opened 3019 in the first place. There is no way to fix the memory leak which is not a hack in the sense that it mixes unrelated concerns.

src/core/bonded_interactions/bonded_tab.cpp
jngrad: Actually this would raise a Clang-Tidy warning because you can simplify it to
```c++
delete bonded_ia_params[bond_type].p.tab.pot;
```
which is safe on a `nullptr`.
However, this would also trigger a segfault, because you can create a tabulated bond on top of an existing FENE bond, in which case `.p.tab.pot` contains a random address. You can overwrite an bond existing bond in ESPResSo, this is used for example in `testsuite/python/interactions_bonded_interface.py`.

RudolfWeeber (approved):

RudolfWeeber:
Let's merge this, also due to the comment fixes contained in the PR.
The "fix" for the mem leak is, of course, a hack.
It will go away, once we convert the bond parameters to a Variant and let the script interface do the parameter broadcasting.

PR #3966, opened 2020-10-22
Remove ROCm integration

### LINKED ISSUE #2973: HIP issue list as discussed in the offline meeting
* https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/134161 (connected to #2878)
  unreadable error message at runtime

### LINKED ISSUE #3895: DipolarBarnesHutGpu and DipolarDirectSumGpu tests fail on ROCm
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/pipelines/13105

jngrad:
Closes #2973, closes #3895, follow-up to espressomd/docker#190

Due to their fast release cycle, ROCm packages are not stable enough for ESPResSo. We are currently supporting ROCm 3.0 to 3.7, which means supporting two compilers (HCC and HIP-Clang) and keeping patches for each ROCm release in our codebase.  Maintaining these patches and the CMake build system for ROCm is time-consuming. The ROCm job also has a tendency to break the CI pipeline (#2973), sometimes due to random or irreproducible software bugs in ROCm, sometimes due to failing hardware in both the main CI runner and backup CI runner. The frequency of false positives in CI is too large compared to the number of true positives. The last true positives were 5da80a92d (April 2020) and https://github.com/espressomd/espresso/issues/2973#issuecomment-514108007 (July 2019). There are no known users of ESPResSo on AMD GPUs according to the [May 2020 user survey](https://lists.nongnu.org/archive/html/espressomd-users/2020-05/msg00001.html). The core team has decided to drop support for ROCm ([ESPResSo meeting 2020-10-20](https://github.com/espressomd/espresso/wiki/Espresso-meeting-2020-10-20)).

The Intel image cannot be built automatically in the espressomd/docker pipeline. Re-building it manually is a time-consuming procedure, usually several hours, due to long response time from the licensing server and the size of the Parallel Studio XE source code. When a new Intel compiler version is available, it usually requires the expertise of two people to update the dockerfile. The core team has decided to remove the Intel job from CI and use the Fedora job to test ESPResSo with MPICH.


src/core/EspressoSystemInterface_cuda.cu
```diff
@@ -17,7 +17,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-#include "cuda_wrapper.hpp"
```
KaiSzuttor, line 20: i'm not sure if we need to remove the wrapper... this can probably stay

src/core/cuda_wrapper.hpp
```diff
@@ -1,89 +0,0 @@
...
- *
- * You should have received a copy of the GNU General Public License
- * along with this program.  If not, see <http://www.gnu.org/licenses/>.
- */
-#ifndef CUDA_WRAPPER_HPP
-#define CUDA_WRAPPER_HPP
-
-#if defined(__CUDACC__)
-
-#include <cuda.h>
```
jngrad, line 24: @KaiSzuttor I'm replying here because your comment doesn't show up in the GitHub interface.

The `cuda_wrapper.hpp` only includes `cuda.h` once the HIP macros are removed. Do you want to keep it so that we can add CUDA macros in the future? Should we also keep the cufft and curand wrappers?

KaiSzuttor: i changed my mind :) it's probably better to remove all of those wrappers.

KaiSzuttor (approved):

PR #3971, opened 2020-10-25
Better separation of concerns in the core

### LINKED ISSUE #3907: Better separation of concerns in the communication framework
**Background**

ESPResSo uses a custom MPI communication framework where everything is managed from the head node. Functions visible in the script interface must be implemented twice: once as a worker function, and once as a head function that contains the same body as the worker function, plus a preamble that triggers the worker function on worker nodes with `mpi_call_*()`. The body usually contains `boost::mpi` and `MPI_*` functions to synchronize data. Often, the body can be implemented only once in a separate free function. These head and worker functions (aka "callbacks") are managed by our `MpiCallbacks` framework which executes them via a visitor pattern.

**Problem statement**

The source file `communication.cpp` is the central place to register callback functions, which should be declared and defined in the relevant cpp resp. hpp files. However for historical reasons, many callbacks have been written directly in `communication.cpp`, which means this file has to know about the implementation details of a variety of classes/structs. This prevents us from hiding implementation details of these classes/structs (e.g. making members private, replacing manual memory management by standard containers, etc.) and can lead to code duplication when the body of the worker function is copy-pasted into the body of the head function. Duplicated code will diverge over time.

**Prior work**

Between 4.1 and 4.3 (waLBerla branch), many callbacks were moved to dedicated source files. For example the NpT callback was moved to `npt.cpp` (d21c7a4e35eb), because `communication.cpp` doesn't need to know the internal structure of the NpT parameters class, it only needs to know that a callback function exists, which is visible via a function prototype in `npt.hpp`. The steepest descent parameters object went one step further and has static storage, making it inaccessible outside the source file:
https://github.com/espressomd/espresso/blob/f760050199b94226b9127b309bfa7db569507605/src/core/integrators/steepest_descent.cpp#L40-L41

**Course of action**

Here are a few candidates that could be simplified further or moved to dedicated files:
* `mpi_bcast_ia_params()` which is written twice, and has now diverged: the worker implementation allocates a pointer with `new` and overwrites the old pointer without a `delete`, leading to a memory leak. Adding `delete` before the `new` doesn't work, because the MPI communication step before that overwrote the address of the allocated memory (see #3906 for details).
* `mpi_bcast_coulomb_params_slave()` needs to know about the Coulomb and dipolar parameter structs. It should rely on `boost::serialization` instead, and have the callbacks written in the relevant electrostatics .cpp files.
* `mpi_gather_stats()` mixes pressure/energy observables with LB communication. The LB communication logic was removed in 4.3. This can be simplified further in 4.3 by replacing the `enum class GatherStats` and the switch statement by a 0/1 integer and a conditional, and removing code duplication between the head and worker functions.

**Challenges**

Moving the remaining callbacks from `communication.cpp` to dedicated .cpp/.hpp files means we have to rely on the `comm_cart` global variable. Passing this global by reference means the dedicated .hpp files that declare callbacks must include MPI headers. However, some parts of the CUDA code are not allowed to include MPI header files, therefore features that are used by CUDA code and rely on MPI communication must either use the `comm_cart` global variable via `extern`, or rely on a second header file, e.g. `npt_mpi.hpp`, to write the prototype of the callback.

**Outlook**

This refactoring will facilitate hiding implementation details of parameter structs and help finding new ways to simplify the callback infrastructure. For example, #3891 removed one level of MPI indirection by having `mpi_steepest_descent()` call `integrate()` via `mpi_steepest_descent_slave()` instead of calling a chain of callbacks. In fact, `mpi_steepest_descent()` could probably call `mpi_integrate_slave()` directly instead of `mpi_steepest_descent_slave()`, removing yet another MPI indirection. This refactoring will also help on the long term if we choose to move in the direction of MPI+OpenMP or pure OpenMP.

jngrad:
Fixes #3907

Move callbacks from `communication.cpp` to the relevant source files. This hides implementation details of many submodules, which no longer need to expose local functions in the public interface with doxygen blocks stating for example "Used by \ref mpi_place_particle, should not be used elsewhere". This also reduces the visibility of `MpiCallbacks.hpp`, `communication.hpp`, `Particle.hpp` and `cuda_init.hpp`. When the head function and worker function have the same code, remove code duplication using `mpi_call_all()`. Remove the `CALLBACK_LIST` macro and split `mpi_gather_stats()` into independent callbacks.


jngrad:
This PR moved the callbacks to narrow interfaces, but didn't attempt to rewrite these callbacks, except for `mpi_gather_stats()` (to disentangle LB from non-LB code) and `calc_mu_max()` (to remove a static global).

There is room for more refactoring on the long term, e.g.:
- merging callbacks `*_slave()`/`*_local()` into the free function they call, or making these free functions static
- refactor callbacks that communicate structs by replacing MPI primitives with `boost::mpi` and `boost::serialize`, but only for callbacks that are not candidates for removal via the Python script interface
- remove integrators struct members from `global.hpp` and communicate them via `boost::mpi` and `boost::serialize`, then move integrators structs to .cpp files to hide implementation details (see steepest descent for a good example)

The other ~100 callbacks that were not originally in `communication.cpp` weren't touched by this PR.

reinaual (dismissed):
lgtm

src/core/electrostatics_magnetostatics/common.hpp
```diff
@@ -0,0 +1,21 @@
+/*
```
reinaual, line 1: i guess we want an header guard [LGTM](https://lgtm.com/projects/g/espressomd/espresso/rev/pr-22060e5ec6266b3319a9a9ed0f7c3443972b76ef)

jngrad: indeed, I'll add one

reinaual (dismissed):

RudolfWeeber (dismissed):

KaiSzuttor (approved):

PR #3976, opened 2020-10-26
Fix -Wfloat-conversion

jngrad:
Port improvement from 4.3.

### DIFF
--- a/src/core/actor/DipolarBarnesHut_cuda.cu
+++ b/src/core/actor/DipolarBarnesHut_cuda.cu
@@ -250,7 +250,7 @@ __global__ __launch_bounds__(THREADS2, FACTOR2) void treeBuildingKernel() {
       j = 0;
       for (l = 0; l < 3; l++)
         if (root[l] < p[l])
-          j += pow(2, l);
+          j += static_cast<int>(pow(2, l));
     }
 
     // Follow path to leaf cell. Should not happen at the first iteration of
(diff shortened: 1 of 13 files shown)


RudolfWeeber (approved):

RudolfWeeber:
The are a few instances of
`static_cast<int>(pow(n,m))`
where n and m seem to be integer as well. IIRC, there is an int_pow somewhere in Utils.
If the sites are not performance-critical, you can leave them as they are.

src/core/actor/DipolarBarnesHut_cuda.cu
```diff
@@ -250,7 +250,7 @@ __global__ __launch_bounds__(THREADS2, FACTOR2) void treeBuildingKernel() {
       j = 0;
       for (l = 0; l < 3; l++)
         if (root[l] < p[l])
-          j += pow(2, l);
+          j += static_cast<int>(pow(2, l));
```
jngrad, line 253: @RudolfWeeber if you referred to these integer powers in the dipolar code, `int_pow<l>()` will not work since the power is not constexpr. Could be a candidate for bit shifting, though, but I don't think this is performance critical because there is an `if()` right above it, which is not a good idea in a CUDA kernel.

src/utils/include/utils/sampling.hpp
```diff
@@ -55,7 +55,8 @@ get_cylindrical_sampling_positions(std::pair<double, double> const &r_limits,
   auto const smallest_bin_volume =
       pi() * pow(r_limits.first + delta_r, 2.0) * delta_phi / (2.0 * pi());
```
jngrad, line 56: this would be a candidate for `Utils::sqr()`, but it's only called in the constructor of profile observables as far as I can tell, so not performance critical

PR #3983, opened 2020-11-03
Ferrofluid tutorial

### LINKED ISSUE #3942: Issue tracker for Tutorial 11
* [x] the HTML versions fail to run due to errors in lowercase/uppercase variable names
* [x] symbols in formula are not always consistent with symbols in the text (e.g. `\varphi` instead of `\phi`)
* [x] the formulation "we commit H.mu as the external magnetic field" is used in the sense of "we apply H.mu as...", and "we commit the orientation of the dipole moments" in the sense of "we set the orientation..."
* [ ] in part 3, check if it would be possible to calculate error bars for the plots

pkreissl:
Fixes #3942

Description of changes:
- minor changes: wording, use `numpy` for simplification


doc/tutorials/11-ferrofluid/11-ferrofluid_part2.ipynb
```diff
@@ -191,7 +191,7 @@
     "    B = \\mu_0 H = \\frac{\\alpha~k_{\\text{B}}T}{\\mu}\n",
     "\\end{equation}\n",
     "\n",
-    "As only the current orientation of the dipole moments, i.e. the unit vector of the dipole moments, is saved in the particle list but not their strength we have to commit $B\\cdot \\mu$ as the external magnetic field to **ESPResSo**.\n",
+    "As only the current orientation of the dipole moments, i.e. the unit vector of the dipole moments, is saved in the particle list but not their magnitude, we have to use $B\\cdot \\mu$ as the strenght of the external magnetic field.\n",
```
jngrad:
```suggestion
    "As only the current orientation of the dipole moments, i.e. the unit vector of the dipole moments, is saved in the particle list but not their magnitude, we have to use $B\\cdot \\mu$ as the strength of the external magnetic field.\n",
```

jngrad:
@pkreissl I think I found the regression in the code, I'll push a solution.

jngrad (approved):

Issue #3883, opened 2020-09-02
Framework for statistical tests


jngrad:
Currently we have frameworks for unit testing, integration testing and sample testing. They are all deterministic, either because the functionality being tested is deterministic, or RNG seeds (both in numpy and espresso) have been set such that the functionality is deterministic. We currently do not have a suitable framework for testing functionality that is stochastic in nature and for which a statistic (e.g. the mean) is known to converge for long simulation times, forcing us to use workarounds, e.g. disable tests in CMake for specific build types, increase test timeout in CI, or reduce the number of integration loops at the expense of accuracy.

**Problem statement**
Several python integration tests are actually not integration tests, but stochastic tests. They do not contribute to code coverage and do not benefit from C++ static assertions and sanitizers, since the underlying functionality already has integration tests. Yet they suffer from significant slowdowns (2-5 times larger runtime) in coverage and sanitizer builds. The Clang sanitizer CI job for example has crossed the 1 hour runtime threshold on multiple occasions over the last few months, prompting us to refactor slow tests to make them run faster at the expense of larger tolerance values in the checks, as well as refactor *unrelated* python tests to make them run faster and gain a few minutes to compensate for slow tests. Statistical tests also tend to undergo large runtime fluctuations on coyote runners that are under load from multiple CI jobs.

Although we should always prefer deterministic tests over statistical tests, there is a need for statistical tests that will not go away. Here are a few candidates for a statistical test framework (times in seconds; `RelWithAssert` builds are compiled with `-O3 -g`):

| Test name | [RelWithAssert](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/250628) | [Coverage](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/250618) | [Sanitizers](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/250626) | Remarks |
| --- | --- | --- | --- | --- |
`langevin_thermostat` | 12.47 | 46.44 | 82.87 | |
`brownian_dynamics` | 6.57 | 14.24 | 23.92 | |
`dpd` | 8.28 | 31.25 | 43.49 | |
`integrator_npt` | 15.65 | 77.96 | 122.07 | |
`lb` | 10.76 | 21.44 | 47.73 | mass & momentum conservation |
`lb_pressure_tensor_acf` | 21.60 | 24.73 | 41.54 | CPU LB is untested (too slow) |
[...]

PR #3988, opened 2020-11-06
Coverage report of scripts

jngrad:
Separate C++ coverage from python coverage. This way, statistical tests (#3883) can appear in the coverage report even though they are excluded from C++ coverage builds, and tutorial/sample tests can also appear in the coverage report.


jngrad:
The coverage file summary shows the testsuite/script files have high coverage, but when one clicks on them, it doesn't show the number of hits. See e.g. [3988/tree/testsuite/scripts/importlib_wrapper.py](https://app.codecov.io/gh/espressomd/espresso/compare/3988/tree/testsuite/scripts/importlib_wrapper.py) where hit lines are not highlighted while uncovered lines are red. This seems to be the default behavior for line only hit once. See e.g. [3809/tree/testsuite/python/dds-and-bh-gpu.py](https://app.codecov.io/gh/espressomd/espresso/compare/3809/tree/testsuite/python/dds-and-bh-gpu.py) where the GPU check was only visited once. In the coverage summary of the python branch, this part of the file has a coverage count if the nightly builds have run at least once (coverage reports accumulate).

KaiSzuttor (approved):

PR #3990, opened 2020-11-11
Make header files self-contained

### LINKED ISSUE #3905: Some tests fail on Fedora Rawhide
On Fedora Rawhide:
```
The following tests FAILED:
	 23 - coulomb_cloud_wall (Failed)
	 71 - domain_decomposition (Failed)
	 83 - coulomb_cloud_wall_duplicated (Failed)
	109 - cluster_analysis (Failed)
Errors while running CTest
```

Build log attached: [build_x86_64.log.txt](https://github.com/espressomd/espresso/files/5255428/build_x86_64.log.txt)

### LINKED ISSUE #3977: Make header files self-contained
TL;DR: ill-formed header files are repeatedly breaking CI in PRs. Impact can be minimized with coding style.

**Problem statement**

Several header files are not self-contained, leading to recurring issues in PRs: 175098b5f7, 5111868bad, 3b95e4c35f, defba4fe2, 70b6da95c1, 395ce4708d, bcf8abc66, 78ade2aec. Investigating compiler errors caused by missing includes can be especially tedious in the case of stream-related headers files, which come in different flavors. For example, finding the `<sstream>` dependency in 2847ae1e23 that caused an "incomplete type" compiler error took me 15 min (in hindsight it should have been obvious, but I bisected boost header files to find it). Some header files only provide a small set of forward declarations of classes, e.g. `<iosfwd>`. STL header files make no guarantee about which other STL header files they include, and it can change with compiler versions. This annoyance can slow down refactoring projects when narrowing includes of header files.

The second issue is portability. Depending on the implementation available on the target machine, ESPResSo may simply fail to compile: 395ce4708d, [`rpms/espresso@77dad47af`](https://src.fedoraproject.org/rpms/espresso/blob/77dad47afa6330ab64b14ce1417ec1f99174b88e/f/espresso-gcc11.patch).

This is not specific to our project:
* C++ Core Guidelines: [SF.11: Header files should be self-contained](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf11-header-files-should-be-self-contained), [SF.10: Avoid dependencies on implicitly #included names](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf10-avoid-dependencies-on-implicitly-included-names)
* NASA's Goddard Space Flight Center *C++ Coding Standard*, Flight Software Branch, Code 582, Version 1.0 ([archive](https://web.archive.org/web/20090412090730/http://software.gsfc.nasa.gov/assetsbytype.cfm?TypeAsset=Standard)):
  > 3.3.7 Each header file shall `#include` the files it needs to compile, rather than forcing users to `#include` the needed files. `#includes` shall be limited to what the header needs; other `#includes` should be placed in the source file.

**Overview of tooling**

From time to time, I run [`include-what-you-use`](https://include-what-you-use.org/) to add missing includes before they become a problem. Unfortunately, this task doesn't lend itself to automation, as the tool can be over-zealous: too many includes, forward declarations, see details in [this post](https://www.incredibuild.com/blog/include-what-you-use-how-to-best-utilize-this-tool-and-avoid-common-issues). The underlying reason is, IWYU looks for header files that directly import symbols used in the current source file, ignoring convenience wrapper header files (these are common in e.g. boost). Therefore, the changes require post-processing by a human, either in vim or with the help of an IDE plugin (e.g. [pokowaka/vscode-iwyu](https://github.com/pokowaka/vscode-iwyu)), or setting up and maintaining a map file. Furthermore, IWYU re-arranges the include statements, making it difficult to extract a clear diff without first patching IWYU. It also requires a recent Clang version (older versions of the tool for Clang < 9 assumed a filename couldn't appear twice in different folders).

The CLion IDE will highlight unused header files and underline variables whose type is unknown, but will not actively highlight types which are not available through one of the includes of the current source file. The Visual Studio Code IDE has a plugin for IWYU, but it doesn't solve the issue of over-zealous changes. The [`deheader`](http://catb.org/~esr/deheader/) tool only removes unused includes.

An automated solution would need to fulfill at least the following criteria:
1. integrate itself in the build process, because the includes depend on ESPResSo features and include paths (`include-what-you-use` does that via `cmake .. -DCMAKE_CXX_INCLUDE_WHAT_YOU_USE="/usr/local/bin/iwyu;-Xiwyu;--no_fwd_decls"`)
2. provide options to control how aggressively to change the include graph
3. support rules to exclude "special" header files
4. produce a clean diff

By "special", I mean:
* `"config.hpp"` which is needed to get the list of features; it should be included last before the first `#ifdef`, but is usually included at the top, probably to avoid compiler errors when it's missing from one of the included header files
* wrapper header files such as the (recently removed) `"cuda_wrapper.hpp"/"curand_wrapper.hpp"/"cufft_wrapper.hpp"` which wrapped around CUDA/HIP includes and added a few macros; these are usually included first, yet should be included last since they provide system header files

Points 2. and 3. are subjective -> human intervention is needed.

**Current status**

A lower estimate of the number of not self-contained header files currently in the project can be obtained with the following algorithm: for each {.cpp file, compiler command}-tuple in `compile_commands.json`, iterate over the N header files in the .cpp file to generate N .cpp files containing only that include and `#include "config.hpp"`, and run the compiler command on them.

A quick implementation in Python generated 1070 valid .cpp files that included 310 header files (out of 422), and revealed that 31 header files failed to compile. Around 15 of them seem to be only missing a single include. The other ones are mostly ScriptInterface header files, where the compiler error log is quite large due to template error messages.

**Proposed solution**

We can minimize the issue by adding to our style guide that include statements should be written in ascending order of precedence. This coding style is already in use in many source files in ESPResSo, and helps detect missing includes early. For example:

```c++
/* mwe.hpp */
void foo() { printf("A\n"); }
/* mwe.cpp */
#include "mwe.hpp"
#include <cstdio> // included last
```
```
$> g++ mwe.cpp
In file included from mwe.cpp:1:
mwe.hpp: In function 'void foo()':
mwe.hpp:1:14: error: 'printf' was not declared in this scope
    1 | void foo() { printf("A\n"); }
      |              ^~~~~~
mwe.hpp:1:1: note: 'printf' is defined in header '<cstdio>'; did you forget to '#include <cstdio>'?
  +++ |+#include <cstdio>
    1 | void foo() { printf("A\n"); }
```

This coding style is easy to adopt, and is already enforced in CI to some extent: the `clang-format` tool moves bracketed includes below quoted includes within an include paragraph.

**Outlook**

The algorithm implemented above is not robust enough to be used in CI, as it will miss a lot of header files. It also cannot detect some issues involving `<sstream>`, probably because the .cpp files don't instantiate any class/struct. I don't think we need a tooling solution, because such a level of quality assurance would be more appropriate for library code rather than for ESPResSo.

jngrad:
Fixes #3977, closes #3905 (ESPResSo 4.2-dev now compiles on GCC 11)

Explicitly include used header files to make source files self-contained and improve portability (see C++ Core Guidelines: [SF.11: Header files should be self-contained](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf11-header-files-should-be-self-contained), [SF.10: Avoid dependencies on implicitly #included names](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf10-avoid-dependencies-on-implicitly-included-names)).

Description of changes:
* apply suggestions from IWYU 10
* add missing includes revealed by an implementation of the `compile_commands.json`-based algorithm outlined in #3977
* remove unused includes and reduce visibility of large header files:
   * `Particle.hpp`: from 332 includes to 245
   * `particle_data.hpp`: from 36 includes to 28
* address several issues:
   * remove a broken ifdef
   * register a unit test in CMake so it can run
   * remove unnecessary static globals in MMM1D GPU
   * dependency inversion for DPD interaction


RudolfWeeber:
What's the rationale for including the stl headers last?

jngrad:
> What's the rationale for including the stl headers last?

I'm ordering header files in reverse generic order (0cb42f615) to help detect headers that are not self-contained. For example, if a header file `mwe.hpp` is using `size_t` but is missing `#include <cstddef>`, then it is not self-contained. If it is included in a source file `mwe.cpp` after the stl headers, we will never be notified that `mwe.hpp` isn't self-contained. If `mwe.cpp` later removes `#include <cstddef>`, compilation fails.

I couldn't find a well-agreed upon style guide for the order of includes:
* the ISO *C++ Core Guidelines* has many recommendations for header file inclusion, but doesn't take a position on how to order these includes
* the Google *C++ Style Guide* advocates for the forward generic order ([Names and Order of Includes](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes)), i.e. the opposite of what I did, but recommend making an exception for the related header, that has to be first to detect if it is not self-contained
* the Lakos *Large-scale C++ Software Design* only mentions "The .c file of every component should include its own .h file as the first substantive line of code." Some people have interpreted it as the Google style, others as the reverse generic order (http://blog.codeimproved.net/posts/large-scale-cpp.html)
* the `clang-format` tool by default orders bracketed includes below quotes includes, and offers a regex-based option to override that order
* discussions about the reverse generic order: https://stackoverflow.com/a/14243688 and https://news.ycombinator.com/item?id=21580829

I opted for the reverse generic order, because it was the most efficient way of detecting missing includes. IWYU missed many header files that were not self-contained. The tool I wrote to compile header files also missed a lot of headers that were not self-contained because it only compiles header files files that are included in .cpp files whose compile options are known, and attempts to apply the same compiler options to header files included in these header files. I stopped at one level of recursion because I get too many compilation errors due to incorrect `-I` compiler options after that. This is a drawback of the Google style order: it only works if each .hpp file has a corresponding .cpp file, or if you have a tool to generate valid `-I` compiler options for any given .hpp file.

RudolfWeeber (approved):

Issue #3955, opened 2020-10-18
Write Tutorial 02 part 2: ELC system with dielectric boundaries


RudolfWeeber:

KaiSzuttor:
please add some details why and how

RudolfWeeber:
Part 1 was changed to charged rod. The crystal stuff was not kept since there were issues with it (@schlaicha) 

The 2nd part still deals with crystal stuff

schlaicha:
I think the general point that should be clarified is if

- we want to provide samples/tutorials that teach physics primarily

or 

- they should primarily show Espresso capabilities.

In the current state part 1 contains a lot of physics and simply uses P3M whereas part 2 is intended as a show case that can use many of the special electrostatics algorithms (ICC, ELC, ...). With @christophlohrmann we started to work on this.

I agree that the current splitting of part 1 and part 2 is arbitrarily/artificial but unless the points above are clarified I would not work further on this specific problem. Note that the clarification concerns also most of the other tutorials.

RudolfWeeber:
@areinauer. Part 2 will be small number of ion pairs with dielectric boundaries and elc.
Crystal part iwll not be kept

PR #3993, opened 2020-11-13
Tutorials maintenance

### LINKED ISSUE #3820: Tutorial 02: electrostatics part 2
* Cite ELC

* The reference to MMM2D has to be removed. No longer part of Espesso

* n_ion_pairs needs to be cast to an int in Python 3

* Some of the parametres need explaining, e.g., gamma.

* The concept of Bjerrum length needs to be explained

* Constraints: 

    - state direction of normal vector (inwards or outwards?)
    - Explain how wall-particle interactions are calculated and that an interaction needs to be setup

* Generate particles in one call per species, Remove id= keyword argument

* The same readability improvements as in part 1 need to be applied to the lj interaction setup

* A loop around the steepest descent is needed. The present code is not guaranteed to result in an overlap-free system

* Replace hand-crafted histogram code by DensityProfile observable

### LINKED ISSUE #3930: Simplify code review of tutorials
Currently, solution cells contain python code wrapped in fenced blocks in markdown cells. The idea was to prevent the user from running the solutions by clicking on `Run All`. However, during hands-on sessions, tutors make sure the participants go through the cells one by one. During self-teaching, participants would presumably also go through cells one by one, because clicking `Run All` would put them at risk of triggering one of the the many assertion statements written between solutions to check participants' code.

Having the solution code wrapped in markdown cells makes it more difficult for tutorial authors and reviewers to run the solutions and edit them. It also makes the CI logic for tutorials more complex and fragile - we missed a bug in #3872, which had to be fixed in a tutorial PR (d6b4d97d8). It's also unclear - without inspecting the python code - what should happen when a solution markdown cell contains a fenced python code block with regular text around it. For future reference, here is the relevant python logic:

https://github.com/espressomd/espresso/blob/46fd5f133a78c9ed7731743e425c035b7d0b751e/doc/tutorials/html_runner.py#L150-L157

It would be simpler from a maintenance perspective to leave code solutions as code cells. *This requires no change to CI.*

If we approve this simplification, the snippet of code above becomes unused and can be removed in the distant future.

### LINKED ISSUE #3934: Issue tracker for Tutorial 01
List of things to fix based on feedback from the summer school.

* `matplotlib` plots
   * should be instantiated with `fig = plt.figure(figsize=(10, 8))` instead of `fig = plt.figure()` to be more readable by taking the entire width of the HTML div
   * should end with `plt.show()` to avoid printing the return value of the last matplotlib method
* `scipy`
   * should be imported with `import scipy.optimize` in the cell where it is used for the first time, i.e. the `def autocor()` cell
* autocorrelation function
   * the function gives a different result than what is calculated in the core with the `Correlator` accumulator, because the normalization doesn't consider that for larger tau values, the sample size is smaller and the variance higher
   * `numpy` and `scipy` are known to be tricky to use for scripting autocorrelation (de Buyl 2018, DOI:[10.21105/joss.00877](https://doi.org/10.21105/joss.00877))
   * the long-term solution here would be to install `tidynamics` in the docker image
* style
   * apply PEP8 formatting
   * `assert` don't need round brackets

### LINKED ISSUE #3959: Consistent numbering for tutorials

jngrad:
Description of changes:
- remove crystallization tutorial (partial fix for #3955, closes #3820)
- remove tutorial numbers and indicate difficulty level (fixes #3959)
- resolve issues in tutorial LJ (fixes #3934)
- create a tool to simplify editing exercise2 notebooks by temporarily converting solution markdown cells to executable code cells (fixes #3930) and apply pep8 formatting
  ```sh
  # convert solution cells to code cells
  ./pypresso doc/tutorials/convert.py exercise2 --to-py doc/tutorials/lennard_jones/lennard_jones.ipynb
  # edit the notebook
  ./ipypresso notebook
  # convert solution cells back to markdown cells
  ./pypresso doc/tutorials/convert.py exercise2 --to-md doc/tutorials/lennard_jones/lennard_jones.ipynb
  # apply autopep formatting rules
  ./pypresso doc/tutorials/convert.py exercise2 --pep8  doc/tutorials/lennard_jones/lennard_jones.ipynb
  ```
- implement an autocorrelation function `espressomd.analyze.autocorrelation()` to calculate the ACF of 1-dimensional and 2-dimensional data


RudolfWeeber (approved):

RudolfWeeber:
I assume, you ran the html build job locally for this. If not, please do before merging.

PR #3999, opened 2020-11-19
New integration tests framework

### LINKED ISSUE #3883: Framework for statistical tests

jngrad:
Fixes #3883

Description of changes:
- write more thorough RNG checks for thermostats
- skip slow/statistical tests in coverage and sanitizer builds
- fix broken ifdefs for Brownian Dynamics
- cleanup IBM code


jngrad:
Here is the summary for the python testsuite, using the same CI runner when possible:

| CI job           | Old runtime | New runtime | speed-up  |
| ---------------- |-------------| ----------- | --------- |
| clang-sanitizer  | 2109.2s     | 988.1s      | 19m (53%) |
| no_rotation      | 647.7s      | 468.5s      | 3m (28%)  |
| maxset           | 961.6s      | 616.0s      | 6m (36%)  |
| default          | 970.5s      | 771.3s      | 3m (20%)  |
| cuda11-maxset    | 1113.4s     | 774.9s      | 6m (30%)  |

The complete pipeline runtime went from 72 min to 38 min. The Clang job completes in 35 min instead of 50 min.

@RudolfWeeber your plan in https://github.com/espressomd/espresso/issues/3883#issuecomment-716630623 was really good, however I had to deviate from it slightly:
* `virtual_sites_tracers` is needed to get code coverage for IBM particle coupling to the fluid (not sure why though, the LB actor is removed before the IBM bonds are created)
* Brownian/NpT were moved fully, but new files with minimum checks were introduced for both integrators, otherwise their code coverage would be 0%

Other remarks:
* `make check_python` still works the same way as before
* `make check_python_skip_long` to run the subset of fast tests
* the comment lines in the thermostat tests about RNG seed and counter state were obtained using [cout.diff](https://github.com/espressomd/espresso/files/5568591/cout.diff.txt)

jngrad:
The slowest tests are now:
* `stokesian_dynamics`: 1 min 10 sec in sanitizer, 40 sec in coverage, 20-30s in others
   * now skipped in sanitizer/coverage builds
   * new test added to check the SD interface
* `elc_vs_analytic`: 1 min 20 sec in sanitizer/coverage
   * removing it from coverage builds would require writing a small test without constant potential to preserve code coverage
* reaction ensemble tests:
   * `constant_pH`: 1 min 30 sec in sanitizer, 45 sec in coverage
   * `reaction_ensemble`: 30-40 sec in sanitizer/coverage, 5 sec in others
   * `widom_insertion`: 30-40 sec in sanitizer/coverage, 5 sec in others
   * `wang_landau_reaction_ensemble`: already skipped in sanitizer/coverage
   * long-term plan would be to skip these tests in sanitizer/coverage and write much smaller tests to check the RE interface (error messages, output files, and run a few reaction steps), like I did here for the thermostats and integrators

testsuite/python/elc.py
```diff
@@ -35,8 +35,8 @@ class ElcTest(ut.TestCase):
     def test_finite_potential_drop(self):
         s = self.system
 
-        p1 = s.part.add(pos=GAP + [0, 0, 1], q=+1)
-        p2 = s.part.add(pos=GAP + [0, 0, 9], q=-1)
+        p1 = s.part.add(pos=[0, 0, 1], q=+1)
```
RudolfWeeber, line 38: I thought it doesn't matter, where in the system the gap is? @reinaual 
Did this change influence the test outcome?

RudolfWeeber (approved):

testsuite/python/elc.py
jngrad: This change did not alter the numerical result. I think the gap is always at the higher value of the z-axis:
https://github.com/espressomd/espresso/blob/127ad4ef7c9cee2b4242e76139cf69f55230d628/src/core/electrostatics_magnetostatics/elc.hpp#L92-L93
https://github.com/espressomd/espresso/blob/127ad4ef7c9cee2b4242e76139cf69f55230d628/src/core/electrostatics_magnetostatics/elc.cpp#L1196

If particles enter the gap, they are misclassified as particles from the dielectric contrast when calculating the forces and energies. The documentation warns against using ELC without walls to block particles from entering the gap:
https://github.com/espressomd/espresso/blob/127ad4ef7c9cee2b4242e76139cf69f55230d628/doc/sphinx/electrostatics.rst#L254-L259

reinaual: In theory it should not matter, to my knowledge in the implementation it does. The lower interface is assumed at `z=0` and the upper one at `z=h`, which is relevant for the error-estimation and the placement of the image charges.
To my understanding the testcase should be correct for every given prefactor, however setting it to `1e-100` allows to explicitly test the potential difference contribution but hides the misplaced particles in this case.

PR #4000, opened 2020-11-20
CI: Pin Python version in mac runners

RudolfWeeber:


jngrad (approved):

.github/workflows/build.yml
```diff
@@ -5,10 +5,13 @@ on:
...
 
 jobs:
-  tests:
+  github_mactest:
     runs-on: macos-latest
     steps:
     - uses: actions/checkout@main
+    - uses: actions/setup-python@v2
+      with:
+          python-version: '3.7'
```
KaiSzuttor, line 14: should we use the opportunity to test also the minimum required version which is 3.6?

RudolfWeeber:
I don't care about the Python version to pin. What was the reason we decided on 3.6 as minimum in the first place? Which ever distro that was by necessity would test that?

@jngrad, please make a decision.

jngrad:
It was OpenSUSE, which will remain on Python 3.6 and Boost 1.66 until the next minor release (15.3) is published in July 2021. Moving to the rolling release OpenSUSE Tumbleweed could make CI more unstable, and we would have to vet which contributed packages to include. If we choose to drop support for OpenSUSE, we can bump the requirements to Python 3.7 and Boost 1.67 (Debian 10). This would help reducing code in h5md (we had to re-implement `dataclasses` in a2d42e1f77), but Boost 1.67 doesn't bring much compared to 1.66.

Python 3.6 is not really relevant for end users. Python 3.6 was dropped from NEP29 in June 2020, and many python packages based on numpy have pledged to stop supporting versions of their packages that were released for Python 3.6.

jngrad:
We could get Python 3.8 on OpenSUSE by installing it from the Tumbleweed packages (jngrad/espresso-docker@71979bb2d908c). That would install both python versions, but by bumping requirements to Python 3.7, CMake would automatically select 3.8. ESPResSo compiles and passes the unit and python tests in this docker image.

Otherwise, I wouldn't bother testing on Python 3.6 with macOS. If any python package breaks, there is nothing we can do but bump them back to Python 3.7, because package maintainers who follow NEP29 are not supposed to provide the same level of support for Python 3.6-related issues compared to 3.7+.

PR #4002, opened 2020-11-20
Make vtk dependency optional

RudolfWeeber:
* Make vtk import in Mayavi visualizer optional. It is only used for log message suppression
* Remove vtk installation from mac build jobs


jngrad:
We can safely remove it from macOS. The other use of vtk in the waLBerla branch is already guarded by a try/catch:
https://github.com/espressomd/espresso/blob/c2d079da811b61a8ae290dd8bfb2316efb9f9e18/testsuite/python/lb_vtk.py#L25-L31

jngrad (approved):

PR #4022, opened 2020-12-02
pairs by type

christophlohrmann:
Description of changes:
- add a ``types`` argument to ``system.cell_system.get_pairs()`` to look only for pairs of particles with the given types
- Interface change: removed the ``_`` from the end of ``get_pairs_()``


src/core/cells.cpp
```diff
@@ -48,23 +50,24 @@
...
  */
-std::vector<std::pair<int, int>> get_pairs(double distance) {
+template <class Filter>
+std::vector<std::pair<int, int>> get_pairs_filtered(double const distance,
+                                                    Filter filter) {
   std::vector<std::pair<int, int>> ret;
+  on_observable_calc();
   auto const cutoff2 = distance * distance;
-
-  cells_update_ghosts(Cells::DATA_PART_POSITION | Cells::DATA_PART_PROPERTIES);
```
fweik: Are you sure this is not needed anymore?

src/core/cells.cpp
```diff
@@ -79,29 +82,66 @@ std::vector<std::pair<int, int>> get_pairs(double distance) {
...
+                              std::vector<int> const &types) {
+  auto local_pairs = get_pairs_of_types(distance, types);
+  Utils::Mpi::gather_buffer(local_pairs, comm_cart);
+}
 
-  boost::mpi::broadcast(comm_cart, distance, 0);
+REGISTER_CALLBACK(get_pairs_of_types_local)
+
+namespace detail {
+void search_distance_sanity_check(double const distance) {
```
fweik: This can be implemented in a cell system agnostic way, by using ```ParticleDecomposition::max_range()```, also I think
this should be place inside the cell system code, not here, if there is no such check already.

fweik (changes requested):

src/core/cells.cpp
christophlohrmann: ``on_observable_calc()`` does the ghost update

src/core/cells.cpp
christophlohrmann: as far as I understand, ``max_range()`` refers to the local box length of the node without taking into consideration the cell decomposition within that local box

src/core/cells.cpp
fweik: Ah right!

src/core/cells.cpp
fweik: Ok, you are right, not a good run for me so far :-). But then this should be implemented, the way you are doing it encodes implementation details of the cell system at the call site. We were working hard for a long time to get those out, I think we shouldn't implement new ones. E.g. either each `ParticleDecomposition` should implement their own range check, or it should expose the maximum range that can be done without a resort.

christophlohrmann: ok how about I rename ``max_range()`` to ``max_cutoff()`` to say which is the maximum cutoff that the chosen decomposition can handle; and then a new function ``max_range()`` to give the actual range that the calculations are performed in after the (local cell-) decomposition has been done?

fweik: Ough, names are hard. But I'd use a new name for the new function, to not cause any unneeded confusion. I think `range` is better suited than `cutoff` (passive vs active), how about `current_range`?

christophlohrmann: can do. But the doc for ``max_range()`` is ``@brief Maximum supported cutoff.`` :D

fweik: Then I'd propose that you implemented `search_distance_sanity_check` in terms of this new function (without metioning the composition type), and we'll leave adding the check in a generic way to `CellStructure` for a later day.

christophlohrmann: alright

christophlohrmann: on a side note: Even for atom decomposition, the maximum cutoff is still bound by the global box size, right?

fweik: I'd do `auto current_cutoff() { return max_cutoff(); }` :-), it's not actually bounded, because every pair is visited no matter the distance...

fweik (dismissed):
LGTM, not so sure about the names tho...

src/core/AtomDecomposition.hpp
```diff
@@ -92,7 +92,9 @@ class AtomDecomposition : public ParticleDecomposition {
     return id_to_cell(p.identity());
   }
 
+  Utils::Vector3d max_cutoff() const override;
```
KaiSzuttor: this needs some docstring

christophlohrmann: are docstrings not inherited?

KaiSzuttor: ah you are right, i overlooked the inheritance

christophlohrmann:
Concerning the maximum search range not considering the skin: the pair finder calls ``on_observable_calc()`` that among other things resorts the particles

jngrad (approved):

PR #4025, opened 2020-12-03
Cells Housekeeping

fweik:
Followup to #4022, some minor improvements to `CellStructure`

Description of changes:
- Made mutable access to the internal `ParticleDecomposition` private
- Avoided one particle copy when switching decomposition
- Split one large function
- Made verlet lists (the data structure) private
- Removed questionable default cutoff from `short_range_loop`.

### DIFF
--- a/src/core/AtomDecomposition.cpp
+++ b/src/core/AtomDecomposition.cpp
@@ -147,6 +147,8 @@ AtomDecomposition::AtomDecomposition(const boost::mpi::communicator &comm,
   mark_cells();
 }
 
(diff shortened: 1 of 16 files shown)


RudolfWeeber:
In general, looks good.

Two questions, though:
* the pair search sanity check uses `max_range()` of the ParticleDecomposition. For domdec, this is cell size. Is a resort guaranteed before the pair search? Otherwise, isn't there a skin missing?
* Are we actually testing somewhere that the particles are preserved when switching the decomposition?

fweik:
@RudolfWeeber I think you commented on the wrong PR.

PR #4031, opened 2020-12-08
Introduce boost qvm and a Quaternion type

### LINKED ISSUE #2289: Replace custom rotation and vector code by boost.qvm

KaiSzuttor:
Fixes #2289

### DIFF
--- a/src/core/Particle.hpp
+++ b/src/core/Particle.hpp
@@ -25,6 +25,7 @@
 
 #include <utils/Vector.hpp>
 #include <utils/math/quaternion.hpp>
+#include <utils/quaternion.hpp>
 
 #include <boost/serialization/vector.hpp>
 
@@ -128,9 +129,10 @@ struct ParticleProperties {
     int to_particle_id = 0;
     double distance = 0;
     /** Relative position of the virtual site. */
-    Utils::Vector4d rel_orientation = {0., 0., 0., 0.};
+    Utils::Quaternion<double> rel_orientation =
+        Utils::Quaternion<double>::zero();
     /** Orientation of the virtual particle in the body fixed frame. */
-    Utils::Vector4d quat = {0., 0., 0., 0.};
+    Utils::Quaternion<double> quat = Utils::Quaternion<double>{0., 0., 0., 0.};
 
     template <class Archive> void serialize(Archive &ar, long int) {
       ar &to_particle_id;
@@ -248,7 +250,7 @@ struct ParticlePosition {
 
 #ifdef ROTATION
   /** quaternion to define particle orientation */
-  Utils::Vector4d quat = {1., 0., 0., 0.};
+  Utils::Quaternion<double> quat = Utils::Quaternion<double>::identity();
   /** unit director calculated from the quaternion */
   Utils::Vector3d calc_director() const {
     return Utils::convert_quaternion_to_director(quat);
--- a/src/core/Particle.hpp
+++ b/src/core/Particle.hpp
@@ -132,7 +132,7 @@ struct ParticleProperties {
     Utils::Quaternion<double> rel_orientation =
         Utils::Quaternion<double>::zero();
     /** Orientation of the virtual particle in the body fixed frame. */
-    Utils::Quaternion<double> quat = Utils::Quaternion<double>{0., 0., 0., 0.};
+    Utils::Quaternion<double> quat = Utils::Quaternion<double>::zero();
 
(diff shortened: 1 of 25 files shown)


src/utils/include/utils/quaternion.hpp
```diff
@@ -0,0 +1,137 @@
...
+#define SRC_UTILS_INCLUDE_UTILS_QUATERNION_HPP
+
+#include <boost/qvm/deduce_scalar.hpp>
+#include <boost/qvm/quat_operations.hpp>
+#include <boost/qvm/quat_traits.hpp>
+
+#include "utils/Vector.hpp"
+
+namespace Utils {
+template <typename T> class Quaternion : private Vector<T, 4> {
```
fweik: Couldn't you just use the quaternion class from qvm?

KaiSzuttor: you mean this?

``` c++
namespace boost { namespace qvm {

    template <class T>
    struct quat {

      T a[4];

      template <class R>
      operator R() const {
        R r;
        assign(r,*this);
        return r;
      }

    };
}
}

Comment (fweik) [reply src/utils/include/utils/quaternion.hpp]: Not enough `enable_if`?

File: src/utils/tests/quaternion_test.cpp [hunk 2]
```diff
@@ -1,7 +1,6 @@
 /*
  * Copyright (C) 2019 The ESPResSo project
  *
- * This file is part of ESPResSo.
```
jngrad: accidental line deletion

src/utils/include/utils/quaternion.hpp
```diff
@@ -0,0 +1,137 @@
+#ifndef SRC_UTILS_INCLUDE_UTILS_QUATERNION_HPP
```
jngrad, line 19: missing license header

fweik:
I'd find it helpful if you would clarify what the final goal of this is. Do you want to do all vector/... operations via qvm? That type should come from where? E.g. it is not clear to my why you redfinde the free function operator overloads for `Quaternion` instead of just inlinening those of qvm, which you also use for the implementation. This looks odd to me, but I'm sure there is a good reason, albeit it not being spelled out. Also it has to be clear that what ever is done in this space will be very hard to change in the future...

KaiSzuttor:
the reason is that this is a draft...

fweik:
Yes, which is why it seems to me the right time to discuss the design. I don't think it helps towards a constructive design process that you are so defensive.

KaiSzuttor:
i just use this draft PR for CI feedback, thanks for your feedback

src/core/Particle.hpp
```diff
@@ -128,9 +129,10 @@ struct ParticleProperties {
     int to_particle_id = 0;
     double distance = 0;
     /** Relative position of the virtual site. */
-    Utils::Vector4d rel_orientation = {0., 0., 0., 0.};
+    Utils::Quaternion<double> rel_orientation =
+        Utils::Quaternion<double>::zero();
     /** Orientation of the virtual particle in the body fixed frame. */
-    Utils::Vector4d quat = {0., 0., 0., 0.};
+    Utils::Quaternion<double> quat = Utils::Quaternion<double>{0., 0., 0., 0.};
```
jngrad:
```suggestion
    Utils::Quaternion<double> quat = Utils::Quaternion<double>::zero();
```

src/utils/include/utils/quaternion.hpp
```diff
@@ -0,0 +1,136 @@
+/*
```
jngrad, line 1: should this file be renamed `Quaternion.hpp`?

jngrad (commented):
besides minor changes, LGTM

src/utils/include/utils/quaternion.hpp
KaiSzuttor: i think we should not use capital letters in filenames at all... that lead to issues on case-insensitive file systems on windows or mac

jngrad (approved):

PR #4037, opened 2020-12-11
Fix clearing the list of actors

### LINKED ISSUE #4033: The list of actors cannot be cleared
`system.actors.clear()` doesn't clear the list of actors if ELC or DLC is present.

MWE:

```python
import espressomd, espressomd.electrostatics
import espressomd.magnetostatics, espressomd.magnetostatic_extensions
system = espressomd.System(box_l=3 * [6], time_step=0.01, periodicity=3*[1])
system.part.add(pos=[3*[0], 3*[1]], q=[-1, 1], dip=[[1,0,0], [-1,0,0]],
                rotation=2*[3*[1]])

p3m = espressomd.electrostatics.P3M(prefactor=1, mesh=32, cao=5, accuracy=1e-3)
dp3m = espressomd.magnetostatics.DipolarP3M(prefactor=1, mesh=32, accuracy=1E-4)
dlc = espressomd.magnetostatic_extensions.DLC(maxPWerror=1E-5, gap_size=2.)
system.actors.add(dp3m)
system.actors.add(dlc)
system.actors.add(p3m)

print(system.actors)
print('clear()')
system.actors.clear()
print(system.actors)
print('clear()')
system.actors.clear()
print(system.actors)
```
Output:
```
[<espressomd.magnetostatics.DipolarP3M object at 0x151b5bcc9680>,
 <espressomd.magnetostatic_extensions.DLC object at 0x151b5bcbd810>,
 <espressomd.electrostatics.P3M object at 0x151b5bcc9640>]
clear()
[<espressomd.magnetostatic_extensions.DLC object at 0x151b5bcbd810>]
clear()
[]
```

Calling `system.actors.clear()` a second time removes DLC. Here is the current `clear()` implementation:
https://github.com/espressomd/espresso/blob/59d100cfe2606533cc1cc46c1364228589f6d607/src/python/espressomd/actors.pyx#L232-L235

Not sure why the for loop exits early. Re-implementing it as a while loop solves the issue with DLC:
```python
    def clear(self):
        """Remove all actors."""
        while len(self.active_actors):
            self.remove(self.active_actors[0])
```

This is a source of errors in the `dipolar_mdlc_p3m_scafacos_p2nfft.py` test: when the DLC check fails, the ScaFaCoS check throws an exception because the list of actors is not emptied by `tearDown()`.

Regarding ELC, it's not possible to remove the actor without triggering "Exception: Unable to remove ELC as the state of the underlying electrostatics method will remain unclear". Is it also the case that removing DLC leaves the magnetostatics method in an undetermined state?

The `clear()` docstring doesn't mention anything about ELC and DLC being "unclearable".

Also, is this discrepancy between `self` and `Actors` intentional? (introduced by 3139ecd517)
https://github.com/espressomd/espresso/blob/59d100cfe2606533cc1cc46c1364228589f6d607/src/python/espressomd/actors.pyx#L213-L214

jngrad:
Fixes #4033

Description of changes:
- rewrite the `Actors.clear()` method so that it removes all actors


RudolfWeeber (approved):

src/python/espressomd/actors.pyx
```diff
@@ -231,8 +231,8 @@ class Actors:
 
     def clear(self):
         """Remove all actors."""
-        for a in self.active_actors:
-            self.remove(a)
+        while len(self.active_actors):
```
KaiSzuttor, line 234: why don't we use the clear method of python lists?

KaiSzuttor, line 234:
```suggestion
    self.active_actors.clear()
```

RudolfWeeber:
>why don't we use the clear method of python lists?

 

IIRc, because actors have a hook for when they are being removed (turn stuff off in the core).

You cannot use Pythons __-hook, because the actual destruction of the object can be deferred by the garbage collection.

PR #4040, opened 2020-12-14
Fix visualizer freezes.

### LINKED ISSUE #3764: Visualizer freezes with Steepest Descent
MWE:
```python
import espressomd
import numpy as np
system.time_step = 0.01
system.cell_system.skin = 0.4
system.periodicity = [1, 1, 1]
system.part.add(pos=[2, 2, 2])
system.part.add(pos=[3, 2, 2])
system.non_bonded_inter[0, 0].wca.set_params(epsilon=10, sigma=1)
system.integrator.run(steps=0)
f_max = 1
assert np.linalg.norm(system.part[0].f) > 10 * f_max
system.integrator.set_steepest_descent(f_max=f_max, gamma=3., max_displacement=0.1)
visualizer.run(1)
```

Behavior: the OpenGL visualizer doesn't start or freezes (on 4.1.3rc 8ee3aa86a50b) or is extremely slow (on 4.2-dev b2ced40e09e368).

Expected behavior: the visualizer starts and two particles become visible, with no time evolution.

Slowing down the main loop solves the freeze, suggesting a race condition between the two threads:
```diff
--- src/python/espressomd/visualization_opengl.pyx
+++ src/python/espressomd/visualization_opengl.pyx
@@ -590,4 +590,5 @@ class openGLLive:
                     try:
                         self.system.integrator.run(integ_steps)
+                        time.sleep(0.0001)
                     except Exception as e:
                         print(e)
```

### LINKED ISSUE #4038: Visualizer: particle info on double click
In the OpenGL-Visualizer one was able to double-click a particle to see the particle traits. Currently this action results in 
`TypeError: _get_particle_info() takes 2 positional arguments but 3 were given`
which is because of a function-signature change introduced with f55cbeab49bc3bed8aebc0a92df17ff1277840da that is incompatible to the assumed callback signature.
Reverting the function signature change in line 1517 of `src/python/espressomd/visualization_opengl.py` of f55cbeab49bc3bed8aebc0a92df17ff1277840da gives back a working version.

```
--- a/src/python/espressomd/visualization_opengl.py
+++ b/src/python/espressomd/visualization_opengl.py
@@ -1530,7 +1530,8 @@ class openGLLive():
         if self.drag_id != -1:
             self.trigger_reset_particle_drag = True
 
-    def _get_particle_info(self, pos):
+    # pylint: disable=unused-argument
+    def _get_particle_info(self, pos, pos_old):
         part_id, _ = self._get_particle_id(pos)
         if self.show_system_info:
             self.show_system_info = False
```

pkreissl:
Fixes #3764, fixes  #4038 

Description of changes:
- Fix freezes when using `visualizer.run()`
- simplify timers
- get rid of if-condition in update loop which prevented visualizer update when using steepest descent
- fix callback function call to `_get_particle_info()` (#4038)

After some profiling, I think I figured out what was causing the freezes:

In multi-threaded Python programs, only one thread is running at a time. To coordinate this, Python (still) uses the General Interpreter Lock (GIL). When a thread is running it holds the GIL, which is typically released on I/O. If a thread doesn't have I/O, the GIL is released regularly after some time (or number of execution cycles) and can then be acquired by any thread (even the one that just released it). The time between GIL releases can depend on the Python/Cython version as well as OS.

In `visualizer.run()`, there are two competing threads:
 1. the deamon thread main loop that calls `visualizer.update()` to trigger a redraw and if not paused `system.integrate()`
 2. the main thread that does all the OpenGL magic, amongst others check for keyboard input, register corresponding callbacks, redraw on idle at a certain rate (60 fps by default)

Visualizer freezes can happen when the rate of GIL releases is smaller or comparable to the update and/or redraw rate, as it can effectively block either updating system data or redrawing if the other thread acquires the GIL.

Typical examples of when that happened are:
 - paused visualizer
 - steepest descent for simple system, converge fast and then basically take no time

On my machine in these cases, time between calls of `update()` were or order 1e-5 so basically running continuously compared to the thread release rate (GIL release) which was by default given as
```
>>> print(sys.getswitchinterval())
0.005
```
and is quite close to the rate at which updates/redraws are triggered (0.017 s).

One can check the above, by removing the `time.sleep()` which triggers a GIL release of the sleeping thread. The described freezes can be observed.
If now the switch rate is increased via
```
sys.setswitchinterval(0.00005)
```
frame rate should be a solid 60 fps.

However, there might still corner cases where both threads compete for the GIL at comparable rates, e.g., run JNs sample script of Issue #3764, where you do delete the line that switches the integrator to steepest descent. If you pause, mark one particle with a double click to view its information, the system visualization becomes quite slow unless you unpause -- cython calls (such as integrate) seem to releases the GIL.

To prevent such corner cases, calling `time.sleep()` with a negligibly small time instead of setting the switch interval seems a more reasonable approach as it triggers a GIL release after each `update()` allowing the main thread to become active and handle both callbacks and redraw.


jngrad (approved):
LGTM, thanks for adding comments and docstrings!

PR #4019, opened 2020-11-27
Cylindrical transformations with well-defined phi angle

jngrad:
Follow-up to #4014

Description of changes:
- add overloads to convert from Cartesian to cylindrical coordinates with custom longitudinal axis and phi angle

Issue #4014, opened 2020-11-26
Coordinate transformations


RudolfWeeber:
In context with #3960, coordinate transforms are needed.
In the particular case, Cartesian -> cylindrical  with basis vectors and origin not identical.

To my understanding, the most general solution is to split this into
1. translation to align the origins
2. basis transform to align the basis vectors
3. convert Cartesian to cylindrical coordinates for matching basis vectors. I.e. theta=0 would refer to the x axis, theta=pi/2 to the y-axis and z =z.

---

1. is probably not worth a function, since it is vector subtraction.

2. The second part can be done most general (supporting also non-orthogonal coordinates)
using a transformation matrix containing the new basis vectors (as columns, i guess)
This only requires matrix vector product which is in the utils library.
https://en.wikipedia.org/wiki/Change_of_basis

3. theta using atan2 (the one that can deal with four quadrants, r=sqrt(x^2+y^2), z=z

The reverse would look like:
x = r cos(theta), y = r sin(theta), z=z

To reverse 2., the caller has to use the original basis transform with the inverse transformation matrix. 
Matrix inversion can be taken from boost::qvm

RudolfWeeber:
This is done, i guess.

PR #4042, opened 2020-12-14
Cylindrical transformations with well-defined phi angle

jngrad:
Follow-up to #4014, closes #4019

Description of changes:
- rewrite coordinate transformations with matrix operations

### DIFF
--- a/src/core/observables/CylindricalLBProfileObservable.hpp
+++ b/src/core/observables/CylindricalLBProfileObservable.hpp
@@ -47,16 +47,15 @@ public:
         limits[0], limits[1], limits[2], n_bins[0], n_bins[1], n_bins[2],
         sampling_density);
     for (auto &p : sampling_positions) {
-      double theta;
-      Utils::Vector3d rotation_axis;
       auto p_cart = Utils::transform_coordinate_cylinder_to_cartesian(
           p, Utils::Vector3d{{0.0, 0.0, 1.0}});
       // We have to rotate the coordinates since the utils function assumes
       // z-axis symmetry.
-      std::tie(theta, rotation_axis) =
-          Utils::rotation_params(Utils::Vector3d{{0.0, 0.0, 1.0}}, axis);
+      constexpr Utils::Vector3d z_axis{{0.0, 0.0, 1.0}};
+      auto const theta = Utils::angle_between(z_axis, axis);
+      auto const rot_axis = Utils::vector_product(z_axis, axis).normalize();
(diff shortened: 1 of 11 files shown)


jngrad:
@KaiSzuttor could you please have a look at the failing tests with cylindrical LB and `HollowConicalFrustum` in commit 764cc32e8f8? Several tests failed because `[1, 0, 0]` was used as both the phi0 axis and cylinder longitudinal axis, which lead to an under-defined phi angle, and thus to a division by zero in the matrix inversion, hence NaN values. But after removing the lines in question, a few tests still fail with NaN values.

src/core/observables/CylindricalLBProfileObservable.hpp
```diff
@@ -47,16 +47,15 @@ class CylindricalLBProfileObservable : public CylindricalProfileObservable {
...
     for (auto &p : sampling_positions) {
-      double theta;
-      Utils::Vector3d rotation_axis;
       auto p_cart = Utils::transform_coordinate_cylinder_to_cartesian(
           p, Utils::Vector3d{{0.0, 0.0, 1.0}});
       // We have to rotate the coordinates since the utils function assumes
       // z-axis symmetry.
-      std::tie(theta, rotation_axis) =
-          Utils::rotation_params(Utils::Vector3d{{0.0, 0.0, 1.0}}, axis);
+      constexpr Utils::Vector3d z_axis{{0.0, 0.0, 1.0}};
```
KaiSzuttor, line 54: why do we still manually rotate at the call site? I thought the idea was to include that into the transformation? Then you can also test it properly.

jngrad:
Alright, so the issues in the testsuite come from test cases where the x-axis is used both for the orientation and the longitudinal axis, which lead to an under-defined phi angle (the rotation matrix is not invertible when two columns are identical). They are temporarily commented out in 90d21a85510b7.

jngrad:
Multiple failures in CI for 90d21a85, each with a different segfault backtrace. Cannot reproduce locally in Release and Debug builds.

christophlohrmann:
I am trying to get this PR to work, but I have a quick question: Should we assert that the ``axis`` and the phi=0 ``orientation`` are orthogonal or should we just take the orthogonal component in case there is a tilt?

KaiSzuttor:
i'm in favour of asserting

christophlohrmann:
In that case we should make it such that you have to provide ``orientation`` if you use ``axis``

KaiSzuttor:
can this be closed because of #4114?

Issue #2071, opened 2018-07-23
Low test coverage: Candidates for downgrade or removal


RudolfWeeber:
This ticket lists parts of Espresso that do not have (sufficient)
tests. 
These parts will likely be marked to require
#define EXPERIMENTAL_FEATURES
in myconfig.hpp, or they may be removed, if the situation persists.

As per https://codecov.io/gh/espressomd/espresso/tree/python/src/core

Potential users/developers/authors are mentioned, below

Accumulators/Correlator.cpp:
* several of the compression functions are untested

Electrostatic interactions:
elc (some 200 lines) @konradbreitsprecher
dipolar direct sum (with replica) (@RudolfWeeber)

Immersed boundary method (@sgekle):
The volume conservation part is entirely untested.
For the bonded interactions, the equations themselves are untested, but the correct minimum is verified.
See also #2305

object-in-fluid @icimrak 
* out_direction
* membrane collision (See also #2214)
* affinity
* For the volume and surface conservation forces, the explicit expressions are not tested, but the restoration of a distorted sphere is checked.
See also #2305

shapes 
stomatocyte (@jdgraaf)
half of rhomboid, hollowcone, slit_pore

Other:
Galilei transform
communication.cpp (200 lines. These are probably more indications of other stuff not being tested)
statistics.cpp (200 lines)
~~metadynamics (no Py interface)~~
spec_func (some 100 lines)

KaiSzuttor:
codecov is currently not updating the coverage...

KaiSzuttor:
thanks @mkuron

mkuron:
@KaiSzuttor, seems like they changed the IP addresses used. I've updated our firewall rules now, so we should be getting coverage reports again.

Also, our coverage reports for the python branch are wrong. We don't mirror it to Gitlab, so it only gets coverage results from the maxset build (from the documentation deployment) and from Travis. Could you change that to mirror all branches to Gitlab?

KaiSzuttor:
i don't quite get what you mean. We have not only the maxset build reporting to codecov, but also e.g. no_cuda_default

jonaslandsgesell:
This issue got closed automatically while merging. I guess we want to reopen it.

jonaslandsgesell:
mmm1d is tested as of merging of PR #2893

Kischy:
I would hate to see the Gay-Berne potential removed. Who is allowed to contribute? If I am allowed to contribute, do you have guidelines for contributions?

[...]

PR #4046, opened 2020-12-16
Test dipolar direct sum methods

jngrad:
Partial fix for #2071

Description of changes:
- add test cases for dds without replica on CPU/GPU, with replica, with DLC


jngrad:
offline discussion: the test doesn't pass anymore if the dipoles are relaxed with steepest descent:
```diff
diff --git a/testsuite/python/dipolar_mdlc_dds_scafacos_direct.py b/testsuite/python/dipolar_mdlc_dds_scafacos_direct.py
index 7ddcb6c65..0a470baff 100644
--- a/testsuite/python/dipolar_mdlc_dds_scafacos_direct.py
+++ b/testsuite/python/dipolar_mdlc_dds_scafacos_direct.py
@@ -157,6 +157,17 @@ class dds(ut.TestCase):
             system.integrator.set_steepest_descent(
                 f_max=1, gamma=0.001, max_displacement=0.01)
             system.integrator.run(100)
+            dds_cpu = espressomd.magnetostatics.DipolarDirectSumCpu(
+                prefactor=1.2)
+            system.actors.add(dds_cpu)
+            system.integrator.set_steepest_descent(
+                f_max=5,
+                gamma=0.001,
+                max_displacement=0.01)
+            print(system.analysis.energy()["dipolar"])
+            print(system.integrator.run(100))
+            print(system.analysis.energy()["dipolar"])
+            system.actors.clear()
             system.integrator.set_vv()
 
             # compute forces and energies for dawaanr
```

RudolfWeeber:
The difference between Espresso's and Scafacos' direct sum with replica seems to be that Espresso uses a spherical cutoff, i.e.
sqrt(i_x^2 +i_y^2 +i_z^2) <= n_replica^2
whereas scafacos does the full cube.
We probably better check, whether the dds with replica gets monotonically closer to p3m with increasing number of replica or compare or a tiny system against results calculate din Python.

jngrad:
I tried with 2 to 8 replica, with and without DLC, but the deviation to ScaFaCoS did not decrease, while the runtime increased to the order of 1 min.

RudolfWeeber:
It is not possible to make Espressos direct sum with replica agree to scafacos direct sum with replica. 
They use different cutoff rules:

* Espresso: Spherical cutoff on the replica lavel. Only include boxes when i_x^2+i_y^2+i_z^2 <=n_replica^2, ix,y,z in N
* Scafacos: Either don't use spherical cutoff at all (i.e., n_replica=4,4,4 also cinludes i_x=-4,i_y=-4, i_z=-4, or cutoff on the particle distance level (using the cutoff argument).

I'm therefore removing this test.

RudolfWeeber:
Dipolar direct sum with replica is broken for 0 replica. In that case, it should match with the direct summation for open boundaries, which it doesn't
dds_cpu without replica, dds from scafacos, dds gpu match

jngrad:
> Dipolar direct sum with replica is broken for 0 replica. In that case, it should match with the direct summation for open boundaries, which it doesn't

Then why does my test pass?

jngrad:
Also, if you test DDS with replica using 0 replica against ScaFaCoS direct summation, you get the same results after energy minimization (using the diff in the second post of this PR), whereas DDS with open borders doesn't match ScaFaCoS anymore.

RudolfWeeber:
dds_cpu and dds_gpu use minimum image convention in case of perioidic bc, dds with replica and scafacos don't.

PR #4049, opened 2020-12-16
Improve LB interface testing

jngrad:
Description of changes:
- rename `print_vtk_*` functions to `write_vtk_*` functions in LB and EK to better reflect what they actually do
- simplify VTK bounding box code and make it coherent with the numpy slice syntax (i.e. `bb2` now needs zero-based indices)
- add assertions to catch an out of bounds bug in VTK (incorrect bounding box values caused the python interpreter to freeze)
- fix broken LB getters in the core and python interface
- fix incorrect LB density getter/setter check in the testsuite
- add extra checks for LB parameters
- add testcase for VTK output


src/core/grid_based_algorithms/lb_interface.cpp
```diff
@@ -563,28 +564,36 @@ void lb_lbfluid_print_vtk_velocity(const std::string &filename,
...
-    bb_low.push_back(std::min(*val1, *val2));
-    bb_high.push_back(std::max(*val1, *val2));
+    auto const lower = std::min(*val1, *val2);
+    auto const upper = std::max(*val1, *val2);
+    if (lower < 0) {
+      throw std::runtime_error("Cannot access index " + std::to_string(lower));
+    }
+    if (upper >= bb_high[it]) {
+      throw std::runtime_error("Cannot access index " + std::to_string(upper));
+    }
```
KaiSzuttor:
```suggestion
    if (lower < 0 or upper >= bb_high[it]) {
      throw std::runtime_error("Cannot access index " + std::to_string(lower));
    }
```

src/core/grid_based_algorithms/lb_interface.cpp
```diff
@@ -563,28 +564,36 @@ void lb_lbfluid_print_vtk_velocity(const std::string &filename,
...
+  if (lattice_switch == ActiveLB::GPU) {
+#ifdef CUDA
+    bb_high = {static_cast<int>(lbpar_gpu.dim_x) - 1,
+               static_cast<int>(lbpar_gpu.dim_y) - 1,
+               static_cast<int>(lbpar_gpu.dim_z) - 1};
+#endif //  CUDA
+  } else {
+    bb_high = {lblattice.global_grid[0] - 1, lblattice.global_grid[1] - 1,
+               lblattice.global_grid[2] - 1};
+  }
```
KaiSzuttor:
```suggestion
  auto const bb_low = Utils::Vector3i{};
  auto const bb_high = lb_lbfluid_get_shape() - Utils::Vector3i{1,1,1};
```

src/python/espressomd/lb.pyx
```diff
@@ -205,23 +205,69 @@ cdef class HydrodynamicInteraction(Actor):
...
+        Parameters
+        ----------
+        path : :obj:`str`
+            Path to the output ASCII file.
+
+        """
         lb_lbfluid_print_vtk_boundary(utils.to_char_pointer(path))
 
     def print_velocity(self, path):
+        """Write the LB fluid velocity to a data file that can be loaded by
```
KaiSzuttor, line 255: we should consider renaming those functions because we actually write to a file and not just print to screen

testsuite/python/lb_vtk.py
```diff
@@ -0,0 +1,225 @@
...
+import unittest as ut
+import unittest_decorators as utx
+
+import os
+import numpy as np
+
+try:
+    import vtk
+    from vtk.util import numpy_support as VN
+    skipIfMissingPythonPackage = ut.case._id
```
KaiSzuttor: what's that?

KaiSzuttor (dismissed):
thanks for those tests! just minor improvements should be done before merging

testsuite/python/lb_vtk.py
jngrad: `vtk` is not a hard dependency

src/core/grid_based_algorithms/lb_interface.cpp
jngrad: wouldn't the error message be confusing if `upper` is the one that causes the problem?

testsuite/python/lb_vtk.py
KaiSzuttor: but what is ```_id```

KaiSzuttor: not sure if it's a good idea to use a private variable

src/python/espressomd/lb.pyx
jngrad: maybe `write_velocity`? Also, do you know why the CPU LB implementation of `print_boundary` casts numbers to 1, while the GPU LB writes some integers?

testsuite/python/lb_vtk.py
jngrad: that's just a lambda that returns the input function, it's called `_id` because `id` is a python built-in function

KaiSzuttor: yes but if it was part of the official interface they'd looked for a proper name, right?

src/core/grid_based_algorithms/lb_interface.cpp
KaiSzuttor: yes of course, overlooked that, but i think it would be best to merge those errors and just print the whole information: lower, upper, and the limits

src/core/grid_based_algorithms/lb_interface.cpp
KaiSzuttor: here, we can probably drop the - 1 and just check for smaller

src/python/espressomd/lb.pyx
KaiSzuttor: i don't only mean this function but also the other ```print_*``` functions in this file. ```write_*``` would be a more accurate description

jngrad: ok, but there are other places in the code with similarly confusing names, e.g. the electrokinetics `print_vtk_*` functions. Is it worth introducing this breaking change in 4.2, considering these LB and EK functions have already been removed in 4.3?

src/core/grid_based_algorithms/lb_interface.cpp
jngrad: good idea!

src/python/espressomd/lb.pyx
KaiSzuttor: don't know, i think we should clean up what we stumble over regardless of the lifetime of the change

testsuite/python/lb_vtk.py
jngrad: we now use our own `_id` lambda in 897c5bb118

src/python/espressomd/lb.pyx
jngrad: done

src/core/grid_based_algorithms/lb_interface.cpp

testsuite/python/h5md.py
```diff
@@ -30,7 +30,7 @@
 import espressomd.io.writer
 try:
     import h5py  # h5py has to be imported *after* espressomd (MPI)
-    skipIfMissingPythonPackage = ut.case._id
+    skipIfMissingPythonPackage = utx._id
```
KaiSzuttor: okay, i still don't get why this should be a private function

testsuite/python/mdanalysis.py
```diff
@@ -24,10 +24,11 @@
 import espressomd.interactions
 import numpy as np
 import unittest as ut
+import unittest_decorators as utx
 try:
     import MDAnalysis as mda
     from espressomd import MDA_ESP
-    skipIfMissingPythonPackage = ut.case._id
+    skipIfMissingPythonPackage = utx._id
```
KaiSzuttor:
```suggestion
    skipIfMissingPythonPackage = lambda x: x
```

KaiSzuttor (commented):
LGTM, please clarify the private id function which must have been introduced elsewhere

testsuite/python/mdanalysis.py
KaiSzuttor: this would be even more readable, right?

testsuite/python/h5md.py
jngrad: I've renamed it to `no_skip` in 08a674fae

testsuite/python/mdanalysis.py
jngrad: we want to hide implementation details of python decorators (that's why `unittest` created `_id()` in the first place)

also, that lambda expression would be transformed into a regular function by autopep8

christophlohrmann (approved):
Changes after Kai's review look good to me

reinaual (approved):

PR #4050, opened 2020-12-17
Improved error message for type error.

### LINKED ISSUE #3908: Improve error message in number_of_particles_with_type()
The method `system.number_of_particles(type)` can throw an exception when `type` is not in the list of tracked particle types.

Here is the relevant core function:

https://github.com/espressomd/espresso/blob/f760050199b94226b9127b309bfa7db569507605/src/core/particle_data.cpp#L1048-L1053

The runtime error should be replaced by something that supports streams to produce a more descriptive error message, e.g.:

```c++
runtimeErrorMsg() << "The provided particle type " << type << " is currently not tracked by the system";
```

which would alert users that they need to call `system.setup_type_map()`. The docstring of `system.number_of_particles()` should be extended with the following paragraph:

```
Raises
------
RuntimeError
    If the particle ``type`` is not currently tracked by the system. To select which particle
    types are tracked, call :meth:`setup_type_map`.
```

Issue reported by @schlaicha.

pstaerk:
Fixes #3908

Description of changes:
- Changed the error message to include missing type.
- Appended docstring to include the error that is thrown.

### DIFF
--- a/src/core/particle_data.cpp
+++ b/src/core/particle_data.cpp
@@ -1181,7 +1181,8 @@ void add_id_to_type_map(int part_id, int type) {
 
 int number_of_particles_with_type(int type) {
   if (particle_type_map.count(type) == 0)
-    throw std::runtime_error("The provided particle type " + std::to_string(type) +
+    throw std::runtime_error("The provided particle type " +
+                             std::to_string(type) +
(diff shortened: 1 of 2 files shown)


jngrad (approved):
thanks!

Issue #4009, opened 2020-11-24
Throw if particles enter ELC gap region


RudolfWeeber:
In case ELC does a pass over all particles, anyway, it could be checked that none is in the gap region. This is, if ELC can distinguish between normal particles and image charges

fweik:
> This is, if ELC can distinguish between normal particles and image charges

Image charges are not actually created as particles, this shouldn't be a problem.

biermanncarl:
I could try this on coding day

jngrad:
There needs to be a check in the python testsuite, for example something like the following:
```python
    @utx.skipIfMissingFeatures(["P3M"])
    def test_elc_gap_exception(self):
        system = self.system
        elc_gap = 2.
        system.actors.add(
            espressomd.electrostatics.P3M(prefactor=1, mesh=3 * [32],
                                          cao=5, accuracy=1e-3, tune=False))
        system.actors.add(electrostatic_extensions.ELC(gap_size=elc_gap,
                                                       maxPWerror=1e-3))
        system.part.add(pos=[0, 0, 1], q=+1)
        system.part.add(pos=[0, 0, system.box_l[2] - elc_gap / 2], q=-1)
        with self.assertRaises(Exception):
            system.analysis.energy()
        with self.assertRaises(Exception):
            system.integrator.run(0)
```

The magnetostatics counterpart, DLC, has the same issue:
https://github.com/espressomd/espresso/blob/59d100cfe2606533cc1cc46c1364228589f6d607/src/core/electrostatics_magnetostatics/mdlc_correction.cpp#L493-L496
The check could be added in `get_DLC_energy_dipolar()` and `get_DLC_dipolar()` before the first `for` loop, using a call to e.g. `dlc_gap_check(particles)` with such an implementation:
```c++
static void dlc_gap_check(const ParticleRange &particles) {
  auto const h = dlc_params.h;  for (auto const &p : particles) {
    if (p.p.dipm > 0 and p.r.p[2] > h) {
      runtimeErrorMsg() << "Particle " << p.identity() << " entered the DLC gap";
      break;
    }
  }
}
```

jngrad:
Beware, `system.actors.clear()` *doesn't* clear the list of electrostatics/magnetostatics actors! This is important to remember when writing the python test outlined above for both ELC and DLC. More details in #4033.

PR #4051, opened 2020-12-17
Added check if particles are not in ELC gap region

biermanncarl:
Fixes Issue #4009

Description of changes:
- add a check in both positive and negative direction with error message

### DIFF
--- a/src/core/electrostatics_magnetostatics/elc.cpp
+++ b/src/core/electrostatics_magnetostatics/elc.cpp
@@ -223,16 +223,17 @@ void distribute(int size) {
   MPI_Allreduce(send_buf, gblcblk, size, MPI_DOUBLE, MPI_SUM, comm_cart);
 }
 
-/** checks if a particle is in the forbidden gap region
+/** Checks if a particle is in the forbidden gap region
  */
 inline void check_gap(const Particle &p) {
-  if ((p.p.q != 0) && ((p.r.p[2] > elc_params.h) || (p.r.p[2] < 0))) {
+  if (p.p.q != 0) {
     if (p.r.p[2] < 0)
       runtimeErrorMsg() << "Particle " << p.p.identity << " entered ELC gap "
                         << "region by " << (p.r.p[2]);
-    else
+    else if (p.r.p[2] > elc_params.h) {
       runtimeErrorMsg() << "Particle " << p.p.identity << " entered ELC gap "
                         << "region by " << (p.r.p[2] - elc_params.h);
+    }
   }
 }
 
--- a/src/core/electrostatics_magnetostatics/elc.cpp
+++ b/src/core/electrostatics_magnetostatics/elc.cpp
@@ -247,6 +247,16 @@ static void add_dipole_force(const ParticleRange &particles) {
   gblcblk[2] = 0; // sum q_i
 
   for (auto const &p : local_particles) {
+    /* check if charged particles are in valid region */
+    if ((p.p.q != 0) && ((p.r.p[2] > elc_params.h) || (p.r.p[2] < 0))) {
+      if (p.r.p[2] < 0)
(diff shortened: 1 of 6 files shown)


src/core/electrostatics_magnetostatics/elc.cpp
```diff
@@ -247,6 +247,16 @@ static void add_dipole_force(const ParticleRange &particles) {
   gblcblk[2] = 0; // sum q_i
 
   for (auto const &p : local_particles) {
+    /* check if charged particles are in valid region */
+    if ((p.p.q != 0) && ((p.r.p[2] > elc_params.h) || (p.r.p[2] < 0))) {
+      if (p.r.p[2] < 0)
+        runtimeErrorMsg() << "Particle " << p.p.identity << " entered ELC gap "
```
jngrad: We probably want the same check in `double dipole_energy()`, in which case this block can be moved to a free function, e.g. `inline void elc_gap_sanity_check(Particle const &p)`. We also need to add a test to make sure the `runtimeErrorMsg()` is properly converted to a Python exception in the Cython code, e.g. with
```python
    def test_elc_gap_exception(self):
        ... # set up an ELC system, or re-use one that has already been set up in another test
        self.system.constraints.clear() # remove walls, if any was used to minimize the system
        self.system.part[0].pos = [box_l, box_l, box_l + elc_gap - 1]
        with self.assertRaises(Exception):
            self.system.analysis.energy()
        with self.assertRaises(Exception):
            self.integrator.run(2)
```

src/core/electrostatics_magnetostatics/elc.cpp
```diff
@@ -247,6 +247,16 @@ static void add_dipole_force(const ParticleRange &particles) {
   gblcblk[2] = 0; // sum q_i
 
   for (auto const &p : local_particles) {
+    /* check if charged particles are in valid region */
+    if ((p.p.q != 0) && ((p.r.p[2] > elc_params.h) || (p.r.p[2] < 0))) {
```
KaiSzuttor: why are you checking for the same condition in the next line?

biermanncarl: Well, only part of the same condition. The outer condition checks whether there is an error, at all. The inner condition to distinguish between the two different directions. I could do two independent checks. But then I would have to check ```q!=0``` twice. Or we could leave away the information about how far the particle is in the forbidden region, then I could leave the inner condition away.

jngrad: what about checking for the charge in the outer conditional, and using `if(p.r.p[2] > elc_params.h) ... else if(p.r.p[2] < 0))` in the inner conditional?

KaiSzuttor: i think this should have a name and should be moved out to a function. the current implementation is happening in the function ```add_dipole_force```.

jngrad:
please use `maintainer/CI/fix_style.sh` :-)

biermanncarl:
> please use `maintainer/CI/fix_style.sh` :-)

I tried, but doesn't work on my local machine properly yet ... It didn't do anything. Sorry.

jngrad (dismissed):
thanks!

jngrad:
Could you please amend these two paragraphs to explain the method will throw if particles enter the gap?
https://github.com/espressomd/espresso/blob/25187188e4271acf6a15493b22bfc526ff72f77d/doc/sphinx/electrostatics.rst#L258-L259
https://github.com/espressomd/espresso/blob/25187188e4271acf6a15493b22bfc526ff72f77d/doc/sphinx/magnetostatics.rst#L99-L100
After that you can remove the WIP from the title, and we're good to merge.

jngrad (approved):

PR #4055, opened 2020-12-17
Document dependency python-wxGTK for mayavi visualizer.

### LINKED ISSUE #4001: Mayavi visualizer depends on python3-wxgtk4
To get rid of
No pyface toolkit plugin for wx
upon 
`import espressomd.visualization_mayavi`

pkreissl:
Fixes #4001


KaiSzuttor:
is it in ```requirements.txt```?

pkreissl:
> is it in `requirements.txt`?

No, it is not. Neither are the vtk python bindings. If you want these to be added to `requirements.txt`, which version would you go for? The one currently available via the standard Ubuntu 20.04 repository? The Python 3 or Python 2 versions? In Fedora, e.g., the packages are called differently -- so pip instead?

Or do we keep it at this general level of: ok, you need `vtk` and `wxGTK4` pythonXY bindings.

I would be fine with all versions.

KaiSzuttor:
just put it in the requirements file without any version, as long as we are not aware of any version that does not work, we can just omit the version.

requirements.txt
```diff
@@ -9,7 +9,9 @@ pint>=0.9
 matplotlib>=2.1.1
 vtk>=7.1.1
 PyOpenGL>=3.1.0
-mayavi>=4.5.0
+mayavi>=4.5.0 # for mayavi-visualizer
+python-vtk # for mayavi-visualizer, e.g. python3-vtk7 (Ubuntu)
+python-wxgtk # for mayavi-visualizer, e.g. python3-wxgtk4.0 (Ubuntu)
```
jngrad: there is no package `python-wxgtk` on pypi (https://pypi.org/project/python-wxgtk/), please remove it

`python-vtk` is a personal project with no documentation nor repository (https://pypi.org/project/python-vtk/), which is then used with `import python_vtk`; are you really sure this is what we should install?

jngrad (changes requested):

requirements.txt
jngrad: I removed that change to `requirements.txt`.

jngrad (approved):

requirements.txt
pkreissl: Fine with me, thanks!

Issue #4029, opened 2020-12-07
Make particle id less prominent in system setup and particle doc


RudolfWeeber:
In the system setup and particle related sections of the user's guide, a lot of emphasis is put on referring to particles by their numerical id.
This is not the preferred way to work with particles, since it results in error-prone scripts and id-based slicing behavior can be surprising.

The docs should be changed to emphasize, that
* specifiying the particle id is optional
* system.part.add() returns either a single particle handle or a particle slice of the created particles. This can be stored in a variable and referred to later
* Bonds can be created using those references
* system.part.select() can be used for creating filtered slices

The underlying discussion is in #3992.

biermanncarl:
I could have a look at this on Coding Day

jngrad:
A lot of particle ids and access-by-ids were removed from the testsuite. There's still a lot of it left, here is how to find them:
```sh
ls -1 testsuite/python/*.py | xargs -d '\n' grep --color -P '\.part\[[^:]+?\]'
ls -1 testsuite/python/*.py | xargs -d '\n' grep --color -P '[^a-zA-Z_]id='
```
These tests typically don't rely on iteration. Instead, the particle ids are carefully chosen for the test logic, sometimes with intentional gaps or patterns. Once we decide on the new particle access mechanism, we can revisit them.

biermanncarl:
Section 7 of the user guide (bonded interactions) needs some rework, too, to get rid of numerical particle IDs

biermanncarl:
What I personally find a bit annoying is the fact that ParticleSlice objects aren't subscriptable. Sure, one can iterate over them with
```python
for p in my_slice:
    ...
```
but for debugging, bond setup and various other purposes, it would be convenient to be able to access one single particle out of a slice without much effort, like ```print(myslice[0].pos)```. This is easy when accessing particles with numerical ids. Would it be hard to make slices subscriptable?

jngrad:
> Would it be hard to make slices subscriptable?

The current particle slice implementation supports slices with gaps: https://github.com/espressomd/espresso/issues/3992#issuecomment-728063238. Once we're done removing particle ids from the docs and tests, we can design a more pythonic slice implementation that supports subscription.

PR #4058, opened 2020-12-17
Reduce direct use of numerical part id (doc, samples)

pkreissl:
Part of #4029

Description of changes:
- make `vs_auto_relate_to()` take `ParticleHandle`, not just particle id
- remove particle id from doc and samples wherever possible
- improve checking of arguments for VS and bonds (e.g. prevent adding the same bond twice by calling `add_bond()` with a particle id and calling it again with the corresponding particle handle)

There are still some places, especially in the `espressomd.observables`, where for now particle ids cannot be omitted, such as `ParticlePositions`.


src/python/espressomd/particle_data.pyx
```diff
@@ -692,11 +692,18 @@ cdef class ParticleHandle:
         # vs_auto_relate_to
         def vs_auto_relate_to(self, _relto):
             """
-            Setup this particle as virtual site relative to the particle with the given id.
+            Setup this particle as virtual site relative to the particle with the given ParticleHandle or id.
 
             """
-            check_type_or_throw_except(
-                _relto, 1, int, "Argument of vs_auto_relate_to has to be of type int.")
+            # If _relto is of type ParticleHandle,
```
KaiSzuttor: none of your business but this variable name is not too descriptive and the underscore is not pep8... just if your over-motivated :)

KaiSzuttor (dismissed):
👍 really good stuff

src/python/espressomd/particle_data.pyx
pkreissl: I agree, the current variable names in this part of the interface are not as I would like them to be. However, this is true for all member variables of the `ParticleHandle` class. I think the idea here is to mirror the core variable but adding a prefixed underscore to not have the variables being **exactly** the same. I think @RudolfWeeber commented in some PR (that I can't find just know), that he would like to have distinct names for the variables in core and interface respectively. If I am wrong here, sure I can quickly go over the class and "pepify" it a bit.

RudolfWeeber (dismissed):

jngrad (approved):

PR #4061, opened 2020-12-18
Dipolar tests

### LINKED ISSUE #1569: mdlc (dipolar layer correction) test needs to be checked
The fact that the test did not fail in spite of the bug removed in #1568 suggests that the test didn't test the method sufficiently. The test data uses sqrt(3) as dipole moment, so the unintended cast to int in the calculation of the total dipole moment should have made a difference.

RudolfWeeber:
fixes #1569 closes #4046 

This is based on #4046, minus the last commit.

Description of changes:
* Switch tests for dipolar interactions with open boundary conditions from comparing two methods against each other to comparing individual methods against reference data
* Use prefactor != 1 in dipolar 2d and 3d periodic tests
* Block DipolarDirectSumWithReplicaCpu from n_replica=0 and periodic bc, as it does not apply minimum image convention. Introducing minimum image convention there is not worth it, because that is provided by DipolarDirectSumCpu. The two methods should be merged anyway.


jngrad (approved):
LGTM

Issue #3680, opened 2020-04-22
Use separate cutoff for bonded and non-bonded interactions


fweik:
Because espresso finds bond partners by an index, and not by spatial neighbor search it is sufficient
for bond partners to be present (real or ghost) on the node where a bond is evaluated. Hence the cutoff for bonds needs only to be considered for the halo size, not for the cell size. This can lead to
considerable performance increase: e.g. currently adding a dihedral bond doubles the cutoff, and hence the cell size. This causes an 8-fold increase in the number of pairs that are considered in the short range pair finding. It  would actually be sufficient die just use a two-cell halo and leave the cell size as is, which would not affect the performance of the hot loop.

This is also required to be able to handle multi-body bonds without the need of a reduction after the force calculation, because by selecting an appropriate halo size for the bond types present (1, 2 or 3) allows for evaluating bonds on ghost particles without influencing the performance of the pair loop.

fweik:
Furthermore the bonded cutoff can then be ignored on a single node and for the n^2 cell system, because in either case all particles are always present.

jngrad:
Several points have already been addressed in the upcoming 4.2.0 release, while the cell size remains an issue:
- the bonded cutoff and non-bonded cutoff are now treated separately in the short-range loop
  https://github.com/espressomd/espresso/blob/ef9f4979ee3ab753f44e803147fb9ae75ace2ffc/src/core/short_range_loop.hpp#L47-L53
- the bonded cutoff is now ignored on 1 MPI rank
  https://github.com/espressomd/espresso/blob/ef9f4979ee3ab753f44e803147fb9ae75ace2ffc/src/core/interactions.cpp#L59-L63
- the bonded cutoff is still used for the skin, and hence for the cell size via `mpi_set_skin()` and `cells_re_init()`
  https://github.com/espressomd/espresso/blob/ef9f4979ee3ab753f44e803147fb9ae75ace2ffc/src/core/integrate.cpp#L448-L451

PR #4066, opened 2020-12-21
Skip bond calculation loop if not needed

jngrad:
Rebase of #4030 on top of `python`
Partial fix for #3680

Description of changes:
- separate bonded IA cutoff from non-bonded IA cutoff
- skip bond loop independently from non-bonded loop

### DIFF
--- a/src/core/short_range_loop.hpp
+++ b/src/core/short_range_loop.hpp
@@ -38,14 +38,17 @@ struct True {
 template <class BondKernel, class PairKernel,
           class VerletCriterion = detail::True>
 void short_range_loop(BondKernel bond_kernel, PairKernel pair_kernel,
-                      double distance_cutoff,
+                      double pair_cutoff, double bond_cutoff,
                       const VerletCriterion &verlet_criterion = {}) {
   ESPRESSO_PROFILER_CXX_MARK_FUNCTION;
 
   assert(cell_structure.get_resort_particles() == Cells::RESORT_NONE);
 
-  cell_structure.bond_loop(bond_kernel);
-  if (distance_cutoff > 0.)
+  if (bond_cutoff >= 0.) {
+    cell_structure.bond_loop(bond_kernel);
+  }
+
+  if (pair_cutoff >= 0.)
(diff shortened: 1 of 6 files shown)


src/core/short_range_loop.hpp
```diff
@@ -38,14 +38,17 @@ struct True {
...
-                      double distance_cutoff,
+                      double pair_cutoff, double bond_cutoff,
                       const VerletCriterion &verlet_criterion = {}) {
   ESPRESSO_PROFILER_CXX_MARK_FUNCTION;
 
   assert(cell_structure.get_resort_particles() == Cells::RESORT_NONE);
 
-  cell_structure.bond_loop(bond_kernel);
-  if (distance_cutoff > 0.)
+  if (bond_cutoff >= 0.) {
```
jngrad, line 47: Why use `>=` here to include zero, now that bonded IAs without range return `0.` as the cutoff instead of `-1.`?
Could it be a mistake, i.e. bonded IAs w/o range should now return `BONDED_INACTIVE_CUTOFF` instead of `0.`?

KaiSzuttor: including the zero does not make sense to me

RudolfWeeber:
So, there are a few bonds that do not have a range (e.g. angle_harmonic). Still, if they are present, the bond loop has to be executed. I'm not 100% sure whether any or all of the oif bonds really fall into that category, since the volume conservation, at least, is handled separately.
However, even if the oif bonds should be incorrectly categorized, it would olnly mean a small performance drop, and only if the bonds in question are active.

i think we can go ahead with this. How do we handle the (probable) merge conflict with Carl's work on the bond data structures?

src/core/short_range_loop.hpp
jngrad: The tests do not pass if I replace `>=` by `>`. We actually need to run the bond loop when there are bonds with zero cutoff, e.g. angle and dihedral bonds, because their range is infinite. The "cutoff" in ESPResSo has a different meaning from "range".

PR #4068, opened 2020-12-22
Guard against ScaFaCoS fatal errors

jngrad:
Description of changes:
- catch runtime errors from ScaFaCoS at the python level instead of crashing ESPResSo
- use standard exception mechanism when setting Coulomb and dipolar prefactors


src/python/espressomd/scafacos.pyx
```diff
@@ -54,7 +54,12 @@ IF SCAFACOS == 1:
             return {"method_name", "method_params", "prefactor"}
 
         def validate_params(self):
-            return True
+            if self._params["method_name"] not in available_methods():
+                raise ValueError(
+                    f"method '{self._params['method_name']}' is unknown or not compiled in ScaFaCoS")
+            if not self._params["method_params"]:
+                raise ValueError(
+                    "ScaFaCoS methods require at least 1 parameter")
```
jngrad, line 62: @RudolfWeeber I looked through `FCSResult fcs_set_parameters()` in `scafacos/src/FCSInterface.c` and from what I understood, the empty string is not allowed. Is my assessment correct, or did I miss something?

reinaual (approved):

src/python/espressomd/scafacos.pyx
jngrad: ok found it, it's by design: when ScaFaCoS fails to match a string (unknown parameter string or empty string) against the list of expected parameters, control flow reaches [`scafacos/scafacos@src/FCSInterface.c#L891-L892`](https://github.com/scafacos/scafacos/blob/066f753f0572c7397508231cb4fc9432d5aeaf04/src/FCSInterface.c#L891-L892), which returns a non-zero error code.

PR #4069, opened 2020-12-22
core: replaced utils::strcat_alloc with printf

### LINKED ISSUE #4020: Remove strcat_alloc()
`Utils::strcat_alloc()` appends a C-string to a char array with manual memory management. This utility function is one of the last consumers of the ESPResSo manual memory management toolset (#2900), the others being EK and LB that are currently being rewritten in the waLBerla branch.

Here is a typical use of `Utils::strcat_alloc()`:
```c++
double p3m_mc_time(char **log, const int mesh[3], int cao, ...) {
  char buffer[200];
  sprintf(buffer, "%-4d %-3d cao too large for this mesh\n", mesh[0], cao);
  *log = strcat_alloc(*log, buffer);
}
```

The same can be achieved with a `std::string`:
```c++
#include <string>
double p3m_mc_time(std::string &log, const int mesh[3], int cao, ...) {
  char buffer[200];
  sprintf(buffer, "%-4d %-3d cao too large for this mesh\n", mesh[0], cao);
  log += buffer;
}
```

Affected files: `p3m.cpp`, `p3m-dipolar.cpp`, `mmm1d.cpp` to remove calls to `Utils::strcat_alloc()`, and `electrostatics.pxd`, `magnetostatics.pxd` to replace `char ** log` by `string &log` from `libcpp.string`. After that, the utility function and its unit test can be removed.

The `sprintf()` statements do not need to be modified. They have two modern `std::string` equivalents, but they cannot be used out-of-the-box: `std::format` is a C++20 feature, `boost::format` uses a different syntax. Anyway, these `sprintf()` calls are found in functions that are probably going to be rewritten in Python in the future, so further beautification isn't worth it.

reinaual:
Fixes #4020

Description of changes:
- replaced manual string `char*` construct with `std::printf`
- introduced verbose parameter to touched tuning functions and interfaces
- removed `utils/strcat_alloc.hpp` with unit tests
- added missing `handle_errors` call in p3m_gpu_init


src/core/electrostatics_magnetostatics/p3m.cpp
```diff
@@ -707,12 +705,11 @@ static double p3m_mcr_time(const int mesh[3], int cao, double r_cut_iL,
  *           -@ref P3M_TUNE_FAIL, -@ref P3M_TUNE_ACCURACY_TOO_LARGE,
  *           -@ref P3M_TUNE_CAO_TOO_LARGE, or -@ref P3M_TUNE_ELCTEST
  */
-static double p3m_mc_time(char **log, const int mesh[3], int cao,
-                          double r_cut_iL_min, double r_cut_iL_max,
-                          double *_r_cut_iL, double *_alpha_L,
-                          double *_accuracy) {
+template <bool verbose>
+static double p3m_mc_time(const int mesh[3], int cao, double r_cut_iL_min,
```
jngrad, line 708: Here and elsewhere: I do not immediately see the advantage of using a template parameter instead of a regular parameter. Could you please let me know why this was necessary?

With a template function, there is no guarantee the `true` and `false` instantiations will do the same calculation. It is, for example, perfectly legal for someone to inject this snippet of code at line 798:
```c++
template <>
double p3m_mc_time<false>(const int mesh[3], int cao, double r_cut_iL_min,
                          double r_cut_iL_max, double *_r_cut_iL,
                          double *_alpha_L, double *_accuracy) {
  return 0.;
}
```
This will cause the tuning to select random values, which means we should duplicate the mmm1d/p3m/dp3m python tests to make sure `verbose=False` gives the same results as `verbose=True`.

reinaual: The template was just personal preference. I dont care, your choice is probably to change it to a regular paramter, i can quickly do that if you want me to.

jngrad: I think it would be more maintainable with a regular parameter. This would also allow you to merge `p3m_adaptive_tune_core()` and `p3m_adaptive_tune()` back into one function.

reinaual: im working on it, this is an artifact of cython not supporting non-type templates

jngrad (approved):

PR #4071, opened 2020-12-23
Investigate bot issues

jngrad:
Over the last two weeks, the bot stopped deleting style comments after a successful pep8 style check. This PR is not meant to be merged. It's only there to help investigate the bot python code.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -33,7 +33,7 @@ variables:
 status_pending:
   <<: *notification_job_definition
   stage: prepare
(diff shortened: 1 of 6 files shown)


KaiSzuttor:
you should probably check the backend log

jngrad:
It's actually just a bug introduced when moving to `pre-commit`: the `gh_post_style_patch.py` script is now only called when `pre-commit` fails, but we actually need to run the python script every time because it contains the logic to delete old posts. I'm currently refactoring the bot scripts.

PR #4075, opened 2020-12-26
Fix Utils::Mpi::gather_buffer() functions

### LINKED ISSUE #4074: gather_buffer() functions are broken
The two `Utils::Mpi::gather_buffer()` functions give random results when parameter `root` is not 0:
* the first `n_elem` of the buffer on rank `root` are overwritten by the first `n_elem` of the buffer on rank 0
* there is a hole of size `n_elem` somewhere in the buffer on rank `root`, which contains either random data (in the pointer overload) or values generated by `class Allocator` (in the container overload)

These functions basically forgot to move the first `n_elem` of the buffer on rank `root` into the correct place before gathering buffers from the other ranks. The corresponding unit test was broken since the beginning due to a copy-paste error. A quick look at the source code of 4.2, 4.1.4, 4.0.2 and the walberla branch revealed no use of these functions for the `root != 0` case.

jngrad:
Fixes #4074

Description of changes:
- fix a bug that caused buffers to overwrite each others in `Utils::Mpi::gather_buffer()` when argument `root` was not 0
- fix two broken unit tests for `Utils::Mpi` functions


src/utils/include/utils/mpi/gather_buffer.hpp
```diff
@@ -104,6 +117,13 @@ void gather_buffer(std::vector<T, Allocator> &buffer,
     /* Resize the buffer */
     buffer.resize(tot_size);
 
+    /* Move the original data to its new location */
```
KaiSzuttor, line 127: btw: why is there so much code duplication? can we get rid of it?

jngrad: The `std::vector` version takes care of memory allocation. The pointer version doesn't and assumes the caller took care of calling `realloc()`. Once we're done removing all the manual memory allocation in the core, the pointer version can be removed, and all the code duplication will be gone.

src/utils/include/utils/mpi/gather_buffer.hpp
```diff
@@ -63,11 +65,19 @@ int gather_buffer(T *buffer, int n_elem, boost::mpi::communicator comm,
     auto const total_size =
         detail::size_and_offset<T>(sizes, displ, n_elem, comm, root);
 
+    /* Move the original data to its new location */
+    if (sizes[root] && displ[root]) {
```
KaiSzuttor: you can move this part out into a function in a detail namespace, e.g.

RudolfWeeber (approved):

PR #4084, opened 2021-01-07
Update build system for Clang 10

### LINKED ISSUE #4082: Implicit float conversions raise warnings in Clang 10
Clang 10 enabled the constant-checked version of `-Wimplicit-float-conversion` by default (see [D64666](https://reviews.llvm.org/D64666#1596634)). It checks that integers can be represented exactly as a float during an implicit conversion. If you pass it to the compiler, it will flag all cases where an integer is implicitely converted to a float. If you don't pass it, Clang will still flag implicit conversions, but only in a constant context, i.e. the warning is raised only when the implicit conversion:

* can be evaluated at compile time
* involves an integer variable explicitely marked as `constexpr` or `const`, or an integer literal
* results in an off-by-1 (or more) deviation from the original integer

To complicate matters, flag [`-Wimplicit-float-conversion`](https://releases.llvm.org/10.0.0/tools/clang/docs/DiagnosticsReference.html#wimplicit-float-conversion) enables a subset of more specialized flags, and in Clang 11, the specialized flag [`-Wimplicit-int-float-conversion`](https://releases.llvm.org/11.0.0/tools/clang/docs/DiagnosticsReference.html#wimplicit-int-float-conversion) enables the more specialized flag `-Wimplicit-const-int-float-conversion`. Here is a MWE for Clang 10 ([compiler explorer output](https://godbolt.org/z/s67dz6)):

```c++
#include<limits>
double foo() {
    // compiler warning because variable 'a' is const
    std::size_t const a = std::numeric_limits<std::size_t>::max();
    double const b = 1. / (a + 1.);
    return b;
}
double bar() {
    // no compiler warning because variable 'a' is not const
    std::size_t a = std::numeric_limits<std::size_t>::max();
    double const b = 1. / (a + 1.);
    return b;
}
```

This is confusing, as simply adding a `const` somewhere can trigger one of those warnings. To avoid inconsistent behavior across compiler versions, we can simply fix these compiler warnings in all of our codebase. This is what the waLBerla codebase did.

ESPResSo is already affected by that warning, when using Clang 10 and enabling the Stokesian Dynamics feature. The MWE is actually adapted from the Stokesian Dynamics library. It was fixed upstream in hmenke/espresso-stokesian-dynamics#3.

jngrad:
Description of changes:
- fix Clang's `-Wimplicit-float-conversion` everywhere (enabled by default in Clang 10 in constant expressions), fixes #4082
- exclude the ScaFaCoS and Stokesian Dynamics external libraries from Clang-Tidy using the `SYSTEM` method already used by Random123 and h5xx (ScaFaCoS raises several alerts in Clang-Tidy 10)
- make NVCC and Clang consistent in which `sm_XY` architecture(s) get compiled depending on the CUDA version


KaiSzuttor (dismissed):

jngrad:
There's no point in adding the automerge label if the CI infrastructure is down :-)

KaiSzuttor:
> There's no point in adding the automerge label if the CI infrastructure is down :-)

so i cant express that i want to merge if our infrastructure is down?

jngrad:
> > There's no point in adding the automerge label if the CI infrastructure is down :-)
> 
> so i cant express that i want to merge if our infrastructure is down?

The merge commit will fail and kodiak will try to merge the following PRs in the queue, which will also fail. In the end we'll have merge commits everywhere, and once CI is up again, kodiak will have to introduce more merge commits since all PRs in the queue will outdate each other. We had this situation on December 19, if you remember.

reinaual (approved):

PR #4094, opened 2021-01-15
fix and test transform_vector_cartesian_to_cylinder

christophlohrmann:
Description of changes:
- Fixes missing ``sqrt`` in calculation of vector transformation
- Unit test the function


jngrad:
Looks like these equations from the failing tests have the wrong unit for the velocity:
https://github.com/espressomd/espresso/blob/python/testsuite/python/observable_cylindricalLB.py#L109-L110
https://github.com/espressomd/espresso/blob/python/testsuite/python/observable_cylindrical.py#L95-L97
Working out `v_x` and `v_y` backwards from the expressions in the `utils` function, I think the correct values should be:
```python
            r = np.sqrt(position[0]**2 + position[1]**2)
            v_y = (r * self.v_phi + position[1] * self.v_r) / r
            v_x = (r * self.v_r - position[1] * v_y) / position[0]
```
but with these, `phi` values from the core are different from the reference values by a factor `sqrt(2)`, and have the wrong signs.

christophlohrmann:
I changed the velocity calculation. Will push in a minute

jngrad (approved):
Thanks a lot!

PR #4095, opened 2021-01-15
CUDA exception handling

### LINKED ISSUE #4085: -lcuda needs to be removed for cuda-11.2
At least in the installation done by spack, libcuda is not present, there is just libcudart.
This needs to be changed in cmake/FindCUDACompilerNVCC.cmake.
Don't know if -lcuda is needed for older cuda versions.

jngrad:
Description of changes:
- convert CUDA error codes into runtime errors using the message associated to the error code
- properly handle CUDA errors by halting the flow of the program instead of ignoring them (Barnes-Hut, LB GPU)
- remove superfluous CUDA global variables (partial fix for #2628)
- restore `cuda_gather_gpus()` functionality and convert the `device_list()` getter to a regular function (API change)
- remove unused `libcuda` dependency (fixes #4085)


src/core/cuda_init.cpp
```diff
@@ -49,49 +51,50 @@ struct CompareDevices {
...
- *  be more than one on one node.
+ *  It relies on <tt>MPI_Get_processor_name()</tt> to get a unique identifier
+ *  of the physical node, as opposed to the logical rank of which there can
+ *  be more than one per node.
  */
-std::vector<EspressoGpuDevice> cuda_gather_gpus() {
-  int n_gpus = cuda_get_n_gpus();
-  char proc_name[MPI_MAX_PROCESSOR_NAME];
-  int proc_name_len;
+static std::vector<EspressoGpuDevice> mpi_cuda_gather_gpus_local() {
```
KaiSzuttor, line 58: why do we collect cuda devices if we only support devices on the head node?

src/core/cuda_utils.cuh
```diff
@@ -0,0 +1,80 @@
...
+#error Do not include CUDA headers in normal C++-code!!!
+#endif
+
+#include "cuda_utils.hpp"
+
+#include <cuda.h>
+
+#include <string>
+
+class cuda_runtime_error_impl : public cuda_runtime_error {
```
KaiSzuttor: not sure if the suffix is appropriate here

src/core/cuda_utils.cuh
```diff
@@ -0,0 +1,80 @@
...
+
+/** CUDA streams for parallel computing on CPU and GPU */
+extern cudaStream_t stream[1];
+
+/** Error output for memory allocation and memory copy
+ *  @param err   cuda error code
+ *  @param file  .cu file were the error took place
+ *  @param line  line of the file were the error took place
+ */
+void _cuda_safe_mem(cudaError_t err, const char *file, unsigned int line);
```
KaiSzuttor: we should not use underscores at the beginning of identifiers, IMHO

src/python/espressomd/cuda_init.pyx
```diff
@@ -52,35 +51,72 @@ cdef class CudaInitHandle:
...
+                devices = cuda_gather_gpus()
+            except RuntimeError:
+                pass
+            resources = dict()
+            for i in range(devices.size()):
+                dev = devices[i]
+                hostname = utils.to_str(dev.proc_name)
+                if hostname not in resources:
+                    resources[hostname] = {}
+                resources[hostname][dev.id] = {
```
KaiSzuttor, line 103: can't we get the device as a string directly?

testsuite/python/gpu_availability.py
```diff
@@ -17,21 +17,61 @@
...
+            with self.assertRaisesRegex(RuntimeError, error_msg):
+                self.system.cuda_init_handle.device = n_gpus + 1
+        else:
+            with self.assertRaisesRegex(RuntimeError, error_msg):
+                self.system.cuda_init_handle.device
+            with self.assertRaisesRegex(RuntimeError, error_msg):
+                self.system.cuda_init_handle.device = 0
+
+    @utx.skipIfMissingGPU()
+    def test_devices(self):
```
KaiSzuttor: what's the purpose of this test?

src/core/cuda_init.cpp
jngrad: That's how it was originally written. Should I remove the MPI functionality?

src/python/espressomd/cuda_init.pyx
jngrad: The string returned by `cudaGetDeviceProperties()` cannot be used as a key: we only collect the first 64 characters of the string returned by the CUDA function. I'm not sure why we truncate it, it was written like this 6 years ago and the commit message doesn't say anything, and there is not comment to explain it. Maybe back different CUDA versions had different char array lengths?

src/core/cuda_init.cpp
KaiSzuttor: no, i was just wondering, why we have the functionality...

testsuite/python/gpu_availability.py
jngrad: Not a good test name. It tests the `list_devices*()` methods.

src/core/cuda_init.cpp
jngrad: This functionality was introduced in the Tcl days and wasn't ported to the python interface until now. I wouldn't mind removing the MPI functionality, since we only support devices on the head nodes. It's also cumbersome to find the local MPI rank name in python to get the correct dict key...

src/python/espressomd/cuda_init.pyx
KaiSzuttor: i meant to get a string representation of ```cudaDeviceProp``` right in the core instead of building it on the python level

jngrad: Sorry, I don't get what you mean. On this line we're building a dict. Most of the device properties the user is interested in are integers.

KaiSzuttor: oh, I was missing the dict, thought we only want to print the information, sorry

KaiSzuttor (approved):

PR #4109, opened 2021-02-02
Remove manual memory management in EK code

### LINKED ISSUE #2900: Remove remaining manual memory management

jngrad:
Fixes #2900

Description of changes:
- remove `Utils::malloc()`
- remove some code duplication in LB and EK
- fix most `-Wconversion` warnings in LB and EK


src/core/grid_based_algorithms/lb_interface.cpp
```diff
@@ -52,6 +52,14 @@ struct NoLBActive : public std::exception {
   const char *what() const noexcept override { return "LB not activated"; }
 };
 
+#ifdef CUDA
```
KaiSzuttor: this is not an interface function

src/core/grid_based_algorithms/lbgpu_cuda.cu
```diff
@@ -2642,13 +2611,10 @@ void lb_get_boundary_flags_GPU(unsigned int *host_bound_array) {
...
                            lbpar_gpu.number_of_nodes * sizeof(unsigned int)));
-  /* values for the kernel call */
-  int threads_per_block = 64;
-  int blocks_per_grid_y = 4;
-  auto blocks_per_grid_x = static_cast<int>(
-      (lbpar_gpu.number_of_nodes + threads_per_block * blocks_per_grid_y - 1) /
-      (threads_per_block * blocks_per_grid_y));
-  dim3 dim_grid = make_uint3(blocks_per_grid_x, blocks_per_grid_y, 1);
+
+  unsigned const threads_per_block = 64;
```
KaiSzuttor: we could move this magic number to a global constexpr

KaiSzuttor (commented):
Minor stuff to fix, thanks @jngrad

src/core/grid_based_algorithms/lbgpu.hpp
```diff
@@ -238,9 +240,16 @@ uint64_t lb_fluid_get_rng_state_gpu();
 void lb_fluid_set_rng_state_gpu(uint64_t counter);
 uint64_t lb_coupling_get_rng_state_gpu();
 void lb_coupling_set_rng_state_gpu(uint64_t counter);
+
+/** Calculate the node index from its coordinates */
+inline unsigned int calculate_node_index(LB_parameters_gpu const &lbpar,
```
KaiSzuttor, line 245: ```Utils::ravel_index(coord, lbpar.dim)``` could probably used with slight modifications

KaiSzuttor: what is the difference between ```Utils::get_linear_index``` and ```Utils::ravel_index```? If they are doing the same, we should drop one of them.

KaiSzuttor (approved):

src/core/grid_based_algorithms/lbgpu.hpp
jngrad: `Utils::get_linear_index` is optimized for 3D grids (2 multiplications and 2 additions). `Utils::ravel_index` is only used in the `ReactionEnsemble` and `Histogram`, where we don't always work with 3D grids, and uses a `for` loop.

PR #4112, opened 2021-02-10
Remove option_if_available()

jngrad:
Make the CMake logic less brittle and more maintainable by removing the `option_if_available()` function. The user now needs to explicitly state which features or libraries have to be included in ESPResSo.


CMakeLists.txt
```diff
@@ -192,25 +191,16 @@ if(WITH_HDF5)
   # who are not familiar with the way hdf5 is distributed in Linux package
   # repositories (libhdf5-dev is the serial version).
   set(HDF5_PREFER_PARALLEL 1)
-  find_package(HDF5 "1.8" COMPONENTS C)
+  find_package(HDF5 "1.8" REQUIRED COMPONENTS C)
```
KaiSzuttor, line 194: why should this feature be required?

jngrad: The idea would be that invoking CMake with `-DWITH_XYZ=ON` guarantees that library `XYZ` gets loaded, with an error message if it's not available. The old behavior is brittle, e.g. if you need to load a module `UVW` that loads `XYZ`, should `UVW` be loaded with `REQUIRED` if `XYZ` was not requested (i.e. `-DWITH_XYZ=ON` is not passed, so it is not required but loaded if available)? What should be the error message be if `UVW` is not available? This is an issue e.g. for ScaFaCos.

KaiSzuttor: sorry, I should look at the whole PR :)

KaiSzuttor (dismissed):

KaiSzuttor (approved):

KaiSzuttor:
thanks!

PR #4114, opened 2021-02-12
Cylinder transformation and observables

### LINKED ISSUE #4023: Add orientation to cylindrical Observables
Depends on #4019.

Observables based on cylindrical coordinates have a `center` and `axis` parameters, but no `orientation`, which is required to get a well-defined phi angle when the `axis` parameter is not aligned with the x, y, z axes.

For performance reasons, when looping over particles, pre-compute the `phi0` angle with the `orientation` vector once, and use the `phi0` variable instead of the `orientation` vector in the for loop.

### LINKED ISSUE #4067: CylindricalLBFluxDensityProfileAtParticlePositions not clear or wrong
My intuition (which is the best I have with no available documentation) would tell me that ``CylindricalLBFluxDensityProfileAtParticlePositions`` would calculate, per bin, the average of [lb fluid velocity * lb density], where the probe positions come from the particle positions.
In the code, we calculate the sum of lb velocities in the bin, normalized by the volume of the bin.
Am I wrong and the intention of the observable is something different?

christophlohrmann:
Fixes #4023, fixes #4067

Description of changes:
- Adds automatic ``orientation`` to core HollowConicalFrustum
- Reworks the cylindrical observables tests
- Fixes the LB fluxdensity observable by implementing a LB density interpolation

- introduces a new object ``CylTrafoParams`` to make sure ``axis`` and ``orientation`` are always orthogonal
- all cylindrical observables use ``CylTrafoParams``


testsuite/python/tests_common.py
```diff
@@ -638,6 +633,13 @@ def gay_berne_potential(r_ij, u_i, u_j, epsilon_0, sigma_0, mu, nu, k_1, k_2):
     return 4. * epsilon * (rr**-12 - rr**-6)
 
 
+class DynamicDict(dict):
```
jngrad: looks like a cherry-pick went wrong here, we definitively don't want `DynamicDict` back ;-)

christophlohrmann: I'll look what went wrong with tests_common

src/python/espressomd/observables.py
```diff
@@ -83,10 +82,10 @@ def __init__(self, **kwargs):
             try_vectors = [np.array([1., 0., 0.]), np.array([0., 0., 1.])]
             axis = np.asarray(kwargs["axis"])
             for vec in try_vectors:
-                proj = np.dot(vec, axis / np.linalg.norm(axis))
-                if np.arccos(proj) > 10 * sys.float_info.epsilon:
-                    vec -= proj * axis
-                    kwargs["orientation"] = vec / np.linalg.norm(vec)
+                orthogonal_component = vec - np.dot(vec, axis) * axis
+                norm = np.linalg.norm(orthogonal_component)
+                if norm > 0.01:
```
KaiSzuttor: what is this magic 0.01 here? do want to check for ```sys.float_info.epsilon```, maybe?

src/shapes/include/shapes/HollowConicalFrustum.hpp
```diff
@@ -59,10 +61,17 @@ class HollowConicalFrustum : public Shape {
...
-    if ((m_orientation - (m_orientation * m_axis) * m_axis).norm() < 0.01) {
-      m_orientation = Utils::Vector3d{{0., 1., 0.}};
+    // orientation for the coordinate transformation.
+    // We need to try two different vectors in case one is parallel to axis.
+    std::list<Utils::Vector3d> try_vectors = {Utils::Vector3d{{1., 0., 0.}},
+                                              Utils::Vector3d{{0., 1., 0.}}};
+    for (auto vec : try_vectors) {
+      auto orth_component = vec - (vec * axis) * axis;
+      auto norm = orth_component.norm();
+      if (norm > 0.01) {
```
KaiSzuttor: ```std::numeric_limits<double>::epsilon()```

src/shapes/include/shapes/HollowConicalFrustum.hpp
```diff
@@ -59,10 +61,17 @@ class HollowConicalFrustum : public Shape {
   void set_axis(Utils::Vector3d const &axis) {
     m_axis = axis;
     // Even though the HCF is cylinder-symmetric, it needs a well defined phi=0
-    // orientation for the coordinate transformation. If the default orientation
-    // is too parallel to axis, choose again
-    if ((m_orientation - (m_orientation * m_axis) * m_axis).norm() < 0.01) {
-      m_orientation = Utils::Vector3d{{0., 1., 0.}};
+    // orientation for the coordinate transformation.
+    // We need to try two different vectors in case one is parallel to axis.
+    std::list<Utils::Vector3d> try_vectors = {Utils::Vector3d{{1., 0., 0.}},
```
KaiSzuttor: ```constexpr Utils::Vector<Utils::Vector3d, 2>>```

src/python/espressomd/observables.py
KaiSzuttor: why do we replicate the logic from the core here?

christophlohrmann: 1.) if the orthogonal component is 2 epsilon long and we scale it to 1, it will probably not be very orthogonal due to precision error. I chose a very generous value of 0.1 because we always have the second vector to try which will be more orthogonal. I guess using 1/sqrt(2) as a threshold will choose the vector that is the most orthogonal already and remove some arbitraryness
2.) The core logic for default orientation is for hollowconicalfrustum, the python one for observables

src/shapes/include/shapes/HollowConicalFrustum.hpp
christophlohrmann: see answer before

src/python/espressomd/observables.py
KaiSzuttor: * okay
* i think the logic is in the wrong place then?

christophlohrmann: Where would you suggest is the right place to unite it?

KaiSzuttor: algorithmically I think we could do something like: https://math.stackexchange.com/a/1681815

christophlohrmann: but that is what I am doing here :)
The Gram-Schmidt-type orthogonalisation works in whatever dimension
You just have to start with a trial vector that is not parallel

src/utils/include/utils/math/orthonormal_vec.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+#include "utils/Vector.hpp"
+#include "utils/constants.hpp"
+
+namespace Utils {
+/**
+ * @brief Return a vector that is orthonormal to vec
+ */
+template <typename T, std::size_t N>
+Vector<T, N> calc_orthonormal_vector(Vector<T, N> const &vec) {
```
KaiSzuttor, line 30: i think it's sufficient to write a version for Vector<T,3>. I'm also not sure if your version would work for an arbitrary dimensional vector.

christophlohrmann: it will
But if 3 is enough for espresso, so be it

christophlohrmann:
I know I still have to adapt the tutorial and do the documentation, I will do that as soon as I get green light on the stuff that is here

KaiSzuttor:
could you please fix the sanitizer warnings

src/python/espressomd/observables.py
```diff
@@ -68,6 +69,16 @@ def bin_centers(self):
         shape = list(map(len, edges)) + [len(edges)]
         return np.array(list(itertools.product(*edges))).reshape(shape)
 
+class CylindricalProfileObservable(ProfileObservable):
+    """
+    Base class for observables that work with cylinder coordinates
+    """
+    def __init__(self, **kwargs):
+        if "oid" not in kwargs and "cyl_trafo_params" not in kwargs:
```
KaiSzuttor: what is this "oid" check for? can't you do the following?:
``` python
def __init__(self, cyl_trafo_params=CylTrafoParams(), **kwargs):
    super().__init__(**kwargs)
```

jngrad: I think the "oid" is the object identifier for building an object from a checkpoint. Also your example doesn't use the variable `cyl_trafo_params` in the call to the parent class constructor.

KaiSzuttor:
``` python
def __init__(self, cyl_trafo_params=CylTrafoParams(), **kwargs):
    kwargs['cyl_trafo_params'] = cyl_trafo_params
    super().__init__(**kwargs)
```
?

KaiSzuttor: checkpointing does not use constructors

jngrad: My bad, the "oid" stuff was replaced by the "sip" framework, and it's only relevant in special cases like constraints (5a67109596ebd). The highlighted code is a duplication of the "oid" check I added by in the RDF observable, but it's most likely not needed. Also the "sip" documentation is wrong. I'll look into it and open a PR.

christophlohrmann: I will use @KaiSzuttor's suggestion then.
Are there general concerns about how the CylTrafoParams work or the naming and that sort of stuff?
If not I will do docs and tutorials etc

jngrad: Regarding naming, it depends on whether we plan on having other coordinate systems in the future. For example, following this pattern we would have something like `spher_trafo_params` for spherical and `cart_trafo_params` for cartesian (e.g. for a rectangular channel connecting two chambers at an angle).

Or we could name the parameter `coord_system` and give a full name to the coordinate system parameter classes, and throw an error if a `SphericalTransformationParameters` object is passed to a cylindrical observable instead of a `CylindricalTransformationParameters` object. That would be my personal preference.

KaiSzuttor: I think that the generic name ```coord_system``` does not directly map to transformation parameters. However, using @jngrad 's idea, I suggest to use ```transform_params``` and have ```CylindricalTransformationParameters```. What do you think?

jngrad: I'm fine with the new names and the suggested change in https://github.com/espressomd/espresso/pull/4114#discussion_r588326876.

KaiSzuttor:
@christophlohrmann could you please take the final steps here? We can then merge this soon I think...

src/python/espressomd/observables.py
christophlohrmann: ok so the observables get the member ``transform_params`` and the CylTrafoParams become ``CylindricalTransformationParameters``.

doc/sphinx/analysis.rst
```diff
@@ -411,10 +411,22 @@ or bin edges for the axes. Example::
...
+Observables based on cylindrical coordinates are also available.
+They require special parameters if the cylindrical coordinate system is non-standard, e.g. if you want the origin of the cylindrical coordinates to be at a special location of the box or if you want to make use of symmetries along an axis that is not parallel to the z-axis.
+For this purpose, use :class:`espressomd.math.CylindricalTransformationParameters` to create a consistent set of the parameters needed. Example::
+
+    import espressomd.math
+    
+    # shifted and rotated cylindrical coordinates
+    cyl_trafo_params = espressomd.math(center=[5.0, 5.0, 0.0], 
+                                       axis=[0, 1, 0],
+                                       orientation=[0, 0, 1])
```
jngrad:
```suggestion
    cyl_trafo_params = espressomd.math.CylindricalTransformationParameters(
        center=[5.0, 5.0, 0.0], axis=[0, 1, 0], orientation=[0, 0, 1])
```

src/python/espressomd/observables.py
```diff
@@ -69,6 +70,18 @@ def bin_centers(self):
...
 
+class CylindricalProfileObservable(ProfileObservable):
+    """
+    Base class for observables that work with cylinder coordinates
+    """
+
+    def __init__(
+            self, trafo_params=CylindricalTransformationParameters(), **kwargs):
+        # Provide default transformation parameters if not user-provided
+        kwargs['trafo_params'] = trafo_params
```
jngrad: here and elsewhere, did you mean to write `transform_params` instead of `trafo_params` for the optional argument?

testsuite/python/observable_cylindrical.py
```diff
@@ -46,180 +49,153 @@ class TestCylindricalObservable(ut.TestCase):
...
-        # Choose the cartesian velocities such that each particle gets the same
-        # v_r, v_phi and v_z, respectively.
-        self.v_r = .75
-        self.v_phi = 2.5
-        self.v_z = 1.5
-        for i in self.params['ids']:
+    def calc_ellipsis_pos_vel(
+            self, n_part, z_min, z_max, semi_x=1., semi_y=1.):
+        """
+        Calculate positions on an elliptical corkscrew line. Calculate cartesian velocities that lead to a constant velocity in cylindrical coordinates
```
jngrad: here and in other python tests, please wrap docstrings to 80 characters per line (the formatting script is not allowed to wrap strings because adding newlines in a string may change the script behavior)

testsuite/python/observable_cylindricalLB.py
```diff
@@ -18,210 +18,153 @@
...
 
+    lb_params = {'agrid': 1.,
+                 'dens': 1.2,
+                 'visc': 2.7,
+                 'tau': 0.1,
+                 }
+    cyl_trafo_params = espressomd.math.CylindricalTransformationParameters(center=3 * [7],
+                                                                           axis=[
+                                                                               1, 0, 0],
+                                                                           orientation=[0, 0, 1])
```
jngrad:
```suggestion
    cyl_trafo_params = espressomd.math.CylindricalTransformationParameters(
        center=3 * [7], axis=[1, 0, 0], orientation=[0, 0, 1])
```

testsuite/python/observable_cylindricalLB.py
```diff
@@ -269,42 +212,69 @@ def test_cylindrical_lb_profile_interface(self):
...
-        np.testing.assert_array_equal(np.copy(observable.axis), [6, 5, 4])
-        # check sampling_density
-        self.assertEqual(observable.sampling_density, 2)
-        observable.sampling_density = 3
-        self.assertEqual(observable.sampling_density, 3)
+        # check center, axis, orientation
+        ctp = espressomd.math.CylindricalTransformationParameters(center=[1, 2, 3],
+                                                                  axis=[
+                                                                      0, 1, 0],
+                                                                  orientation=[0, 0, 1])
```
jngrad:
```suggestion
        ctp = espressomd.math.CylindricalTransformationParameters(
            center=[1, 2, 3], axis=[0, 1, 0], orientation=[0, 0, 1])
```

testsuite/python/observable_cylindrical.py
```diff
@@ -266,15 +242,16 @@ def test_cylindrical_pid_profile_interface(self):
...
-        np.testing.assert_array_equal(np.copy(observable.center), [3, 2, 1])
-        # check axis
-        np.testing.assert_array_equal(np.copy(observable.axis), params['axis'])
-        observable.axis = [6, 5, 4]
-        np.testing.assert_array_equal(np.copy(observable.axis), [6, 5, 4])
+        # check center, axis, orientation
+        ctp = espressomd.math.CylindricalTransformationParameters(center=[1, 2, 3],
+                                                                  axis=[
+                                                                      0, 1, 0],
+                                                                  orientation=[0, 0, 1])
```
jngrad:
```suggestion
        ctp = espressomd.math.CylindricalTransformationParameters(
            center=[1, 2, 3], axis=[0, 1, 0], orientation=[0, 0, 1])
```

testsuite/python/tests_common.py
```diff
@@ -167,11 +167,28 @@ def transform_vel_from_cartesian_to_polar_coordinates(pos, vel):
...
     return np.dot(A.transpose(), vec)
 
 
+def rodrigues_rot(vec, axis, angle):
+    """
+    https://en.wikipedia.org/wiki/Rodrigues%27_rotation_formula#Statement
+    """
+    axis /= np.linalg.norm(axis)
+    return np.cos(angle) * vec + np.sin(angle) * np.cross(axis,
+                                                          vec) + (1 - np.cos(angle)) * np.dot(axis, vec) * axis
```
jngrad:
```suggestion
    return np.cos(angle) * vec + np.sin(angle) * np.cross(axis, vec) + \
        (1 - np.cos(angle)) * np.dot(axis, vec) * axis
```

jngrad (commented):
a few formatting issues

testsuite/python/tests_common.py
KaiSzuttor: In general I think it is not feasible to manually format the source code and we should also not spend too much time on it. Maybe we should have a look at https://github.com/psf/black.

src/python/espressomd/observables.py
christophlohrmann: I was thinking that transform_params sounds a bit like the params get transformed

KaiSzuttor: no, I think it's fine 😊

jngrad: I agree, `transform_params` is transparent and the docstring is clear about its role

christophlohrmann: ok

src/core/observables/CylindricalDensityProfile.hpp
```diff
@@ -41,7 +41,8 @@ class CylindricalDensityProfile : public CylindricalPidProfileObservable {
 
     for (auto p : particles) {
       histogram.update(Utils::transform_coordinate_cartesian_to_cylinder(
-          folded_position(traits.position(p), box_geo) - center, axis));
+          folded_position(traits.position(p), box_geo) - trafo_params->center(),
+          trafo_params->axis(), trafo_params->orientation()));
```
KaiSzuttor: since you wanted to make sure we all agree on the naming... I think it would not harm the readability to use the same parameter name as in the python interface 😁 "trafo" is also a german abbreviation IIRC

PR #4120, opened 2021-02-24
non bonded interactions refactoring draft

KaiSzuttor:
Addresses #3060, and the non-bonded version of #3019

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -29,6 +29,7 @@ if(WITH_TESTS)
 endif(WITH_TESTS)
 
 add_subdirectory(utils)
+add_subdirectory(interactions)
 add_subdirectory(shapes)
 add_subdirectory(core)
 add_subdirectory(config)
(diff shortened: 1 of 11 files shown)


fweik:
I think this draft touches interesting design questions that are worth discussing. *BUT* upon inspection I think that this touches
on at least three separate issues that can and should be discussed independently. Those are

1. The storage of the interaction data
2. The design of the actual single interactions
3. And the interaction data type per pair.

I think splitting those will actually make the discussion easier due to the limited scope of each.

KaiSzuttor:
> I think this draft touches interesting design questions that are worth discussing. _BUT_ upon inspection I think that this touches
> on at least three separate issues that can and should be discussed independently. Those are
> 
> 1. The storage of the interaction data
> 2. The design of the actual single interactions
> 3. And the interaction data type per pair.
> 
> I think splitting those will actually make the discussion easier due to the limited scope of each.

I have tried to describe the ideas in the [discussions section](https://github.com/espressomd/espresso/discussions)

PR #4121, opened 2021-02-24
Script interface maintenance

jngrad:
Description of changes:
- fix regressions introduced in the last script interface refactor
- add python `dict` to the list of types allowed in the script interface (gets converted to `std::unordered_map<int, Variant>`)
- make coupling-based fields throw an error when calling a coupling parameter setter, instead of silently ignoring the new value (the coupling object is const), and take a `dict` instead of having the user manually cast `dict_items` to `list` (API change)
- improve testing of field-based constraints and shapes

### DIFF
--- a/src/script_interface/get_value.hpp
+++ b/src/script_interface/get_value.hpp
@@ -160,6 +160,26 @@ template <> struct get_value_helper<std::vector<double>, void> {
   }
 };
 
+template <typename K, typename T>
+struct GetMapOrEmpty : boost::static_visitor<std::unordered_map<K, T>> {
+  /* Catch all case -> wrong type. */
+  template <typename U> std::unordered_map<K, T> operator()(U const &) const {
+    throw boost::bad_get{};
+  }
+
+  /* Standard case, correct type */
+  std::unordered_map<K, T> operator()(std::unordered_map<K, T> const &v) const {
+    return v;
+  }
+};
+
+/* std::unordered_map cases */
+template <> struct get_value_helper<std::unordered_map<int, Variant>, void> {
+  std::unordered_map<int, Variant> operator()(Variant const &v) const {
+    return boost::apply_visitor(GetMapOrEmpty<int, Variant>{}, v);
+  }
+};
+
 /* This allows direct retrieval of a shared_ptr to the object from
    an ObjectId variant. If the type is a derived type, the type is
    also checked.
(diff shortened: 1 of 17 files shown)


src/script_interface/get_value.hpp
```diff
@@ -256,6 +256,21 @@ template <typename T>
 void set_from_args(T &dst, VariantMap const &vals, const char *name) {
   dst = get_value<T>(vals, name);
 }
+
+/**
+ * @brief Convert a list of 2-tuples to an unordered map.
+ * Typically used for python dict objects, which have to be passed as
```
KaiSzuttor: i think we should not refer to the python interface here

src/script_interface/get_value.hpp
```diff
@@ -256,6 +256,21 @@ template <typename T>
...
+ * lists of tuples using <tt>list(argument.items())</tt>.
+ */
+template <typename K, typename T>
+std::unordered_map<K, T> get_map(const std::vector<Variant> &vv) {
+  std::unordered_map<K, T> ret{};
+  for (auto const &kv : vv) {
+    auto const kv_vec = get_value<std::vector<Variant>>(kv);
```
KaiSzuttor: i'm not sure if i understand this correctly... is the input a vector of a variant and a possible type of this variant is a vector of a variant?

jngrad: We use this `get_map()` function when we know the Variant object is packing a python `dict` which has been converted to a `list`. Each item in the `std::vector<Variant>` is itself a `std::vector<Variant>` of size 2, the first item is of type `K` (typically  `int`) and the second item is of type `T`.

I've tried to update the `Variant` object to pack a `std::unordered_map<int, Variant>` and avoid using this construction, but a template error message resisted for 1 hour...

jngrad: ok I've finally got the `dict <-> std::unordered_map<int, Variant>` conversion to work, the code above can be removed if we change the API to pass a `dict` to coupling-based fields instead of a flattened `dict_items` object of the form `[(pid1, scale1), (pid2, scale2), ...]`.

src/shapes/unit_tests/NoWhere_test.cpp
```diff
@@ -0,0 +1,63 @@
...
+bool check_distance_function(const Shapes::Shape &s) {
+  constexpr auto infinity = std::numeric_limits<double>::infinity();
+  for (int i = 0; i < 10; i++) {
+    for (int j = 0; j < 10; j++) {
+      for (int k = 0; k < 10; k++) {
+        Utils::Vector3d pos = {i * 0.5, j * 0.5, k * 0.5};
+        Utils::Vector3d dist{};
+        double d;
+
+        s.calculate_dist(pos, d, dist);
```
KaiSzuttor: not sure if it is necessary to check more than two independent positions here

src/shapes/unit_tests/NoWhere_test.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include <boost/test/unit_test.hpp>
+
+#include <shapes/NoWhere.hpp>
+#include <shapes/Shape.hpp>
+
+#include <utils/Vector.hpp>
+
+#include <limits>
+
+bool check_distance_function(const Shapes::Shape &s) {
```
KaiSzuttor: the function name could be more expressive, *e.g.* ```dist_is_always_inf```

KaiSzuttor (approved):

PR #4130, opened 2021-03-03
Fix Union shape

jngrad:
Fixes partially #4128

Description of changes:
- fix Union shape iterator
- skip long tests on macOS sanitizers build

### DIFF
--- a/.github/workflows/push_pull.yml
+++ b/.github/workflows/push_pull.yml
@@ -22,6 +22,7 @@ jobs:
         with:
           asan: false
           ubsan: false
+          check_skip_long: false
 
   sanitizer_check:
     runs-on: macos-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/push_pull.yml
```diff
@@ -38,6 +39,7 @@ jobs:
         with:
           asan: true
           ubsan: true
+          check_skip_long: true
```
KaiSzuttor, line 42: shouldn't it be the other way around? we don't have any time constraints for the nightly build, right?

jngrad: The Wang-Landau test can take an arbitrary amount of time depending on the load of the runner, and it can take so much resources that other tests can end up timing out too. We disabled it in our own GitLab sanitizer build, including in the nightly builds.

KaiSzuttor: but shouldn't we fix the issue instead of the symptom?

jngrad: We already did our best to reduce the runtime of the Wang-Landau test without sacrificing the statistical power of the check. The same is true for all statistical tests that are skipped by this command. The sanitizers are only needed on regular tests.

KaiSzuttor (approved):

PR #4136, opened 2021-03-04
Core: Skip ibm bond loop if no ibm objects

RudolfWeeber:


src/core/immersed_boundary/ImmersedBoundaries.hpp
```diff
@@ -27,7 +27,8 @@
 
 class ImmersedBoundaries {
 public:
-  ImmersedBoundaries() : MaxNumIBM(1000), VolumeInitDone(false) {
+  ImmersedBoundaries()
+      : MaxNumIBM(1002), VolumeInitDone(false), BoundariesFound(false) {
```
KaiSzuttor: what is that magic number and why did you increase it by 2?

RudolfWeeber:
IIRC, Magic number is the max number of immersed objects, the implementation can handle. Don't remember the details, since I didn't write the implementation.
Change to 1002 was unintentional, although I have no idea how I managed that, even using vim.

src/core/immersed_boundary/ImmersedBoundaries.hpp
```diff
@@ -40,8 +41,8 @@ class ImmersedBoundaries {
 
   const int MaxNumIBM;
   std::vector<double> VolumesCurrent;
-  bool VolumeInitDone = false;
-  bool BoundariesFound = false;
+  bool VolumeInitDone;
```
KaiSzuttor, line 44: > Use of an indeterminate value obtained by default-initializing a non-class variable of any type is undefined behavior

https://en.cppreference.com/w/cpp/language/default_initialization

RudolfWeeber:
I’m confused. Both bools are in the initializer list of the constructor. Why shouldn’t that get executed?

KaiSzuttor (approved):

KaiSzuttor:
> I’m confused. Both bools are in the initializer list of the constructor. Why shouldn’t that get executed?

never mind, i overlooked the change in the constructor

PR #4138, opened 2021-03-05
Script interface maintenance

jngrad:
Description of changes:
* remove unused `oid` argument guard in RDF class
* document how to extend the `Variant` with extra types
* add include guard for `boost::serialization` bug in Boost 1.74.0


RudolfWeeber:
What's the rdf change?

jngrad:
The `oid` (aka object id) is a relic from a previous generation of the python script interface. It has been superseded with the `sip` keyword argument. The `sip` is however irrelevant for observables, it's currently only used for shape-based constraints, where the user is allowed to pass a `Shape` object to the `Constraints` class instead of a `ShapeBasedConstraint`, in which case the script interface will instantiate a `ShapeBasedConstraint` from that `Shape` object on behalf of the user via the `sip` keyword argument. This has surfaced in the context of https://github.com/espressomd/espresso/pull/4114#discussion_r588342816 and caused some confusion.

KaiSzuttor (approved):

PR #4148, opened 2021-03-09
maintainer: run step if case of failing pipeline.

KaiSzuttor:
Description of changes:
- circumvent early exit in case of failure


KaiSzuttor:
to actually see the changes in the pipeline, you have to check https://github.com/KaiSzuttor/espresso/runs/2066385804?check_suite_focus=true

jngrad (dismissed):

jngrad (approved):

PR #4152, opened 2021-03-10
Refactor cylindrical coordinates transformation and dependent observables

### LINKED ISSUE #4023: Add orientation to cylindrical Observables

### LINKED ISSUE #4067: CylindricalLBFluxDensityProfileAtParticlePositions not clear or wrong

KaiSzuttor:
Fixes #4023, fixes #4067, closes #4114

Description of changes:
- add `orientation` to `HollowConicalFrustum`
- add an optional argument to control the orientation of cylindrical observables
- fix the LB fluxdensity observable by implementing a LB density interpolation


christophlohrmann (dismissed):

christophlohrmann:
A very good PR. Naming is spot on!

jngrad (approved):

RudolfWeeber:
Thanks alot! I guess this turned out a bigger project than initially planned.

PR #4161, opened 2021-03-16
Refactor bonded interaction parameter handling

### LINKED ISSUE #3019: Bond_parameters should be a variant

biermanncarl:
Fixes #3019, partial fix for #3680, closes #4066

Description of changes:
- Bonded parameters are now stored inside a ```boost::variant```
- Most bond types are now in self-contained header files (dependency reversals compared to former state)
- Most bond kernels have become members of the respective structs
- Each bond type is represented by its own C++ type (the three tabulated types are now separate, too), which is important because C++ type is now equivalent to bond type
- Removed the former bond type identification scheme from the C++ core
- Separate bonded IA cutoff from non-bonded IA cutoff
- Skip bond loop independently from non-bonded loop


src/config/config.hpp
```diff
@@ -90,6 +90,11 @@
 #define TINY_OIF_ELASTICITY_COEFFICIENT 1e-10
 #endif
 
+/** Maximal number of soft particles per IBM volcons bond (?) */
+#ifndef IBM_MAX_NUM
+#define IBM_MAX_NUM 1000
```
KaiSzuttor: why?

src/core/bonded_interactions/angle_cosine.hpp
```diff
@@ -67,18 +98,16 @@ angle_cosine_force(Utils::Vector3d const &r_mid, Utils::Vector3d const &r_left,
...
+    Utils::Vector3d const &r_right) const {
   auto const vectors = calc_vectors_and_cosine(r_mid, r_left, r_right, true);
   auto const cos_phi = std::get<4>(vectors);
   auto const sin_phi = sqrt(1 - Utils::sqr(cos_phi));
-  auto const cos_phi0 = iaparams.p.angle_cosine.cos_phi0;
-  auto const sin_phi0 = iaparams.p.angle_cosine.sin_phi0;
-  auto const k = iaparams.p.angle_cosine.bend;
+  auto const cos_phi0 = this->cos_phi0;
+  auto const sin_phi0 = this->sin_phi0;
+  auto const k = this->bend;
```
KaiSzuttor: these variables are not needed

src/core/bonded_interactions/angle_cossquare.hpp
```diff
@@ -27,32 +27,59 @@
...
+Angle_cossquare_bond_parameters::angle_force(
+    Utils::Vector3d const &r_mid, Utils::Vector3d const &r_left,
+    Utils::Vector3d const &r_right) const {
 
-  auto forceFactor = [&iaparams](double const cos_phi) {
-    auto const cos_phi0 = iaparams.p.angle_cossquare.cos_phi0;
-    auto const k = iaparams.p.angle_cossquare.bend;
+  auto forceFactor = [this](double const cos_phi) {
+    auto const cos_phi0 = this->cos_phi0;
+    auto const k = this->bend;
```
KaiSzuttor: these variables are not needed

src/core/bonded_interactions/angle_cossquare.hpp
```diff
@@ -63,16 +90,14 @@ angle_cossquare_force(Utils::Vector3d const &r_mid,
...
-                                     Bonded_ia_parameters const &iaparams) {
+inline double Angle_cossquare_bond_parameters::angle_energy(
+    Utils::Vector3d const &r_mid, Utils::Vector3d const &r_left,
+    Utils::Vector3d const &r_right) const {
   auto const vectors = calc_vectors_and_cosine(r_mid, r_left, r_right, true);
   auto const cos_phi = std::get<4>(vectors);
-  auto const cos_phi0 = iaparams.p.angle_cossquare.cos_phi0;
-  auto const k = iaparams.p.angle_cossquare.bend;
+  auto const cos_phi0 = this->cos_phi0;
+  auto const k = this->bend;
```
KaiSzuttor: here and elsewhere, please remove those variables, also you do not need to use "this" explicitely, I think we use  the convention of using a "m_" prefix for member variables

src/core/bonded_interactions/bonded_coulomb.hpp
```diff
@@ -29,51 +29,67 @@
...
 #include <cmath>
 
-/** Set the parameters for the bonded Coulomb potential
- *
- *  @retval ES_OK on success
- *  @retval ES_ERROR on error
- */
-int bonded_coulomb_set_params(int bond_type, double prefactor);
+/** Parameters for %Coulomb bond Potential */
+struct Bonded_coulomb_bond_parameters {
```
KaiSzuttor:
```suggestion
struct BondedCoulombParameters {
```

src/core/bonded_interactions/bonded_interaction_data.cpp
```diff
@@ -28,68 +28,29 @@
...
+};
 
 double maximal_cutoff_bonded() {
-  auto const max_cut_bonded =
-      boost::accumulate(bonded_ia_params, -1.,
-                        [](auto max_cut, Bonded_ia_parameters const &bond) {
-                          return std::max(max_cut, cutoff(bond.type, bond.p));
-                        });
+  auto const max_cut_bonded = boost::accumulate(
+      bonded_ia_params, -1.,
```
KaiSzuttor: why do you initialize with -1 and not 0?

src/core/bonded_interactions/bonded_interaction_data.cpp
```diff
@@ -99,31 +60,28 @@ double maximal_cutoff_bonded() {
   return (any_dihedrals) ? 2 * max_cut_bonded : max_cut_bonded;
 }
 
+int bonded_ia_params_num_partners(int bond_id) {
```
KaiSzuttor:
```suggestion
int no_of_partners(int bond_id) {
```

src/config/config.hpp
biermanncarl: It's basically a relocation of a constant from [here](https://github.com/espressomd/espresso/pull/4161/files#diff-17cc54295fcd4d75a78d8790cb5c73e2330c7f30d7a504f86ca89c16513bf527L30)

src/core/bonded_interactions/angle_cossquare.hpp
biermanncarl: Well, it's a tradeoff. Either we change the name of ALL bonded parameter members, or we keep ```this->```. It's actually only needed in the constructor, and nowhere else if I am not mistaken. I'm unsure what is the better solution.

KaiSzuttor: ```this``` is only ever needed for disambiguation. And if you are touching those variables we should change the naming to be consistent with our conventions

jngrad: There is no consensus in core team yet for adding a prefix `m_` to public members. The `this->` dereference however can be removed in all bonded force/energy kernels.

src/core/bonded_interactions/bonded_interaction_data.cpp
jngrad: I think `-1.` is the magic value that will be replaced by the more expressive macro `BONDED_INACTIVE_CUTOFF` in #4066.

src/core/bonded_interactions/bonded_interaction_data.hpp
```diff
@@ -23,349 +23,114 @@
...
+/** Field containing the parameters of the bonded ia types */
+extern std::vector<Bonded_ia_parameters> bonded_ia_params;
 
-/** Parameters for n-body tabulated potential (n=2,3,4). */
-struct Tabulated_bond_parameters {
-  TabulatedBondedInteraction type;
-  TabulatedPotential *pot;
+/** These following functions are necessary to hide boost::variant from Cython.
+ *  It can't be used there because Cython doesn't support variadic templates.
+ */
```
jngrad: It's unclear to which function this doxygen block applies. You can use `@name` and delimiters `/**@{*/` and `/**@}*/` to convey that information ([example](https://github.com/espressomd/espresso/blob/a20d99b94f9ee8c798afdc56f83d51858206beb2/src/core/thermostat.hpp#L58-L67)).

src/core/bonded_interactions/bonded_interactions.dox
```diff
@@ -24,131 +24,164 @@
...
- *  * core:
- *   - define a data structure for the interaction parameters (prefactors, etc.)
- *   - write a setter function, which takes the parameters of the interactions
- *     and stores them in the interaction data structure
- *   - write functions for calculating force and energy
- *   - add calls to the force and energy calculation functions to the force
- *     calculation in the integration loop as well as to energy and
- *     pressure tensor analysis
+ *  * C++ core:
+ *    - Create a new header filefor your new bond type, preferably inside the
```
jngrad:
```suggestion
 *    - Create a new header file for your new bond type, preferably inside the
```

src/core/bonded_interactions/bonded_interactions.dox
```diff
@@ -24,131 +24,164 @@
...
- *  * \c bonded_ia_params[bond_type].num is the number of particles involved
- *    in the bond, minus 1, i.e. 1 for a pairwise bonded potential such as the
- *    FENE bond.
- *  * The parameters for the individual bonded interaction go to the member of
- *    @ref Bond_parameters for your interaction defined in the previous step.
- *    For the FENE bond, this would be:
+ *    Replace all the <tt>ar &</tt> commands with the new bond's parameters.
+ *    Every data member of your struct needs to be transmitted. This template
+ *    function can be made private, but then the following line needs to be
+ *    added before for the serialization routine to be able to access it:
```
jngrad: serialization methods are typically private, therefore I would rephrase it like so:
```suggestion
 *    Every data member of your struct needs to be transmitted. This template
 *    function is private, hence the following line needs to be added before:
```

src/core/bonded_interactions/bonded_interactions.dox
```diff
@@ -24,131 +24,164 @@
...
- *  * The functions calculating force and energy can make use of a
- *    pre-calculated distance vector (\p dx) pointing from particle 2 to
- *    particle 1, that takes periodic boundary conditions into account.
- *  * The force and energy functions cannot alter the particle states.
- *
- *  @subsection bondedIA_new_integration Including the bonded interaction in \
- *                                       the force calculation and the energy \
- *                                       and pressure analysis
+ *    Refer to @ref fene.hpp as an example.
+
```
jngrad:
```suggestion
```

src/core/bonded_interactions/bonded_interactions.dox
```diff
@@ -503,3 +533,58 @@
  *  @f$ \theta_{ijk} = \pi @f$.
  *
  */
+
+/*
```
jngrad: This break transformed the next block into a regular comment instead of a doxygen comment.
```suggestion
 *
```

src/python/espressomd/interactions.pxd
```diff
@@ -298,45 +298,97 @@ IF SOFT_SPHERE:
...
-        int type
-        TabulatedPotential * pot
+    cdef extern from "bonded_interactions/bonded_interaction_data.hpp":
+        # Parameters for n-body tabulated potential (n=2,3,4).
+        cdef cppclass Tabulated_bond_parameters:
+            Tabulated_bond_parameters()
+            Tabulated_bond_parameters(double min, double max,
+                                      const vector[double] & energy,
+                                      const vector[double] & force) except +
+            int type
```
jngrad:
```suggestion
```

src/python/espressomd/interactions.pxd
```diff
@@ -298,45 +298,97 @@ IF SOFT_SPHERE:
...
+            TabulatedPotential * pot
```
jngrad: Should this raw pointer become a `shared_ptr[TabulatedPotential] pot` with the appropriate `from libcpp.memory cimport shared_ptr` at the top of the Cython file?

src/config/config.hpp
jngrad: This constant had to be moved out to help with dependency inversion, IIRC.

src/core/bonded_interactions/bonded_interaction_data.cpp
jngrad: The `bonded_ia_params_` prefix is used in this function and the ones above and below because it's acting on the variant called `bonded_ia_params`. This will become clear once we add the correct doxygen group delimiters. We could even move these functions to `src/script_interface`.

src/core/bonded_interactions/bonded_coulomb.hpp
jngrad: All bonded parameter structs are now CamelCase.

src/core/bonded_interactions/bonded_interaction_data.cpp
jngrad: Function moved to `src/script_interface/interactions/bonded.hpp` and modified to call a new function with signature `number_of_partners(Bonded_ia_parameters const &iaparams)` in `src/core/bonded_interactions/bonded_interaction_data.cpp`.

jngrad (dismissed):
good to go from my side

src/core/bonded_interactions/angle_cosine.hpp
```diff
@@ -27,37 +27,65 @@
...
+
+  double cutoff() const { return -1.; }
+
+  static constexpr int num = 2;
+
+  AngleCosineBondParameters() = default;
+  AngleCosineBondParameters(double bend, double phi0);
+
+  std::tuple<Utils::Vector3d, Utils::Vector3d, Utils::Vector3d>
+  angle_force(Utils::Vector3d const &r_mid, Utils::Vector3d const &r_left,
```
KaiSzuttor: i'm not sure about ```angle_force```... wouldn't ```forces``` more precisely reflect that there are multiple forces returned and calculated?

src/core/bonded_interactions/angle_cosine.hpp
```diff
@@ -67,21 +95,17 @@ angle_cosine_force(Utils::Vector3d const &r_mid, Utils::Vector3d const &r_left,
...
  *  @param[in]  r_left    Position of first/left particle.
  *  @param[in]  r_right   Position of third/right particle.
- *  @param[in]  iaparams  Bonded parameters for the angle interaction.
  */
-inline double angle_cosine_energy(Utils::Vector3d const &r_mid,
-                                  Utils::Vector3d const &r_left,
-                                  Utils::Vector3d const &r_right,
-                                  Bonded_ia_parameters const &iaparams) {
+inline double
+AngleCosineBondParameters::angle_energy(Utils::Vector3d const &r_mid,
```
KaiSzuttor:
```suggestion
AngleCosineBondParameters::energy(Utils::Vector3d const &r_mid,
```

src/core/bonded_interactions/angle_harmonic.hpp
```diff
@@ -27,36 +27,58 @@
...
 #include <utils/Vector.hpp>
 #include <utils/math/sqr.hpp>
 
 #include <cmath>
 #include <tuple>
 
-/** Set parameters for the angle potential. */
-int angle_harmonic_set_params(int bond_type, double bend, double phi0);
+/** Parameters for three-body angular potential (harmonic). */
+struct AngleHarmonicBondParameters {
```
KaiSzuttor: why are these types have a ```Parameters``` suffix if they actually also implement the energy and force kernels?

src/core/bonded_interactions/bonded_coulomb_sr.hpp
```diff
@@ -42,40 +39,58 @@
...
+  double q1q2;
+
+  double cutoff() const { return -1.; }
+
+  static constexpr int num = 1;
+
+  BondedCoulombSRParameters() = default;
+  BondedCoulombSRParameters(double q1q2);
+
+  boost::optional<Utils::Vector3d> pair_force(Utils::Vector3d const &dx) const;
```
KaiSzuttor:
```suggestion
  boost::optional<Utils::Vector3d> force(Utils::Vector3d const &dx) const;
```

src/core/bonded_interactions/bonded_coulomb_sr.hpp
```diff
@@ -42,40 +39,58 @@
...
+  boost::optional<double> pair_energy(Particle const &p1, Particle const &p2,
```
KaiSzuttor:
```suggestion
  boost::optional<double> energy(Particle const &p1, Particle const &p2,
```

src/core/bonded_interactions/bonded_interaction_data.cpp
KaiSzuttor: Still don't understand this... The minimum value for the maximum cut off is zero and not some magic number.

src/core/bonded_interactions/bonded_interaction_data.hpp
```diff
@@ -23,346 +23,80 @@
...
-  BondedInteraction type;
-  /** Number of bonds (N_part - 1) for that interaction type. */
-  int num;
-  /** Interaction parameters. */
-  Bond_parameters p;
-};
+/** Variant in which to store the parameters of an individual bonded
+ *  interaction
+ */
+using Bonded_ia_parameters = boost::variant<
```
KaiSzuttor:
```suggestion
using Bonded_IA_Parameters = boost::variant<
```

src/core/bonded_interactions/bonded_tab.cpp
```diff
@@ -21,66 +21,54 @@
...
 #include <vector>
 
-int tabulated_bonded_set_params(int bond_type,
-                                TabulatedBondedInteraction tab_type, double min,
-                                double max, std::vector<double> const &energy,
-                                std::vector<double> const &force) {
-  if (bond_type < 0)
-    return ES_ERROR;
-
+Tabulated_bond_parameters::Tabulated_bond_parameters(
```
KaiSzuttor:
```suggestion
TabulatedBondParameters::TabulatedBondParameters(
```

src/core/bonded_interactions/bonded_tab.hpp
```diff
@@ -32,53 +32,112 @@
...
- *
- *  @retval ES_OK on success
- *  @retval ES_ERROR on error
- */
-int tabulated_bonded_set_params(int bond_type,
-                                TabulatedBondedInteraction tab_type, double min,
-                                double max, std::vector<double> const &energy,
-                                std::vector<double> const &force);
+/** Base class for n-body tabulated potential (n=2,3,4). */
+struct Tabulated_bond_parameters {
```
KaiSzuttor:
```suggestion
struct TabulatedBonParameters {
```

src/core/bonded_interactions/bonded_tab.hpp
```diff
@@ -32,53 +32,112 @@
...
+
+  static constexpr int num = 3;
+
+  TabulatedDihedralBondParameters() = default;
+  TabulatedDihedralBondParameters(double min, double max,
+                                  std::vector<double> const &energy,
+                                  std::vector<double> const &force);
+  boost::optional<std::tuple<Utils::Vector3d, Utils::Vector3d, Utils::Vector3d,
+                             Utils::Vector3d>>
+  dihedral_force(Utils::Vector3d const &r1, Utils::Vector3d const &r2,
```
KaiSzuttor:
```suggestion
  forces(Utils::Vector3d const &r1, Utils::Vector3d const &r2,
```

src/core/bonded_interactions/bonded_tab.hpp
```diff
@@ -32,53 +32,112 @@
...
  *  extrapolation based on the first two tabulated force values.
  *
- *  @param[in]  iaparams  Bonded parameters for the pair interaction.
  *  @param[in]  dx        %Distance between the particles.
  */
 inline boost::optional<Utils::Vector3d>
-tab_bond_force(Bonded_ia_parameters const &iaparams,
-               Utils::Vector3d const &dx) {
-  auto const *tab_pot = iaparams.p.tab.pot;
+TabulatedDistanceBondParameters::pair_force(Utils::Vector3d const &dx) const {
```
KaiSzuttor:
```suggestion
TabulatedDistanceBond::force(Utils::Vector3d const &dx) const {
```

KaiSzuttor (changes requested):
Please check that all types follow our naming convention (SnakeCase) and adjust the naming of the bond classes (remove the ```Parameters``` suffix)

KaiSzuttor (approved):

PR #4162, opened 2021-03-16
ICC maintenance

reinaual:
Description of changes:
- renamed functions/struct in the core to reflect independence of P3M
- provide parameter setter function including validation instead of writing to glubal struct directly, including testcase
- made relative dielectric permittivity a required parameter (what is the purpose of the algorithm if not provided?)
- removed typos and default params for non-optional parameters
- I don't understand the the error-estimation formula in the core, it is different from the one shown in the paper [https://doi.org/10.1063/1.3376011](https://doi.org/10.1063/1.3376011)


jngrad (approved):
LGTM

Issue #4154, opened 2021-03-11
Use Philox PRNG in the Reaction Ensemble


jngrad:
The Reaction Ensemble (RE) is one of the last consumers of the Mersenne Twister (MT) pseudo-random number generator (PRNG). The MT19937 generator is known to produce correlated data in the first 1 million generated values (Figure 4 in Panneton et al. 2006, DOI:[10.1145/1132973.1132974](https://doi.org/10.1145/1132973.1132974)), which we discard in ESPResSo (5ac12d2aeacdee). The Philox PRNG doesn't suffer from this correlation.

ESPResSo already provides methods to generate a single random value drawn from a uniform or Gaussian distribution:
```c++
uint64_t counter = 0;
uint32_t const seed = 42;
int const key_uniform = 1;
int const key_gaussian = 2;
double random_uniform_value  = Random::noise_uniform<RNGSalt::RE,  1>(counter++, seed, key_uniform)
double random_gaussian_value = Random::noise_gaussian<RNGSalt::RE, 1>(counter++, seed, key_gaussian)[0];
```

The salt `RNGSalt::RE` needs to be added to the enum class ([src/core/random.hpp#L40-L61](https://github.com/espressomd/espresso/blob/415f8d4f66bd8b1bb417a7617c14b01577834dad/src/core/random.hpp#L40-L61)) beforehand. The key decorrelates sequences with the same seed. The mean and standard deviation of the distribution may differ between MT19937 and Philox.

Once the core has been updated, the samples, tutorial and tests may need adjustments. Many of them rely on a fixed random seed, and with the new PRNG the random sequence will change. Here is a list:

```
testsuite/python/reaction_ensemble.py
testsuite/python/constant_pH.py
testsuite/python/widom_insertion.py
testsuite/python/wang_landau_reaction_ensemble.py
testsuite/scripts/samples/test_wang_landau_reaction_ensemble.py
testsuite/scripts/samples/test_reaction_ensemble_complex_reaction.py
testsuite/scripts/samples/test_reaction_ensemble.py
testsuite/scripts/samples/test_widom_insertion.py
testsuite/scripts/samples/test_grand_canonical.py
testsuite/scripts/tutorials/test_constant_pH__ideal.py
testsuite/scripts/tutorials/test_constant_pH__interactions.py
```

[...]

PR #4164, opened 2021-03-17
Reaction Ensemble refactor and unit testing

### LINKED ISSUE #4157: Improve testing of the Reaction Ensemble
At the time this ticket is written, RE code coverage is 51% for [`reaction_ensemble.hpp`](https://codecov.io/gh/espressomd/espresso/src/415f8d4f66bd8b1bb417a7617c14b01577834dad/src/core/reaction_ensemble.hpp) and 32% for [`reaction_ensemble.cpp`](https://codecov.io/gh/espressomd/espresso/src/415f8d4f66bd8b1bb417a7617c14b01577834dad/src/core/reaction_ensemble.cpp). Lines without coverage are not checked for undefined behavior and memory corruption in the automated testing. Code without coverage is also harder to maintain, because there is no way to detect regressions during bugfixes and refactoring projects.

Most RE tests under `testsuite/python` are statistical tests, i.e. they check a statistic of the RE system like the potential or heat capacity, instead of checking components of the RE code (like `testsuite/python/reaction_ensemble.py` does). These statistical tests are labeled as `long` in [`testsuite/python/CMakeLists.txt`](https://github.com/espressomd/espresso/blob/415f8d4f66bd8b1bb417a7617c14b01577834dad/testsuite/python/CMakeLists.txt) and cannot be executed in code coverage builds and sanitizers builds due to their running time.

**Making the code unit testable**

The `SingleReaction` constructor is written in a separate class (`ReactionAlgorithm`). Moving the constructor to the `SingleReaction` class (7add1c62323aef02) makes the class self-contained, and therefore unit testable ([`974e8e9`](https://github.com/espressomd/espresso/commit/974e8e9f21e2d8db49a76d9ecfbdbed1400878b8#diff-45b815ff7150960d0366bfd97f1cde005eaf9c46d309f2f59d410312c42860d8R120-R138)).

Most methods in RE classes can be made `const`, and often their arguments can be made const too (for `std::map` to be const, the setter/getter [`operator[]`](https://en.cppreference.com/w/cpp/container/map/operator_at) must be changed to the getter [`at()`](https://en.cppreference.com/w/cpp/container/map/at)). This change helps identifying testable methods and simplifies the unit test logic, since a const method cannot alter the state of the tested object.

Methods `calculate_acceptance_probability()` are implementation details and are currently `private`; switching to `protected` (acf5786d2b5) would still hide them from outside code, but would make them unit testable via public inheritance ([`3d22a1d`](https://github.com/espressomd/espresso/commit/3d22a1d8e872b94ad1b643dc15200d2635d1d24f#diff-45b815ff7150960d0366bfd97f1cde005eaf9c46d309f2f59d410312c42860d8R244-R277)).

**Writing unit tests**

By linking a Boost unit test against `Boost::mpi` and `MPI::MPI_CXX`, it is possible to instantiate a minimally working ESPResSo system via `Communication::init(mpi_env)`. In such a system, particles can be created and deleted (see unit test in 974e8e9f21e2d). However some components of the ESPResSo system might not necessarily work, e.g. observables or energy kernels.

**Writing python tests**

Attempts at reducing the runtime of the RE statistical tests without reducing their statistical power have been unfruitful. The best way forward is to come up with new python tests where individual components of RE classes can be tested, for example a test that would check the outcome of a single reaction step. Such a test needs to be fully deterministic. Maybe there is a way to set up a chemical reaction so that a Monte Carlo move is guaranteed to be accepted in the next time step, for example due to a large excess of an adduct type, to a large pH-pKa difference, or to a large potential difference.

jngrad:
Fixes #4157

Description of changes:
- make `SingleReaction` class self-contained
- make RE methods and arguments `const` whenever possible
- write unit tests for the acceptance probability of all RE classes
- write integration tests to increase code coverage of RE to 90%


src/core/reaction_ensemble.hpp
```diff
@@ -38,21 +40,41 @@
 namespace ReactionEnsemble {
 
 struct SingleReaction {
+  SingleReaction() = default;
+  SingleReaction(double gamma, std::vector<int> const &reactant_types,
+                 std::vector<int> const &reactant_coefficients,
+                 std::vector<int> const &product_types,
+                 std::vector<int> const &product_coefficients) {
+    std::copy(reactant_types.begin(), reactant_types.end(),
```
KaiSzuttor: why not just ```this->reactant_types = reactant_types```?

src/core/unit_tests/reaction_ensemble_classes_test.cpp
```diff
@@ -0,0 +1,389 @@
...
+#include <utils/math/sqr.hpp>
+
+#include <boost/mpi.hpp>
+
+#include <cmath>
+#include <limits>
+#include <memory>
+#include <stdexcept>
+#include <vector>
+
```
KaiSzuttor, line 51: all the tests need documentation

src/core/reaction_ensemble.hpp
jngrad: yes, these `std::copy` are unnecessary

src/core/unit_tests/reaction_ensemble_classes_test.cpp
jngrad: I'm working on it

jngrad: The unit test comments now show the equation of the chemical reactions and explain the different Monte Carlo schemes being tested. The particle types are also now stored in variables instead of hardcoded.

src/core/unit_tests/reaction_ensemble_classes_test.cpp
```diff
@@ -0,0 +1,499 @@
...
+    BOOST_CHECK_EQUAL(r_algo.collective_variables[0]->CV_minimum, 1.);
+    BOOST_CHECK_EQUAL(r_algo.collective_variables[0]->CV_maximum, 5.);
+    BOOST_CHECK_CLOSE(r_algo.min_boundaries_energies[0], 1., tol);
+    BOOST_CHECK_CLOSE(r_algo.min_boundaries_energies[1], 2., tol);
+    BOOST_CHECK_CLOSE(r_algo.max_boundaries_energies[0], 4., tol);
+    BOOST_CHECK_CLOSE(r_algo.max_boundaries_energies[1], 5., tol);
+    // check Wang-Landau histogram
+    {
+      r_algo.write_wang_landau_results_to_file(filename_output);
+      std::ifstream outfile;
```
KaiSzuttor: please avoid writing to disk in the tests

jngrad: This check and the corresponding RE code was changed to use a file-like object (`stringstream`) instead of an output file. The collective variable however still uses an input file that is created at the start of the test; the collective variable and Wang-Landau classes are so intertwined, that adapting them to take a file-like object would require a lot of refactoring.

src/core/reaction_ensemble.cpp
```diff
@@ -69,26 +78,22 @@ void EnergyCollectiveVariable::load_CV_boundaries(
...
-      m_current_wang_landau_system.min_boundaries_energies.begin(),
-      m_current_wang_landau_system.min_boundaries_energies.end()));
-  CV_maximum = *(std::max_element(
-      m_current_wang_landau_system.max_boundaries_energies.begin(),
-      m_current_wang_landau_system.max_boundaries_energies.end()));
+    assert(values.size() >= 2);
+    auto it = values.end();
+    wl_system.max_boundaries_energies.push_back(*(--it));
+    wl_system.min_boundaries_energies.push_back(*(--it));
+  }
```
KaiSzuttor, line 82: if you'd extract this part as a function returning the values, we could even unit test this part and get rid of the explicit file on disk dependency...

src/core/reaction_ensemble.cpp
```diff
@@ -69,26 +78,22 @@ void EnergyCollectiveVariable::load_CV_boundaries(
...
-  CV_minimum = *(std::min_element(
-      m_current_wang_landau_system.min_boundaries_energies.begin(),
-      m_current_wang_landau_system.min_boundaries_energies.end()));
-  CV_maximum = *(std::max_element(
-      m_current_wang_landau_system.max_boundaries_energies.begin(),
-      m_current_wang_landau_system.max_boundaries_energies.end()));
+    assert(values.size() >= 2);
+    auto it = values.end();
+    wl_system.max_boundaries_energies.push_back(*(--it));
+    wl_system.min_boundaries_energies.push_back(*(--it));
```
KaiSzuttor:
```suggestion
    wl_system.max_boundaries_energies.push_back(values.back());
    wl_system.min_boundaries_energies.push_back(values[values.size()-2]));
```

src/core/reaction_ensemble.cpp
```diff
@@ -69,26 +78,22 @@ void EnergyCollectiveVariable::load_CV_boundaries(
...
+  CV_maximum = *boost::range::max_element(wl_system.max_boundaries_energies);
 }
 
 /**
  * Performs a randomly selected reaction in the reaction ensemble
  */
 int ReactionAlgorithm::do_reaction(int reaction_steps) {
   for (int i = 0; i < reaction_steps; i++) {
-    int reaction_id = i_random(reactions.size());
+    int reaction_id = i_random(static_cast<int>(reactions.size()));
```
KaiSzuttor, line 93: can't you move up the RNG as an arg?

src/core/reaction_ensemble.cpp
```diff
@@ -972,18 +944,13 @@ double WangLandauReactionEnsemble::calculate_delta_degree_of_association(
...
-                                      current_collective_variable->CV_minimum) /
-                                     current_collective_variable->delta_CV) +
-                                 1); // plus 1 needed for degrees of association
-                                     // related part of histogram (think of only
-                                     // one acid particle)
+long WangLandauReactionEnsemble::get_num_needed_bins() const {
+  long needed_bins = 1;
+  // add 1 for degrees of association-related part of histogram (think of only
+  // one acid particle)
+  for (const auto &cv : collective_variables) {
```
KaiSzuttor: this is  a ```std::accumulate```

src/core/reaction_ensemble.cpp
```diff
@@ -1332,19 +1295,20 @@ void WangLandauReactionEnsemble::write_out_preliminary_energy_run_results(
...
     // use unraveled index
-    for (int i = 0; i < collective_variables.size(); i++) {
-      fprintf(pFile, "%f ",
-              unraveled_index[i] * collective_variables[i]->delta_CV +
-                  collective_variables[i]->CV_minimum);
+    for (std::size_t i = 0; i < collective_variables.size(); i++) {
+      auto const value = static_cast<double>(unraveled_index[i]) *
+                             collective_variables[i]->delta_CV +
+                         collective_variables[i]->CV_minimum;
+      fprintf(pFile, "%f ", value);
```
KaiSzuttor: this function should definitely use a ```ostream```

src/core/unit_tests/reaction_ensemble_classes_test.cpp
```diff
@@ -0,0 +1,502 @@
...
+  int const type_C = 2;
+  SingleReaction const reaction(2., {type_A}, {1}, {type_B, type_C}, {3, 4});
+  r_algo.add_reaction(reaction.gamma, reaction.reactant_types,
+                      reaction.reactant_coefficients, reaction.product_types,
+                      reaction.product_coefficients);
+
+  // exception if not enough particles
+  BOOST_CHECK_THROW(r_algo.measure_excess_chemical_potential(0),
+                    std::runtime_error);
+}
```
KaiSzuttor: this is very verbose for only checking if an exception is thrown

testsuite/python/constant_pH_stats.py
```diff
@@ -0,0 +1,119 @@
...
+        # constant pH ensemble, since the volume is totally arbitrary and does
+        # not influence the average number of protons
+        pH = ReactionEnsembleTest.pH
+        pKa = ReactionEnsembleTest.pKa
+        target_alpha = ReactionEnsembleTest.ideal_alpha(pH)
+        rel_error_alpha = abs(average_alpha - target_alpha) / target_alpha
+        # relative error
+        self.assertLess(
+            rel_error_alpha,
+            0.015,
```
KaiSzuttor, line 105: do we have any idea how the difference depends on given parameters?

KaiSzuttor (commented):
Overall looks good, it's just not the prettiest neighbourhood of espresso :P

src/core/reaction_ensemble.cpp
jngrad:
```c++
int ReactionAlgorithm::do_reaction(int reaction_steps) {
  printf("======\n");
  for (int i = 0; i < reaction_steps; i++) {
    int reaction_id = i_random(static_cast<int>(reactions.size()));
    printf("%d %zu\n", reaction_id, reactions.size());
    generic_oneway_reaction(reaction_id);
  }
  return 0;
}
```
output:
```
======
0 2
0 2
0 2
0 2
1 2
1 2
1 2
0 2
1 2
0 2
======
0 2
1 2
0 2
0 2
1 2
1 2
0 2
0 2
0 2
1 2
```

src/core/unit_tests/reaction_ensemble_classes_test.cpp
jngrad: After more refactoring, the unit test no longer depends on files.

src/core/reaction_ensemble.cpp
jngrad: done

src/core/reaction_ensemble.cpp
jngrad: done

src/core/unit_tests/reaction_ensemble_classes_test.cpp
jngrad: test removed

src/core/reaction_ensemble.cpp
KaiSzuttor: What do you want to show here?

testsuite/python/constant_pH_stats.py
jngrad: I'm not sure, and the comment about the pH-pKa difference having to be small in the test setup doesn't seem to always be true, e.g. in the faster test `constant_pH.py ` the pH-pKa difference is larger yet convergence is reached faster. The RNG sequence seems to be more important.

src/core/reaction_ensemble.cpp
jngrad: ok I misunderstood your question. Do you want to pass a RNG lambda function as argument to `do_reaction()`?

KaiSzuttor: ah sorry, maybe it was not clear... yes, i wanted the RNG to be an input from outside of the function such that we could inject the randomness

jngrad: Maybe this can be done in the context of #4154? Right now injecting the randomness wouldn't help with unit testing, because RE classes still rely on the energy functions, that cannot be executed in a unit test.

KaiSzuttor: alright

KaiSzuttor (approved):

PR #4166, opened 2021-03-19
fix clang-tidy-10 not recognized

biermanncarl:
Fixes the issue that clang-tidy-10 was not recognized by the CMake logic.


jngrad (approved):
Thanks!

PR #4171, opened 2021-03-23
LBGPU: Global Stress.

### LINKED ISSUE #4134: Implement more efficient gpu lb fluid pressure calculation
this is needed for our various rheology projects.

This is currently very unefficient.
lb_interface.cpp::lb_lbfluid_get_pressure_tensor()
copies the the stress for all nodes to the cpu and does the reduction there.

Instead:
* a trhust transform-reduce operation could do the trick: 
https://thrust.github.io/doc/group__transformed__reductions_ga0d4232a9685675f488c3cc847111e48d.html
* The transform part an be some wrapper around lbgpu_cuda.cu::calc_values_from_m which returns only the pressure
* The reduction operation will be a sum.
Thrust should launch all of that in parallel, since the transform operations for different nodes are independent and the reduciont can be done in hierarchical steps.

See lb_get_inerpolated_velocities_gpu() for inspiration. The first iterator needs to be on current_nodes. Begin and end have to be passed as pointers, as current_nodes isn't a trhust vector.
Alternatively, if the effort is reasonable, the node array could be made a thrust device vector. However, we don't want to spend too much time on refactoring the current lbgpu.

KaiSzuttor:
Fixes #4134

### DIFF
--- a/src/core/grid_based_algorithms/electrokinetics_cuda.cu
+++ b/src/core/grid_based_algorithms/electrokinetics_cuda.cu
@@ -270,8 +270,7 @@ __device__ void ek_displacement(float *dx, LB_nodes_gpu n,
   float mode[19];
 
   for (unsigned i = 0; i < 19; i++) {
-    mode[i] =
-        n.populations[i * ek_lbparameters_gpu->number_of_nodes + node_index];
+    mode[i] = n.populations[node_index][i];
   }
 
   rho += mode[0] + mode[1] + mode[2] + mode[3] + mode[4] + mode[5] + mode[6] +
--- a/src/core/grid_based_algorithms/electrokinetics_cuda.cu
+++ b/src/core/grid_based_algorithms/electrokinetics_cuda.cu
@@ -349,8 +349,9 @@ __device__ void ek_diffusion_migration_lbforce_linkcentered_stencil(
     force = flux * ek_parameters_gpu->T * ek_parameters_gpu->agrid * D_inv;
     force *= force_conv;
 
-    atomicAdd(&node_f.force_density[index], force * 0.5f);
-    atomicAdd(&node_f.force_density[neighborindex[EK_LINK_U00]], force * 0.5f);
+    atomicAdd(&node_f.force_density[index][0], force * 0.5f);
+    atomicAdd(&node_f.force_density[neighborindex[EK_LINK_U00]][0],
+              force * 0.5f);
   } else {
     force = -1.0f * ek_parameters_gpu->valency[species_index] *
             (((cufftReal *)ek_parameters_gpu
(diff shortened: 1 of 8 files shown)


mkuron:
Here's a patch that fixes it:
```c++
diff --git a/src/utils/include/utils/get.hpp b/src/utils/include/utils/get.hpp
index 41bdf379e..85711fe71 100644
--- a/src/utils/include/utils/get.hpp
+++ b/src/utils/include/utils/get.hpp
@@ -21,12 +21,21 @@
 
 #include <cstddef>
 #include <tuple>
+#if defined(__CUDACC__)
+#include <thrust/tuple.h>
+#endif
 
 namespace Utils {
 template <std::size_t I, typename T> auto get(const T &v) {
   return std::get<I>(v);
 }
 
+#if defined(__CUDACC__)
+template <std::size_t I, typename... T> auto get(const thrust::tuple<T...> &v) {
+  return thrust::get<I>(v);
+}
+#endif
+
 template <class T> struct tuple_size : std::tuple_size<T> {};
 
 template <std::size_t I, class Tuple>
```

RudolfWeeber (approved):

pkreissl:
The new code, although faster, results in values that differ quite a bit from reference values obtained with the old implementation (e.g.: ed96b9220d0), that is on the order of 1e-2 after only 2 time steps of integration:

```
First tensor
[[ 8.37624901e+02 -3.11898913e-02 -2.13092766e-02]
 [-3.11898913e-02  8.37578210e+02 -1.95547613e-02]
 [-2.13092766e-02 -1.95547613e-02  8.37608834e+02]]
should be
[[ 8.37976466e+02 -3.64415686e-02 -1.62209976e-02]
 [-3.64415686e-02  8.37927312e+02 -2.11635250e-02]
 [-1.62209976e-02 -2.11635250e-02  8.37960618e+02]]
```
I would expect SOME deviation for the gpu code, however this seems to be rather large. Are deviations on that level expected?
Ah, my simulation setup is `l_box=50, time_step=0.1, agrid=0.5, dens=1, visc-1.8, kT-0.7`. If possibly too large, I can also post my test script.

KaiSzuttor:
the summation is now with floats vs double before. concerning the deviation to reference data,  I thought you are the expert 😉

pkreissl:
Sure, that's why I definitely do expect some deviations. But really I am not too sure how large a deviation one would expect for such a system when switching from double to single float summation -- maybe @mkuron can comment, for such a question probably he is more expert than me ;)
It is only that the off-diagonal elements are the one with the large (?!) deviations, at the same time the ones I am interested in... I am currently sampling to check against viscosity, If that is nicely reproduced, I guess we are fine?

mkuron:
You're adding up lots of small stresses to make one big stress, so deviations at single precision are certainly expected. However, this is a reduction, so stresses should be summed hierarchically. If stress is rather homogeneous in your simulation, you would thus not lose precision. However, if you have regions of very low and very high stress, the low stresses can get lost completely in the summation.

RudolfWeeber:
Please obtain the standard deviation of the distribution of single node pressures in a single time step. 

 

If floating point accuracy is really an issue, the summation can be switched to double precision, I guess. (Changing the return type of the single node pressure to function and the result vector to double).

Issue #3958, opened 2020-10-20
Error estimation in ES + tutorial


RudolfWeeber:
@cbiermann with @schlaicha and @RudolfWeeber

biermanncarl:
The tutorial side of this issue is being worked on in [this](https://github.com/biermanncarl/espresso/tree/error-analysis-tutorial) branch. I am waiting for a review by someone who has experience in this area, before I move on to the auto-covariance analysis part.

KaiSzuttor:
can't you just open a draft pull request and reference this issue?

PR #2382, opened 2018-11-22
Rotation cleanup

RudolfWeeber:

PR #4174, opened 2021-03-24
Add error analysis tutorial

biermanncarl:
Description of changes:
- Adds a tutorial that teaches users the basics of error analysis.

This will solve the tutorial side of issue #3958

### DIFF
--- a/doc/tutorials/error_analysis/error_analysis_part1.ipynb
+++ b/doc/tutorials/error_analysis/error_analysis_part1.ipynb
@@ -1,0 +1,520 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Tutorial: Error Estimation - Part 1 (Introduction and Binning Analysis)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Table of contents\n",
+    "0. [Data generation](#Data-generation)\n",
+    "1. [Introduction](#Introduction)\n",
+    "2. [Uncorrelated samples](#Uncorrelated-samples)\n",
+    "3. [Binning analysis](#Binning-analysis)\n",
+    "4. [References](#References)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Data generation\n",
+    "\n",
+    "In this tutorial, you will learn how to estimate the accuracy of your simulation results. Because we are going to emply statistical methods, we need a fair amount of data to play with. The following code cell will generate two data sets which will be used throughout the tutorial."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "scrolled": true
+   },
+   "outputs": [],
+   "source": [
+    "import numpy as np\n",
+    "import matplotlib.pyplot as plt\n",
+    "plt.rcParams.update({'font.size': 18})\n",
+    "import sys\n",
+    "import logging\n",
+    "logging.basicConfig(level=logging.INFO, stream=sys.stdout)\n",
+    "\n",
+    "np.random.seed(43)\n",
+    "\n",
+    "def ar_1_process(n_samples, c, phi, eps):\n",
+    "    '''\n",
+    "    Generate a correlated random sequence with the AR(1) process.\n",
+    "\n",
+    "    Parameters\n",
+    "    ----------\n",
+    "    n_samples: :obj:`int`\n",
+    "        Sample size.\n",
+    "    c: :obj:`float`\n",
+    "        Constant term.\n",
+    "    phi: :obj:`float`\n",
+    "        Correlation magnitude.\n",
+    "    eps: :obj:`float`\n",
+    "        Shock magnitude.\n",
+    "    '''\n",
+    "    ys = np.zeros(n_samples)\n",
+    "    if abs(phi) >= 1:\n",
+    "        raise ValueError(\"abs(phi) must be smaller than 1.\")\n",
+    "    # draw initial value from normal distribution with known mean and variance\n",
+    "    ys[0] = np.random.normal(loc=c / (1 - phi), scale=np.sqrt(eps**2 / (1 - phi**2)))\n",
+    "    for i in range(1, n_samples):\n",
+    "        ys[i] = c + phi * ys[i - 1] + np.random.normal(loc=0., scale=eps)\n",
+    "    return ys\n",
+    "\n",
+    "# generate simulation data using the AR(1) process\n",
+    "\n",
+    "logging.info(\"Generating data sets for the tutorial ...\")\n",
+    "\n",
+    "N_SAMPLES = 100000\n",
+    "\n",
+    "C_1 = 2.0\n",
+    "PHI_1 = 0.85\n",
+    "EPS_1 = 2.0\n",
+    "time_series_1 = ar_1_process(N_SAMPLES, C_1, PHI_1, EPS_1)\n",
+    "\n",
+    "C_2 = 0.05\n",
+    "PHI_2 = 0.999\n",
+    "EPS_2 = 1.0\n",
+    "time_series_2 = ar_1_process(N_SAMPLES, C_2, PHI_2, EPS_2)\n",
+    "\n",
+    "logging.info(\"Done\")"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "scrolled": true
+   },
+   "outputs": [],
+   "source": [
+    "fig = plt.figure(figsize=(10, 6))\n",
+    "plt.title(\"The first 1000 samples of both time series\")\n",
+    "plt.plot(time_series_1[0:1000], label=\"time series 1\")\n",
+    "plt.plot(time_series_2[0:1000], label=\"time series 2\")\n",
+    "plt.xlabel(\"$i$\")\n",
+    "plt.ylabel(\"$X_i$\")\n",
+    "plt.legend()\n",
+    "plt.show()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Introduction\n",
+    "\n",
+    "In this tutorial, you will learn how to do statistical analysis of your simulation data.\n",
+    "This is an important topic, because the statistics of your data determine how precise your simulation result is. Furthermore, knowing about the statistics can help you optimize your disk space usage.\n",
+    "\n",
+    "ESPResSo provides a lot of ways to take measurements of your system. Usually, you will sample a quantity many times during a simulation and in the end average over all samples. Intuitively, the simulation result will be more precise the more samples are taken during the simulation. However, this is not the whole truth. There are some things that need to be considered, which we will cover in this tutorial.\n",
+    "\n",
+    "Formally, if you determine a physical quantity by averaging over several samples, you only *approximate* the unknown, true mean value. Usually, the quantity is expected to fluctuate around its mean; therefore, you can never directly measure the mean. You are bound to take repeated measurements and in the end average over all samples (a finite number). In your report, you will present this average as your result. Additionally, you should express the precision of your measurements to give a proper meaning to your result. And this is where things get more involved.\n",
+    "\n",
+    "There are several different ways to express the precision of your measurements. We will begin by briefly discussing what they are and what their differences are. After that, we will continue with the *standard error of the mean* as a viable option to be presented in your simulation results.\n",
+    "\n",
+    "### Standard deviation\n",
+    "The standard deviation is a measure for how much individual samples are expected to deviate from the mean. We want to use precise terminology, and therefore need to state that, in fact, we cannot directly measure the standard deviation but only estimate it. A commonly used estimator for the standard deviation is\n",
+    "\n",
+    "$\n",
+    "\\begin{align}\n",
+    "    \\hat{\\sigma} = \\sqrt{\\frac{1}{N-1.5}\\sum_{i=1}^{N}(X_i-\\overline{X})^2}\\tag{1}\n",
+    "\\end{align}\n",
+    "$\n",
+    "\n",
+    "where $\\hat{\\sigma}$ is the estimator of the standard deviation $\\sigma$, $N$ the number of samples, $X_i$ the individual samples and $\\overline{X}$ their mean. This estimator somewhat resembles the \"square root of the variance\". The curious $-1.5$ in the denominator is a necessary correction to make the estimator less biased (for further reading, see <a href='#[1]'>[1]</a>).\n",
+    "\n",
+    "### Standard error of the mean\n",
+    "The standard error of the mean (often abbreviated as SEM, or $s$, and its estimator is designated $\\hat{\\sigma}_\\overline{X}$) describes how much the mean value of your sample is expected to deviate from the true mean value $\\mu$. Imagine repeating the whole simulation over and over again, taking $N$ samples every time and averaging over them. The SEM quantifies how much those averages will fluctuate around the true mean $\\mu$. In fact, it is defined as the standard deviation of the averages.\n",
+    "\n",
+    "At first glance, it might seem to be very expensive to compute the SEM, because one would have to repeat the whole simulation many times. However, under the right circumstances, the SEM can be estimated from *a single series* of $N$ measurements. We will discuss how this can be done.\n",
+    "\n",
+    "### Confidence interval\n",
+    "A confidence interval (CI) specifies a range of numbers within which the unknown true mean value $\\mu$ lies with a certain probability $1-\\alpha$. A common confidence level is $1-\\alpha=95~\\%$. A $95~\\%$ CI would contain the true value $\\mu$ with probability $95~\\%$. Care must be taken interpreting the CI, since the lower and upper bound of a CI are themselves random variables. Just as a simulation run drafts samples from the overall ensemble, determining a CI from a simulation run is drafting a CI from all possible CIs. When the upper and lower bound of a CI have been calculated, this range either contains the true value or not, so there no longer is a probability attached to it. However, for repeated simulations with subsequent computation of the corresponding CIs, on average $95~\\%$ of CIs will contain the true value, while $5~\\%$ won't.\n",
+    "\n",
+    "If the samples are normally distributed and the SEM is known, the upper and lower bounds of the $95~\\%$ CI are $\\overline{X} \\pm 1.96 \\, \\hat{\\sigma}_\\overline{X}$.\n",
+    "\n",
+    "### Interquartile range\n",
+    "The interquartile range denotes the range, within which the central $50~\\%$ of all samples lie, if one were to order them by their size. This leaves one quarter of all samples lying below the interquartile range, and another quarter of all samples above it.\n",
+    "\n",
+    "### Now &ndash; what do we use?\n",
+    "We are interested in the precision of our overall, averaged, simulation result, and not in the precision of the individual samples. Those are expected to fluctuate, and in many cases, those fluctuations are uninteresting for the end result. Out of the options presented above, the SEM and the CI are the only ones doing this requirement justice. Since they are related, the question boils down to how to compute the SEM, which will be the topic of the rest of this tutorial."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Uncorrelated samples\n",
+    "How the SEM can be computed depends on the data itself. For uncorrelated samples, it is nearly trivial:\n",
+    "\n",
+    "$\n",
+    "\\begin{align}\n",
+    "    \\hat\\sigma_\\overline{X} = \\frac{\\hat\\sigma}{\\sqrt{N}}\\tag{2}\n",
+    "\\end{align}\n",
+    "$\n",
+    "\n",
+    "where $\\hat\\sigma_\\overline{X}$ is the estimated SEM, $\\hat\\sigma$ is the estimated standard deviation (see eq. 1) and $N$ is the number of samples. But what does it mean for samples to be uncorrelated?\n",
+    "\n",
+    "An example for uncorrelated samples would be the rolling of a dice. The outcome of each trial is completely independent to the previous trials. We might guess any number from 1 to 6, regardless of what has been the last result. The same could be true if we ran an experiment many times independently from one another and measured a quantity each time. By looking at one experimental value, we would'nt be able to predict the next one. The best guess would be simply the mean value of the entire series. In the case of rolling a dice, correlations could for example be observed if it was more probable to obtain the same result as in the previous dice roll rather than another result.\n",
+    "\n",
+    "Usually, when you run a molecular dynamics simulation, the particles will only move by a tiny amount during a time step. Consequently, most observables also change only by a small amount during a time step and it is, therefore, more probable to obtain a similar result rather than a completely different result. If we were to sample an observable in every time step, we would get a lot of samples with very similar values. It is said that the samples are *correlated*. Only if we wait for a sufficiently long time, the system will eventually have evolved to a completely different configuration, and we can expect the observable to assume a truly independent, *uncorrelated* value.\n",
+    "\n",
+    "It is often easy to see when samples are correlated. Execute the code cell below for an example, where a small part of `time_series_1` is plotted."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "fig = plt.figure(figsize=(10, 6))\n",
+    "plt.plot(time_series_1[1000:1050], \"x\")\n",
+    "fig.axes[0].margins(y=0.1)\n",
+    "plt.xlabel(\"$i$\")\n",
+    "plt.ylabel(\"$X_i$\")\n",
+    "plt.show()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "One can clearly see that each sample lies in the vicinity of the previous one.\n",
+    "\n",
+    "Below is an example for almost completely uncorrelated samples. The data points are taken from the same time series as in the previous example, but this time they are chosen with large gaps in between (every 800th sample is used). These samples appear to fluctuate a lot more randomly."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "fig = plt.figure(figsize=(10, 6))\n",
+    "plt.plot(np.arange(2000, 42000, 800), time_series_1[2000:42000:800], \"x\")\n",
+    "fig.axes[0].margins(y=0.1)\n",
+    "plt.xlabel(\"$i$\")\n",
+    "plt.ylabel(\"$X_i$\")\n",
+    "fig.axes[0].xaxis.set_major_locator(plt.MultipleLocator(base=8000))\n",
+    "plt.show()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "However, you should not trust your eye in deciding whether or not a time series is correlated. In fact, when running molecular dynamics simulations, your best guess is to always assume that samples are correlated, and that you should use one of the following techniques for statistical analysis, and rather not just use equation (2)."
+   ]
+  },
+  {
+   "cell_type": "markdown",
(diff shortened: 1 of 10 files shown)


doc/tutorials/error_analysis/error_analysis_part1.ipynb
```diff
@@ -0,0 +1,553 @@
...
+    "The standard deviation is a measure for how much individual samples are expected to deviate from the mean. We want to use precise terminology, and therefore need to state that, in fact, we cannot directly measure the standard deviation but only estimate it. A commonly used estimator for the standard error is\n",
+    "\n",
+    "$\n",
+    "\\begin{align}\n",
+    "    \\bar\\sigma = \\sqrt{\\frac{1}{N-1}\\sum_{i=1}^{N}(x_i-\\bar{x})^2}\\tag{1}\n",
+    "\\end{align}\n",
+    "$\n",
+    "\n",
+    "where $\\bar\\sigma$ is the estimator of the standard deviation $\\sigma$, $N$ the number of samples, $x_i$ the individual samples and $\\bar{x}$ their mean. This estimator somewhat resembles the \"square root of the variance\". However, instead of dividing by $N$ as for the variance, we divide by $N-1$. This accounts for the fact that we do not know the true mean value, but only estimate it (by averaging over all samples); and therefore we have one additional degree of freedom. Another way of viewing the $N-1$ divisor is to understand that for a single measurement, it is impossible to define the variance, which is represented by a \"division by zero\".\n",
```
jonaslandsgesell: Dividing by N-1 is due to the fact that the real average value is unknown and only estimated. In this case N-1 is the number of (statsical) degrees of freedom. Also the sample variance would involve division by N-1.

doc/tutorials/error_analysis/error_analysis_part1.ipynb
```diff
@@ -0,0 +1,553 @@
...
+    "\n",
+    "$\n",
+    "\\begin{align}\n",
+    "    \\bar\\epsilon = \\frac{\\bar\\sigma}{\\sqrt{N}}\\tag{2}\n",
+    "\\end{align}\n",
+    "$\n",
+    "\n",
+    "where $\\bar\\epsilon$ is the estimated SEM, $\\bar\\sigma$ is the estimated standard deviation (see eq. 1) and $N$ is the number of samples. But what does it mean for samples to be uncorrelated?\n",
+    "\n",
+    "An example for uncorrelated samples would be the rolling of a dice. The outcome of each trial is completely independent to the previous trials. We might guess any number from 1 to 6, regardless of what has been the last result. The same could be true if we ran an experiment many times independently from one another and measured a quantity each time. By looking at one experimental value, we would'nt be able to predict the next one. The best guess would be simply the mean value of the entire series.\n",
```
jonaslandsgesell: This is my favourite example!! You can also directly explain correlation here. If the outcome of rolling a dice would make it more probable to get this outcome the next time you roll a dice, then the dice would produce a correlated time series.

jonaslandsgesell:
I would propose to generate a correlated time-series via for example the AR-1 process. This is trivial. It will save a lot of time in the generation of the time-series and you can make the discussion independent of MD/MC sampling. I would also refer to the Markov-Chain central limit theorem. This is the straight forward way to introduce the error estimate like it can be found in the paper by Janke.

doc/tutorials/error_analysis/error_analysis_part1.ipynb
jngrad: `it is impossible to define the variance, which is represented by a \"division by zero\"`
I'm not certain this is a good illustration. Did you find it in a textbook?

IIRC, the `N - 1` denominator can be worked out from the definition of the variance, but it only holds for the *variance* estimator. For the *standard deviation* estimator, there is no exact denominator, as it depends on the probability distribution of the random process. You can have e.g. `N - 1.5` in the denominator, or more complicated expressions. For example the Gaussian distribution has an expression involving a ratio of Gamma functions (more details in [unbiased sample standard deviation](https://en.wikipedia.org/wiki/Standard_deviation#Unbiased_sample_standard_deviation)), which is built in the probability distribution of the [Student's *t*-distribution](https://en.wikipedia.org/wiki/Student%27s_t-distribution#Probability_density_function) for convenience (so that you can use an integer nu parameter, although there are applications in Bayesian statistics where a non-integer nu is used).

jonaslandsgesell: For further reading:
https://en.m.wikipedia.org/wiki/Unbiased_estimation_of_standard_deviation

jngrad: You are using \overline{\sigma} to refer to an estimator and \overline{x} to refer to an average. Textbooks usually one of these notations for estimators: \hat{\sigma}, \sigma_{\overline{X}} with *X* instead of *x* for the variable (*X* is the random process). A realization of an estimator has a different symbol, typically the latin version of the greek symbol, in this case *s*.

doc/tutorials/error_analysis/error_analysis_part1.ipynb
```diff
@@ -0,0 +1,553 @@
...
+    "\n",
+    "At first glance, it might seem to be very expensive to compute the SEM, because one would have to repeat the whole simulation many times. However, under the right circumstances, the SEM can be estimated from *a single series* of $N$ measurements. We will discuss how this can be done.\n",
+    "\n",
+    "### Confidence interval\n",
+    "A common confidence interval is the $95~\\%$ confidence interval. If we were given a number and a range around it, called the $95~\\%$ confidence interval, we would know that only $5~\\%$ of all samples are expected to lie outside that range. If the samples are normally distributed with a standard deviation $\\sigma$, the $95~\\%$ confidence interval would be roughly $\\pm1.96\\sigma$.\n",
```
jngrad: `95 % confidence interval, we would know that only 5 % of all samples are expected to lie outside that range`

I'm not exactly sure about this. In probability theory, \mu \pm 1.96\sigma excludes 5% of the probability distribution. However in statistics, the CI is an estimator of the lower and upper bounds of \mu for repeated experiments. I'm not convinced that the CI doubles as an estimator of the proportion of future samples to lie outside its range (assuming that by "expected", you mean "future").

doc/tutorials/error_analysis/error_analysis_part1.ipynb
jonaslandsgesell: I worked out the correlated dice process a bit: https://github.com/jonaslandsgesell/correlated_dice/blob/master/correlated%20dice.ipynb

doc/tutorials/error_analysis/error_analysis_part1.ipynb
biermanncarl: Thanks for the input on this! I think I'll go with Jonas' first suggestion.

biermanncarl: Thanks for the input!

doc/tutorials/error_analysis/error_analysis_part1.ipynb
biermanncarl: Thanks for your effort. I cannot open this file. May it be private?

jonaslandsgesell: I can open it in google chrome without being logged in. Let me know if the problem persists, then I can also send it to you via mail.

biermanncarl: Today, I was able to open it. Wow, that is a significant effort! From your first reply I imagined a short explanation of one or two extra sentences. I am not sure if *this* will fit into the existing tutorial. An intuitive picture of correlation is sufficient at this point in the tutorial. I, for one, am often struggling with texts that contain many formulas, and I imagine, some of the users would, too. However, it is a nice work, detailing many aspects of correlation, and I am thinking about a way we could include this.

doc/tutorials/error_analysis/error_analysis_part1.ipynb
biermanncarl: Good point, thanks!

doc/tutorials/error_analysis/error_analysis_part1.ipynb
jonaslandsgesell: I find the example of the correlated dice very educative and just wanted to make the point that correlations can be illustrated with this exampls :) So maybe you can build up this intuition behind correlations via stating: "In the case of rolling a dice, correlations could be observed if it was (for example) more probable to obtain the same result as in the previous dice roll rather than another result."

Probably, this example is not suited to showcase the uncertainty estimation of the mean value though. Probably an AR-1 model is more suited here.

biermanncarl:
Do these tutorials need to be tested? They don't rely on ESPResSo, at all, but only NumPy, Matplotlib and SciPy.

jngrad:
> Do these tutorials need to be tested? They don't rely on ESPResSo, at all, but only NumPy, Matplotlib and SciPy.

Yes, all tutorials need to be tested. It's the only way to be sure that the solution cells indeed work. You can add `np.random.seed(42)` at the beginning of the test to fix the seed and get reproducible random sequences, and check that the final standard error of the mean and autocorrelation time are correct.

doc/tutorials/error_analysis/error_analysis_part1.ipynb
```diff
@@ -0,0 +1,459 @@
...
+    "\n",
+    "$\n",
+    "\\begin{align}\n",
+    "    \\hat\\sigma_\\overline{X} = \\frac{\\hat\\sigma}{\\sqrt{N}}\\tag{2}\n",
+    "\\end{align}\n",
+    "$\n",
+    "\n",
+    "where $\\hat\\sigma_\\overline{X}$ is the estimated SEM, $\\hat\\sigma$ is the estimated standard deviation (see eq. 1) and $N$ is the number of samples. But what does it mean for samples to be uncorrelated?\n",
+    "\n",
+    "An example for uncorrelated samples would be the rolling of a dice. The outcome of each trial is completely independent to the previous trials. We might guess any number from 1 to 6, regardless of what has been the last result. The same could be true if we ran an experiment many times independently from one another and measured a quantity each time. By looking at one experimental value, we would'nt be able to predict the next one. The best guess would be simply the mean value of the entire series.\n",
```
jonaslandsgesell: Consider:
```suggestion
    "An example for uncorrelated samples would be the rolling of a dice. The outcome of each trial is completely independent to the previous trials. We might guess any number from 1 to 6, regardless of what has been the last result. The same could be true if we ran an experiment many times independently from one another and measured a quantity each time. By looking at one experimental value, we would'nt be able to predict the next one. The best guess would be simply the mean value of the entire series. In the case of rolling a dice, correlations could for example be observed if it was more probable to obtain the same result as in the previous dice roll rather than another result.\n",
```

doc/tutorials/error_analysis/error_analysis_part1.ipynb
```diff
@@ -0,0 +1,459 @@
...
+    "\n",
+    "However, usually, when you run a molecular dynamics simulation, the particles will only move by a tiny amount during a time step. Consequently, most observables also change only by a small amount during a time step. If we were to sample an observable in every time step, we would get a lot of samples with very similar values. It is said that the samples are *correlated*. Only if we wait for a sufficiently long time, the system will eventually have evolved to a completely different configuration, and we can expect the observable to assume a truly independent, *uncorrelated* value.\n",
```
jonaslandsgesell:
```suggestion
    "Usually, when you run a molecular dynamics simulation, the particles will only move by a tiny amount during a time step. Consequently, most observables also change only by a small amount during a time step and it is, therefore, more probable to obtain a similar result rather than a completely different result. If we were to sample an observable in every time step, we would get a lot of samples with very similar values. It is said that the samples are *correlated*. Only if we wait for a sufficiently long time, the system will eventually have evolved to a completely different configuration, and we can expect the observable to assume a truly independent, *uncorrelated* value.\n",
```

doc/tutorials/error_analysis/error_analysis_part1.ipynb
```diff
@@ -0,0 +1,459 @@
...
+    "\n",
+    "def ar_1_process(n_samples, y0, c, phi, eps, n_warmup):\n",
+    "    y = y0\n",
+    "    for i in range(n_warmup):\n",
+    "        y = c + phi*y + eps*np.random.uniform(-1.0,1.0)\n",
+    "    ys = np.zeros(n_samples)\n",
+    "    ys[0] = y\n",
+    "    for i in range(1,n_samples):\n",
+    "        ys[i] = c + phi*ys[i-1] + eps*np.random.uniform(-1.0,1.0)\n",
+    "    return ys\n",
```
jngrad: This looks a little bit verbose.

I did not immediately understand the purpose of the warmup phase, given that AR(1) processes have a fixed initial condition. It looks like generating an AR(1) sequence of size `n_samples+n_warmup` starting from 0, but the first `n_warmup` data points are "burned". Did I guess right? If that's the case, I'm not sure if it's better than having the user generate a random initial value with `np.random.uniform(-1.0,1.0)` directly, but I could be wrong.

This function is also quite slow since it uses Python loops instead of the C-extensions from numpy and scipy. The two `for` loops can be replaced by a single call to a scipy function: `scipy.signal.lfilter((1.,), (1., -phi), ys)` (implemented in c618ff9986b86e3f6fac3b7374c05a59d4584bde). Although it looks cryptic at first, the [scipy docs](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.lfilter.html) show the implementation:
```
a[0]*y[n] = b[0]*x[n] + b[1]*x[n-1] + ... + b[M]*x[n-M]
                      - a[1]*y[n-1] - ... - a[N]*y[n-N]
```
This is exactly how the ARMA(p, q) process is defined. We need p=1 and q=0. In the function call, `b = [1]` so only the self-term (constant + shock at time t) stays, and `a = [1, -phi]` so that the first term of the second line stays (phi times value at t - 1). For an AR(p) process, use `a = [1] + p * [-phi]`.

The benefit of using scipy over a custom implementation of the AR(1) process is that there is less room for bugs.

doc/tutorials/error_analysis/error_analysis_part1.ipynb
```diff
@@ -0,0 +1,459 @@
...
+    "\n",
+    "At first glance, it might seem to be very expensive to compute the SEM, because one would have to repeat the whole simulation many times. However, under the right circumstances, the SEM can be estimated from *a single series* of $N$ measurements. We will discuss how this can be done.\n",
+    "\n",
+    "### Confidence interval\n",
+    "A confidence interval (CI) specifies a range of numbers within which the unknown true mean value lies with a certain probability. A common confidence interval (CI) is the $95~\\%$ CI, which would contain the true value with probability $95~\\%$. Care must be taken interpreting the CI, since the lower and upper bound of a CI are themselves random variables. Just as a simulation run drafts samples from the overall ensemble, determining a CI from a simulation run is drafting a CI from all possible CIs. When the upper and lower bound of a CI have been calculated, this range either contains the true value or not, so there no longer is a probability attached to it. However, for repeated simulations with subsequent computation of the corresponding CIs, on average $95~\\%$ of CIs will contain the true value, while $5~\\%$ won't.\n",
```
jngrad: Here I would personally introduce the Greek symbol mu to represent the true mean and make the explanation clearer (implemented in b1b89faaf6dc6543c6a54cb4df07c73a63007204).

doc/tutorials/error_analysis/error_analysis_part2.ipynb
```diff
@@ -0,0 +1,534 @@
...
+    "plt.xlabel(\"lag time\")\n",
+    "plt.ylabel(\"auto-covariance\")\n",
+    "plt.show()\n",
+    "\n",
+    "print(\"Exponential autocorrelation time: {:.2f} sampling intervals\".format(popt[1]))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "behind-knife",
```
jngrad: Most of these easter eggs are funny, but some like this one are puzzling. They all appeared in bright red in `git diff` because my version of Jupyter doesn't have the plugin that introduced them, hence Jupyter deleted all of them. If we need these ids, we have to make sure everyone has the right plugins installed before editing the notebooks.

jngrad (commented):
First round of comments, more to come this week.

doc/tutorials/error_analysis/error_analysis_part1.ipynb
biermanncarl: You guessed right about the warmup phase. The reason why it is included is because the mean value of the series is unknown, or at least I haven't come across a simple formula how to compute it from the input parameters. So without the warmup phase, there would be a bias in the first part of the data series.
I agree on this method being slow, but it is no concern here. Executing the cell is almost instantaneously. Would it not be good to have something rather verbose in the tutorial, so the trainee can understand what's going on? I, for one, find the idea of an AR(1) process quite simple and intuitive, but would be deterred to find out more about the process if a more abstract "scipy.signal.lfilter" method would be called. And if there are bugs in the code, would the following error analysis not detect them and report them in the tests?
I agree that joining the two loops would be a good idea in order to not have duplicate code. Other than that, I think it is a matter of taste?

doc/tutorials/error_analysis/error_analysis_part2.ipynb
biermanncarl: I don't know what caused that, and I doubt it is necessary. If the notebook works as intended, including the hidden solutions, after the removal of these lines, they should be removed. Are you referring to `"id": "behind-knife",` ? Because I would guess that `"cell_type": "markdown",` is necessary to distinguish between code cells and markdown cells

doc/tutorials/error_analysis/error_analysis_part1.ipynb
jngrad: > The reason why it is included is because the mean value of the series is unknown, or at least I haven't come across a simple formula how to compute it from the input parameters. So without the warmup phase, there would be a bias in the first part of the data series.

Good point, the bias is actually visible in the time series with a warmup of 0.

> And if there are bugs in the code, would the following error analysis not detect them and report them in the tests?

Hopefully. We once had a flipped sign in a cross product in the C++ core ([diff](https://github.com/espressomd/espresso/pull/2382/commits/5945d070b7037d98d88fb2fe3efa3055c77ef1b7#diff-d45782ab798b1f002476e4ac52098301d1bf8d74ab0de0b17c04229fef1570a6L335-L343)). The test was always green. It took 4 years for someone to notice.

I've reverted my changes to your original implementation (00dbceb5ad5c1a055813fe9f1c5da497cbcc27c3) and moved the scipy function to the test case (4cf00de72f16503c231ab7d12a4e79e72586422f).

doc/tutorials/error_analysis/error_analysis_part2.ipynb
jngrad: Yes, I just meant the ids. The are removed in af5396d446beaca8c5bbb96891aa4678d166ecc8.

doc/tutorials/error_analysis/error_analysis_part1.ipynb
jonaslandsgesell: Isn't the mean value of the AR-1 process mu=c/(1-phi) for |phi|<1 ?https://en.m.wikipedia.org/wiki/Autoregressive_model

biermanncarl: > Isn't the mean value of the AR-1 process mu=c/(1-phi) for |phi|<1 ?https://en.m.wikipedia.org/wiki/Autoregressive_model

Good hint. Using the expressions there, it is possible to omit the warmup phase and the initial value y_0. 

> moved the scipy function to the test case ([4cf00de](https://github.com/espressomd/espresso/commit/4cf00de72f16503c231ab7d12a4e79e72586422f)).

Thanks. Hmmm... well it is annoying that the two notebooks are supposed to have some identical code, so they require identical tests.
One obvious solution would be to join the two tutorials together. Or just live with the duplicate code.

doc/tutorials/error_analysis/error_analysis_part2.ipynb
biermanncarl: These fields are added in newer versions (4.5) of the notebooks. See [here](https://github.com/jupyterlab/jupyterlab/issues/9729). Since part 2 has been begun later than part 1, it is in notebook version 4.5, while the former is in 4.4 and doesn't contain any IDs. I'll revert the notebook to 4.4, so this won't occur. But likely we will have to deal with this in the future.

doc/tutorials/error_analysis/error_analysis_part1.ipynb
```diff
@@ -0,0 +1,464 @@
...
+    "```python\n",
+    "def do_binning_analysis(data, bin_size):\n",
+    "    n_samples = len(data)\n",
+    "    n_bins = n_samples // bin_size\n",
+    "    bin_avgs = np.zeros(n_bins)\n",
+    "    for i in range(n_bins):\n",
+    "        bin_avgs[i] = np.average(data[i*bin_size:(i+1)*bin_size])\n",
+    "    avg = np.average(bin_avgs)\n",
+    "    sem = np.sqrt(np.sum((bin_avgs-avg)**2)/((n_bins-1.5)*n_bins))\n",
+    "    return sem\n",
```
jngrad: This function needs to be tested: 2346874116a0324ab914be5d26ed4c8a317487a4

jngrad: Optionally, the python for loop could be replaced with numpy to decrease the test runtime from 12 seconds to 2.3 seconds (63cdea90bccf0b2b11e4d5236defbb9973870b8c).

doc/tutorials/error_analysis/error_analysis_part1.ipynb
```diff
@@ -0,0 +1,464 @@
...
+   "source": [
+    "```python\n",
+    "sizes = np.arange(3,5001,dtype=int)\n",
+    "sems = np.zeros(5001-3,dtype=float)\n",
+    "for s in range(len(sizes)):\n",
+    "    sems[s] = do_binning_analysis(time_series_1,sizes[s])\n",
+    "\n",
+    "plt.plot(sizes,sems, \"x\")\n",
+    "plt.xscale(\"log\")\n",
+    "plt.show()\n",
```
jngrad: By default, plots in Jupyter notebooks are quite small. We like to use a combination of `plt.rcParams.update({'font.size': 18})` and `fig = plt.figure(figsize=(10, 6))` to make them more readable. I've updated the [tutorial instructions](https://github.com/espressomd/espresso/wiki/Documentation#Tutorials) accordingly, it was an oversight on my part; it should have been documented since the beginning.

I've changed all plots accordingly and labeled all the axes: 67840aa3d9c19a776ff0363c4de96fbd9ceaa229

doc/tutorials/error_analysis/error_analysis_part1.ipynb
```diff
@@ -0,0 +1,464 @@
...
+    "fit_params, _ = curve_fit(fit_fn,sizes,sems[:CUTOFF], (0.05,1,0.5))\n",
+    "\n",
+    "fit_sems = fit_fn(sizes,*fit_params)\n",
+    "\n",
+    "plt.plot(sizes,sems[:CUTOFF], \"x\")\n",
+    "plt.plot(sizes,fit_sems,\"-\")\n",
+    "plt.xscale(\"log\")\n",
+    "plt.show()\n",
+    "\n",
+    "print(\"Final Standard Error of the Mean: {:.3f}\".format(fit_params[2]))"
```
jngrad: f-strings look more modern: e3258a4ee6ba01b697fdfe9a25c8acf216e424a3

jngrad: You probably also want to look into the `logging` module instead of `print`. We use it in the LB tutorial part 2. Example:
```python
import logging
logging.basicConfig(level=logging.INFO, stream=sys.stdout)
# ...
logging.info("Equilibrating the system.")
```
Output:
```
INFO:root:Equilibrating the system.
```

doc/tutorials/error_analysis/error_analysis_part2.ipynb
jngrad: Nice find! I'll try to keep that in mind the next time we generate the docker images with newer versions of Jupyter.

doc/tutorials/error_analysis/error_analysis_part1.ipynb
jngrad: We could move the function to a python file and import it in the notebooks, but then users would have to manually open the python file if they want to see the function implementation...

jonaslandsgesell:
Plotting the analytical autocorrelation function (see B_n in https://en.m.wikipedia.org/wiki/Autoregressive_model ) in the plot where you show the estimated auto correlation function could be interesting to the reader?

doc/tutorials/error_analysis/error_analysis_part2.ipynb
```diff
@@ -0,0 +1,550 @@
...
+    "\n",
+    "# plot\n",
+    "fig = plt.figure(figsize=(10, 6))\n",
+    "plt.plot(j_max_v,C*tau_int_v)\n",
+    "plt.plot(j_max_v, j_max_v)\n",
+    "plt.plot([j_max],[C*tau_int_v[j_max]],\"ro\")\n",
+    "plt.xscale(\"log\")\n",
+    "plt.ylim((0,50))\n",
+    "plt.xlabel(r\"sum length $j_\\mathrm{max}$\")\n",
+    "plt.ylabel(r\"$\\tau_{X, \\mathrm{int}}$\")\n",
```
jngrad: This y-axis should actually read C times tau. Also, several axis labels are missing the hat symbol on estimators, like tau here and Rxx elsewhere. That was fixed as part of dd07036631122.

doc/tutorials/error_analysis/error_analysis_part2.ipynb
```diff
@@ -0,0 +1,550 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "solution2": "hidden"
+   },
+   "source": [
+    "```python\n",
+    "# naive Python solution\n",
+    "autocov = np.zeros(1000)\n",
```
jngrad: This simple implementation takes a lot of time. We could limit it to 300 iterations (0db6bc6db7fa), since the result isn't used in the rest of the tutorial.

doc/tutorials/error_analysis/error_analysis_part2.ipynb
```diff
@@ -0,0 +1,550 @@
...
+    "j = np.linspace(0,N_MAX-1,N_MAX)\n",
+    "popt, pcov = curve_fit(exp_fnc, j, autocov[:N_MAX], p0=[15,10])\n",
+    "\n",
+    "# compute analytical ACF of AR(1) process\n",
+    "AN_SIGMA_1 = np.sqrt(EPS_1 ** 2 / (1 - PHI_1 ** 2))\n",
+    "AN_TAU_EXP_1 = -1 / np.log(PHI_1)\n",
+    "an_acf_1 = AN_SIGMA_1**2*np.exp(-j/AN_TAU_EXP_1)\n",
+    "\n",
+    "fig = plt.figure(figsize=(10, 6))\n",
+    "plt.plot(j,autocov[:N_MAX], 'x', label=\"numerical ACF\")\n",
```
jngrad: Here and elsewhere, it looks like we use log scales for value ranges that include 0, which make the curves look strange. I've tried to address some of them in e54637ed47c, but you should probably have a closer look.

doc/tutorials/error_analysis/error_analysis_part2.ipynb
```diff
@@ -0,0 +1,550 @@
...
+    "import numpy as np\n",
+    "import matplotlib.pyplot as plt\n",
+    "plt.rcParams.update({'font.size': 18})\n",
+    "import sys\n",
+    "import logging\n",
+    "logging.basicConfig(level=logging.INFO, stream=sys.stdout)\n",
+    "\n",
+    "np.random.seed(43)\n",
+    "\n",
+    "def ar_1_process(n_samples, c, phi, eps):\n",
```
jngrad, line 46: docstrings: 6798a90125

jngrad (commented):
The proposed changes are in [jngrad/espresso@error-analysis-tutorial](https://github.com/jngrad/espresso/tree/error-analysis-tutorial).

jonaslandsgesell:
Did you consider teaching bootstrap and showcasing https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.bootstrap.html#scipy.stats.bootstrap ?

Scipy does not seem to use block bootstrap which is needed for correlated time series data.
Potentially, this python package could be showcased:
https://pypi.org/project/recombinator/

jngrad:
Teaching bootstrapping would make this tutorial quite long. We also want to limit the number of python dependencies in ESPResSo.

jngrad (approved):

PR #4178, opened 2021-03-25
Implement slicing for LB

### LINKED ISSUE #4143: Implement slicing support for LB
i.e., `lbf[4:20,:,1].velocity`
or
`lbf[:,:,:].velocity = np.zeros(lbf.shape))` 

* the __getitem__() method of The LB Python class should accept slices
* If slices are passed in, rather than an instance of LB Node, an instance of a new LBSlice class shuold be returned

An initial implementation of the LBSlice class should just act as proxy to single node getters. this can be done with 20-30 lines of Python. This will give the usability benefit immediately. A more efficient implementation can be done after the Walberla transition, if the performance increase is ever needed.

Steps:

* In src/python/espressomd.lb.pyx, create a new class LBSlice
    
    * constructor should take three slice objects for x,y,z
    * indivudal properties:
    * Using those, calcualte the x,y,z dimensions of the resulting array (i.e., how many lattice sites are requested in each direction) should be (max-min)/step, or 1 if the value for the coordinate is a single int.

        * Find out the number and type of values per lb site (either hard-code 1 float for density, 3 floats for velocity, 3 ints for index, ...) or fetch the first value (e.g., lower left corner) using a single node getter
        * Shape the result array accordingly
        * Use a nested loop to fetch all the individual values and fill the result array

* Modify the __getimem__() method of the lb fluid class in src/python/espresso/lb.pyx:
    
    If any of the arguments is an instance of slice, create an instance of LBSlice with the three arguments for x,y,z

Draft:
```
class LBSlice:

    def __init__(self, x,y,z, lb_fluid):
        self.x = x
        ...
        self.lb_fluid = lb_fluid
        self.dim = self.calculate_dimensions(x,y,z)

    def calculate_dimensions(x,y,z):
        # x,y,z either an instance of slcie or an int
        # Return tubple of 3 ints for the dimensions calculated from x,y,z
        return (...)

    
    def get_values(property_name, values_per_site, result_type):
        dim = self.dim
        res =  np.zeors((dim[0],dim[1],dim[22], values_per_site), dtype=result_type)
        for x in ...
           for y in ...
              for z in ...
                  res[x,y,z,:] = getattr(self.lb_fluid[x,y,z], property_name)
    def set_values(property_name, values): 
        for x in ...
           for y in ...
              for z in ...
                  setattr(self.lb_fluid[x,y,z], property_name, values[x,y,z])

    
    property velocity:
        def __get__(self):
           return self.get_vavlues("velocity", 3, float)
       def __set__(values):
           self.set_values("velocity", values)
# analogous for other properties like density
```

KaiserMartin:
Fixes #4143

Description of changes:
- Modified __getitem__ to detect slices in input keys
- Introduced LBSlice class that calculates node indices from slice input and calls single node getter from LBFLuidRoutines class for every node in the slice.

The output is a multidimensional array containing the requested quantity and is of dimension len(slice_x)*len(slice_y)*len(slice_z)*len(quantity), where len(quantity) is the dimensions of the quantity, for example 3 in case of velocity. The output array is arranged in the way numpy cycles through slices. Therefore, when setting attributes one has to set the dimensions of the input array exactly the same way numpy cycles through the sliced nodes.
An error is issued if the dimensions of the slices and the input does not match.

### DIFF
--- a/src/utils/include/utils/Array.hpp
+++ b/src/utils/include/utils/Array.hpp
@@ -26,8 +26,9 @@
 
 #include <cassert>
 #include <cstddef>
+#include <iterator>
+#include <ostream>
 #include <stdexcept>
-
 namespace Utils {
 
 namespace detail {
@@ -151,6 +152,16 @@ private:
   template <typename Archive>
   void serialize(Archive &ar, const unsigned int /* version */) {
     ar &m_storage;
+  }
+
+  friend std::ostream &operator<<(std::ostream &out, Array const &a) {
+    if (not a.empty())
+      out << a[0];
+
+    for (auto it = std::next(a.begin()); it != a.end(); ++it)
+      out << ", " << *it;
+
+    return out;
   }
 };
 
--- a/src/utils/tests/Array_test.cpp
+++ b/src/utils/tests/Array_test.cpp
@@ -117,3 +117,23 @@ BOOST_AUTO_TEST_CASE(tuple_protocol) {
 
   BOOST_CHECK_EQUAL(Utils::get<1>(A{1, 2, 3, 4}), 2);
 }
+
+BOOST_AUTO_TEST_CASE(streaming_operator) {
+  {
+    auto const a = Utils::Array<int, 1>{1};
+
+    std::stringstream ss;
+    ss << a;
+
+    BOOST_CHECK_EQUAL(ss.str(), "1");
+  }
+
+  {
+    auto const a = Utils::Array<int, 3>{1, 2, 3};
+
+    std::stringstream ss;
+    ss << a;
+
+    BOOST_CHECK_EQUAL(ss.str(), "1, 2, 3");
+  }
+}


src/python/espressomd/lb.pyx
```diff
@@ -542,3 +548,144 @@ cdef class LBFluidRoutines:
 
         def __set__(self, value):
             raise NotImplementedError
+
+cdef class LBSlice:
+    cdef Vector3i node 
+    cdef np.ndarray x_indices
+    cdef np.ndarray y_indices
+    cdef np.ndarray z_indices
```
KaiSzuttor: are you sure those variables should be static? i think you can just remove those statements.

src/python/espressomd/lb.pyx
```diff
@@ -542,3 +548,144 @@ cdef class LBFluidRoutines:
 
         def __set__(self, value):
             raise NotImplementedError
+
+cdef class LBSlice:
```
KaiSzuttor: i think this does not need to be a cython class

KaiserMartin: is there a drawback if it is? its more convenient to handle types of cython and python arrays if both classes are cython classes.

src/python/espressomd/lb.pyx
KaiserMartin: if i keep it as a cython class they should be static.
expect the "cdef Vector3i node ", i deleted that since it is not effectively used anywhere

src/python/espressomd/lb.pyx
KaiSzuttor: In general we try to avoid cython classes if possible. Cython is only used to get bindings to the c++ core interface. The  interface to this C++ interface is implemented in the ```LBFluid``` class. So to have a better separation of concerns you should not use a cython class here and it should also not be necessary.

jngrad: @KaiSzuttor Don't we need to have an intermediate class, though? Since we want to be able to access properties, we cannot simply return a numpy array of the `LBFluidRoutines` objects.

KaiserMartin: sorry if I'm uneducated about that (I'm not that familiar with mixing python and cython), but is it even possible to simply call a cython class from a python one, like i do it here? as far as i know there is no proper workaround for that

KaiserMartin: > sorry if I'm uneducated about that (I'm not that familiar with mixing python and cython), but is it even possible to simply call a cython class from a python one, like i do it here? as far as i know there is no proper workaround for that

in line wit @jngrad comment

KaiSzuttor: we are only accessing LB fluid properties that are already interfaced or do we want to change the way we retrieve information from the c++ interface?

src/python/espressomd/lb.pyx
```diff
@@ -542,3 +548,144 @@ cdef class LBFluidRoutines:
...
+
+        def __set__(self, value):
+            raise NotImplementedError
+
+    property population:
+        def __get__(self):
+            prop_name = "population"
+            shape_res = (19,)
+            cdef Vector19d double_return
+            double_return = lb_lbnode_get_pop(self.node)
```
KaiSzuttor: this is not needed since the populations are interfaced in the ```LBFluid```

src/python/espressomd/lb.pyx
```diff
@@ -542,3 +548,144 @@ cdef class LBFluidRoutines:
...
+    property index:
+        def __get__(self):
+            prop_name = "index"
+            shape_res = (3,)
+            return self.get_values(
+                self.x_indices, self.y_indices, self.z_indices, prop_name, shape_res)
+
+    property velocity:
+        def __get__(self):
+            prop_name = "velocity"
```
KaiSzuttor, line 617: those strings are also not necessary

src/python/espressomd/lb.pyx
KaiserMartin: true, will be fixed

src/python/espressomd/lb.pyx
```diff
@@ -542,3 +548,144 @@ cdef class LBFluidRoutines:
...
+        res = np.zeros(
+            (_x_indices.size,
+             _y_indices.size,
+             _z_indices.size,
+             *shape_res))
+        for i, x in enumerate(_x_indices):
+            for j, y in enumerate(_y_indices):
+                for k, z in enumerate(_z_indices):
+                    res[i, j, k] = getattr(LBFluidRoutines(
+                        np.array([x, y, z])), prop_name)
```
KaiSzuttor, line 582: this is repeated in every property and can probably be reused if put into a function

KaiserMartin (commented):
- Removing unnecessary c variables from population and squeezing output from property boundary so it has the right dimensions

src/python/espressomd/lb.pyx
```diff
@@ -542,3 +548,137 @@ cdef class LBFluidRoutines:
...
+
+cdef class LBSlice:
+    cdef np.ndarray x_indices
+    cdef np.ndarray y_indices
+    cdef np.ndarray z_indices
+
+    def __init__(self, key):
+        shape = lb_lbfluid_get_shape()
+        self.x_indices, self.y_indices, self.z_indices = self.get_indices(
+            key, shape[0], shape[1], shape[2])
```
KaiSzuttor:
```suggestion
    def __init__(self, key, shape):
        self.shape = shape
        self.x_indices, self.y_indices, self.z_indices = self.get_indices(
            key, shape[0], shape[1], shape[2])
```

src/python/espressomd/lb.pyx
```diff
@@ -542,3 +548,137 @@ cdef class LBFluidRoutines:
 
         def __set__(self, value):
             raise NotImplementedError
+
+cdef class LBSlice:
```
KaiSzuttor:
```suggestion
class LBSlice:
```

src/python/espressomd/lb.pyx
```diff
@@ -542,3 +548,137 @@ cdef class LBFluidRoutines:
...
+    cdef np.ndarray x_indices
+    cdef np.ndarray y_indices
+    cdef np.ndarray z_indices
```
KaiSzuttor:
```suggestion
```

src/python/espressomd/lb.pyx
```diff
@@ -542,3 +548,137 @@ cdef class LBFluidRoutines:
...
+    def set_values(self, _x_indices, _y_indices, _z_indices, prop_name, value):
+        for i, x in enumerate(_x_indices):
+            for j, y in enumerate(_y_indices):
+                for k, z in enumerate(_z_indices):
+                    setattr(LBFluidRoutines(
+                        np.array([x, y, z])), prop_name, value[i, j, k])
+
+    property density:
+        def __get__(self):
+            prop_name = "density"
```
KaiSzuttor:
```suggestion
```

src/python/espressomd/lb.pyx
```diff
@@ -542,3 +548,137 @@ cdef class LBFluidRoutines:
...
+            shape_res = (1,)
+            return np.squeeze(self.get_values(
+                self.x_indices, self.y_indices, self.z_indices, prop_name, shape_res), axis=3)
+
+        def __set__(self, value):
+            prop_name = "density"
```
KaiSzuttor:
```suggestion
```

src/python/espressomd/lb.pyx
```diff
@@ -542,3 +548,137 @@ cdef class LBFluidRoutines:
...
+            self.set_values(
+                self.x_indices,
+                self.y_indices,
+                self.z_indices,
+                prop_name,
+                value)
+
+    property index:
+        def __get__(self):
+            prop_name = "index"
```
KaiSzuttor:
```suggestion
```

src/python/espressomd/lb.pyx
```diff
@@ -542,3 +548,137 @@ cdef class LBFluidRoutines:
...
+            shape_res = (3,)
+            return self.get_values(
+                self.x_indices, self.y_indices, self.z_indices, prop_name, shape_res)
+
+    property velocity:
+        def __get__(self):
+            prop_name = "velocity"
```
KaiSzuttor:
```suggestion
```

src/python/espressomd/lb.pyx
```diff
@@ -542,3 +548,137 @@ cdef class LBFluidRoutines:
...
+            shape_res = (3,)
+            return self.get_values(
+                self.x_indices, self.y_indices, self.z_indices, prop_name, shape_res)
+
+        def __set__(self, value):
+            prop_name = "velocity"
```
KaiSzuttor:
```suggestion
```

src/python/espressomd/lb.pyx
```diff
@@ -542,3 +548,137 @@ cdef class LBFluidRoutines:
...
+                    prop_name,
+                    value)
+            else:
+                raise ValueError(
+                    "Input-dimensions of velocity array", value.shape, "does not match slice dimensions",
+                    (len(self.x_indices), len(self.y_indices), len(self.z_indices), 3), ".")
+
+    property pressure_tensor:
+        def __get__(self):
+            prop_name = "pressure_tensor"
```
KaiSzuttor:
```suggestion
```

src/python/espressomd/lb.pyx
```diff
@@ -542,3 +548,137 @@ cdef class LBFluidRoutines:
...
+            shape_res = (3, 3)
+            return self.get_values(
+                self.x_indices, self.y_indices, self.z_indices, prop_name, shape_res)
+
+        def __set__(self, value):
+            raise NotImplementedError
+
+    property pressure_tensor_neq:
+        def __get__(self):
+            prop_name = "pressure_tensor_neq"
```
KaiSzuttor:
```suggestion
```

src/python/espressomd/lb.pyx
```diff
@@ -542,3 +548,137 @@ cdef class LBFluidRoutines:
...
+            shape_res = (3, 3)
+            return self.get_values(
+                self.x_indices, self.y_indices, self.z_indices, prop_name, shape_res)
+
+        def __set__(self, value):
+            raise NotImplementedError
+
+    property population:
+        def __get__(self):
+            prop_name = "population"
```
KaiSzuttor:
```suggestion
```

src/python/espressomd/lb.pyx
```diff
@@ -542,3 +548,137 @@ cdef class LBFluidRoutines:
...
+            shape_res = (19,)
+            return self.get_values(
+                self.x_indices, self.y_indices, self.z_indices, prop_name, shape_res)
+
+        def __set__(self, value):
+            prop_name = "population"
```
KaiSzuttor:
```suggestion
```

src/python/espressomd/lb.pyx
```diff
@@ -542,3 +548,137 @@ cdef class LBFluidRoutines:
...
+                    value)
+            else:
+                raise ValueError(
+                    "Input-dimensions of population array", value.shape, "does not match slice dimensions",
+                    (len(self.x_indices), len(self.y_indices), len(self.z_indices), 19), ".")
+
+    property boundary:
+        def __get__(self):
+            shape_res = (1,)
+            prop_name = "boundary"
```
KaiSzuttor:
```suggestion
```

src/python/espressomd/lb.pyx
```diff
@@ -542,3 +548,137 @@ cdef class LBFluidRoutines:
...
+
+        def __set__(self, value):
+            prop_name = "density"
+            self.set_values(
+                self.x_indices,
+                self.y_indices,
+                self.z_indices,
+                prop_name,
+                value)
+
```
KaiSzuttor:
```suggestion
    @property
    def density(self):
        prop_name = "density"
        shape_res = (1,)
        return np.squeeze(self.get_values(
               self.x_indices, self.y_indices, self.z_indices, prop_name, shape_res), axis=3)
    @density.setter
    def density(self, value):
            self.set_values(
                self.x_indices,
                self.y_indices,
                self.z_indices,
                prop_name,
                value)

```

KaiSzuttor: this is the modern python syntax for properties

src/python/espressomd/lb.pyx
jngrad: > i think this does not need to be a cython class

I'm currently working on a Python version of the Cython class, I'll push the commits once I'm done updating the test.

jngrad:
@KaiserMartin  Here is the result of applying Kai's review: [jngrad/espresso@fix-4143](https://github.com/jngrad/espresso/tree/fix-4143) ([diff view](https://github.com/KaiserMartin/espresso/compare/python...jngrad:fix-4143))

src/python/espressomd/lb.pyx
jngrad: This is now a regular Python class.

jngrad (commented):
PR ready for merge from my side.

RudolfWeeber (dismissed):

jngrad:
please don't merge it yet, we're looking into a simplification using a python reflection

jngrad:
The `LBSlice` now returns numpy arrays as `array_locked`. Properties are generated automatically.
The broken `LBSlice.boundary` raise assertion check is now fixed.

jngrad:
I accidentally pushed the `python` branch instead of `fix-4143` to KaiserMartin's fork, and now GitHub rejects further pushes:
```
$ git push -f KaiserMartin fix-4143:python
X11 forwarding request failed on channel 0
Enumerating objects: 99, done.
Counting objects: 100% (99/99), done.
Delta compression using up to 32 threads
Compressing objects: 100% (88/88), done.
Writing objects: 100% (88/88), 17.57 KiB | 1.46 MiB/s, done.
Total 88 (delta 71), reused 0 (delta 0)
remote: Resolving deltas: 100% (71/71), completed with 11 local objects.
To github.com:KaiserMartin/espresso.git
 ! [remote rejected]     fix-4143 -> python (permission denied)
error: failed to push some refs to 'git@github.com:KaiserMartin/espresso.git'
```
GitHub also closed the PR using my username. Moving over to #4195. Sorry for the inconvenience.

PR #4179, opened 2021-03-25
Hollowconical Frustum with cutout

### LINKED ISSUE #4172: Add central angle option to HollowConicalFrustum shape

christophlohrmann:
Fixes #4172 

Description of changes:
- add ``central_angle`` option to HollowConicalFrustum
- change ``axis`` and ``center`` to ``CylindricalTransformationParameters`` for HCF
- more default behaviour for ``CylindricalTransformationParameters``


KaiSzuttor:
please check the shape unit test

src/shapes/src/HollowConicalFrustum.cpp
```diff
@@ -20,55 +20,87 @@
...
-  auto const pos_intersection =
-      Utils::transform_coordinate_cylinder_to_cartesian(
-          {r_intersection, pos_cyl[1], z_intersection}, m_axis, m_orientation) +
-      m_center;
 
-  auto const u = (pos - pos_intersection).normalize();
-  auto const d = (pos - pos_intersection).norm() - 0.5 * m_thickness;
-  auto const in_out = (d < 0.0) ? -m_direction : m_direction;
-  dist = in_out * std::abs(d);
+  Utils::Vector3d pos_closest;
```
KaiSzuttor: not sure if this is easy to follow... can't you also recycle the old code for the case where the angle is within the segment without the gap?

christophlohrmann: I tried to, but it does not work. The point of interest, ``axis`` and the edge of the mantle are not in a common plane

KaiSzuttor: what i meant: shouldn't the distance calculation just be the prior one if the point of interest is in the region without the gap? Think of the the case where the central angle is 0 or 2pi (no gap)...

christophlohrmann: We can revert to the old calculation. I thought it is more clear if for both cases the calculation is similar

KaiSzuttor: not if you remove all my comments and don't add new ones for your calculation ;). Also you could have saved your time implementing this again... So please add some more comments and explain what you are calculating. Without additional informations it's really hard to follow what you are doing here.

KaiSzuttor (approved):
thanks a lot @christophlohrmann, I think this version has made quite an improvement to the readability

PR #4180, opened 2021-03-25
Make particle id less prominent in the testsuite

jngrad:
Partial fix for #4029

Description of changes:
- use `ParticleHandle` objects instead of `system.part[pid]` in the python tests
- improve tests clarity (batch particle creation, utility functions from `itertools` and `tests_common.py`, etc.)

### DIFF
--- a/testsuite/python/dawaanr-and-bh-gpu.py
+++ b/testsuite/python/dawaanr-and-bh-gpu.py
@@ -92,9 +92,6 @@ class BHGPUTest(ut.TestCase):
             self.system.actors.add(bh_gpu)
             self.system.integrator.run(steps=0, recalc_forces=True)
 
-            bhgpu_f = []
-            bhgpu_t = []
-
             bhgpu_f = np.copy(self.system.part[:].f)
             bhgpu_t = np.copy(self.system.part[:].torque_lab)
             bhgpu_e = self.system.analysis.energy()["total"]
--- a/testsuite/python/dawaanr-and-bh-gpu.py
+++ b/testsuite/python/dawaanr-and-bh-gpu.py
@@ -79,14 +79,9 @@ class BHGPUTest(ut.TestCase):
             self.system.actors.add(dds_cpu)
             self.system.integrator.run(steps=0, recalc_forces=True)
 
-            dawaanr_f = []
-            dawaanr_t = []
-
-            for i in range(n):
-                dawaanr_f.append(self.system.part[i].f)
-                dawaanr_t.append(self.system.part[i].torque_lab)
-            dawaanr_e = espressomd.analyze.Analysis(
-                self.system).energy()["total"]
+            dawaanr_f = np.copy(self.system.part[:].f)
+            dawaanr_t = np.copy(self.system.part[:].torque_lab)
+            dawaanr_e = self.system.analysis.energy()["total"]
 
             del dds_cpu
             self.system.actors.clear()
@@ -100,11 +95,9 @@ class BHGPUTest(ut.TestCase):
             bhgpu_f = []
             bhgpu_t = []
 
-            for i in range(n):
-                bhgpu_f.append(self.system.part[i].f)
-                bhgpu_t.append(self.system.part[i].torque_lab)
-            bhgpu_e = espressomd.analyze.Analysis(
-                self.system).energy()["total"]
+            bhgpu_f = np.copy(self.system.part[:].f)
+            bhgpu_t = np.copy(self.system.part[:].torque_lab)
+            bhgpu_e = self.system.analysis.energy()["total"]
 
             # compare
             for i in range(n):
--- a/testsuite/python/interactions_dihedral.py
+++ b/testsuite/python/interactions_dihedral.py
@@ -92,10 +92,7 @@ class InteractionsBondedTest(ut.TestCase):
         self.system.cell_system.skin = 0.4
         self.system.time_step = .1
 
-        self.system.part.add(pos=self.start_pos, type=0)
-        self.system.part.add(pos=self.start_pos, type=0)
-        self.system.part.add(pos=self.start_pos, type=0)
-        self.system.part.add(pos=self.start_pos, type=0)
+        self.system.part.add(pos=4 * [self.start_pos], type=4 * [0])
 
     def tearDown(self):
         self.system.part.clear()
--- a/testsuite/python/interactions_dihedral.py
+++ b/testsuite/python/interactions_dihedral.py
@@ -92,10 +92,10 @@ class InteractionsBondedTest(ut.TestCase):
         self.system.cell_system.skin = 0.4
         self.system.time_step = .1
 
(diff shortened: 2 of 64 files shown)


testsuite/python/dawaanr-and-bh-gpu.py
```diff
@@ -100,11 +95,9 @@ def test(self):
             bhgpu_f = []
             bhgpu_t = []
```
reinaual:
```suggestion
```

testsuite/python/interactions_dihedral.py
```diff
@@ -92,10 +92,10 @@ def setUp(self):
...
         self.system.time_step = .1
 
-        self.system.part.add(id=0, pos=self.start_pos, type=0)
-        self.system.part.add(id=1, pos=self.start_pos, type=0)
-        self.system.part.add(id=2, pos=self.start_pos, type=0)
-        self.system.part.add(id=3, pos=self.start_pos, type=0)
+        self.system.part.add(pos=self.start_pos, type=0)
+        self.system.part.add(pos=self.start_pos, type=0)
+        self.system.part.add(pos=self.start_pos, type=0)
+        self.system.part.add(pos=self.start_pos, type=0)
```
reinaual:
```suggestion
        self.system.part.add(pos=4 * [self.start_pos], type=4 * [0])
```

testsuite/python/mmm1d.py
```diff
@@ -67,11 +67,11 @@ def test_energy(self):
 
     def test_with_analytical_result(self, prefactor=1.0, accuracy=1e-4):
         self.system.part.clear()
-        self.system.part.add(pos=[0, 0, 0], q=1)
-        self.system.part.add(pos=[0, 0, 1], q=1)
+        p = self.system.part.add(pos=[0, 0, 0], q=1)
+        _ = self.system.part.add(pos=[0, 0, 1], q=1)
```
reinaual: Is there a reason to save the handle? aesthetics?
```suggestion
        self.system.part.add(pos=[0, 0, 1], q=1)
```

testsuite/python/pair_criteria.py
```diff
@@ -33,10 +33,9 @@ class PairCriteria(ut.TestCase):
     es.bonded_inter.add(f1)
     f2 = FeneBond(k=1, d_r_max=0.05)
     es.bonded_inter.add(f2)
-    es.part.add(id=0, pos=(0, 0, 0))
-    es.part.add(id=1, pos=(0.91, 0, 0))
-    p1 = es.part[0]
-    p2 = es.part[1]
+    es.part.add(pos=(0, 0, 0))
+    es.part.add(pos=(0.91, 0, 0))
+    p1, p2 = es.part[:]
```
reinaual:
```suggestion
    p1 = es.part.add(pos=(0, 0, 0))
    p2 = es.part.add(pos=(0.91, 0, 0))
```

testsuite/python/rdf.py
```diff
@@ -107,20 +107,23 @@ def test_mixed(self):
...
-        s.part.add(id=1, pos=[0, 0, 0], type=1)
-        observable = espressomd.observables.RDF(ids1=s.part[:].id[0::2],
-                                                ids2=s.part[:].id[1::2],
+        s.part.add(pos=[0, 0, 0], type=0)
+        s.part.add(pos=[0, 0, 0], type=1)
+        s.part.add(pos=[0, 0, 0], type=0)
+        s.part.add(pos=[0, 0, 0], type=1)
+        pids1 = [s.part[:].id[0]]
+        pids2 = [s.part[:].id[1]]
+        observable = espressomd.observables.RDF(ids1=pids1, ids2=pids2,
```
reinaual: I dont get the change in this testcase.
Is there a reason to double the particle count and not use them in the observable?
Why dont we add them in one call and use the returned particle handles for the ids?

testsuite/python/rotation_per_particle.py
```diff
@@ -28,47 +28,49 @@ class Rotation(ut.TestCase):
...
         """Applies langevin thermostat and checks that correct axes get
            thermalized"""
         s = self.s
         s.thermostat.set_langevin(gamma=1, kT=1, seed=42)
-        for x in (0, 1):
-            for y in (0, 1):
-                for z in (0, 1):
+        for x in (False, True):
+            for y in (False, True):
+                for z in (False, True):
```
reinaual: I would suggest to replace the three loops with 
`for x, y, z in itertools.product((False, True), repeat=3):`

testsuite/python/rotational_inertia.py
```diff
@@ -60,17 +58,17 @@ def test_stability(self):
         # < J[0] < J[2].
         J = np.array([5, 0.5, 18.5])
 
-        self.system.part[0].rinertia = J[:]
+        p.rinertia = J[:]
```
reinaual: If i saw it correctly `J` is not used anywhere, so these lines can be merged

testsuite/python/lb_switch.py
```diff
@@ -50,33 +50,33 @@ def switch_test(self, GPU=False):
...
         system.integrator.run(0)
 
-        np.testing.assert_allclose(np.copy(system.part[0].f), 0.0)
+        np.testing.assert_allclose(np.copy(p.f), 0.0)
 
         system.actors.add(lb_fluid_2)
         system.thermostat.set_lb(LB_fluid=lb_fluid_2, gamma=friction_2)
 
-        for p in product(range(5), range(5), range(5)):
+        for pid in product(range(5), range(5), range(5)):
```
reinaual:
```suggestion
        for pid in product(range(5), repeat=3):
```

reinaual (commented):
This review was a lot of work... Overall LGTM
Sorry about addressing some things you have not touched.

testsuite/python/mmm1d.py
jngrad: `_` has a special meaning in Python and is often used to discard temporary values, in this case I used if for aesthetics but I wouldn't mind removing it

testsuite/python/rdf.py
jngrad: Nice catch, that's a copy paste error, I meant to take a slice, not the first element. The idea is to initialize the object with `pids1=[0, 2]` and `pids2=[1, 3]` and later update the observable with `pids1=[0]` and `pids2=[1]` to check the script interface setter correctly updated the core object.

testsuite/python/rotation_per_particle.py
jngrad: good idea!

testsuite/python/mmm1d.py
reinaual: I know about the convention, I only use it for tuple-unpacking. I was pointing it out because there are other occasions with the same situation
https://github.com/espressomd/espresso/blob/2293827912cfc8ab0657aadd11455c179a11eef7/testsuite/python/interactions_bond_angle.py#L41-L43

reinaual (approved):

PR #4181, opened 2021-03-25
Remove mis-use in Rattle

### LINKED ISSUE #4141: Rattle: Remove mis-use of particle data members
Rattle stores velocity corrections in the particle force attribte and then forces in the p_old attribute.
This stands in the way of separating state (position, velocity) and observable (force).

If, during the execution of Rattle, the particles are not resorted, and hence, the order in which they are processed, does not change, the temporary values can be stored in a vector.
* If the order does change, but no ghost communication is needed, an unordered_map with the particle id could be used
* If ghost communication is needed, a particle property (for the velocity correction I assume) needs to be added and also added to the ghost communicator

* The ParticlePosition::p_old property of rattle should be renamed (e.g. to p_last_time_step) to avoid confusion with ParticleLocal::p_old (which should be p_last_verlet_update).
* It needs to be checked whether Rattle's p_old can be moved to ParticleLocal, or if it actually needs to be ghost ommunicated.

https://www.sciencedirect.com/science/article/pii/0021999183900141

reinaual:
Fixes #4141

Description of changes:
- removed mis-use of the force-reduction mechanism by introducing new particle property
- reduction mechanism for the new particle property
- cleanup of the rattle-implementation
- introduced early-breakout

### DIFF
--- a/src/core/Particle.hpp
+++ b/src/core/Particle.hpp
@@ -365,6 +365,26 @@ struct ParticleLocal {
   }
 };
 
+#ifdef BOND_CONSTRAINT
+struct ParticleRattle {
+  /** position/velocity correction */
+  Utils::Vector3d correction = {0, 0, 0};
+
+  friend ParticleRattle operator+(ParticleRattle const &lhs,
+                                  ParticleRattle const &rhs) {
+    return {lhs.correction + rhs.correction};
+  }
+
+  ParticleRattle &operator+=(ParticleRattle const &rhs) {
+    return *this = *this + rhs;
+  }
+
+  template <class Archive> void serialize(Archive &ar, long int /* version */) {
+    ar &correction;
+  }
+};
+#endif
+
 /** Struct holding all information for one particle. */
 struct Particle { // NOLINT(bugprone-exception-escape)
   int &identity() { return p.identity; }
@@ -391,6 +411,10 @@ struct Particle { // NOLINT(bugprone-exception-escape)
   ParticleForce f;
   ///
   ParticleLocal l;
+#ifdef BOND_CONSTRAINT
+  ///
+  ParticleRattle rattle;
+#endif
 
 private:
   BondList bl;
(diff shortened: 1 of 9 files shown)


src/core/Particle.hpp
```diff
@@ -268,7 +268,7 @@ struct ParticlePosition {
 
 #ifdef BOND_CONSTRAINT
   /** particle position at the previous time step (RATTLE algorithm) */
-  Utils::Vector3d p_old = {0., 0., 0.};
+  Utils::Vector3d p_last_timestep = {0., 0., 0.};
```
jngrad, line 271: From what I understand, this member cannot be moved to `ParticleLocal` (7713989ad856a) because we need to communicate it on ghosts, correct?

jngrad (commented):
Overall LGTM, although I couldn't follow everything in the last commit.

src/core/Particle.hpp
reinaual: The problem is that the correction term is calculated using `p.r.p` as well as `p.r.p_old` and both of them have to be present on the ghost particles. If my understanding is correct this means that i can't use `ParticleLocal` for that.
https://github.com/espressomd/espresso/blob/294f4905578f26b08826e463426b61688fc4a18e/src/core/rattle.cpp#L82-L91

reinaual: I might be missing something here, but i should be able to get this value from the start of the integration and store it locally because it is only a copy of the `p.r.p` value... Let me have a look at that tomorrow with a fresh mind.

reinaual: If my understand is correct it is not possible to store `p_last_timestep` in `ParticleLocal` because of this call which is resorting the cellstructure.
https://github.com/espressomd/espresso/blob/1e37bedd09f9569a45a4129db375e624abdbd157/src/core/rattle.cpp#L134

If we need to get rid of it from the `ParticlePosition` struct i guess a new one needs to be introduced because highjacking the `ParticleRattle` struct is probably not a good idea since a reduction is operating on it and a different communication-operation is expected for `p_last_timestep`.

RudolfWeeber (approved):

jngrad (approved):

PR #4182, opened 2021-03-26
structure factor testing and CMakeList edit

### LINKED ISSUE #270: analyze structurefactor has no testcase

stekajack:
Fixes #270

Description of changes:

- added a simple test case for Structure Factor. Make a simple cubic lattice with a!=b!=c. Compare scattering peaks with known results (multiples of pi, and all diagonal combinations of those). Test truncated to match up to 1.e-02 precision 9 scattering peaks.
- updated testsuite/python/CMakeList.txt so test is run with make check.


jngrad:
Two containers failed in CI because the structure factor C++ code is broken. The Clang compiler helped me find three instances of undefined behavior. There might be more bugs hiding in the C++ code, but I'm not sure what to look for. You can find my changes in [jngrad/espresso@pr-4182](https://github.com/stekajack/espresso/compare/080930fde4bfb3d8bfd1dc772013096ef6a95ad0...jngrad:pr-4182).

testsuite/python/sf_simple_lattice.py
```diff
@@ -0,0 +1,63 @@
...
+        self.sf_data = [(x, y)
+                        for (x, y) in zip(sf_observable_x, sf_observable_y)]
+        self.sf_data.sort(key=lambda tup: tup[1])
+
+    def test(self):
+        self.make_lattice()
+        self.peaks()
+        self.scatter_lattice()
+        self.assertTrue([np.any(np.isclose(element, self.sf_data, rtol=1e-02))
+                         for element in self.diags])
```
jngrad, line 59: This checks the wavevector in `element` against `self.sf_data`, which contains tuples of wavevectors and their structure factor. You probably meant to check that the wavevector `element` was contained in wavevectors `sf_observable_x`, and also that the structure factor in `sf_observable_y` are highest for the diagonals.

stekajack: so I only wanted to check here whether the peak positions from SF correspond to the tabulated values contained in the self.diags. previously I sort the sf_data, so the intensity test is basically implicit, and I wanted to slice out most of sf_data (I did forget to do so xD), because fitting 9 peaks is enough. I see your point, because I leave the tuple, not just the peak positions. however, it works as intended. Also I'm sorry to see is calls on C level, i just compiled on my machine and ran the tests, passed just fine.

jngrad: Alright, if I understand your test correctly you are looking at the scattering of a tetragonal lattice.

I've rewritten the test in [jngrad/espresso@pr-4182](https://github.com/stekajack/espresso/compare/080930fde4bfb3d8bfd1dc772013096ef6a95ad0...jngrad:pr-4182) to check for the simple cubic, body-centered cubic and face-centered cubic lattices. In each test case, the structure factor must be either 0 (no reflection) or equal to the number of particles in the lattice (reflection). For those reflections, the wavevector should match the reflection rule for the h, k, l indices of the lattice.

Could you please have a look at the new test and let me know if this makes sense to you? There are a few unknowns from my side:
* The simple cubic lattice is missing reflections at `q=7` and `q=15`. This is mentioned in the [C++ docstring](https://github.com/espressomd/espresso/blob/ee9f3010f595a809ac1024e1c698b6482bf4eb5f/src/core/statistics.hpp#L93-L99), but I don't understand the explanation.
* The S(q) plots for the c-centered orthorhombic and tetragonal lattices (jngrad/espresso@4a0c3d24cb950bff7) don't make sense to me; for example, there are peaks whose intensity is not an integer multiple of the number of particles.

KaiSzuttor: as a side note: if it is not clear what is being tested, please add a docstring to your test

stekajack: having played with the code in jngrad/espresso@pr-4182, everything seems legit, with expected results. Before, the test was aimed to really only test for positions of the peaks and whether it correctly recognises lengths (lattice constant was different for in each direction). That would suffice as a test but like this it is maybe more systematic. 

So the intensity checking is I presume there to demonstrate that the average works as expected. In can case properly normalised S(q) shouldn't be size dependant for isotropic and center-to-center distance.

The reason that peaks at q=7 and 15 are missing is just because from the definition with miller indices, one cannot find integer solutions for the (I, j, k) set. That is in line with what should come out (also for fcc bcc ..).

I think that the reason that one gets these interference terms for c-centered orthorhombic lattice  (jngrad/espresso@4a0c3d2) is a finite size effect. If you draw a box around the lattice you will notice that it basically misses 2 rows of particles, which when averaged gives these interference floats. Not sure what to do with that...

jngrad: > Before, the test was aimed to really only test for positions of the peaks and whether it correctly recognises lengths (lattice constant was different for in each direction). That would suffice as a test but like this it is maybe more systematic.

The new test is also testing for the lattice constants, since they are component of the wavevectors, and now all wavectors are tested. I would really like to have a test case for a lattice with lower symmetry such as tetragonal or c-centered orthorhombic, so as to get different lattice constants in each direction.

> So the intensity checking is I presume there to demonstrate that the average works as expected. In can case properly normalised S(q) shouldn't be size dependant for isotropic and center-to-center distance.

Yes, in the new test case I use the special case of S(q) where it is proportional to |Fhkl|^2.

> I think that the reason that one gets these interference terms for c-centered orthorhombic lattice (jngrad/espresso@4a0c3d2) is a finite size effect. If you draw a box around the lattice you will notice that it basically misses 2 rows of particles, which when averaged gives these interference floats.

Hmm, I can't find the missing rows. With periodic images activated, the c-centered orthorhombic lattice looks complete to me:
```python
visualizer = espressomd.visualization_opengl.openGLLive(
    system, background_color=[1, 1, 1], periodic_images=[1, 1, 1])
visualizer.run(1)
```
The radial distribution function also looks normal:
```python
import espressomd.observables
obs = espressomd.observables.RDF(ids1=system.part[:].id, min_r=0.1,
                                 max_r=6., n_r_bins=100)
rdf = obs.calculate()
r = obs.bin_centers()
plt.plot(r, rdf)
plt.show()
```

stekajack: So it took me a while to test out my suspicion but i have arrived at an educated guess. Firstly, i didn't want to imply that the lattice that you construct was wrong. Indeed with pbc you reproduce a proper lattice. But it became apparent that results are overly reliant on having a good choice of box size, and the c-centered orthorhombic primitive cell is difficult to replicate in to a full lattice with pbc. You would need a particular multiple of primitive cells to manage that, which you actually did with your code. 

But one should be able to do scattering on a singe primitive cell and reproduce literature results. So i tested again for bcc and fcc just for a single primitive cell. In tests that worked for you you always had a lattice spacing of 4 and a even multiple of lattice spacings. So just making a primitive and wrapping it in a simulations box doesn't work, probably because of how pbc plays a role in the calculation. A simulation box overhang is necessary apparently. So one should stress in the documentation this limitation. However one can place a bravais lattice snuggly in a simulation box and that works perfectly. But an odd multiple of lattice spacings in a simulation box doesn't work.

 Let's take the FCC example. The formula espresso uses 2*pi/L*(l*x+m*y+*z), where for example x=k*a. Now if a=4 and box_l=8 this means that L=2*a, so stuff cancels out, cosine values end up either 1 or 0, and we get reflections at 3,4,8,... as we should. However, take now instead of a=1 in the same box. Now L=8*a, we get quarter of pi multiple, that end up giving values not necessarily 0 and 1, and this doesn't work. I think in this case we get reflections at q=32 if i remember correctly. in other cases we get interference terms which shouldn't be there, like for the c-centered orthorhombic test. Furthermore, due to the fact that box size is a part of this calculation, if the lattice constant and box length don't coincide in a way that the calculation above works out nicely, the S(q) calculation can understand wrongly the lattice size, and even give accurate results, just for a different lattice.

This is where i am with the understanding of the current S(q) implementation. As it is, I would say the testing is successful, and the calculation accurate, given that we make clear under which conditions.

jngrad: ok, I went back to your original test and figured out how to extend the `peaks(self)` function to the general case where we filter peaks by the reflection conditions on (h, k, l): [`generate_peaks(self, a, b, c, condition, cutoff=box_l)`](https://github.com/jngrad/espresso/blob/a5d4223516f904327f92c7c46d236dd23a4183d7/testsuite/python/sf_simple_lattice.py#L41-L62). Your original tetragonal test is also back in. I opened a new PR to avoid editing your branch.

There is indeed something special about the orthorhombic c-centered lattice, maybe we should keep it in the test as an edge case. If a subtle bug is introduced in the C++ code, the orthorhombic c-centered case will be more sensitive to it than the cubic and tetragonal lattices.

KaiSzuttor:
so this PR can be closed? I think you could have just pushed onto that one @jngrad

jngrad:
> so this PR can be closed? I think you could have just pushed onto that one @jngrad

The last time I tried to push to the `python` branch of a fork, I got a nasty surprise after the second force push:
```
 ! [remote rejected]     fix-4143 -> python (permission denied)
error: failed to push some refs to 'git@github.com:KaiserMartin/espresso.git'
```

PR #4193, opened 2021-03-30
utils: Made Array streamable

### LINKED ISSUE #4189: Implement string stream << for Utils::Vector
Otherwise, it is not possible to write BOOST_CHECK_EQUAL and friends with vectors.

fweik:
Fixes #4189.

Description of changes:
- Implemented `operator<<` for Utils::Array

Notes:
It is not easily possible to add a templated `operator<<` (as one normaly would),
because this is picked up by boost.serialization, which in turn breaks. So I implemented
it only for `std::ostream`, which I think is fine for most cases.

### DIFF
--- a/src/utils/include/utils/Array.hpp
+++ b/src/utils/include/utils/Array.hpp
@@ -26,8 +26,9 @@
 
 #include <cassert>
 #include <cstddef>
+#include <iterator>
+#include <ostream>
 #include <stdexcept>
-
 namespace Utils {
 
 namespace detail {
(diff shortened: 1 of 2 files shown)


src/utils/include/utils/Array.hpp
```diff
@@ -26,6 +26,8 @@
 
 #include <cassert>
 #include <cstddef>
+#include <iostream>
```
jngrad: Isn't `ostream` sufficient? `iostream` brings in initialization code and a global state.

fweik: Right.

KaiSzuttor (approved):

PR #4195, opened 2021-03-31
Implement slicing for LB

### LINKED ISSUE #4143: Implement slicing support for LB

jngrad:
Fixes #4143

Description of changes:
- Modified __getitem__ to detect slices in input keys
- Introduced LBSlice class that calculates node indices from slice input and calls single node getter from LBFLuidRoutines class for every node in the slice.

The output is a multidimensional array containing the requested quantity and is of dimension len(slice_x)*len(slice_y)*len(slice_z)*len(quantity), where len(quantity) is the dimensions of the quantity, for example 3 in case of velocity. The output array is arranged in the way numpy cycles through slices. Therefore, when setting attributes one has to set the dimensions of the input array exactly the same way numpy cycles through the sliced nodes.
An error is issued if the dimensions of the slices and the input does not match.


KaiSzuttor (approved):

KaiSzuttor:
@KaiserMartin thanks a lot for your implementation

Issue #4015, opened 2020-11-26
Investigate switching integrator per particle


RudolfWeeber:
Right now, the integrator is chosen globally (via integ_switch).
Specific intregrators contain their own loop over all particles.

This has two weaknesses:
* some common functions like Verlet list updates have to copied into each integrator, or an extra iteration over all particles is incurred, if Verlet list management is to be separated
* Inertialess tracers are currently implemented as virtual sites, although, they conceptually are more like an integrator (#2695, #3059). 

Using definitions:
* virtual sites ar particles who's properties position/orientation/velocity are dreived from other particles
* integrators implement rules that propagate the particles

We should investigate, whether
* exposing integrator per-particle kernels rather than sweeps
* putting the sweep over all particle in the central integration function and using switching kernels on a per-particle level
* Verlet list criterion check can then be integrated as an additional per-particle kernel
* particles can have an additional flag which overrides the default integrator This would be used for the inertialess tracers

In a second step (or in one go), integrator parameters can be put into classes. This would be time step for velocity verlet, pre-factor and cap for steepest descent, etc.

@fweik, this is how I understood our discussions on the topic. Is there something missing?

fweik:
Yes that sounds reasonable. If you want to coerce the virtual sites into that abstraction (they are really not per particle), than they will always need special treatment. Also it has to be considered if rattle can be integrated into this framework.

For the virtual sites I think abstractly what you want to do is in the first sweep defer some particles for later integration.
You could e.g. have integrators return either new positions for a particle or a defer tag, in which the particle is noted for
later propagation. Then, if there are deferred particles, you do the next loop only with the left ones and so on until there
are none left. Potentially rattle also fits into this construction e.g. rattle could defer particles until it is happy with the tolerance.

jngrad:
# Equations of motion for integrators and thermostats

Used notation:
* position: $x(t)$
* quaternion: $q(t)$
* velocity: $v(t)$
* angular velocity: $\omega(t)$
* force: $f(t)$
* torque: $\tau(t)$
* mass: $m$
* moment of inertia: $I$
* interpolated fluid velocity: $u(x(t))$
* box length: $L(t)$
* time step: $dt$
* temperature: $T$
* friction: $\gamma$
* noise: $\eta()$
* quaternion rotation matrix: $R()$

## Velocity Verlet

* [documentation](http://espressomd.org/html/doc/running.html#velocity-verlet-algorithm)
[...]

PR #4198, opened 2021-04-03
core: Consider rotational kinetic energy of virtual particles.

### LINKED ISSUE #3596: Rotational degrees of virtual sites do not contribute to kinetic energy
From `energy_inline.hpp`:
```c++
inline void add_kinetic_energy(Particle const &p1) {
  if (p1.p.is_virtual)
    return;

...

    // Note that rotational degrees of virtual sites are integrated
    // and therefore can contribute to kinetic energy
#ifdef ROTATION
  if (p1.p.rotation) {
 ...
```

The situation is actually more complicated, because this depends on the virtual sites
parameters. Maybe `VirtualSites::m_have_quaternion` should be dropped in favor of
obeying the rotation flag of the virtual particle. Also there is of course no test.

fweik:
Fixes #3596.

This is a fix closest to the existing behavior. I think there should be a discussion here how
to handle active/inactive degrees of freedom, and if this should be consistent between
translation and rotation, and how it should interact with the virtual sites relative feature.

Description of changes:
- Don't drop rotational kinetic energy of virtual particles.


RudolfWeeber:
>I think there should be a discussion here how
>to handle active/inactive degrees of freedom, and if this should be consistent between
>ranslation and rotation, and how it should interact with the virtual sites relative feature.

 

Yes. The current behaviour with regards to VS relative and rotation is hard to understand.

My current understanding is that, the propagation for rotational and translational degrees should be controllable independently and per particle. E.g., users could specify

 

*	Posiotn propagation: Relative to other particle
*	Rotational propagation: Rotational Langevin eqn

 

From that information, it should be derivable whether the dof is ‘active’ or not.

 

This would be part of #4015.

fweik:
In my mind, in an ideal world you'd be able to add or remove degrees of freedom to a particle/point, and they would themselves know how they are propagated and what their energy/momentum contribution is. But probably such a fundamental design change will be hard to pull of, so I think what you are proposing is a reasonable direction. We should discuss beforehand who (in terms of components) should be responsible for that, and which choices yield that best separation of concerns, given the circumstances.

KaiSzuttor (dismissed):

jngrad (approved):

PR #4199, opened 2021-04-05
Remove code duplication in non-bonded interactions test

### LINKED ISSUE #4168: Remove code duplication in non bonded test
Without the code duplication, tests could be expanded, e.g., with regards to checkpointing, effectively.

In testsuite/python/interactions_non-bonded.py, for each non bonded interaction, the same test code is repeated:
* loop over some distances
* move particle
* compare energy/force/pressure

this should be done analogous to the interactions_bonde.dpy:
There, for each interaction, a lambda which maps distance to force and energy is defined.
This is then passed to a common test function, which runs the loop.

This task requires Python knowledge only and there is a template  (the bonded test) to work from.

davidbbeyer:
Fixes #4168

Description of changes:
- remove code duplication in test of non-bonded interactions (except Gay-Berne potential) by using a common test function run_test


jngrad (commented):
Thanks @davidbbeyer! From my side, this is ready to be merged.

KaiSzuttor (approved):
it would probably more expressive to use kwargs for the test function but I think we can live with it for now

PR #4200, opened 2021-04-06
Introduced a bugfix to solve the infinite loop in constant pH  algorithm 

### LINKED ISSUE #4197: Constant pH method can enter an infinite loop if some particles are not created initially
Due to a mistake in my code, I have learned that there is a small bug in the constant pH algorithm. I have observed that this bug happens in the actual development version of espresso. Consider the typical acid/base equilibrium of a weak acid

HA <-> A- + H+

HA: type 0
A-: type 1
H+: type 2

Consider that, by mistake, the user does not create any particle of types 0 and 1, but it creates particles of type 2. In these conditions, it is of course impossible to have the reaction in either direction. When attempting to do a reaction step, espresso enters in an infinite loop-like behaviour that needs to be killed from the terminal. Below I attach a code with a minimal setup that reproduces this bug. I have also noticed that if the user does not create any particle at all the program returns a segmentation fault error. These errors do not happen in using the reaction ensemble algorithm, which simply does not do the reaction if it is not possible.

In my opinion, this bug should be easily solved by checking the number of particles of each reactant at the beginning of the constant pH algorithm and exiting the subroutine if the reaction is impossible. I will work on a bugfix myself and submit it for your consideration.

Best,

Pablo M. Blanco

Test code with the described bug:

``` python
import espressomd

charge={0:0, 1:1, 2:-1}

typ=2

system.part.add(id=[0], pos=[[0,0,0]], type=[typ], q=[-1])

RE = reaction_ensemble.ConstantpHEnsemble(temperature=1, exclusion_radius=1, seed=1)
RE.constant_pH = 7

RE.add_reaction(gamma=1, reactant_types=[0], reactant_coefficients=[1],
                product_types=[1,2],
                product_coefficients=[1, 1], default_charges=charge)

print("Before reaction")
RE.reaction(1)
print("After reaction")

```

pm-blanco:
Fixes #4197

When attempting to do a reaction step using the constant pH algorithm, if there are not enough reactants nor product particles, the program simply not does the operation. The apparition of an infinite loop is prevented by existing the reaction subroutine in case that neither direction of the reaction is possible. 

Description of changes:
- The program checks if at least one direction of the reaction is possible and otherwise exits the subroutine returning a 1. This value should be later processed so the user gets a warning message informing of this error (work in progress).

### DIFF
--- a/src/core/reaction_ensemble.cpp
+++ b/src/core/reaction_ensemble.cpp
@@ -1497,18 +1497,12 @@ int ConstantpHEnsemble::get_random_valid_p_id() {
  */
 int ConstantpHEnsemble::do_reaction(int reaction_steps) {
 
-  bool reaction_possible = false;
 
   for (int reaction_id = 0; reaction_id < reactions.size(); reaction_id++) {
-    if (all_reactant_particles_exist(reaction_id)) {
-      reaction_possible = true;
+    if (all_reactant_particles_exist(reaction_id) == false){return 1;
     }
   }
-
-  if (reaction_possible == false) {
-    return 1;
-  }
-
+      
   for (int i = 0; i < reaction_steps; ++i) {
     // get a list of reactions where a randomly selected particle type occurs in
     // the reactant list. the selection probability of the particle types has to
--- a/src/core/reaction_ensemble.cpp
+++ b/src/core/reaction_ensemble.cpp
@@ -1497,12 +1497,19 @@ int ConstantpHEnsemble::get_random_valid_p_id() {
  */
 int ConstantpHEnsemble::do_reaction(int reaction_steps) {
 
+
+  bool reaction_impossible = true;
+
   for (int reaction_id = 0; reaction_id < reactions.size(); reaction_id++) {
-    if (all_reactant_particles_exist(reaction_id) == false) {
-      throw std::runtime_error("There is not enough reactants nor products to "
-                               "perform the reaction in either direction");
-      ;
+    if (all_reactant_particles_exist(reaction_id)) {
+      reaction_impossible = false;
     }
+  }
+
+  if (reaction_impossible){
+   
+   throw std::runtime_error("There is not enough reactants nor products to perform the reaction in either direction");
+
   }
 
   for (int i = 0; i < reaction_steps; ++i) {
--- a/src/core/reaction_ensemble.cpp
+++ b/src/core/reaction_ensemble.cpp
@@ -51,8 +51,6 @@
 #include <utility>
 #include <vector>
 
-
-
 namespace ReactionEnsemble {
 
 /** Load minimum and maximum energies as a function of the other collective
@@ -1499,13 +1497,17 @@ int ConstantpHEnsemble::get_random_valid_p_id() {
  */
 int ConstantpHEnsemble::do_reaction(int reaction_steps) {
 
-  bool reaction_possible=false;
+  bool reaction_possible = false;
 
   for (int reaction_id = 0; reaction_id < reactions.size(); reaction_id++) {
-	if (all_reactant_particles_exist(reaction_id)){reaction_possible=true;}	
-	}
+    if (all_reactant_particles_exist(reaction_id)) {
+      reaction_possible = true;
+    }
+  }
 
-  if(reaction_possible==false){return 1;}
+  if (reaction_possible == false) {
+    return 1;
+  }
 
   for (int i = 0; i < reaction_steps; ++i) {
     // get a list of reactions where a randomly selected particle type occurs in
--- a/src/core/reaction_ensemble.cpp
+++ b/src/core/reaction_ensemble.cpp
@@ -51,6 +51,8 @@
 #include <utility>
 #include <vector>
 
+
+
 namespace ReactionEnsemble {
 
 /** Load minimum and maximum energies as a function of the other collective
@@ -1497,6 +1499,14 @@ int ConstantpHEnsemble::get_random_valid_p_id() {
  */
 int ConstantpHEnsemble::do_reaction(int reaction_steps) {
 
+  bool reaction_possible=false;
+
+  for (int reaction_id = 0; reaction_id < reactions.size(); reaction_id++) {
+	if (all_reactant_particles_exist(reaction_id)){reaction_possible=true;}	
+	}
+
+  if(reaction_possible==false){return 1;}
+
   for (int i = 0; i < reaction_steps; ++i) {
     // get a list of reactions where a randomly selected particle type occurs in
     // the reactant list. the selection probability of the particle types has to
@@ -1517,6 +1527,7 @@ int ConstantpHEnsemble::do_reaction(int reaction_steps) {
 
       // construct list of reactions with the above reactant type
       for (int reaction_i = 0; reaction_i < reactions.size(); reaction_i++) {
+
         SingleReaction &current_reaction = reactions[reaction_i];
         for (int reactant_i = 0; reactant_i < 1;
              reactant_i++) { // reactant_i < 1 since it is assumed in this place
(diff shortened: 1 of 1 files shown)


src/core/reaction_ensemble.cpp
```diff
@@ -1497,6 +1497,18 @@ int ConstantpHEnsemble::get_random_valid_p_id() {
...
+  for (int reaction_id = 0; reaction_id < reactions.size(); reaction_id++) {
+    if (all_reactant_particles_exist(reaction_id)) {
+      reaction_possible = true;
+    }
+  }
+
+  if (reaction_possible == false) {
+    return 1;
+  }
+
```
KaiSzuttor: isn't it possible to get this kind of logic implicitly? If no particle of the reactant type is found nothing should happen automatically without manually checking?

pm-blanco: The problem that causes the infinite loop has its source in the while loop in lines 1521-1546. This while loop searches  the particle ids that have a type that matches the reactant/product types. Of course, if you do not create such particle, then this loop runs for ever.
As you suggest, another possible solution would be to do this task without using a while statement so that if there is no reactant then the program does nothing. However, in my opinion, manually checking it has the advantage that we can integrate an exception error preventing the user from doing such simulation (since I cannot think where one would attempt to run the algorithm without reactants if it is not by mistake).

pm-blanco: The problem that causes the infinite loop has its source in the while loop in lines 1521-1546. This while loop searches the particle ids that have a type that matches the reactant/product types. Of course, if you do not create such particle, then this loop runs for ever.
As you suggest, another possible solution would be to do this task without using a while statement so that if there is no reactant then the program does nothing. However, in my opinion, manually checking it has the advantage that we can integrate an exception error preventing the user from doing such simulation (since I cannot think where one would attempt to run the algorithm without reactants if it is not by mistake).

```

Comment (kosovan) [reply src/core/reaction_ensemble.cpp]: The point is that such situation should not occur. If it occurs, then it is an error in the system setup, so it cannot be a simple no-op. It should raise an error.

Comment (KaiSzuttor) [reply src/core/reaction_ensemble.cpp]: in my mind the reaction algorithm itself should be independent of the state of the system. I can also define a non bonded interaction of two non existing particle types without having an infinite loop. The reaction of reactants is also some kind of particle interaction, right? Why should it be special here?

Comment (jngrad) [reply src/core/reaction_ensemble.cpp]: All methods default to a no-op when not enough particles are present, except for `WidomInsertion` which throws an error. I don't immediately see why it should be an issue to not have enough particles of type HA if there is at least one particle of type A-. Could you please clarify this for us?
Also if you reverse the reactant types with the adduct types, you will get a bug in the other direction, if I understand e4104d28e6 correctly.

Comment (pm-blanco) [reply src/core/reaction_ensemble.cpp]: @jngrad  Considering the reaction HA <--> A- + H+; If you don't have any particle HA, in addition of having at least one particle of type A-  you also need at least a counter-ion particle H+ so the reaction is possible in the reverse way. In the current develop version creating either only A- or H+ particles leads to an infinite loop error because the program searches in vain particles matching the missing type. You are right, in e4104d2  I introduced the reverse bug, I think that I have solved the problem in the most recent commit.
@KaiSzuttor If I understood you right, you suggest that it could be defined that a reaction is possible between some particle types even if they are not actually created in the system. In this case however, what the user is attempting is to actually perform the reaction, which is impossible if there are not enough particle reactants or products.

**Comment by RudolfWeeber** [comment]
In my view, there are two issues:
* The function `ConstantPHEnsemble::get_random_valid_p_id()` is incorrect, because it contains an empty loop if it has no solution. This should be fixed by changing the return type to `std::optional<int>` and returning `{}` if the system is empty. The quickest fix is to check that the total number of particles is >0. Long-term, the function can be replaced by something more efficient for the case that there are many holes in the particle id list.
* The question of what should happen if no adducts are present, more more technically, what should happen in `do_reaction()` if `get_random_valid_pid()` returns `{}`. In my view, no-op is the correct behavior. This is analogous to running 
```python
system.cell_system.skin = 0
system.time_step = 1
system.integrator.run(1)
```
without particles (no-op, as-well)

RudolfWeeber:
Espresso also allows definition of non-bonded interactions for types for which there are (at that point) no particles.
The interaction will become used as soon as relevant particles are there.
So, there as well no-op would be the consistent behavior.

The Python call `re.reaction()` might return the number of attempted and performed reactions, just like system.integrator.run() does.
In this way, the user could handle the case of no educts present, if they wish to.

kosovan:
> The Python call `re.reaction()` might return the number of attempted and performed reactions, just like system.integrator.run() does.
> In this way, the user could handle the case of no educts present, if they wish to.

I like the solution proposed by @RudolfWeeber Indeed, the user might want to first create a reaction and then particles.

However, if we change the behaviour and return value of `re.reaction()`, then it would require more extensive changes, beyond a simple bugfix. I have some other suggestions on my list when it comes to modifying the reaction ensemble behaviour. Therefore, I propose to implement a simple bugfix now (throw an exception) and start creating a wishlist of deeper changes for the next coding day.

src/core/reaction_ensemble.cpp
jngrad: To elaborate on @KaiSzuttor's remark, shouldn't it be possible to use the Constant pH method with competitive reactions? For example HA <--> A- + H+ together with A- + BX <--> AB + X- so that reaction B consumes all AH (by driving the acid-base reaction forward) and leaves HX as the acid. One application would be the S<sub>N</sub>2 reaction between HBr and alkyl chlorides.

pm-blanco: @jngrad  thank you for your clarification, now I understand your point. Now I agree with you that the desired output would be a no-op so one could couple the acid/base reaction with other reactions

kosovan: Indeed, @jngrad is right that all reactants could be consumed if different reactions are competing for the same product or reactant. However, I would discourage anybody from intentionally using RxMC in this regime. It suffers from serious finite-size artifacts if the number of (any type of) reacting particles is on the order of unity. It is related to the fact that particle numbers in a simulation have to be integers whereas most existing theories work with concentrations as continuous variables.

kosovan: Note: SN2 refers to a specific reaction mechanism. A reaction of the type A- + BX <--> AB + X- is not compatible with the way currently reactions are implemented in Espresso. The current implementation allows to create or destroy particles at random positions but it does not allow to transfer a particle between two binding partners. It is also one of the points on our wishlist.

jonaslandsgesell: Be aware it definitely depends on what you mean:
Of course it is totally valid to create this reaction 
A+B <-> C+D 
Whit artbitrary A, B, C, D. I.e. also :
 A- + BX <--> AB + X-
However all the types will be "primitive" beads i.e. not have bonds in the bead "BX". But it is completely fine for the "BX"  type to have different lenard jones parameters (i.e. appear bigger) etc...
Maybe this would require having a type dependent exclusion radius for some more extreme variations...

jonaslandsgesell:
The problem #4197 is fixed via circumventing the asymetric proposal probability in #4207. The PR effectively shortens the code and makes it simpler. There is no need for a more complicated solution or further discussion of symptomatic fixes ;)

src/core/reaction_ensemble.cpp
jngrad: > A reaction of the type A- + BX <--> AB + X- is not compatible with the way currently reactions are implemented in Espresso. The current implementation allows to create or destroy particles at random positions but it does not allow to transfer a particle between two binding partners.

Good point. I also thought about the case of the titration curve of a buffer solution, i.e. a mixture of a weak acid and a strong acid. Titration by a strong base can trigger the bug once the buffer capacity is used up. But that's an edge case, one probably wouldn't want to add more base than there is buffer.

jngrad:
PR superseded by #4207.

PR #4202, opened 2021-04-06
core: Avoid some redudant lookups in the particle type map

fweik:
Description of changes:
- The particle type map stores an index of particle ids for the particle types, this switches the
  the data type from an unordered_set to a flat_set, which is more efficient for the typical use
  of the index, which is picking a random particle of a type in the hot loop of MC methods.
- Avoid some redundant lookups in the type index


fweik:
@KaiSzuttor

jonaslandsgesell:
Hi Floh, 

Just out of curiosity do you have some timings showing a significant improvement for the reaction ensemble for O(200) or O(1000) particles where this set is touched in every accepted reaction step?
I just read this Blogpost which has timings indicating that flat set gets slower for bigger sizes  https://www.google.com/amp/s/www.r-bloggers.com/2014/01/insert-and-remove-performance-of-boosts-flat_set-v-s-stdset/amp/

fweik:
From my understanding the performance relevant operation here is getting the n-th element from the set, which is a constant time op for the flat_set and linear (with poor cache performance) for the unordered_set. You are right that the asymptotic complexity for insert/remove for the unordered_set is better, but there are other things to consider: The flat_set has all data contiguously in memory (it really is just a sorted vector), while the unordered_set is a heap structure with indirection, so the flat version has way better cache performance and (amortized) does not need allocations on insert, both of which are not true for the unordered_set.
For very small sets (like a dozen ints or similar) it is even faster to just use a vector with linear search, due to the even better cache performance (bisection jumps around, while linear search does not). Overall for the expected sizes in this case (at max a couple of hundred particles) I think the flat_set is a good choice, also of course because it is readily available. There are also (much) more efficient hash table implementations than the one in the STL, but that would require a library or implementing one
which I would defer to such time as to when this becomes a bottleneck :-) 

I did micro-benchmarks on this a long time ago (https://github.com/fweik/flat_map_bm), but not sure about the quality,
YMMV.

Also be aware that the link you posted compares to std::set, which is an ordered set implemented as a binary tree typically, while here the candidate is std::unordered_set, which is a hash table.

KaiSzuttor:
interestingly the mac test is broken but the gitlab CI is green

fweik:
Upon inspection, I think I slightly changed the behavior. Will check.

For the performance questions it would be nice to have a benchmark/typical scenario to consider, I don't have the domain knowledge here to come up with this.

KaiSzuttor:
@kosovan maybe you or one of your students is interested in checking the performance impact of this change

fweik:
I think its less about checking this particular change, but about having a benchmark script for a typical scenario as there exist e.g. for MD and electrostatics.

KaiSzuttor:
well, checking the impact of this change would imply providing such a scenario

fweik:
The macOS issue persists. I don't think it is worthwhile to investigate this further, so I'm going to abandon this. Sorry for the noise.

fweik:
@KaiSzuttor I think we can keep the cleanup, if you care to review...

KaiSzuttor (approved):

kosovan:
> @kosovan maybe you or one of your students is interested in checking the performance impact of this change

@pm-blanco says that he would be interested in testing it but it is not clear to us yet what is the expected impact.

KaiSzuttor:
> > @kosovan maybe you or one of your students is interested in checking the performance impact of this change
> 
> @pm-blanco says that he would be interested in testing it but it is not clear to us yet what is the expected impact.

sorry, @fweik decided to give up here since not all tests passed. the original idea was to change the container type of the particle type map to improve on cache hits during traversal by using the better memory layout of the flat set in the boost library.

jonaslandsgesell:
@fweik thanks for the answer
@KaiSzuttor was the merge intended?

fweik:
@jonaslandsgesell yes the map changed was dropped for now, due to problems on macos. Sorry for the confusion ^^

PR #4203, opened 2021-04-06
add ext_force_density to ekin

christophlohrmann:
Description of changes:
- Add interface for external forces on the fluid that ekin passes to lb
- test together with the lb poiseuille testcase


RudolfWeeber:
@itischler, could you review this one, since you are familiar with the EK code?

The way EK "remote-controls" LB is really odd, but as I understand it, that's the way it's currently implemented for all EK/LB features.

This will be different, when we switch to Walberla.

I foudn one issue: As I see it, you added EK to the list of required featues for the Poisseuille test. Please avoid that. The pure LB version should run independently of whether or  not EK is built in. In addition, the EK+LB-poiseuille test should run if EK is built in. I.e., you need two test funcitons `test_lb_poisseuille()` and `test_lb_ek_poisseuille()` and a third funciton, which will contain all the code shared between the two.

jngrad:
> I foudn one issue: As I see it, you added EK to the list of required featues for the Poisseuille test. Please avoid that. The pure LB version should run independently of whether or not EK is built in. In addition, the EK+LB-poiseuille test should run if EK is built in. I.e., you need two test funcitons `test_lb_poisseuille()` and `test_lb_ek_poisseuille()` and a third funciton, which will contain all the code shared between the two.

I think what he did is correct: a new test class was introduced that is skipped if Ekin is not compiled in. The pure LB still runs independently.

testsuite/python/lb_poiseuille.py
```diff
@@ -86,9 +95,7 @@ def prepare(self):
         self.system.lbboundaries.add(wall1)
         self.system.lbboundaries.add(wall2)
 
-        mid_indices = [int((self.system.box_l[0] / AGRID) / 2),
-                       int((self.system.box_l[1] / AGRID) / 2),
-                       int((self.system.box_l[2] / AGRID) / 2)]
+        mid_indices = [int(l / AGRID / 2) for l in self.system.box_l]
```
jngrad:
```suggestion
        mid_indices = (self.system.box_l / AGRID / 2).astype(int)
```

src/core/grid_based_algorithms/electrokinetics_cuda.cu
```diff
@@ -2241,49 +2241,39 @@ int ek_init() {
...
+        ek_parameters.lb_ext_force_density[2] != 0;
+    auto force_dens_conversion_factor = lbpar_gpu.agrid * lbpar_gpu.agrid *
+                                        lbpar_gpu.time_step *
+                                        lbpar_gpu.time_step;
+    lbpar_gpu.ext_force_density[0] =
+        ek_parameters.lb_ext_force_density[0] * force_dens_conversion_factor;
+    lbpar_gpu.ext_force_density[1] =
+        ek_parameters.lb_ext_force_density[1] * force_dens_conversion_factor;
+    lbpar_gpu.ext_force_density[2] =
+        ek_parameters.lb_ext_force_density[2] * force_dens_conversion_factor;
```
jngrad: with the help of a (safe) object slicing:
```suggestion
    lbpar_gpu.ext_force_density =
        Utils::Vector3f(ek_parameters.lb_ext_force_density) *
        Utils::sqr(lbpar_gpu.agrid * lbpar_gpu.time_step);
```

jngrad (commented):
minor readability suggestions

itischler (approved):
Seems good to me. 
Just noticed, that modification to external force after the method was activated, would not apply, since thy are not communicated to the gpu. However this is also true to ext forces applied to the species. So this is probably a task for an other pr.

jngrad (approved):
LGTM

KaiSzuttor:
> Seems good to me.
> Just noticed, that modification to external force after the method was activated, would not apply, since thy are not communicated to the gpu. However this is also true to ext forces applied to the species. So this is probably a task for an other pr.

please open a ticket

PR #4206, opened 2021-04-06
observables: make ids of ID based observable readonly.

KaiSzuttor:


jngrad (approved):

RudolfWeeber:
If we decide we want to prevent users from changing what an observables observe over time, I think it needs to be done everywhere by making all the parameters read-only and only settable at construction, just like with correlators.
This PR changes the behavior for PidObservables, but looking at the code in `src/script_interface/observables/ProfileObservable.hpp`, one can also change histogram dimensions after the fact.
E.g., one an write to `max_z` of a density profile.

KaiSzuttor:
yes, all of the observables should be immutable

PR #4211, opened 2021-04-08
observables: make observables immutable.

### LINKED ISSUE #4208: make observables immutable

KaiSzuttor:
Fixes #4208 

Description of changes:
- made observables immutable

ToDo:
- fix issue in python tests of cylindrical observables

### DIFF
--- a/src/core/observables/ProfileObservable.hpp
+++ b/src/core/observables/ProfileObservable.hpp
@@ -34,20 +34,20 @@ namespace Observables {
 
 /** Cartesian profile observable */
 class ProfileObservable : virtual public Observable {
-protected:
+private:
   /** Range of the profile edges. */
-  const std::array<std::pair<double, double>, 3> limits;
+  std::array<std::pair<double, double>, 3> m_limits;
   /** Number of bins for each coordinate. */
-  const std::array<size_t, 3> n_bins;
+  std::array<size_t, 3> m_n_bins;
 
 public:
   ProfileObservable(int n_x_bins, int n_y_bins, int n_z_bins, double min_x,
                     double max_x, double min_y, double max_y, double min_z,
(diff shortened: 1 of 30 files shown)


src/core/observables/PidObservable.hpp
```diff
@@ -51,7 +51,7 @@ using ParticleReferenceRange =
  */
 class PidObservable : virtual public Observable {
   /** Identifiers of particles measured by this observable */
-  std::vector<int> m_ids;
+  const std::vector<int> m_ids;
```
fweik: I think, as a general rule, const data members should be avoided.

KaiSzuttor: ok, changed that

src/core/observables/ProfileObservable.hpp
```diff
@@ -34,25 +34,37 @@ namespace Observables {
 
 /** Cartesian profile observable */
 class ProfileObservable : virtual public Observable {
+protected:
+  /** Range of the profile edges. */
+  const std::array<std::pair<double, double>, 3> limits;
```
fweik: Here ...

src/core/observables/ProfileObservable.hpp
```diff
@@ -34,25 +34,37 @@ namespace Observables {
...
+  /** Number of bins for each coordinate. */
+  const std::array<size_t, 3> n_bins;
```
fweik: ... too.

There reason is that this limits how the type can be used, which breaks the separation, but has no actual benefit. Also `protected` is usually no a great sign, design-wise. This is even a core guideline.

jngrad (dismissed):
LGTM

jngrad (approved):

PR #4226, opened 2021-04-13
Remove numerical particle ID from samples and user guide

### LINKED ISSUE #4029: Make particle id less prominent in system setup and particle doc

biermanncarl:
Fixes #4029

Description of changes:
- Remove numerical particle IDs from samples where possible
- Modify `particles.rst` and `inter_bonded.rst` in the user guide to prefer particle handles over numerical IDs

The following samples have been left out for now:
- `drude_bmimpf6.py` because `add_drude_particle_to_core` only accepts numerical IDs
- `lj-demo.py` because the sample appears to be broken


christophlohrmann:
From manual inspection: The visualiser samples work fine.

jngrad (approved):
Thanks!

PR #4228, opened 2021-04-14
make ElectrokineticsRoutines inherit from LBFluidRoutines

christophlohrmann:
Description of changes:
- ElectrokineticsRoutines so far was an incomplete copy of LBFluidRoutines with one extra property
- removed unneeded ekin code that reads quantities from GPU LB
- interface change: EK does not have ``.pressure`` anymore, instead ``.pressure_tensor``, which is what it actually returns
- bugfix: EK node access no longer accepts node indices as floating point values


jngrad (approved):
Thanks!

PR #4232, opened 2021-04-19
Unit test ESPResSo simulations

jngrad:
Description of changes:
- create an ESPResSo system manager to set up the MPI callbacks loop
- use `Utils::Vector3d` instead of raw pointers in particle callbacks
- unit test a typical MD simulation (LJ, FENE, P3M) with MPI

### DIFF
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -43,6 +43,7 @@ set(EspressoCore_SRC
     PartCfg.cpp
     AtomDecomposition.cpp
     reduce_observable_stat.cpp
+    EspressoSystemStandAlone.cpp
     DomainDecomposition.cpp)
 
 if(CUDA)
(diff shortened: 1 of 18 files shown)


jngrad:
Note: the new class `EspressoSystemStandAlone` is minimalist on purpose. It provides convenience methods to encapsulate the `global.hpp` logic for the most common properties of the system. Once we're done refactoring the `FIELD_*` globals as `on_*_change()` callbacks, the class methods can be removed and the constructor converted to a free function.

reinaual (approved):
I struggled a bit with the electrostatics threshold value but i don't have a better solution for it.

jngrad:
> I struggled a bit with the electrostatics threshold value but i don't have a better solution for it.

The k-space contribution the the P3M energy represents less than a percent of the real-space energy. The tolerance of 0.01% is just large enough for this case (2 particles). I looked into Madelung formula for crystal structures but couldn't find an analytical solution for the tetragonal lattice. If we have a need for a more precise P3M unit test in the future, we could look into building a NaCl crystal.

PR #4233, opened 2021-04-20
Cleanup non-bonded interaction kernels

jngrad:
Description of changes:
- merge kernel helper functions into their respective kernels if there are no other consumers in the core
- remove unused (and untested) `Utils::Vector3d` wrappers around force-factor-based kernels
- reduce code complexity and replace magic values by mathematical constants

### DIFF
--- a/src/core/nonbonded_interactions/bmhtf-nacl.hpp
+++ b/src/core/nonbonded_interactions/bmhtf-nacl.hpp
@@ -54,12 +54,6 @@ inline double BMHTF_pair_force_factor(IA_parameters const &ia_params,
   return 0.0;
 }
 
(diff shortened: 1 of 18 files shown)


reinaual (approved):

KaiSzuttor:
maybe you should add your opinion on the discussion about the non bonded interaction refactoring in https://github.com/espressomd/espresso/discussions/4214

jngrad:
> maybe you should add your opinion on the discussion about the non bonded interaction refactoring in #4214

That was the idea behind this PR, getting everything tidy before converting them to structs, like for bonded interactions. I looked everywhere in the list of issues to find the ticket, but it was in the discussions :-)

PR #4235, opened 2021-04-24
BoxGeometry members private

### LINKED ISSUE #4170: Make BoxGeometry members private
To allow storing values calculated from the box geometry which are guaranteed to be up to date.

E.g. inverse box length, which is used in minimum image convention in the hot loop depdning on the particle decomposition.

1. Replace the mechanism by which box size and periodicity are communicated. This is currently relies on direct read/write access to the data member (See #4133).
2. Make the data members in BoxGeometry private
3. Implement setters/getters

Note: Making the entire BoxGeometry class serializeable is probably not the solution, because algorithms like NPT change only the box size but not the periodicity. Also certain box changles, like isotropic ones, can be handled more efficiently than others.

reinaual:
Fixes #4170

Description of changes:
- privatized BoxGeometry members
- mpi-callbacks for periodicity and box_l
- extracted `on_periodicity_change` and `on_skin_change` from `on_parameter_change`
- added `skip_method_adaption` parameter to `on_boxl_change` to replace custom implementation in `velocity_verlet_npt`
- added python test for box_l and periodicity interface

### DIFF
--- a/src/core/BoxGeometry.hpp
+++ b/src/core/BoxGeometry.hpp
@@ -28,12 +28,13 @@
 #include <cmath>
 
 class BoxGeometry {
-public:
+private:
   /** Flags for all three dimensions whether pbc are applied (default). */
   std::bitset<3> m_periodic = 0b111;
   /** Side lengths of the box */
   Utils::Vector3d m_length = {1, 1, 1};
 
+public:
   /**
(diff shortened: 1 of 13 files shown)


jngrad (approved):
Thanks a lot!

PR #4237, opened 2021-04-26
Deploy documentation to GitHub pages

jngrad:
Deploy documentation in nightly builds. Write the espressomd/espresso commit SHA in the commit message for bookkeeping. When the last merge commit goes through nightly builds several days in a row, only deploy on the first night.


jngrad:
The deploy script works as expected. If the SHA of the current commit is different from the SHA written in the commit message of the GitHub Pages, overwrite the old documentation with a force push ([logfile](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/282063)). If the CI pipeline runs multiple times for the same commit SHA, then we don't need to deploy ([logfile](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/282068)). The verbosity level of the `rsync` and `git commit` have been changed to 0, because the first logfile was really long.

.gitlab-ci.yml
```diff
@@ -524,53 +527,47 @@ check_with_odd_no_of_processors:
...
+    - git add doc dox tutorials
+    - PREV_COMMIT_MSG=$(git log -1 --pretty=format:"%s")
+    - case "${PREV_COMMIT_MSG}" in "Documentation for "*) git reset --soft HEAD^; esac
+    - git commit --quiet -m "${NEXT_COMMIT}"
+    - git push -f origin main
+  interruptible: false
+  tags:
+    - docker
+    - linux
+    - icp
```
KaiSzuttor: I think we should start splitting our gitlab ci yaml files. This file get's too crowded.

jngrad:
Deployment now takes place via GitHub Actions. You can see the logfile in [pipeline 2450109192](https://github.com/jngrad/espresso/runs/2450109192?check_suite_focus=true).

The GitHub Pages at [espressomd/espressomd.github.io](https://github.com/espressomd/espressomd.github.io) now has 3 branches. The `landing_page` contains the logic to generate the HTML files from the Markdown files. The `releases` branch contains the user guide and tutorial of old releases. When a new version of ESPResSo is released, add the generated files to `releases` and update the release version in the metadata file in `landing_page`. The `main` branch contains the generated documentation and is force pushed at every new merge commit.

The tutorials summary in `doc/tutorials/Readme.md` is now used to generate the HTML file, this way we don't need to touch the GitHub Pages when tutorials are created or deleted.

.github/actions/deploy_docs/action.yml
```diff
@@ -0,0 +1,46 @@
...
+description: 'Deploy ESPResSo documentation'
+runs:
+  using: "composite"
+  steps:
+    - run: |
+        # checkout github pages
+        cd "${HOME}"
+        git clone --quiet git@github.com:espressomd/espressomd.github.io.git
+        cd espressomd.github.io
+        LAST_COMMIT=$(git log -1 --pretty=format:"%s" remotes/origin/main -- doc dox tutorials)
```
RudolfWeeber: Please add  a comment explaining which commit is or isn't selected, here

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,24 @@
+name: deploy
+
+on:
+  push:
+  pull_request:
```
RudolfWeeber: Why is it supposed to run on pull_request?

.github/actions/deploy_docs/action.yml
```diff
@@ -0,0 +1,46 @@
...
+        if grep -F '<!DOCTYPE html>' *.zip; then
+          echo "The artifacts could not be downloaded.";
+          exit 1;
+        fi
+        unzip -q '*.zip'
+        # create a fresh main branch
+        git config --global user.email "noreply@icp.uni-stuttgart.de"
+        git config --global user.name "espresso-ci"
+        git checkout -b new_main remotes/origin/releases
+        # generate the landing page
```
RudolfWeeber: Please add a comment mentioning the branch structure of the github pages repo and how the final page is constructed.

.github/actions/deploy_docs/action.yml
```diff
@@ -0,0 +1,46 @@
...
+          echo "Documentation already up-to-date.";
+          exit 0;
+        fi
+        # download artifacts
+        curl https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/artifacts/python/download?job=run_tutorials -L -o tutorials.zip
+        curl https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/artifacts/python/download?job=run_doxygen   -L -o doxygen.zip
+        curl https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/artifacts/python/download?job=check_sphinx  -L -o sphinx.zip
+        if grep -F '<!DOCTYPE html>' *.zip; then
+          echo "The artifacts could not be downloaded.";
+          exit 1;
```
RudolfWeeber, line 25: Are you sure that's enough to verify that all 3 downloads worked?

RudolfWeeber (changes requested):

.github/actions/deploy_docs/action.yml
jngrad: After further reflection, we don't actually need to filter files here, because the commit we're looking for is always the most recent one. I'll remove the superfluous `-- doc dox tutorials`.

.github/workflows/deploy.yml
jngrad: Good point, we only need it on `schedule`.

.github/actions/deploy_docs/action.yml
jngrad: We can add the `curl` flag `--fail`, which prints out the HTTP error code in case of failure. The `curl` man pages say it's not working on all HTTP error codes, but I tried a few common ones (typo in url and in GET query) and it printed out the expected 404 error code instead of downloading the fancy 404 HTML page, which is helpful for debugging.

RudolfWeeber (approved):

PR #4261, opened 2021-05-27
Refactor globals.cpp:bcast_parameter

### LINKED ISSUE #4133: Refactor globals.cpp:bcast_parameter can of worms...globals
This relies on gloals.cpp::fields which maps flags to memory localtion of globals.
Ghey are then sent via mpi without serialization.
Afterwards some logic in events.cpp is used to do follow-up changes that aris from the parameter broadcast.

All of this needs to be replaced:
* For each of the objects (box_l, periodicity, skin, ...) an mpi callback pair needs tobe created that does the data transfer and handles the follow up logic on the receiving end.

I stumbled up on this, because I wanted to make m_length and m_periodic in BoxGeometry private (as they should be). This would have allowed to pre-cauclate half_box_l and box_l_inv for more efficient get_mi_vector in bond-heavy and Lees-Edwards systems.

reinaual:
Fixes #4133

- created MPI-Callbacks for each variable in the old `global.hpp:field`
- removed `globals.hpp` file
- removed the `kT=None, gamma=None` default logic in thermostats.pyx


src/core/global.hpp
```diff
@@ -1,88 +0,0 @@
-/*
```
jngrad, line 1: With this file gone, we can finally remove external linkage of time-related global variables: f78df9f8b7bacf66e315b40ffdc7efc3fafb67d4

src/core/thermostat.cpp
```diff
@@ -144,3 +145,124 @@ void philox_counter_increment() {
...
 }
+
+#ifndef PARTICLE_ANISOTROPY
+void mpi_set_brownian_gamma_local(double gamma) {
+  brownian.gamma = GammaType{gamma};
+}
+
+void mpi_set_brownian_gamma_rot_local(double gamma) {
+  brownian.gamma_rotation = GammaType{gamma};
+}
```
jngrad, line 155: CI didn't pass the coverage checks because a few callbacks like this one are never executed with the config files available in CI. This can be solved by removing duplicated callbacks: e6043ef6d3a123a1266a96bed133bf5e28302813

jngrad (changes requested):
Besides suggested changes ([jngrad/espresso:`globals_refactor_bcase`](https://github.com/jngrad/espresso/tree/globals_refactor_bcase)), looks good to me.

src/core/thermostat.cpp
jngrad: Fixed.

src/core/global.hpp
jngrad: This will be addressed in another PR, many more globals can now be made static.

jngrad (approved):

PR #4262, opened 2021-05-27
Update the constant-pH tutorial

### LINKED ISSUE #3941: Issue tracker for Tutorial 12
Regarding the version with WCA active and P3M active:

* it takes 1 hour to run the tutorial
* the tutorial only says "Observe how the ionization equilibrium is affected by various interactions" without interpreting the plot
   * the sigmoid is shifted to higher pH values, i.e. the degree of dissociation is smaller at pH > pKa, probably because once several monomers are basic, protons lost by neighboring monomers have a higher chance of being recaptured by basic monomers due to the Coulomb interaction
   * it's difficult to assess if the resulting plots are correct; are analytical formula available?
* the test for the WCA+P3M case doesn't check numerical results

kosovan:
Fixes #3941

Description of changes:
- add prior knowledge, make equations and units more consistent
- add a flag to run electrostatics with Debye-Hückel instead of P3M
- accelerate tutorial with DH, cell system, non-interacting particle type
- fix ambiguous explanations


kosovan:
Some interesting observations on speedup:
* `RE.set_non_interacting_type(len(TYPES))` speeds up the calculation of interacting system roughly by a factor of 2.0.
*  Replacing P3M with DH makes the calculation with electrostatics almost as fast as without
*  If we do reaction in every cycle and integration just in some cycles, we save the amount of runtime roughly proportional to `PROB_INTEGRATION`. In the previous version we were doing the opposite, so we were wasting some 50% of the reaction samples.

If the non_interacting_type has such an effect on the speed, I think it is desired to give it a fixed and low value, or to set dynamically to the lowest possible value that is available.

Maybe we should consider using some of these tricks also in the tests of the RxMC and cpH algorithms. It is not important to get perfectly correct physics but to ensure that we can still reproduce the titration curve of an interacting system if we change something in the algorithm. What do you think of that? @jngrad @RudolfWeeber

jngrad:
> Maybe we should consider using some of these tricks also in the tests of the RxMC and cpH algorithms.

Good idea, I can have a look at this later this week to speed up the testsuite, and mention it in the docs too.

A couple of tests did not pass CI. Let us know when you are ready with the PR and we'll review it.

doc/tutorials/constant_pH/constant_pH.ipynb
```diff
@@ -4,7 +4,20 @@
    "cell_type": "markdown",
```
davidbbeyer, line 4: <p>In the first paragraph I would add "particle size \sigma <strong>in nm </strong>..."</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/espressomd/espresso/pull/4262/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='11'/>

davidbbeyer, line 4: <p>I think we should change the beginning of some of the sentences, it is not so nice to have successive sentences starting with "If" or "We".</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/espressomd/espresso/pull/4262/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='13'/>

davidbbeyer, line 4: <p>There is a typo here: "tne number ..." instead of "the number ..."</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/espressomd/espresso/pull/4262/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='19'/>

davidbbeyer, line 4: <p>I think it would be better to write this as a list like this:</p><ul><li>time-step ...</li><li>skin ...</li><li>seed ...</li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/espressomd/espresso/pull/4262/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='25'/>

kosovan: @davidbbeyer I have not yet cleaned up the text. Modifications to the python code are mostly done but I only did those changes in the text that were directly related to changes in the code.

kosovan: <p>@davidbeyer: I do not understand, where would you like to add "particle size \sigma <strong>in nm"</strong> and why? In the current implementation, you can set particle size in any unit of length and pint takes care of converting it.</p>

kosovan: <p>OK, I can rephrase it. However, I do not see it as an important issue.</p>

davidbbeyer: <p>I'm okay with leaving it out, but then you should also remove the reference to Kelvin ("We start by setting the desired temperature&nbsp;T in Kelvin") since it is an arbitrary unit as well.</p>

kosovan: <p>OK, now I see the point. Right, we should remove the Kelvin from there.</p>

kosovan:
TODO: before this PR is merged, we should change the initial set of parameters back to the ideal system and clear the cells with solutions of the excercises.

jngrad (approved):
Thanks for the improvements!

PR #4264, opened 2021-05-27
Pass time step and temperature as function arguments

### LINKED ISSUE #4247: Add time step and kT as parameters to components that use them
In preparation for the integrator refactor (#4015):
In order to remove time_step (#3769) and temperature as globals, components that use them should take them as an initialization argument and store their own copy.

This affects e.g.: 
* dpd (time step and temperature)
* thermalized bonds (tim step and temperature)
* lb (temperature, time step because of unit conversion)
* reaciton ensemble (temperature)
* electrostatics module (no idea, wy. Perhaps Drude)

The best way to identify uses is probably, using and IDE of your choice,  to open the headers which define the globals (integrate.hpp and thermostat.hpp), find the definition of the global and use the "find uses" functio of the IDE.

jngrad:
Fixes #4247, partial fix for #2628

Description of changes:
- reduce usage of the `temperature` and `time_step` globals
- combine the `int fluidstep` and `double fluidstep` static globals into a single static global
- only increment LB RNG during LB integration (changes the thermalization random sequence when `tau != time_step`)
- reduce code duplication, replace raw pointers by references, improve separation of concerns

### DIFF
--- a/src/core/grid_based_algorithms/lbgpu.cpp
+++ b/src/core/grid_based_algorithms/lbgpu.cpp
@@ -64,8 +64,6 @@ LB_parameters_gpu lbpar_gpu = {
     -1.0,
     // tau
     -1.0,
-    // time_step
-    0.0,
     // dim_x;
     0,
     // dim_y;
@@ -102,9 +100,9 @@ bool ek_initialized = false;
 /*-----------------------------------------------------------*/
 
 /** %Lattice Boltzmann update gpu called from integrate.cpp */
-void lattice_boltzmann_update_gpu() {
+void lattice_boltzmann_update_gpu(double time_step) {
 
-  auto factor = (int)round(lbpar_gpu.tau / time_step);
+  auto factor = static_cast<int>(round(lbpar_gpu.tau / time_step));
 
   fluidstep += 1;
 
@@ -130,7 +128,6 @@ void lb_reinit_fluid_gpu() {
  *  See @cite dunweg07a and @cite dhumieres09a.
  */
 void lb_reinit_parameters_gpu() {
-  lbpar_gpu.time_step = static_cast<float>(time_step);
   lbpar_gpu.mu = 0.0;
 
   if (lbpar_gpu.viscosity > 0.0 && lbpar_gpu.agrid > 0.0 &&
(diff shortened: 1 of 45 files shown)


src/core/grid_based_algorithms/lbgpu.cpp
```diff
@@ -102,9 +100,9 @@ bool ek_initialized = false;
 /*-----------------------------------------------------------*/
 
 /** %Lattice Boltzmann update gpu called from integrate.cpp */
-void lattice_boltzmann_update_gpu() {
+void lattice_boltzmann_update_gpu(double time_step) {
 
-  auto factor = (int)round(lbpar_gpu.tau / time_step);
+  auto factor = static_cast<int>(round(lbpar_gpu.tau / time_step));
```
KaiSzuttor: the decision to update the LB should be made in the integration loop and not within the LB. Then the time step is not needed here

src/core/grid_based_algorithms/lbgpu_cuda.cu
```diff
@@ -1366,23 +1366,23 @@ __device__ void calc_viscous_force(
       (1 - flag_cs_float) * viscforce_density.z;
 
   // only add to particle_force for particle centre <=> (1-flag_cs) = 1
-  delta_j[0] -= ((1 - flag_cs_float) * viscforce_density.x) * para->time_step *
+  delta_j[0] -= ((1 - flag_cs_float) * viscforce_density.x) * time_step *
```
KaiSzuttor, line 1369: i'm not sure if this is correct, the LB should not depend on the time step. it should only have it's own unit of time, namely tau

KaiSzuttor: maybe @RudolfWeeber can comment on this

RudolfWeeber:
I think it’s unit conversion. dt^2/agrid is the conversion factor for a force, I guess.

The particle force is stored in MD units on the GPU, so the conversion from LB units has to happen somewhere. But the details of LBGPU-uints are a mystery to me.

 

This will only get disentangled with the switch to Walberla, I guess.

KaiSzuttor:
> I think it’s unit conversion. dt^2/agrid is the conversion factor for a force, I guess. The particle force is stored in MD units on the GPU, so the conversion from LB units has to happen somewhere. But the details of LBGPU-uints are a mystery to me. This will only get disentangled with the switch to Walberla, I guess.

ah I thought that I once had removed almost all MD units. But probably only from the CPU version.

src/core/grid_based_algorithms/lbgpu.cpp
jngrad: Thanks! This is now fixed.

KaiSzuttor (approved):

Issue #4251, opened 2021-05-14
Reaction method refactoring wish list


RudolfWeeber:
actoring wish list (Disclaimer: I'm not terribly familiar with that part of the code)

This is a WIP

## ReactionAlgorithm::make_reaction_attempt()

* [x] Are the non const vector arguments actually used only as outputs? If so, they should be returned rather than being arguments, to make the dependencies clear.

## ReactionAlgorithm::generic_oneway_reaction()
* [x] This should take a reference to a SingleReaction to decouple the reaction from the storage of the reactions.
* [x] Looks like some Wang Landau stuff is mixed in
* [x] The "exponentials" accumulating accumulates something else than the Boltzmann factor used in the acceptance probability. Is that intended?
* [ ] The particle (property) hiding/deleting/restoring is not clear to me. It needs to be explained somewhere, or if it is, the explanation needs to be referenced. Why is hiding preferred to outright deletion?
* [x] Rather than calling the `on_.*()` methods, the function should communicate the outcome via return value (a typed enum). This would allow to avoid mixing in code for a special case (Wang Landau) in the implementation of the general case. I guess, outcomes are: early rejection, rejection, acceptance.

## ReactionAlgorithm::delete_particle() and ReactionAlgorithm::create_particle(int)
* [ ] The functionality of managing the max seen particle counter and obtaining an unused particle id for particle creation should be pulled out of the reaction ensemble. Other algorithms need that too (and have their own private implementations...)
* [x] The expensive `partCfg()`and `distto()` stuff can be skipped if the exclusion radius is 0 (several instances throughout the code)

## general
* [ ] The task of finding/checking for a close-by particle (currently done with `distto()`) can be accelerated considerably using the domain decomposition cell system. O(1) vs O(N)

## ReactionAlgorithm::do_global_mc_move_for_particles_of_type()
* [x] The task of finding ids and positions of particles to move should be outsourced to a function returning the two vectors. The logic can probably be simplified. Starting L563. Then, it could be tested independently.
* [x] rather than immediately moving the particles, generate all new positions/velocities and store in an array. When setting positions, the particle data caches are invalidated. Or even better: Collect the particle masses when collecting the old positions and ids. then, no additional `get_particle_data()` is needed at all.
* [x] there is an other instance of the expensive `distto()` logic for the exclusion radius
* [x] Creating a dummy reaction and then using its acceptance probability method is an indication that the acceptance probability should be a free function.

## ConstantpHEnsemble::do_reaction()
[...]

PR #4265, opened 2021-05-27
Refactor reaction methods

pm-blanco:
Fixes minor issues of #4251
Today we have been working with @davidbbeyer  on some of the minor issues of  #4251.

Description of changes:
- `ReactionAlgorithm::make_reaction_attempt()` now returns vectors instead of relying on output parameters.
- `ReactionAlgorithm::generic_oneway_reaction()` takes now as an argument a `SingleReaction` variable. We have done the same change in `ReactionAlgorithm::save_old_particle_numbers`, `ReactionAlgorithm::all_reactant_particles_exist()` and `WidomInsertion::measure_excess_chemical_potential()`.
- We improved separation of concerns for particle creation/deletion/move in `ReactionAlgorithm` and reduced code duplication.


jngrad:
@pm-blanco Could you please resolve the merge conflicts?

pm-blanco:
@jngrad done.

jngrad (approved):
Thanks @davidbbeyer and @pm-blanco!

src/core/reaction_methods/ReactionAlgorithm.hpp
```diff
@@ -109,11 +111,11 @@ class ReactionAlgorithm {
   virtual void on_mc_reject(int &old_state_index){};
   virtual int on_mc_use_WL_get_new_state() { return -10; }
 
-  void make_reaction_attempt(
-      SingleReaction const &current_reaction,
-      std::vector<StoredParticleProperty> &changed_particles_properties,
-      std::vector<int> &p_ids_created_particles,
-      std::vector<StoredParticleProperty> &hidden_particles_properties);
+  std::tuple<std::vector<StoredParticleProperty>, std::vector<int>,
```
jonaslandsgesell, line 114: @jngrad , @KaiSzuttor  , I found an interesting comment on whether to use custom structs or tuples https://google.github.io/styleguide/cppguide.html#Structs_vs._Tuples linked in https://abseil.io/tips/176
Would it have been better to go for the solution implemented here: https://github.com/espressomd/espresso/pull/4265/commits/7b5bb27d8bc20cb25b43891dc70061a1e2849d65 ?

jngrad: The drawback of `auto` as a return type of a non-inline method is that the return type is only known in the translation unit where the method is defined. If you try to call the method from a child class, you will get
```c++
struct A {
  auto foo();
};
struct B: public A {
  auto bar() { return foo(); }
};
```
```
$> g++ -std=c++14 -c src.hpp
src.hpp: In member function ‘auto B::bar()’:
src.hpp:5:27: error: use of ‘auto A::foo()’ before deduction of ‘auto’
   auto bar() { return foo(); };
```
This means you will need to bring the struct declaration from the method scope to the class scope as soon as you find yourself in a situation where you need to call the method from outside the translation unit where it is defined, for example when writing a unit test.

I'm not totally convinced wrapping the return values in a struct to provide a namespace is helpful either. It also locks us with the names in the struct: if we later decide to rename any of the struct members, we have to update all call sites.

Personally, I find structured bindings sufficiently expressive:
```c++
#include<tuple>
#include<vector>
#include<iostream>

std::tuple<std::vector<int>, std::vector<float>, std::vector<double>> foo();

int main() {
  auto [p_ids_created, hidden_particles, changed_particles] = foo(); // C++17 feature
  std::cout << p_ids_created[0] << " "
            << hidden_particles[0] << " "
            << changed_particles[0] << "\n";
}
```
In this PR the `std::tie(p_ids_created, hidden_particles, changed_particles) = foo()` construction had to be used, because we're stuck with C++14 for hardware reasons. This has the unfortunate consequence that we need to spell out the types of the bound variables before the call to `std::tie()`.

In another place in the PR, you see the following construction:

```c++
#include<array>
#include<vector>
#include<utility>

using Vector3d = std::array<double, 3>;

void place_particle(int pid, Vector3d const &new_pos);

void generate_new_particle_positions(int n_part) {
  using ParticleProperties = std::pair<int, Vector3d>;
  std::vector<ParticleProperties> cache;
  cache.reserve(n_part);
  for (int i = 0; i < n_part; ++pid) {
    auto const pid = i_random(/*...*/);
    auto const new_pos = /*...*/;
    cache.emplace_back(ParticleProperties{pid, new_pos});
  }
  /* ... */
  // restore original particle positions
  for (auto const & item : cache)
    place_particle(std::get<0>(item), std::get<1>(item));
}
```
Of course, I meant to forward the pair directly to the `place_particle()` function, but it didn't compile with C++14:
```c++
  // restore original particle positions
  for (auto const &item : cache)
    std::apply(place_particle, item); // C++17 feature
```

jonaslandsgesell: Thanks for your thoughts! I will think about it a bit more :)

PR #4268, opened 2021-05-28
Implement == for cdef class LBNode and LBSlice #4204

### LINKED ISSUE #4204: Implement == for cdef class LBNode and LBSlice
Nodes and slcies shoudl be considered equal, if the index/slice is equal﻿

Otherwise 
```
lb_nodes1 = [some list of lb node objects retrieved by index]
lb_nodes2 = [some list of lb nodes overlapping with lb_nodes1 also retrieved by index]
all_nodes =set(lb_nodes1+lbnodes2)
```
can have duplicate entries

This would be solved by implementing `__eq__()`.

Particle handles might also be affected.

stekajack:
Resolves #4204

Description of changes:

- implemented __eq__() for LBFluidRoutines() . Now logical lbf[index, index, index] == lbf[index, index, index] give expected results.

- modified default __hash__() so that the comparisons are done based on node indices rather than the name attribute. This is done because hash will give different names to pointers so 2 different handles to the same node would be considered not equal. Now they are considered equal so that set(list_of_nodes, also_list_of_nodes) is possible and cannot have duplicate entries now.

- implemented __iter__() for LBSlice class. now given a slice object, one can directly access nodes in any iteration context (using generator functionality for this). Therefore LBSlice object can now be transformed in to for example a list of nodes. This enables smth like set(list(LBSlice), list_of_nodes) to work correctly. 

- However set(LBSice, list_of_nodes) will NOT work!  This is deliberate because LBSlice as implemented, cannot in general be used as a node and as such it made more sense to first ask for an explicit conversion to a iterable_of_nodes before doing logical comparisons.


RudolfWeeber:
Thanks. The code looks good. Could you please also add test functions in testsuite/python/lb.py testing the new methosd:
For the iteration, you could, e.g., compare the index attribute from the slice directly and from the loop variable of an iteration

stekajack:
> Thanks. The code looks good. Could you please also add test functions in testsuite/python/lb.py testing the new methosd:
> For the iteration, you could, e.g., compare the index attribute from the slice directly and from the loop variable of an iteration

I hope this encompasses well what you had in mind. In lb_slice.py test I moved the lb_fluid handle outside of the method, so now test methods access the fluid as a class attribute, so I don't have to teardown.

RudolfWeeber (approved):

RudolfWeeber:
Yes, that's what I had in mind. Thanks.

PR #4273, opened 2021-06-01
Demonstrate constant pH method with more complex reactions

### LINKED ISSUE #4251: Reaction method refactoring wish list

jonaslandsgesell:
With this PR I demonstrate a generalization of the constant pH method in Espresso (see also the question by @jngrad in https://github.com/espressomd/espresso/issues/4251#issuecomment-851641704 ). 

The change can be summarized in one sentence: the constant pH ensemble does exactly the same as the reaction ensemble but the number of protons N_H appearing in the acceptance probability is not inferred from counting protons in the box (like in the reaction ensemble) but from computing 10**-pH mol/l *volume, see src/core/reaction_methods/ConstantpHEnsemble.cpp https://github.com/espressomd/espresso/pull/4273/files#diff-31488bc3f9a69b6dd9b0349f71bdfba05fb8a1d20d4c640d837fd15bd43eb7feR50

partially Fixes https://github.com/espressomd/espresso/issues/4251#issuecomment-849888239

Description of changes: allow more complex reactions in the constant pH method using the rule cH^nu_H = 10**(-pH*nu_H) (which neglects any mean activity coefficient). If you want to follow the reasoning, please compare the reaction ensemble acceptance criterion (eq.9) with eq. 17 in https://arxiv.org/pdf/1702.04911.pdf
Please note that one needs to track *nu_H* which is done via the new parameter called "product_index_protons" which is the index of the proton type in the product list.

Also a new required parameter "conversion_factor_from_mol_per_l_to_1_div_sigma_cubed" was introduced needed for converting cH=10**-pH mol/l to simulation units.

The more complex example reaction which is implemented is 
2 A + 3B <-> 4 C + 1 H + 3E (see https://github.com/espressomd/espresso/pull/4273/files#diff-4e17e965e0a44a038dae3f6c1efcbaf0dae92c39e416d42eb11e9ef25a148aa3R88 )


samples/constant_pH_ensemble_complex_reaction.py
```diff
@@ -0,0 +1,168 @@
...
+        system.part.add(pos=np.random.random(3) * system.box_l, type=type_A)
+    for j in range(abs(nu_B)):
+        system.part.add(pos=np.random.random(3) * system.box_l, type=type_B)
+
+# use an exclusion radius of 0 to simulate an ideal gas
+RE = reaction_ensemble.ConstantpHEnsemble(
+    temperature=1, exclusion_radius=0, seed=4, conversion_factor_from_mol_per_l_to_1_div_sigma_cubed=1.0/37.1)
+
+RE.add_reaction(
+    gamma=K*(c_ref_in_1_div_sigma_cubed)**nu_bar,
```
jonaslandsgesell: Since conversion_factor_from_mol_per_l_to_1_div_sigma_cubed is now passed to the reaction method, one could do this conversion internally and only provide the dimensionless equilibrium constant K = prod a_i**nu_i to the reaction method. This would probably simplify the useage of the reaction methods.

samples/constant_pH_ensemble_complex_reaction.py
```diff
@@ -0,0 +1,168 @@
...
+# setup N0 pairs of reactants
+for i in range(N0):
+    for j in range(abs(nu_A)):
+        system.part.add(pos=np.random.random(3) * system.box_l, type=type_A)
+    for j in range(abs(nu_B)):
+        system.part.add(pos=np.random.random(3) * system.box_l, type=type_B)
+
+# use an exclusion radius of 0 to simulate an ideal gas
+RE = reaction_ensemble.ConstantpHEnsemble(
+    temperature=1, exclusion_radius=0, seed=4, conversion_factor_from_mol_per_l_to_1_div_sigma_cubed=1.0/37.1)
```
jonaslandsgesell: Maybe pH should be a mandatory argument for the ConstantpHEnsemble constructor and the getters and setters should be used only when changing the initial pH value. Right now it is possible to create a ConstantpHEnsemble with undefined pH and running a simulation with undefined pH.

samples/constant_pH_ensemble_complex_reaction.py
jonaslandsgesell: I implemented the changes, gamma which is passed to the reaction methods is now the dimenstionless reaction constant.
The discussion about gamma in https://espressomd.github.io/doc/advanced_methods.html#reaction-ensemble needs to be adapted.

kosovan:
I think that I was already posting a comment on this PR but I do not see it here. Perhaps I forgot to submit it.

I disagree with the statement that it partially Fixes #4251 (comment) This comment remains valid: that constant-pH algorithm as it has been introduced in the literature, works only with one possible set of values of stoichiometric coefficients, so it is pointless to ask the user to provide these numbers. Instead of removing the redundancy, this generalization introduces extra complexity in the algorithm, that is not needed and not properly documented. 

This generalization is perhaps algorithmically feasible, but I find several problematic issues:
1. Most important: there is no literature reference to this generalization, no white paper where the correct acceptance probabilities would be derived, and against which the implementation could be checked.  So, the only reference and documentation of the acceptance probabilities is the code itself. I do not dispute that such generalization is possible but the reference mentioned in the description of this PR ( https://arxiv.org/pdf/1702.04911.pdf ) does not provide a formula that could be directly implemented. 
2. We do not have a clear analytical result that could be used as a reference for a non-trivial test case
3. I do not see a clear use case for such a generalization
4.There is probably no user who would like to use it at the moment.
5. This suggestion comes in the middle of the work on refactoring the reaction ensemble code. Since it is not needed, it should be postponed and implemented (if needed) in the refactored code.

jonaslandsgesell:
Hello Peter,

I see it differently.

1. The change in this PR  can be summarized in one sentence: the constant pH ensemble does exactly the same as the reaction ensemble but the number of protons N_H appearing in the acceptance probability is not inferred from counting protons in the box (like in the reaction ensemble) but from computing  N_H =10**-pH mol/l *volume. 
This is obviously equivalent to the constant pH method with the reaction HA <-> A + H (compare the acceptance probabilities in the  Arxiv paper). This generalizes for other reactions. If this was never published before, then it could be published.

2. One can easily obtain analytical results for non-interacting systems. A numerical check using scipy.opzimize.fsolve is done in the sample and the accompanying test case (which is automatically executed) samples/constant_pH_ensemble_complex_reaction.py which is exactly the corresponding reaction ensemble test samples/reaction_ensemble_complex_reaction.py but with the replacement N_H =10**-pH mol/l *volume (and a slightly different reaction)
I do not see a problem with this way of testing. Analytical results would just mean solving the equations in samples/constant_pH_ensemble_complex_reaction.py symbolically instead of solving them numerically with a well tested library. I do not see benefit in doing so for the test.

3. I see potential use cases. Possible examples of reactions involving multiple protons are reactions in acidic environments. For example:
https://www.chemteam.info/Redox/Balance-Redox-Acid.html
You might argue those are not relevant because, maybe, you can represent a big reaction involving e.g. 8 H+ as a chain of protonation steps. But even if this was possible, you should consider that it can be beyond the scope of a simple model to introduce all these single protonation steps. A simple model would not model potential intermediate protonation steps, while sacrificing modeling depth/accuracy. Deciding on the importance of model accuracy (with all the potential intermediate protonation steps) and model simplicity (only having one reaction to deal with) cannot be done in general - also not in this PR which is why it would be good to  provide modeling freedom in the constant pH method (which is done with this PR). A model with only one reaction might give the needed results already leaving out not needed intermediate steps which could also help in computational efficiency.

4. I am not aware of an open (WIP) PR on refactoring parts of the code which I touched. In any case the merge conflicts should be easy to resolve based on the simplicity of the change (see point 1)

jngrad:
The core team decided to not move forward with this PR due to the added maintenance costs. It is also unclear, how many users would benefit from the feature. If a research group is interested in exploring this extension of the constant pH method and is willing to maintain the associated code, testsuite and samples, please contact us here or on the mailing list. Closing for now.

PR #4281, opened 2021-06-15
Fix ObjectList checkpointing regression

### LINKED ISSUE #4280: Checkpointing doesn't support recursion with more than 1 MPI rank
In ESPResSo 4.2-dev, a regression from #4167 broke the checkpointing mechanism. When reloading from a checkpoint file, script interface objects with global creation policy are not broadcasted, i.e. they are only restored on rank 0. This affects ~63 python classes. 

The checkpointing tests don't check that these objects are restored on all MPI ranks. This could be achieved by e.g. introducing a new particle with type 10 on a cell belonging to MPI rank 1 or rank 2 that only interacts with a shape-based constraint. Then we load the system from the checkpoint, recalculate the force on that particle and check that it is non-zero.

MWE:

```python
# mwe_save.py
import os
import espressomd.shapes

checkpoint = espressomd.checkpointing.Checkpoint(checkpoint_id='mwe', checkpoint_path=".")
if checkpoint.has_checkpoints(): # cleanup old checkpoint files
    for filepath in os.listdir(checkpoint.checkpoint_dir):
        os.remove(os.path.join(checkpoint.checkpoint_dir, filepath))

system = espressomd.System(box_l=3 * [10.0], time_step = 0.01)
system.part.add(pos=[system.box_l / 2. - 1., system.box_l / 2. + 1.], type=[0, 0])
system.constraints.add(shape=espressomd.shapes.Sphere(center=system.box_l / 2., radius=1), particle_type=1)
system.non_bonded_inter[1, 0].lennard_jones.set_params(epsilon=1., sigma=1., cutoff=3.0, shift=0.1)
checkpoint.register("system")
checkpoint.save(0)

system.integrator.run(0)
print(system.part[:].f)
```

```python
# mwe_load.py

checkpoint = espressomd.checkpointing.Checkpoint(checkpoint_id='mwe', checkpoint_path=".")
checkpoint.load(0)

system.integrator.run(0)
print(system.part[:].f)
```

Output:

```
$> mpiexec -n 2 ./pypresso mwe_save.py
[[-1475.26213767 -1475.26213767 -1475.26213767]
 [ 1475.26213767  1475.26213767  1475.26213767]]
$> mpiexec -n 2 ./pypresso mwe_load.py
[[-1475.26213767 -1475.26213767 -1475.26213767]
 [    0.             0.             0.        ]]
```

jngrad:
Fixes #4280

Description of changes:
- deserialize objects contained in `ObjectList` containers on all MPI ranks if the containers have global creation policy
- update checkpointing tests to check that multiple MPI nodes agree on objects contained in `ObjectList` containers

### DIFF
--- a/src/core/accumulators.cpp
+++ b/src/core/accumulators.cpp
@@ -62,9 +62,17 @@ int auto_update_next_update() {
 
 void auto_update_add(AccumulatorBase *acc) {
   assert(acc);
+  assert(std::find_if(auto_update_accumulators.begin(),
+                      auto_update_accumulators.end(), [acc](auto const &a) {
+                        return a.acc == acc;
+                      }) == auto_update_accumulators.end());
   auto_update_accumulators.emplace_back(acc);
 }
 void auto_update_remove(AccumulatorBase *acc) {
+  assert(std::find_if(auto_update_accumulators.begin(),
+                      auto_update_accumulators.end(), [acc](auto const &a) {
+                        return a.acc == acc;
+                      }) != auto_update_accumulators.end());
   auto_update_accumulators.erase(
       boost::remove_if(
           auto_update_accumulators,
--- a/src/core/constraints/Constraints.hpp
+++ b/src/core/constraints/Constraints.hpp
@@ -23,6 +23,7 @@
 #include "grid.hpp"
 #include "statistics.hpp"
 
+#include <algorithm>
 #include <memory>
 #include <vector>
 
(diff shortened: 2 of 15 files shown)


RudolfWeeber:
I just took a look. The approach you took looks correct to me.

Would it maybe helpful to ad test code to the script interface that compares the content of
Globalcontext::m_local_objects? The ids present should agree directly. For the objects, it should at least be possible to compare name, the vairant map with the parameters, and maybe the internal state string.
This would be a more direct test than what we do from Python.

RudolfWeeber:
It might also be useful to printf the object ids of the objects in ObjectList::remove() and ::add() on the different nodes.

Issue #4278, opened 2021-06-09
system.part does not behave like an array


itischler:
choosing a random particle via ```np.random.choice``` will fail, if there is only one particle in the system and if this particle has not the id of 0

```
import espressomd
import numpy as np

sys = espressomd.System(box_l=[10,10,10])
sys.part.add(pos=[4,4,4])
for i in range(10):
  print(np.random.choice(sys.part).id) ### works fine

sys.part.add(pos=[5,5,5])
sys.part.add(pos=[6,6,6])

for i in range(10):
  print(np.random.choice(sys.part).id) ### works fine

sys.part[0].remove()

for i in range(10):
  print(np.random.choice(sys.part).id) ### still works fine

sys.part[2].remove()

print(np.random.choice(sys.part).id) ### fails but particle 1 is still in there
```

KaiSzuttor:
what is "failing"?

itischler:
```  File "choice_test.py", line 22, in <module>
print(np.random.choice(sys.part).id) ### fails 
File "particle_data.pyx", line 110, in espressomd.particle_data.ParticleHandle.id.__get__
File "particle_data.pyx", line 58, in espressomd.particle_data.ParticleHandle.update_particle_data
RuntimeError: Particle node for id 0 not found!
```

KaiSzuttor:
so, your MWE can be made more "minimal":

``` python
import espressomd
import numpy as np

sys = espressomd.System(box_l=[10,10,10])
sys.part.add(pos=[4,4,4], id=1)
print(np.random.choice(sys.part))
```

KaiSzuttor:
the issue is that the particle list in espresso is not "array-like" because it can have gaps and the index is interpreted as the id (which is not the best idea), right?

itischler:
yes, and np.random.choice thinks if the length of a list is one the index must be 0

KaiSzuttor:
okay, but is it a bug or expected behavior? looking at https://espressomd.github.io/doc/espressomd.html#espressomd.particle_data.ParticleList I'd say it's documented behavior

jhossbach:
> so, your MWE can be made more "minimal":
> 
> ```python
> import espressomd
> import numpy as np
> 
> sys = espressomd.System(box_l=[10,10,10])
> sys.part.add(pos=[4,4,4], id=1)
> print(np.random.choice(sys.part))
> ```

[...]

PR #4283, opened 2021-06-22
Fix system.part choice for Gibbs sample

jhossbach:
Fixes / Avoids #4278 causing a RuntimeError if a box runs empty

### DIFF
--- a/samples/gibbs_ensemble/gibbs_ensemble.py
+++ b/samples/gibbs_ensemble/gibbs_ensemble.py
@@ -80,7 +80,8 @@ class Client(Process):
 
     def move_particle(self, DX_MAX):
         """ Move random particle inside the box """
-        self.old_particle = np.random.choice(self.system.part)
+        random_particle_id = np.random.choice(self.system.part[:].id)
+        self.old_particle = self.system.part[random_particle_id]
         self.old_pos = self.old_particle.pos
         self.old_particle.pos = self.old_pos + \
             (0.5 - np.random.random(size=3)) * DX_MAX
@@ -104,7 +105,8 @@ class Client(Process):
 
     def remove_particle(self):
         """ Remove a random particle """
-        self.old_particle = np.random.choice(self.system.part).to_dict()
+        random_particle_id = np.random.choice(self.system.part[:].id)
+        self.old_particle = self.system.part[random_particle_id].to_dict()
         self.system.part[self.old_particle["id"]].remove()
         self.send_energy()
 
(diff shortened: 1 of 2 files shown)


samples/gibbs_ensemble/gibbs_ensemble.py
```diff
@@ -80,7 +80,9 @@ def send_info(self):
 
     def move_particle(self, DX_MAX):
         """ Move random particle inside the box """
-        self.old_particle = np.random.choice(self.system.part)
+        self.random_particle_id = np.random.choice(self.system.part[:].id)
+        self.old_particle = self.system.part[self.random_particle_id]
+        self.random_particle_id = None
```
jngrad:
```suggestion
        random_particle_id = np.random.choice(self.system.part[:].id)
        self.old_particle = self.system.part[random_particle_id]
```

samples/gibbs_ensemble/gibbs_ensemble.py
```diff
@@ -104,7 +106,9 @@ def revert_add_particle(self):
 
     def remove_particle(self):
         """ Remove a random particle """
-        self.old_particle = np.random.choice(self.system.part).to_dict()
+        self.random_particle_id = np.random.choice(self.system.part[:].id)
+        self.old_particle = self.system.part[self.random_particle_id].to_dict()
+        self.random_particle_id = None
```
jngrad:
```suggestion
        random_particle_id = np.random.choice(self.system.part[:].id)
        self.old_particle = self.system.part[random_particle_id].to_dict()
```

jngrad (changes requested):
Besides requested changes, looks good to me.

samples/gibbs_ensemble/gibbs_ensemble.py
```diff
@@ -80,7 +80,8 @@ def send_info(self):
 
     def move_particle(self, DX_MAX):
         """ Move random particle inside the box """
-        self.old_particle = np.random.choice(self.system.part)
+        self.random_particle_id = np.random.choice(self.system.part[:].id)
+        self.old_particle = self.system.part[self.random_particle_id]
```
jngrad: why do we need to store the `random_particle_id` as a member of the object if it is not used elsewhere?

jhossbach: Looks like I wrote that out of habit, I'll force update it not to store it as a member.

jngrad (approved):

PR #4286, opened 2021-06-24
IBM volume conservation maintenance

### LINKED ISSUE #4284: IBM volume conservation edge case
There is an edge case for the `softID` argument of the `IBM_VolCons` bond that leads to undefined behavior:
* for `softID = IBM_MAX_NUM + 0` the forces are not calculated for particles with that softID
* for `softID = IBM_MAX_NUM + 1` the call to the `std::vector` constructor actually lands us in the destructor according to GDB
* for `softID = IBM_MAX_NUM + 2` there is a segfault
* for `softID = IBM_MAX_NUM + 3` there is a segfault
* for `softID = IBM_MAX_NUM + 4` there is a segfault
* for `softID = IBM_MAX_NUM + 5` there is an ESPResSo runtime error picked up by a `handle_errors()` pretty late in the simulation script (during integration)

Here is the culprit:
https://github.com/espressomd/espresso/blob/b2d3341684425e3abb2c51ebeaa00de0221a1d91/src/core/immersed_boundary/ibm_volcons.cpp#L24-L29

There are three separate issues:
* the sanity check should read `softID >= IBM_MAX_NUM`
* there is no `handle_errors()` in the corresponding Cython code
* this code has zero code coverage

MWE:
```python
from espressomd.virtual_sites import VirtualSitesInertialessTracers
system = espressomd.System(box_l=3 * [10.], time_step=0.01)
system.cell_system.skin = 0.1
system.virtual_sites = VirtualSitesInertialessTracers()
positions = [(0, 0, 0), (1, 0, 0), (0, 1, 0), (0, 0, 1)]
triangles = [(0, 1, 2), (0, 3, 1), (3, 1, 2), (3, 0, 2)]
system.part.add(pos=positions, virtual=4 * [True])
for i1, i2, i3 in triangles:
    tri = espressomd.interactions.IBM_Triel(ind1=i1, ind2=i2, ind3=i3, elasticLaw="Skalak", k1=0, maxDist=3)
    system.bonded_inter.add(tri)
    system.part[i1].add_bond((tri, i2, i3))

volCons = espressomd.interactions.IBM_VolCons(softID=1000, kappaV=1.) # <--- here change the softID
system.bonded_inter.add(volCons)
for p in system.part:
    p.add_bond((volCons,))

system.part[3].pos = (0, 0, 2)
print(system.part[:].f)
```
Expected output:
```
[[-0.5  -0.5  -0.25]
 [-0.5  -1.   -0.5 ]
 [-1.   -0.5  -0.5 ]
 [-1.   -1.   -0.25]]
```

jngrad:
Fixes #4284

Description of changes:
- use a `std::vector` for memory management
- remove `IBM_MAX_NUM` macro, now that any `softID` is valid
- don't crash the simulation when measuring the system energy with IBM bonds
- write a testcase for the IBM volume conservation (partial fix for #2071)


testsuite/python/ibm.py
```diff
@@ -127,6 +128,55 @@ def test_triel(self):
...
+        # Run the integrator to initialize the mesh reference volume.
+        system.integrator.run(0, recalc_forces=True)
+
+        # The restorative force is zero at the moment.
+        np.testing.assert_almost_equal(np.copy(self.system.part[:].f), 0.)
+
+        # Double the cube dimensions. The volume increases by a factor of 8.
+        system.part[:].pos = 2. * positions + system.box_l / 2.
+        system.integrator.run(0, recalc_forces=True)
+        ref_forces = 1.75 * np.array(
```
reinaual, line 189: I had to look into the implementation to understand where the values in `ref_forces` are coming from. Not sure if this is obvious or if it is worth to put a comment somewhere.

reinaual (dismissed):

testsuite/python/ibm.py
jngrad: All I understood was that these forces are acting on the vertices of the mesh to shrink it. I don't understand why they aren't symmetric, e.g. in the form `(+/- a, +/- a, +/- a)` with `a` a constant. Drawing the vectors in `ref_forces` on a piece of paper didn't help me either. I tried using a tetrahedron, but I got `[(-0.5, -0.5, -0.5), (0.5, 0, 0), (0, 0.5, 0), (0, 0, 0.5)]` in one case, and in another case where the particles were placed at different positions, the integration loop halted due to broken triangle bonds... I guess this bond algorithm is tied to the way meshes are triangulated.

jngrad: The volume change is now checked in 46a7c7789a874143.

reinaual (dismissed):

reinaual (approved):

PR #4289, opened 2021-07-02
WIP: Propagation refactoring

RudolfWeeber:
Work in progress.

* Propagation is controlled per particle by a flag called `propagation`. This can be SYSTEM_DEFAULT, VIRTUALSITES_RELATIVE or INERTIALESS_TRACERS.
* When SYSTEM_DEFAULT is set, the system's integrator settings will be applied to the particle, i.e., whaever is chosen by integrator.set_... and thermostat.set_... In the future, there will be a possibility to set the system default with a singel command, combining integrator and thermostat.

current status:
* Particle selection is currently only present for Langevin eqn, Langevin LB coupling, inertialless tracers, virtual sites relative.
* NPT and steepest descent still need to be done.
* currently, the same propagation is applied for rotation and propagation. This should be settable individully.
* For now, particles still need to be marked virtual when used with inertialless tracers or vs relative.
* For now, vs relative and inertialess tracers code is always executed, even if that propagation is not chosen for any particle. This incurs a performance cost.

Decisions to make:
* Is there any reason to keep the virtual flag? In principle the particle propagation flag contains the infromation already. Some code will have to be written which decides whether a degree of freedom contributes to kinetic energy, center of mass, etc based on the system and particle propagation settings.


RudolfWeeber:
Usage:
* Setting of VS implementation via `system.virtual_sites` no longer needed
* `from espressomd.particle_data import Propagation`
* Set inertialess tracers using `p = system.part.add(..., propagation=Propagation.INERTIALESS_TRACER, virtual=TrueS)`
* Set up virtual sites relative usinv `p.vs_auto_relate_to(...)` This will set the particle propagation to `Propagation.VIRTUALSITES_RELATIVE` and `virtual=True`, which can also be done manually.

For now virtual sites relative always updates the quaternion of the vs according to the vs_quat attribute (for ENGINE to work)

src/core/propagation/virtual_sites_relative.hpp
```diff
@@ -16,15 +16,14 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
-#ifndef VIRTUAL_SITES_VIRTUAL_SITES_OFF_HPP
-#define VIRTUAL_SITES_VIRTUAL_SITES_OFF_HPP
 
-#include "config.hpp"
-#ifdef VIRTUAL_SITES
-#include "VirtualSites.hpp"
+#pragma once
```
jngrad, line 20: We cannot use `#pragma once` in ESPResSo because we copy-paste files from `src/` to `build/`. See [Pragma once: Caveats](https://en.wikipedia.org/wiki/Pragma_once#Caveats) for more details.

src/core/grid_based_algorithms/lbgpu.hpp
```diff
@@ -109,13 +109,11 @@ struct LB_rho_v_pi_gpu {
 
 struct LB_node_force_density_gpu {
   Utils::Array<float, 3> *force_density;
-#if defined(VIRTUAL_SITES_INERTIALESS_TRACERS) || defined(EK_DEBUG)
 
   // We need the node forces for the velocity interpolation at the virtual
   // particles' position. However, LBM wants to reset them immediately
   // after the LBM update. This variable keeps a backup
   Utils::Array<float, 3> *force_density_buf;
-#endif
```
jngrad, line 118: We need the `EK_DEBUG` ifdef here.

src/core/EspressoSystemStandAlone.cpp
```diff
@@ -43,9 +41,6 @@ EspressoSystemStandAlone::EspressoSystemStandAlone(int argc, char **argv) {
   Communication::init(mpi_env);
 
   // default-construct global state of the system
-#ifdef VIRTUAL_SITES
-  set_virtual_sites(std::make_shared<VirtualSitesOff>());
-#endif
```
jngrad, line 48: The comment on line 43 is now referring to nothing and will cause confusion, since `mpi_loop()` on line 46 doesn't default construct anything.

src/core/propagation/velocity_verlet_inline.hpp
```diff
@@ -35,7 +35,9 @@
  *  v(t) + 0.5 \Delta t f(t)/m \f] <br> \f[ p(t+\Delta t) = p(t) + \Delta t
  *  v(t+0.5 \Delta t) \f]
  */
-inline void velocity_verlet_propagate_vel_pos(const ParticleRange &particles) {
+template <typename ParticleIterable>
+inline void
+velocity_verlet_propagate_vel_pos(const ParticleIterable particles) {
```
jngrad, line 40: Here and elsewhere, is there any performance penalty in passing the `ParticleIterable` (aka `boost::iterator_range`) by copy instead of by reference?

jngrad:
Closed by #4820.

PR #4297, opened 2021-07-13
Benchmarks maintenance

jngrad:
Description of changes:
- add particle coupling in LB benchmark
- move common code to free functions
- re-enable benchmark tests (fully deterministic)
- write tests for all benchmarks


maintainer/benchmarks/ferrofluid.py
```diff
@@ -101,111 +91,65 @@
...
 system.non_bonded_inter[0, 0].lennard_jones.set_params(
     epsilon=lj_eps, sigma=lj_sig, cutoff=lj_cut, shift="auto")
 
-print("LJ-parameters:")
-print(system.non_bonded_inter[0, 0].lennard_jones.get_params())
-
 # Particle setup
 #############################################################
+system.part.add(
+    pos=np.random.random((n_part, 3)) * system.box_l,
```
itischler, line 98: remove the comment line 84-86 as in the other benchmarks
```
# PRNG seeds
#############################################################
# np.random.seed(1)
```

itischler (changes requested):
one small comment removal, everything else is fine.
I also noticed that the `runner.sh` assumes that the build folder is inside of the espresso folder, which may not always be the case.

maintainer/benchmarks/ferrofluid.py
jngrad: good catch!

jngrad:
> one small comment removal, everything else is fine.
> I also noticed that the `runner.sh` assumes that the build folder is inside of the espresso folder, which may not always be the case.

Thanks for the review. Regarding the assumptions of `runner.sh`, it is so because it is not meant to be executed directly, but rather by `suite.sh`, which guarantees the build directory is inside the espresso folder:
https://github.com/espressomd/espresso/blob/e57687d049aa9f69d3377b01d75767d0e6608d89/maintainer/benchmarks/suite.sh#L41-L42
This is for safety reasons: the script deletes the build folder with `rm -rf` and we don't want to accidentally delete important files if the user makes a typo when running the script with a custom build path and custom espresso path as arguments.

RudolfWeeber (approved):

PR #4298, opened 2021-07-13
lgtm python dependencies

jngrad:
GUI-related Python packages (e.g. `mayavi`) are currently listed as dependencies of `lgtm`, although it doesn't need them and has trouble installing them (`lgtm` logfiles contain many error and warnings). This PR excludes them from the `pip` command and adds the `wheel` package as a dependency (to fix `wheel`-related warnings).

### DIFF
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -8,3 +8,16 @@ extraction:
       - "cd build && cmake .."
     index:
       build_command: "cd build && make -j2"
+  python:
+    python_setup:
+      requirements:
+        - wheel
+      requirements_files:
+        - requirements.txt
+      exclude_requirements:
+        - PyOpenGL
+        - mayavi
+        - pyface
+        - pygame
+        - traits
+        - traitsui


jngrad:
@reinaual @RudolfWeeber

There was an extra noop commit that changed python files to trigger the analysis in `lgtm`, before the last force push (c9c66a8884f4d). Here was the result: [build merge logfile](https://lgtm.com/projects/g/espressomd/espresso/logs/rev/pr-dc3e5694adfa99266ffd88affa264749081c3d73/lang:python/stage:Build%20merge_d2bb93501d315f56567f7fb365354177a338ffad). If you click on the Extraction tab, you will see that there aren't many errors left (look for `stderr`), and they aren't critical. The most frequent one is due to `vtk`, because `lgtm` is first looking for `vtk==7.*,>=7.1.1` (the first condition is derived from the second condition), then falls back to the second condition `>=7.1.1` (which is the one in ESPResSo) to install `vtk` version 9. The [build python logfile](https://lgtm.com/projects/g/espressomd/espresso/logs/rev/pr-dc3e5694adfa99266ffd88affa264749081c3d73/lang:python/stage:Build%20python_9738a65eb80b6e0d3044b11aba05816b6fcb085f) failed due to the timeout. Again, clicking on the Extraction tab shows the job itself, and it is cut after 3 min at line `Running setup.py install for mayavi: started`.

Merging this would remove mayavi and friends from the equation; if `lgtm` is still failing at the build python stage after that, then I'm out of ideas, and the next course of action would probably be to disable the scanning of python files.

reinaual (approved):

PR #4304, opened 2021-07-14
active matter tutorial to notebook

### LINKED ISSUE #4183: Code quintuplication in active matter tutorial
Code for the rectification of active particle is in the ``.ipynb`` notebook, the geometry exercises, the geometry solutions, the simulation exercise and the simulations solution.
It could all be united in one notebook.

christophlohrmann:
Fixes #4183 

Description of changes:
- move the active matter tutorial scripts to a single notebook
- add exercise blocks
- rework text


doc/tutorials/active_matter/active_matter.ipynb
```diff
@@ -112,67 +103,704 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import espressomd.observables\n",
+    "import espressomd.accumulators\n",
+    "import matplotlib.pyplot as plt\n",
```
jngrad, line 143: please use a font size of 18 as explained in the [matplotlib instructions for tutorials](https://github.com/espressomd/espresso/wiki/Documentation#Tutorials)

doc/tutorials/active_matter/active_matter.ipynb
```diff
@@ -112,67 +103,704 @@
...
+    "avacf_result = np.sum(avacf.result(), axis=2)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "fig_msd = plt.figure()\n",
```
jngrad: here and in all subsequent plots, please use the `figsize=(10, 6)` argument as explained in the [matplotlib instructions for tutorials](https://github.com/espressomd/espresso/wiki/Documentation#Tutorials) and check that `make tutorial_am_html` generates an `active_matter.html` file with legible plots

doc/tutorials/active_matter/active_matter.ipynb
```diff
@@ -112,67 +103,704 @@
...
+   "source": [
+    "fig_vacf = plt.figure()\n",
+    "plt.plot(taus_vacf, vacf_result[:, 0], label='active')\n",
+    "plt.plot(taus_vacf, vacf_result[:, 1], label='passive')\n",
+    "plt.xlim((taus_vacf[1], None))\n",
+    "plt.loglog()\n",
+    "plt.xlabel('t')\n",
+    "plt.ylabel('VACF(t)')\n",
+    "plt.legend()\n",
+    "plt.show()"
```
jngrad: In this plot and in the plot thereafter, you are feeding negative values to a plotting area with log-log scales.

Here is a quick and dirty way of removing most (but not all) of the noise (maybe there is a better approach):
```python
def acf_stable_regime(x, y):
    """
    Remove the noisy tail in autocorrelation functions of finite time series.
    """
    cut = np.argmax(y <= 0.) - 2
    assert cut >= 1
    return (x[1:cut], y[1:cut])

fig_vacf = plt.figure()
plt.plot(*acf_stable_regime(taus_vacf, vacf_result[:, 0]), label='active')
plt.plot(*acf_stable_regime(taus_vacf, vacf_result[:, 1]), label='passive')
```

The ACF of finite time series is always noisy in the tail. In some applications, e.g. integrating the ACF, it is necessary to discard *at least* the second half of the ACF (see eq. S36 and the paragraph below it in `/work/jgrad/zeman20b/supporting_information/si.pdf`), because it cancels the first half.

doc/tutorials/active_matter/active_matter.ipynb
```diff
@@ -112,67 +103,704 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The geometry we will use is a cylindrical system with a funnel dividing two halves of the box as shown in the figure below.\n",
+    "\n",
+    "<figure><img src=\"figures/geometry.svg\" style=\"float: center; width: 75%\"/>\n",
+    "<center>\n",
+    "<figcaption>Fig. 4: (a) Sketch of the rectifying geometry which we\n",
+    "simulate for this tutorial. (b) The parameters of the cone funnel.</figcaption>\n",
```
jngrad: The parameters in image 4b don't have the same name as the python variables. Consider removing 4b.

doc/tutorials/active_matter/active_matter.ipynb
```diff
@@ -210,549 +837,189 @@
...
-    "0.25 to visualize the inside of the geometry that you have created. The result\n",
-    "should look like <a href='#fig:rectify'>Fig. 5</a>(a).\n",
-    "\n",
-    "<a id='fig:rectify'></a>\n",
-    "<figure>\n",
-    "<img src=\"figures/rectify.jpg\" style=\"float: center; width: 50%\"/>\n",
-    "<center>\n",
-    "<figcaption>Fig. 5: (a) Snapshot of the rectifying geometry, as visualized with ParaView. (b) Final state for the system with 500 particles and <code>vel=5.0</code>. Note that there are more particles in the left-hand chamber, due to the rectification.</figcaption>\n",
-    "</center>\n",
-    "</figure>"
```
jngrad, line 567: I think this entire section about Paraview should stay in. It's the only tutorial where we explain how to use it. We have a few scripts in `/samples` that generate vtk files and tell the user to open them in Paraview (the samples assume the user is already familiar with the Paraview interface).

It's not an issue for people who don't have Paraview installed on their workstation because an image of the viewport is provided in the tutorial.

doc/tutorials/active_matter/CMakeLists.txt
```diff
@@ -1,23 +1,6 @@
...
-  solutions/enhanced_diffusion.py
-  solutions/flow_field.py
-  solutions/rectification_geometry.py
-  solutions/rectification_simulation.py
-  solutions/solutions.txt)
+  TARGET tutorial_am DEPENDS active_matter.ipynb figures/friction.svg
+  figures/pusher-puller.svg figures/enhanced.svg figures/geometry.svg
+  figures/rectify.jpg figures/rectification.svg figures/flow_field.svg)
 
 nb_export(TARGET tutorial_am SUFFIX "" FILE "active_matter.ipynb" HTML_RUN)
```
jngrad:
```
$> time make tutorial_am_html
[100%] Built target tutorial_am_html
real	5m55.387s
user	5m46.537s
sys	0m7.399s
```
6 min to build the HTML file is really long. The `run_tutorials` CI job already takes 25 min to complete, which is less than 2 min per tutorial. Most of the time is spent in the rectification exercise (4 min 40 sec).

Finding a way to pack the big `RECT_PARAMS` dict variable in a CMake string and pass it as argument to `VAR_SUBST` at line 6 without getting either CMake, the CMake formatting script or one of the many bash implementations used in CI to complain (or crash) will be challenging. What would you think of having the `*_PARAMS` dict only contain simulation state parameters, and having extra variables `*_n_samples`/`*_n_steps_per_sample` for parameters that don't alter the time evolution of the system? This way, it should be possible to only reduce the number of samples in CMake and get under the 2 min bar. Likewise, the unittest file could be made smaller if the number of steps is the only thing you changed in the dicts.

doc/tutorials/active_matter/active_matter.ipynb
```diff
@@ -210,549 +837,189 @@
...
-    "and the information provided here. We will therefore focus on the use of this\n",
-    "script.\n",
-    "\n",
-    "<a id='fig:flow_field'></a>\n",
-    "<figure>",
-    "<img src=\"figures/flow_field.svg\" style=\"float: center; width: 40%\"/>\n",
-    "<center>\n",
-    "<figcaption>Fig. 7: (a) Snapshot of the flow field around a pusher particle visualized with ParaView.</figcaption>\n",
-    "</center>\n",
-    "</figure>\n",
```
jngrad, line 724: Maybe this Paraview picture would be nice to have, so that people who find it more convenient to use a GUI over writing matplotlib scripts know that Paraview is a viable option. But I don't have a strong opinion on that one.

doc/tutorials/active_matter/active_matter.ipynb
```diff
@@ -112,67 +103,704 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def a2quat(theta, phi):\n",
+    "    q1w = np.cos(theta / 2.0)\n",
+    "    q1x = 0\n",
+    "    q1y = np.sin(theta / 2.0)\n",
+    "    q1z = 0\n",
```
jngrad: This functions needs to be documented and tested. Feel free to cherry-pick jngrad/espresso@c747b04e9b93d624599

doc/tutorials/active_matter/active_matter.ipynb
```diff
@@ -210,549 +837,189 @@
...
+    "\n",
+    "fig_vels, ax_vels = plt.subplots()\n",
+    "im = plt.pcolormesh(vel_abs.T, cmap='YlOrRd')\n",
+    "plt.quiver(xs, zs, vels[:, :, 0].T, vels[:, :, 2].T, angles='xy', scale=0.005)\n",
+    "circ = plt.Circle(particle.pos_folded[[0, 2]], 0.5, color='blue')\n",
+    "ax_vels.add_patch(circ)\n",
+    "ax_vels.set_aspect('equal')\n",
+    "plt.xlabel('x')\n",
+    "plt.ylabel('z')\n",
+    "cb = plt.colorbar(im, label=r'$|v_{fluid}|$')\n",
```
jngrad: The subscript appears slanted, did you mean to write `$|v_{\text{fluid}}|$`?

jngrad (changes requested):

doc/tutorials/active_matter/active_matter.ipynb
jngrad: @jhossbach please do `make tutorial_am_html` and look at this particular plot. This type of issue cannot be detected in CI because we only generate the html files on the python branch, i.e. after a PR is merged.

jhossbach: @jngrad You're right, there was a problem with the `\text` part, changing it to `\mathrm` solved it.

doc/tutorials/active_matter/active_matter.ipynb
christophlohrmann: I agree with including some paraview. But not at this stage, where we have to use LB just to get a rasterized boundary.
We can include a paraview part later when we talk about the swimmer flowfield

jngrad (approved):
LGTM

PR #4310, opened 2021-07-22
Refactor reaction method constraints

jngrad:
Description of changes:
- unit test reaction constraints
- make reaction constraint fields private
- make invalid constraint parameters raise a `ValueError`
- make it safe to change a reaction constraint from cylindrical to slab
- make it possible to remove a reaction constraint


jngrad:
@reinaual  This isn't a perfect solution, but there is no need to spend more time on this part of the code for now. The long-term plan is to extract the reaction constraint logic from the `ReactionAlgorithm` class to improve extensibility and separation of concerns. Shouldn't be too hard now that the logic is fully private. An `enum` was the quickest way to make the code safely unit testable.

reinaual (approved):

PR #4325, opened 2021-08-11
Set hidden particle type in Reaction Methods scripts

### LINKED ISSUE #4321: Set hidden particle type in Reaction Methods scripts
In Reaction Method tutorials, samples and python tests labeled as `long`, set the hidden particle type to a value not used in the script and check if the runtime improves. If it does, please open a PR.

davidbbeyer:
Fixes #4321

I checked if manually setting the hidden particle types in the various reaction method scripts to the lowest possible value leads to a speedup. Since it does in all cases (typically between 2x and 4x), I added it to the sample scripts and python tests.


samples/reaction_ensemble.py
```diff
@@ -92,6 +92,10 @@
 print(RE.get_status())
 system.setup_type_map([0, 1, 2])
 
+
+# Set the hidden particle type to the lowest possible number to speed up the simulation
+RE.set_non_interacting_type(3)
```
jonaslandsgesell: Consider introducing a types list or a types dictionary

testsuite/python/constant_pH.py
```diff
@@ -55,6 +55,9 @@ def test_ideal_alpha(self):
             default_charges={type_HA: 0, type_A: -1, type_H: +1})
         RE.constant_pH = pH
 
+        # Set the hidden particle type to the lowest possible number to speed up the simulation
+        RE.set_non_interacting_type(3)
```
jonaslandsgesell: Consider making use of a types list or len(default_charges.keys())

testsuite/python/constant_pH_stats.py
```diff
@@ -71,6 +71,10 @@ def test_ideal_titration_curve(self):
         type_HA = ReactionEnsembleTest.type_HA
         system = ReactionEnsembleTest.system
         RE = ReactionEnsembleTest.RE
+
+        # Set the hidden particle type to the lowest possible number to speed up the simulation
+        RE.set_non_interacting_type(3)
```
jonaslandsgesell: Consider making use of a types list

testsuite/python/reaction_ensemble.py
```diff
@@ -90,6 +90,9 @@ def test_ideal_titration_curve(self):
         RE = ReactionEnsembleTest.RE
         target_alpha = ReactionEnsembleTest.target_alpha
 
+        # Set the hidden particle type to the lowest possible number to speed up the simulation
+        RE.set_non_interacting_type(3)
```
jonaslandsgesell: same here, a little refactoring to avoid magic numbers would be helpful

samples/widom_insertion.py
```diff
@@ -113,6 +113,10 @@
 print(widom.get_status())
 system.setup_type_map([0, 1, 2])
 
+
+# Set the hidden particle type to the lowest possible number to speed up the simulation
+widom.set_non_interacting_type(len(types))
```
jngrad: `max(types)` would be safer here: there is no guarantee that particle types are numbered sequentially starting from 0

testsuite/python/constant_pH.py
jngrad: or `max(default_charges)` which returns the largest dict key

jngrad:
@davidbbeyer please run the formatting script `maintainer/CI/fix_style.sh`, we do test-driven development

davidbbeyer:
The remaining suggestion looks nonsensical to me and does not make the code more legible IMO.

jngrad (approved):
Thanks!

Issue #4219, opened 2021-04-09
Robust exception mechanism in core setters


jngrad:
**Problem statement**

Setters in the core return an error code to indicate whether the core object was successfully initialized or not. The Cython code is expected to check this error code and react to it if it is non-zero. There are however many places in the Cython code where the returned error code was simply ignored and the range checks from the core setter were copy-pasted to the `validate_params()` method of the corresponding Cython class. This code duplication is both unnecessary and error-prone: if the range checks diverge, the user cannot be notified by the Python interpreter that a core object wasn't initialized. See #4217 for an example.

**Solution**

All setters are called from the head node, which means it's safe to throw C++ exceptions. Simply add the `except +` modifier to prototype declarations in .pxd files, and the C++ exception are converted to Python exceptions. The range checks in core setters can be adapted to throw an exception instead of returning an error code. All range checks in `validate_params()` methods should be moved to their corresponding core setters; the `validate_params()` should only do type checking and vector length checking.

The range checks need to be placed before the first line that alters the state of the core object. This way, the system doesn't end up in an undefined state and the user can recover from the error. For example, when adding an actor with incorrect parameters, the error is recovered by deleting the actor from the list of active actors. When updating an active actor with incorrect parameters, no cleanup action is required.

Here are the four main C++ exceptions:
* [`std::domain_error`](https://en.cppreference.com/w/cpp/error/domain_error): the value is outside the domain of validity
   * use when e.g. a positive value was expected but a negative value was passed (pressure, time step, etc.)
   * gets converted to a Python `ValueError`
* [`std::invalid_argument`](https://en.cppreference.com/w/cpp/error/invalid_argument): the value was rejected
   * use when e.g. a keyword was expected but an empty string or a keyword outside the set of allowed keywords was passed (name of the correlation operation, etc.)
   * gets converted to a Python `ValueError`
* [`std::out_of_range`](https://en.cppreference.com/w/cpp/error/out_of_range): access out of bounds
   * use when e.g. accessing an array with a negative index
   * gets converted to a Python `IndexError`
* [`std::runtime_error`](https://en.cppreference.com/w/cpp/error/runtime_error): something unexpected happened
   * use when the three other exceptions aren't suitable
   * gets converted to a Python `RuntimeError`

jngrad:
This proposal has been successfully implemented in the 4.2 and 4.3 lines of ESPResSo. Closing.

PR #4327, opened 2021-08-12
Make ekin immutable and disable ekin checkpointing

### LINKED ISSUE #4210: EK parameter changes after activation do not apply
Changing the EK parameters (like external force density) after the method was activated was activated would not apply, because those parameters are only communicated to the GPU during ``ek_init()`` which is only called on activation or species creation.

christophlohrmann:
Closes #4210

Description of changes:
- Calling an ekin value setter after the method is initialized will throw an error
- This does not play nice with ekin checkpointing, so checkpointing gets disabled


src/core/grid_based_algorithms/electrokinetics_cuda.cu
```diff
@@ -3641,100 +3642,143 @@ void ek_print_lbpar() {
...
 }
 
 int ek_set_fluidcoupling(bool ideal_contribution) {
+  if (ek_initialized)
+    throw std::runtime_error(
+        "Electrokinetics parameters cannot be set after initialisation");
   if (ek_parameters.stencil != 0)
-    return 1; // combination not implemented
-
+    return 1; // combination not implemente
```
jngrad: The error code returned here is not handled in Cython. Please convert it to a `throw std::runtime_error()`. You can then convert all these setter functions to return type `void` and remove their `return 0;` (details in #4219).

src/core/grid_based_algorithms/electrokinetics_cuda.cu
```diff
@@ -3641,100 +3642,143 @@ void ek_print_lbpar() {
...
+    throw std::runtime_error(
+        "Electrokinetics parameters cannot be set after initialisation");
   ek_parameters.fluctuations = fluctuations;
   return 0;
 }
 
 int ek_set_fluctuation_amplitude(float fluctuation_amplitude) {
+  if (ek_initialized)
+    throw std::runtime_error(
+        "Electrokinetics parameters cannot be set after initialisation");
```
jngrad: Not that important, but you could move this logic to a local helper function `inline void setter_throw_if_ek_initialized();` so as to reduce code duplication.

jngrad (changes requested):

jngrad (approved):
Thanks!

PR #4328, opened 2021-08-12
remove a2quat from active matter tutorials, sort and extend rotation doc

### LINKED ISSUE #4320: add quaternion helper to ``espressomd.math``
It would be nice to have a helper function to get the quaternion representation of a particle orientation from more commonly known spherical coordinates. I suggest moving the respective function from the active matter tutorial to espresso's math module. Also links to #4054 
Edit: I just saw that the director can be set directly on a particle. In that case it could be good to reduce the visibility of ``quat`` in the documentation, instead focus on setting ``director`` and leave the quaternion details to the espresso core

christophlohrmann:
Fixes #4320 

Description of changes:
- remove quats from the active matter tutorial
- move rotation description to the particle creation doc
- sort the particle doc so we go from single particle properties to multiple particle properties


jngrad (approved):
LGTM

PR #4333, opened 2021-08-16
Add non-bonded loop trace function

RudolfWeeber:
this is to check that the non bonded loop in conjunction with Verlet Lists and (re)-sorting works fully.

For Lees Edwards and other cases where we do changes to the cell system and Verlet infrastructure.


jngrad (approved):
LGTM

PR #4346, opened 2021-09-13
Testing: fix IBMVolCons test

biermanncarl:
Description of changes:
The IBMVolCons test includes a test of a contracting or expanding cube. The underlying LB fluid, however, was not able to move due to too low resolution of the grid. This pull request fixes the test and allows the cube to actually relax with a reasonable rate. The cost is a higher LB fluid resolution and consequently more computational effort needed to complete the test.

### DIFF
--- a/testsuite/python/ibm.py
+++ b/testsuite/python/ibm.py
@@ -138,7 +138,7 @@ class IBM(ut.TestCase):
     def test_volcons(self):
         '''Check volume conservation forces on a simple mesh (cube).'''
         system = self.system
-        system.virtual_sites = espressomd.virtual_sites.VirtualSitesInertialessTracers()
+        system.virtual_sites = espressomd.virtual_sites.VirtualSitesOff()
         system.thermostat.set_langevin(kT=0, gamma=1, seed=1)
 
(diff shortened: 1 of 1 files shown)


biermanncarl:
Should have checked that... this fix does not work since the domain cannot be sub-partitioned (7 is a prime number)

RudolfWeeber:
It should in principle not be necessary to check the IBM volume conservation with LB. The LB particle coupling is independently tested in the VS tracers tests. You should be able to jut use Langevin with kT=0.

biermanncarl:
> It should in principle not be necessary to check the IBM volume conservation with LB. The LB particle coupling is independently tested in the VS tracers tests. You should be able to jut use Langevin with kT=0.

Do we test the `act_on_virtual=False` option of the LB thermostat, though?

RudolfWeeber:
The lb_thermo_virtual test should do that.

biermanncarl:
The current version uses the Langevin thermostat instead of InertialessTracers in a LB fluid and should correctly test IBMVolCons.

jngrad (approved):
LGTM

PR #4350, opened 2021-09-17
Refactor bonded interactions storage

### LINKED ISSUE #4169: Bonded interactions cannot be deleted

### LINKED ISSUE #4225: Replace bonded interactions cython interface
Bonded interactions should be managed by the script interface and the hand-crafted Cython code should be removed.

Steps:
* Wrap the bonded interaction parameter store by a map-like container
    * should support `at()` and `insert()` and `contains()` as in `std::unordered_map`
    * and an additional `insert()` which auto-generates a free key (as we would like to avoid manual bond id handling in the future)

* Reading clients such as the bond loop should probably only have const access
    * convert read access throughout Espresso source to use `bonded_ia_params.at()`
    * adapt `set_bonded_ia_params()` in `bonded_interactions/bonded_interactions_utils.hpp`

* Remove MPI broadcasting of bonds

* Write an `ObjectMap` analogue to the script interface `ObjectList` which wraps an associative container:
    * `add()` method takes a `KeyType` and a `ManagedType`
    * `remove()` takes a `KeyType`

* Write Python bindings for the `ObjectMap`
    * `__getitem__()`, `__iter__()`, and dictionary-related friends

* Implement the script interface `BondedInteractInteractionsMap`
    * making use of `ObjectMap`, otherwise analogous to `Constraints` or `LBBoundaries`

* Write script interface classes for Espresso's bonded interaction classes

* `@script_interface`-register the classes `BondedInteractionsStore` and the bonded interaction script interface classes in Python

### LINKED ISSUE #4377: Manually overwriting bond objects of different type is undefined behavior
The python interface currently allows overwriting bond objects. This is undefined behavior when a bond is overwritten by a bond of a different type. For example, when replacing a 2-center bond by a 3- or 4-center bond, the list of partners in the `BondView` is not updated and the force calculation will access span elements out of bounds. This bug can be fixed without impacting the collision detection feature (see #4225).

MWE:
```python
system = espressomd.System(box_l=[20.0, 20.0, 20.0])
p1, p2, p3, p4 = system.part.add(pos=[[1, 0, 0], [0, 0, 0], [0, 1, 0], [1, 1, 0]])
system.time_step = 0.01
system.cell_system.skin = 0.4
system.bonded_inter[0] = espressomd.interactions.HarmonicBond(k=0.5, r_0=1.1)
p1.add_bond((system.bonded_inter[0], p2))
print(p1.bonds)
system.bonded_inter[0] = espressomd.interactions.Dihedral(mult=3, bend=5.2, phase=3.)
print(p1.bonds)  # issue: there is only one bonded partner, but Dihedral needs 3 partners
system.integrator.run(0, recalc_forces=True)  # segfault: access out of bounds
```
Output in a debug build:
```
((HarmonicBond(0): {'k': 0.5, 'r_0': 1.1, 'r_cut': 0.0}, 1),)
((Dihedral(0): {'mult': 3.0, 'bend': 5.2, 'phase': 3.0}, 1),)
python3: /work/jgrad/espresso/src/utils/include/utils/Span.hpp:103: constexpr T& Utils::Span<T>::operator[](Utils::Span<T>::size_type) const [with T = Particle*; Utils::Span<T>::reference = Particle*&; Utils::Span<T>::size_type = long unsigned int]: Assertion `(i < size())' failed.
[seraue:3978640] *** Process received signal ***
[seraue:3978640] Signal: Aborted (6)
[seraue:3978640] Signal code:  (-6)
[seraue:3978640] [ 0] /lib/x86_64-linux-gnu/libc.so.6(+0x46210)[0x1540b9cc1210]
[seraue:3978640] [ 1] /lib/x86_64-linux-gnu/libc.so.6(gsignal+0xcb)[0x1540b9cc118b]
[seraue:3978640] [ 2] /lib/x86_64-linux-gnu/libc.so.6(abort+0x12b)[0x1540b9ca0859]
[seraue:3978640] [ 3] /lib/x86_64-linux-gnu/libc.so.6(+0x25729)[0x1540b9ca0729]
[seraue:3978640] [ 4] /lib/x86_64-linux-gnu/libc.so.6(+0x36f36)[0x1540b9cb1f36]
[seraue:3978640] [ 5] /work/jgrad/espresso/build/src/core/EspressoCore.so(Utils::Span<Particle*>::operator[](unsigned long) const+0x45)[0x1540b84920d3]
[seraue:3978640] [ 6] /work/jgrad/espresso/build/src/core/EspressoCore.so(add_bonded_force(Particle&, int, Utils::Span<Particle*>)+0xfa)[0x1540b84c37d9]
[seraue:3978640] [ 7] /work/jgrad/espresso/build/src/core/EspressoCore.so(void CellStructure::execute_bond_handler<bool (*)(Particle&, int, Utils::Span<Particle*>)>(Particle&, bool (*)(Particle&, int, Utils::Span<Particle*>))+0x12a)[0x1540b84c91de]
[seraue:3978640] [ 8] /work/jgrad/espresso/build/src/core/EspressoCore.so(void CellStructure::bond_loop<bool (*)(Particle&, int, Utils::Span<Particle*>)>(bool (* const&)(Particle&, int, Utils::Span<Particle*>))+0xc6)[0x1540b84c7712]
...
```

biermanncarl:
Fixes #4225
Fixes #4377
Fixes #4169

Description of changes:
- Bonded interactions are now stored as ScriptInterface objects, are immutable and can be removed
- It is no longer possible to overwrite a bond object by a different type of bond (avoids segmentation faults)


jngrad (dismissed):
Thanks a lot! From my side, this is ready for merge.

biermanncarl:
Also fixes #4169 ?

RudolfWeeber:
Thank you, Carl. This is a big step forward.

I think we can merge this now, also to unblock related work.

However, some improvements should still be considered:
* It looks like the BondedInteraction script interface class still knows about the numerical id (i,e, key) of the bond in the BondedInteractions storage. That should be avoided. The individual bonded interaction should not (need to) know, how it is stored.
* At least the valid_keys() Python function can probably be autogenerated from the parameter list of the script object.

jngrad:
> At least the valid_keys() Python function can probably be autogenerated from the parameter list of the script object.

Good point! The `valid_keys()` and `required_keys()` methods are now deduced from the script object.

> It looks like the BondedInteraction script interface class still knows about the numerical id (i,e, key) of the bond in the BondedInteractions storage. That should be avoided. The individual bonded interaction should not (need to) know, how it is stored.

This has to remain so for now, unfortunately. The `BondedInteractions` container serializes the bond objects by storing the hashmap key (aka bond id) and the bond object parameter dict in the pickle file. During deserialization, a new bond is created from the parameter dict and it is stored in the hashmap with the same key. So in principle we don't need to store the key in the bond object itself, however the bond objects are also stored as a copy in the `ParticleHandle` objects, and the only way to make the particle bond lists agree with the system bond list after deserialization is to store a unique identifier in the bond objects, e.g. the bond id.

This is not a good design, because each bond is serialized and deserialized multiple times. There is also the issue of the IBM bonds that are hiding the particles ids, and therefore cannot be checkpointed (this is already an issue in the current python branch). I tried to work around that by re-implementing the script interface hashmap in python (4f509f121653e68844650db410f8cd0b955306a5), but this adds more complexity and doesn't work anyway, because IBM bonds as a side effect do a particle fetch in the constructor, but there is no guarantee that all particles in an IBM bond have already been loaded by the time the first particle in the bond is being loaded.

We can probably revisit this design later, for example in the planned project to store bonds symmetrically on particles.

jngrad (approved):

PR #4352, opened 2021-09-21
MC benchmarks

### LINKED ISSUE #4209: Add benchmark for MC methods
Akin to the MD benchmark scripts there should be a benchmark script for the MC methods, which measures the performance of a typical (=important to stakeholders) scenario, so that performance impacts introduced by changes can be identified, and to make
it possible for people without domain knowledge to identify bottlenecks.

kosovan:
Fixes #4209 

Description of changes:
- add a script `mc_acid_base_reservoir.py` to benchmark MC simulations by simulating the grand-reaction ensemble
- add a 'label' column to the CSV file to handle multiple timing outputs per benchmark


kosovan:
@jngrad can you please have a look at our `CMakeLists.txt` ? Clearly, we did not do the last step correctly.

jngrad:
`CMakeLists.txt` looks ok to me.

maintainer/benchmarks/MC-acid-base-reservoir.py
```diff
@@ -0,0 +1,296 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# ionization degree alpha calculated from the Henderson-Hasselbalch equation for an ideal system
+def calc_ideal_alpha(pH, pKa):
+    return 1. / (1 + 10**(pKa - pH))
+
+def calc_donnan_coefficient(c_acid, I_res, charge = -1):
+    return charge*c_acid/(2*I_res) + np.sqrt( ( charge*c_acid/(2*I_res) )**2 + 1 )
```
jngrad: numpy hasn't been imported yet. It would be better to move these files after the imports and argparse logic

maintainer/benchmarks/MC-acid-base-reservoir.py
```diff
@@ -0,0 +1,296 @@
...
+system.integrator.run(steps=1000)
+
+COULOMB_PREFACTOR=BJERRUM_LENGTH_REDUCED * KT_REDUCED
+if USE_P3M:
+    coulomb = espressomd.electrostatics.P3M(prefactor = COULOMB_PREFACTOR, 
+                                            accuracy=1e-3)
+else:
+    coulomb = espressomd.electrostatics.DH(prefactor = COULOMB_PREFACTOR, 
+                                           kappa = KAPPA_REDUCED, 
+                                           r_cut = 1. / KAPPA_REDUCED)
```
jngrad: `KAPPA_REDUCED` is not declared. Consider installing the python linter and formatter with `pip3 install --user "autopep8==1.5.0" "pycodestyle==2.5.0" "pylint>=2.4.4" "astroid>=2.3.3" "isort>=4.3.4" "setuptools>=39.0.1" "pre-commit>=2.2.0" "cmake-format==0.6.11" ` and running them with `pre_commit run --all-files` to detect such issues early. These tools are really useful.

pm-blanco:
@jngrad  do we need to create a test for our Monte Carlo benchmark similar to `testsuite/scripts/benchmarks/test_lb.py` ?

jngrad:
> @jngrad do we need to create a test for our Monte Carlo benchmark similar to `testsuite/scripts/benchmarks/test_lb.py` ?

yes, please use [`testsuite/scripts/benchmarks/test_lb.py`](https://github.com/espressomd/espresso/blob/python/testsuite/scripts/benchmarks/test_lb.py) as template and update `testsuite/scripts/benchmarks/CMakeLists.txt` accordingly

pm-blanco:
@jngrad should I rebase this PR also?

maintainer/benchmarks/MC-acid-base-reservoir.py
```diff
@@ -0,0 +1,336 @@
...
+def equilibrate_reaction(reaction_steps=1):
+    RE.reaction(reaction_steps)
+
+
+def report_progress(system, i, next_i):
+    n_A = system.number_of_particles(type=TYPES['A']) 
+    n_Na = system.number_of_particles(type=TYPES['Na']) 
+    n_Cl = system.number_of_particles(type=TYPES['Cl']) 
+    n_All = len(system.part)
+    if(i == next_i):
```
jngrad: here and elsewhere: in Python you don't need parentheses around `if` statements, the only time we add parentheses is when you need work around the formatting script, which is not the case in this file

maintainer/benchmarks/MC-acid-base-reservoir.py
```diff
@@ -0,0 +1,336 @@
...
+                system.time,
+                i /
+                NUM_SAMPLES *
+                100,
+                n_All,
+                n_Na,
+                n_Cl,
+                n_A,
+                n_A /
+                N_ACID))
```
jngrad: the formatting script is know to be over-zealous, please rewrite this `print` statement in 3 lines or less

maintainer/benchmarks/MC-acid-base-reservoir.py
```diff
@@ -0,0 +1,336 @@
...
+    # caluclate the averages at the end in order to check that we are having
+    # the correct number of particles
+    n_A_avg = np.average(np.array(n_As))
+    n_All_avg = np.average(np.array(n_Alls))
+    print(
+        "Average values: n_All {:.2f} , n_A {:.2f}, alpha {:.4f}".format(
+            n_All_avg,
+            n_A_avg,
+            n_A_avg /
+            N_ACID))
```
jngrad: please use `f-strings`

maintainer/benchmarks/benchmarks.py
```diff
@@ -72,13 +77,23 @@ def get_timings(system, n_steps, n_iterations, verbose=True):
...
         tick = time.time()
-        system.integrator.run(n_steps)
+        if(n_steps_MC > 0):  
+            RE.reaction(n_steps_MC)
+        system.integrator.run(n_steps_integration)
         tock = time.time()
-        t = (tock - tick) / n_steps
+        # FIXME the MC and MD steps are not really additive, we should count
+        # each of them separately
+        t = (tock - tick) / (n_steps_integration + n_steps_MC)
```
jngrad: This is more invasive that what I imagined. It also doesn't scale well, i.e. if we want to do the same for another feature, this function will no longer be maintainable. Could you please copy-paste the body of that function to the call site in the MC script, and then restore this file to its original state? This would allow you to address the FIXME by having two arrays to accumulate the integration time and MC time separately.

You could then write two lines in the CSV output file, one for the MD timing and one for the MC timing. Unfortunately there is currently no column to distinguish these. Right now data is written in ['long format'](https://en.wikipedia.org/wiki/Wide_and_narrow_data), maybe in the future we want to write it in 'narrow format' and let the user figure out how to pivot/melt the dataframe in pandas or Rstudio? Or we add a `flag` column that contains a benchmark-specific integer, i.e. 0 for all benchmarks except for the MC benchmark, where a 0 would mean MD and a 1 would mean MC?

jngrad:
> @jngrad should I rebase this PR also?

Once you're done with the code review, yes. Don't forget to add `Co-authored-by: Peter Košovan <kosovan@icp.uni-stuttgart.de>` at the end of the commit message after a new line, so that Peter gets credited as an author (like in 622497aa0d1bfb).

After that you could also do `git fetch upstream python; git rebase upstream/python` to move the commit on top of the most recent commit, so that you can see the effect of your recently merged loop performance improvement in the benchmark.

pm-blanco:
@jngrad  I have finished correcting all your revision.  I was trying to rebase this PR before pushing but I see now that I have mess it up badly. For some reason, now all the commits in between are now showing up here.

maintainer/benchmarks/benchmarks.py
pm-blanco: For now I have added the two different timings as two lines in the CSV output file. I think that it would be desirable in the future to add some kind of flag as you suggest so one can distinguish the two lines.

pm-blanco:
@jngrad I added all your revisions, could you please take a final look just to make sure I did not screw up anything while trying to rebase this PR?

maintainer/benchmarks/benchmarks.py
jngrad: I've added a 'label' column to distinguish MC from MD rows (0ec5b89cc0e8736).

jngrad (approved):
Thanks a lot for your work!

PR #4355, opened 2021-09-23
Refactor LB exception mechanism

### LINKED ISSUE #4354: LB setters and getters ignore runtime errors
**TL;DR:** The exception mechanism in the LB interface is broken. Runtime errors raised in the core are ignored by the Cython code. This is another manifestation of #4219.

**Background information**

When a C++ function throws an exception, the signal can be caught at the Cython level by marking the included C++ function with the modifier `except X`, where `X` is usually `+` (other symbols exists, and they modify how the signal is forwarded from Cython to Python). When an included C++ function is not called directly from Python, but indirectly via an intermediate Cython function, the Cython function needs to be marked `except X` to forward the signal.

These intermediate Cython functions are declared in the `.pxd` file as `cdef inline`, with or without a C++ return type:

* the `cdef inline` statement doesn't mention the return type: then the return type is a Python object, and the signal is wrapped inside that object; when it is sent to the Python interpreter, the signal is deserialized as a Python exception.
* the `cdef inline` statement mentions a C++ return type: then either the signal is encoded as a sentinel value (only for pointers, `int`, `float`, `enum`) if it doesn't collide with a sentinel value from the C++ function (e.g. -1 for `ES_ERROR` and uninitialized global variables...), or it cannot be encoded (any other C++ type, as well as the `void` return type) and the error message is printed to the screen without halting the flow of the program.

For more details, see the Cython documentation section on [error return values](https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#error-return-values) and click on "Cython" to get the version of the documentation that is relevant to us.

**Problem statement**

The LB interface uses intermediate Cython functions which are not marked as `except X`, so they won't forward errors from the core.

Here is a MWE:
```python
import espressomd.lb
system = espressomd.System(box_l=3 * [6.0])
system.time_step = 0.1
lbf = espressomd.lb.LBFluid(visc=1.0, dens=1.0, agrid=1.0, tau=0.1)
system.actors.add(lbf)
node = lbf[0, 0, 0]
system.actors.remove(lbf)
print('start')
print(node.density) # undefined behavior
node.density = 5 # undefined behavior
print('end')
```
Output:
```
$> ./pypresso mwe.py
start
RuntimeError: LB not activated
Exception ignored in: 'espressomd.lb.python_lbnode_get_density'
  File "mwe3.py", line 9, in <module>
    print(node.density)
RuntimeError: LB not activated
0.0
RuntimeError: LB not activated
Exception ignored in: 'espressomd.lb.python_lbnode_set_density'
  File "mwe3.py", line 10, in <module>
    node.density = 5
RuntimeError: LB not activated
end
$> echo $?
0
```
Expected behavior: the script should halt when printing the value of `node.density` and return an error code.

The intermediate Cython function also have a C++ return type, which complicates the way we need to handle the signal. In particular, we cannot use `except +`. Instead, we need to use specializations depending on the return type. For example, functions with `void` return type require `except *`, which introduces overhead in the form of a function call to `PyErr_Occurred()`. When the chosen `except X` specialization isn't correct for the Cython function return type, a malformed Python exception is raised that halts the flow of the program, but crashes the `unittest` framework with a cryptic error message:

```python
...
        with self.assertRaises(RuntimeError):
            print(lbf.viscosity) # raises a malformed RuntimeError object
...
```
```
  File "../testsuite/python/lb.py", line 135, in test_raise_if_not_active
    print(lbf.viscosity)
  File "/usr/lib/python3.8/unittest/case.py", line 227, in __exit__
    self._raiseFailure("{} not raised".format(exc_name))
AttributeError: '_AssertRaisesContext' object has no attribute '_raiseFailure'
```

This is also an issue in the `walberla` branch and all derived branches (`walberla_ubb`, `walberla_ekin`, `lees_edwards`). Some exceptions halt the flow of the program but are still broken, e.g. the node density setter prints a cryptic error message when called in the global namespace of a Python session, but not in a function namespace, nor in the global namespace of a script:

```
>>> node.density = 5
  File "lb.pxd", line 178, in espressomd.lb.python_lbnode_set_density
RuntimeError: LB not activated

The above exception was the direct cause of the following exception:

SystemError: PyEval_EvalFrameEx returned a result with an error set
```

**Solution**

Remove return types from the intermediate Cython functions, such that they can be marked as `except +`. When the return type is a `VectorXd`, convert it to a list. Write python tests to check the exceptions.

jngrad:
Fixes #4354

Description of changes:
- complete rewrite of `lb.pxd`
- bugfix: C++ exception `NoLBActive()` now raises a Python exception

### DIFF
--- a/testsuite/python/lb.py
+++ b/testsuite/python/lb.py
@@ -106,17 +106,99 @@ class TestLB:
             self.lbf[0, 1].velocity = [1, 2, 3]
 
     def test_raise_if_not_active(self):
-        lbf = self.lb_class(visc=1.0, dens=1.0, agrid=1.0, tau=0.1)
+        class MockLBFluid(self.lb_class):
+            '''LB class mock that ignores runtime errors from agrid and tau.'''
+            @property
+            def agrid(self):
+                return 1.
+
+            @agrid.setter
+            def agrid(self, value):
+                pass
+
+            @property
+            def tau(self):
+                return 0.01
+
+            @tau.setter
+            def tau(self, value):
+                pass
+
+        self.check_raise_if_not_active(self.lb_class, False)
+        self.check_raise_if_not_active(MockLBFluid, True)
+
+    def check_raise_if_not_active(self, lb_class, mock):
+        lbf = lb_class(visc=1.0, dens=1.0, agrid=1.0, tau=0.1)
+
+        # check exceptions from LB actor
         with self.assertRaises(RuntimeError):
-            lbf.viscosity = 0.2
-        with self.assertRaises(RuntimeError):
-            lbf.bulk_viscosity = 0.2
+            _ = lbf.density
         with self.assertRaises(RuntimeError):
             lbf.density = 0.2
         with self.assertRaises(RuntimeError):
+            _ = lbf.viscosity
+        with self.assertRaises(RuntimeError):
+            lbf.viscosity = 0.2
+        with self.assertRaises(RuntimeError):
+            _ = lbf.bulk_viscosity
+        with self.assertRaises(RuntimeError):
+            lbf.bulk_viscosity = 0.2
+        with self.assertRaises(RuntimeError):
+            _ = lbf.seed
+        with self.assertRaises(RuntimeError):
             lbf.seed = 2
         with self.assertRaises(RuntimeError):
-            lbf.agrid = 0.2
+            _ = lbf.kT
+        with self.assertRaises(RuntimeError):
+            lbf.kT = 2
+        with self.assertRaises(RuntimeError):
+            _ = lbf.shape
+        if not mock:
+            with self.assertRaises(RuntimeError):
+                _ = lbf.agrid
+            with self.assertRaises(RuntimeError):
+                lbf.agrid = 0.2
+            with self.assertRaises(RuntimeError):
+                _ = lbf.tau
+            with self.assertRaises(RuntimeError):
+                lbf.tau = 0.01
+        with self.assertRaises(RuntimeError):
+            _ = lbf.pressure_tensor
+        with self.assertRaises(NotImplementedError):
+            lbf.pressure_tensor = np.eye(3, 3)
+        with self.assertRaises(RuntimeError):
+            _ = lbf.ext_force_density
+        with self.assertRaises(RuntimeError):
+            lbf.ext_force_density = [1, 1, 1]
+
+        # check exceptions from LB node
+        self.system.actors.add(lbf)
+        node = lbf[0, 0, 0]
+        self.system.actors.remove(lbf)
+        with self.assertRaises(RuntimeError):
+            _ = node.density
+        with self.assertRaises(RuntimeError):
+            node.density = 1.
+        with self.assertRaises(RuntimeError):
+            _ = node.velocity
+        with self.assertRaises(RuntimeError):
+            node.velocity = [1, 1, 1]
+        with self.assertRaises(RuntimeError):
+            _ = node.pressure_tensor
+        with self.assertRaises(NotImplementedError):
+            node.pressure_tensor = np.eye(3, 3)
+        with self.assertRaises(RuntimeError):
+            _ = node.pressure_tensor_neq
+        with self.assertRaises(NotImplementedError):
+            node.pressure_tensor_neq = np.eye(3, 3)
+        with self.assertRaises(RuntimeError):
+            _ = node.boundary
+        with self.assertRaises(NotImplementedError):
+            node.boundary = 1
+        with self.assertRaises(RuntimeError):
+            _ = node.population
+        with self.assertRaises(RuntimeError):
+            node.population = np.zeros(19)
 
     def test_pressure_tensor_observable(self):
         """
--- a/testsuite/python/lb.py
+++ b/testsuite/python/lb.py
@@ -170,6 +170,8 @@ class TestLB:
             _ = lbf.ext_force_density
         with self.assertRaises(RuntimeError):
             lbf.ext_force_density = [1, 1, 1]
+        with self.assertRaises(RuntimeError):
+            lbf.get_interpolated_velocity([0, 0, 0])
 
         # check exceptions from LB node
         self.system.actors.add(lbf)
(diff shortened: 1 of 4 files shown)


testsuite/python/lb_pressure_tensor.py
```diff
@@ -104,7 +104,7 @@ def test_averages(self):
         # ... globally,
         self.assert_allclose_matrix(
             np.mean(self.p_global, axis=0),
-            p_avg_expected, atol_diag=c_s_lb**2 / 6, atol_offdiag=c_s_lb**2 / 9)
+            p_avg_expected, atol_diag=c_s_lb**2 / 5, atol_offdiag=c_s_lb**2 / 9)
```
reinaual, line 107: why is this tolerance increased?

jngrad: The old `lb.pyx` file used to do unit conversions with `grid` and `tau` as Python floats, while now it is done with C++ doubles, which don't have the same precision. In very long simulations like this statistical tests, a tiny numerical error in the properties of the fluid can lead to large numerical fluctuations after 5000 steps. If we look at the current python branch, the diagonal of the pressure tensor deviates from the expectation value by `0.0600`, and the absolute tolerance is `c_s_lb**2 / 6 = 0.0555`. The test still passes because the diagonal has a large value (`50006.38`), and thus the relative tolerance is `50006.38 * 1e-07 = 0.0050`, which brings the total tolerance to `0.605`, which is larger than the deviation, but not by much. In fact, this was an issue in the walberla branch, where I already had to adjust this tolerance a few weeks ago when the pressure tensor code was added. With the new absolute tolerance of `c_s_lb**2 / 5 = 0.0666` we are no longer in the danger zone.

testsuite/python/lb.py
```diff
@@ -106,17 +106,101 @@ def test_properties(self):
...
+            @property
+            def tau(self):
+                return 0.01
+
+            @tau.setter
+            def tau(self, value):
+                pass
+
+        self.check_raise_if_not_active(self.lb_class, False)
+        self.check_raise_if_not_active(MockLBFluid, True)
```
jngrad, line 128: Note: this mocked LB class was introduced to help testing exceptions thrown from core getters. This works around the function `assert_agrid_tau_set()` that intercepts some of the calls to core functions if agrid or tau is not set. Since this is a function from a compiled Cython file, it's not possible to redefine it from the Python interpreter. The mocked class solves the issue without too much code. In the walberla branch, we don't need this workaround.

testsuite/python/lb_pressure_tensor.py
reinaual: I don't really see why Python floats in CPython should have a different precision than C++ doubles but since the relative tolerance is that large this should not matter.

reinaual (approved):

testsuite/python/lb_pressure_tensor.py
jngrad: For example with the LB density setter, we have to calculate `p_dens * agrid * agrid * agrid` in the unit conversion. The original `lb.pxd` would do `__pyx_PyFloat_AsFloat(__pyx_v_p_dens)` and call `PyNumber_Multiply(__pyx_t_5, __pyx_v_agrid)` three times, the new `lb.pxd` simply does `__pyx_v_dens * pow(__pyx_v_agrid, 3.0)`. This difference is probably enough to introduce an error of the order of `std::numeric_limits<double>::epsilon`.

<details>
<summary>Generated C++ code</summary>

Here is the original `lb.pxd` code:
```c++
/* "espressomd/lb.pxd":127
 *     # get pointers
 *     if isinstance(p_dens, float) or isinstance(p_dens, int):
 *         c_dens = <float > p_dens * agrid * agrid * agrid             # <<<<<<<<<<<<<<
 *     else:
 *         c_dens = p_dens * agrid * agrid * agrid
 */
    __pyx_t_4 = __pyx_PyFloat_AsFloat(__pyx_v_p_dens); if (unlikely((__pyx_t_4 == (float)-1) && PyErr_Occurred())) __PYX_ERR(0, 127, __pyx_L1_error)
    __pyx_t_5 = PyFloat_FromDouble(((float)__pyx_t_4)); if (unlikely(!__pyx_t_5)) __PYX_ERR(0, 127, __pyx_L1_error)
    __Pyx_GOTREF(__pyx_t_5);
    __pyx_t_6 = PyNumber_Multiply(__pyx_t_5, __pyx_v_agrid); if (unlikely(!__pyx_t_6)) __PYX_ERR(0, 127, __pyx_L1_error)
    __Pyx_GOTREF(__pyx_t_6);
    __Pyx_DECREF(__pyx_t_5); __pyx_t_5 = 0;
    __pyx_t_5 = PyNumber_Multiply(__pyx_t_6, __pyx_v_agrid); if (unlikely(!__pyx_t_5)) __PYX_ERR(0, 127, __pyx_L1_error)
    __Pyx_GOTREF(__pyx_t_5);
    __Pyx_DECREF(__pyx_t_6); __pyx_t_6 = 0;
    __pyx_t_6 = PyNumber_Multiply(__pyx_t_5, __pyx_v_agrid); if (unlikely(!__pyx_t_6)) __PYX_ERR(0, 127, __pyx_L1_error)
    __Pyx_GOTREF(__pyx_t_6);
    __Pyx_DECREF(__pyx_t_5); __pyx_t_5 = 0;
    __pyx_t_7 = __pyx_PyFloat_AsDouble(__pyx_t_6); if (unlikely((__pyx_t_7 == (double)-1) && PyErr_Occurred())) __PYX_ERR(0, 127, __pyx_L1_error)
    __Pyx_DECREF(__pyx_t_6); __pyx_t_6 = 0;
    __pyx_v_c_dens = __pyx_t_7;
```

and here is the new `lb.pxd` code:
```c++
/* "espressomd/lb.pxd":125
 * 
 * cdef inline python_lbfluid_set_density(double dens, double agrid) except +:
 *     lb_lbfluid_set_density(dens * agrid**3)             # <<<<<<<<<<<<<<
 * 
 * cdef inline python_lbfluid_set_viscosity(double visc, double agrid, double tau) except +:
 */
  try {
    lb_lbfluid_set_density((__pyx_v_dens * pow(__pyx_v_agrid, 3.0)));
  } catch(...) {
    __Pyx_CppExn2PyErr();
    __PYX_ERR(3, 125, __pyx_L1_error)
  }
```
</details>

PR #4368, opened 2021-10-15
Summer school wrap-up

### LINKED ISSUE #3898: Custom criteria for Steepest Descent: minimal distance vs. relative energy
Until now (as of 07f2a3ae7c8) the documentation gives examples of custom convergence criteria for Steepest Descent that are only based on inter-particle distances. This is not always re-usable, for example it's not suitable for dense, mixed LJ liquids where particle sizes are significantly different (the inter-particle distance must be as large as the largest sigma, yet there might not be enough space in the box to satisfy this criterion for small particles). Using a relative energy criterion to halt integration when, for example, the system energy changed by less than 5%, could be more robust and applicable to more cases. This would however make steepest descent an energy-minimization method.

If we decide to update the documentation in that direction, the samples and tutorials would have to be updated to use a relative energy as the convergence criterion. We could also consider updating the steepest descent class to take either a number of steps or a lambda function to decide when to halt.

### LINKED ISSUE #4369: Fix docstring in energy
Minor fix of the docstrings:
The following need to be `non_bonded` instead of `nonbonded`:

https://github.com/espressomd/espresso/blob/9bd4b845fc312abfcfead8b8934d9166390cfaab/src/python/espressomd/analyze.pyx#L320-L323
In addition, the entry `external_fields` should also be mentioned.

### LINKED ISSUE #4370: Make test fails without error messages
MWE: Clone latest commit from python branch; run `cmake .. && make && make test`
Some tests will fail without error messages

jngrad:
Description of changes:
- document broken CMake feature `make test` (closes  #4370)
- improve documentation of energy/pressure calculation (fixes #4369)
- document energy minimization with Steepest Descent (fixes #3898)
- modernize visualizer tutorial with Steepest Descent, tqdm and logging
- cleanup core and LB test cases to synchronize with the walberla branch
- fix broken tests on configurations not tested in CI (feedback from the summer school)


jngrad:
During today's online tech support session, participants found issues in the unit tests and python tests. I'll push the fixes here.

biermanncarl:
I also found some minor things during my preparation for the tutorials. I considered making a separate pull-request, but could also make some changes on top of this branch.

jngrad:
@biermanncarl good idea, you can write your changes on top of my branch in your fork, and I'll forward them here. I have one more thing add to this PR (documenting something in the reaction methods), if you are faster than me I'll commit it after your changes.

doc/sphinx/analysis.rst
```diff
@@ -578,17 +636,22 @@ Example: Calculating a particle's diffusion coefficient
...
 To obtain the velocity auto-correlation function of particle 0, use::
 
-    obs = ParticleVelocities(ids=(0,))
+    obs = ParticleVelocities(ids=[p1.id])
     c_vel = Correlator(obs1=vel_obs, tau_lin=16, tau_max=20., delta_N=1,
                        corr_operation="scalar_product", compress1="discard1")
 
 The full example can be found in :file:`samples/diffusion_coefficient.py`.
+Note that in this example, the operation :literal:`square_distance_componentwise`
+is used, which is not a true correlation function but rather the squared
```
christophlohrmann: Why are we so critical of the poor MSD correlator?

doc/sphinx/installation.rst
```diff
@@ -87,9 +87,9 @@ are required:
 
 .. code-block:: bash
 
-    sudo apt install python3-matplotlib python3-pint ipython3 jupyter-notebook
+    sudo apt install python3-matplotlib python3-pint python3-tqdm ipython3 jupyter-notebook
```
christophlohrmann, line 90: why do we not install pint and tqdm via pip?

jngrad: We do not want to use pip in general, because users end up with a complex mix of package versions. This can be a source of frustration when running samples and tutorials that require a different version of the packages. Installing packages from Ubuntu repositories ensures that everyone has the same versions, that packages are compatible with each other and that they remain in sync with the NEP 29 schedule.

There are a few packages that need to be installed via pip. For example `scipy` is too old in the Ubuntu repository and is missing the rotation matrix features that we use in python tests. Same issue with Sphinx, which doesn't contain features that are crucial to the ESPResSo user guide. There used to be a time where `pint` also had to be installed from pip because of a critical bug when used with `numpy.cbrt()`.

doc/sphinx/analysis.rst
```diff
@@ -578,17 +636,22 @@ Example: Calculating a particle's diffusion coefficient
...
+distance of each component to the first sample is being computed. Other
+available operations include actual correlation functions, as described
+in the source documentation of :class:`espressomd.accumulators.Correlator`.
```
jngrad:
```suggestion
Note that in this example, the operation :literal:`square_distance_componentwise`
is used, which is not a correlation function in the mathematical sense. Other
available operations include actual correlation functions, as described
in the source documentation of :class:`espressomd.accumulators.Correlator`.
```

doc/sphinx/analysis.rst
jngrad: Should be fixed by 24801f50825.

christophlohrmann (approved):

reinaual (approved):

PR #4372, opened 2021-10-20
jean noel implemented fast dict part-neighbours search

stekajack:
Fixes #

Description of changes:
-

### DIFF
--- a/src/core/cells.cpp
+++ b/src/core/cells.cpp
@@ -84,6 +84,63 @@ std::vector<std::pair<int, int>> get_pairs_filtered(double const distance,
   return ret;
 }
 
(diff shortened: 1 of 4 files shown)


RudolfWeeber:
In what way do the …_deniz versions in cells.cp/hpp differ from the standard get pairs by type?

stekajack:
> In what way do the …_deniz versions in cells.cp/hpp differ from the standard get pairs by type?

yeah Im sorry, had no intention to merge! Jean Noel was kind enough to implement an idea centered about what we talked about, but i am just about to look in to the code and see what is what. Really sorry about this, entirely accidental!

PR #4373, opened 2021-10-20
Implementation of HybridDecomposition

### LINKED ISSUE #4351: WIP: Introduce hybrid ParticleDecomposition
Implement a ParticleDecomposition which is, in turn, backed by two ParticleDecompositions optimized for different respective particle sizes.

This will increase the efficientcy for short range calculations for systems with few very large and much more small particles.
The reason is that even a single large particle forces a larger cell size on the entire system. Due to that large cell size, a large number of non-interacting pairs of small particles are visited during the short range calculation.

## Sketch of possible design

Implmenet a class `HybridDecomposition`, inheriting from `ParticleDecomposition`

The HybridDecompositoin 
* Instances two ParticleDecompositions, of which one is DomainDecomposition, the other probably n-square
* Has a list of particle types to put in the n-square cell system. 
* local_cels and ghost_cells contain cells of both chiel decompositions
* particle_to_cell() forward to particle_to_cell() of the child decompositoin depending on the particle's type
* max_cut() and max_range(): unclear. Maybe type pair specific, or that of n-squre and manual override by user.
* minimum_image_disitance(): Return {box_geo} for now, i.e. use minimum image sitance always (note: domain decomposition works with Euclidan, N-Square does not, but MI is always correct)
* resort() calls resort on both children
* exchange_ghot_comm, collect_ghost_force_comm(): Returns a communicator containing all the communications of both child decompositions
### Takss of the constructor
* Construct both child decompositions (probably domain decompoision first)
* Create the local_cells and ghost_cells spans (in private member vars) containing thre respectige contents of both the child decompositions
* Same for both ghost communicators (note: Ghost communicator=list of communication tasks to achieve a purpose. GhostCommunication: individual communication, e.g. cend cell x to y)
* Add the local cell of the n-square decomposition as a neighbor to all cells in the domain decompoistion

### Destructor
* Destructs both child cecompoistions

### Distance calculation
In the initial implementation, we use minimum image distance everywhere. In a later stage, the decision between eucllidian and MI distance ccan be made in the linked_cell loop depending on the cells involved.

### Cut-off management
When all large particles are in the N-Square  part of the atom decomposition, their size can be ignored in the cutoff calculation of the n-square cell system.
In the experimental phase, we let the user chose this. In the final implementation, the cutoff calculation routines will need the option to calc the cutoff disregarding specified particle types.

pkreissl:
Fixes #4351 

Description of changes:
- Implement HybridDecomposition
- Add tests and doc
- Make DomainDecomposition `class` (instead of `struct`)


RudolfWeeber:
* Please initialize m_n_square_types to empty set
* In resort, please make sure the child decompositions don't clear the diff vector
* in particle to cell you are using the paricle id in the lookup. Should be the particle type, I guess.
* In the same function, return directly from the if/else branches. No need for a temporary variable.
* In the constructor: The child decompositions (and HybridDecomposition) currently use move on the communicator. I'm not sure that will work. With move, the original can be rendered unusable. You might need copy semantics
* You still need to set up the cell neighborships across the two child decompositions. Note that the link cell algorithm (algoriht/linked_cell.hpp) only goes over local_cells and only over red neighbors. So, ghost cells have to be made neighbors of local cells. Probably the easiest solution is to add all cells (local and ghost) of the n_square as red neighbors to the local cells of the domain decomposition.
* Either you have to make the list of particle types to go into the n_square a consturctor argument of the Hybrid Decomposition and thus immutable, or you have to write a resort, which will move wrongly placed particles between the decompositoins afterwards.

pkreissl:
>     * Please initialize m_n_square_types to empty set
Can so so, I didn't initialize empty as ~of~ for now this cannot be set via the python interface, so for initial testing -- but of course you can also change that manually when trying out.

>     * in particle to cell you are using the paricle id in the lookup. Should be the particle type, I guess.
Sure, my bad.

>     * In the same function, return directly from the if/else branches. No need for a temporary variable.
Had that originally, clion complained as it deems this bad style. I will go back to initial solution if you want me to.

RudolfWeeber:
>>* In the same function, return directly from the if/else branches. No need for a temporary variable.

 

>Had that originally, clion complained as it deems this bad style. I will go back to initial solution if you want me to.

At least initialize the variable to a nullptr then, so it doesn’t look like UB af first glance.
 
 
>>* In the constructor: The child decompositions (and HybridDecomposition) currently use move on the communicator. I'm not sure that will work. With move, the original can be rendered unusable. You might need copy >semantics

>This is also how it is done in DomainDecomposition so I guess it should work.

Yes, but there, it is used only once. You are moving the same constructor twice to the two child decompositions.

pkreissl:
Ready for review...

RudolfWeeber (dismissed):
JN, could you maybe also tae a look?

jngrad (approved):
I don't have too much expertise in our cell system, but looks good from the technical side

PR #4374, opened 2021-10-20
Widom excess chemical potential calculation and RxMC performance improvement

pm-blanco:
Description of changes:
-  `WidomInsertion.measure_excess_chemical_potential()` was replaced by `WidomInsertion.calculate_excess_chemical_potential()` which returns the instantaneous value of the excess chemical potential instead of the accumulated mean and std. The mean value and std of the excess chemical potential must be now be calculated by the user elsewhere. An example of how to do it is given in `samples/widom_insertion.py`.
- `ReactionAlgorithm::do_reaction()` now only computes the potential energy of the old state at the begininng of the reaction loop. This value is updated after each successful reaction trial move.

### DIFF
--- a/samples/widom_insertion.py
+++ b/samples/widom_insertion.py
@@ -118,18 +118,21 @@ system.setup_type_map([0, 1, 2])
 # up the simulation
 widom.set_non_interacting_type(max(types) + 1)
 
-n_iterations = 100
+particle_insertion_potential_energy_samples = []
+
+n_iterations = 500
 for i in range(n_iterations):
-    for j in range(50):
-        widom.measure_excess_chemical_potential(insertion_reaction_id)
+    particle_insertion_potential_energy_samples.append(
+        widom.calculate_particle_insertion_potential_energy(0))
     system.integrator.run(steps=500)
+
     if i % 20 == 0:
-        print("mu_ex_pair ({:.4f}, +/- {:.4f})".format(
-            *widom.measure_excess_chemical_potential(insertion_reaction_id)))
         print(f"HA {system.number_of_particles(type=0)}",
               f"A- {system.number_of_particles(type=1)}",
               f"H+ {system.number_of_particles(type=2)}")
 
+mu_ex_mean, mu_ex_Delta = widom.calculate_excess_chemical_potential(
+    particle_insertion_potential_energy_samples=particle_insertion_potential_energy_samples)
 
-print("excess chemical potential for an ion pair ",
-      widom.measure_excess_chemical_potential(insertion_reaction_id))
+print(
+    f"excess chemical potential for an ion pair {mu_ex_mean:.4g} +/- {mu_ex_Delta:.4g}")
--- a/src/core/reaction_methods/ReactionAlgorithm.cpp
+++ b/src/core/reaction_methods/ReactionAlgorithm.cpp
@@ -49,9 +49,13 @@ namespace ReactionMethods {
  * Performs a randomly selected reaction in the reaction ensemble
  */
 int ReactionAlgorithm::do_reaction(int reaction_steps) {
+  // calculate potential energy; only consider potential energy since we
+  // assume that the kinetic part drops out in the process of calculating
+  // ensemble averages (kinetic part may be separated and crossed out)
+  auto current_E_pot = calculate_current_potential_energy_of_system();
   for (int i = 0; i < reaction_steps; i++) {
     int reaction_id = i_random(static_cast<int>(reactions.size()));
-    generic_oneway_reaction(reactions[reaction_id]);
+    generic_oneway_reaction(reactions[reaction_id], current_E_pot);
   }
   return 0;
 }
@@ -278,20 +282,8 @@ std::map<int, int> ReactionAlgorithm::save_old_particle_numbers(
   return old_particle_numbers;
 }
 
-/**
- * Generic one way reaction
- * A+B+...+G +... --> K+...X + Z +...
- * You need to use 2A --> B instead of A+A --> B since in the last case you
- * assume distinctness of the particles, however both ways to describe the
- * reaction are equivalent in the thermodynamic limit (large particle numbers).
- * Furthermore, it is crucial for the function in which order you provide the
- * reactant and product types since particles will be replaced correspondingly!
- * If there are less reactants than products, new product particles are created
- * randomly in the box. Matching particles simply change the types. If there
- * are more reactants than products, old reactant particles are deleted.
- */
 void ReactionAlgorithm::generic_oneway_reaction(
-    SingleReaction &current_reaction) {
+    SingleReaction &current_reaction, double &E_pot_old) {
 
   current_reaction.tried_moves += 1;
   particle_inside_exclusion_radius_touched = false;
@@ -300,18 +292,6 @@ void ReactionAlgorithm::generic_oneway_reaction(
     // rollback of complete reactions
     return;
   }
-
-  // calculate potential energy
-  const double E_pot_old =
-      calculate_current_potential_energy_of_system(); // only consider potential
-                                                      // energy since we assume
-                                                      // that the kinetic part
-                                                      // drops out in the
-                                                      // process of calculating
-                                                      // ensemble averages
-                                                      // (kinetic part may be
-                                                      // separated and crossed
-                                                      // out)
 
   // find reacting molecules in reactants and save their properties for later
   // recreation if step is not accepted
@@ -363,6 +343,8 @@ void ReactionAlgorithm::generic_oneway_reaction(
       delete_particle(to_be_deleted_hidden_ids[i]); // delete particle
     }
     current_reaction.accepted_moves += 1;
+    E_pot_old = E_pot_new; // Update the system energy
+
   } else {
     // reject
     // reverse reaction
--- a/src/core/reaction_methods/WidomInsertion.cpp
+++ b/src/core/reaction_methods/WidomInsertion.cpp
@@ -29,14 +29,14 @@
 
 namespace ReactionMethods {
 
-std::pair<double, double> WidomInsertion::measure_excess_chemical_potential(
+double WidomInsertion::calculate_particle_insertion_potential_energy(
     SingleReaction &current_reaction) {
 
   if (!all_reactant_particles_exist(current_reaction))
     throw std::runtime_error("Trying to remove some non-existing particles "
                              "from the system via the inverse Widom scheme.");
 
-  const double E_pot_old = calculate_current_potential_energy_of_system();
+  auto const E_pot_old = calculate_current_potential_energy_of_system();
 
   // make reaction attempt
   std::vector<int> p_ids_created_particles;
@@ -45,13 +45,13 @@ std::pair<double, double> WidomInsertion::measure_excess_chemical_potential(
 
   // save p_id, charge and type of the reactant particle, only thing we
   // need to hide the particle and recover it
-  const int number_of_saved_properties = 3;
+  auto constexpr number_of_saved_properties = 3;
 
   std::tie(changed_particles_properties, p_ids_created_particles,
            hidden_particles_properties) =
       make_reaction_attempt(current_reaction);
 
-  const double E_pot_new = calculate_current_potential_energy_of_system();
+  auto const E_pot_new = calculate_current_potential_energy_of_system();
   // reverse reaction attempt
   // reverse reaction
   // 1) delete created product particles
@@ -62,15 +62,11 @@ std::pair<double, double> WidomInsertion::measure_excess_chemical_potential(
   restore_properties(hidden_particles_properties, number_of_saved_properties);
   // 3) restore previously changed reactant particles
   restore_properties(changed_particles_properties, number_of_saved_properties);
-  std::vector<double> exponential = {exp(-1.0 / kT * (E_pot_new - E_pot_old))};
-  current_reaction.accumulator_potential_energy_difference_exponential(
-      exponential);
 
-  // calculate mean excess chemical potential and standard error of the mean
-  auto const &accumulator =
-      current_reaction.accumulator_potential_energy_difference_exponential;
-  return {-kT * log(accumulator.mean()[0]),
-          std::abs(-kT / accumulator.mean()[0] * accumulator.std_error()[0])};
+  // calculate the particle insertion potential energy
+  auto const E_pot_insertion = E_pot_new - E_pot_old;
+
+  return E_pot_insertion;
 }
 
 } // namespace ReactionMethods
--- a/src/core/reaction_methods/WidomInsertion.hpp
+++ b/src/core/reaction_methods/WidomInsertion.hpp
@@ -29,8 +29,8 @@ namespace ReactionMethods {
 class WidomInsertion : public ReactionAlgorithm {
 public:
   WidomInsertion(int seed) : ReactionAlgorithm(seed) {}
-  std::pair<double, double>
-  measure_excess_chemical_potential(SingleReaction &current_reaction);
+  double calculate_particle_insertion_potential_energy(
+      SingleReaction &current_reaction);
 };
 
 } // namespace ReactionMethods
--- a/src/python/espressomd/reaction_ensemble.pyx
+++ b/src/python/espressomd/reaction_ensemble.pyx
@@ -612,20 +612,75 @@ cdef class WidomInsertion(ReactionAlgorithm):
 
         self._set_params_in_es_core()
 
-    def measure_excess_chemical_potential(self, reaction_id=0):
+    def calculate_particle_insertion_potential_energy(self, reaction_id=0):
         """
-        Measures the excess chemical potential in a homogeneous system for
-        the provided ``reaction_id``. Please define the insertion moves
-        first by calling the method :meth:`~ReactionAlgorithm.add_reaction`
-        (with only product types specified).
-        Returns the excess chemical potential and the standard error for
-        the excess chemical potential. The error estimate assumes that
-        your samples are uncorrelated.
+        Measures the potential energy when particles are inserted in the
+        system following the reaction  provided ``reaction_id``. Please
+        define the insertion moves first by calling the method
+        :meth:`~ReactionAlgorithm.add_reaction` (with only product types
+        specified).
 
+        Note that although this function does not provide directly
+        the chemical potential, it can be used to calculate it.
+        For an example of such an application please check
+        :file:`/samples/widom_insertion.py`.
         """
         # make inverse widom scheme (deletion of particles) inaccessible.
         # The deletion reactions are the odd reaction_ids
         if(reaction_id < 0 or reaction_id > (deref(self.WidomInsertionPtr).reactions.size() + 1) / 2):
             raise ValueError("This reaction is not present")
-        return deref(self.WidomInsertionPtr).measure_excess_chemical_potential(
+        return deref(self.WidomInsertionPtr).calculate_particle_insertion_potential_energy(
             deref(self.WidomInsertionPtr).reactions[int(2 * reaction_id)])
+
+    def calculate_excess_chemical_potential(
+            self, particle_insertion_potential_energy_samples, N_blocks=16):
+        """
+        Given a set of samples of the particle insertion potential energy,
+        calculates the excess chemical potential and its statistical error.
+
+        Parameters
+        ----------
+        particle_insertion_potential_energy_samples : array_like of :obj:`float`
+            Samples of the particle insertion potential energy.
+        N_blocks : :obj:`int`, optional
+            Number of bins for binning analysis.
+
+        Returns
+        -------
+        mean : :obj:`float`
+            Mean excess chemical potential.
+        error : :obj:`float`
+            Standard error of the mean.
+        """
+
+        def do_block_analysis(samples, N_blocks=16):
+            """
+            Performs a binning analysis of samples.
+            Divides the samples in ``N_blocks`` equispaced blocks
+            and returns the mean and its uncertainty
+            """
+            size_block = int(len(samples) / N_blocks)
+            block_list = []
+            for block in range(N_blocks):
+                block_list.append(
+                    np.mean(samples[block * size_block:(block + 1) * size_block]))
+
+            sample_mean = np.mean(block_list)
+            sample_std = np.std(block_list, ddof=1)
+            sample_uncertainty = sample_std / np.sqrt(N_blocks)
+
+            return sample_mean, sample_uncertainty
+
+        gamma_samples = np.exp(-1.0 * np.array(
+            particle_insertion_potential_energy_samples) / self._params["kT"])
+
+        gamma_mean, gamma_std = do_block_analysis(
+            samples=gamma_samples, N_blocks=N_blocks)
+
+        mu_ex_mean = -1 * np.log(gamma_mean) * self._params["kT"]
+
+        # full propagation of error
+        mu_ex_Delta = 0.5 * self._params["kT"] * abs(-np.log(gamma_mean + gamma_std) -
+                                                     (-np.log(gamma_mean - gamma_std)))
+
+        return mu_ex_mean, mu_ex_Delta
--- a/testsuite/python/widom_insertion.py
+++ b/testsuite/python/widom_insertion.py
@@ -94,19 +94,28 @@ class WidomInsertionTest(ut.TestCase):
             default_charges={self.TYPE_HA: self.CHARGE_HA})
 
     def test_widom_insertion(self):
-        num_samples = 100000
+
+        num_samples = 10000
+        particle_insertion_potential_energy_samples = []
+
         for _ in range(num_samples):
             # 0 for insertion reaction
-            self.Widom.measure_excess_chemical_potential(0)
-        mu_ex = self.Widom.measure_excess_chemical_potential(0)
-        deviation_mu_ex = abs(mu_ex[0] - self.target_mu_ex)
+            particle_insertion_potential_energy = self.Widom.calculate_particle_insertion_potential_energy(
+                0)
+            particle_insertion_potential_energy_samples.append(
+                particle_insertion_potential_energy)
+
+        mu_ex_mean, mu_ex_Delta = self.Widom.calculate_excess_chemical_potential(
+            particle_insertion_potential_energy_samples=particle_insertion_potential_energy_samples)
+
+        deviation_mu_ex = abs(np.mean(mu_ex_mean) - self.target_mu_ex)
 
         self.assertLess(
             deviation_mu_ex,
             1e-3,
             msg="\nExcess chemical potential for single LJ-particle computed via Widom insertion is wrong.\n"
-            + f"  average mu_ex: {mu_ex[0]:.4f}"
-            + f"   mu_ex_std_err: {mu_ex[1]:.5f}"
+            + f"  average mu_ex: {np.mean(mu_ex_mean):.4f}"
+            + f"   mu_ex_std_err: {np.std(mu_ex_Delta):.5f}"
             + f"  target_mu_ex: {self.target_mu_ex:.4f}"
         )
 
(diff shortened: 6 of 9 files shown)


RudolfWeeber:
Contentwise this looks good. However, we use semantic versioning. I.e., as long as the major version (4 in 4.1.2) is the same, we promise not to change the outcome of a simulation script silently.
To make your change non-silent, please rename the python function, i.g., calc_instantaneous_chemical_potential()

samples/widom_insertion.py
```diff
@@ -118,18 +118,37 @@
...
+N_block = 16
+size_block = int(n_iterations / N_block)
+gamma_block = []
+
+for block in range(N_block):
+    gamma_block.append(
+        np.mean(gamma_samples[block * size_block:(block + 1) * size_block]))
+
+mu_ex_mean = -1 * np.log(np.mean(gamma_block)) * temperature
+mu_ex_std = np.sqrt(np.var(gamma_block) / np.mean(gamma_block)) * temperature
```
jonaslandsgesell: np.std(..., dof=1) instead of np.sqrt(np.var(...))

jonaslandsgesell: Please add a note that this is due to linearized error propagation. In principle you can do better via not using linearization: mu_ex_Delta =abs( -1 * np.log(np.mean(gamma_block)+np.std(gamma_block, dof =1)) * temperature -(-1 * np.log(np.mean(gamma_block)-np.std(gamma_block, dof =1)) * temperature)

src/python/espressomd/reaction_ensemble.pyx
```diff
@@ -564,20 +564,17 @@ cdef class WidomInsertion(ReactionAlgorithm):
...
         """
         Measures the excess chemical potential in a homogeneous system for
         the provided ``reaction_id``. Please define the insertion moves
         first by calling the method :meth:`~ReactionAlgorithm.add_reaction`
         (with only product types specified).
-        Returns the excess chemical potential and the standard error for
-        the excess chemical potential. The error estimate assumes that
-        your samples are uncorrelated.
-
+        Returns the instantaneous excess chemical potential
```
jonaslandsgesell: There is no instantaneous chemical potential (which is always an ensemble average, please find a better name), it would also miss the application of -kT log().

If you change the logic of how this function works, please describe the needed computations in the doc string

testsuite/python/widom_insertion.py
```diff
@@ -94,19 +94,38 @@ def setUp(self):
...
+            mu_ex = self.Widom.calc_instantaneous_chemical_potential(0)
+            # One needs to average the activity coefficient not mu_ex
+            gamma = np.exp(-1.0 * mu_ex / self.TEMPERATURE)
+            gamma_samples.append(gamma)
+
+        # Binning procedure to estimate the error
+
+        N_block = 16
+        size_block = int(num_samples / N_block)
+        gamma_block = []
```
jonaslandsgesell: Please encapsulate the logic in a blocking method

jonaslandsgesell:
There is no such thing as an instanteous chemical potential. The chemical potential is per definition an ensemble average. Additionally, the value returned cannot be energy: The application of -kT log is missing (surely because log and the ensemble average do not commute). You could name it compute_exp_of_potential_energy_change_for_excess_chemical_potential_computatiom() which would at be a telling name.

src/core/reaction_methods/WidomInsertion.cpp
```diff
@@ -29,7 +29,7 @@
 
 namespace ReactionMethods {
 
-std::pair<double, double> WidomInsertion::measure_excess_chemical_potential(
+double WidomInsertion::calc_instantaneous_chemical_potential(
```
jonaslandsgesell: The name is misleading

src/core/reaction_methods/WidomInsertion.cpp
```diff
@@ -62,15 +62,12 @@ std::pair<double, double> WidomInsertion::measure_excess_chemical_potential(
...
-  // calculate mean excess chemical potential and standard error of the mean
-  auto const &accumulator =
-      current_reaction.accumulator_potential_energy_difference_exponential;
-  return {-kT * log(accumulator.mean()[0]),
-          std::abs(-kT / accumulator.mean()[0] * accumulator.std_error()[0])};
+  // calculate instantaneous excess chemical potential
+
+  const double mu = E_pot_new - E_pot_old;
+
+  return mu;
```
jonaslandsgesell: The name is misleading it is a potential energy change not an instanteous chemical potential. By the way: the chemical potential mu also has contributions from the kinetic energy of particles (this is why also an ideal gas has a chemical potential). The Widom method only deals with the excess chemical potential due to interactions. Just stating mu would be misleading.

jonaslandsgesell (changes requested):
The current naming is misleading, please improve

src/core/reaction_methods/ReactionAlgorithm.cpp
```diff
@@ -49,9 +49,22 @@ namespace ReactionMethods {
  * Performs a randomly selected reaction in the reaction ensemble
  */
 int ReactionAlgorithm::do_reaction(int reaction_steps) {
+
+  // calculate potential energy
+  double E_pot_old =
+      calculate_current_potential_energy_of_system(); // only consider potential
```
jonaslandsgesell: Naming

src/core/reaction_methods/ReactionAlgorithm.cpp
```diff
@@ -291,7 +304,7 @@ std::map<int, int> ReactionAlgorithm::save_old_particle_numbers(
  * are more reactants than products, old reactant particles are deleted.
  */
 void ReactionAlgorithm::generic_oneway_reaction(
-    SingleReaction &current_reaction) {
+    SingleReaction &current_reaction, double E_pot_old) {
```
jonaslandsgesell, line 287: Why would you submit E_pot old from extern? You can just compute it at the beginning of this function before any change happens. To me reusing old E_pot_old values looks wrong. Please check the logic of your change!

pm-blanco: @jonaslandsgesell  Thank you for your carefull revision of the changes. Let me answer to some of your points:

- I agree with you that the name "instantaneous chemical potential" is missleading. What do you think about renaming it as "particle insertion potential energy"?

- Regarding the reason of submiting E_pot_old from extern. The point is that you only need to compute the energy of the old state the first time that you call the reaction routine. You can avoid to have to compute two times the system energy per reaction by simply updating E_pot_old when a reaction is accepted. Since it is an important change I checked that this implementation does not have any artifact comparing the results of the constant pH tutorial with electrostatics with the current 4.1.4. version of espresso.

jonaslandsgesell: I would be fine with your suggested name.
Right now, I do not see how your change does not change the behavior of the generic one way function. If you perform multiple reactions say n> 2 it has to change behaviour, if e.g. the first reaction is accepted and the energy of the system changes? If not, why?

pm-blanco: Then the value of `E_pot_old` is updated to mach the energy of the new state, see in line 367 in `src/core/reaction_methods/ReactionAlgorithm.cpp`,  `E_pot_old = E_pot_new;`

jonaslandsgesell: You pass by value, in order for E_pot_old to be modified in the outside scope you would need to pass by reference.

samples/widom_insertion.py
```diff
@@ -123,7 +123,7 @@
 n_iterations = 5000
 for i in range(n_iterations):
 
-    mu_ex = widom.calc_instantaneous_chemical_potential(insertion_reaction_id)
+    E_ins = widom.calc_particle_insertion_potential_energy(insertion_reaction_id)
```
jonaslandsgesell: consider E_pot_insertion. The inserted particles also carry kinetic energy which is not reported in this observable

src/core/reaction_methods/WidomInsertion.cpp
```diff
@@ -63,11 +63,11 @@ double WidomInsertion::calc_instantaneous_chemical_potential(
   // 3) restore previously changed reactant particles
   restore_properties(changed_particles_properties, number_of_saved_properties);
 
-  // calculate instantaneous excess chemical potential
+  // calculate calc_particle_insertion_potential_energy
 
-  const double mu = E_pot_new - E_pot_old;
+  const double E_ins = E_pot_new - E_pot_old;
```
jonaslandsgesell: I think we should avoid abbreviation whenever possible https://medium.com/ltunes/what-is-clean-code-naming-conventions-part-1-426d383eb85d

src/core/reaction_methods/WidomInsertion.hpp
```diff
@@ -30,7 +30,7 @@ class WidomInsertion : public ReactionAlgorithm {
 public:
   WidomInsertion(int seed) : ReactionAlgorithm(seed) {}
   double
-  calc_instantaneous_chemical_potential(SingleReaction &current_reaction);
+  calc_particle_insertion_potential_energy(SingleReaction &current_reaction);
```
jonaslandsgesell: Abbreviation for calculate?

jonaslandsgesell (changes requested):

src/core/reaction_methods/WidomInsertion.hpp
pm-blanco: I don't mind changing the nomenclature but is not "calc" as abbreviation for calculate clear enough? I agree that one should try to avoid to abbreviate too much but also I think that too long names can become confusing

jonaslandsgesell: I don't think it s worth to abbreviate words like this

pm-blanco:
@jonaslandsgesell  I belive I have resolved all your comments, please take a look and let me know if you find anything else

jonaslandsgesell:
Thank you for taking care.

samples/widom_insertion.py
```diff
@@ -118,18 +118,54 @@
...
+    sample_mean = np.mean(block_list)
+    sample_std = np.std(block_list)
+
+    return sample_mean, sample_std
+
+
+gamma_mean, gamma_std = do_block_analysis(gamma_samples)   
+
+mu_ex_mean = -1 * np.log(gamma_mean) * temperature
+mu_ex_std = abs(-1 * np.log(gamma_mean) + gamma_std) * temperature - \
```
jonaslandsgesell: I would write: 
`#full propagation of error`
`mu_ex_Delta = 0.5*temperature* abs(-np.log(gamma_mean + gamma_std)  - \
    (-np.log(gamma_mean - gamma_std)) )`
where I
* made sure abs goes around everything
* simplified -1*log to -log (I would also do so in line 163)
* added a comment
* pulled out temperature
* moved gamma_std within the argument of the first log
* the 0.5 stems from the standard deviation also just measuring something like a "half width"


Could you please consider to add the function

`compute_excess_chemical_potential_and_error(mean_potential_insertion_energy, standard_deviation_potential_insertion_energy)` into the WidomInsertion class in the .pyx file? Like this your effort of coming up with the correct computations is available for everyone.
You could also consider making the function signature

`compute_excess_chemical_potential_and_error(potential_energy_changes_due_to_insertions, N_block)` 

such that even the blocking analysis is available for everyone and the mean value and standard deviation computation happens inside compute_excess_chemical_potential, given a list potential_energy_changes_due_to_insertions

jonaslandsgesell:
This PR would be a good check for your benchmark PR. What is the performance before/after your suggested change, when running e.g. 10 reactions in a row for 1000 times?

pm-blanco:
@jonaslandsgesell I did a performance test with the benchmark script of my PR. Performing 1000 MC reactions steps  prior my implementation takes ~ 1.5 s on my laptop. Performing the same number of MC steps with the implementation proposed in this PR takes ~ 0.9 s on my laptop. For a sufficiently large number of reaction steps you gain roughly a factor of 2 in computational time (only in the reaction side, of course).

samples/widom_insertion.py
pm-blanco: I have changed the line as you suggest and I will probably try to make the function in the cython file as well soon (although I do not have the time today).

src/core/reaction_methods/ReactionAlgorithm.cpp
jonaslandsgesell: - [x] @RudolfWeeber @pm-blanco  Ideally, we would now write a test which fails if the function signature is such that values are passed by value and not by reference.

jonaslandsgesell:
Could you please add the unittests which checks if the energy is altered after it is passed in? Then this PR is fine from my side

jngrad:
Once the comments by Jonas are resolved, it would be nice to rebase the PR. I find it challenging to follow the changes, because the 19 commits alternate between the Widom refactoring and the `ReactionAlgorithm::do_reaction` for loop optimization (plus a few random merge and formatting commits). Also most commits don't pass CI, so if I select a commit revision and the code doesn't build or doesn't pass the tests on my workstation, I don't know if the issue comes from my build environment or from the PR.

src/python/espressomd/reaction_ensemble.pyx
```diff
@@ -612,20 +612,59 @@ cdef class WidomInsertion(ReactionAlgorithm):
...
         # make inverse widom scheme (deletion of particles) inaccessible.
         # The deletion reactions are the odd reaction_ids
         if(reaction_id < 0 or reaction_id > (deref(self.WidomInsertionPtr).reactions.size() + 1) / 2):
             raise ValueError("This reaction is not present")
-        return deref(self.WidomInsertionPtr).measure_excess_chemical_potential(
+        return deref(self.WidomInsertionPtr).calculate_particle_insertion_potential_energy(
             deref(self.WidomInsertionPtr).reactions[int(2 * reaction_id)])
+
+    def calculate_excess_chemical_potential(
+            self, particle_insertion_potential_energy_samples, N_blocks=16):
```
jngrad, line 636: I'm not a big fan of long function names and variable names like `particle_insertion_potential_energy_samples`. Wouldn't it be sufficient to call it `energies`? The docstring already explains this argument contains the particle insertion potential energy.

src/python/espressomd/reaction_ensemble.pyx
```diff
@@ -612,20 +612,59 @@ cdef class WidomInsertion(ReactionAlgorithm):
...
+            """
+            size_block = int(len(samples) / N_blocks)
+            block_list = []
+            for block in range(N_blocks):
+
+                block_list.append(
+                    np.mean(samples[block * size_block:(block + 1) * size_block]))
+
+            sample_mean = np.mean(block_list)
+            sample_std = np.std(block_list)
```
jngrad: 1. by default `np.std()` divides the variance by `n` instead of `n - 1`, shouldn't we use `ddof=1` here?
2. when doing binning analysis, aren't we interested in the standard error of the mean? i.e. `sample_std/np.sqrt(N_blocks)`

src/python/espressomd/reaction_ensemble.pyx
```diff
@@ -612,20 +612,59 @@ cdef class WidomInsertion(ReactionAlgorithm):
...
+        """
+        Given a set of samples of the particle insertion potential energy, calculates the excess chemical potential 
+        and its statistical error.
+        Returns excess_chemical_potential_mean, excess_chemical_potential_error   
+        """
+
+        def do_block_analysis(samples, N_blocks=16):
+            """
+            Performs a binning analysis of samples. 
+            Divides the samples in N_blocks equispaced blocks
```
jngrad:
```suggestion
            Divides the samples in ``N_blocks`` equispaced blocks
```
The two backticks will format that word in monospace font and label it as source code. My spellchecker software relies on that information to skip variable names and source code samples.

src/python/espressomd/reaction_ensemble.pyx
```diff
@@ -612,20 +612,59 @@ cdef class WidomInsertion(ReactionAlgorithm):
...
+        return deref(self.WidomInsertionPtr).calculate_particle_insertion_potential_energy(
             deref(self.WidomInsertionPtr).reactions[int(2 * reaction_id)])
+
+    def calculate_excess_chemical_potential(
+            self, particle_insertion_potential_energy_samples, N_blocks=16):
+        """
+        Given a set of samples of the particle insertion potential energy, calculates the excess chemical potential 
+        and its statistical error.
+        Returns excess_chemical_potential_mean, excess_chemical_potential_error   
+        """
```
jngrad: Please follow the numpydoc + napoleon style ([wiki page](https://github.com/espressomd/espresso/wiki/Documentation#Sphinx)). It *could* look like this:
```suggestion
        """
        Given a time series of the particle insertion potential energy,
        calculates the excess chemical potential and its statistical error.

        Parameters
        ----------
        particle_insertion_potential_energy_samples : array_like of :obj:`float`
            Time series of the particle insertion potential energy.
        N_blocks : :obj:`int`, optional
            Number of bins.

        Returns
        -------
        mean : :obj:`float`
            Mean excess chemical potential.
        error : :obj:`float`
            Standard error of the mean.
        """
```
Do not commit this suggestion directly, as it *does not* describe what your code is currently doing (e.g. the error is not the standard error of the mean).

samples/widom_insertion.py
```diff
@@ -118,18 +118,28 @@
...
+mu_ex_mean, mu_ex_Delta = widom.calculate_excess_chemical_potential(
+    particle_insertion_potential_energy_samples=particle_insertion_potential_energy_samples)
 
-print("excess chemical potential for an ion pair ",
-      widom.measure_excess_chemical_potential(insertion_reaction_id))
+print(
+    "excess chemical potential for an ion pair ",
+    mu_ex_mean,
+    " +/- ",
+    mu_ex_Delta)
```
jngrad: f-strings are more readable (I'm not sure about the floating-point precision we need here, though):
```suggestion
print(
   f"excess chemical potential for an ion pair {mu_ex_mean:.4f} +/- {mu_ex_Delta:.4f}")
```

src/python/espressomd/reaction_ensemble.pyx
jonaslandsgesell: I am not a fan of too short variable names. For me this would be fine. I prefer longer names over shorter names.

src/python/espressomd/reaction_ensemble.pyx
jonaslandsgesell: In the Janke publications the formula is provided as suggested by Jean-Noel.
I suppose the derivation of dividing by N_blocks would go by the fact that one computes the variance of the blocking mean

\frac{1}{\mathrm{N_blocks}} \sum_{i=1}^{\mathrm{N_blocks}} \mu_{\mathrm{block}, i}

i.e.

\mathrm{Var}(\frac{1}{\mathrm{N_blocks}} \sum_{i=1}^{\mathrm{N_blocks}} \mu_{\mathrm{block}, i})

(and since each block is uncorrelated to the next one) we get:

\frac{1}{\mathrm{N_blocks}^2} \sum_{i=1}^{\mathrm{N_blocks}} \mathrm{Var}(\mu_{\mathrm{block}, i}) = \frac{1}{\mathrm{N_blocks}} \mathrm{Var}(\mu_{\mathrm{block}})

so that finally the estimated blocking estimate for the standard deviation of the blocking mean is:

\sqrt{\frac{1}{\mathrm{N_blocks}} \mathrm{Var}(\mu_{\mathrm{block}})}

pm-blanco:
@jonaslandsgesell @jngrad  I will try to attend your comments ASAP. Honestly, I have not yet done the unit test because I am unsure on how to implement such test straightforwardly, but I will try to propose something.

jonaslandsgesell:
> @jonaslandsgesell @jngrad I will try to attend your comments ASAP. Honestly, I have not yet done the unit test because I am unsure on how to implement such test straightforwardly, but I will try to propose something.

@jngrad , could you please assist in the unittest? It might require refactoring to get something which is easily testable.

src/core/reaction_methods/ReactionAlgorithm.cpp
jngrad: I guess this could be unit tested in `src/core/reaction_methods/tests/ReactionAlgorithm_test.cpp` by adding a `using ReactionAlgorithm::generic_oneway_reaction;` in the declaration of `ReactionAlgorithmTest` between line 56 and 57, and then adding a new scope at line 232 (between the Monte-Carlo scope and the random number generator scope) with a non-uniform Hamiltonian that would force the move to be accepted.

pm-blanco:
@jonaslandsgesell it took me a while to figure it out how to do it properly, but I managed to implement one unit test in `src/core/reaction_methods/tests/ReactionAlgorithm_test.cpp`. Basicaly I define a simple identity change reaction D <-> E with a reaction constant Gamma = infinity (i.e. completly shifted to products). Then I create one particle D in the system and I call `ReactionAlgorithm::generic_oneway_reaction` providing a fake energy of the system equal to 1. Since it is an ideal system with no interactions and the reaction always will be accepted, after this call the energy should be updated to 0. I checked and the test fails if in `ReactionAlgorithm::generic_oneway_reaction` the energy is provided by value and not by reference and it passes otherwise.
@jngrad I am sorry, but I did not rebase the git history this time. I prefer to practice and learn how to do it properly than mess this PR. Next time I will try to do a more clean PR.  I applied all your other revisions.

src/python/espressomd/reaction_ensemble.pyx
jonaslandsgesell: @pm-blanco, to my and Jean-Noels @jngrad understanding you need to return sample_std/np.sqrt(N_blocks) in the blocking analysis for estimating the uncertainty of the blocking mean

pm-blanco:
@jngrad it is now failing the CI due to some weird error due to an "unexpected indentation in the docstring" in `\src\python\espressomd\reaction_ensemble.pyx`, any idea why?

src/python/espressomd/reaction_ensemble.pyx
pm-blanco: @jonaslandsgesell  solved, I thought that it was encapsulated when one gaves the argument dof=1 to np.std()

src/python/espressomd/reaction_ensemble.pyx
```diff
@@ -612,20 +612,71 @@ cdef class WidomInsertion(ReactionAlgorithm):
...
+        particle_insertion_potential_energy_samples : array_like of :obj:`float`
+            Samples of the particle insertion potential energy.
+        N_blocks : :obj:`int`, optional
+            Number of bins for binning analysis.
+        Returns
+        -------
+        mean : :obj:`float`
+            Mean excess chemical potential.
+        error : :obj:`float`
+            Standard deviation of the mean.
```
jngrad: I don't think that's true. If I'm reading the block analysis code correctly, it currently returns the standard error of the mean, not the standard deviation.

src/core/reaction_methods/tests/ReactionAlgorithm_test.cpp
```diff
@@ -26,12 +26,12 @@
 #include <boost/test/unit_test.hpp>
 
 #include "reaction_methods/ReactionAlgorithm.hpp"
+#include "reaction_methods/ReactionEnsemble.hpp"
```
jngrad: The `ReactionEnsemble` header cannot be included it this file, because it's a unit test for the `ReactionAlgorithm` class. Please either rewrite the new test logic to use the `ReactionAlgorithm` class, or move it to the `ReactionEnsemble_test.cpp` file.

src/core/reaction_methods/tests/ReactionAlgorithm_test.cpp
```diff
@@ -41,6 +41,13 @@
 #include <utility>
 #include <vector>
 
+#define OBSERVABLES_ENERGY_HPP
+#include "electrostatics_magnetostatics/coulomb.hpp"
+#include "energy.hpp"
+#include "observables/Observable.hpp"
+#include <iostream>
+using namespace std;
```
jngrad: Please don't write `#define OBSERVABLES_ENERGY_HPP`. This will silence error messages that would otherwise help in debugging linker errors. You might also instantiate the wrong class templates and function templates without knowing it. We also don't use `using namespace std;` because it interferes with the namespace resolution mechanism. We also don't need `<iostream>`. The other includes need to be moved to the top, in the following order of precedence: espresso core headers before `utils` headers before `boost` before `stl`.

src/core/reaction_methods/tests/ReactionAlgorithm_test.cpp
```diff
@@ -230,6 +238,68 @@ BOOST_AUTO_TEST_CASE(ReactionAlgorithm_test) {
     remove_particle(1);
   }
 
+  // checks that the system energy is updated after a succesful reaction
+
+  {
+
+// Create an instance of the Reaction ensemble class
+#include "reaction_methods/ReactionEnsemble.hpp"
```
jngrad: An include statement in the middle of a function, namespace or boost unittest is dangerous. Please remove it.

src/core/reaction_methods/tests/ReactionAlgorithm_test.cpp
```diff
@@ -230,6 +238,68 @@ BOOST_AUTO_TEST_CASE(ReactionAlgorithm_test) {
...
+
+    int const type_D = 0;
+    int const type_E = 1;
+
+    test_reaction.charges_of_types[0] = 0;
+    test_reaction.charges_of_types[1] = 0;
+
+    float Gamma = INFINITY; // Reaction completly shifted to product creation
+    ReactionMethods::SingleReaction reaction(Gamma, {type_D}, {1}, {type_E},
+                                             {1});
```
jngrad: Variable names start with a lowercase to distinguish them from struct/class names. We use `double` for floating-point values, except in CUDA code. We don't use `INFINITY` nor `std::numeric_limits<double>::infinity()` because infinity values are ignored when ESPResSo is compiled in fast-math mode. Would 10^100 be enough? I thought about `std::numeric_limits<double>::max()`, but then if any value is added to it when calculating the probability, there is a risk of overflow.

```suggestion
    auto const gamma = 1e100; // Reaction completely shifted to product creation
    ReactionMethods::SingleReaction reaction(gamma, {type_D}, {1}, {type_E},
                                             {1});
```

src/core/reaction_methods/tests/ReactionAlgorithm_test.cpp
```diff
@@ -230,6 +238,68 @@ BOOST_AUTO_TEST_CASE(ReactionAlgorithm_test) {
...
+    espresso::system->set_box_l(box_l);
+
+    // Create a D particle in the system
+    auto const ref_position = Utils::Vector3d{0.1, 0.2, 0.3};
+
+    place_particle(0, ref_position);
+    set_particle_type(0, type_D);
+
+    // Add a fake system energy just to check it gets updated
+    double energy = 1.0;
```
jngrad:
```suggestion
    // sentinel value to check energies get updated
    double energy = 1.0;
```

src/python/espressomd/reaction_ensemble.pyx
```diff
@@ -612,20 +612,71 @@ cdef class WidomInsertion(ReactionAlgorithm):
...
+        Parameters
+        ----------
+        particle_insertion_potential_energy_samples : array_like of :obj:`float`
+            Samples of the particle insertion potential energy.
+        N_blocks : :obj:`int`, optional
+            Number of bins for binning analysis.
+        Returns
+        -------
+        mean : :obj:`float`
+            Mean excess chemical potential.
```
jngrad: numpydoc mandates empty lines before sections titles such as `Parameters` and `Returns`, otherwise you get indentation errors in sphinx.

jngrad (changes requested):

jngrad:
For the rebase, it's as simple as
```sh
# backup current state of the branch
git branch Widom+RxMC_fix_backup
# squash all changes
git reset 5bab6bd51e7
```
Then do `git add -p` to select which lines belong to the performance improvement change for the reaction algorithm, commit it, and add the rest to the Widom commit. You can recover the original history at any time with the backed up branch.

pm-blanco:
@jngrad  the PR is now blocked because it does not detect that I did the changes you requested, I belive

jngrad (approved):

PR #4384, opened 2021-10-29
Fix division by zero returning incorrect vector

### LINKED ISSUE #4375: Division by zero in distance calculation of sphere constraint
When placing a particle at the center of a sphere constraint, the distance between the particle and the constraint is calculated incorrectly, resulting in a force on the particle which contains NANs.

Here is a MWE:
```python
import espressomd
import espressomd.shapes

PARTICLE_TYPE = 0
CONSTRAINT_TYPE = PARTICLE_TYPE

system = espressomd.System(box_l=3*[10], periodicity=3*[False])
system.time_step = 1
system.cell_system.skin = 0.1

sphere_shape = espressomd.shapes.Sphere(
    center=0.5*system.box_l,
    radius=0.25*system.box_l[0],
    direction=-1
)
sphere_constraint = espressomd.constraints.ShapeBasedConstraint(
    shape=sphere_shape, penetrable=True, particle_type=CONSTRAINT_TYPE)
system.constraints.add(sphere_constraint)

system.non_bonded_inter[PARTICLE_TYPE,
                        CONSTRAINT_TYPE].wca.set_params(epsilon=1., sigma=1.)

assert(system.constraints[0].penetrable == True)

system.part.add(pos=system.box_l/2, type=PARTICLE_TYPE)

print(system.constraints[0].min_dist())  # 2.5
system.integrator.run(0)
print(system.part[:].f)  # [[nan, nan, nan]]

system.integrator.run(1)  # ERROR: Constraint violated by particle 0 dist -nan
```
The scalar distance is actually calculated correctly, but the distance vector `vec` in [Sphere.cpp](https://github.com/espressomd/espresso/blob/f02fe8d42f13840cea3ebce0cf77344eb5933bba/src/shapes/src/Sphere.cpp#L27) contains NANs due to [`c_dist`](https://github.com/espressomd/espresso/blob/f02fe8d42f13840cea3ebce0cf77344eb5933bba/src/shapes/src/Sphere.cpp#L30) being zero.

RiccardoFrenner:
Fixes #4375

Description of changes:
- Check if distance from center of sphere is zero and if so, return a vector with the correct length instead of {inf, inf, inf}.
- Added unit test for the sphere shape distance function.

### DIFF
--- a/src/shapes/src/Sphere.cpp
+++ b/src/shapes/src/Sphere.cpp
@@ -32,8 +32,13 @@ void Sphere::calculate_dist(const Utils::Vector3d &pos, double &dist,
   if (m_direction == -1) {
     /* apply force towards inside the sphere */
     dist = m_rad - c_dist;
-    auto const fac = dist / c_dist;
-    vec *= fac;
+    if (c_dist == 0) {
+      // any vector of length `dist` is correct in this case
+      vec = Utils::Vector3d{dist, 0, 0};
+    } else {
+      auto const fac = dist / c_dist;
+      vec *= fac;
+    }
   } else {
     /* apply force towards outside the sphere */
     dist = c_dist - m_rad;
--- a/src/shapes/src/Sphere.cpp
+++ b/src/shapes/src/Sphere.cpp
@@ -42,8 +42,13 @@ void Sphere::calculate_dist(const Utils::Vector3d &pos, double &dist,
   } else {
     /* apply force towards outside the sphere */
     dist = c_dist - m_rad;
-    auto const fac = dist / c_dist;
-    vec *= -fac;
+    if (c_dist == 0) {
+      // any vector of length `dist` is correct in this case
+      vec = Utils::Vector3d{dist, 0, 0};
+    } else {
+      auto const fac = dist / c_dist;
+      vec *= -fac;
+    }
   }
 }
 } // namespace Shapes
--- a/src/shapes/unit_tests/Sphere_test.cpp
+++ b/src/shapes/unit_tests/Sphere_test.cpp
@@ -1,0 +1,66 @@
+/*
+ * Copyright (C) 2010-2021 The ESPResSo project
+ * Copyright (C) 2002,2003,2004,2005,2006,2007,2008,2009,2010
+ *   Max-Planck-Institute for Polymer Research, Theory Group
+ *
+ * This file is part of ESPResSo.
+ *
+ * ESPResSo is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * ESPResSo is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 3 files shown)


src/shapes/src/Sphere.cpp
```diff
@@ -32,8 +32,13 @@ void Sphere::calculate_dist(const Utils::Vector3d &pos, double &dist,
...
     dist = m_rad - c_dist;
-    auto const fac = dist / c_dist;
-    vec *= fac;
+    if (c_dist == 0) {
+      // any vector of length `dist` is correct in this case
+      vec = Utils::Vector3d{dist, 0, 0};
+    } else {
+      auto const fac = dist / c_dist;
+      vec *= fac;
+    }
```
jngrad, line 41: 1. I'm not sure this will solve your MWE if you change `direction` to `+1`; the `else` branch needs the same fix
2. This won't pass CI, because it introduces code that is untested; please consider creating a new unit test `src/shapes/unit_tests/Sphere_test.cpp` using e.g. `src/shapes/unit_tests/Ellipsoid_test.cpp` as a guide to test a few trivial points, e.g. `(0, 0, 0), (1, 0, 0), (0, 1, 0), (0, 0, 1)`, with direction +1 and -1

jngrad (changes requested):

src/shapes/unit_tests/Sphere_test.cpp
```diff
@@ -0,0 +1,66 @@
...
+  auto const eps = std::numeric_limits<double>::epsilon();
+
+  s.rad() = 1.0;
+
+  pos = {0., 0., 0.};
+  s.calculate_dist(pos, dist, vec);
+  double always_pos_dist = -s.direction() * dist;
+  BOOST_REQUIRE_GE(always_pos_dist, 0.0);
+  BOOST_REQUIRE_LE(always_pos_dist - s.rad(), eps);
+  BOOST_REQUIRE_LE(vec.norm() - always_pos_dist, eps);
```
jngrad: Here and at 3 other lines, you probably meant to write `BOOST_REQUIRE_CLOSE`. For example at line 45, `BOOST_REQUIRE_LE` will evaluate to true if `vec.norm()` is 0 and `always_pos_dist` is 10^30.

RiccardoFrenner: But why should the norm be 0 and the distance be much larger?

RiccardoFrenner: Never mind, that's the reason I test it.

src/shapes/unit_tests/Sphere_test.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include <utils/Vector.hpp>
+
+#include <limits>
+
+void check_distance_function(Shapes::Sphere &s) {
+  Utils::Vector3d pos;
+  Utils::Vector3d vec;
+  double dist;
+  // multiply by 100 because BOOST_REQUIRE_CLOSE takes a percentage tolerance
+  auto const tol = std::numeric_limits<double>::epsilon()*100;
```
jngrad: Not sure this line will pass the formatting checks, but other than that the PR looks fine from my side. Maybe consider rebasing this PR into a single commit to make the history cleaner.

jngrad (approved):

Issue #4391, opened 2021-11-12
Bond objects retrieved from particles don't match original


RudolfWeeber:
MNWE:
```
import espressomd
from espressomd.interactions import HarmonicBond

s=espressomd.System(box_l=[1]*3)
h=HarmonicBond(k=1,r_0=0)
s.bonded_inter.add(h)
p1=s.part.add(pos=[0]*3)
p2=s.part.add(pos=[0]*3)
b=[(h,p2.id)]
p1.bonds=b
print(p1.bonds)
print(b)
assert p1.bonds == b # which fails
```

I susepct the reason is that a new ScriptObject is instanced when retrieving a bond by id

```
    def __getitem__(self, bond_id):
        return self._get_bond(bond_id)

    def _get_bond(self, bond_id):
[...]
        # Find out the type of the interaction from ESPResSo
        bond_type = get_bonded_interaction_type_from_es_core(bond_id)
[...]
        # Find the appropriate class representing such a bond
        bond_class = bonded_interaction_classes[bond_type]

        # Create a new script interface object (i.e. a copy of the shared_ptr)
        # which links to the bonded interaction object
        return bond_class(bond_id)
```

I think that, whil it represents the sem bond, it creates a new script object.
Wouldn't it be possible to return the existing bond script object from the ObjectMap (via call-method). See the implementation in the breaking_bonds branch on my github repo.

jngrad:
It does create a new script object, but we adapted the `__richcmp__()` method such that we compare the sip. Bond objects are perfectly comparable, and we have a test for it:
https://github.com/espressomd/espresso/blob/b801493a3390b0d03d03bfe4306161f404648d0f/testsuite/python/interactions_bonded_interface.py#L121-L123
Your MNWE fails because you're comparing a tuple to a list, which must evaluate to False. Change the last line to this:
```python
assert p1.bonds == tuple(b)
```

RudolfWeeber:
Ah, missed that. Sorry for the noise.

That aside, what was the reason for instancing a new ScriptObject rather than returning the existing one stored by the ObjectMap?

jngrad:
> That aside, what was the reason for instancing a new ScriptObject rather than returning the existing one stored by the ObjectMap?

I can have a look.

jngrad:
> > That aside, what was the reason for instancing a new ScriptObject rather than returning the existing one stored by the ObjectMap?
> 
> I can have a look.

[...]

PR #4394, opened 2021-11-22
Script interface maintenance

jngrad:
Description of changes:
- allow `@property`-setters in `ScriptInterfaceHelper` classes instead of silently skipping them
- unit test the `ScriptInterfaceHelper` class in python
- re-enable benchmark tests
- fix regressions introduced by #4350
- store bond objects in the ObjectMap (see https://github.com/espressomd/espresso/issues/4391#issuecomment-968664361)
- stop using the core global variable `this_node` in the script interface

### DIFF
--- a/src/script_interface/LocalContext.hpp
+++ b/src/script_interface/LocalContext.hpp
@@ -39,10 +39,11 @@ namespace ScriptInterface {
  */
 class LocalContext : public Context {
   Utils::Factory<ObjectHandle> m_factory;
+  bool m_is_head_node;
 
 public:
-  explicit LocalContext(Utils::Factory<ObjectHandle> factory)
-      : m_factory(std::move(factory)) {}
+  explicit LocalContext(Utils::Factory<ObjectHandle> factory, int mpi_rank)
+      : m_factory(std::move(factory)), m_is_head_node(mpi_rank == 0) {}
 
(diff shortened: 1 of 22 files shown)


src/script_interface/LocalContext.hpp
```diff
@@ -39,10 +39,11 @@ namespace ScriptInterface {
  */
 class LocalContext : public Context {
   Utils::Factory<ObjectHandle> m_factory;
+  bool m_is_head_node;
 
 public:
-  explicit LocalContext(Utils::Factory<ObjectHandle> factory)
-      : m_factory(std::move(factory)) {}
+  explicit LocalContext(Utils::Factory<ObjectHandle> factory, int mpi_rank)
```
reinaual, line 45: I guess the `explicit` was only necessary because of [Core Guideline C46](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c46-by-default-declare-single-argument-constructors-explicit)

reinaual (approved):

src/script_interface/LocalContext.hpp
jngrad: I wasn't sure. In any case, I have more changes pending on the script interface. If it is not needed anymore, I'll remove the `explicit` in one of the next PRs.

PR #4396, opened 2021-11-29
Test cylindrical LB velocity profile observable

### LINKED ISSUE #4149: CylindricalLBVelocityProfile is not tested
The corresponding Test only checks getters/setters, not the actual profile

jngrad:
Fixes #4149

Description of changes:
- write test cases for `CylindricalLBVelocityProfile`
- remove code duplication in LB checkpointing code


testsuite/python/lb_circular_couette.py
```diff
@@ -0,0 +1,186 @@
...
+
+        # get Taylor-Couette coefficients
+        (a_sim, b_sim), _ = scipy.optimize.curve_fit(
+            taylor_couette_kernel, r, v_phi)
+        # check deviations are small (the outer cylinder isn't perfect)
+        v_phi_fit = taylor_couette_kernel(r, a_sim, b_sim)
+        np.testing.assert_allclose(v_phi, v_phi_fit, atol=1e-2)
+        # check the correct coefficients are recovered
+        a_ref, b_ref = (-0.0388038, 0.019088255)
+        np.testing.assert_allclose([a_sim, b_sim], [a_ref, b_ref], atol=1e-6)
```
jngrad: @christophlohrmann This test is a backport of the `walberla` branch circular Couette test. Since we don't have per-node slip velocity conditions on the `python` branch, I approximated the outer cylinder with 8 moving walls. I hardcoded the reference coefficients here because the Taylor-Couette equations seems unstable when the outer cylinder rotates instead of the inner cylinder. Let me know if you think there is a better way to write this test.

reinaual: The reference values are obtained using the walberla branch with the per-node boundary condition?

reinaual (dismissed):

testsuite/python/lb_circular_couette.py
christophlohrmann: Tbh I am surprised the coefficients match to 1e-6 with the octagon construction. Seems suspiciously good.
Also why do we not check against the analytic coefficients? a_ref and b_ref should be calculated

jngrad: This is now tested against the analytical formula. The deviation is 1e-3 due to the inaccuracy of the octagon.

christophlohrmann (approved):

reinaual (approved):

PR #4401, opened 2021-12-07
Add short range neighbor methods

### LINKED ISSUE #4399: Efficient search for interacting particles
Currently the search for (short-range) interacting particles, e.g., in the reaction ensemble, is n-square, as all other particles in the system are considered. This is not necessary when the cell system is a regular domain decomposition.

Objectives:
* Provide a function template in `CellStructure` that can execute a given kernel function on all particles within the short-range interaction range of a given particle p.
* Using that, implement a kernel that collects the ids of the particles in the neighborhood of p.
* In the same way, implement a kernel calculating the short range energy contribution of a particle with its neighbors
* Implement python interface functions for those two cases

## Implement `CellStructure::run_on_particle_short_range_neighbors`

This should take a kernel and a particle, i.e., 

```c++
template <typename Kernel> 
bool run_on_particle_short_range_neighbors(const Particle& p) 
```

Steps:
* Identify the cell in which the particle is stored using `particle_to_cell` (It might be necessary to use the position at the last resort, which is also in p.r).
* If a `nullptr` is returned, the particle should not be on the mpi rank. Return `false`
* For the returned cell:
    * iterate over the particles in the cell itself (`cell.particles()`
    * Iterate over neighboring cells, both in in red_neighbors and black_neighbors (see `Cell.hpp`). Iterate over the particles in those neighboring cells
    * For all these particles, except p itself (compare using addresses, not values), run `kernel(p)`
* Return `true`

## Implement `get_short_range_neighbors(const Particle& p)` in cells.cpp/hpp

* Initialize a `std::vector<int>` to store the particle ids.
* As kernel, to `CellStrucutre.run_on_particle_short_range_neighbors()`, use a lambda `[&result_vector](const Particle& p)` which appends the particle id to the result vector. See the get_pairs code in `cells.cpp` for inspiration

## Implement `particle_short_range_energy_contribution(const Particle& p)` in energy.cpp/hpp

Same as above, but 
* use a double as result type
* Capture the Particle p in the lambda `[&result, &p](const Particle& q)`
* In the lamdba, add the pair energy provided by `add_non_bonded_pair_energy()` (`energy_inline.hpp`) to the result

## Python interface

* You need the head node function called from cython `mpi_...`
* The callback function (which should return a `boost::optional`)
* Register the callback via REGISTER_CALLBACK
* The callbacks should be declares such that a result from exactly one mpi rank is expected.
See `get_particle_data_local`, corresponding `REGISTER_CALLBACK_ONE_RANK` and the mpi call in `get_particle_data()` (below where it says "cache miss") FOR INSPIRATION

### LINKED ISSUE #4438: Reaction methods should use the cell system to find neighboring particles
When inserting/deleting a particle in the reaction methods, instead of calculating its distance to all particles the cell system should be used to find the neighboring particles and calculate the distance only to those to check for the exclusion radius.

jhossbach:
Fixes #4399, fixes #4438

Description of changes:
- `CellStructure.cpp`: Added the `CellStructure::run_on_particle_short_range_neighbors()` method which runs a kernel function over all particles inside the cell and its neighbors of a given particle
- `cells.hpp`: Added the `mpi_get_short_range_neighbors()` function to execute a parallel search
   - can be called from the Python interface via `system.cell_system.get_neighbors.get_neighbors(p, distance)`
- `energy.hpp`: Added the `compute_non_bonded_pair_energy()` function which returns both the short-range Coulomb interactions plus the non-bonded energy contributions  of two particles
   - can be called from the Python interface via `system.analysis.particle_energy(p)`
- Reaction methods can use the new neighbor search algorithm with constructor argument `search_algorithm="parallel"` (default is `search_algorithm="order_n"`); on 1 MPI rank the original order N algorithm is faster since the parallel algorithm introduces some overhead due to the ghost update (this overhead is negligible with 2 or more MPI ranks)

### DIFF
--- a/src/core/cells.cpp
+++ b/src/core/cells.cpp
@@ -33,6 +33,7 @@
 
 #include "Particle.hpp"
 #include "communication.hpp"
+#include "errorhandling.hpp"
 #include "event.hpp"
 #include "grid.hpp"
 #include "integrate.hpp"
@@ -113,7 +114,50 @@ static void search_distance_sanity_check(double const distance) {
                             std::to_string(range));
   }
 }
+static void search_neighbors_sanity_check(double const distance) {
+  search_distance_sanity_check(distance);
+  if (cell_structure.decomposition_type() ==
+      CellStructureType::CELL_STRUCTURE_HYBRID) {
+    throw std::runtime_error("Cannot search for neighbors in the hybrid "
+                             "decomposition cell system");
+  }
+}
 } // namespace detail
+
+boost::optional<std::vector<int>>
+mpi_get_short_range_neighbors_local(int const pid, double const distance,
+                                    bool run_sanity_checks) {
+
+  if (run_sanity_checks) {
+    detail::search_neighbors_sanity_check(distance);
+  }
+  on_observable_calc();
+
+  auto const p = cell_structure.get_local_particle(pid);
+  if (not p or p->is_ghost()) {
+    return {};
+  }
+
+  std::vector<int> ret;
+  auto const cutoff2 = distance * distance;
+  auto kernel = [&ret, cutoff2](Particle const &p1, Particle const &p2,
+                                Utils::Vector3d const &vec) {
+    if (vec.norm2() < cutoff2) {
+      ret.emplace_back(p2.id());
+    }
+  };
+  cell_structure.run_on_particle_short_range_neighbors(*p, kernel);
+  return ret;
+}
+
+REGISTER_CALLBACK_ONE_RANK(mpi_get_short_range_neighbors_local)
+
+std::vector<int> mpi_get_short_range_neighbors(int const pid,
+                                               double const distance) {
+  detail::search_neighbors_sanity_check(distance);
+  return mpi_call(::Communication::Result::one_rank,
+                  mpi_get_short_range_neighbors_local, pid, distance, false);
+}
 
 std::vector<std::pair<int, int>> get_pairs(double const distance) {
   detail::search_distance_sanity_check(distance);
--- a/src/core/energy.cpp
+++ b/src/core/energy.cpp
@@ -128,3 +128,36 @@ double observable_compute_energy() {
   auto const obs_energy = calculate_energy();
   return obs_energy->accumulate(0);
 }
+
+double particle_short_range_energy_contribution_local(int pid) {
+  double ret = 0.0;
+
+  if (cell_structure.get_resort_particles()) {
+    cells_update_ghosts(global_ghost_flags());
+  }
+
+  auto const p = cell_structure.get_local_particle(pid);
+
+  if (p) {
+    auto kernel = [&ret](Particle const &p, Particle const &p1,
+                         Utils::Vector3d const &vec) {
+#ifdef EXCLUSIONS
+      if (not do_nonbonded(p, p1))
+        return;
+#endif
+      auto const &ia_params = *get_ia_param(p.type(), p1.type());
+      // Add energy for current particle pair to result
+      ret += calc_non_bonded_pair_energy(p, p1, ia_params, vec, vec.norm());
+    };
+    cell_structure.run_on_particle_short_range_neighbors(*p, kernel);
+  }
+  return ret;
+}
+
+REGISTER_CALLBACK_REDUCTION(particle_short_range_energy_contribution_local,
+                            std::plus<double>())
+
+double particle_short_range_energy_contribution(int pid) {
+  return mpi_call(Communication::Result::reduction, std::plus<double>(),
+                  particle_short_range_energy_contribution_local, pid);
+}
--- a/src/core/energy.hpp
+++ b/src/core/energy.hpp
@@ -42,4 +42,15 @@ double calculate_current_potential_energy_of_system();
 /** Helper function for @ref Observables::Energy. */
 double observable_compute_energy();
 
+/**
+ * @brief Compute short-range energy of a particle.
+ *
+ * Iterates through particles inside cell and neighboring cells and computes
+ * energy contribution for a specificparticle.
+ *
+ * @param pid    Particle id
+ * @return Non-bonded energy of the particle.
+ */
+double particle_short_range_energy_contribution(int pid);
+
 #endif
--- a/src/python/espressomd/analyze.pyx
+++ b/src/python/espressomd/analyze.pyx
@@ -29,6 +29,7 @@ from .utils cimport Vector3i, Vector3d, Vector9d
 from .utils cimport make_Vector3d
 from .utils cimport make_array_locked
 from .utils cimport create_nparray_from_double_array
+from .particle_data cimport particle, ParticleHandle
 
 
 def autocorrelation(time_series):
@@ -343,6 +344,24 @@ class Analysis:
         obs = analyze.get_energy()
         utils.handle_errors("calculate_energy() failed")
         return obs
+
+    def particle_energy(self, particle):
+        """
+        Calculate the non-bonded energy of a single given particle.
+
+        Parameters
+        ----------
+        particle : :class:`~espressomd.particle_data.ParticleHandle`
+
+        Returns
+        -------
+        :obj: `float`
+            non-bonded energy of that particle
+
+        """
+        energy_contribution = particle_short_range_energy_contribution(
+            particle.id)
+        return energy_contribution
 
     def calc_re(self, chain_start=None, number_of_chains=None,
                 chain_length=None):
--- a/src/core/cell_system/CellStructure.hpp
+++ b/src/core/cell_system/CellStructure.hpp
@@ -692,6 +692,58 @@ public:
 
     return particle_to_cell(p);
   }
+
+  /**
+   * @brief Run kernel on all particles inside local cell and its neighbors.
+   *
+   * @param p      Particle to find cell for
+   * @param kernel Function with signature <tt>double(Particle const&,
+   *               Particle const&, Utils::Vector3d const&)</tt>
+   * @return false if cell is not found, otherwise true
+   */
+  template <class Kernel>
+  bool run_on_particle_short_range_neighbors(Particle const &p,
+                                             Kernel &kernel) {
+    auto const cell = find_current_cell(p);
+
+    if (cell == nullptr) {
+      return false;
+    }
+
+    auto const maybe_box = decomposition().minimum_image_distance();
+
+    if (maybe_box) {
+      auto const distance_function = detail::MinimalImageDistance{box_geo};
+      short_range_neighbor_loop(p, cell, kernel, distance_function);
+    } else {
+      auto const distance_function = detail::EuclidianDistance{};
+      short_range_neighbor_loop(p, cell, kernel, distance_function);
+    }
+    return true;
+  }
+
+private:
+  template <class Kernel, class DistanceFunc>
+  void short_range_neighbor_loop(Particle const &p1, Cell *const cell,
+                                 Kernel &kernel, DistanceFunc const &df) {
+    /* Iterate over particles inside cell */
+    for (auto const &p2 : cell->particles()) {
+      if (p1.id() != p2.id()) {
+        auto const vec = df(p1, p2).vec21;
+        kernel(p1, p2, vec);
+      }
+    }
+    /* Iterate over all neighbors */
+    for (auto const neighbor : cell->neighbors().all()) {
+      /* Iterate over particles in neighbors */
+      if (neighbor != cell) {
+        for (auto const &p2 : neighbor->particles()) {
+          auto const vec = df(p1, p2).vec21;
+          kernel(p1, p2, vec);
+        }
+      }
+    }
+  }
 };
 
 #endif
--- a/src/core/cells.hpp
+++ b/src/core/cells.hpp
@@ -56,6 +56,8 @@
 
 #include "Particle.hpp"
 
+#include <boost/optional.hpp>
+
 #include <utility>
 #include <vector>
 
@@ -115,6 +117,15 @@ get_pairs_of_types(double distance, std::vector<int> const &types);
 
 /** Check if a particle resorting is required. */
 void check_resort_particles();
+
+/**
+ * @brief Get ids of particles that are within a certain distance
+ * of another particle.
+ */
+std::vector<int> mpi_get_short_range_neighbors(int pid, double distance);
+boost::optional<std::vector<int>>
+mpi_get_short_range_neighbors_local(int pid, double distance,
+                                    bool run_sanity_checks);
 
 /**
  * @brief Find the cell in which a particle is stored.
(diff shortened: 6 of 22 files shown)


jhossbach:
The most recent version can now be found here: https://github.com/jhossbach/espresso/tree/short_range_neighbors_patch
@stekajack: Most tests in the unit test produce the expected results now, but there is still a problem with PBC neighbors which I will look at during the next coding day.

jhossbach:
## Left ToDo's :

- [x] Write `unit_tests/short_range_neighbors_test.cpp` to test that neighbor cells are chosen correctly
- [x] Add python interface for the short range neighbors (see #4399)
- [x] Add python interface for the short range energy contribution
- [ ] ~Test short range neighbors via python interface (or as a cpp unit test within in `unit_tests/short_range_neighbors_test.cpp`)~
- [x] Test `particle_short_range_energy_contribution` within python interface

jhossbach:
One of the changes introduced an error with the mpi_... functions, I will check and push an updated version.

src/python/espressomd/cellsystem.pyx
```diff
@@ -129,6 +130,27 @@ cdef class CellSystem:
...
+
+        Parameters
+        ----------
+        particle : :attr:`~espressomd.particle_data.ParticleHandle.type`
+        distance : :obj:`float`
+            Pairs of particles closer than ``distance`` are found.
+
+        Returns
+        -------
+        list of tuples of :obj:`int`
```
jngrad:
```suggestion
        (N,) array_like of :obj:`int`
```

src/python/espressomd/cellsystem.pyx
```diff
@@ -129,6 +130,27 @@ cdef class CellSystem:
         handle_errors("Error in get_pairs()")
         return pairs
 
+    def get_neighbors(self, ParticleHandle particle, distance):
+        """
+        Get neighbors of a given particle up to distance
+
+        Parameters
+        ----------
+        particle : :attr:`~espressomd.particle_data.ParticleHandle.type`
```
jngrad:
```suggestion
        particle : :class:`~espressomd.particle_data.ParticleHandle`
```

src/python/espressomd/cellsystem.pyx
```diff
@@ -129,6 +130,27 @@ cdef class CellSystem:
         handle_errors("Error in get_pairs()")
         return pairs
 
+    def get_neighbors(self, ParticleHandle particle, distance):
```
jngrad:
```suggestion
    def get_neighbors(self, particle, distance):
```
Also don't forget `utils.check_type_or_throw_except()` on both `distance` and `particle` to make sure we pass to correct types to the C++ function.

src/python/espressomd/cellsystem.pxd
```diff
@@ -39,6 +39,11 @@ cdef extern from "CellStructureType.hpp":
         CELL_STRUCTURE_REGULAR "CellStructureType::CELL_STRUCTURE_REGULAR"
         CELL_STRUCTURE_NSQUARE "CellStructureType::CELL_STRUCTURE_NSQUARE"
 
+cdef extern from "particle_data.hpp":
+    ctypedef struct particle "Particle"
+    const particle & get_particle_data(int pid) except +
+
+
```
jngrad:
```suggestion
```

Note: each .pyx file includes all .pxd files. This struct and function are already declared in `particle_data.pxd`, so it's not necessary to include them here again.

src/python/espressomd/analyze.pyx
```diff
@@ -344,6 +345,23 @@ class Analysis:
...
+        """
+        Calculate the non-bonded energy of a single given particle.
+
+        Parameters
+        ----------
+        particle : :attr:`~espressomd.particle_data.ParticleHandle.type`
+
+        Returns
+        -------
+        :obj: `float` non-bonded energy of that particle
```
jngrad:
```suggestion
        :obj: `float`
            non-bonded energy of that particle
```

src/python/espressomd/analyze.pyx
```diff
@@ -344,6 +345,23 @@ class Analysis:
         utils.handle_errors("calculate_energy() failed")
         return obs
 
+    def particle_energy(self, ParticleHandle particle):
+        """
+        Calculate the non-bonded energy of a single given particle.
+
+        Parameters
+        ----------
+        particle : :attr:`~espressomd.particle_data.ParticleHandle.type`
```
jngrad:
```suggestion
        particle : :class:`~espressomd.particle_data.ParticleHandle`
```

src/python/espressomd/analyze.pyx
```diff
@@ -344,6 +345,23 @@ class Analysis:
         utils.handle_errors("calculate_energy() failed")
         return obs
 
+    def particle_energy(self, ParticleHandle particle):
```
jngrad:
```suggestion
    def particle_energy(self, particle):
```

src/core/energy.hpp
```diff
@@ -42,4 +42,16 @@ double calculate_current_potential_energy_of_system();
...
 
+/**
+ * @brief Compute short range energy of particle p
+ *
+ * Iterates through particles inside cell and neighboring cells and computes
+ * energy contribution for particle p
+ *
+ * @param pid Particle id
+ * @return Double containing short-range Coulomb and non-bonded energy of the
+ * particle
```
jngrad: Is Coulomb part of `calc_non_bonded_pair_energy()`?

src/core/energy.cpp
```diff
@@ -128,3 +128,33 @@ double observable_compute_energy() {
...
+
+  auto const p = cell_structure.get_local_particle(pid);
+
+  if (p) {
+    auto kernel = [&ret](Particle const &p, Particle const &p1,
+                         Utils::Vector3d const &vec) {
+      auto const dist = sqrt(vec.norm2());
+      auto const &ia_params = *get_ia_param(p.p.type, p1.p.type);
+      // Add energy for current particle pair to result
+      ret += calc_non_bonded_pair_energy(p, p1, ia_params, vec, dist);
```
jngrad:
```suggestion
#ifdef EXCLUSIONS
      if (not do_nonbonded(p, p1))
        return;
#endif
      auto const &ia_params = *get_ia_param(p.type(), p1.type());
      // Add energy for current particle pair to result
      ret += calc_non_bonded_pair_energy(p, p1, ia_params, vec, vec.norm());
```

src/core/cells.cpp
```diff
@@ -111,6 +112,38 @@ std::vector<PairInfo> non_bonded_loop_trace() {
...
+    return {};
+  }
+
+  std::vector<int> ret;
+  auto const cutoff2 = distance * distance;
+  auto kernel = [&ret, cutoff2](Particle const &p, Particle const &p1,
+                                Utils::Vector3d const &vec) {
+    if (vec.norm2() < cutoff2) {
+      ret.emplace_back(p1.p.identity);
+    };
```
jngrad:
```suggestion
      ret.emplace_back(p1.id());
    }
```

src/core/CellStructure.hpp
```diff
@@ -663,6 +663,58 @@ struct CellStructure {
...
+      auto const distance_function = detail::EuclidianDistance{};
+      short_range_neighbor_loop(p, cell, kernel, distance_function);
+    }
+    return true;
+  }
+
+private:
+  template <class Kernel, class DistanceFunc>
+  void short_range_neighbor_loop(Particle const &p, Cell *cell, Kernel kernel,
+                                 DistanceFunc df) {
```
jngrad:
```suggestion
  void short_range_neighbor_loop(Particle const &p, Cell *const cell,
                                 Kernel &kernel, DistanceFunc const &df) {
```

src/core/CellStructure.hpp
```diff
@@ -663,6 +663,58 @@ struct CellStructure {
...
+public:
+  /**
+   * @brief Run kernel on all particles inside cell and neighbors.
+   *
+   * @param p Particle to identify cell and neighbors
+   * @param kernel Function with (Particle, Particle, Vector)
+   * @return false if cell is not found via particle_to_cell, otherwise true
+   */
+  template <class Kernel>
+  bool run_on_particle_short_range_neighbors(Particle const &p, Kernel kernel) {
```
jngrad:
```suggestion
  bool run_on_particle_short_range_neighbors(Particle const &p,
                                             Kernel &kernel) {
```

src/core/CellStructure.hpp
```diff
@@ -663,6 +663,58 @@ struct CellStructure {
...
     return particle_to_cell(p);
   }
-};
 
+public:
+  /**
+   * @brief Run kernel on all particles inside cell and neighbors.
+   *
+   * @param p Particle to identify cell and neighbors
+   * @param kernel Function with (Particle, Particle, Vector)
```
jngrad:
```suggestion
   * @param kernel Function with signature <tt>double(Particle, Particle, Vector)</tt>
```

src/core/CellStructure.hpp
```diff
@@ -663,6 +663,58 @@ struct CellStructure {
...
+   * @return false if cell is not found via particle_to_cell, otherwise true
```
jngrad:
```suggestion
   * @return false if cell is not found via @ref particle_to_cell, otherwise true
```

src/core/CellStructure.hpp
```diff
@@ -663,6 +663,58 @@ struct CellStructure {
...
+    return true;
+  }
+
+private:
+  template <class Kernel, class DistanceFunc>
+  void short_range_neighbor_loop(Particle const &p, Cell *cell, Kernel kernel,
+                                 DistanceFunc df) {
+    /* Iterate over particles inside cell */
+    for (auto const &part : cell->particles()) {
+      if (&part == &p) {
```
jngrad: would this equality check work with `part.id() == p.id()`?

jngrad (changes requested):

src/core/CellStructure.hpp
jhossbach: I was following https://github.com/espressomd/espresso/issues/4399#issue-1072323710, but using `id()` looks more elegant to me.
Is there any difference between `.id()` and `.identity()`?

src/core/energy.hpp
jhossbach: No. Corrected in the wrap up commit

src/core/CellStructure.hpp
jngrad: Both methods are identical, we'll most likely remove one of them soon.
Actually here you could even write `if (part == p)` thanks to:
https://github.com/espressomd/espresso/blob/a91b0dc6bbbaee101eeb5d93ec36ec378483b350/src/core/Particle.hpp#L408-L410

src/core/cells.cpp
```diff
@@ -28,6 +28,7 @@
 
 #include "Particle.hpp"
 #include "communication.hpp"
+#include "energy_inline.hpp"
```
jngrad:
```suggestion
```

src/core/CellStructure.hpp
```diff
@@ -663,6 +663,61 @@ struct CellStructure {
...
-};
 
+public:
+  /**
+   * @brief Run kernel on all particles inside cell and neighbors.
+   *
+   * @param p Particle to identify cell and neighbors
+   * @param kernel Function with signature <tt>double(Particle, Particle,
+   * Vector)</tt>
+   * @return false if cell is not found via @ref particle_to_cell, otherwise
```
jngrad:
```suggestion
   * @return false if cell is not found via @c particle_to_cell, otherwise
```
sorry, I forgot that private methods couldn't be referenced in Doxygen

jngrad (changes requested):
Looks good to me! Once the remaining proposed changes are in, we can move forward with the merge.

jhossbach:
There is a problem with getting the right distances when using more than one MPI rank, this needs to be adressed first

jngrad:
@RudolfWeeber could you please have a look at this PR? Somehow running a short-range loop (e.g. to get the list of neighbors) updates the ghost information, but only in the cellsystem, so we end up with a corrupted particle node cache. This is why running analysis functions that act on particle positions after particles have been moved return results based on the old position, unless we run the integrator with 0 steps.

To reproduce the issue, run the test on 3 MPI ranks. This triggers a segfault due to a corrupted particle node cache.

When replacing
```c++
  if (pnode == this_node) {
    assert(cell_structure.get_local_particle(part));
    return *cell_structure.get_local_particle(part);
  }
```
by
```c++
  if (pnode == this_node) {
    auto const p_ptr = cell_structure.get_local_particle(part);
    if (p_ptr) {
      return *p_ptr;
    }
  }
```
in `src/core/particle_data.cpp:get_particle_data()`, the segfault moves to the particle insertion code. When commenting out the block
```c++
  if (cell_structure.get_resort_particles()) {
    cells_update_ghosts(global_ghost_flags());
  }
```
that was introduced by this PR in `src/core/cells.cpp:mpi_get_short_range_neighbors_local()`, the segfault persists.

The more fundamental issue here is that doing
```python
system.part.all().pos = np.random.random((20, 3)) * system.box_l
```
does **not** update the particle properties, but instead queues the change until the next integration loop. So at the minimum one has to do
```python
system.integrator.run(0, recalc_forces=False)
```
before attempting to update particle positions again. This also explains why the `ParticleHandle` property `node` still returns the old MPI rank after moving the particle to a different MPI domain, until the next integration.

RudolfWeeber:
Please try `on_observable_calc()` from event.cpp rather than the manual resort call at the beginning of the neigbor search.

RudolfWeeber:
The resort which was already present means that particles can move from one node to an other. Then the particle node cache needs to be invalidated, which is done in the above mentioned funciton.

jhossbach:
The assertion in the `check_neighbors` function fails when running with 5 and 7 MPI cores. Also, we should implement a runtime check to ensure that the distance is less than the cell size, otherwise particles that are within distance but not in the neighborlist of that cell will not be included

jhossbach:
I added a sanity check to prevent looking for particles for distances further than the cell size.
I also adapted the unit test, can you have another look @jngrad?

jngrad:
The `HybridDecomposition` cell system is now partially supported thanks to @pkreissl. This was achieved by iterating over the N-square cells and injecting the local regular decomposition cell as a neighbor. This means that in a system with ions on a regular decomposition cell system and colloids on a N-square cell system, ions can detect any colloid without any distance restriction, however colloids can only detect ions in the regular cell of the same MPI node. I tried adding the regular cell neighbors on the local node to the local N-square cell, but then I get double counting since the regular cell neighbors already contain the N-square cells. To fix that, one would have to create temporary vectors in the hybrid decomposition constructor. However, there is another issue: since the colloid cell has no geometric boundary, a colloid fixed in space can potentially detect all ions in regular cell of MPI rank 0 at time step 0, and all ions in regular cell of MPI rank 1 at time step 1, if the N-square cells are resorted. To fully support `HybridDecomposition`, one probably would have to add all regular cells as neighbors of the N-square cells.

jngrad (approved):

PR #4402, opened 2021-12-07
New particle access syntax

### LINKED ISSUE #4389: Re-work Python particle access
The goal is to get rid of the ambiguities concerning access to particle ranges by id slicing.

Due to the fact that python uses `[]` for both list and dictionary access, the semantics of particle access is currently unclear. Does `system.part[:10]` refer to particles with id <10 or to the ten particles with lowest ids?

## Proposed solution:
* Where ever possible, references should be used: `p = system.part.add(...)`
* An individual particle can be accessed by id using `system.part.by_id(id)`, where `id` is the numerical particle id. This will return a `ParticleHandle`
* A slice of particles can be accessed via an iterable of ids with `system.part.by_ids(id_list)`, which returns a `ParticleSlice`
* `system.part.all()` returns a slice containing all particles
* The order in which particles will be provided when iterating over a slice will stay fixed, once the slice has been instanced. I.e, `for pos, vel in zip(system.part.all().pos, system.part.all().v)` will produce the expected result.
* The `[]`-operator on `system.part` will be removed
* Selection by criterion (either using `key=value` or a lambda) can be done with the existing `system.part.select()`

This is a non-silent interface change. Users are encouraged to avoid access to particles via their ids in any case.

christophlohrmann:
Fixes #4389 

Description of changes:
- The `[]`-operator on `system.part` was removed
- An individual particle can be accessed by id using `system.part.by_id(id)`, where `id` is the numerical particle id. This returns a `ParticleHandle`
- A slice of particles can be accessed via an iterable of ids with `system.part.by_ids(id_list)`, which returns a `ParticleSlice`
- `system.part.all()` returns a slice containing all particles
- The order in which particles will be provided when iterating over a slice will stay fixed, once the slice has been instanced. I.e, `for pos, vel in zip(system.part.all().pos, system.part.all().v)` will produce the expected result.
- Selection by criterion (either using `key=value` or a lambda) can be done with the existing `system.part.select()`

### DIFF
--- a/doc/sphinx/inter_bonded.rst
+++ b/doc/sphinx/inter_bonded.rst
@@ -41,7 +41,7 @@ Note that alternatively to particle handles, the particle's ids can be
 used to setup bonded interactions. For example, to create a bond between the
 particles with the ids 12 and 43::
 
-    system.part[12].add_bond((fene, 43))
+    system.part.by_id(12).add_bond((fene, 43))
 
 .. _Distance-dependent bonds:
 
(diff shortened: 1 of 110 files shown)


jngrad (approved):
Thanks for the hard work!

PR #4404, opened 2021-12-10
CUDA code maintenance

jngrad:
Partial fix for #2628 and #4219

Description of changes:
- simplify CUDA kernels and GPU interface code
   - remove superfluous GPU-related macros and unused code
   - around 300 lines of CUDA code were deleted
- remove GPU magnetostatics global variables
   - they are now shared pointers in the Cython interface
- rewrite `SystemInterface` framework
   - document framework
   - remove unused methods
   - remove broken past-the-end device memory pointers (access to device memory with the wrong alignment is dangerous)
   - throw an error when requesting a particle list conversion from AoS to SoA for features that are not compiled in
   - unit test `EspressoSystemInterface` implementation
- improve separation of concerns:
   - `magnetostatics.pyx` no longer knows about the `dipole` global nor the associated enum values
   - header files of GPU long-range methods no longer expose implementation details
   - narrow includes of `cells.hpp`, `thermostat.hpp`, `rotation.hpp` to the strict minimum
      - the majority of indirect includes were unnecessary, exposed global variables and increased compile time
      - where necessary, functions were moved to different files and global variables `this_node` and `cell_structure` were forwarded by function argument
- bugfixes:
   - fix device memory leak in GPU Barnes-Hut (can fill up device memory quickly when creating, using and then deleting a Barnes-Hut actor in a loop)
   - GPU dipolar direct sum and Barnes-Hut used to emit a runtime error message, but `handle_errors()` was missing in the Cython interface

### DIFF
--- a/src/core/EspressoSystemInterface.hpp
+++ b/src/core/EspressoSystemInterface.hpp
@@ -27,6 +27,16 @@
 
 #include <cstddef>
 
+/**
+ * @brief CUDA implementation of @ref SystemInterface.
+ *
+ * When data is synchronized between host and device memory, a subset
+ * of the @ref Particle struct is copied from each particle on the host
+ * to the corresponding @ref CUDA_particle_data struct on the device via
+ * @ref EspressoSystemInterface::gatherParticles(). Once the transfer is
+ * complete, the particle AoS on the device is copied (or "split") to
+ * a SoA via @ref EspressoSystemInterface::split_particle_struct().
+ */
 class EspressoSystemInterface : public SystemInterface {
 public:
   EspressoSystemInterface() = default;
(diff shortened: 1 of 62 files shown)


src/core/EspressoSystemInterface.hpp
```diff
@@ -20,136 +20,102 @@
...
     m_gpu |= m_needsTorqueGpu;
-    if (m_gpu)
-      enableParticleCommunication();
-    return m_needsTorqueGpu;
+    enableParticleCommunication();
   };
 #endif
 
+  float *eGpu() override {
+    // cast pointer to struct of floats to array of floats
```
reinaual:
```suggestion
    // cast pointer from struct of floats to array of floats
```

reinaual (dismissed):

reinaual (approved):

PR #4420, opened 2022-01-11
Write additional unit tests

jngrad:
Description of changes:
- improve unit testing of utils, core and script interface
- run MPI-capable unit tests in parallel
- improve separation of concerns
    - significantly narrow inclusion of `MpiCallbacks.hpp`
    - run include what you use
- fix minor regressions in the core

### DIFF
--- a/testsuite/python/npt_thermostat.py
+++ b/testsuite/python/npt_thermostat.py
@@ -19,6 +19,7 @@
 import unittest as ut
 import unittest_decorators as utx
 import espressomd
+import tests_common
 import numpy as np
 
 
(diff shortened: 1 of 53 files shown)


testsuite/python/npt_thermostat.py
```diff
@@ -112,6 +112,36 @@ def reset_particle_and_box():
         self.assertTrue(np.all(np.not_equal(force4, force5)))
         self.assertTrue(np.all(np.not_equal(boxl4, boxl5)))
 
+    def test_02__direction(self):
+        """Test for NpT constrained in one direction."""
+
+        system = self.system
+
+        ref_force_dir = [75.3778877, 35.1267878, 21.5026612]
+        ref_boxl_dir = [0.7669542, 0.7809505, 0.719799]
```
reinaual: where do these magic values come from?

jngrad: They came from the simulation. I don't really like that solution, because these values are tied to the RNG and need to be adapted every time we update the test RNG seed. It's hard to come up with an analytical solution for a NpT simulation; even macroscopic values like the compressibility are not stable (7f7dda58447f9f9).

The original plan was to check the fixed directions wouldn't change, but as we can see in `ref_boxl_dir` the three values changed. The particle force could maybe serve as a proxy since the first component is much larger than the two other, but I doubt this is reliable, as it depends on the particle and RNG seed. I rewrote that test using `integrator_npt_stats.py`  (c77ae1d8b7fd15) and now we get the expected behavior.

reinaual (approved):

PR #4439, opened 2022-01-31
Remove code duplication in actors

jngrad:
Description of changes:
- actors should check for parameter values, while `espressomd.utils` should check for parameter names and pretty-print error messages (single responsibility principle)

### DIFF
--- a/src/python/espressomd/interactions.pyx
+++ b/src/python/espressomd/interactions.pyx
@@ -22,6 +22,7 @@ cimport cpython.object
 import collections
 
 include "myconfig.pxi"
+from . import utils
 from .utils import is_valid_type
 from .utils cimport check_type_or_throw_except
 from .script_interface import ScriptObjectRegistry, ScriptInterfaceHelper, script_interface_register
@@ -56,16 +57,11 @@ cdef class NonBondedInteraction:
 
         # Or have we been called with keyword args describing the interaction
         elif len(args) == 0:
+            utils.check_required_keys(self.required_keys(), kwargs.keys())
+            utils.check_valid_keys(self.valid_keys(), kwargs.keys())
             # Initialize default values
             self._params = self.default_params()
             self._part_types = [-1, -1]
-
-            # Check if all required keys are given
-            for k in self.required_keys():
-                if k not in kwargs:
-                    raise ValueError(
-                        "At least the following keys have to be given as keyword arguments: " + self.required_keys().__str__())
-
             self._params.update(kwargs)
             self.validate_params()
         else:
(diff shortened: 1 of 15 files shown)


src/python/espressomd/interactions.pyx
```diff
@@ -1696,7 +1686,7 @@ class BondedInteraction(ScriptInterfaceHelper):
                 params.update(kwargs)
                 self.validate_params(params)
                 super().__init__(*args, **params)
-                self._check_keys(params.keys(), check_required=True)
+                utils.check_valid_keys(self.valid_keys(), kwargs.keys())
```
jngrad, line 1689: Note: it was not necessary to check required keys here because `super().__init__(*args, **params)` already raises a `RuntimeError` if a required key is missing.

testsuite/python/ek_charged_plate.py
```diff
@@ -164,11 +164,18 @@ def test(self):
...
 
         # Test error when trying to change ekin parameters after initialisation
-        ek._params.update({'agrid': 3,
-                           'T': 0.01})
         with self.assertRaises(RuntimeError):
+            ek._params.update({'agrid': 3, 'T': 0.01})
             ek._set_params_in_es_core()
 
+        # Check errors from the constructor
+        with self.assertRaisesRegex(ValueError, r"The following keys have to be given as keyword arguments: "
```
christophlohrmann, line 172: should D not pop up in the 'have to be given' list?

christophlohrmann (commented):
In general I am not a fan of checking the error msg of every actor in a test.
If the msg is tested once, I would be fine with it as subsequent tests just check if the required/valid set of each actor is correct.
We can keep it, though

testsuite/python/ek_charged_plate.py
jngrad: `r"\[.+\]"` means any sequence of characters, which includes D.

christophlohrmann: oh ok

christophlohrmann (approved):

jngrad:
> In general I am not a fan of checking the error msg of every actor in a test.
> If the msg is tested once, I would be fine with it as subsequent tests just check if the required/valid set of each actor is correct.

When inheritance is involved (e.g. `IntegratorHandle` resp. `Actor`), then I test the error message of only one derived class (e.g. `SteepestDescent` resp. `ReactionField`). For classes that do not inherit from anything, they must all be tested. When a class uses the new utility function in two places (e.g. constructor and params setter), we need to test both. We had issues in 4.1.4 where LB, ELC and DH did not properly handle error messages, and as a result users couldn't know when they were using partially initialized actors.

There are concrete plans for rewriting the ReactionMethods, Electrostatics, Magnetostatics, LB and EK actors as script interface objects. Some of these have already been implemented in draft PRs and local branches. Once deployed, we will no longer need to check parameters against `required_keys()`. This will also give us the freedom of having multiple mutually-exclusive required arguments (see e.g. walberla LB actor), which is something we currently cannot express with `required_keys()`.

reinaual (approved):

PR #4444, opened 2022-02-04
Particle property getters

### LINKED ISSUE #4430: Implementation-independent interface for particle properties
Goal: The Particle class should have interface funciton to access particle properties. The actual detial of the storage should be hidden from the customers of the Particle class.

## Example
```C++
  auto const& pos() const { return r.p; };
  auto & pos() { return r.p; };
  
  const auto & mass() const { return p.mass; };
  auto & mass() { return p.mass; };
```

## Notes

* Both, the writeable reference and the `const auto& ... const` version are needed. The first, so that `p.pos() = ...;` can be used, the latter to allolw access on a `const Particle`.
* Property names shoudl match those from the Python interface, where possible
* Flags should have `is_...()` getters and `set_...(bool)` setters, to allow us to merge several flags into a bitset, later
* When properties have constexpr values, if the relevant feature is not compiled in (e.g., `constexpr const mass =1`, getter function shoudl be declared `constexpr const` as well, and settes shoudl throw.
* The actual roll-out of the new interface throughout the code base will not be done immediately, to limit the number of merge conflicts

christophlohrmann:
Fixes #4430 

Description of changes:
- introduce getters for particle proprties
- start rolling out, but keep old access compatible

### DIFF
--- a/src/core/Particle.hpp
+++ b/src/core/Particle.hpp
@@ -29,9 +29,20 @@
 
 #include <boost/serialization/vector.hpp>
 
+#include <bitset>
 #include <cstdint>
 #include <stdexcept>
 #include <vector>
+
+namespace detail {
+inline void check_axis_idx_valid(int const axis) {
+  assert(axis >= 0 and axis <= 2);
+}
+
+inline bool get_nth_bit(uint8_t const bitfield, int const bit_idx) {
+  return bitfield & (1u << bit_idx);
+}
+} // namespace detail
 
 enum : uint8_t {
   ROTATION_FIXED = 0u,
@@ -57,9 +68,9 @@ enum : uint8_t {
 struct ParticleParametersSwimming {
   /** Is the particle a swimmer. */
   bool swimming = false;
-  /** Constant velocity to relax to. */
-  double f_swim = 0.;
   /** Imposed constant force. */
+  double f_swim = 0.;
+  /** Constant velocity to relax to. */
   double v_swim = 0.;
   /** Flag for the swimming mode in a LB fluid.
    *  Values:
@@ -421,11 +432,6 @@ struct Particle { // NOLINT(bugprone-exception-escape)
 private:
   BondList bl;
 
-public:
-  auto &bonds() { return bl; }
-  auto const &bonds() const { return bl; }
-
-private:
 #ifdef EXCLUSIONS
   /** list of particles, with which this particle has no non-bonded
    *  interactions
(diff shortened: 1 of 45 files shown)


src/core/Particle.hpp
```diff
@@ -434,21 +436,148 @@ struct Particle { // NOLINT(bugprone-exception-escape)
...
+  auto &pos() { return r.p; }
+  auto const &v() const { return m.v; }
+  auto &v() { return m.v; }
+  auto const &force() const { return f.f; }
+  auto &force() { return f.f; }
+
+  bool is_ghost() const { return l.ghost; }
+  void set_ghost(bool const ghost_flag) { l.ghost = ghost_flag; }
+  auto &p_old() { return l.p_old; }
+  auto const &p_old() const { return l.p_old; }
```
jngrad: I would personally prefer to name getters `v()` resp. `p_old()` as `vel()` resp. `old_pos()`. @RudolfWeeber @reinaual, any objection?

src/core/Particle.hpp
```diff
@@ -434,21 +436,148 @@ struct Particle { // NOLINT(bugprone-exception-escape)
...
   }
-
-  std::vector<int> const &exclusions() const {
-#ifdef EXCLUSIONS
-    return el;
+  bool can_rotate_around(int const axis) const {
+    detail::check_axis_idx_valid(axis);
+    std::bitset<3> rot_flags(p.rotation);
+    return rot_flags[axis];
+  }
```
jngrad: After reading the documentation of `std::bitset`, I'm having second thoughts on using it in ESPResSo. Although it's currently not used in a way that could lead to performance issues in this PR, it could become an issue in the future if used improperly.

For example, it's reserving 8 bytes internally instead of 1 byte for `uint8_t` which is really unfortunate and prevents us from using it as a field in a struct. It also generates assembly code for a `throw std::out_of_range("...")` which introduces overhead every time the method is called, because there is no guarantee at runtime that the value of `axis` is less or equal to 2 (the assertion in `detail::check_axis_idx_valid(axis)` is ignored by the compiler in a release build). This could become an issue if we call this method from the integrator loop when rolling out this PR to the rest of the codebase. @RudolfWeeber @reinaual do you think this is an issue?

Here are live examples: [getter](https://godbolt.org/z/M3qPxfMjT) and [setter](https://godbolt.org/z/rWbGcPv1G).

src/core/Particle.hpp
```diff
@@ -434,21 +436,148 @@ struct Particle { // NOLINT(bugprone-exception-escape)
...
+  auto &gamma() { return p.gamma; }
+#ifdef ROTATION
+  auto const &gamma_rot() const { return p.gamma_rot; }
+  auto &gamma_rot() { return p.gamma_rot; }
+#endif // ROTATION
+#endif // THERMOSTAT_PER_PARTICLE
+#ifdef EXTERNAL_FORCES
+  bool has_fixed_coordinates() const {
+    std::bitset<4> fix_flags(p.ext_flag);
+    fix_flags >> 1;
```
jngrad: Line 544 doesn't generate assembly code. You probably meant to write `fix_flags >>= 1u;`

Consider adding idef'ed test cases in `src/core/unit_tests/Particle_test.cpp` to check for these new methods. Most of them currently have no code coverage (codecov doesn't consider them as code because they are never called), so we don't know if they actually work.

jngrad (commented):
Besides suggestions, looks good to me. Thanks for the hard work!

src/core/Particle.hpp
reinaual: I was already blaming @christophlohrmann that he is doing the exact opposite regarding the naming than what he talked about in the code-quality meeting. But it was not his fault, he just followed the rule 
> Property names should match those from the Python interface, where possible

I would actually really appreciate to get rid of these single-character names.

When reading through the exposed code-snippet from this comment: what is the reason that we have a proper getter and setter for the `ghost`-property but not for the remaining properties?

src/core/Particle.hpp
reinaual: For the exceptions it works analogously to a vector access. The exception code is only generated when using the `set()/test()` methods and can be avoided by using the `[]` operator.

The allocation-size is a real pity and also compiler dependent [(gcc)](https://github.com/gcc-mirror/gcc/blob/1ce5395977f37e8d0c03394f7b932a584ce85cc7/libstdc%2B%2B-v3/include/std/bitset#L77) Not sure if this makes it feasable. Maybe we should just go back to the old-school manual bitshifting solution for now and rethink this when everything is hidden behind the interface.

jngrad: Alright, we should replace `rot_flags.set(axis, rot_flag);` by `rot_flags[axis] = rot_flag;` to remove the branching and exception.

There is still 3 times more assembly code and CPU flags being set and discarded with the non-throwing bitset solution, compared to the manual bitshifting solution proposed in the live examples above (using both Clang and GCC). My preference would be to use bitwise operations even if they are less readable. @RudolfWeeber would you like to weigh in?

src/core/Particle.hpp
jngrad: dev meeting summary: keep `v()` for consistency with the python interface, but rename `p_old()` to something more explicit like `old_pos()`

src/core/Particle.hpp
jngrad: dev meeting summary:
* avoid throwing code due to overhead
* avoid bitset due to size
* bitarray implementation were considered, but it might increase runtime in debug and coverage builds
* manual bitshift is the preferred solution due to the minimal overhead
   * you can use enum values from an `enum : int { ROT_X = 0, ROT_Y = 1, ROT_Z = 2 };` for the axis when calling `can_rotate_around(int const axis)`
   * just keep in mind that in C++, bit operations involving types smaller than `int` (e.g. `unit8_t`) are implicitly cast to `signed int`, therefore you probably won't be able to get anything better than the following implementation:
   ```c++
   bool can_rotate_around(int const axis) const {
     return rotation & (1<<axis);
   }
  ```

src/core/Particle.hpp
RudolfWeeber: If we are talking about the p_old that gets set at resort or verlet rebuild, I'd suggest `pos_at_last_resort` or the like.
If it is the rattle p_old it should probably be `pos_prev_time_Step`.

Also, once the new interface is used everywhere, we should renmae the Parrticle substrcut members such as `ParticleProperties p`.

jngrad (approved):

PR #4445, opened 2022-02-11
Make CellStructureType an enum class.

### LINKED ISSUE #4428: Add checks for cell system compatibility
Goal: Algorithms that rely on specific cell structures should check for them either at the point where the method is called or (if that's impractical) in a sanity check on integration start
(In preparation for #3662)

Steps:
* In `cellStructure.hpp` make the `type` an enum class
* Move the definition of this enum calss to a separate file `CellStructureType.hpp`. This let's you use it elsehere without giving access to the full cell structure header.
* Add a check for the domain decompositoin type to the following places (list may be incompletet). Use Espresso's RuntimeErrorMSG() mechanism where possible.
* Avoid giving various places in Espresso access to the cell_structure, if they do not need it. Preferrably pass in the cell structure type as an argument.

Relevant places 

* CPU P3M, ELC, ICC (see force_calc() in forces.cpp)
* Dipolar P3m and DLC
* The Scafacos electrostatics and magnetostatics solvers
* LB CPU (propagate and particle coupling. Dispatched from the main integraiotn loop (integrate.cpp) and fore_calc() (forces.cpp), respectively)
* The external potential featues (?)
* The Scafacos electrostatics and magnetostatics solvers* Anything else that uses regular grids
Err on the side of caution.

Places that proabbly don't care about the decompositoin:
* cluster analysis
* collision detection
* reaction ensemble
* all GPU methods
* dipolar direct summation,
* Debye-Hueckel

Unclear:
* MMM1D

pkreissl:
Partial fix for #4428

Description of changes:
- `CellStructureType` is now an `enum class`.
- Removed access to the cell system in several files by storing the information about the current `CellStructureType` in the global variable `LocalBox<double> local_geo`.

### DIFF
--- a/src/core/CellStructure.cpp
+++ b/src/core/CellStructure.cpp
@@ -242,15 +242,18 @@ void CellStructure::resort_particles(int global_flag) {
 }
 
 void CellStructure::set_atom_decomposition(boost::mpi::communicator const &comm,
-                                           BoxGeometry const &box) {
+                                           BoxGeometry const &box,
+                                           LocalBox<double> &local_geo) {
   set_particle_decomposition(std::make_unique<AtomDecomposition>(comm, box));
   m_type = CellStructureType::CELL_STRUCTURE_NSQUARE;
+  local_geo.set_cell_structure_type(m_type);
 }
 
 void CellStructure::set_regular_decomposition(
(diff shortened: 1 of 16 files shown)


jngrad (approved):
LGTM, thanks!

PR #4450, opened 2022-02-14
Verlet unit test

### LINKED ISSUE #4306: unit test VerletCriterion
currently, the VerletCriterion relies on the onn-bonded interaction data structures being set up, so it is nont trivial to test.
This could be changed by making the retrieval of non-bondex cutoffs for type pairs injectable.

* Change 
`auto const max_cut = get_ia_param(p1.p.type, p2.p.type)->max_cut;`
to a new function `get_nonbonded_cutoff(int type_i, int type_j);`

* Make VerletCriterion a class template taking <typename PairCutoffGetter> so the pair cutoff retrieval can be injected.

* Write a unit test for VerletCriterion with a mock pair cutoff getter to test all possibel cases

mebrito:
Fixes #4306 

Description of changes:
- Make VerletCriterion a class template
- Add unit test for VerletCriterion

### DIFF
--- a/src/core/forces.cpp
+++ b/src/core/forces.cpp
@@ -187,8 +187,8 @@ void force_calc(CellStructure &cell_structure, double time_step, double kT) {
 #endif
       },
       maximal_cutoff(), maximal_cutoff_bonded(),
-      VerletCriterion{skin, interaction_range(), coulomb_cutoff, dipole_cutoff,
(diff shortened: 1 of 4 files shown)


jngrad (commented):
LGTM

reinaual (approved):

PR #4451, opened 2022-02-15
Rewrite reaction methods as script interface classes

### LINKED ISSUE #4398: Reaction methods should be script interface classes
Many Cython classes in the reaction methods have the following design:

```cython
cdef class ReactionEnsemble(ReactionAlgorithm):

    cdef unique_ptr[CReactionEnsemble] REptr

    def __init__(self, *args, **kwargs):
        self._params = {"kT": 1, "exclusion_radius": 0, **kwargs}
        self.REptr.reset(new CReactionEnsemble(int(self._params["seed"])))
        self.RE = < CReactionAlgorithm * > self.REptr.get()
```

The `ScriptInterface` framework provides a convenient interface to store a shared pointer to a core class. It has many advantages:
* no need to manage a unique pointer, this is done by the script interface
* no need to store a raw pointer to the parent class of the managed object
* no need to re-declare (and maintain) the core class in the .pxd file
* the Cython class can be converted to a regular Python class
* the class can be checkpointed

Instructions can be found in the developer guide, section [Using handles to core and script interface objects](https://github.com/espressomd/espresso/wiki/Software-design#using-handles-to-core-and-script-interface-objects).

While this can be done without breaking the API, there is an opportunity to replace the setters and getters by properties, which would e.g. allow doing `reaction.volume = box_l**3` and `print(reaction.volume)` instead of doing `reaction.set_volume(box_l**3)` resp. `print(reaction.get_volume())`.

pm-blanco:
Fixes #4398

Description of changes:
- Converted the reaction methods Cython interface to a Python interface using the `ScriptInterface` framework
- Now reaction methods only take keyword arguments
- Adapted tutorials, samples, tests and benchmarks
- Fixed minor documentation issues


jngrad (commented):
Looks good to me, thanks for the hard work!

samples/reaction_ensemble.py
```diff
@@ -105,14 +105,20 @@
...
-print("reaction 0 has acceptance rate: ", RE.get_acceptance_rate_reaction(0))
-print("reaction 1 has acceptance rate: ", RE.get_acceptance_rate_reaction(1))
+print(
+    "reaction 0 has acceptance rate: ",
+    RE.get_acceptance_rate_reaction(
+        reaction_id=0))
+print(
+    "reaction 1 has acceptance rate: ",
+    RE.get_acceptance_rate_reaction(
+        reaction_id=1))
```
reinaual, line 123: @jngrad Probably out of scope for this PR but do we want to enforce the usage of f-strings at some point?

src/python/espressomd/reaction_ensemble.py
```diff
@@ -0,0 +1,567 @@
...
+
+        Returns
+        -------
+        mean : :obj:`float`
+            Mean excess chemical potential.
+        error : :obj:`float`
+            Standard error of the mean.
+        """
+
+        def do_block_analysis(samples, N_blocks=16):
```
reinaual:
```suggestion
        def do_block_analysis(samples, N_blocks):
```
This default-value was never necessary in this code, so we can just remove it

testsuite/python/reaction_methods.py
```diff
@@ -0,0 +1,241 @@
...
+            method.get_volume(),
+            self.system.volume(),
+            delta=1e-10)
+        method.set_volume(volume=1.)
+        self.assertAlmostEqual(method.get_volume(), 1., delta=1e-10)
+        self.assertEqual(method.get_non_interacting_type(), 100)
+        method.set_non_interacting_type(type=9)
+        self.assertEqual(method.get_non_interacting_type(), 9)
+        if isinstance(method, espressomd.reaction_ensemble.ConstantpHEnsemble):
+            self.assertAlmostEqual(method.constant_pH, -10., delta=1e-10)
```
reinaual: just out of interest, is there a reason to set the default pH value to `-10.0`? In an ideal world, shouldn't that be a parameter of the constructor?

reinaual (dismissed):
LGTM, @jngrad you can decide if it is necessary to work on any of my comments.

samples/reaction_ensemble.py
jngrad: I'll add that task to my upcoming coding-day wrap-up PR.

testsuite/python/reaction_methods.py
jngrad: `-10` is the historical sentinel value for uninitialized class members in reaction method classes. In this specific case, I think there is no obstacle to making the constant pH a part of the core class constructor. @pm-blanco would you agree with that change? From the users perspective, it would still be possible to do a titration with `for pH in range(10): RE.constant_pH = pH`, but it would be mandatory to provide a sensible value in `RE = ConstantpHEnsemble(kT=1.5, exclusion_radius=0.8, seed=12, constant_pH=6)`.

If there is a constraint on the pH range, we should actually write a setter function in the core class that throws an exception if the pH is outside the valid range. I don't think that's the case though, because negative pH values are well-defined in non-aqueous solvents and obey the same thermodynamic equations.

pm-blanco: I agree with that change, I think that actually it is desired that one has to provide the pH when doing an instance of the constant pH ensemble, since one cannot use this method without setting the pH value. You are right @jngrad, it should not be any a priori restriction for the pH value.

src/python/espressomd/reaction_ensemble.py
jngrad: Done.

testsuite/python/reaction_methods.py
jngrad: Done: `kT`, `exclusion_radius` and `constant_pH` are are now required arguments of the core class constructors.

reinaual (approved):

PR #4457, opened 2022-02-25
Implement Lees-Edwards boundary conditions

### LINKED ISSUE #2051: Lees Edwards Rewrite tracking ticket
The current Lees-Edwards implementation creates duplicte pairs in the short range loop. It produces incorrect short range interactions even for an offset of 0.

The idea is to use a columnar cell and node grid, such that the neighborship relations do not change when the Lees Edwards offset changes.
I.e., the length of the cells in shear directoin is the box length. In the other directions, the cell size is determined by the interaction range and skin as for domain decomposition.

The shift of ghost particles' positions will be removed, also for conventional domain decomposition.

Lees Edwards needs the following:
* enforce columnar cell and node grid
* Distance function must refelect LE. Will be used both, in the short range loop and everywhere else
* Veleocity update, when particles cross the shear plane box boundary.

In the code
* Change the signature of get_mi_vector to take particles rather than positions, wherever possible
* Check cell geometry when setting LE offset to values other tthan 0.
* Remove position shift from ghost particles and adapt short range loop to use get_mi_vector
* Implement LE-aware velocity difference function for use in DPD, thermalized bonds, etc.
* Re-introduce code for velocity change when particles cross shear boundary

### LINKED ISSUE #2083: Modify `get_mi_vector` to support Lees-Edwards
Currently `get_mi_vector` (from `grid.hpp`) takes two points as arguments and returns their minimum image convention distance. When Lees-Edwards BC are used this distance needs to be modified. The
calculation of the modified distance needs more information, which is currently not available in the function.
So two things need to be done:

 - Determine what information is needed for the LE distance calculation
 - Adapt the signature of `get_mi_vector` and the call sites to pass that information

I *think* it is sufficient to pass the image box to the function but please check. Also a appropiate
choice of default parameters might make it unnecessary to touch all call sites. Also please get
back to after you have completed step one so we can choose good types for the interface.

I believe we agreed that @bindgens1 will do this task.

jngrad:
Fixes #2051
Fixes #2083


jngrad:
Note: this PR blocks the `Particle` interface rollout, the HybridDecomposition PR and the `walberla` branch.

src/script_interface/lees_edwards/LinearShear.hpp
```diff
@@ -0,0 +1,59 @@
...
+#include <memory>
+
+namespace ScriptInterface {
+namespace LeesEdwards {
+
+class LinearShear : public Protocol {
+public:
+  LinearShear()
+      : m_protocol{
+            new ::LeesEdwards::ActiveProtocol{::LeesEdwards::LinearShear()}} {
```
reinaual, line 38: For all the shear-cases, is there a reason to use `new` instead of `std::make_shared` for the construction of the core object? Intuitively i would expect it to work out of the box and to be more explicit than this solution.

jngrad: This `new` construction is actually quite common in the script interface. Unfortunately, the reasons for this design decision were not documented, as far as I know. The resulting `shared_ptr` is much smaller since it only contains a pointer instead of the pointee, and the rules for deallocation are slightly different (I think the pointee is deallocated immediately when the reference counter is zero, regardless of the value of the weak pointer counter).

reinaual (approved):

src/script_interface/lees_edwards/LinearShear.hpp
reinaual: To me it is not clear where this difference comes from, especially because we just construct a shared-pointer from it. I would have expected that `std::make_shared` does exactly that, but I never bothered to check what `std::make_shared` actually does internally. I always had the impression that it is designed to replace `new` with a shared-ownership, which does not seem to be the whole story.

jngrad: The `std::shared_ptr` constructor takes a raw pointer as argument, and takes ownership of it, i.e. we are no longer allowed to call `free()` or `delete` on the original raw pointer. The resulting shared pointer is lightweight, since it only wraps a raw pointer. Whether the pointee is allocated with `new` or `malloc` is not really relevant, but the `new` syntax is a bit more readable. This strategy has some overhead due to the double dereference, but then we can have a lightweight `std::vector<std::shared_ptr<T>>` that is easy to resize and resort.

The `std::shared_ptr` copy constructor is designed to work with `std::make_shared`, which embeds the pointee. For our use case in the script interface, I think it's safe to replace the `new` syntax with the `std::make_shared` everywhere. It also bugged me a few weeks ago when I realized we were using the `new` syntax in the script interface.

PR #4460, opened 2022-03-02
Roll out new particle interface

jngrad:
Description of changes:
- use the new `Particle` setters and getters in the core, except in electrostatics/magnetostatics
- hide `ext_flag` and `rotation` implementation details from the script interface
- make `lees_edwards` and `bond_breakage` submodules of `src/core` for consistency with `src/script_interface`
- make the bond breakage global variable static and unit test bond breakage actions
- cleanup comments (and clang-format tags) in source files modified by this rollout


src/python/espressomd/particle_data.pxd
```diff
@@ -46,7 +40,6 @@ cdef extern from "particle_data.hpp":
         int    mol_id
         int    type
         double mass
-        stdint.uint8_t rotation
```
jngrad, line 49: Note: I intentionally left these particle substructs ctypedefs.
They can be removed in the next coding day (#4461).

src/core/bond_breakage/actions.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+namespace BondBreakage {
+
+// Delete Actions
+struct DeleteBond {
+  int particle_id;
+  int bond_partner_id;
+  int bond_type;
+  std::size_t hash_value() const {
+    std::size_t seed = 3875;
```
christophlohrmann, line 36: How do the seed magic values come about?

src/core/integrators/stokesian_dynamics_inline.hpp
```diff
@@ -37,12 +37,12 @@ inline void stokesian_dynamics_propagate_vel_pos(const ParticleRange &particles,
...
-    p.r.p += p.m.v * time_step;
+    p.pos() += p.v() * time_step;
 
     // Perform rotation
-    auto const norm = p.m.omega.norm();
-    if (norm != 0) {
-      auto const omega_unit = (1. / norm) * p.m.omega;
+    auto const norm = p.omega().norm();
+    if (norm != 0.) {
+      auto const omega_unit = (1. / norm) * p.omega();
```
christophlohrmann, line 45: is there normalize() for our Vectors? If not would be nice to have

src/core/particle_data.cpp
```diff
@@ -780,9 +775,16 @@ constexpr Utils::Vector3d ParticleProperties::rinertia;
 #endif
 
 #ifdef ROTATION
-void set_particle_rotation(int part, int rot) {
-  mpi_update_particle_property<uint8_t, &ParticleProperties::rotation>(part,
-                                                                       rot);
+void set_particle_rotation(int part, Utils::Vector3i const &flag) {
```
christophlohrmann, line 778: Is Vector3bool possible?

src/core/rotation.cpp
```diff
@@ -146,56 +146,56 @@ void propagate_omega_quat_particle(Particle &p, double time_step) {
...
 
   /* and rescale quaternion, so it is exactly of unit length */
-  auto const scale = p.r.quat.norm();
+  auto const scale = p.quat().norm();
   if (scale == 0) {
-    p.r.quat = Utils::Quaternion<double>::identity();
+    p.quat() = Utils::Quaternion<double>::identity();
   } else {
-    p.r.quat /= scale;
+    p.quat() /= scale;
```
christophlohrmann, line 157: p.quat().normalize() ?

christophlohrmann (approved):
LGTM. Big fan of the new flag handling and the removal of pointers in virtualsites

src/core/particle_data.cpp
jngrad: I guess so, but Cython doesn't like templated classes, so we would need to first define a `Vector3b` in the core, copy paste its struct definition in `utils.pxd`, and then deal with `bool` being a keyword and not a type in Cython.

christophlohrmann: ok. too much hassle for 2 usecases

src/core/integrators/stokesian_dynamics_inline.hpp
jngrad: The method exists, but then we are computing the norm twice: once to decide whether it is zero, and once to normalize the vector. Clang 13.0 and GCC 11.2 do not realize that the same value can be re-used, even at `-O3`. Since this is a hot loop, performance is more important than readability.

src/core/rotation.cpp
jngrad: Same issue as in the integrator, only worse: the quaternion normalize method uses `boost::qvm` under the hood, therefore that single C++ line, which takes 50 lines of assembly code, becomes 700 lines of assembly code with `p.quat().normalize()`.

src/core/bond_breakage/actions.hpp
jngrad: There is a discussion in [walberla/walberla#160](https://i10git.cs.fau.de/walberla/walberla/-/issues/160) (this actually where this code comes from). Some people like to use the golden ratio sequence because of its entropy, but the choice is somewhat arbitrary. Since the number of bonds is bounded to a small value, probably to the order of 100 for very large simulations, entropy is not as much an issue as it was for walberla. The seed differ between the two structs to avoid collisions, since both can appear in the same `std::vector`.

reinaual (approved):

PR #4463, opened 2022-03-04
Lees-Edwards with collision detection and breakable bonds

bindgens1:
Description of changes:
- There is a new test that combines LE and collision detection and breakable bonds.


bindgens1:
Things are working also with the collision detection.
Naming still depends on #4464.

bindgens1:
There is a problem with the test because we can't guarantee the order of created VS in the test cases.
is there a common way to solve this?

bindgens1:
Can someone review this / have a look?

jngrad (approved):

PR #4464, opened 2022-03-04
Bond breakage interface

jngrad:
Description of changes:
- accept bond objects as key for bond breakage
- document the bond breakage feature
- encapsulate `ScriptObjectMap` checkpointing logic


doc/sphinx/advanced_methods.rst
```diff
@@ -118,6 +118,63 @@ The following limitations currently apply for the collision detection:
 * The ``"bind at point of collision"`` approach cannot handle collisions
   between virtual sites
 
+.. _Deleting bonds when particles are pulled:
```
bindgens1: "pulled apart" or "exceed a critical distance"

doc/sphinx/advanced_methods.rst
```diff
@@ -118,6 +118,63 @@ The following limitations currently apply for the collision detection:
...
+
+Deleting bonds when particles are pulled
+----------------------------------------
+
+With this feature, bonds between particles can be deleted automatically
+when a critical bond extension is reached. This is useful to simulate
+weak bonds between particles.
```
bindgens1: Not only week bonds (all bonded interactions can be used). Maybe "reversible bonds"? or simply "braking bonds".

doc/sphinx/advanced_methods.rst
```diff
@@ -118,6 +118,63 @@ The following limitations currently apply for the collision detection:
...
+
+The bond breakage action is specified for individual bonds via the system
+:attr:`~espressomd.system.System.bond_breakage` attribute.
+
+Several modes are available:
+
+* ``"revert_center_bond"``: delete a bond from the first particle
```
bindgens1: analog to the second item here: this can be used together with the collision detection: bond centers.

doc/sphinx/advanced_methods.rst
```diff
@@ -118,6 +118,63 @@ The following limitations currently apply for the collision detection:
...
+* ``"revert_vs_bond"``: delete a virtual bond, can be used together with
+  the :ref:`collision detection<Creating bonds when particles collide>`
+  feature (configured in mode ``"bind_at_point_of_collision"``) to
+  reversibly create and delete a bond
```
bindgens1: with "bind_at_point_of_collision" also the real particles are bonded together via, e.g., a virtual bond. If you want to revert this, you need to run both modes of the breaking bonds.

doc/sphinx/advanced_methods.rst
```diff
@@ -118,6 +118,63 @@ The following limitations currently apply for the collision detection:
...
+* ``"none"``: cancel an existing bond breakage specification
+
```
bindgens1, line 138: We should add:
Please note that virtual sites are not automatically removed from the simulation. A simultaneous use of `"bind_at_point_of_collision"`` and breakable bonds, will sequentially always increase the particle number. If you want to remove virtual sites, you need to do so manually.

src/python/espressomd/bond_breakage.py
```diff
@@ -17,14 +17,41 @@
...
 #
 from .script_interface import script_interface_register, ScriptObjectMap, ScriptInterfaceHelper
+from . import interactions
 
 
 @script_interface_register
 class BreakageSpec(ScriptInterfaceHelper):
+    """
+    Specifications for bond breakage.
+    See :ref:`Deleting bonds when particles are pulled` for more details.
```
bindgens1: pulled apart or exceed a certain distance.

doc/sphinx/advanced_methods.rst
bindgens1: Also: There is no special treatment implemented for treating the energy release/ consumed by bond removal. This can lead to physical inconsistencies.

bindgens1 (dismissed):

doc/sphinx/advanced_methods.rst
```diff
@@ -118,6 +118,87 @@ The following limitations currently apply for the collision detection:
...
+  used to create or delete a bond between two real particles
+* ``"revert_vs_bond"`` mode for breakable bonds together with
+  ``"bind_at_point_of_collision"`` mode for collision detection:
+  used to create or delete virtual sites (the implicitly created
+  bond between the real particles isn't affected)
+
+Please note that virtual sites are not automatically removed from the
+simulation, therefore the particle number will increase. If you want to
+remove virtual sites, you need to do so manually.
+
```
RudolfWeeber, line 204: Please clarify which vs you are talking about

doc/sphinx/advanced_methods.rst
```diff
@@ -118,6 +118,87 @@ The following limitations currently apply for the collision detection:
...
+With this feature, bonds between particles can be deleted automatically
+when the bond length exceeds a critical distance. This is used to model
+breakable bonds.
+
+The bond breakage action is specified for individual bonds via the system
+:attr:`~espressomd.system.System.bond_breakage` attribute.
+
+Several modes are available:
+
+* ``"revert_center_bond"``: delete a bond from the first particle
```
RudolfWeeber: I'm not happy with the names. The word center should not appear in the name for the methdo that just deletes the bond. `delete_bond` or similar should do. Please take into account that this might be used independently from the collision detection.  The revert_vs_bond action actually does not only that but also deletes bonds between the related physcal paricles. Maybe `revert_vs_non_sliding_contact` if we want something independent from collision detection, or otherwise the name from the core `rvert_bind_at_point_of_collision`.

RudolfWeeber (dismissed):

doc/sphinx/advanced_methods.rst
bindgens1: I agree with "delete bond".

For the latter I would fall back on the name of the collision detection. People can use this independently, but might not do so. And I think it is hard to imagine what it means when you are not using the collision detection.

Also another thought: What happens when a user connect vs manually without connecting the center particles...

doc/sphinx/advanced_methods.rst
```diff
@@ -118,6 +118,90 @@ The following limitations currently apply for the collision detection:
...
+    system.cell_system.skin = 0.4
+    system.time_step = 0.1
+    system.min_global_cut = 2.
+
+    h1 = espressomd.interactions.HarmonicBond(k=0.01, r_0=0.4)
+    h2 = espressomd.interactions.HarmonicBond(k=0.01, r_0=0.5)
+    system.bonded_inter.add(h1)
+    system.bonded_inter.add(h2)
+    system.bond_breakage[h1] = espressomd.bond_breakage.BreakageSpec(
+        breakage_length=0.5, action_type="revert_center_bond")
```
itischler: `"revert_center_bond"` is non of the modes mentioned above or below.

src/python/espressomd/interactions.pyx
```diff
@@ -2850,13 +2850,27 @@ def get_bonded_interaction_type_from_es_core(bond_id):
...
     Represents the bonded interactions list.
 
     Individual interactions can be accessed using ``BondedInteractions[i]``,
     where ``i`` is the bond id.
+
+    Methods
+    -------
+    remove():
+        Remove a bond from the list.
+        This is a noop if the bond does not exist.
```
itischler: according to wikipedia it is called NOP, no-op or NOOP

src/python/espressomd/script_interface.pyx
```diff
@@ -406,14 +406,10 @@ class ScriptObjectMap(ScriptObjectRegistry):
 
     def remove(self, key):
         """
-        Removes the element with the given key
+        Remove the element with the given key.
+        This is a noop if the key does not exist.
```
itischler: NOP, no-op or NOOP

itischler (changes requested):
change `revert_center_bond` to `delete_bond` in the .rst docu.

doc/sphinx/advanced_methods.rst
jngrad: Thanks!

itischler (dismissed):

reinaual (dismissed):

reinaual (approved):

PR #4468, opened 2022-03-09
CMake maintenance

jngrad:
Description of changes:
- rewrite CMake logic for Python tests
   - checkpoint tests: convert the body of the list loop into a function `checkpoint_test()` and call it once for each combination of features (improves readability and separation of concerns)
   - tests called multiple times with a different number of MPI ranks no longer require creating a new copy of the original file, instead a different CTest name is used
   - Python tests that used to depend on CMake variable substitution are now configured with a class factory; the CMake variable is passed as a command-line argument to the test file and the corresponding test class is generated on-the-fly
- improve code coverage
   - checkpoint tests coverage data is no longer discarded at random, instead it is accumulated over all feature combinations


pm-blanco:
Now that I see this PR, I noticed that it exists one instruction  of cmake `make test` whose utility is unclear to me. Its wording seems confusing to me compared to `make check`, which is the one that I use to check that all tests run properly.  When I run `make test`, it runs 379 tests, some of them failing on my computer.  However, the continuous integration of github passes in my PRs so I guess that this command is somehow deprecated.

jngrad:
> Now that I see this PR, I noticed that it exists one instruction of cmake `make test` whose utility is unclear to me. Its wording seems confusing to me compared to `make check`, which is the one that I use to check that all tests run properly. When I run `make test`, it runs 379 tests, some of them failing on my computer. However, the continuous integration of github passes in my PRs so I guess that this command is somehow deprecated.

`make test` is indeed deprecated. It is a built-in feature of CMake, se we cannot remove it. It doesn't support file dependencies, therefore several tests will fail in a clean build folder where `make check` was never executed before. More details in #4370.

It also ignores the CMake variable `CTEST_ARGS`, which is why `make test` runs the tests one after the other, while `make check` runs them in parallel batches.

pm-blanco:
> > Now that I see this PR, I noticed that it exists one instruction of cmake `make test` whose utility is unclear to me. Its wording seems confusing to me compared to `make check`, which is the one that I use to check that all tests run properly. When I run `make test`, it runs 379 tests, some of them failing on my computer. However, the continuous integration of github passes in my PRs so I guess that this command is somehow deprecated.
> 
> `make test` is indeed deprecated. It is a built-in feature of CMake, se we cannot remove it. It doesn't support file dependencies, therefore several tests will fail in a clean build folder where `make check` was never executed before. More details in #4370.
> 
> It also ignores the CMake variable `CTEST_ARGS`, which is why `make test` runs the tests one after the other, while `make check` runs them in parallel batches.

@jngrad thank you for the explanation!

jngrad:
This PR is needed for the upcoming electrostatics/magnetostatics refactoring PR.

reinaual (dismissed):

reinaual (approved):

PR #4469, opened 2022-03-09
reaction_methods: new optional keyword exclusion_radius_per_type

### LINKED ISSUE #4437: Reaction methods should take an exclusion radius per particle type
Many systems  have particles with rather different radii (e.g. colloidal systems). If one wants to simulate such systems including a reaction that creates/destroys particles (e.g. acid/base reactions), it is necessary more than one exclusion radius.  The reason is that there is no good choice for the exclusion radius for such system since:
i) if one sets the exclusion radius to match the size of the small particles then one can get insertions inside the bigger ones.
ii) if one sets the exclusion radius to match the size of the big particles then the smaller ones effectively occupy more than they should. If you have many small particles (imagine systems with explicit salt) then one basically has no space to insert particle while the system actually does have it. 

I like the solution proposed by @jngrad  to simply change the exclusion radius to a dictionary (similar to `default_charges`) where the user defines one exclusion radius per particle type. If we all agree in this change, I can take care of implementing it.

pm-blanco:
Fixes #4437

Description of changes:
- The  keyword `exclusion_radius`  has been renamed to `exclusion_range`.
- If the keyword `exclusion_radius` is provided, a keyerror warns the user that `exclusion_radius` is obsolete and the current equivalent keyword is `exclusion_range`
- The reaction methods now take `exclusion_radius_per_type` as an optional argument.
- The exclusion range between two particles is calculated as their sum of their radii in `exclusion_radius_per_type`, following the Lorentz-Berthelot combining rule. 
- If the exclusion radius of one particle type is not defined, the value of the parameter provided in  `exclusion_range` is used by default.
- If the value in `exclusion_radius_per_type` is equal to 0, then the exclusion range of that particle type with any other particle is 0.
- A new section have been added in `doc/sphinx/advanced_methods.rst` to explain how to couple the reaction methods to molecular dynamics, including the new features of this PR.
- New unit tests have been included in `src/core/reaction_methods/tests/ReactionAlgorithm_test.cpp` to check the behavior of `exclusion_radius_per_type`
-  The loop over all particles in `ReactionAlgorithm::check_exclusion_range`  now breaks if one particle is found within the exclusion range of the inserted particle. 
- All benchmarks, samples, tests and tutorials have been changed accordingly. 

I checked the computational efficiency using the benchmark in `maintainer/benchmarks/mc_acid_base_reservoir.py` and this PR improves the MC  computational by a 15%  compared to the current  python branch (MC timing: 1.268e-04 vs 1.395e-04, this PR vs current python, respectively). Probably the gain could be even higher for dense systems.


jonaslandsgesell:
May I suggest to make the exclusion radius a required keyword, forcing the user to explicitly consider what parameter makes sense. Ideal reactions are rarely executed in Espresso which is why I would not use a default value which will only work for ideal systems, when not only MC moves are performed (but also MD moves).

pm-blanco:
> May I suggest to make the exclusion radius a required keyword, forcing the user to explicitly consider what parameter makes sense. Ideal reactions are rarely executed in Espresso which is why I would not use a default value which will only work for ideal systems.

@jonaslandsgesell  the exclusion radius is still a required keyword. An ideal system system is only assumed if the user provides `exclusion_radius={}`.

If the user does not provide an exclusion radius for a particle type, such particles are assumed to have an exclusion radius equal to 0, which I agree that could be potentially dangerous. Since in principle the user can create new particle types after the reaction methods have been initialized, the only alternative way that I can think of is to check it during the particle insertion and throw an error if a particle type without an exclusion radius is found. In practice, this would force the users to define an exclusion radius for all particle types that exist within their simulations. However, such change would force us to make the exclusion radius a required keyword for the widom insertion too, which would  be annoying since it is not actually necessary.

pm-blanco:
@jonaslandsgesell  I have figured out a way of checking that all the existing particle types have a defined excluded radius without having to force the user to define an exclusion radius when using the widom insertion method. I belive that the issue is now solved.

jonaslandsgesell:
@pm-blanco thanks for having a look. 

Short disclaimer: I did not have a deeper look at your changes due to time :) maybe you already have a great solution but here are some things which popped up in my mind when thinking about the exclusion radii.

* Would it make sense to require the user to supply a dictionary contains at least one numeric exclusion radius (either 0 or something else) so that he had to consider which number(s) make sense here? This would explicitly disallow an empty dictionary.

* Would it make sense to provideexclusion radii for pairs of particle types? Something like Exclusion_radius[1,1]=0.9, exclusion_radius[1,2]=2 etc, due to the various different combination rules of the LJ potential? https://en.m.wikipedia.org/wiki/Combining_rules

pm-blanco:
>    * Would it make sense to require the user to supply a dictionary contains at least one numeric exclusion radius (either 0 or something else) so that he had to consider which number(s) make sense here? This would explicitly disallow an empty dictionary.

I can add a check during the initialization of the reaction methods to prevent the user to provide an empty dict. It feels a little bit redundant to me  since the code already checks that the user has defined an exclusion_radius for all existing particle types during the particle insertion, effectively dealing with this issue. It could however be a nice way to make the user aware in a early stage that he/she should define one exclusion radius per particle type.
 
>    * Would it make sense to provideexclusion radii for pairs of particle types? Something like Exclusion_radius[1,1]=0.9, exclusion_radius[1,2]=2 etc, due to the various different combination rules of the LJ potential? https://en.m.wikipedia.org/wiki/Combining_rules

In my opinion, using that pair definition of the exclusion radius can complicate its setting up unnecessarily. Since there are only a few established combination rules, I believe that it would be rather more convenient to  define an optional keyword e.g.  `combination_rule="Lorentz-Berthelot"` and deal with it internally.  By default could be the Lorentz-Berthelot  combing rule since (up to my knowledge) is the more commonly used one.

src/core/reaction_methods/ReactionAlgorithm.cpp
```diff
@@ -375,16 +375,65 @@ void ReactionAlgorithm::hide_particle(int p_id) const {
...
-void ReactionAlgorithm::check_exclusion_radius(int p_id) {
-  if (exclusion_radius == 0.) {
-    return;
+void ReactionAlgorithm::check_exclusion_range(int inserted_particle_id) {
+
+  auto const &inserted_particle = get_particle_data(inserted_particle_id);
+
+  /* Check the excluded radius of the inserted particle */
+
+  if (check_exclusion_radius_defined(inserted_particle.type())) {
```
jngrad: Here and elsewhere, you can probably replace this call to `check_exclusion_radius_defined()` by `exclusion_radius_per_type.count(inserted_particle.type()) == 0` and then delete the `check_exclusion_radius_defined()` method.

src/core/reaction_methods/ReactionAlgorithm.hpp
```diff
@@ -47,17 +48,30 @@ struct StoredParticleProperty {
 class ReactionAlgorithm {
 
 public:
-  ReactionAlgorithm(int seed, double kT, double exclusion_radius)
+  ReactionAlgorithm(int seed, double kT, double exclusion_range,
+                    std::unordered_map<int, double> exclusion_radius_per_type)
```
jngrad: Here and in the constructor of the other reaction method classes, you can pass `exclusion_radius_per_type` by const reference. This avoid making multiple copies when the derived classes forward the argument to the base class.

doc/sphinx/advanced_methods.rst
```diff
@@ -1772,6 +1772,25 @@ be converted to :math:`K_c` as
 where :math:`p^{\ominus}=1\,\mathrm{atm}` is the standard pressure.
 Consider using the python module pint for unit conversion.
 
+Coupling Reaction Methods to Molecular Dynamics
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The Monte Carlo (MC) sampling of the reaction can  be coupled with a configurational sampling using Molecular Dynamics (MD).
+For non-interacting systems this coupling is not an issue, but for interacting systems the insertion of new particles
+can lead to inestabilities in the MD integration ultimately leading to a crash of the simulation.
```
jngrad: typo: inestabilities -> instabilities

doc/sphinx/advanced_methods.rst
```diff
@@ -1772,6 +1772,25 @@ be converted to :math:`K_c` as
...
+
+This integration instabilities can be avoided by defining a distance around the particles which already exist in the system
+where new particles will not be inserted, which is defined by the required keyword `exclusion_range`.
```
jngrad: Here and later in the text: in Sphinx, use double backticks to format text in teletype, i.e. ``` ``exclusion_range`` ```. The single backtick version formats the text in italic.

src/core/reaction_methods/tests/ReactionEnsemble_test.cpp
```diff
@@ -46,7 +46,7 @@ namespace espresso {
 // ESPResSo system instance
 std::unique_ptr<EspressoSystemStandAlone> system;
 } // namespace espresso
-
+std::unordered_map<int, double> exclusion_radius_per_type;
```
jngrad: Here and in the other C++ tests, you can remove the global variable `exclusion_radius_per_type` and pass the symbol `{}` as argument of the constructors to default-create an empty map. Using a mutable global variable for the map is risky, because if a test case inserts data in the map, the testsuite outcome becomes dependent on the order in which the test cases are executed (the order can be controlled from e.g. the command line).

The `system` global is unfortunately an exception to this rule, because we have to guarantee there is only one ESPResSo system.

src/script_interface/reaction_methods/ReactionEnsemble.hpp
```diff
@@ -39,9 +39,20 @@ class ReactionEnsemble : public ReactionAlgorithm {
...
   void do_construct(VariantMap const &params) override {
+    std::unordered_map<int, double> exclusion_radius_per_type;
+
+    for (auto const &item : get_value_or<std::unordered_map<int, Variant>>(
+             params, "exclusion_radius_per_type", {})) {
+      exclusion_radius_per_type[item.first] =
+          get_value<double>(get_value<std::unordered_map<int, Variant>>(
+                                params, "exclusion_radius_per_type")
+                                .at(item.first));
+    }
```
jngrad: Here and in all the other `do_construct()` methods, how about replacing this logic by the following one-liner?
```c++
   get_map<int, double>(get_value_or<std::unordered_map<int, Variant>>(
      params, "exclusion_radius_per_type", {}))
```

src/script_interface/reaction_methods/ReactionAlgorithm.hpp
```diff
@@ -56,19 +56,38 @@ class ReactionAlgorithm : public AutoParameters<ReactionAlgorithm> {
...
+         {"exclusion_radius_per_type",
+          [this](Variant const &v) {
+            std::unordered_map<int, double> exclusion_radius_per_type;
+
+            for (auto const &item :
+                 get_value<std::unordered_map<int, Variant>>(v)) {
+              exclusion_radius_per_type[item.first] = get_value<double>(
+                  get_value<std::unordered_map<int, Variant>>(v).at(
+                      item.first));
+            }
```
jngrad: How about the following?
```c++
  get_map<int, double>(get_value_or<std::unordered_map<int, Variant>>(
      v, "exclusion_radius_per_type", {}))
```

testsuite/python/reaction_methods.py
```diff
@@ -148,17 +148,17 @@ def check_reaction_parameters(reactions, parameters):
...
         method = espressomd.reaction_ensemble.ConstantpHEnsemble(
-            kT=1.5, exclusion_radius=0.8, seed=12, constant_pH=10)
-        self.check_interface(method, kT=1.5, exclusion_radius=0.8, gamma=1.2)
+            kT=1.5, exclusion_range=0.8, seed=12, constant_pH=10)
+        self.check_interface(method, kT=1.5, exclusion_range=0.8, gamma=1.2)
 
         # Widom insertion
         method = espressomd.reaction_ensemble.WidomInsertion(kT=1.6, seed=12)
-        self.check_interface(method, kT=1.6, exclusion_radius=0., gamma=1.)
+        self.check_interface(method, kT=1.6, exclusion_range=0., gamma=1.)
```
jngrad: consider adding an argument `exclusion_radius_per_type={1: 0.1}` in each class and `check_interface` call, and then after line 73 adding the following:
```python
        self.assertEqual(list(method.exclusion_radius_per_type.keys()), [1])
        self.assertAlmostEqual(
            method.exclusion_radius_per_type[1],
            exclusion_radius_per_type[1],
            delta=1e-10)
        method.exclusion_radius_per_type = {2: 0.2}
        self.assertEqual(list(method.exclusion_radius_per_type.keys()), [2])
        self.assertAlmostEqual(
            method.exclusion_radius_per_type[2],
            0.2,
            delta=1e-10)
```
This will provide code coverage for the setters and getters and guarantee they have been properly implemented.

The goal of `self.assertEqual(list(method.exclusion_radius_per_type.keys()), [2])` is to make sure the key exists before we check the corresponding exclusion radius with ` self.assertAlmostEqual()` (if the key is not in the dict, we want the test case to halt with a meaningful error message, instead of halting the entire testsuite with a `KeyError` exception).

jngrad (commented):
Thanks for this work! Please find a few suggestions below.

src/script_interface/reaction_methods/ReactionAlgorithm.hpp
```diff
@@ -56,19 +56,38 @@ class ReactionAlgorithm : public AutoParameters<ReactionAlgorithm> {
...
+
+            RE()->exclusion_radius_per_type = exclusion_radius_per_type;
+          },
+          [this]() {
+            std::unordered_map<int, Variant> exclusion_radius_per_type;
+            for (auto const &item : RE()->exclusion_radius_per_type) {
+              exclusion_radius_per_type[item.first] = item.second;
+            }
+            return exclusion_radius_per_type;
```
jngrad: how about the following one-liner?
```c++
return make_map(RE()->exclusion_radius_per_type);
```

src/script_interface/reaction_methods/ReactionAlgorithm.hpp
pm-blanco: Maybe I don't quite get how it should work but this line
`{"exclusion_radius_per_type",
          [this](Variant const &v) {
            RE()->exclusion_radius_per_type =   get_map<int, double>(get_value_or<std::unordered_map<int, Variant>>(
      v, "exclusion_radius_per_type", {}));
          }`
gives me the following error:
     In file included from /home/pablo/Desktop/espresso_pablo/src/script_interface/reaction_methods/ConstantpHEnsemble.hpp:23,
                 from /home/pablo/Desktop/espresso_pablo/src/script_interface/reaction_methods/initialize.cpp:24:
/home/pablo/Desktop/espresso_pablo/src/script_interface/reaction_methods/ReactionAlgorithm.hpp: In lambda function:
/home/pablo/Desktop/espresso_pablo/src/script_interface/reaction_methods/ReactionAlgorithm.hpp:76:41: error: no matching function for call to ‘get_value_or<std::unordered_map<int, boost::variant<boost::detail::variant::recursive_flag<ScriptInterface::None>, bool, int, long unsigned int, double, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::vector<int, std::allocator<int> >, std::vector<double, std::allocator<double> >, std::shared_ptr<ScriptInterface::ObjectHandle>, std::vector<boost::recursive_variant_, std::allocator<boost::recursive_variant_> >, Utils::Vector<double, 2>, Utils::Vector<double, 3>, Utils::Vector<double, 4>, std::unordered_map<int, boost::recursive_variant_, std::hash<int>, std::equal_to<int>, std::allocator<std::pair<const int, boost::recursive_variant_> > > >, std::hash<int>, std::equal_to<int>, std::allocator<std::pair<const int, boost::variant<boost::detail::variant::recursive_flag<ScriptInterface::None>, bool, int, long unsigned int, double, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::vector<int, std::allocator<int> >, std::vector<double, std::allocator<double> >, std::shared_ptr<ScriptInterface::ObjectHandle>, std::vector<boost::recursive_variant_, std::allocator<boost::recursive_variant_> >, Utils::Vector<double, 2>, Utils::Vector<double, 3>, Utils::Vector<double, 4>, std::unordered_map<int, boost::recursive_variant_, std::hash<int>, std::equal_to<int>, std::allocator<std::pair<const int, boost::recursive_variant_> > > > > > > >(const Variant&, const char [26], <brace-enclosed initializer list>)’
       v, "exclusion_radius_per_type", {}));
                                         ^
In file included from /home/pablo/Desktop/espresso_pablo/src/script_interface/auto_parameters/AutoParameter.hpp:23,
                 from /home/pablo/Desktop/espresso_pablo/src/script_interface/auto_parameters/AutoParameters.hpp:24,
                 from /home/pablo/Desktop/espresso_pablo/src/script_interface/ScriptInterface.hpp:30,
                 from /home/pablo/Desktop/espresso_pablo/src/script_interface/reaction_methods/SingleReaction.hpp:24,
                 from /home/pablo/Desktop/espresso_pablo/src/script_interface/reaction_methods/initialize.cpp:22:
/home/pablo/Desktop/espresso_pablo/src/script_interface/get_value.hpp:301:3: note: candidate: ‘template<class T> T ScriptInterface::get_value_or(const VariantMap&, const string&, const T&)’
 T get_value_or(VariantMap const &vals, std::string const &name,
   ^~~~~~~~~~~~
/home/pablo/Desktop/espresso_pablo/src/script_interface/get_value.hpp:301:3: note:   template argument deduction/substitution failed:
In file included from /home/pablo/Desktop/espresso_pablo/src/script_interface/reaction_methods/ConstantpHEnsemble.hpp:23,
                 from /home/pablo/Desktop/espresso_pablo/src/script_interface/reaction_methods/initialize.cpp:24:
/home/pablo/Desktop/espresso_pablo/src/script_interface/reaction_methods/ReactionAlgorithm.hpp:76:7: note:   cannot convert ‘v’ (type ‘const Variant’ {aka ‘const boost::variant<boost::detail::variant::recursive_flag<ScriptInterface::None>, bool, int, long unsigned int, double, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::vector<int, std::allocator<int> >, std::vector<double, std::allocator<double> >, std::shared_ptr<ScriptInterface::ObjectHandle>, std::vector<boost::recursive_variant_, std::allocator<boost::recursive_variant_> >, Utils::Vector<double, 2>, Utils::Vector<double, 3>, Utils::Vector<double, 4>, std::unordered_map<int, boost::recursive_variant_, std::hash<int>, std::equal_to<int>, std::allocator<std::pair<const int, boost::recursive_variant_> > > >’}) to type ‘const VariantMap&’ {aka ‘const std::unordered_map<std::__cxx11::basic_string<char>, boost::variant<boost::detail::variant::recursive_flag<ScriptInterface::None>, bool, int, long unsigned int, double, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::vector<int, std::allocator<int> >, std::vector<double, std::allocator<double> >, std::shared_ptr<ScriptInterface::ObjectHandle>, std::vector<boost::recursive_variant_, std::allocator<boost::recursive_variant_> >, Utils::Vector<double, 2>, Utils::Vector<double, 3>, Utils::Vector<double, 4>, std::unordered_map<int, boost::recursive_variant_, std::hash<int>, std::equal_to<int>, std::allocator<std::pair<const int, boost::recursive_variant_> > > > >&’}
       v, "exclusion_radius_per_type", {}));
       ^
make[2]: *** [src/script_interface/CMakeFiles/ScriptInterface.dir/build.make:323: src/script_interface/CMakeFiles/ScriptInterface.dir/reaction_methods/initialize.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:6649: src/script_interface/CMakeFiles/ScriptInterface.dir/all] Error 2
make: *** [Makefile:141: all] Error 2

jngrad: oh sorry, that was a copy-paste error, it should look more like this:
```c++
{"exclusion_radius_per_type",
  [this](Variant const &v) {
    RE()->exclusion_radius_per_type =
      get_map<int, double>(get_value<std::unordered_map<int, Variant>>(v));
  },
```

jngrad (approved):

PR #4470, opened 2022-03-10
LB coupling for RegularDecomposition without ghost shifts.

### LINKED ISSUE #4432: `non_bonded_loop_trace` seems to be broken
`system.cell_system.non_bonded_loop_trace()` doesn't seem to work properly. Given two particles in the system (`p0` and `p1`) it returns two entries for interactions `(0, 1)` and `(1, 0)`. Talking to @RudolfWeeber it seems that this function is supposed to only return one of these.
More importantly still, for certain combinations of `box_l` and particle positions one of the two distance vectors does not obey the minimum image convention. I observed this for a number of different position vector combinations, as an example see
```python
import numpy as np
import espressomd

box_l = 28.968631620212868
system = espressomd.System(box_l=[box_l, box_l, box_l])
system.time_step = 0.01
system.cell_system.skin = 0.4

p0 = system.part.add(pos=[14.276456611971197, 13.16738226901207, 8.927150084980797])
p1 = system.part.add(pos=[28.881982622730717, 0.39909834518046605, 7.269467872077882])

    epsilon=1, sigma=4, cutoff=10, shift=0)

system.integrator.run(0)

def obeys_mic(v):
    return np.all(abs(v) < box_l / 2)

# non_bonded_loop_trace_returns (id1, id2, pos1, pos2, vec21)
cs_pairs = system.cell_system.non_bonded_loop_trace()
print(cs_pairs, "\n")

# distance vectors via loop trace
core01 = cs_pairs[0][4]
core10 = cs_pairs[1][4]
# distance vectors from python
py01 = -system.distance_vec(p0, p1)
py10 = -system.distance_vec(p1, p0)

for vector in [core01, py01, core10, py10]:
    print(vector, np.linalg.norm(vector), obeys_mic(vector))
```
producing the following output
```bash
pkreissl@host:~/espresso/build$ ./pypresso minimal_mic.py 
[(0, 1, array([14.27645661, 13.16738227,  8.92715008]), array([28.88198262,  0.39909835,  7.26946787]), array([-14.60552601,  12.76828392,   1.65768221]), 0), (1, 0, array([28.88198262,  0.39909835,  7.26946787]), array([43.24508823, 13.16738227,  8.92715008]), array([-14.36310561, -12.76828392,  -1.65768221]), 0)] 

[-14.60552601  12.76828392   1.65768221] 19.470448755216818 False
[14.36310561 12.76828392  1.65768221] 19.289266119447404 True
[-14.36310561 -12.76828392  -1.65768221] 19.289266119447404 True
[-14.36310561 -12.76828392  -1.65768221] 19.289266119447404 True
```

### LINKED ISSUE #4466: Support LB coupling for RegularDecomposition without ghost shifts
Tis can either be done on the Walberla branch, or by backporting the n-square lb coupling support from the Walberla branch to Phon
(commit b9b077, 502f2, 614df). 

Start from #4452 (merge, if using the Walberla branch)

Modify the logic in lb_particle_coupling as follows
* Add a std::unordered_set `coupled_ghost_particles` 
* do not call couple_particle() on particles which are ghosts for which we also have the physical particles (use CellStructure::get_local_particle(), which will always spoint to the pysical paricle, if present. Check the ghost flag of the returend particle). Outsource the decision to a function `is_ghost_for_local_particle()`
* If we don't have the physical particle: in certain cases, we may have two ghosts for the same particle. Therefore, check if the particle id is in `coupled_ghost_particles`. If not, couple the particle nd add its id to the set. If the id is already in the set, continue without coupling.
* Alwyas couple physical particles
* In couple_particle() make the `hast_ghost==false` case the default, i.e., always call md force on all possible combinatinos of shift by one box length. The has_ghosts parameter can be removed.

pkreissl:
Fixes #4466, fixes #4432, closes #4440

Description of changes:
 - ports n-square lb coupling support from the `walberla` branch back into the `python` branch
 - merges PR #4440 and fixes CI issues along the way
 - expose some funtions in `lb_particle_coupling.hpp` to get rid of code duplication in (the fixed version of) `lb_inertialess_tracers.cpp`


RudolfWeeber:
@pkreissl, what's the status. Given that this Pr is high-risk and the feature freeze is fast aproaching.

pkreissl:
Resolving the code duplication in the LB coupling of both, regular particles and inertialess tracers, proved to be not quite as straight-forward as anticipated. I am open for suggestions for improvement.
 - In `lb_inertialess_tracers`: former `in_local_domain()` was basically the same as `in_local_halo()` of `lb_particle_coupling.cpp`
 - In the current commit `inline in_local_domain()` is defined in both `.cpp` files. Putting it into the `lb_particle_coupling.hpp` is not directly possible, as the function requires an `include "grid.hpp"`. This would lead to an indirect include of `mpi.h` in `electrokinetics_cuda.cu`, which is not permissible.

@RudolfWeeber , can you please also have an extra-close look at the coupling logic in the inertialess tracer coupling -- I think I got it right, and the tests pass, but still...

jngrad (commented):
From a C++ perspective, this looks good to me. I ran the `lb_momentum_conservation.py` test on CPU with particle coupling, `kT=0`, 500 time steps, random initial positions and velocities for 20 particles on MPI topologies `[1 1 1], [2 1 1], [3 1 1], [2 2 1], [1 2 2], [4 1 1]`. The particle forces at the end of the simulations were identical within 7 decimal places, and were also identical to the same simulation on 1 MPI node with the N-square cell system.

RudolfWeeber (dismissed):

pkreissl:
I think @christophlohrmann also wanted to have a look at this PR. What is the status?

pkreissl:
I just also ran the following tests on 1, 2, 3, 4, and 6 cores, respectively, to run them with different MPI topologies:
 - `engine_lb.py` as additional test for active particles.
 - `virtual_sites_tracers.py`; for that, I had to adapt `virtual_sites_tracers_common.py`, i.e., I used `box_l = [6, 6, 12]` and increased the number of integration steps for establishing the steady state flow field in the advection test (line 81) to 800.

All these additional test passed.

jngrad (approved):

PR #4471, opened 2022-03-14
throw error for bonded interactions if particles are at the same posi…

### LINKED ISSUE #4453: Fene bond for distance=0 and r_0 != 0 wrong
When the equilibrium distance is not 0, the force function for a particle distance of 0 should either throw or return a random force.
Currently it returns (0,0,0), which is wrong.

Currently, we have a situation where the bonded energy is non-zero but the bonded pressure tensor is zero, which is an excellent sink for user time.

The harmoinc bond needs to be checked, too.

christophlohrmann:
…tion

Fixes #4453 

Description of changes:
- if particle dist is 0 and the equilibrium extension of the bond is nonzero, throw error (because direction of the force cannot be determined)
- affects FENE, harmonic and quartic bond

### DIFF
--- a/src/core/bonded_interactions/fene.hpp
+++ b/src/core/bonded_interactions/fene.hpp
@@ -27,6 +27,7 @@
  */
 
 #include "config.hpp"
+#include "errorhandling.hpp"
 
 #include <utils/Vector.hpp>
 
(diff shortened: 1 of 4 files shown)


jngrad (approved):
LGTM

PR #4473, opened 2022-03-14
Particle property getters

### LINKED ISSUE #4461: Roll out new particle property getters in the Python interface
The Python interface should reflect the recent changes to the `Particle` struct. The goal is to adapt all property getters to the new syntax, except for `ext_flag` and `rotation` which can be ignored (they rely on special getters outside the class definition). The setters don't need to change. This task is suitable for a coding day and can be done in about half an hour. No C++ knowledge required.

The `src/python/espressomd/particle_data.pyx` file currently looks like this:

```cython
cdef class ParticleHandle:
    property type:
        def __get__(self):
            self.update_particle_data()
            return self.particle_data.p.type
```

It should look like this:

```cython
cdef class ParticleHandle:
    property type:
        def __get__(self):
            self.update_particle_data()
            return self.particle_data.type()  # <<< now we simply call the struct method
```

To achieve this, the `src/python/espressomd/particle_data.pxd` needs to be changed from this:

```cython
cdef extern from "particle_data.hpp":
    ctypedef struct particle_properties "ParticleProperties":
        int    identity
        int    type

    ctypedef struct particle "Particle":
        particle_properties p
        Vector3d calc_dip()
```

to this:

```cython
cdef extern from "particle_data.hpp":
    ctypedef struct particle_properties "ParticleProperties":
        int    identity  # <<< 'type' is gone, 'identity' can be removed too

    ctypedef struct particle "Particle":
        particle_properties p
        Vector3d calc_dip()
        int type()  # <<< struct method is declared, do the same for all other properties
```

The same changes have to be applied to 2-3 lines in `src/python/espressomd/visualization_mayavi.pyx`.

Once all getters are replaced, `particle_data.pxd` should no longer need to declare the particle substructs.

christophlohrmann:
Fixes #4461 

Description of changes:
- Update all properties to the new getters except gamma and VS properties

### DIFF
--- a/src/core/particle_data.hpp
+++ b/src/core/particle_data.hpp
@@ -373,32 +373,6 @@ int number_of_particles_with_type(int type);
 // This is needed, because cython does not support conditional compilation
 // within a ctypedef definition
 
-#ifdef LB_ELECTROHYDRODYNAMICS
-inline Utils::Vector3d get_particle_mu_E(Particle const *p) {
-  return p->mu_E();
-}
-#endif
-
-#ifdef ROTATION
-inline Utils::Vector3d get_particle_omega_body(Particle const *p) {
-  return p->omega();
-}
-
-inline Utils::Vector3d get_particle_torque_body(Particle const *p) {
-  return p->torque();
-}
-
-inline Utils::Quaternion<double> get_particle_quat(Particle const *p) {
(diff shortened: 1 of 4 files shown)


jngrad (approved):
Thanks!

PR #4476, opened 2022-03-14
Always overwrite values in std::unordered_map objects

### LINKED ISSUE #4448: std::unordered_map objects discard values with duplicate key
There are several places in the core where `std::unordered_map` and `std::map` are assigned values in a way that actually discards the value, e.g. [`src/utils/include/utils/Cache.hpp#L106-L123`](https://github.com/espressomd/espresso/blob/7e29f9052e710fe1ebf0f5d2a8076b32921fbc6a/src/utils/include/utils/Cache.hpp#L106-L123). The [`std::unordered_map<Key,T>::emplace(std::pair<Key, T>)`](https://en.cppreference.com/w/cpp/container/unordered_map/emplace) documentation says the following (emphasis mine):
> Inserts a new element into the container constructed in-place with the given args **if there is no element with the key in the container**.
> [...] The element may be constructed even if there already is an element with the key in the container, **in which case the newly constructed element will be destroyed immediately**.

Same issue with `std::unordered_map<Key,T>::insert()` and the `std::map` equivalents. Old values are never overwritten, which can lead to undefined behavior due to stale references, e.g. in the script interface object manager. To overwrite an existing value, one has to use the `std::unordered_map<Key,T>::operator[]` operator. C++17 introduced `std::unordered_map<Key,T>::insert_or_assign()` which overwrites the existing object, and `std::unordered_map<Key,T>::try_emplace()` which preserves the original object.

MWE:
```c++
#include <unordered_map>
int main() {
  std::unordered_map<int, int> dict;
  dict.emplace(1, 2);
  dict.emplace(1, 3);
  return dict[1];
}
```
Output:
```
Program returned: 2
```

Affected code in ESPResSo: `Utils::Cache` (used in the particle cache), `Utils::NumeratedContainer` (used in MPI callbacks), `ScriptInterface::AutoParameters` (multiple parameter definitions don't overwrite each other) and `ScriptInterface::GlobalContext` (stale references stored in MPI ranks >=1). Affects all ESPResSo versions since 4.0.0.

jngrad:
Fixes #4448

Affected code (all ESPResSo versions since 4.0.0):
- `Utils::Cache` (used for particle caches)
- `Utils::Factory` (used in the script interface registry)
- `Utils::NumeratedContainer` (used in MPI callbacks)
- `ScriptInterface::AutoParameters` (multiple parameter definitions didn't overwrite each other)
- `ScriptInterface::GlobalContext` (risk of stale references being stored in MPI ranks >=1)
- `bond_breakage` (multiple specifications didn't overwrite each other)

### DIFF
--- a/src/core/PartCfg.hpp
+++ b/src/core/PartCfg.hpp
@@ -29,12 +29,11 @@
  *
  * This class implements cached access to all particles in a
  * particle range on the head node.
- * This implementation fetches all particles to
- * the head node on first access. Updates of the particle data are
- * triggered automatically on access. The data in the cache
- * is invalidated automatically on_particle_change, and then
+ * This implementation fetches all particles to the head node on first access,
+ * which invalidates all existing particle pointers. Updates of the particle
+ * data are triggered automatically on access. The data in the cache
+ * is invalidated automatically by @ref on_particle_change, and then
  * updated on the next access.
- *
  */
 class PartCfg {
   /** The particle data */
--- a/src/core/bond_breakage/bond_breakage.cpp
+++ b/src/core/bond_breakage/bond_breakage.cpp
@@ -43,7 +43,7 @@ namespace BondBreakage {
 static std::unordered_map<int, std::shared_ptr<BreakageSpec>> breakage_specs;
 
 void insert_spec(int key, std::shared_ptr<BreakageSpec> obj) {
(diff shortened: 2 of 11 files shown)


jngrad:
This PR is needed for the upcoming electrostatics/magnetostatics refactoring PR.

reinaual (approved):

Issue #4347, opened 2021-09-14
lj-demo.py is broken


jngrad:
The physics in [`samples/lj-demo.py`](https://github.com/espressomd/espresso/blob/cc0b79f46f89a796c689a9b8ea26b5e2e1ee6df7/samples/lj-demo.py) is broken since ESPResSo 4.0. The script is also completely untested, and has diverged, e.g. the NpT option doesn't work because the integrator function signature changed in 4.1. It's possible to fix the sample without a MIDI board.

Tasks:

* Remove the phase diagram plot. It is unclear how it was obtained, and the dot temporarily moves into the solid phase when the volume is increased, which is not correct; as the box volume increases, it takes time for the fluid to occupy the newly available space, which is why the dot temporarily enters the solid region. However a phase diagram describes the state of a system at equilibrium, and during the time it takes for the fluid to occupy the new space, the system is strongly out-of-equilibrium. I also couldn't find an analytical solution for the phase diagram of a simple LJ system, and according to the literature it depends strongly on the cutoff value for the LJ potential (Smit 1992, doi:[10.1063/1.462271](https://aip.scitation.org/doi/10.1063/1.462271)).
* Fix the overflow issue when the GUI sliders are moved too quickly.
* Fix NpT error caused by the change of signature of the NpT integrator.
* Add a minimal test case, if feasible. It's not going to be pretty. `testsuite/scripts/samples/test_visualization_npt.py` can be used as template, should *probably* work with the following patch:

<details>
  <summary>patch (click to unroll)</summary>

```python
def disable_GUI_and_MIDI(code):
    # disable MIDI and traits interface
    breakpoint = 'from traits.api import HasTraits, Any, Range, List, Enum, Float\nfrom traitsui.api import View, Group, Item, CheckListEditor, RangeEditor'
    assert breakpoint in code
    code.replace(
        breakpoint,
        'from unittest.mock import MagicMock()\n' +
        '\n'.join(
            f'{x} = MagicMock()' for y in breakpoint.split('\n') for x in y.split(' import ')[1].split(', ')))
    breakpoint = 'system.thermostat.set_langevin(kT=controls.temperature, gamma=1.0)'
    assert breakpoint in code
    code.replace(breakpoint, 'system.thermostat.set_langevin(kT=1, gamma=1.0)')
    breakpoint = 'new_box = np.ones(3) * controls.volume**(1. / 3.)'
    assert breakpoint in code
[...]

PR #4479, opened 2022-03-18
Improve exception handling and fix various regressions

jngrad:
Description of changes:
- improve exception handling in MPI code (partial fix for #4219)
   - several runtime errors used to be queued for far too long and would only surface during integration
      - runtime errors from non-bonded interactions instantiated with a cutoff too large now immediately raise an exception, from which the user needs to recover by either reducing the cutoff or disabling the interaction
      - runtime errors from virtual sites relative applied to particles too far away now immediately raise an exception, from which the user needs to recover by increasing the minimum global cutoff
   - all automatically-generated script interface methods used to call `handle_errors("")` with an empty string; now it's called with the method name to help identify which feature threw the exception
   - virtual sites fatal errors are now runtime errors
   - quaternion fatal errors are now runtime errors
   - invalid particle ids fatal errors are now value errors
- fix bugs
   - virtual sites can no longer relate to themselves
   - particles can no longer be created with a negative id
   - LB node property `boundary` no longer returns a random integer when `LB_BOUNDARIES` is not compiled in
- improve code coverage
   - common exceptions are now properly covered by tests (e.g. error messages from the cell system and integrator)
   - rotation code is now unit tested
   - h5md exceptions are now tested
- fix regressions in the testsuite
   - checkpoint tests were only running on 1 MPI core
   - the LB VTK had a wall boundary misplaced
   - virtual sites relative test cases can now run in any order
- fix regression in the OpenGL visualizer
   - activating the `LB_draw_boundaries` option without any other `LB_draw_*` option no longer throws an exception
   - rescaling the box size in `lj-demo.py` now properly rescales the particle coordinates (partial fix for #4347)


jngrad:
This work is needed for the electrostatics/magnetostatics refactoring in 4.2.0.

testsuite/python/interactions_non-bonded_interface.py
```diff
@@ -25,7 +25,12 @@
 
 
 class Non_bonded_interactionsTests(ut.TestCase):
-    system = espressomd.System(box_l=[20.0, 20.0, 20.0])
+    system = espressomd.System(box_l=[30.0, 30.0, 30.0])
```
reinaual, line 28: does this number even matter in the test?

reinaual (approved):

testsuite/python/interactions_non-bonded_interface.py
jngrad: Yes, with `20` on the x-axis the test fails on 3 MPI ranks because the LJ range of 8.4 is too large. This error message never surfaced before because runtime errors were never checked when setting a interaction. It's only when the integrator runs that it throws, which can be really confusing when the integrator is run indirectly, e.g. via P3M tuning...

Issue #4382, opened 2021-10-27
Look into PyStar


jngrad:
Source code: [alexander.d.kazakov/pystar](https://gitlab.com/alexander.d.kazakov/pystar)

Installation: `pip3 install simlearn`

Running PyStar:

```python
import espressomd
from simlearn import PyStar
PyStar.run()
```

At the time of writing, this requires ESPResSo 4.1.4 (4.2-dev doesn't work out-of-the-box).

Most of the GUI python code was automatically generated from .ui files with `pyuic5`.

Tasks for the coding day:
* determine which simulation ensembles are supported for a simple LJ fluid (NpT, NVT)
* compare features to the ESPResSo OpenGL visualizer:
   * particle bonds drawing
   * shape-based constraints drawing
   * LB boundary geometry drawing
   * LB velocity field drawing
   * user interaction with the system via the mouse or keyboard

RudolfWeeber:
More generally, it needs to be determined, what exactly can be tought using the program, and if/how this could/should tie in with our other Espresso teaching activities.

RiccardoFrenner:
I will look into it tomorrow.

RiccardoFrenner:
# Visualization
- Uses `espressomd.visualization_opengl.OpenGLLive` and creates a separate window and thread for it once the user starts the simulation.
- That means that there are no differences in the display of particle bonds, etc. from the ESPResSo OpenGL visualizer.

# Usage
## As teacher
Use [predefined example systems](#predefined-example-systems) or [create own systems](#create-own-systems) to
- display a window which allows for the initialization of system parameters and interactive control over parameters via e.g. sliders
- create live plots
- save data to files while simulation is running
## As student
- Play with the GUI created by the teacher to see the effects in the OpenGL visualizer or the live plots.
- Analyze the data provided by the teacher

# Predefined example systems
## `LJ fluid 2D/3D`
- Particles with LJ interaction
- Interactive control over
    - number of particles
    - thermostat (on/off, temperature),
    - gravity (on/off, strength),
    - particle type (LJ sigma and eps) with some presets (Argon, Oxygen)
    - live plots of energies, pressure and radial distribution function (RDF)
- Teaching:
    - Could be integrated in a lecture about ideal gases and pressure since one can see clearly the effect on the pressure when changing some parameters like temperature
    - Some parameters also influence the RDF in interesting ways which helps in the understanding of it
[...]

PR #4482, opened 2022-03-29
Script interface maintenance

### LINKED ISSUE #4347: lj-demo.py is broken

jngrad:
Description of changes:
- API change: rename `espressomd.reaction_ensemble` to `espressomd.reaction_methods`
- feature configuration: remove obsolete `EXPERIMENTAL_FEATURES`
- script interface:
   - use shared pointers instead of `new` expressions in the script interface
   - make type conversion error messages less ambiguous by putting quotes around C++ types
   - require EK node grid indices to be exactly 3 integers (2 indices would skip the exception and return `None`)
- python interface:
   - replace `from X import Y` by `import X` and `X.Y` in samples and python interface
   - use `tqdm`, `numpy` and f-strings more often in the samples
   - remove `lj-demo.py` (fixes  #4347)
   - remove mentions to non-existent functions in the user guide


src/python/espressomd/analyze.pyx
```diff
@@ -560,7 +559,7 @@ class Analysis:
             raise ValueError("type_list_b has to be a list!")
 
         if r_max is None:
-            box_l = make_array_locked(< Vector3d > box_geo.length())
```
christophlohrmann, line 563: How was this never caught by anyone?

src/python/espressomd/electrokinetics.pyx
```diff
@@ -496,13 +491,11 @@ IF ELECTROKINETICS:
...
-                            self.id, self.node[0], self.node[1], self.node[2], value) != 0:
-                        raise Exception("Species has not been added to EK.")
-                else:
-                    raise ValueError(
-                        "Type of property is wrong. Expected: float.")
+                utils.check_type_or_throw_except(
+                    value, 1, float, "Property 'density' has to be a float")
+                if ek_node_set_density(
+                        self.id, self.node[0], self.node[1], self.node[2], value) != 0:
+                    raise Exception("Species has not been added to EK.")
```
christophlohrmann: That's a very broad Exception

christophlohrmann (dismissed):
PR seems very fine to me. One point though:
As far as I know, we need lj-demo.py for Tag der Wissenschaft. Is there anyone actively working to fix the demo?

src/python/espressomd/analyze.pyx
jngrad: I'm not sure, there should have been a warning about the redundant cast according to the [Cython docs on type casting](https://cython.readthedocs.io/en/stable/src/userguide/language_basics.html#type-casting).

jngrad:
> As far as I know, we need lj-demo.py for Tag der Wissenschaft. Is there anyone actively working to fix the demo?

Not that I know. I sent an email to David and Julian on March 22 to inform them of the planned removal and invite them to voice their objection, but got no reply. For the TdW event, people can use PyStar (#4382). When I used `lj-demo.py` for TdW 2019, people asked why the LJ simulation was following `P=VNRT` instead of `PV=NRT`, and I had no answer. The sample is broken since release 4.0.0, and unfortunately no one here was successful in implementing a working version. It is time to retire it.

src/python/espressomd/electrokinetics.pyx
jngrad: done

reinaual (approved):

PR #4502, opened 2022-04-19
Fix bond visualization; data handling.

pkreissl:
Add updates for bond data in OpenGL visualizer.

Up to now, the OpenGL visualizer collected information about the bonds in a system only once, on initialization. Simulations that dynamically add or remove bonds while running were therefore not correctly visualized.

Note that still dynamic removal of particles or bonds may result in race conditions, if during the drawing process new data has been fetched from the core where particles are no longer available. To lower the chance of this to happen, the data collection now happens with just one communication of `system.parts.all()` as well as bond drawing catching (and ignoring) this exception. (Another point where that might happen is `_draw_system_particles()`, if this proves to be an issue, I can give it another look.)

While this increases the amount of data being communicated, it also reduces the number of communications. To check the performance impact, both the old and new version of the `_update_particles` method were tested on a simple LJ fluid (868 particles). When communicating the minimum number of attributes (`pos` and `type`) the slowdown is about 4 %, when communicating all attributes we get about 4 % speedup. In practice, this does not seem to have a noticeable impact on visualization performance of common simulation setups.

Also, parts of the visualizer, such as the `_draw_system_particles()` method, relied on the particle ids communicated by the core to be contiguous.

Description of changes:
 - fetch bond data at every visualizer update
 - reduce number of communications of particle data
 - store a dictionary mapping the particle id to array index to access the correct data -- not relying on contiguous particle ids


jngrad (commented):
Overall the changes look alright to me, but the new particle id to index mapping logic assumes particle ids start at 0, and breaks down when id 0 doesn't exist:
```python
system = espressomd.System(box_l=[10., 10., 10.])
system.time_step = 0.1
system.cell_system.skin = 0.4
system.non_bonded_inter[0, 0].wca.set_params(epsilon=1., sigma=0.8)
p1 = system.part.add(id=1, pos=[2.0, 0.0, 0.0], v=[0.0, 0.0, 0.0])
p2 = system.part.add(id=2, pos=[0.5, 0.0, 0.0], v=[0.1, 0.0, 0.0])
visualizer.run(1)
```
Output:
```
  File "/ssd/espresso/build/src/python/espressomd/visualization_opengl.py", line 1413, in display
    self._display_all()
  File "/ssd/espresso/build/src/python/espressomd/visualization_opengl.py", line 1314, in _display_all
    self._draw_system()
  File "/ssd/espresso/build/src/python/espressomd/visualization_opengl.py", line 951, in _draw_system
    self._draw_system_particles()
  File "/ssd/espresso/build/src/python/espressomd/visualization_opengl.py", line 1044, in _draw_system_particles
    part_type = int(self.particles['type'][index])
IndexError: index 2 is out of bounds for axis 0 with size 2
```

jngrad (approved):

PR #4342, opened 2021-09-06
fix math in error calculation

jhossbach:
Fixes the calculation of the autocorrelation time in the `create_fits.py` script, also updates the particle exchange code for clarity

PR #4503, opened 2022-04-19
fix math in error calculation

jhossbach:
Fixes the calculation of the autocorrelation time in the create_fits.py script, also updates the particle exchange code for clarity
See #4342


jngrad (approved):
Approved by @RudolfWeeber in #4342.

Issue #4220, opened 2021-04-09
Core objects should have the life expectancy of the corresponding Python objects


jngrad:
Many python objects (e.g. actors and integrators) have a constructor that call a C++ setter that recycles the old core object, i.e. the core setter copies the new parameters passed as arguments into the corresponding fields of an already-existing global struct object, which is default-constructed when the ESPResSo system is instantiated and destroyed when the python session terminates. This approach is error-prone, as it's easy to forget to update one field of the struct, in which case the value that existed before the setter was called will remain in the object (the compiler or IDE cannot issue a warning). This also makes it impossible to exit gracefully from an object setter upon error, e.g. if a combination of parameters is incompatible, since the old state of the object has already been partially or fully overwritten. For an example, see #4211 where the histogram dimensions could be resized to negative values in `LBProfileObservable.hpp` (e.g. `n_samples_x` < 0), in which case the negative value was cast to a large unsigned integer.

This can be remediated by instantiating a new object in the setter and running the sanity checks on that new object. If the checks fail, the setter throws and the old state remains untouched. If the checks pass, the old state is replaced by the new one. This approach was implemented for the NpT and Steepest Descent integrators (8977e50c597486d441690d) and the RF and DH electrostastic methods (383da42b6f7e0c43aae4).

jngrad:
In the long term, these core objects need to be managed by the ScriptInterface instead of being stored as global variables.

jngrad:
This recommendation has now been implemented in almost all features. The last remaining ones (collision detection, integrators, thermostats) have dedicated tickets. Closing.

PR #4506, opened 2022-05-03
Rewrite long-range actors

### LINKED ISSUE #239: P3M GPU has no energy calculation.

### LINKED ISSUE #4454: Long-range actors should be script interface objects
Many electrostatic and magnetostatic methods implemented for the GPU have the following design:

```cython
IF ELECTROSTATICS and MMM1D_GPU:
    cdef class MMM1DGPU(ElectrostaticInteraction):
        cdef shared_ptr[Mmm1dgpuForce] sip
        cdef EspressoSystemInterface * interface

        def __cinit__(self):
            self.interface = &EspressoSystemInterface.Instance()
            self.sip = make_shared[Mmm1dgpuForce](dereference(self.interface))

        def _deactivate_method(self):
            dereference(self.sip).deactivate()
```

The `ScriptInterface` framework provides a convenient interface to store a shared pointer to a core class. It has many advantages:
* no need to manage a shared pointer, this is done by the script interface
* no need to store a raw pointer to the `EspressoSystemInterface` global variable
* no need to re-declare (and maintain) the core class in the .pxd file
* the Cython class can be converted to a regular Python class

The CPU implementations can also benefit from such a refactoring:
* no need to duplicate class declarations in the .pxd files
* all electrostatics and magnetostatics core global variables can be removed

## Roadmap

**Rewrite electrostatic and magnetostatic methods as C++ classes**
* [x] improve interface segregation and encapsulate algorithms in classes
* [x] do dependency inversion (e.g. `dipoles.hpp` should include all magnetostatic methods, instead of all magnetostatic methods including `dipoles.hpp`)
* [x] extract common code (P3M data structs and utility functions, FFT code)
* [x] move actor source files to the electrostatics or magnetostatics modules
* [x] remove the `Actors` framework in the core
* [x] split electrostatic methods from magnetostatic methods
   * the core should have three new folders: `electrostatics`, `magnetostatics`, `p3m` (for common P3M code and FFT)
   * the core should no longer have folders `actors`, `electrostatics_magnetostatics`
* [x] split ScaFaCoS electrostatic methods from ScaFaCoS magnetostatic methods

**Write the script interface**
* [x] write `ScriptInterface` classes for electrostatic methods
* [x] write `ScriptInterface` classes for magnetostatic methods
* [x] write `ScriptInterface` classes for ScaFaCoS methods
* [x] sort out issues related to MPI errors when throwing exceptions on worker nodes, e.g. in the constructors and tuning functions (see #4449)
* [x] sort out issues related to MPI deadlock when mixing `boost::mpi` communication with `MpiCallbacks` communication

**Roll out changes to the core**
* [x] remove all electrostatics/magnetostatics global variables in the core
* [x] add `boost::optional<boost::variant<Method1,Method2,Method3>> magnetostatics_actor` global variables in the core
* [x] implement boost visitors for the energy, force and pressure calculation
* [x] make DLC a stand alone actor instead of an extension to break dependencies in the core and lift ambiguities in the script interface (see #2155)
   * e.g. what should be the state of `DipolarP3M` when removing the `DLC` "extension"? The `DipolarP3M` force and energy kernels are not `const`, therefore the actor will be altered and no longer be in sync with the `DP3M` python class. Making the `DP3M` actor an argument of `DLC` removes the ambiguity. In addition, the Cython `Actors` class no longer has to keep track of active actors vs. active extensions.
* [x] clarify when `on_coulomb_change()`, parameter broadcasts and long-range actor sanity checks are really needed; right now they are called in too many places (e.g. sanity checks are run at each integration step during tuning) and parameter broadcasts override each other when an actor is adapted by `ELC` or `DLC`. Parameter broadcasts need to be removed entirely by making the tuning code synchronous, sanity checks should only be executed when an actor is active, and the `on_coulomb_change()` should only be called when activating, deactivating or modifying the state of an actor.
* [x] runtime errors triggered by long-range actor sanity checks are queued and their handling is sometimes deferred to the integration loop, which is too late (fixed by #4479)
* [x] break dependencies in the core: at the moment, `coulomb.hpp` is indirectly included in around 60 files, half of which are totally unrelated to electrostatics; to break this chain of includes, we need to make the `BondedCoulombSR` kernels non-inline (i.e. remove its `#include "coulomb_inline.hpp"` and move it to a .cpp file) and split the pair criteria into individual classes
* [x] decouple magnetostatics events from electrostatics events (for example, modifying the `DipolarP3M` parameters should not reinitialize the `P3MGPU` parameters)

jngrad:
Fixes #4454

Description of changes:
- apply the SOLID principles
   - split electrostatics from magnetostatics
   - factor out code duplication
   - extract P3M/DP3M tuning code
   - hide ELC details in private class methods (the energy/force correction logic used to be exposed in `coulomb.cpp`)
   - hide Coulomb implementation details by no longer indirectly including `coulomb.hpp` in 30 unrelated files in the core
- write clear pre-conditions as event hooks
   - these events are triggered upon box size change, periodicity change, node grid change and cell structure change
      - partial fix for #4428
      - setting a `node_grid` incompatible with a long-range actor is no longer possible (the old `node_grid` is restored)
   - these preconditions always run once before the integration loop starts
   - write test cases to check for these events
   - bugfix: charge neutrality checks now run once before the integration loop starts, instead of once at actor activation
- encapsulate the state of long-range methods in classes managed by the script interface
   - bugfix: the cached `DipolarP3M` energy correction is now invalidated when the box length changes
   - remove global variables for CPU methods (partial fix for #2628) and introduce one global variable for each type of long-range actor: dipolar solver, Coulomb solver, Coulomb extension
  - convert Cython files for long-range actors to Python files using the `ScriptInterface` framework
     - long-range actors and their python object now have the same lifetime, regardless of whether the actor is active or not (partial fix for #4220)
  - the list of active long-range actors can no longer end up in an invalid state (partial fix for #4219):
     - updating an active actor with an invalid parameter now automatically restores the original parameter
     - inserting an invalid long-range actor in the list of actors now automatically removes the actor from the list of active actors, even when the exception happens during tuning
- API changes:
   - `ELC` and `DLC` are no longer extensions, instead they take another actor as argument
   - `P3MGPU` is now able to execute the CPU energy and pressure kernels using parameters tuned for the GPU force kernel
      - closes #239
      - P3M-based reaction scripts run with 2x speed-up when reaction steps take place every 1000 integration steps
   - `P3MGPU` is now supported by `ELC`
   - long-range actor parameters are now immutable, since changing one parameter (e.g. the prefactor) would usually invalidate the tuned parameters
   - the charge neutrality check sensitivity can now be modified via `actor.charge_neutrality_tolerance`; this is mostly relevant to actors coupled to `ICC`, whose induced charges can have value spanning several orders of magnitude
- under-the-hood changes:
   - CPU-based long-range actors no longer rely on the `MpiCallbacks` framework
   - the subtracted Coulomb bonded interaction and pair criteria no longer indirectly include `coulomb.hpp`


src/core/electrostatics/icc.cpp
```diff
@@ -0,0 +1,283 @@
...
+    runtimeErrorMsg() << err.what();
+    return;
+  }
+
+  auto prefactor = 0.;
+  if (electrostatics_actor) {
+    prefactor =
+        boost::apply_visitor(GetCoulombPrefactor{}, *electrostatics_actor);
+  }
+  assert(prefactor != 0.);
```
reinaual: Just out of interest, is there a way to reach that point with no `electrostatics_actor` activated? In my ideal world ICC*-iterations should never be executed without an electrostatics solver. That would make the `if (electrostatics_actor)` to always evalute to `true`, right?

src/core/unit_tests/EspressoSystemStandAlone_test.cpp
```diff
@@ -245,12 +265,8 @@ BOOST_FIXTURE_TEST_CASE(espresso_system_stand_alone, ParticleFactory,
...
 
     // set up P3M
-    auto const prefactor = 2.0;
-    auto const mesh = std::vector<int>{8, 8, 8};
-    Coulomb::set_prefactor(prefactor);
-    p3m_set_eps(0.0);
-    p3m_set_params(3.5, mesh.data(), 5, 0.654, 1e-3);
-    p3m_set_mesh_offset(0.5, 0.5, 0.5);
+    mpi_set_tuned_p3m();
+    auto const prefactor = 2.;
```
reinaual, line 271: I'm not to sure if it is a good idea to rely on this variable to be the same as the hard-coded one above in
`auto solver = std::make_shared<CoulombP3M>(std::move(p3m), 2., 1, false);`

src/python/espressomd/actors.py
```diff
@@ -0,0 +1,99 @@
...
+        if actor in Actors.active_actors:
+            raise highlander.ThereCanOnlyBeOne(actor)
+
+        if isinstance(actor, script_interface.ScriptInterfaceHelper):
+            actor._activate()
+
+        self.active_actors.append(actor)
+
+        if not isinstance(actor, script_interface.ScriptInterfaceHelper):
+            actor._activate()
```
reinaual, line 59: where does it come from, that for `ScriptInterfaceHelper`-objects we need to call `_activate()` before appending it to the list?

jngrad: Here is the intended workflow:
- `_activate()` calls the C++ actor registration code that adds the actor in the global optional variant
- the now active actor in the core runs the sanity checks and tuning algorithm (if the method supports tuning)
- if checks or tuning fail, a C++ exception is thrown, which is handled in the [registration code](https://github.com/espressomd/espresso/pull/4506/files/0d6f2c1ba994eeac24eb41c0b11581d668811b07#diff-f9f14648861a7b0f5fa3655eedb94ffb693693d1e0410be86582aae1cdb47d38R54-R62)
   - the actor is safely removed from the core global
   - the exception is re-thrown
   - on MPI rank 0, the C++ exception becomes a Python exception
   - line 54 in actors.py throws
   - the python object is never added to the list of active actors

Since removing an actor from the actor list will trigger sanity checks in the core, we cannot remove a python actor if it is not also in the core. We also cannot keep an invalid actor in the core. LB and EK actors still rely on the old workflow.

src/python/espressomd/lb.pyx
```diff
@@ -24,19 +24,160 @@ import functools
...
+
+    def _deactivate(self):
+        self._deactivate_method()
+        utils.handle_errors("Deactivation of an actor")
+        self._isactive = False
+        inter = self._get_interaction_type()
+        if inter in FluidActor.active_list:
+            if not FluidActor.active_list[inter]:
+                raise Exception(
+                    "Class not registered in Actor.active_list: " + self.__class__.__bases__[0].__name__)
```
reinaual: What is the reason we use `+` and `%`-operators for strings instead of f-strings in this whole file?

jngrad: I had moved the original actor code without modifying it. It's now using f-strings everywhere in the latest commit.

src/core/unit_tests/EspressoSystemStandAlone_test.cpp
jngrad: good point, this is fixed in the latest commit

src/script_interface/electrostatics/ICCStar.hpp
```diff
@@ -0,0 +1,122 @@
...
+         }},
+        {"ext_field", AutoParameter::read_only,
+         [this]() { return actor()->icc_cfg.ext_field; }},
+        {"relaxation", AutoParameter::read_only,
+         [this]() { return actor()->icc_cfg.relaxation; }},
+        {"citeration", AutoParameter::read_only,
+         [this]() { return actor()->icc_cfg.citeration; }},
+        {"first_id", AutoParameter::read_only,
+         [this]() { return actor()->icc_cfg.first_id; }},
+    });
```
reinaual, line 77: Technically it would not be necessary to prevent users from overwriting any parameter here (except for `citeration`). Is there a reason we don't allow it?

src/core/tuning.cpp
```diff
@@ -39,36 +37,63 @@
...
+  if (this_node == 0) {
+    check_statistics(running_average);
   }
 
-  /* MPI returns s, return value should be in ms. */
-  return 1000. * running_average.avg();
+  /* MPI returns in seconds, returned value should be in ms. */
+  auto retval = 1000. * running_average.avg();
+  boost::mpi::broadcast(comm_cart, retval, 0);
+  return retval;
```
reinaual, line 96: Am i right that it wasn't necessary to broadcast this value before because the timing only ran on the head-node?

jngrad: that's correct; now all MPI ranks must have the same timings, otherwise the bisection will diverge

src/script_interface/electrostatics/ICCStar.hpp
jngrad: Actually, most of these parameters depend on one another. For example, we cannot append a value to `sigmas` because it would have a different length as `epsilons` and `areas`. We also cannot move the vector length checks from the constructor to a new method `check_vector_consistency`, because from the python interface we can only update ICC properties one by one. We would need to create a lock system, like a context manager, to guarantee that a set of changes is self-consistent, and if it is not roll back to the old properties. That's doable, but then what should we do if checkpointing is triggered by a `Ctrl+C` signal in the middle of the context manager?

I chose this solution because it seems easier to me to create a new ICC object with the new properties to replace the existing one. Let me know if there is a use case for dynamically changing these properties during the simulation.

reinaual: okay, i see the trouble with that... altough the only criterion that has to be fulfilled is that `sigmas, areas, epsilons, normals` have to have a length `>= n_icc` for ICC to work properly. It only uses the first `n_icc` values of the arrays anyway, so they could also each have a different length, as long as they are long enough.

AFAIK there is no real use-case right now for dynamically changing the amount of ICC charges, although I did have a look at such systems once where i did ICC-refinements. What i would argue should be changeable are all the remaining parameters that don't have any side-effect outside of ICC and can't leave it in a problematic state. This includes `max_iterations, eps_out, convergence, ext_field, relaxation` (personally I don't mind about `first_id`, altough this also has no side-effect outside of ICC). For the `ext_field` parameter I could also see a use-case, e.g. for alternating electric fields. Probably this would also be close to no effort to implement.

TL;DR: How about we leave all ICC particle-data immutable, but one can change all the other stuff that is independent from each other and has no influence outside of ICC?

src/scafacos/src/Coulomb.cpp
```diff
@@ -0,0 +1,97 @@
...
+                  std::vector<double> &potentials) {
+
+  auto const n_part = charges.size();
+  fields.resize(3ul * n_part);
+  potentials.resize(n_part);
+
+  tune(charges, positions);
+  auto const size = static_cast<int>(n_part);
+  handle_error(fcs_run(m_handle, size, positions.data(), charges.data(),
+                       fields.data(), potentials.data()));
```
jngrad, line 87: Does anyone remember why in the original implementation, we need to re-tune ScaFaCoS before each integration step? Maybe it has to do with side effects when the `node_grid` or number of particles changes, or when we reload from a checkpoint (tuned parameters cannot be checkpointed). Commenting out line 84 doesn't affect the testsuite, but the edge case it's supposedly guarding against is not documented in the original implementation.

reinaual (commented):
LGTM

reinaual (approved):

PR #4508, opened 2022-05-09
Store absolute paths in h5md files

jngrad:
Description of changes:
- resolve relative paths and symlinks when writing h5md file paths to avoid discrepancies when reloading from a checkpoint


pkreissl (approved):
lgtm.

reinaual (approved):

PR #4516, opened 2022-05-25
Python maintenance

### LINKED ISSUE #4517: Slab neighborhood search is broken
The [`system.analysis.nbhood(pos=None, r_catch=None, plane='3d')`](https://espressomd.github.io/doc4.1.4/espressomd.html#espressomd.analyze.Analysis.nbhood) function has two search mode: spherical region (default) and slab region. The slab search is incorrect, because the bitmask is inverted. Instead of searching for particles within a distance `r_catch` of a plane, it searches for particles in a cylinder of radius `r_catch` perpendicular to the plane. There is no python test for the slab search mode. The slab search mode was already incorrect in the Tcl version of ESPResSo, and it doesn't take Lees-Edwards boundary conditions into account.

MWE:

```python
import espressomd
import numpy as np
origin = np.array([10., 10., 0.])
offset = 4.
search_dist = 1.
system.part.add(pos=origin + [0., 0., 0.]) # inside sphere, distance to center is 0
system.part.add(pos=origin + [0., 0., offset]) # inside cylinder, distance to z-axis is 0
system.part.add(pos=origin + [offset, 0., 0.]) # inside slab, distance to xy-plane is 0
print(system.analysis.nbhood(origin, search_dist, plane='3d'))
print(system.analysis.nbhood(origin, search_dist, plane='xy'))
```

Output:

```
[0]
[0, 1]
```

Expected output:

```
[0]
[0, 2]
```

Visual inspection (only particles in the neighborhood appear in the visualizer):

```python
system = espressomd.System(box_l=[10] * 3, time_step=0.1)
system.cell_system.skin = 0.
for i in range(10):
    for j in range(10):
        for k in range(10):
            system.part.add(pos=[i,j,k])

ids = system.analysis.nbhood([5., 5., 5.], 4.5, plane='xy')
coords = system.part.by_ids(ids).pos
system.part.clear()
for pos in coords:
    system.part.add(pos=pos)

espressomd.visualization.openGLLive(system).run(0)
```

jngrad:
Description of changes:
- user guide:
   - cleanup code examples and fix bond breakage documentation (e.g. mode `"bind_centers"` vs. `"bond_centers"`)
   - document ghost particles side-effects on observables
   - where changes were made, wrap text to ~80 chars per line to make git diffs more readable
- python testsuite:
   - reduce code duplication by merging similar tests together and by using `tests_common.py` more often
   - improve code coverage by rewriting assertions
- bug fixes:
   - the incorrectly implemented slab search mode of  `system.analysis.nbhood()` was removed (fixes #4517)
   - `TabulatedNonBonded.is_active()` now returns `False` instead of `None` when the bond is inactive
   - always checks for runtime errors when the cell system changes or when a `ScriptInterface` object is instantiated
- API changes:
   - script interface methods no longer return `True` to signal they didn't raise an exception
- script interface improvements:
   - simplify `System` class and document why the `_Globals` class (now renamed to `_BoxGeometry`) has to exist
   - simplify `ScriptInterfaceHelper` classes using automatically generated member methods
   - simplify several core functions that are included in pxd files


doc/sphinx/advanced_methods.rst
```diff
@@ -12,36 +12,47 @@ literature before using them.
...
-  * ``bond_centers`` is the bonded interaction (an instance of :class:`espressomd.interactions.HarmonicBond`) to be created between the particles. No guarantees are made regarding which of the two colliding particles gets the bond. Once there is a bond of this type on any of the colliding particles, no further binding occurs for this pair of particles.
-
-* ``"bind_at_point_of_collision"``: this mode prevents sliding of the colliding particles at the contact. This is achieved by
-  creating two virtual sites at the point of collision. They are
-  rigidly connected to the colliding particles, respectively. A bond is
+  * ``distance`` is the distance between two particles at which the binding is triggered.
+    This cutoff distance, ``1.5`` in the example above, is typically chosen slightly larger
+    than the particle diameter. It is also a good choice for the equilibrium length of the bond.
+  * ``bond_centers`` is the bonded interaction (an instance of :class:`~espressomd.interactions.HarmonicBond`)
```
pkreissl: Does this have to be a `HarmonicBond`? Or is any bonded interaction permissible, in principle?

doc/sphinx/analysis.rst
```diff
@@ -90,7 +90,7 @@ Particles in the neighborhood
 
 :meth:`espressomd.analyze.Analysis.nbhood`
 
-Returns a list of the particle ids of that fall within a given radius of a target position.
+Returns a list of the particle ids that fall within a given radius of a target position.
```
pkreissl: To be precise, grammatically this statement is incorrect. While one likelyt understands the intended meaning, the particle **ids** do not fall within the given radius. Rather should be along these lines: "Returns a list of the ids of particles that fall [...]"

doc/sphinx/lb.rst
```diff
@@ -89,7 +97,9 @@ expert, leave their defaults unchanged. If you do change them, note that they
 are to be given in LB units.
 
 Before running a simulation at least the following parameters must be
-set up: ``agrid``, ``tau``, ``visc``, ``dens``. For the other parameters, the following are taken: ``bulk_visc=0``, ``gamma_odd=0``, ``gamma_even=0``, ``ext_force_density=[0,0,0]``.
+set up: ``agrid``, ``tau``, ``visc``, ``dens``. For the other parameters,
+the following are taken: ``bulk_visc=0``, ``gamma_odd=0``, ``gamma_even=0``,
+``ext_force_density=[0,0,0]``.
```
pkreissl: ``ext_force_density=[0, 0, 0]``?

doc/sphinx/lb.rst
```diff
@@ -102,8 +112,8 @@ Checkpointing
     lb.load_checkpoint(path, binary)
 
 The first command saves all of the LB fluid nodes' populations to an ascii
-(``binary=False``) or binary (``binary=True``) format respectively. The load command
-loads the populations from a checkpoint file written with
+(``binary=False``) or binary (``binary=True``) format respectively.
+The load command loads the populations from a checkpoint file written with
```
pkreissl: The ``lb.load_checkpoint`` command [...]?

doc/sphinx/lb.rst
```diff
@@ -188,7 +200,10 @@ Appending three indices to the ``lb`` object returns an object that represents t
     lb[x, y, z].boundary             # flag indicating whether the node is fluid or boundary (fluid: boundary=0, boundary: boundary != 0)
     lb[x, y, z].population           # 19 LB populations (a numpy array of 19 floats, check order from the source code)
 
-All of these properties can be read and used in further calculations. Only the property ``population`` can be modified. The indices ``x,y,z`` are integers and enumerate the LB nodes in the three directions, starts with 0. To modify ``boundary``, refer to :ref:`Setting up boundary conditions`.
+All of these properties can be read and used in further calculations.
+Only the property ``population`` can be modified. The indices ``x,y,z``
```
pkreissl: ``x, y, z`` (with spaces)?

src/core/polymer.cpp
```diff
@@ -67,6 +66,23 @@ template <class RNG> static Utils::Vector3d random_unit_vector(RNG &rng) {
   return v;
 }
 
+/** Calculate minimal distance to point.
+ *  Note: Particle handles may be invalidated!
+ *  @param partCfg particle selection
+ *  @param pos  point from which to check for close neighbors
```
pkreissl, line 72: Double space after "pos"; although, may just be the default indent before the parameter description... caught my eye, though

src/python/espressomd/system.pyx
```diff
@@ -16,55 +16,56 @@
...
 
-if VIRTUAL_SITES:
-    setable_properties.append("_active_virtual_sites_handle")
-    checkpointable_properties.append("_active_virtual_sites_handle")
 
+cdef class _BoxGeometry:
+    """
+    Wrapper class required for technical reasons only.
+
+    When reloading from a checkpoint file, the box length, periodicity and
```
pkreissl: depending on your philosophy, an Oxford-comma after periodicity...

src/python/espressomd/system.pyx
```diff
@@ -76,48 +77,36 @@ cdef class _Globals:
...
-        (3,) array_like of :obj:`float`:
-            Dimensions of the simulation box
-
-        """
-
-        def __set__(self, _box_l):
-            if len(_box_l) != 3:
+        def __set__(self, box_l):
+            if len(box_l) != 3:
                 raise ValueError("Box length must be of length 3")
```
pkreissl: mus be **array_like** of length 3?

src/python/espressomd/system.pyx
```diff
@@ -76,48 +77,36 @@ cdef class _Globals:
...
-
-        def __set__(self, _periodic):
-            if len(_periodic) != 3:
+        def __set__(self, periodic):
+            if len(periodic) != 3:
                 raise ValueError(
-                    f"periodicity must be of length 3, got length {len(_periodic)}")
-            mpi_set_periodicity(_periodic[0], _periodic[1], _periodic[2])
-            handle_errors("Error while assigning system periodicity")
+                    f"periodicity must be of length 3, got length {len(periodic)}")
```
pkreissl: maybe also "array_like of lenght 3"

src/python/espressomd/utils.pyx
```diff
@@ -235,6 +234,13 @@ cdef Vector3d make_Vector3d(a):
     return v
 
 
+cdef Vector3i make_Vector3i(a):
```
pkreissl, line 238: should there be an assert that ``len(a) == 3``?

src/python/espressomd/visualization.py
```diff
@@ -1278,18 +1266,17 @@ def _draw_lb_vel(self):
     def _modulo_indexing(l, t):
         return l[t % len(l)]
 
-    # FADE PARTICLE CHARGE COLOR FROM WHITE (q=0) to PLUSCOLOR (q=q_max) RESP
-    # MINUSCOLOR (q=q_min)
+    # FADE PARTICLE CHARGE COLOR FROM WHITE (q=0) to PLUS_COLOR (q=q_max) RESP
```
pkreissl: At some point, we should probably just convert the comments in `visualization.py` to normal capitalization instead of UPPERCASE...

pkreissl (changes requested):

src/core/polymer.cpp
jngrad: In doxygen comments, the amount of whitespace between a parameter name and its description is arbitrary. In the past I tried to make it more consistent in the codebase by properly aligning the descriptions, but it's time-consuming. And as soon as a new parameter is introduced whose name is a bit too long, or a `@param[in,out]` is changed to a `@param[in]`, the entire list of `@param` needs to be reformatted, possibly with extra line breaks (which can't be masked with `git diff --color-words` because newlines start with an asterisk). Not really worth the effort, and on some occasions it made the resolution of merge conflicts more difficult.

src/python/espressomd/visualization.py
jngrad: done! it took forever

doc/sphinx/advanced_methods.rst
jngrad: Any bond type is allowed. This sentence was supposed to read "in the example above, an instance of HarmonicBond".

src/python/espressomd/utils.pyx
jngrad: I don't think it's necessary, the helper functions `make_Vector3i()` and `make_Vector3d()` should only be called on a variable `a` after it passed the sanity check `check_type_or_throw_except(a, 3, float, "...")`.

pkreissl (approved):
Thanks a lot for the nice code improvements and for putting in the extra work of capitalization in the visualizer!

testsuite/python/linear_momentum_lb.py
```diff
@@ -92,7 +92,7 @@ class LBGPULinearMomentum(LinearMomentumTest, ut.TestCase):
     """Test for the GPU implementation of the LB."""
 
     lb_class = espressomd.lb.LBFluidGPU
-    atol = 1e-6
+    atol = 1e-5
```
reinaual, line 95: what happened here that the tolerance needs to be increased?

reinaual (approved):

testsuite/python/linear_momentum_lb.py
jngrad: This test is failing at random in CI, e.g. [logfile 317102](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/317102) in the long-range actors refactoring PR. The numerical error is random with a probability distribution that looks Chi-squared. The test fixes all random seeds and the core is supposed to sum the densities in a deterministic order, so I'm not sure what the root cause is. I ran that test in a loop for several thousand times and the largest deviation was 4e-6.

PR #4533, opened 2022-07-06
Fixed typo doc 8.1

simongravelle:
Description of changes:
- Fixed typo in the bond documentation (8.1)
- Added instruction to instantiate a Quartic bond, and specified that the r_cut is not optional in that case.


jngrad (approved):
Thanks!

PR #4544, opened 2022-07-27
Test P3M stress tensor

### LINKED ISSUE #1711: p3m cpu stress tensor lacks a test

jngrad:
Fixes #1711

Description of changes:
- test P3M and DP3M against the Madelung constant of 1D, 2D and 3D crystals
- re-enable python tests on 3 cpu cores in CI


testsuite/python/p3m_madelung.py
```diff
@@ -0,0 +1,294 @@
...
+            energy_per_dip = self.get_normalized_obs_per_dipole(dipm, spacing)
+            np.testing.assert_allclose(energy_per_dip, mc, atol=0., rtol=3e-4)
+
+        dp3m = espressomd.magnetostatics.DipolarP3M(**dp3m_params)
+        self.system.actors.add(dp3m)
+        check()
+
+    @utx.skipIfMissingFeatures(["P3M"])
+    def test_infinite_ionic_wire(self):
+        n_pairs = 64
```
reinaual, line 191: what is the reason for using that many pairs of ions? The result should be independent on the number of ion pairs...

jngrad: The precision of the energy and stress tensor is between 2 and 3 orders of magnitude larger than the P3M accuracy, therefore to test with a tolerance of 1e-6, the P3M accuracy needs to be in the range 1e-8 to 1e-9. But in 1D it's really hard to find a r_cut and cao that allows for such a precision. I'm not really sure why adding more particles, because the real and k-space error magnitudes should actually increase with the square root of the number of particles. Yet in the 3D case you can get away with 8 pairs. Maybe the numerical error gets down when you do the reduction to get the energy or stress, since you divide by the number of particles, i.e. the error gets overall divided by the square root of the number of particles?

reinaual (approved):

PR #4549, opened 2022-08-10
Modernize CMake

jngrad:
Description of changes:
- use [`FindPython`](https://cmake.org/cmake/help/v3.22/module/FindPython.html) instead of our custom CMake functions to detect Python and NumPy
- better use [`GNUInstallDirs`](https://cmake.org/cmake/help/v3.22/module/GNUInstallDirs.html), for example when building ESPResSo without the python interface, install the C++ shared objects in `${prefix}/lib` instead of the (possibly non-existing) python site-package folder
- rename shared object targets to lowercase for consistency with other libraries


jngrad:
This PR is blocking the MpiCallbacks refactoring PR.

reinaual (approved):

PR #4557, opened 2022-08-23
Pitchforkize config

### LINKED ISSUE #4554: Re-organize C++ modules
This is a pre-requisite for #2530.

Several C++ modules mix public header files with private header files, which leads to violations of the SOLID principles. It makes regular maintenance more difficult, and can lead to cryptic compiler bugs. For example, `#include <version.hpp>` is ambiguous: should it include `config/version.hpp` or `script_interface/code_info/Version.hpp`? Note that some compilers use case-sensitive paths, while others use case-insensitive paths.

The Pitchfork layout solves this issue by encapsulating the public header files in an isolated leaf of the project tree, as follows:

```
./
└── module_name/
    ├── CMakeLists.txt
    ├── src/
    │   ├── CMakeLists.txt
    │   ├── private.hpp
    │   └── private.cpp
    ├── tests/
    │   ├── CMakeLists.txt
    │   └── unit_test.cpp
    └── include/
        └── module_name/
            └── public.hpp
```

Then in CMake, add `-I ${CMAKE_CURRENT_SOURCE_DIR}/include/` to any consumer target. These targets can include the public API with `#include <module_name/public.hpp>`, but it is impossible to include `private.hpp`.

Most submodules are already using the Pitchfork layout. The `src/config` submodule needs to be converted to the Pitchfork layout, and both the `SOURCE` and `BUILD` include folders need to be included, since the `version.hpp` file is procedurally generated in the build folder.

The waLBerla bridge also needs to be converted to the Pitchfork layout. The long-term plan is to extract the waLBerla bridge into a self-contained git repository, so that other projects like GROMACS can re-use it. The waLBerla bridge could look like this:

```
./src/
└── walberla_bridge/
    ├── src/
    │   ├── lattice_boltzmann/
    │   │   └── lb_walberla_init.cpp
    │   ├── electrokinetics/
    │   │   └── ek_walberla_init.cpp
    │   ├── generated_kernels/*.cpp
    │   ├── LatticeModel.cpp
    │   ├── LatticeWalberla.cpp
    │   └── lb_walberla_init.cpp
    ├── tests/
    └── include/
        └── walberla_bridge
            ├── lattice_boltzmann/
            │   ├── LBWalberlaBase.hpp
            │   ├── LBWalberlaImpl.hpp
            │   └── lb_walberla_init.hpp
            ├── electrokinetics/
            │   ├── PoissonSolver/
            │   ├── reactions/
            │   ├── EKinWalberlaBase.hpp
            │   ├── EKinWalberlaImpl.hpp
            │   └── ek_walberla_init.hpp
            ├── generated_kernels/*.h
            ├── utils/
            │   ├── boundary_utils.hpp
            │   └── walberla_utils.hpp
            ├── BlockAndCell.hpp
            ├── LatticeModel.hpp
            └── LatticeWalberla.hpp
```

The `src/script_interface` could be converted as well, although it is not as critical, since its header files are actually all private. The `src/core` module cannot be converted to the Pitchfork layout at the moment, since it is much more complex and also will undergo significant changes in the coming few months.

christophlohrmann:
Partially fixes #4554

Description of changes:
- introduce ``include`` and ``src`` directory in ``src/config``


jngrad (approved):
Thanks a lot!

PR #4558, opened 2022-08-24
Rewrite non-bonded interactions as script interface objects

### LINKED ISSUE #4555: Rewrite non-bonded interactions as script interface objects
This is a self-contained task for #4542.

The python interface currently relies on MpiCallbacks to broadcast interaction structs to all MPI ranks. This communication can be taken care of by the `ScriptInterface` framework.

Use the [`jngrad/nonbonded_refactor`](https://github.com/espressomd/espresso/compare/python...jngrad:nonbonded_refactor) branch as the starting point. The first commit lays out the new interface, and can be ignored. The next two commits show how to use the new interface with LJ and WCA. Simply apply that formula to all remaining interactions. Some interactions might require extra care, for example DPD requires two structs and interacts with thermostats. Be sure to use the maxset myconfig file and to run the testsuite frequently to catch regressions.

Once this is done and the python testsuite shows no regression, it will be possible to delete the old interface in the core. To that end, delete the code marked with `// TODO` in ca9f833a1223048bf875a9dad58e9baae9d3eccb, then build again and delete all lines that trigger *any* compiler warning.

### LINKED ISSUE #4569: Non-bonded interactions `set_params()` is broken
The non-bonded interactions `set_params()` method has undocumented behavior that can lead to unphysical potentials.

Once a non-bonded interaction is activated with valid parameters using `set_params()`, the second call to `set_params()` silently changes behavior: required parameters become optional, and missing arguments are populated using the values currently in the core, completely bypassing the `default_params()` method.

This unexpected behavior can compromise simulation scripts in two ways:
1. The user intends to reset an interaction with only the required parameters, expecting the default parameters to populate the missing non-required parameters: the old parameters are silently recycled
2. The user discovers this undocumented behavior and intends to use it to update one parameter of a `LennardJonesInteraction` or `GenericLennardJonesInteraction` potential that was previously set up with `shift="auto"`: the `shift` parameter is not automatically recalculated and the potential has a jump in energy.

Here is a MWE to reproduce the bug in the second scenario:

```python
import espressomd
import numpy as np

system = espressomd.System(box_l=3 * [10.])
system.time_step = .1

def sample():
    start_pos = [0., 0., 0.]
    step = np.array([0.01 / 600, 0., 0.])
    system.part.clear()
    p0, p1 = system.part.add(pos=[start_pos, start_pos + 600 * step], type=[0, 0])
    xval = []
    yval = []
    for _ in np.arange(180):
        p1.pos = p1.pos + step
        xval.append(np.linalg.norm(p1.pos - p0.pos))
        yval.append(system.analysis.energy()["non_bonded"])
    return np.array(xval), np.array(yval)

system.non_bonded_inter[0, 0].lennard_jones.set_params(epsilon=1.92, sigma=0.05, cutoff=0.11, shift="auto")
# trigger bug: forget to use shift="auto", so the shift isn't recalculated for the new cutoff
system.non_bonded_inter[0, 0].lennard_jones.set_params(cutoff=0.012)
xval1, yval1 = sample()
# fix bug: manually add shift="auto"
system.non_bonded_inter[0, 0].lennard_jones.set_params(cutoff=0.012, shift="auto")
xval2, yval2 = sample()

plt.plot(xval2, yval2, label="set_params(cutoff=0.012, shift=\"auto\")")
plt.plot(xval1, yval1, label="set_params(cutoff=0.012)")
plt.xlabel("Inter-particle distance")
plt.ylabel("Energy")
plt.legend()
plt.show()
```
Plot of the MWE showing the energy jump for a LJ potential

jhossbach:
Fixes #4555 and fixes #4569

Description of changes:
- rewrite non-bonded interactions as script interface objects
   - convert Cython file `interaction.pyx` to Python file `interaction.py`
   - 1600 lines of Cython code were removed
- bugfix: the `set_params()` method now raises an error if required arguments are not provided
   - scripts may silently break when a potential is set and later updated in a way that is affected by the bug described in #4569
- new feature: possibility to deactivate/reset non-bonded interactions
- reduce code duplication in the script interface of bonded interactions


src/core/nonbonded_interactions/nonbonded_interaction_data.cpp
```diff
@@ -60,21 +61,33 @@ static double min_global_cut = INACTIVE_CUTOFF;
...
+    if (ia_params == nullptr) {
+      ia_params = std::make_shared<IA_parameters>();
+    }
+  }
 
-  max_seen_particle_type = new_size;
-  std::swap(nonbonded_ia_params, new_params);
+  ::max_seen_particle_type = new_size;
+  std::swap(::old_nonbonded_ia_params, new_params_);
+  std::swap(::nonbonded_ia_params, new_params);
```
reinaual: why do we need a swap here? Probably a move-assignment would be more expressive

src/python/espressomd/interactions.pyx
```diff
@@ -205,6 +205,85 @@ cdef class NonBondedInteraction:
         raise Exception(
             "Subclasses of NonBondedInteraction must define the required_keys() method.")
 
+
+class NewNonBondedInteraction(ScriptInterfaceHelper):
```
reinaual: Have we ever tested the [python abstract base class](https://docs.python.org/3/library/abc.html) to work together with the ScriptInterface? Probably a lot of the methods would be perfect candidates for `@abc.abstractmethod`

src/core/nonbonded_interactions/nonbonded_interaction_data.cpp
jngrad: I just looked it up and apparently, a swap was the only way to do a cheap move for `std::vector` back in C++03. I'll change it to a `std::move`.

src/python/espressomd/interactions.pyx
jngrad: Well, it wasn't used before because everything was in Cython, and abstract classes don't work out-of-the-box in Cython:

```
Error compiling Cython file:
------------------------------------------------------------
...

import abc
cdef class HydrodynamicInteraction(FluidActor, metaclass=abc.ABCMeta):
                                                                    ^
------------------------------------------------------------
/work/jgrad/es/src/python/espressomd/lb.pyx:172:69: C classes cannot take keyword bases.
```

or

```
Error compiling Cython file:
------------------------------------------------------------
...

import abc
cdef class FluidActor(abc.ABC):
                        ^
------------------------------------------------------------
/work/jgrad/es/src/python/espressomd/lb.pyx:34:25: First base of 'FluidActor' is not an extension type
```

The usual workaround was to declare a python abstract class and use it to register the cdef class to make it abstract, which is less elegant and error-prone. Now that the module is in Python, we can use ABC metaclasses.

reinaual (dismissed):

reinaual (approved):

PR #4559, opened 2022-08-29
MPI-parallel DDS with replicas

jngrad:
Re-implementation of #2499

Description of changes:
- the `DipolarDirectSumCpu` method was rewritten to support replicas and MPI-parallelization
   - API change: a new argument `n_replicas` was added, defaults to 0 so that old ESPResSo scripts still work
- the original `DipolarDirectSumWithReplicaCpu` method was removed
- the minimum image convention is now tested


stekajack:
Hi @jngrad @RudolfWeeber @fweik Did testing of the MDdS implementation vs. p3m and theory. Will write with some detail the conclusion for completeness.

The test is magnetisation of a ferrofluid, 3d pbc, \phi=4%, lambda=3. Validation against Langevin magnetisation law and modified mean-field theory (MMFT2; 2nd order). Key subtlety is the fact that direct sum and p3m use different boundary conditions (bc), therefore are not comparable as such. There are several ways to go about this. 

By default, p3m use the so-called metallic bc. MMFT2 should agree with p3m in this case. Direct sum solves for essentially a finite system (given enough replicas one can argue a pseudo-infinite system) in vacuum, which in effect means that there is implicitly a demagnetisation field in the results obtained from the direct sum. I expanded the MMFT2 to account for the demagnetisation field and should therefore be able to fit the direct sum results. Alternatively, p3m has a surface correction method implemented which is run if the \epsilon parameter is not 0. To my understanding setting the \epsilon param to 1 corresponds to p3m using vacuum boundary conditions and should therefore match MDdS  and MMFT2 with the demagnetisation field. The figure attached fits my expectations. I would say that MDdS is functioning as intended. Hope this is helpful. Cheers.



mdds_test

src/core/magnetostatics/CMakeLists.txt
```diff
@@ -22,7 +22,6 @@ target_sources(
           ${CMAKE_CURRENT_SOURCE_DIR}/barnes_hut_gpu.cpp
           ${CMAKE_CURRENT_SOURCE_DIR}/dds.cpp
```
RudolfWeeber: pls consider renaming the file to dipolar_direct_sum or the like

src/core/magnetostatics/dds.cpp
```diff
@@ -30,110 +30,376 @@
...
-  return energy;
+  /* Wait for the rest of the data to arrive */
+  boost::mpi::wait_all(reqs.begin(), reqs.end());
+
+  /* Output iterator for the force */
+  p = local_particles.begin();
+
+  /* Interaction with all the other particles */
+  for (auto pi = local_posmom_begin; pi != local_posmom_end; ++pi, ++p) {
+    // red particles
```
RudolfWeeber: For interactions among local particles, each pair is only looked at once and symmetry is used to calculate the oposite force/torque. For interactions between particles at different nodes, this seems not to be the case, so f_ij and f_ji are calculated separately. the result is correct. However, in the future, this could/should be expanded to use the same symmetry and have a final reduciton of the forces. Due to the n^2 scaling, this is probably worth it. Not now, though.

src/core/magnetostatics/dds.cpp
```diff
@@ -30,110 +30,376 @@
...
-    throw std::runtime_error(
-        "MPI parallelization not supported by DipolarDirectSumCpu.");
+/**
+ * @brief Calculate the interaction potential.
+ *
+ * This employs a parallel N-square loop over all particle pairs.
+ */
+double
+DipolarDirectSum::long_range_energy(ParticleRange const &particles) const {
+  auto const &box_l = ::box_geo.length();
```
RudolfWeeber: This kernel looks structurally different from the force kernel. Is there an mpi reduce somewhere else in Es whic accumulates the energy contributions from all the nodes?

RudolfWeeber (commented):
this will be very useful for the folks in Vienna and Stuttgart. Thanks Flo for the initial work and Jn and Deniz for finishing it.

src/core/magnetostatics/CMakeLists.txt
jngrad: done

src/core/magnetostatics/dds.cpp
jngrad: Yes, this is managed by the `Observable_stat` framework, which is always called in parallel from the corresponding observable and analysis function.

RudolfWeeber (approved):

PR #4564, opened 2022-09-05
Simplify VS_Relative, fix VS support for Lees Edwards

RudolfWeeber:
* Removs the less-than-transparent shifting logic which prevents posiiton folding of vs relative. this is no longer needed, since we now use minimum_image_distance everywhere.
* Apply Lees Edwards position shift to VS on update
* Alwasy fold shear plane normal coordinates in Lees Edwards minimum image dsitance

The trajectory reconstruciton stuff isn't working yet.


RudolfWeeber:
I sorted out the trajectory reconstruciotn (p.lees_edwards_offset) as follows:
The lees_edwards_offset tracked in the particle now exactly matches the sum of all position offset aplied on Lees-Edwards boundary crossings. I also formulated the test like that. IMO, this is more consistent than what was there previously.

Should we re-instate a 2 step position update (as we have a two-step integrator and one can argue that the LE position offset has changed during the half-step), this can be done by having a pos update in the Push and UpdateOffset, each.
I'd not re-introduce a two step velocity update, as the velocity shift should match teh particle position: when the particle has crossed the boudnary, the velocity shift has to be applied completely, otherwise, there will be huge forces in a simulation with Lattice-Boltzmann coupling.
There, once, once the boundary is crossed, the flow velocity changes from +v_shear/2 to -v_shear/2, so it would lead to a big force to couple a particle with v=+v_shear/2 - v_shear/2 = 0. Note that thee is only one LB couling, even if we use a two-step integrator.

jngrad:
This work was merged in the walberla branch. Closing.

jngrad (approved):

Issue #4560, opened 2022-08-29
Tons of feature guards missing


RudolfWeeber:
Starting from maxset, removing the following will cause compile errors:

* ELECTROSTATICS (in the script interface)
* DIPOLES (in the script interface)
* NPT (in the script interface)
* EXTERNAL_FORCEs (in virtual sites inertialess tracers due to missing ext_torque paritcel property, might requrie build with cuda)
* VIRUTAL_SITES_INERTIALESS_TRACERS (in script interface)

jngrad:
I tried removing all of these features from maxset, and also removing them individually, but couldn't reproduce any compiler error. Did you use the python branch? Also it is worth noting that our file server has synchronization issues, for example `/work` is 6 seconds in the future since Monday, so if you modified the myconfig file and ran `make` less than 6 seconds later on an affected drive, the linker will link together object files that have been compiled with a different version of the myconfig file. It happened to me yesterday.

RudolfWeeber:
The failling myconfig is:
`
```
#define ROTATION
#define ROTATIONAL_INERTIA
#define MASS
#define THERMOSTAT_PER_PARTICLE
#define LENNARD_JONES
#define VIRTUAL_SITES_RELATIVE
```

and related ones, with some of the featues listed in the initial issue added. This particular one doesn't compiel, others have undefined symbol when running.

PR #4565, opened 2022-09-06
Missing guard ext torque

pkreissl:
Partial fix for #4560 

Add missing `EXTERNAL_FORCES` guard for `ext_torque()`; revealed by the following `myconfig.hpp`:
```c++
#define ROTATION
#define ROTATIONAL_INERTIA
#define MASS
#define THERMOSTAT_PER_PARTICLE
#define LENNARD_JONES
#define VIRTUAL_SITES_RELATIVE
```
The second commit adds missing guards to fix tests that failed with the feature configuration given above.

### DIFF
--- a/src/core/Particle.hpp
+++ b/src/core/Particle.hpp
@@ -492,8 +492,10 @@ public:
   auto &torque() { return f.torque; }
   auto const &omega() const { return m.omega; }
   auto &omega() { return m.omega; }
+#ifdef EXTERNAL_FORCES
   auto const &ext_torque() const { return p.ext_torque; }
   auto &ext_torque() { return p.ext_torque; }
+#endif // EXTERNAL_FORCES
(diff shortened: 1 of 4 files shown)


jngrad:
Could you please check if you fixed more regressions than what was already fixed by 6be18da3cf184bcc6a4b9b6973e0e43d11922dfb? I was not able to find the combination of missing features that lead to missing symbols, for example.

pkreissl:
I was unaware on your working on this issue. Still, I think this PR is more selective with the `Verlet_list_test.cpp` unit test, maybe it is worth another look. Otherwise just close this PR.

jngrad:
Sorry for not marking the other PR as a partial fix, forgot about it.

jngrad (approved):
LGTM

PR #4576, opened 2022-09-23
Checkpointing maintenance

jngrad:
Description of changes:
- allow checkpointing of an empty ESPResSo system
- document how to checkpoint variables at function scope
- document which imports are necessary when loading from a checkpoint

### DIFF
--- a/src/script_interface/integrators/IntegratorHandle.cpp
+++ b/src/script_interface/integrators/IntegratorHandle.cpp
@@ -82,9 +82,19 @@ IntegratorHandle::IntegratorHandle() {
   });
 }
 
+static bool checkpoint_filter(typename VariantMap::value_type const &kv) {
+  /* When loading from a checkpoint file, defer the integrator object to last,
+   * and skip the time_step if it is -1 to avoid triggering sanity checks.
+   */
+  return kv.first != "integrator" and
+         not(kv.first == "time_step" and ::integ_switch == INTEG_METHOD_NVT and
+             get_time_step() == -1. and is_type<double>(kv.second) and
+             get_value<double>(kv.second) == -1.);
+}
+
 void IntegratorHandle::do_construct(VariantMap const &params) {
   for (auto const &kv : params) {
-    if (kv.first != "integrator") {
+    if (checkpoint_filter(kv)) {
       do_set_parameter(kv.first, kv.second);
     }
   }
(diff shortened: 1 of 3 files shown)


src/script_interface/integrators/IntegratorHandle.cpp
```diff
@@ -82,9 +82,19 @@ IntegratorHandle::IntegratorHandle() {
...
 }
 
+static bool checkpoint_filter(typename VariantMap::value_type const &kv) {
+  /* When loading from a checkpoint file, defer the integrator object to last,
+   * and skip the time_step if it is -1 to avoid triggering sanity checks.
+   */
+  return kv.first != "integrator" and
+         not(kv.first == "time_step" and ::integ_switch == INTEG_METHOD_NVT and
+             get_time_step() == -1. and is_type<double>(kv.second) and
+             get_value<double>(kv.second) == -1.);
```
jngrad, line 92: This is inelegant, but I'm unsure what can be done at the moment. The old Cython interface would bypass the public interface and directly inject the values from the checkpoint file into the global variables. Since we want to remove global variables, we need to use the public interface. Here are the alternatives:
* create a new public setter for `time_step` that has no sanity checks
   * anyone could use that setter to bypass sanity checks
   * we would need to create two `IntegratorHandle` constructors, one for checkpointing and one for normal construction (only called once when the system object is created) -> code duplication, risk of divergence
* make the `time_step` a mandatory parameter like `box_l`
   * this is probably the "right" solution, since many sanity checks in the core depend on the time step
   * downside: all ESPResSo scripts will break
* create a global bool variable that indicates whether we are currently reloading from a checkpoint file, such that sanity checks can decide whether to fire or not
   * adding more globals is problematic, maybe once we have all globals in a struct we could have a member to indicate whether we're reloading, but we're still far away from that
   * having such a flag means the original simulation state and the checkpoint reloaded state had to follow two different code paths; this could lead to simulation reproducibility issues if other components of ESPResSo decide to use that global flag too to take different paths and accidentally introduce a regression

reinaual: Personally I'm in favor of the solution to make `time_step` a mandatory argument, since it also reflects the necessity of this parameter.  However, that's a decision we need to discuss at least in the core team.

reinaual (approved):

PR #4589, opened 2022-10-07
Reaction methods bug fixes

### LINKED ISSUE #4587: Monte Carlo displacement move is broken
There are several issues with method [`RE.displacement_mc_move_for_particles_of_type()`](https://espressomd.github.io/doc4.2.0/espressomd.html?highlight=displacement_mc_move_for_particles_of_type#espressomd.reaction_methods.ReactionAlgorithm.displacement_mc_move_for_particles_of_type):

1. it is completely untested and unused in the tutorials and samples
2. the docstring and user guide don't mention it treats particles as distinguishable (sampling without replacement), i.e. when moving 3 particles, you can never be in a situation where you move particle 1, 2, and then 1 again
3. the code enters an infinite loop when passing argument `particle_number_to_be_changed=N` when the system contains less than `N` particles of the chosen type
4. when the move is rejected, the original particle position is restored, but not the original particle velocity
5. when moving `N` particles, if move `j` is rejected, the remaining `N - j - 1` moves are still attempted, even though all `N` moves will necessarily be reverted at the end; this leaves N particles in their original positions but with different velocities

### LINKED ISSUE #4588: Particles without user-defined type are not tracked by reactions
When inserting particles of type 0 with the following syntax:

```python
system.part.add(pos=[1, 2, 3])
```

the particle is created with a default-constructed value for the type, in this case 0. Since the `set_particle_type()` callback is not called, the particle id is not added to the global type map used by reaction methods, and therefore won't be involved in any reaction where one of the reactants has type 0. This affects all ESPResSo versions since 4.0.0.

MWE for 4.2.0 and 4.3-dev:

```python
import espressomd.reaction_methods
import numpy as np
np.random.seed(seed=42)
system = espressomd.System(box_l=3 * [35.0], time_step=0.02)
system.cell_system.skin = 0.4
# 2A + 3B <-> 4C
type_A, type_B, type_C = (0, 1, 2)
nu_A, nu_B, nu_C = (-2, -3, 4)
nu_bar = sum([nu_A, nu_B, nu_C])
K = 1000.
RE = espressomd.reaction_methods.ReactionEnsemble(
    kT=1., exclusion_range=0., seed=4)
RE.set_non_interacting_type(type=3)
RE.add_reaction(
    gamma=K * (0.027)**nu_bar, reactant_types=[type_A, type_B],
    reactant_coefficients=[abs(nu_A), abs(nu_B)],
    product_types=[type_C],
    product_coefficients=[nu_C],
    default_charges={type_A: 0, type_B: 0, type_C: 0})
N0 = 50
for i in range(N0):
    for j in range(abs(nu_B)):
        system.part.add(pos=np.random.random(3) * system.box_l, type=type_B)
    for j in range(abs(nu_A)):
        system.part.add(pos=np.random.random(3) * system.box_l) # missing type=type_A
RE.reaction(reaction_steps=2000)
print(f"n_A = {system.number_of_particles(type=type_A)}")
print(f"n_B = {system.number_of_particles(type=type_B)}")
print(f"n_C = {system.number_of_particles(type=type_C)}")
```

MWE for 4.1.4:
```python
import espressomd.reaction_ensemble
import numpy as np
np.random.seed(seed=42)
system = espressomd.System(box_l=3 * [35.0], time_step=0.02)
system.cell_system.skin = 0.4
# 2A + 3B <-> 4C
type_A, type_B, type_C = (0, 1, 2)
nu_A, nu_B, nu_C = (-2, -3, 4)
nu_bar = sum([nu_A, nu_B, nu_C])
K = 1000.
RE = espressomd.reaction_ensemble.ReactionEnsemble(
    temperature=1., exclusion_radius=0., seed=4)
RE.set_non_interacting_type(3)
RE.add_reaction(
    gamma=K * (0.027)**nu_bar, reactant_types=[type_A, type_B],
    reactant_coefficients=[abs(nu_A), abs(nu_B)],
    product_types=[type_C],
    product_coefficients=[nu_C],
    default_charges={type_A: 0, type_B: 0, type_C: 0})
N0 = 50
for i in range(N0):
    for j in range(abs(nu_B)):
        system.part.add(pos=np.random.random(3) * system.box_l, type=type_B)
    for j in range(abs(nu_A)):
        system.part.add(pos=np.random.random(3) * system.box_l) # missing type=type_A
RE.reaction(reaction_steps=2000)
print(f"n_A = {system.number_of_particles(type=type_A)}")
print(f"n_B = {system.number_of_particles(type=type_B)}")
print(f"n_C = {system.number_of_particles(type=type_C)}")
```
Output:
```
n_A = 0
n_B = 150
n_C = 0
```
Expected output:
```
n_A = 40
n_B = 60
n_C = 120
```

Possible patch for 4.3-dev:
```diff
diff --git a/src/script_interface/particle_data/ParticleHandle.cpp b/src/script_interface/particle_data/ParticleHandle.cpp
index 70b9330a93..19e3f15323 100644
--- a/src/script_interface/particle_data/ParticleHandle.cpp
+++ b/src/script_interface/particle_data/ParticleHandle.cpp
@@ -484,4 +484,7 @@ void ParticleHandle::do_construct(VariantMap const &params) {
       }
     }
+    if (params.count("type") == 0) {
+      do_set_parameter("type", 0);
+    }
   }
 }
```

Possible patch for 4.2.0:
```diff
diff --git a/src/python/espressomd/particle_data.pyx b/src/python/espressomd/particle_data.pyx
index 3dc80634b..964143eee 100644
--- a/src/python/espressomd/particle_data.pyx
+++ b/src/python/espressomd/particle_data.pyx
@@ -1814,4 +1814,7 @@ Set quat and scalar dipole moment (dipm) instead.")
         pid = p_dict.pop("id")
 
+        if "type" not in p_dict:
+            p_dict["type"] = 0
+
         if p_dict != {}:
             self.by_id(pid).update(p_dict)
```

Possible patch for 4.1.4:
```diff
diff --git a/src/python/espressomd/particle_data.pyx b/src/python/espressomd/particle_data.pyx
index 4e7799a29..c93b752fe 100644
--- a/src/python/espressomd/particle_data.pyx
+++ b/src/python/espressomd/particle_data.pyx
@@ -1912,4 +1912,7 @@ Set quat and scalar dipole moment (dipm) instead.")
         del P["id"]
 
+        if "type" not in P:
+            P["type"] = 0
+
         if P != {}:
             self[id].update(P)
```

jngrad:
Description of changes:
- track particles with default-constructed type 0 (fixes #4588)
- restore the original particle velocity when a Monte Carlo displacement move is rejected (fixes #4587)


pm-blanco:
Thanks for involving me in this PR! There are some things that I am not convinced about the MC displacement move method:
- I think that  the input variable `n_part` is ambiguous: one could propose a new configuration moving n_part (as in the current implentation) or one could choose n_part particles and perform an MC trial move with each. For consistency with the rest of MC methods, I propose that instead we use `steps` as input variable, standing for the number of MC steps that the method would do. Then each MC step one particle is randomly chosen and moved, and the move is accepted or rejected according the probability given by the canonical ensemble. 
- Instead of having a method for moving only particles of one specific type, I think that the method should by default try to move any particle. Then, one could provide an optional argument (something like `list_particle_types_to_move`) which could be used to restrict which particles are selected.

@davidbbeyer  do you think that such an implementation would still be compatible for your project? If yes and all of you agree with this alternative implementation I volunteer to change the code accordingly

jngrad:
Offline discussion with @pm-blanco: the Monte Carlo displacement move method in its current state could benefit from a few improvements, but that work might take a couple of weeks to complete. There is also a newly discovered bug when combining constant-pH with Widom insertion, which may be relevant for Monte Carlo displacement moves too, since displacement moves can be carried out from any reaction method class instance.

Since this PR fixes important bugs and is blocking the parallel Monte Carlo project, we agreed that this PR should be merged now. The other points can be taken care of in an upcoming PR that will be orthogonal to the parallel MC project.

pm-blanco (dismissed):
As discussed with @jngrad we accept this PR as it is for now to allow the developing of the parallelization of the MC and we will continue improving it in a future PR

src/core/reaction_methods/ReactionAlgorithm.cpp
```diff
@@ -564,38 +564,49 @@ ReactionAlgorithm::generate_new_particle_positions(int type, int n_particles) {
...
   m_tried_configurational_MC_moves += 1;
   particle_inside_exclusion_range_touched = false;
 
-  auto const particle_number_of_type = number_of_particles_with_type(type);
-  if (particle_number_of_type == 0 or
-      particle_number_of_type_to_be_changed == 0) {
+  if (n_part == 0) {
+    // reject
+    return false;
+  }
```
reinaual: Is this really intentional that we increase the counter of `m_tried_configurational_MC_moves` when `n_part` is zero? Did one really try a configurational move if one doesn't change anything?

jngrad: That's an oversight. Calling this method with 0 particles doesn't make sense, and it isn't meaningful to return true or false because no move was actually attempted. It's even worse if `n_part` is a negative number... I'll fix that.

src/core/reaction_methods/ReactionAlgorithm.cpp
```diff
@@ -564,38 +563,53 @@ ReactionAlgorithm::generate_new_particle_positions(int type, int n_particles) {
...
+  if (n_part <= 0) {
+    throw std::domain_error(
+        "Parameter 'particle_number_to_be_changed' must be >= 1");
+  }
+  auto const n_particles_of_type = number_of_particles_with_type(type);
+  if (n_part > n_particles_of_type) {
+    throw std::domain_error(
+        "Parameter 'particle_number_to_be_changed' must be less or equal to "
+        "the number of particles in the system");
   }
```
reinaual: We should spend a second thought on this change, because it will probably break old espresso scripts.
What I was thinking of is a reactive system (e.g. GCMC) where you have initially setup a species that will react with time. At some point it might happen that this species has completely reacted, but one might have used displacement moves in the script to change the configuration. In such a case one would have just rejected the move in the old implementation, but now an error is thrown. An additional, explicit check before the displacement moves would be necessary to prevent this error.
The very same happens when providing zero, which might happen when using the current particle number as the number of displacement moves.

Disclaimer: I'm not claiming any of this is good practice, but it used to work and I have written scripts where this change would also break my code.

reinaual (approved):

Issue #4595, opened 2022-10-16
Electroneutrality breaks when combining Widom + cpH methods


pm-blanco:
@davidbbeyer and I have observed that when combining the Widom insertion method with the constant pH method (and possibly also with the Reaction ensemble) leads to the destruction of an additional ion, ultimately breaking the electroneutrality of the system. 

Below I provide a minimal example script that produces the bug. This only happens in the 4.2.0 version of espresso but it does not happen in the previous version 4.1.4, so it could be related to one of the newer changes we introduced. 

```python
import espressomd
from espressomd import version
import numpy as np

pH=7
pKa=7
exclusion_range=1
seed=12345
kT=1
BOX_LENGTH=10
N_SALT=2
N_acid = 2
bug=False

types = {"H": 0, "Na": 1, "Cl": 2, "Ha": 3, "a": 4}
charges = {"H": 1.0, "Na": 1.0, "Cl": -1.0, "Ha": 0.0, "a": -1.0}

system = espressomd.System(box_l=[BOX_LENGTH,]*3)
system.part.add(type=[types["Na"]]*N_SALT, pos=np.random.rand(N_SALT, 3) * BOX_LENGTH,
                q=[charges["Na"]]*N_SALT)
system.part.add(type=[types["Cl"]]*N_SALT, pos=np.random.rand(N_SALT, 3) * BOX_LENGTH,
                q=[charges["Cl"]]*N_SALT)
if bug:

    system.part.add(type=[types["Ha"]]*N_acid, pos=np.random.rand(N_acid, 3) * BOX_LENGTH,
                q=[charges["Ha"]]*N_acid)
else:
    system.part.add(type=[types["a"]]*N_acid, pos=np.random.rand(N_acid, 3) * BOX_LENGTH,
                q=[charges["a"]]*N_acid)
    system.part.add(type=[types["H"]]*N_acid, pos=np.random.rand(N_acid, 3) * BOX_LENGTH,
                q=[charges["H"]]*N_acid)

if version.friendly() == "4.2" or version.friendly() == "4.3":
    from espressomd import reaction_methods
    RE = espressomd.reaction_methods.ConstantpHEnsemble(
        constant_pH=pH,
        kT=kT,
        exclusion_range=exclusion_range,
        seed=seed)
    widom = espressomd.reaction_methods.WidomInsertion(
        kT=kT,
        seed=seed)

elif version.friendly() == "4.1":

    from espressomd import reaction_ensemble
[...]

PR #4605, opened 2022-11-08
Refactor particle management

jngrad:
Description of changes:
- particle state tracking is now fully encapsulated
   - helps with separation of concerns and reducing code duplication
- particle creation and particle moves are now carried out by 2 independent functions
    - this makes the reaction methods bugs throw an error instead of silently reacting particles of the wrong type (#4595)


pm-blanco:
@jngrad Maybe I did not understand properly what you meant in this PR but it does not seem to me that it actually throws an error when  reacting particles of the wrong type. I tested your PR with the script that I provide in #4595 and it still does not show any error when producing the bug.

jngrad:
> @jngrad Maybe I did not understand properly what you meant in this PR but it does not seem to me that it actually throws an error when reacting particles of the wrong type. I tested your PR with the script that I provide in #4595 and it still does not show any error when producing the bug.

Here is the output on my workstation:
```
jgrad@seraue:/work/jgrad/espresso-fork-PR/build-lama-release2$ mpirun -n 1 ./pypresso mwe_pablo.py 
step  0
Before widom: a: 1 H: 1 Ha: 1 Na: 2 Cl: 2
total charge  0.0
/work/jgrad/espresso-fork-PR/build-lama-release2/src/python/espressomd/reaction_methods.py:462:
FutureWarning: arguments 'reactant_coefficients' and 'product_coefficients' are deprecated and
are no longer necessary for the constant pH ensemble. They are kept for backward compatibility
but might be deleted in future versions.
  warnings.warn(warn_msg, FutureWarning)
  File "mwe_pablo.py", line 105, in <module>
    widom.calculate_particle_insertion_potential_energy(reaction_id=0)
  File "/work/jgrad/espresso-fork-PR/build-lama-release2/src/python/espressomd/reaction_methods.py", line 515, in calculate_particle_insertion_potential_energy
    return self.call_method(
  File "script_interface.pyx", line 159, in espressomd.script_interface.PScriptInterface.call_method
RuntimeError: Invalid local particle index entry.
```

In the end, the bug needs to be removed with your PR.

pm-blanco:
That's strange, somehow I cannot reproduce it in my workstation.  I get
```
mpirun -n 1 ../../espresso_jn/build/pypresso  bug_widom.py 
/home/pablo/espresso_jn/build/src/python/espressomd/reaction_methods.py:462: FutureWarning: arguments 'reactant_coefficients' and 'product_coefficients' are deprecated and are no longer necessary for the constant pH ensemble. They are kept for backward compatibility but might be deleted in future versions.
  warnings.warn(warn_msg, FutureWarning)
step  0
Before widom: a: 2 H: 2 Ha: 0 Na: 2 Cl: 2
total charge  0.0
After widom: a: 2 H: 1 Ha: 0 Na: 2 Cl: 2
total charge  -1.0

```

> 
> In the end, the bug needs to be removed with your PR.

it is  quite right that we need to solve it in another PR but, just to avoid confusion, my current PR is just on the refactoring of the MC function to move particles. I was waiting for your green light on my idea of checking the bookeeping list each time the `do_reaction` function is called as you mentioned that you wanted to discuss it with Rudolf in my issue, since it comes with an associated performance loss. 

Anyway, I will not oppose to merge this PR for this error message  since I agree with you that we need to solve the bug in another PR anyway.

reinaual (approved):

PR #4623, opened 2022-12-07
Implement override_cutoff_check property for vs.

pkreissl:
Under specific circumstances one might have to override the cutoff check for virtual sites.

E.g. on `n>1` cores, when using the hybrid decomposition scheme to simulate a large virtual sites rigid object such as a large raspberry particle, the `min_global_cut` necessary to set up said object becomes quite large. Say the raspberry is surrounded by lots of small particles, e.g. a polymer suspension, which are put in the regular child decomposition of the `HybridDecomposition` cell system, the resulting minimum cell size of the regular cell decomposition is limited by this large `min_global_cut` -- even if the raspberry central particle resides in the N-square child decomposition, where it already interacts with all particles making the `min_global_cut` requirement superfluous. However, one forfeits the possible computational advantage of the hybrid decomposition.

### DIFF
--- a/doc/sphinx/particles.rst
+++ b/doc/sphinx/particles.rst
@@ -354,6 +354,10 @@ Please note:
    you need to set the system's :attr:`~espressomd.system.System.min_global_cut`
    attribute to this largest distance.
    An error is generated when this requirement is not met.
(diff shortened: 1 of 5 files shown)


jngrad (approved):
LGTM

PR #4626, opened 2022-12-12
Dipole field calculation in parallel with direct sum

stekajack:
New feature: calculate and track the total dipole field for each particle, in parallel. This implementation is related to the recently merged magnetostatics actor DirectSumCpu that uses MPI and can use replicas to simulate open systems, with vacuum boundary conditions.

Description of changes:
- added feature `DIPOLE_FIELD_TRACKING` which implies `DIPOLES`
- in addition to `LongRangeForce` and `LongRangeEnergy`, there is now a `LongRangeField`, but only for DDS on CPU
- added particle property called `dip_fld` if `DIPOLE_FIELD_TRACKING` is defined
   - call `system.analysis.dipole_fields()` to update the particle fields
   - or use `espressomd.observables.ParticleDipoleFields()`


stekajack:
@RudolfWeeber managed to fix the compilation issue and tested quickly again. There was a combination of #ifdef incompatibility, std:tie use seems to be replaced across the board, gather_particle_data() return changed.  Don't know how to assign you as a reviewer, probably need write privileges to do that. Hopefully this is a good place to start.

jngrad (approved):

PR #4628, opened 2022-12-12
fixed system.part.pairs() 

### LINKED ISSUE #4622: Py: system.part.pairs() is wrong
if the particle numbering is not consecuitve (contains holes), this misses pairs.

Koppeprojects:
Fixes #4622

Description of changes:
- the `System.part.pairs()` method now returns the correct particle pairs when particle ids aren't both contiguous and starting from 0


jngrad (approved):
Thanks!

PR #4631, opened 2022-12-15
Improve testing and code coverage

jngrad:
Description of changes:
- improve unit testing of core functionality: P3M, MMM1D, OIF, virtual sites, script interface factory
- remove unreachable code and factor out code duplication
- convert exceptions that are unreachable from the public interface by assertions
- improve Sphinx and Doxygen documentation

### DIFF
--- a/testsuite/python/virtual_sites_relative.py
+++ b/testsuite/python/virtual_sites_relative.py
@@ -84,12 +84,22 @@ class VirtualSites(ut.TestCase):
         self.assertIsInstance(
             self.system.virtual_sites,
             espressomd.virtual_sites.VirtualSitesOff)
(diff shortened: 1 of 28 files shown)


testsuite/python/virtual_sites_relative.py
```diff
@@ -382,4 +394,4 @@ def test_zz_pressure_tensor(self):
 
 
 if __name__ == "__main__":
-    ut.main()
+    ut.main(verbosity=2)
```
jngrad, line 397: The point of this change is to make this test output more information when it times out. The test is flaky and times out at random quite infrequently in CI (not reproducible outside of CI).

reinaual (approved):

PR #4662, opened 2023-02-05
Add particle neighbor kernel method

### LINKED ISSUE #4620: Implement neighbor list loop
## Goal

* Implement a function template that calls a kernel on each particle, providing a list of neighboring particles
* To be used, e.g., for ML potentials, which calculate a descriptor based on a article's surrounding, and then use an ML model to obtain a force from that descriptor
* This is different from the `short_range_loop()` that calculates the LJ interaction
    * as it does not work on pairs, but for each particle provides a list of all short range neighbors
    * there is no assumption of pairwise symmetry ($f_{ab} = f_{ba}$)

## Sketch/example

```c++
template <typename Kernel>
particles_neighbor_list_loop(Kernel kernel) {...};

void add_force_ml(Particle& p, Utils::Span<Particle*> neighbors) {
auto descriptor = calculate_descriptor(neighbors);
auto force = ml_model(descriptor);
p.force() +=force;
}

particles_neighbor_list_loop(add_force_ml)
```

## Steps

* The function template that runs a kernel on all short range neighbors of a single particle is `CellStructure::run_on_particel_short_range_neighbors()`
* Use that to write a `Utils::Span<Particle*> get_particle_short_range_neighbors()`. See `src/core/cells.cpp:mpi_get_short_range_neighbors_local` for inspiration, omit the use of the optional, use a `Particle&` reference rather than a numerical particle id as input
* Write the `particles_neighbor_list_loop` template
    * Loop over all local particles `CellStructure::local_particles()` 
    * Obtain the span containing pointers to the neighbors using the function from the previous step
    * Call the kernel with the current particle and the neighborhood span.
* to test, write a kernel that fills a `std::vector<std::vector<int>>` with lists of particle ids, where the first id in the inner list is the particle for which the neighbor list is obtained and the remaining ints are the ids of the neighbors. See `src/core/cells:non_bonded_loop_trace()` for inspiration.

jhossbach:
Closes #4620

Description of changes:
- add a short-range loop method to generate a list of the particle ids of all particles within interaction range of a central particle (to be used in machine-learned potentials based on descriptors of a particle's neighborhood)
- expose `cell_structure.max_range()` to the python interface for unit testing
- refactor short-range loop methods (dependency inversion and separation of concerns)


src/core/cells.cpp
```diff
@@ -160,6 +166,24 @@ std::vector<int> mpi_get_short_range_neighbors(int const pid,
                   mpi_get_short_range_neighbors_local, pid, distance, false);
 }
 
+std::vector<Particle *> get_particle_neighbors_local(Particle const &p) {
```
jngrad:
```suggestion
static auto get_particle_neighbors(Particle const &p) {
```

* `static`: this function doesn't need to be visible outside of this source file, so marking it as `static` hides it and allows the compiler to inline it
* `auto`: the return type can be deduced since there is only one `return` statement
* `_local`: no longer needed, we have removed almost all `mpi_*_local` functions at this point, developers should now assume that core functions run in parallel by default
* the corresponding prototype can be removed from `cells.hpp`

src/core/cells.cpp
```diff
@@ -160,6 +166,24 @@ std::vector<int> mpi_get_short_range_neighbors(int const pid,
...
+  std::vector<Particle *> ret;
+
+  auto const distance = *boost::min_element(cell_structure.max_range());
+  double const cutoff2 = distance * distance;
```
jngrad:
```suggestion
  auto const cutoff2 = Utils::sqr(distance);
```

src/core/cells.cpp
```diff
@@ -160,6 +166,24 @@ std::vector<int> mpi_get_short_range_neighbors(int const pid,
...
+  detail::search_neighbors_sanity_check(distance);
+
+  auto kernel = [&ret, &cutoff2](Particle const &p1, Particle const &p2,
```
jngrad:
```suggestion
  auto const kernel = [&ret, cutoff2](Particle const &p1, Particle const &p2,
```

src/core/cells.cpp
```diff
@@ -160,6 +166,24 @@ std::vector<int> mpi_get_short_range_neighbors(int const pid,
...
+                                 auto vec) {
+    auto _p2 = const_cast<Particle *>(&p2);
```
jngrad: This will most likely invoke undefined behavior. Why not use `std::vector<Particle const *>` storage instead? Note how the `const` is on the left of the asterisk to signify it's a mutable value type (requirement of `std::vector`) that points to a `const Particle` (requirement of the short-range loop).

src/core/cells.cpp
```diff
@@ -160,6 +166,24 @@ std::vector<int> mpi_get_short_range_neighbors(int const pid,
...
 
+std::vector<Particle *> get_particle_neighbors_local(Particle const &p) {
+  std::vector<Particle *> ret;
+
+  auto const distance = *boost::min_element(cell_structure.max_range());
+  double const cutoff2 = distance * distance;
+  detail::search_neighbors_sanity_check(distance);
+
+  auto kernel = [&ret, &cutoff2](Particle const &p1, Particle const &p2,
+                                 auto vec) {
```
jngrad:
```suggestion
                                 auto const &vec) {
```
The `auto` keyword cannot resolve to a reference type, even when you pass a reference-valued type to the function. As a result, your original kernel would call the vector copy constructor ([live example](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIM6SuADJ4DJgAcj4ARpjEEgCspAAOqAqETgwe3r7%2ByanpAiFhkSwxcVyJdpgOGUIETMQEWT5%2BAVU1AnUNBEUR0bEJtvWNzTltwz2hfaUDFQCUtqhexMjsHOYAzKHI3lgmG24AbtVExPvYJhoAgpc3ZltUWFQA1G4A8soAmgD67%2BFCABUAEpyNwAt5A25MLxEZ78VAQaGwtAMBQEZ4AegAVOYAGxYjFUObPEwAdis12eVOexEwBGWDDhADo0gAvTAQOb7Ck3UkAEVum0wtCUUJhqDhqARSIlKLRJLMuKJJPJt2pNLpDOZbI5XI2PLJAuuQoY%2BCoguuMueUQaEDR6BAIGODhI%2BzcoQI52ecvReOdxLJPPVtPpxEZ8Ig/u5gv5HAWtE48V4fg4WlIqE4bms1meCiWK0wCo2PFIBE0cYWAGsQPENPpOJJk%2BX05xeAoQHWy6m46Q4LAYIgUKgWEk6LFyJQ0COx3FgFwzAEsIc8KsAGp4TAAdzeSUYnBLNFoBFi7YgUWbUVCDQAnvveJfmMRr28otoTnfSFO2II3gxaLfu1ILAWEMYBxEA/BaRqY520AzBVGqGE1hLD1hWbWg8CiYgbw8LBmwIYg8BYO8FioAxgAUdctx3PduF4fhBBEMR2CkGRBEUFR1EA3QuH0UCUGzSx9Ew9tIAWVAkkcARYIAWgAdTEWhnnk%2BCCKYZT7X2PlTEsawzA0ZS3mAZS%2BXeL5fjef5gVBcEgTbYUTgyFxTVGPxeOCKYSjKPQUjSKTMk8FofPyfzei82ZbAcjoGC6EZApyXj2n82LJmKfpyiGbpXL0NFujC9KJAWPNllWQr6w4JNSBTNMMw4Z5VAADlxGTcUkZ5gGQZBnnnJkzGeCBcEIEgiy4OZeC7LQ5irGs6wTDhG1IYjayq5tarbDtS3LBY%2B0HKdR3oMgKAgPaZxAOcFyAzBlzXDdt13FMDzoY9iFPc9AIfG8Pw%2Bp8XzfBwPy/RgCF/f9m2A0DwLTSDHJg5t4MQ48P1Qua0wwrCcIwNY0wIoiSL4cjKNumiHvo2QmPEViGPkJQ1GbXQAgMIwBN0oS0dEzl00kjJZIU2glJU1Q1I0gh0C0nSrEsfTDOAez7H85z3HitzAlNfKZgy3yCgC7Jlc10LPIKxKouSiZsqNuXagmNXvMS02lZyq2DfVsrioLMq5sq6reFq%2BqmpatqOq6nq%2BoG/BThGsbNu7KbyoWpa6y9lsOHWzstum5a5o2JtALWqPJvjTgzGzmrWzzitSGOF6nMkIA%3D)) in the hot loop.

src/core/cells.cpp
jngrad: As a side-note, you could consider the following construction:
```c++
  auto out_ids = std::back_inserter(ret);
  auto const kernel = [&out_ids, cutoff2](int i) {*out_ids = i;};
```
This make the intent clear that your lambda can only push new elements to the vector, and not modify existing elements, or clear all elements, or call the vector destructor.

src/core/cells.cpp
```diff
@@ -192,6 +216,30 @@ std::vector<PairInfo> non_bonded_loop_trace() {
...
+template <typename Kernel>
+void run_on_particle_neighbors(Particle const &p, Kernel kernel) {
+  std::vector<Particle *> const neighbors = get_particle_neighbors_local(p);
+  kernel(p, neighbors);
+}
+
+std::vector<NeighborPIDs> get_neighbor_pids() {
+  std::vector<NeighborPIDs> ret;
+  auto pid_kernel = [&ret](Particle const &p,
+                           std::vector<Particle *> const neighbors) {
```
jngrad:
```suggestion
                           std::vector<Particle *> const &neighbors) {
```

jngrad: General question: do you currently return a vector of pointers, because you would like to add more kernels in the future? If not, you could directly return a vector of `int` from the hot-loop kernel.

src/core/cells.cpp
```diff
@@ -192,6 +216,30 @@ std::vector<PairInfo> non_bonded_loop_trace() {
...
+    for (auto const &neighbor : neighbors) {
+      neighbor_pids.emplace_back(neighbor->id());
+    }
+    ret.emplace_back(p.id(), neighbor_pids);
+  };
+  on_observable_calc();
+  for (auto const &p : cell_structure.local_particles()) {
+    run_on_particle_neighbors(p, pid_kernel);
+  }
+  Utils::Mpi::gather_buffer(ret, comm_cart);
```
jngrad: Maybe this gather operation belongs to the script interface. We should try to keep the core fully MPI-parallel and deal with MPI reductions in the interface. The interface has its own communicator, obtained via a call to `context()->get_comm()`. The `comm_cart` global variable used here will be removed in the near future.

src/core/cells.hpp
```diff
@@ -119,6 +119,41 @@ boost::optional<std::vector<int>>
 mpi_get_short_range_neighbors_local(int pid, double distance,
                                     bool run_sanity_checks);
 
+/**
+ * @brief Get all neighbors of a particle.
+ *
+ * Uses cell_structure.max_range() to identify all neighbors of the particle.
```
jngrad: Consider using Doxygen `@ref` with the proper class method namespace resolution to create a link. This way, if the method name or namespace changes, the CI will notify us this docstring has become outdated.

src/core/cells.hpp
```diff
@@ -119,6 +119,41 @@ boost::optional<std::vector<int>>
...
+ *
+ * @param p Particle const &
```
jngrad: Avoid copy-pasting the argument type as description, since Doxygen already displays the argument type. Either write a description, or don't write `@param` if the argument is self-explanatory.

src/core/cells.hpp
```diff
@@ -119,6 +119,41 @@ boost::optional<std::vector<int>>
...
+ *
+ * @return vector of pointer of all neighboring particles.
+ */
+std::vector<Particle *> get_particle_neighbors_local(Particle const &p);
```
jngrad: Can probably be removed, this is an implementation detail.

src/core/cells.hpp
```diff
@@ -119,6 +119,41 @@ boost::optional<std::vector<int>>
...
+
+/**
+ * @brief Run kernel on all neighbors of a particle.
+ *
+ * @param p Particle const &
+ * @param kernel Function with signature <tt>(Particle const &,
+ *		 std::vector<Particle *>)</tt>
+ */
+template <typename Kernel>
+void run_on_particle_neighbors(Particle const &p, Kernel kernel);
```
jngrad: Remove this prototype and move the docstring to `cells.cpp`.

Function templates rarely need a prototype. It can actually be dangerous to have a prototype, because multiple translation units can write calls to instances of the function template, and the developer must guarantee that there is a translation unit where these template instances exist, and that this translation unit is linked against all translation units where the template instances are called.

src/core/cells.cpp
```diff
@@ -91,6 +91,12 @@ std::vector<std::pair<int, int>> get_pairs_filtered(double const distance,
 
 namespace boost {
 namespace serialization {
+template <class Archive>
+void serialize(Archive &ar, NeighborPIDs &n, const unsigned int /* version */) {
+  ar &n.pid;
+  ar &n.neighbor_pids;
+}
```
jngrad: This code is never called. Consider running the test case in parallel with:
```diff
diff --git a/testsuite/python/CMakeLists.txt b/testsuite/python/CMakeLists.txt
index e878840e1..b4b408242 100644
--- a/testsuite/python/CMakeLists.txt
+++ b/testsuite/python/CMakeLists.txt
@@ -242,3 +242,3 @@ python_test(FILE observable_cylindricalLB.py MAX_NUM_PROC 2 LABELS gpu)
 python_test(FILE analyze_chains.py MAX_NUM_PROC 1)
-python_test(FILE analyze_distance.py MAX_NUM_PROC 1)
+python_test(FILE analyze_distance.py MAX_NUM_PROC 2)
 python_test(FILE analyze_acf.py MAX_NUM_PROC 1)
```

src/python/espressomd/analyze.py
```diff
@@ -192,6 +192,16 @@ class Analysis(ScriptInterfaceHelper):
         (N,) array_like of :obj:`int`
             The neighbouring particle ids.
 
+    particle_neighbor_pids()
+        Get a dictionary, where each key-value-pair represents a particle
+        and it's respective neighbors.
```
jngrad:
```suggestion
        Get a list of all short-range neighbors for each particle.
```
To avoid repetition with the return paragraph.

src/script_interface/analysis/Analysis.cpp
```diff
@@ -81,6 +82,15 @@ Variant Analysis::do_call_method(std::string const &name,
     auto const local = particle_short_range_energy_contribution(pid);
     return mpi_reduce_sum(context()->get_comm(), local);
   }
+  if (name == "particle_neighbor_pids") {
+    auto const neighbor_pids = get_neighbor_pids();
+    std::unordered_map<int, std::vector<int>> dict;
+    std::for_each(neighbor_pids.begin(), neighbor_pids.end(),
+                  [&dict](NeighborPIDs const &neighbor_pid) {
+                    dict[neighbor_pid.pid] = neighbor_pid.neighbor_pids;
+                  });
```
jngrad:
```suggestion
    std::unordered_map<int, std::vector<int>> dict;
    context()->parallel_try_catch([&]() {
      auto const neighbor_pids = get_neighbor_pids();
      std::for_each(neighbor_pids.begin(), neighbor_pids.end(),
                    [&dict](NeighborPIDs const &neighbor_pid) {
                      dict[neighbor_pid.pid] = neighbor_pid.neighbor_pids;
                    });
    });
```
Because the core function can throw an exception, we need a parallel try catch block.

jngrad (changes requested):
Overall, looks good to me. Please find below suggestions to improve performance and design.

src/core/cells.cpp
jhossbach: ~The goal is to be able to change e.g. the force of neighboring particles with this method. I am removing the const qualifier from `short_range_neighbor_loop` instead.~
We can change it to be `std::vector<Particle const *>)`.

src/core/cells.cpp
jhossbach: Yes, the intent of `run_on_particle_neighbors` is to be able to find properties of neighboring particles, such as charge etc.

src/core/cells.cpp
jhossbach: `Utils::Mpi::gather_buffer` does not take a `const boost::mpi::communicator &` which is what `context()->get_comm()` provides, it takes `boost::mpi::communicator`. Any reason why the `comm` should not be a const reference?

jhossbach:
I have added your suggestions @jngrad, please re-review.
Some details:
 - Removed the prototype of `run_on_particle_neighbors` from `cells.hpp` and moved the docstring to the cpp file
 - Applied your suggestions to `get_particle_neighbors_local`
   - This method is not needed outside of cells.cpp, so I removed the .hpp declaration and moved the docstring to the definition in the .cpp file
 - Moved the `gather_buffer` operation to the script interface and added the `serialize` method for the custom `NeighborPIDS` struct to `script_interface/analysis/Analysis.cpp`
   - I had to change `gather_buffer` in `src/utils/include/utils/mpi/gather_buffer.hpp` to take a const reference of the comm

jngrad (approved):
Thanks!

reinaual (approved):

Issue #4617, opened 2022-11-27
Prototype reaction method in Python


RudolfWeeber:
* MC methods in Python would be much easier to develop\
* MC methods probably do not benefit form parallelization, as the sytem is only locally modified most of the time
* However, as we are remoming mpi callbacks, it would have to be parallelized if remainign in the core.
* this is non-trivial.
* Time consuming tasks like energy calculaiton will still be done by the core
* The current C++ implementaion is not parallelized and essentially calls the same core funcitons, the Python script would use.

To assess the performance cost, a reaction ensemble simulation of a polymer with dissociating ions (HA -> H+ +A- and back reaction) should be written in Python.

## Notes

* energy calculation should be done via the core (`system.analysis.energy()`, in a later stage, this can be made mor specific, e.g., to omit kinetic energy.
* the methods needs a list of ids for each particle type. For the prototype, this shuold be constructed once and updated on the fly. Later an effiient implementaiton in the core can be added.

    ```python
    all_types = np.unique(system.part.all().type())
    ids_by_type = {t: system.part.select(type=t).id.shape) for t in all_types} # maybe shape[0]
    ```

* Exclusion radius checking can be done with `system.analysis.nbhood()` We need to make usre, the O(1)  implementaiton is used.
* If this is significantly slower than the current core implementaiton, bottlenecks can be identified with a line-by-line profiling: https://stackoverflow.com/questions/3927628/how-can-i-profile-python-code-line-by-line

Issue #4614, opened 2022-11-26
Get rid of mpi callbacks


RudolfWeeber:
* To allow for several `espressomd.System` instances in a single script
* to allow for shared-memory and/or task-based parallelism
* to get rid of most Cython code

## Cell system / pair finding related

* [x] cells.cpp:mpi_get_short_range_neighbors_local
* [x] EspressoSystemStandAlone.cpp:mpi_set_node_grid_local
* [x] integrate.cpp:mpi_set_skin_local
 
## Analysis and status reporting

* [x] dpd.cpp:dpd_viscous_stress_local
* [x] energy.cpp:REGISTER_CALLBACK_MAIN_RANK(calculate_energy)
* [x] pressure.cpp:REGISTER_CALLBACK_MAIN_RANK(calculate_pressure)
* [ ] errorhandling.cpp:mpi_gather_runtime_errors_local
 
## System and interaction properties
 
* [x] EspressoSystemStandAlone.cpp:mpi_set_box_length_local
 
## Propagation, equations of motion
 
* [x] EspressoSystemStandAlone.cpp:mpi_set_time_step_local
* [x] integrate.cpp:REGISTER_CALLBACK_MAIN_RANK(mpi_steepest_descent_local)
* [x] integrate.cpp:REGISTER_CALLBACK_MAIN_RANK(integrate)
* [x] thermostat.cpp:  mpi_##thermostat##_set_rng_seed
* [x] thermostat.cpp:  mpi_##thermostat##_set_rng_counter
* [x] thermostat.cpp:mpi_set_brownian_gamma_local
* [x] thermostat.cpp:mpi_set_brownian_gamma_rot_local
* [x] thermostat.cpp:mpi_set_langevin_gamma_local
* [x] thermostat.cpp:mpi_set_langevin_gamma_rot_local
* [x] thermostat.cpp:mpi_set_thermo_virtual_local
* [x] thermostat.cpp:mpi_set_temperature_local
* [x] thermostat.cpp:mpi_set_thermo_switch_local
* [x] thermostat.cpp:mpi_set_nptiso_gammas_local
 
## Interactions
 
* [x] nonbonded_interactions/nonbonded_interaction_data.cpp:mpi_realloc_ia_params_local
 
## Particles management
* [x] particle_data.cpp:mpi_send_update_message_local
* [x] particle_data.cpp:mpi_remove_exclusion_local
* [x] particle_data.cpp:mpi_add_exclusion_local
[...]

PR #4666, opened 2023-02-10
Make reaction methods run in parallel

jngrad:
Partial fix for #4614
Partial fix for #4617
Follow-up to #4637

Description of changes:
- reaction methods are now fully parallel and no longer rely on MpiCallbacks
- reaction methods are no longer entangled with the particle management code
- the Monte Carlo acceptance probability is now implemented in Python
- API changes:
   - it is no longer possible to change the reaction constant of an existing reaction with `gamma <= 0`
   - `ReactionAlgorithm.reaction()` now takes `steps` instead of `reaction_steps` as argument
   - when setting up a MC method with two or more reactions, a runtime error is raised if a reaction accidentally overwrites the default charge of a specific type with a different value
- under-the-hood changes:
   - the `particle_data.cpp` source file was removed
   - the `EspressoSystemStandAlone` class no longer relies on MpiCallbacks
   - frequency of cell system invalidation during Monte Carlo trial moves has been significantly reduced
   - the new Python interface of reaction methods is around 40% slower than the original C++ implementation


src/python/espressomd/reaction_methods.py
```diff
@@ -374,26 +399,122 @@ def _check_charge_neutrality(self, type2charge, reaction):
...
+            reaction = {"reactant_coefficients": reaction.reactant_coefficients,
+                        "reactant_types": reaction.reactant_types,
+                        "product_types": reaction.product_types,
+                        "product_coefficients": reaction.product_coefficients,
+                        "reactant_types": reaction.reactant_types,
+                        "gamma": reaction.gamma}
             reactions_list.append(reaction)
 
-        return {"reactions": reactions_list, "kT": self.kT, 
+        return {"reactions": reactions_list, "kT": self.kT,
```
pm-blanco, line 511: I like this new feature, but it should also return `exclusion_radius_per_type` for completeness

src/python/espressomd/reaction_methods.py
```diff
@@ -374,26 +399,122 @@ def _check_charge_neutrality(self, type2charge, reaction):
...
 
+    def _rebuild_reaction_cache(self):
+        """
+        Maintain a local cache of the Python representation of the
+        script interface reactions. This helps reducing the overhead
+        when calculating the acceptance probability.
+        """
+        self._reactions_cache = [x for x in self.reactions]
+
+    def reaction(self, reaction_steps):
```
pm-blanco: I think that `reaction_steps` should be renamed to `steps`, as it was in the current version. I think that  in this case the change in the naming convention does not bring any clarification and will probably only annoy the users.

src/core/reaction_methods/ReactionAlgorithm.hpp
```diff
@@ -166,37 +135,86 @@ class ReactionAlgorithm {
...
+  /**
+   * Accept or reject moves made by @ref generic_oneway_reaction_part_1 based
+   * on a probability acceptance @c bf.
+   * @returns Potential energy of the system after the move was accepted or
+   * rejected.
+   */
+  double generic_oneway_reaction_part_2(int reaction_id, double bf,
+                                        double E_pot_old, double E_pot_new);
+  /**
+   * Attempt a global MC move for particles of a given type.
```
pm-blanco: I would generally avoid to use the terminology "global" MC move in the docs because it is ambiguous (what does "global" refer to? to do MC move around the globe? :)) This also applies to  ` make_displacement_mc_move_attempt`

src/core/reaction_methods/ReactionAlgorithm.cpp
```diff
@@ -196,107 +177,102 @@ ReactionAlgorithm::make_reaction_attempt(SingleReaction const &reaction) {
...
-  return old_particle_numbers;
+  return particle_numbers;
 }
 
-void ReactionAlgorithm::generic_oneway_reaction(
-    SingleReaction &current_reaction, double &E_pot_old) {
-
-  current_reaction.tried_moves += 1;
+std::optional<double>
+ReactionAlgorithm::generic_oneway_reaction_part_1(int reaction_id) {
```
pm-blanco: I think that we should reconsider the name of the methods `generic_oneway_reaction_part_1` and `generic_oneway_reaction_part_2`, since they do not provide much information about what is going on on them and I think that they will be confusing for future developers navigating the code. Since this two methods are technically new anyway, I propose that we completely rename them. I suggest the following:
-  `generic_oneway_reaction_part_1` becomes `create_new_trial_state`: generates a new trial state following the reaction given by `reaction_id` and returns the new potential energy of this trial state
- `generic_oneway_reaction_part_2` becomes `attempt_MC_reaction_move`:  accepts or rejects the MC reaction move with a  `bf` probability. If rejected, restores the espresso system to its original state.

pm-blanco (commented):
Thank you again @jngrad for all your hard work on this PR. I think that it looks good to me, except some specific comments that I leave here. Since this PR comes with quite some changes in the code, I tested it for the case of a oligopeptide for which we have published reference data (See fig. attached). The results look good to me within the statistics (the simulations are rather short), so on my side the PR can be merged after addressing my other comments. 
test_peptide_short

pm-blanco (dismissed):
Looks good to me

src/python/espressomd/reaction_methods.py
```diff
@@ -374,26 +399,122 @@ def _check_charge_neutrality(self, type2charge, reaction):
...
-                        "reactant_types": core_reaction.reactant_types,
-                        "gamma": core_reaction.gamma}
+        for reaction in self.reactions:
+            reaction = {"reactant_coefficients": reaction.reactant_coefficients,
+                        "reactant_types": reaction.reactant_types,
+                        "product_types": reaction.product_types,
+                        "product_coefficients": reaction.product_coefficients,
+                        "reactant_types": reaction.reactant_types,
+                        "gamma": reaction.gamma}
             reactions_list.append(reaction)
```
reinaual: I personally don't like the name-overriding of variables like here

reinaual (dismissed):

src/python/espressomd/reaction_methods.py
jngrad: Nice catch. It's quite dangerous. Not to mention the W0109 pylint violation (key `"reactant_types"` appears twice) and the amount of copy-pasting that could be elided with a `getattr` in a list comprehension.

reinaual (approved):

pm-blanco (approved):

PR #4678, opened 2023-02-23
Enable P3M for hybrid decomposition with one MPI node

RiccardoFrenner:
Description of changes:
- removed sanity checks preventing hybrid decomposition from being used with P3M/Dipolar P3M on a single MPI node
- added tests to ensure forces/energies of the hybrid decomposition with P3M are equal to those calculated using the regular decomposition


jngrad (approved):
Thanks!

PR #4680, opened 2023-03-05
Fix bugs in testcases.

pkreissl:
Description of changes:
- fix feature guards in the testsuite

To reproduce the errors that this PR fixes, build without `EXTERNAL_FORCES` and `DPD`, respectively. E.g. use only the following features in your `myconfig.hpp`:
```c++
#define ROTATION
#define ROTATIONAL_INERTIA
#define MASS
#define THERMOSTAT_PER_PARTICLE
#define LENNARD_JONES
#define VIRTUAL_SITES_RELATIVE
#define PARTICLE_ANISOTROPY
```


jngrad (approved):
Thanks!

PR #4683, opened 2023-03-09
Fix operator transformation from body frame to lab frame

### LINKED ISSUE #4667: Langevin thermostat with anisotropic friction is broken
Brownian themorstat is fine, though
```python
import numpy as np
import espressomd

use_brownian = False

system.time_step = 0.1
system.cell_system.skin = 0.4
if use_brownian:
    system.thermostat.set_brownian(kT=0.0, gamma=1.0, seed=42)
    system.integrator.set_brownian_dynamics()
else:
    system.thermostat.set_langevin(kT=0.0, gamma=1.0, seed=42)

ext_force = 1.5
gamma_parallel = 0.8
gamma_ortho = 0.2
mass = 0.05
part = system.part.add(pos=3 * [0], 
                       gamma=[gamma_ortho, gamma_ortho, gamma_parallel],
                       mass=mass)

# check particle orientation unchanged
part.director = [0, 0, 1]
part.ext_force = [ext_force, 0, ext_force]
system.integrator.run(1000)
np.testing.assert_allclose(
    part.v, [ext_force / gamma_ortho, 0, ext_force / gamma_parallel]
)

# check after rotation: switch parallel and perpendicular
part.director = [1, 0, 0]
part.ext_force = [ext_force, 0, ext_force]
system.integrator.run(10000)
np.testing.assert_allclose(
    part.v, 
    [ext_force / gamma_parallel, 0, ext_force / gamma_ortho], 
    atol=1e-10
)

# check noise
kT = 0.1234
if use_brownian:
    system.thermostat.set_brownian(kT=kT, gamma=1.0, seed=42)
    system.integrator.set_brownian_dynamics()
else:
    system.thermostat.set_langevin(kT=kT, gamma=1.0, seed=42)
    
part.ext_force = 3 * [0.]
part.rotation = 3*[False]

def check_temperature():
    n_samples = 10000
    steps_per_sample = 3

    vels = []
    for _ in range(n_samples):
        system.integrator.run(steps_per_sample)
        vels.append(np.copy(part.v))
    
    # https://en.wikipedia.org/wiki/Thermal_velocity
    # use 1D-Formula for each direction separately 
    thermal_vel = np.sqrt(np.mean(np.array(vels)**2, axis=0))
    kT_measured = mass * thermal_vel**2

    np.testing.assert_allclose(kT_measured,kT, rtol=3e-2)

part.director = [0,0,1]
check_temperature()
part.director = [0,1,0]
check_temperature()
```

I only checked friction. 
This https://github.com/espressomd/espresso/blob/6d08b88b3b1cf414deb046cc0562e7faa065f312/src/core/thermostats/langevin_inline.hpp#L84 also looks wrong to me. The friction force is converted from body to lab frame, but the noise is not.

EDIT:
Added temperature check that fails due to not-body-to-lab-converted noise prefactor

mebrito:
Partially fixes #4667

Description of changes:
- Fix operator transformation
- Add unit test

### DIFF
--- a/src/core/unit_tests/rotation_test.cpp
+++ b/src/core/unit_tests/rotation_test.cpp
@@ -189,6 +189,38 @@ BOOST_AUTO_TEST_CASE(propagate_omega_quat_particle_test) {
   }
 }
 
+BOOST_AUTO_TEST_CASE(convert_operator_body_to_space_test) {
+  auto constexpr sqrt_2_half = Utils::sqrt_2() / 2.0;
+  // rotation around z-axis by pi/2
+  Utils::Quaternion<double> const quat = {sqrt_2_half, 0.0, sqrt_2_half, 0.0};
+  // rotation around z-axis by pi/4
+  Utils::Matrix<double, 3, 3> const linear_transf_body = {
+      {sqrt_2_half, -sqrt_2_half, 0.0},
+      {sqrt_2_half, sqrt_2_half, 0.0},
+      {0.0, 0.0, 1.0}};
+  // rotation around x-axis by pi/4
+  Utils::Matrix<double, 3, 3> const linear_transf_space_ref = {
+      {1.0, 0.0, 0.0},
+      {0.0, sqrt_2_half, -sqrt_2_half},
+      {0.0, sqrt_2_half, sqrt_2_half}};
+
+  auto const linear_transf_space =
+      convert_body_to_space(quat, linear_transf_body);
+
+  for (int i = 0; i < 3; i++) {
+    for (int j = 0; j < 3; j++) {
+      if (linear_transf_space_ref(i, j) == 0.0) {
+        BOOST_CHECK_SMALL(
+            std::abs(linear_transf_space(i, j) - linear_transf_space_ref(i, j)),
+            tol);
+      } else {
+        BOOST_CHECK_CLOSE(linear_transf_space(i, j),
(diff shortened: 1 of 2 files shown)


jngrad (approved):
Thanks!

src/core/unit_tests/rotation_test.cpp
```diff
@@ -189,6 +189,38 @@ BOOST_AUTO_TEST_CASE(propagate_omega_quat_particle_test) {
   }
 }
 
+BOOST_AUTO_TEST_CASE(convert_operator_body_to_space_test) {
+  auto constexpr sqrt_2_half = Utils::sqrt_2() / 2.0;
+  // rotation around z-axis by pi/2
```
christophlohrmann, line 194: sorry that I am late but what does rotation around z-axis mean here?
I do not get the test

jngrad: It's a typo in the comment. The quaternion actually describes a rotation around the y-axis. This can be fixed as part of the 4.2.1 release spellcheck or as part of the fix for #4667. The idea behind this test is to make sure the old behavior is not re-introduced by mistake. As I understood it , the behavior in ticket #4667 is actually caused by two bugs: the rotation matrix was incorrectly applied (the particle quaternion describes the rotation from the z-axis to the director, so we need to revert the direction) and the prefactor was not transformed.

PR #4691, opened 2023-03-21
Improve freeglut support in OpenGL visualizer

jngrad:
Description of changes:
- always check for the presence of GLE functions before invoking them


jngrad:
This PR successfully fixes the bug reported by Le Qiao on the mailing list thread [problem with simplepore live visualization](https://lists.nongnu.org/archive/html/espressomd-users/2023-03/msg00008.html).

reinaual (approved):

PR #4693, opened 2023-03-22
core: partially inline LB particle coupling

RudolfWeeber:
Fixes #

Description of changes:
-


RudolfWeeber:
I'm not continuing on this. The changes are too far reahing fo r a bugfix release, considering this concerns only performance.
Also, it would have to be re-done for hte Walberla branch.

PR #4698, opened 2023-03-29
Fix issues with cluster analysis and chain analysis

### LINKED ISSUE #4689: bug: analysis.calc_rg and analysis.calc_re break when particles are in different periodic images
While debugging a student code I found out the following behavior which looks like a bug to me. If one initializes the system with particles in different periodic images (for example some inside the simulation box and some in the next periodic image) and one calculates the radius of gyration or end-to-end distance, one obtains a value that disregards the periodicity of the system. Here you can find a minimal script that reproduces the bug:

```python3
import espressomd

box_l=20
bug=True # Switch on/off the bug

espresso_system=espressomd.System(box_l=[box_l]*3, periodicity=[True,True,True])

if bug:
    # Initialize the system with one particle inside the simulation box and another in a periodic image
    # Both particles are separated only by a distance of 0.1 by PBC
    espresso_system.part.add(id=0, pos = [0.1,0,0])
    espresso_system.part.add(id=1, pos = [box_l+0.2,box_l,box_l])
else:
    # Initialize the system with both particles inside the same periodic image
    espresso_system.part.add(id=0, pos = [box_l+0.1,box_l,box_l])
    espresso_system.part.add(id=1, pos = [box_l+0.2,box_l,box_l])

dist = espresso_system.distance(espresso_system.part.by_id(0),espresso_system.part.by_id(1))
Rg = espresso_system.analysis.calc_rg(chain_start=0, number_of_chains=1, chain_length=2)
Re = espresso_system.analysis.calc_re(chain_start=0, number_of_chains=1, chain_length=2)
print(f'The distance between the two particles is {dist:.4g}')
print(f'Their end-to-end distance is {Re[0]:.4g} and their radius of gyration is {Rg[0]:.4g}')
```
Actual output:
```
The distance between the two particles is 0.1
Their end-to-end distance is 34.7 and their radius of gyration is 17.35
```
Expected output:
```
The distance between the two particles is 0.1
Their end-to-end distance is 0.1 and their radius of gyration is 0.05
```

jngrad:
Fixes #4689

Description of changes:
- disable cluster analysis methods when Lees-Edwards boundary conditions are used
- document caveats from chain analysis methods
- cleanup doxygen and comment blocks

### DIFF
--- a/doc/sphinx/analysis.rst
+++ b/doc/sphinx/analysis.rst
@@ -295,57 +295,16 @@ assuming the ``image_box`` values were properly set up. This is important to
 consider when setting up systems where the polymers are much longer than the
 box length, which is common when simulating polymer melts.
 
-.. _End to end distance:
+.. _Available chain analysis functions:
 
-End-to-end distance
-^^^^^^^^^^^^^^^^^^^
-:meth:`espressomd.analyze.Analysis.calc_re`
+Available chain analysis functions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Returns the mean end-to-end-distance and its standard deviation,
-as well as its mean squared value over all chains.
+* :meth:`espressomd.analyze.Analysis.calc_re`: average end-to-end-distance
 
-.. _Radius of gyration:
+* :meth:`espressomd.analyze.Analysis.calc_rg`: average radius of gyration
 
-Radius of gyration
-^^^^^^^^^^^^^^^^^^
-:meth:`espressomd.analyze.Analysis.calc_rg`
-
-Returns the radius of gyration averaged over all chains.
-It is a radius of a sphere, which would have the same moment of inertia as the
-molecule, defined as
-
-.. math::
-
-   \label{eq:Rg}
-   R_{\mathrm G}^2 = \frac{1}{N} \sum\limits_{i=1}^{N} \left(\vec r_i - \vec r_{\mathrm{cm}}\right)^2\,,
-
-where :math:`\vec r_i` are position vectors of individual particles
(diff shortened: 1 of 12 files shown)


jngrad:
@RudolfWeeber @reinaual @pkreissl @bindgens1
This PR disables cluster analysis methods when LEbc are active.
The faulty behavior can be reproduced with the MWE in #4699.

doc/sphinx/analysis.rst
```diff
@@ -266,22 +266,43 @@ Chains
...
 
 .. _End to end distance:
 
 End-to-end distance
 ^^^^^^^^^^^^^^^^^^^
 :meth:`espressomd.analyze.Analysis.calc_re`
 
-Returns the quadratic end-to-end-distance and its root averaged over all chains.
+Returns the mean end-to-end-distance and its standard deviation,
+as well as its mean squared value over all chains.
```
reinaual: Apparently it was wrong before, but to me it's still not obvious now that it returns 4 values, as it is stated in https://github.com/espressomd/espresso/blob/f5bbb2408f3bce6e444ce9ed5bbd254a6cd25a4d/src/python/espressomd/analyze.py#L225-L228

reinaual (dismissed):

doc/sphinx/analysis.rst
jngrad: Nice catch. Actually, none of the chain analysis function descriptions in the user guide match the corresponding Sphinx docstrings. There is also a lot of copy-pasting involved. Maybe it's best to simply move these user guide paragraphs to the corresponding docstrings, just like we did with the `Observable` classes.

reinaual (approved):

Issue #4542, opened 2022-07-22
De-cythonize ESPResSo


jngrad:
**tl;dr:** Most Cython files in ESPResSo can be rewritten as Python files using the `ScriptInterface` framework.

## Background and problem statement

The Python interface of ESPResSo was originally designed using Cython, which could access C++ header files from the core and generate all the necessary glue code to connect the C++ shared objects to the Python interpreter. Cython is a convenient solution, and its syntax is easy to understand to contributors who have some Python experience but no C++ experience.

However, this design decision comes with several drawbacks that are not immediately visible to occasional contributors:
* C++ files automatically generated from Cython files are several thousand lines long
   * they are tedious to debug in GDB
   * they take a long time to compile in CI
   * they trigger many compiler warnings and diagnostics due to how Cython v0.29 sometimes rely on sloppy (yet harmless) C++ constructs, e.g. unused variables, undocumented switch fallthroughs, etc.
   * exceptions in Cython code don't produce a useful traceback: only the last frame is shown (as a filename and line number)
* removing global variables from the core is difficult, since they are used in the Cython code
* Cython classes only run on the head node, which in turn requires introducing MPI callbacks in the core
* Cython classes cannot easily "own" a resource (like an electrostatic actor)
   * the ownership information is necessary to control the lifetime of an object
   * storing a smart pointer to the owned resource in the Cython class requires a lot of knowledge about the internal mechanism for Cython/Python object construction, and makes pickling in a checkpoint file quite difficult

## Proposed solution

The `ScriptInterface` framework was introduced to automatically generate the glue code between a Python class and a C++ class that holds a shared pointer to the owned resource. Although it requires writing a C++ class, there are many benefits:
* clear separation of concerns in the core, with code split between public header files and private source files
* no need for MPI callbacks anymore: the C++ interface class is already running in parallel, which mean standard `boost::mpi` code can be used instead of our custom `MpiCallbacks` framework
* no need to use `utils.check_type_or_throw_except()` or `utils.is_valid_type()` anymore: this information is already encoded in the C++ class when using e.g. `auto const value = get_value<Utils::Vector3d>(value)`, and a clear error message is generated automatically if the conversion failed
* uniform interface for class methods: all methods use `**kwargs`, which reduces the risk of introducing silent API changes
* no need to write `__getstate__()` and `__setstate__()` methods to checkpoint Python objects: they are generated automatically

[...]

PR #4713, opened 2023-04-21
Rewrite integrators script interface as Python code

### LINKED ISSUE #4712: On releasing the GIL in ESPResSo
Here is the definition of the global interpreter lock (GIL) in the Cython documentation ([link](https://cython.readthedocs.io/en/latest/src/userguide/glossary.html#term-Global-Interpreter-Lock-or-GIL)):

> A lock inside the Python interpreter to ensure that only one Python thread is run at once. This lock is purely to ensure that race conditions do not corrupt internal Python state. Python objects cannot be manipulated unless the GIL is held. It is most relevant to Cython when writing code that should be run in parallel. If you are not aiming to write parallel code then there is usually no benefit to releasing the GIL in Cython.

There is one place in the ESPResSo code where the GIL is released: the `integrator.run()` method, if the integrator is not steepest descent. While in principle we could release the GIL everywhere by making the script interface `call_method()` run within a `nogil` context, this doesn't lead to performance improvements (tested with the testsuite on a 16-core AMD Ryzen). It is however useful for integrators, since we often run them concurrently with a python thread to update the OpenGL visualizer scene in real time; in that case, not releasing the GIL leads to a serious drop in frames per second.

For historical reasons, the integrator script interface classes had to rely on custom and poorly-documented Cython code to call the C++ method with `nogil` while preserving both C++ exceptions and interrupt signals. Now that the development branch of ESPResSo requires Python 3.9, we can replace the Cython signal handling code by regular Python code. In addition, we now require a Cython version that no longer requires `nogil` functions to be `noexcept`, which means we can use the `call_method()` instead of re-implementing it in `integrate.pyx`. We just need to keep in mind that `except *` is not recommended due to its prohibitive cost (the GIL is re-acquired for each statement in the `nogil` context), while all other exception specifiers come without overhead. The body of a `nogil` context cannot contain Python objects, because implicit coercion of Python types to C++ types requires calls to Cython functions that can throw Python exceptions, which is unsafe without the GIL (potential for race conditions).

jngrad:
Fixes #4712

Description of changes:
- rewrite integrator as Python (partial fix for #4542)
- remove two MPI callbacks (partial fix for #4614)
- remove a global variable (partial fix for #2628)


jngrad (commented):
There is no code coverage in the core for the signal handling code, probably because the interrupt signal interferes with the profiler. But the signal error message is actually tested in the python testcase now. Also, another benefit of no longer exposing the integrator internal state to Cython is that we can encapsulate its 9 global variables into a single struct.

For a proof-of-concept, see `integrate.cpp` in [`c933b1b`](https://github.com/espressomd/espresso/commit/c933b1be107e30a866b2e9cfe5dfbda7907e5f64#diff-d7a42dacac8665929e80a826633769e775a73a9a9717cf6adf4f51d7dffeb512). I will not work further on it for now, because it is not entirely satisfactory, e.g. some of the core functionality like ICC cannot get a pointer to it via function arguments without a significant refactoring. Also, we need to define what is supposed to go in the integrator struct once we enable per-particle integration. And more urgently, we need to get the observable framework to run in parallel to help further simplify the integrator code.

src/python/espressomd/lb.pxd
```diff
@@ -119,6 +119,9 @@ cdef extern from "grid_based_algorithms/lb_interpolation.hpp" namespace "Interpo
     cdef InterpolationOrder linear
     cdef InterpolationOrder quadratic
 
+cdef extern from "integrate.hpp":
+    double get_time_step()
+
```
reinaual, line 124: Is this necessary, because `integrate.pyx` is no longer a cython file? Don't we have a better place to access the `time_step`?

jngrad: Unfortunately, it is still necessary in the python branch, because the old LB implementation needs it. That's the last consumer, so this will be gone in the walberla branch.

reinaual (approved):

PR #4715, opened 2023-04-26
Add support for recent compilers

jngrad:
Description of changes:
- fix compiler diagnostics from Clang 16.0.0 and GCC 13.0
- reduce the amount of `-Wconversion` diagnostics

### DIFF
--- a/src/core/BondList.hpp
+++ b/src/core/BondList.hpp
@@ -139,9 +139,9 @@ public:
       auto const id_pos = find_end(m_it);
       auto const partners_begin = m_it;
       auto const partners_end = id_pos;
-      return {-(*id_pos) - 1,
-              Utils::make_span(std::addressof(*partners_begin),
-                               std::distance(partners_begin, partners_end))};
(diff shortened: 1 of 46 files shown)


jngrad:
This is the work that helped me detect the LB inertialess tracers bug. I concentrated my efforts on the `-Wconversion` from header files, which flood the terminal, and tried to avoid LB/EK/FFT, since we have concrete plans to replace them.

The DP3M `memcpy` issue was likely a false positive due to the default compile-time value of 0 for `cao`, which could trick the `-Warray-bounds` heuristic into thinking `std::copy` would copy null bytes from an empty span.

jngrad (commented):
* the `-Warray-bounds` fix was submitted to the Fedora package, where the issue was originally reported
* all `-Wconversion` diagnostics from the walberla bridge and kernels have been addressed in the walberla branch (e847d7135d7f192b005c81978f0e805640715ca4, be6c7336789e0d81cc002fa1e3af0189b5e0263b)

reinaual (approved):

PR #4716, opened 2023-04-28
Bond breakage support for angle bonds

RudolfWeeber:
The counterpart to collisoin deteciotn with angl ebonds on the vs


RudolfWeeber:
@jngrad coul you please take a look at the doxygen warning. It's not clear to me what's wrong.
Thanks

jngrad (approved):

PR #4728, opened 2023-05-17
LB non-equilibrium pressure tensor

### LINKED ISSUE #3517: Walberla: Decide about the future of stress_neq.
Walberla provides only one stress getter per node. It is not entirely clear, what that returns. Likely the hydrostatic part is not included, but the diagonals are not zero for the unthermalized lb_shear test either.
Once a decision has been made, the lb node stress_neq test can be re-activated in array_properties.py (search for stress_neq)

jngrad:
Fixes #3517

Description of changes:
- implement getter for the LB non-equilibrium pressure tensor
- add extra LB checks in the testsuite


itischler (approved):
seems all good to me

reinaual (approved):

PR #4734, opened 2023-05-23
Add lattice-Boltzmann CUDA kernels

jngrad:
Description of changes:
- Add support for waLBerla GPU kernels
   - not yet supported: Lees-Edwards boundary conditions and VTK writers on the GPU
   - performance issues: CUDA kernels are not launched in parallel but sequentially without CUDA MPS, only affects simulations with more than 1 MPI rank; also slice getters/setters are quite slow


src/walberla_bridge/src/lattice_boltzmann/LBWalberlaImpl.hpp
```diff
@@ -695,6 +752,101 @@ class LBWalberlaImpl : public LBWalberlaBase {
     }
   }
 
+  [[nodiscard]] bool is_gpu() const noexcept override {
+#if defined(__CUDACC__)
+    return true;
```
RudolfWeeber: Should this really be true unconditionally or actually depend on the arch template param?

src/core/lb/LBNone.hpp
```diff
@@ -41,8 +42,16 @@ struct LBNone {
                                            bool) const {
     throw NoLBActive{};
   }
-  bool add_force_at_pos(Utils::Vector3d const &pos,
-                        Utils::Vector3d const &force) const {
+  bool add_force_at_pos(Utils::Vector3d const &,
+                        Utils::Vector3d const &) const {
+    throw NoLBActive{};
+  }
+  void add_forces_at_pos(std::vector<Utils::Vector3d> const &,
```
RudolfWeeber, line 49: Teh get_velocities_at_pos and add_forces_at_pos share logic for transforming coordinates. It looks similar and might be outsourced to a central funciton. Also, i think the implementation implies there being exaclty one lb block per mpi task. While we currently do this, we should probably at checks at the places where we imply it. On CPU at least, more blocks per core can result in better cache locallity.

src/core/lb/particle_coupling.cpp
```diff
@@ -172,45 +183,109 @@ Utils::Vector3d ParticleCoupling::get_noise_term(Particle const &p) const {
   return m_noise_pref_wo_gamma * Utils::hadamard_product(sqrt(gamma), noise);
 }
 
-void ParticleCoupling::kernel(Particle &p) {
-  auto const agrid = m_lb.get_agrid();
```
RudolfWeeber, line 176: I've been looking at this funcito for a while now. 
The complexity arises mainly, as in some cases forces are added in more than one place for a single particle.

I wonder if it cannot be simplified by passing over the vector of particle pointers twice.
Precondition: only particles within the local domain +halo (agrid/2) should be in the vector

1. Get all interpolated velocities at particles' folded position. This yields exactly one velocity per particle.
2. Populate arrays with coupling positions and coupling forces. Coupling position and coupling force arrays will again have the same length, but there can be more coupling positions/forces than particles.

    * For nromal particles, the coupling forces comes from the difference of particle and lb velocity. Coupling positions are teh particle position plus, in some caes, images shifted by +-box_l. 
    * For swimmers, the force comes from the particles'swimming properites. Again, forces are applied at the particle osition, plus positions shifted by +- box_l, if within agrid/2 of the boudnary

Or am I missing sth?

jngrad: Yes. This is more or less what the new `commit()` method does. It was introduced for the GPU kernels. In an upcoming commit, the CPU kernels will use `commit()` instead of `kernel()` to reduce code duplication, then another commit will fix a regression in `commit()` that caused swimmers configured with `force_on_fluid` to be skipped, and finally another commit will delete the unused `kernel()` method and rename `commit()` to `kernel()`.

src/walberla_bridge/src/lattice_boltzmann/LBWalberlaImpl.hpp
jngrad: Good point.

src/core/lb/LBNone.hpp
jngrad: Agreed. The conversion factors can be stored as data members in the class. This is what we actually do in the script interface. These data members can be initialized when a LB solver is attached to the system. The same data members must be erased when detaching the LB solver.

jngrad: Regarding your point about the blocks, I agree and will add assertions in the relevant methods. We actually made this assumption in other parts of the `LBWalberlaImpl` class, for example the slice getters, but forgot to add assertions.

src/core/lb/particle_coupling.cpp
jngrad: offline discussion: postpone this refactoring to the coding day

RudolfWeeber:
I'm in the process of reviewing.
Here are some points I found so far

For the Cuda accessor templates:
* Pls add (brief) docstrings explainling what the individual funcitons do, especially, where it is not obvious (e.g. broadcast)
* Mention the difference between the two linear index getter functions and when to use which
* The use of the interpolation getters/adders is not limited to coupling. If you want to put it in a name space, better use Interpolation as name
* In the same vain, part_index and n_part in the interpolation code should be pos_index n_pos or the like
* The weight and index calculation code between getter and adder shold be shared 
* I'm not sure about the weight calculation. There seem to be 3 liens where always w_x[2] is assignted to. However, indices on the rhs go from 0 to 2. Should this also be hte case on the lhs? If so, why do the tests pass?

RudolfWeeber:
* Great idea to collect the conversion factors in one place. Would it be better to make them funcitons though, then we cannot forget to update them, if we allow for time_step changes later?
* In the initialization of m_conv = Conversions[...] are we on a new enough cc++ to use stuctured bingins [.vel_to_lb=...]
* How about other quantities we have to convert, e.g. viscosity?

RudolfWeeber:
Is it intentional theat the GPU verison fo the lb_thermostat test is commented out? It does pass.

RudolfWeeber:
W.r.t performance (tested on Gerenuk with a 3070):
* The GPU LB is faster than the CPU one starting at ~30^3 grid points
* The runtiem (lb benchmark, 1000 particles at 0.1 volume fraction) is 2ms per iteration even for tiny LBs. Only at ~40^3 the number of LB nodes starts to matter. 
* Single precision is slightly faster than double precision, but the differene is small given that I ran on a gaming GPU with throttled double precision performance
* In the perf profile, pthread_mutex_unlock@@GLIBC_2.2.5 and the corresponding lock are the leading contributors to runtime by a wide margin. 

I think we need to look into the follwing:
I understand pystencils_espresso.py sets up the fields with the same memory layout for cpu and gpu. Is that actually the best choice? IIRC, optimal indexing is different on GPU?

RudolfWeeber:
I verified that the runtime is not donminated by particle coupling (by commenting it out). Without coupling, runtime changed from 2.0 to 1.9 ms



The memory layout hypothesis is supported by nvidia NSight (nsys stats)


* In the Cuda Api Summary, the leading oen is CudaMemCpy3D
* In the Mem-Ops summary, the leadign one with 99% is Cuda MemCopy DtoD, there are 8 million calls for 8000 time steps
* In the memory by volume summary, the max size of the DtoD transfers is 4k,

jngrad:
Thank you for looking into this. I'll apply the changes you suggested in your first post: docstrings, namespace, argument names, removing code duplication of weight/index calculation.

As for the other points:

> There seem to be 3 liens where always w_x[2] is assignted to.

These three lines actually define 3 similarly-named C-style arrays of size 2. This is actually a bit verbose and can be simplified by declaring a 2-dimensional C-style array instead, with sizes `{{D}}` (jinja template variable for the number of dimensions) and 2 (interpolation of order 2), which also allows us to factor out the C-style array `distance`, which only holds temporary values.

> Great idea to collect the conversion factors in one place. Would it be better to make them funcitons though, then we cannot forget to update them, if we allow for time_step changes later?

Should changing the MD time step affect unit conversion? If so, then the script interface unit conversions are wrong too.

> In the initialization of m_conv = Conversions[...] are we on a new enough cc++ to use stuctured bingins [.vel_to_lb=...]

That would be nice, but requires C++20, which requires CUDA 12. Then we can't use Clang, Clang-Tidy and scan-build on ICP workstations anymore, since only Clang 17 and 18 are compatible with CUDA Toolkit 12. Clang-format *might* still work, since it uses a heuristic instead of the abstract syntax tree.

> How about other quantities we have to convert, e.g. viscosity?

The script interface has its own unit conversion code, which I duplicated here in the core. I rewrote the Ekin and LB script interface slightly to make the instantiation of new objects and the generation of the unit conversion tables more isolated from the script interface logic, but there is still room for improvement. I would personally defer that task to an ulterior refactoring PR, because it's a time sink.

> Is it intentional theat the GPU verison fo the lb_thermostat test is commented out? It does pass.

Yes: running the code with 2 MPI ranks makes the GPU kernels 10 times slower. I'll update `CMakeLists.txt` to run this test twice, once with 1 MPI ranks, and once with 2 MPI ranks but without GPUs.

> I understand pystencils_espresso.py sets up the fields with the same memory layout for cpu and gpu. Is that actually the best choice? IIRC, optimal indexing is different on GPU?

I ended up using the same layout as on the CPU, because it's how the waLBerla GPU tutorial does it, and it saves me from redesigning the host memory data layout when e.g. implementing the slice setters and getters. Re-generating the GPU kernels with `zyxf` instead of `fzyx` breaks a few tests, without affecting runtime much:
```
$ ./pypresso ../maintainer/benchmarks/lb.py  --particles_per_core 1000 --lb_sites_per_particle 64
CPU layout fzyx: average: 9.4ms
GPU layout fzyx: average: 4.5ms
GPU layout zyxf: average: 4.4ms
```

RudolfWeeber (approved):

PR #4737, opened 2023-05-24
Particle property ranges, remove ParticleRanges from P3M

### LINKED ISSUE #4696: particle property iterators
Create iterators over single particle propreties and common groups of particle properties
```C++
auto positions = PositionIterator(cellstrucutre.local_particles());
for (std::vector3d& pos: positions) {
};
```

This can, to my understanding, be implemented by a boost transform iterator.

For combinations such as
PositionChargeIterator,

a design choice needs to be made:
Either one uses a boost zip iterator, then the individual properties (position, velocities) are numbered. (zip itertaor returns tuples)
Or we create proxy classes which forward the poiters and references as named members

```C++
class PositionChargeRefProxy() {
   PositionVelcoityProxy(Vector3d& pos_ref, Vector3d charge_ref) {
   m_pos_ref = pos_ref;
   m_charge_ref = charge_ref;
   };
   Vector3d& m_pos_ref;
   Vector3d& charge_ref;
   Vector3d& pos() { return pos_ref; };
   doube& charge() { return m_charge_ref; };
};
```

### LINKED ISSUE #4721: Make P3M independent of ESPResSo implementation details
## Objective

The P3M algorithm (and in the future: all electrostatic algorithms) should not need to know how Espresso handles details like the particle data structure.

In particular, the various functions should no longer take a `ParticleRange` but a `PositionRange`, `ChargeRange`, `ForceRange`.

## Steps

* Write the single property iterators and ranges for position, charge and force. Be careful to return references, not values. For the input, const references can be used. See #4696 
* At the beginning of `P3M::long_range_kernel()` create ranges for positions/charges/forces from the `ParticleRange` using the results from the previous step
* Change the functions which take a `ParticleRange` called in `long_range_kernel()` to use the ranges for the individual properties rather than `ParticleRange`
* The calls of `Utils::integral_parameters` are a special case. this function does not have to be changed, but the signatures of the call operators of the `AssignForces` functor does. Then, in the call to `integral_parameters`, the `ParticleRange` can also be replaced by a `PositionRange` and a `ChargeRange`

## Notes

* See #4696 for the property iterators/ranges
* This ticket is part of #4695
* The signature of the  public interface functions of the P3M class such as `add_long_range_forces()` have to be kept unchanged for now, since this interfaces is shared among all long range actors
* boost::combine can be used to do a zipped iteration over two ranges (e.g., `pos_and_charge_range = boost::combine(pos_range, charge_range)`. This can be used in a range-based loop.

jhossbach:
Partial fix for #4721, closes #4696

Description of changes:
- Implements the `ParticlePropertyRange` struct containing static methods to transform a `ParticleRange` to a `boost::iterator_range` of properties of the particles, such as `ForceRange` or `ParticleRange`, which can then be used in range-based for-loops.

The refactoring of the P3M part is not finished yet, but I think this PR could also be extended to remove the ParticleRanges from the remaining Coulomb interactions.

### DIFF
--- a/src/core/ParticlePropertyIterator.hpp
+++ b/src/core/ParticlePropertyIterator.hpp
@@ -1,0 +1,64 @@
+/*
+ * Copyright (C) 2023 The ESPResSo project
+ *
+ * This file is part of ESPResSo.
+ *
+ * ESPResSo is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * ESPResSo is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "BoxGeometry.hpp"
+#include "Particle.hpp"
+#include "ParticleRange.hpp"
+
+#include <boost/iterator/transform_iterator.hpp>
+#include <boost/range/iterator_range.hpp>
+
+#include <utils/Vector.hpp>
+
+namespace ParticlePropertyRange {
+
+namespace detail {
+template <class Kernel>
+auto create_transform_range(ParticleRange const &particles, Kernel kernel) {
+  auto transform_iterator_begin =
+      boost::make_transform_iterator(particles.begin(), kernel);
+  auto transform_iterator_end =
+      boost::make_transform_iterator(particles.end(), kernel);
+  return boost::make_iterator_range<decltype(transform_iterator_begin)>(
+      transform_iterator_begin, transform_iterator_end);
+}
+} // namespace detail
+
+auto unfolded_pos_range(ParticleRange const &particles,
+                        BoxGeometry const &box) {
+  auto return_unfolded_pos = [&box](Particle &p) {
+    return ::detail::unfolded_position(p.pos(), p.image_box(), box.length());
+  };
+  return detail::create_transform_range(particles, return_unfolded_pos);
+}
+
+auto charge_range(ParticleRange const &particles) {
+  auto return_charge = [](Particle &p) -> double & { return p.q(); };
+  return detail::create_transform_range(particles, return_charge);
+}
+auto force_range(ParticleRange const &particles) {
+  auto return_force = [](Particle &p) -> Utils::Vector3d & {
+    return p.force();
+  };
+  return detail::create_transform_range(particles, return_force);
+}
+
+} // namespace ParticlePropertyRange
--- a/src/core/electrostatics/p3m.cpp
+++ b/src/core/electrostatics/p3m.cpp
@@ -41,6 +41,7 @@
 #include "BoxGeometry.hpp"
 #include "LocalBox.hpp"
 #include "Particle.hpp"
+#include "ParticlePropertyIterator.hpp"
 #include "ParticleRange.hpp"
 #include "actor/visitors.hpp"
 #include "cell_system/CellStructure.hpp"
@@ -64,6 +65,7 @@
 #include <boost/mpi/collectives/broadcast.hpp>
 #include <boost/mpi/collectives/reduce.hpp>
 #include <boost/mpi/communicator.hpp>
+#include <boost/range/combine.hpp>
 #include <boost/range/numeric.hpp>
 
 #include <algorithm>
@@ -354,8 +356,9 @@ void CoulombP3M::assign_charge(double q, Utils::Vector3d const &real_pos) {
 }
 
 template <int cao> struct AssignForces {
+  template <typename combined_ranges>
   void operator()(p3m_data_struct &p3m, double force_prefac,
-                  ParticleRange const &particles) const {
+                  combined_ranges const &p_q_force_range) const {
     using Utils::make_const_span;
     using Utils::Span;
     using Utils::Vector;
@@ -365,9 +368,11 @@ template <int cao> struct AssignForces {
     /* charged particle counter */
     auto p_index = std::size_t{0ul};
 
-    for (auto &p : particles) {
-      if (p.q() != 0.0) {
-        auto const pref = p.q() * force_prefac;
+    for (auto zipped : p_q_force_range) {
+      auto p_q = boost::get<0>(zipped);
+      auto &p_force = boost::get<1>(zipped);
+      if (p_q != 0.0) {
+        auto const pref = p_q * force_prefac;
         auto const w = p3m.inter_weights.load<cao>(p_index);
 
         Utils::Vector3d force{};
@@ -376,22 +381,23 @@ template <int cao> struct AssignForces {
                                        p3m.E_mesh[2][ind]};
         });
 
-        p.force() -= pref * force;
+        p_force -= pref * force;
         ++p_index;
       }
     }
   }
 };
 
+template <typename combined_ranges>
 static auto calc_dipole_moment(boost::mpi::communicator const &comm,
-                               ParticleRange const &particles,
-                               BoxGeometry const &box_geo) {
-  auto const local_dip = boost::accumulate(
-      particles, Utils::Vector3d{},
-      [&box_geo](Utils::Vector3d const &dip, auto const &p) {
-        return dip + p.q() * box_geo.unfolded_position(p.pos(), p.image_box());
-      });
-
+                               combined_ranges const &p_q_unfolded_pos_range) {
+  auto const local_dip =
+      boost::accumulate(p_q_unfolded_pos_range, Utils::Vector3d{},
+                        [](Utils::Vector3d const &dip, auto const &q_pos) {
+                          auto const p_q = boost::get<0>(q_pos);
+                          auto const &p_unfolded_pos = boost::get<1>(q_pos);
+                          return dip + p_q * p_unfolded_pos;
+                        });
   return boost::mpi::all_reduce(comm, local_dip, std::plus<>());
 }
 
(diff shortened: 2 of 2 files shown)


jhossbach:
I have successfully changed all ParticleRange signatures of methods inside the `long_range_kernel` method. In order to make the whole `p3m.cpp` free of ParticleRanges one would need to remove both the signature of `long_range_kernel` itself and the `AssignCharge` struct. The latter would require touching some other files, e.g. `elc.cpp`. Is this still a part of this PR or should this be moved to a larger refactoring @RudolfWeeber ?

src/core/ParticlePropertyIterator.hpp
```diff
@@ -0,0 +1,71 @@
...
+#include <boost/iterator/transform_iterator.hpp>
+#include <boost/range/iterator_range.hpp>
+#include <tuple>
+#include <utils/Vector.hpp>
+
+namespace ParticlePropertyRange {
+
+namespace {
+template <class Kernel>
+auto create_transform_range(ParticleRange const &particles, Kernel kernel) {
```
jngrad: I would prefer to write `namespace detail {` here, and to add a `detail::` namespace at every call site. Using an unnamed namespace in a header file is not wrong (and one can find several such examples in legacy .cpp files in ESPResSo), but it's not exactly good practice, because unnamed namespaces do not behave like named namespaces: they add a `static` prefix in front of every symbol they encapsulate.

It could get pretty dangerous if one were to add a global variable in there, because one would assume the variable would be hidden, when in reality it would still be available in the `ParticlePropertyRange` namespace. In addition, every .cpp file that include this header file would disagree about the value stored in that global variable, since each translation unit would hold a different copy of the variable.

src/core/ParticlePropertyIterator.hpp
```diff
@@ -0,0 +1,71 @@
...
+      boost::make_transform_iterator(particles.begin(), kernel);
+  auto transform_iterator_end =
+      boost::make_transform_iterator(particles.end(), kernel);
+  return boost::make_iterator_range<decltype(transform_iterator_begin)>(
+      transform_iterator_begin, transform_iterator_end);
+}
+} // namespace
+
+auto pos_range(ParticleRange const &particles) {
+  auto return_pos = [](Particle &p) { return p.pos(); };
```
jngrad: Do we want ranges to yield particle properties by copy or by reference?
```suggestion
  auto return_pos = [](Particle &p) -> Utils::Vector3d & { return p.pos(); };
```

src/core/ParticlePropertyIterator.hpp
```diff
@@ -0,0 +1,71 @@
...
+auto unfolded_pos_range(ParticleRange const &particles,
+                        BoxGeometry const &box) {
+  auto return_unfolded_pos = [&box](Particle &p) {
+    return unfolded_position(p.pos(), p.image_box(), box.length());
+  };
+  return create_transform_range(particles, return_unfolded_pos);
+}
+
+auto charge_range(ParticleRange const &particles) {
+  auto return_charge = [](Particle &p) { return p.q(); };
```
jngrad: Maybe we need to yield by reference here, but I'm not sure. The ICC algorithm for example needs to be able to modify particle charges.
```suggestion
  auto return_charge = [](Particle &p) -> double & { return p.q(); };
```

src/core/ParticlePropertyIterator.hpp
```diff
@@ -0,0 +1,71 @@
...
+ * ESPResSo is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+#ifndef ESPRESSO_SRC_CORE_PARTICLE_EXTRACT_PROPERTIES_HPP
+#define ESPRESSO_SRC_CORE_PARTICLE_EXTRACT_PROPERTIES_HPP
```
jngrad:
```suggestion
#pragma once
```
Include pragmas are allowed in ESPResSo since 2023.

This avoids accidentally defining the same include guard label in two different files. The label doesn't match the file name.

src/core/ParticlePropertyIterator.hpp
```diff
@@ -0,0 +1,71 @@
...
+auto force_range(ParticleRange const &particles) {
+  auto return_force = [](Particle &p) -> Utils::Vector3d & {
+    return p.force();
+  };
+  return create_transform_range(particles, return_force);
+}
+
+} // namespace ParticlePropertyRange
+
+#endif
```
jngrad:
```suggestion
```
See include pragmas.

src/core/electrostatics/p3m.cpp
```diff
@@ -366,24 +371,28 @@ template <int cao> struct AssignForces {
...
+        p_force -= pref * force;
         ++p_index;
       }
     }
   }
 };
 
-auto dipole_moment(Particle const &p, BoxGeometry const &box) {
-  return p.q() * unfolded_position(p.pos(), p.image_box(), box.length());
+auto dipole_moment(double const q, Utils::Vector3d const unfolded_pos) {
```
jngrad:
```suggestion
static auto dipole_moment(double const q, Utils::Vector3d const &unfolded_pos) {
```

src/core/electrostatics/p3m.cpp
```diff
@@ -366,24 +371,28 @@ template <int cao> struct AssignForces {
...
-      particles, Utils::Vector3d{}, [&box](Utils::Vector3d dip, auto const &p) {
-        return dip + dipole_moment(p, box);
-      });
+                        combined_ranges const &p_q_unfolded_pos_range) {
+  auto const local_dip =
+      boost::accumulate(p_q_unfolded_pos_range, Utils::Vector3d{},
+                        [](Utils::Vector3d dip, auto const &q_pos) {
+                          double p_q{};
+                          Utils::Vector3d p_unfolded_pos{};
+                          boost::tie(p_q, p_unfolded_pos) = q_pos;
```
jngrad: Please double-check if this compiles, and if this improves the generated code. The idea would be to skip the copy of argument `dip`, skip the zero-initialization of `p_unfolded_pos`, and skip the copy of `p_unfolded_pos`.
```suggestion
                        [](Utils::Vector3d const &dip, auto const &q_pos) {
                          auto const p_q = boost::get<0>(zipped);
                          auto const &p_unfolded_pos = boost::get<1>(zipped);
```

src/core/ParticlePropertyIterator.hpp
jngrad: Also, according to the code coverage report, this function is never called. Consider removing it.

src/core/ParticlePropertyIterator.hpp
```diff
@@ -0,0 +1,71 @@
...
+#define ESPRESSO_SRC_CORE_PARTICLE_EXTRACT_PROPERTIES_HPP
+
+#include "BoxGeometry.hpp"
+#include "Particle.hpp"
+#include "ParticleRange.hpp"
+#include "grid.hpp"
+#include <boost/iterator/transform_iterator.hpp>
+#include <boost/range/iterator_range.hpp>
+#include <tuple>
+#include <utils/Vector.hpp>
```
jngrad:
```suggestion
#include "BoxGeometry.hpp"
#include "Particle.hpp"
#include "ParticleRange.hpp"

#include <utils/Vector.hpp>

#include <boost/iterator/transform_iterator.hpp>
#include <boost/range/iterator_range.hpp>
```

src/core/ParticlePropertyIterator.hpp
```diff
@@ -0,0 +1,71 @@
+/*
+ * Copyright (C) 2010-2022 The ESPResSo project
```
jngrad:
```suggestion
 * Copyright (C) 2023 The ESPResSo project
```

jngrad (changes requested):
Looks good, pending suggested changes.

src/core/electrostatics/p3m.cpp
jhossbach: The `static` keyword is redundant here since we are in an anonymous namespace. Alternatively we can also replace the anonymous namespace with a `detail` namespace as above

jngrad: I have already replaced the anonymous namespaces by `static` functions to make it more explicit in the PR that removes global variables ([diff](https://github.com/espressomd/espresso/pull/4783/commits/11e1df1c9d6d3639a249d6bc6e129d65f382ae66#diff-c72eb42c3b85ab12f7d6b7b86dfa8d29ee3aa84fd4fa190f5ce6012e5a5740a9R386)).

jhossbach: Updating when #4783 is merged.

src/core/ParticlePropertyIterator.hpp
jhossbach: Yes, I would also assume this to be a reference (so similar to `return_force`)

src/core/ParticlePropertyIterator.hpp
jhossbach: You are right, I forgot to remove this now redundant function

src/core/electrostatics/p3m.cpp
jngrad: This is still not resolved: the `static` is missing (we're no longer in an anonymous namespace after the rebase), and we want to avoid a pass-by-copy.

src/core/electrostatics/p3m.cpp
```diff
@@ -376,22 +381,27 @@ template <int cao> struct AssignForces {
...
+auto dipole_moment(double const q, Utils::Vector3d const unfolded_pos) {
+  return q * unfolded_pos;
+}
 
+template <typename combined_ranges>
+static auto calc_dipole_moment(boost::mpi::communicator const &comm,
+                               combined_ranges const &p_q_unfolded_pos_range) {
+  auto const local_dip =
+      boost::accumulate(p_q_unfolded_pos_range, Utils::Vector3d{},
+                        [](Utils::Vector3d dip, auto const &q_pos) {
```
jngrad:
```suggestion
                        [](Utils::Vector3d const &dip, auto const &q_pos) {
```

jngrad (approved):

PR #4741, opened 2023-06-09
Introduce a System class in the core

### LINKED ISSUE #4615: Introduce an `EspressoSystem` class in the core
... which holds the state of the simulation.

* in the long run, each `espressomd.System` python class whoudl one one of them, via the script interface
* to ease the transition, the core class can be made available via a global, initially.
* globals, as far as they are still needed, would become members of this class

jngrad:
Fixes #4615

Description of changes:
- completely rewrite the `EspressoSystemInterface` class
   - the new `System` class uses composition for GPU particle data management, resource deallocation, and globals
   - GPU particle data management now uses standard `thrust` vectors and properly deallocates device memory
   - simplify the `espressomd.system.System` class
- simplify GPU code
   - 9 CUDA global variables were removed (partial fix for #2628)
   - ~300 lines of CUDA code were removed


src/script_interface/system/SystemFacade.hpp
```diff
@@ -0,0 +1,73 @@
+/*
+ * Copyright (C) 2013-2022 The ESPResSo project
```
reinaual, line 2: Do you put these specific copyright years here on purpose? Also in the "new" `System` implementation in the core.

reinaual (approved):
Code looks good to me, as far as I understood the structure.

src/script_interface/system/SystemFacade.hpp
jngrad: If I remember correctly, this file was actually renamed from `Globals.hpp`. But the changes are so significant that git cannot deduce it's the same file. Internally git doesn't keep track of filename changes and has to deduce it from a diff algorithm with ranking, which is why sometimes adding a CMakeLists.txt file and removing a Python file can be misinterpreted as a filename change by git (they format their license header in the exact same way).

PR #4743, opened 2023-06-12
Thermostat cleanup

### LINKED ISSUE #3734: LB coupling should use per particle gamma
Currently the behavior of the LB and the Langevin thermostat differ in that the latter does use per partticle gamma's and particle anisotropy, while the former only support a global, scalar friction coefficient. This shouldn't be hard to change, and would allow for anisotropic point particles with dispersity. (In principle this is also true for the per particle temperature, but this is even more of a foot gun than for Langevin, so not a big fan...)

christophlohrmann:
Fixes #3734 

Description of changes:
- use particle-specific gamma in LB
- check that particle gamma is not anisotropic (not supported by LB)
-  update test with more 'physical' paameters

### DIFF
--- a/src/core/grid_based_algorithms/lb_particle_coupling.cpp
+++ b/src/core/grid_based_algorithms/lb_particle_coupling.cpp
@@ -24,6 +24,7 @@
 #include "errorhandling.hpp"
 #include "grid.hpp"
 #include "random.hpp"
+#include "thermostat.hpp"
 
 #include "grid_based_algorithms/lb_interface.hpp"
 #include "grid_based_algorithms/lb_interpolation.hpp"
@@ -35,6 +36,7 @@
 
 #include <boost/mpi.hpp>
 
+#include <cmath>
 #include <cstdint>
 #include <initializer_list>
 #include <limits>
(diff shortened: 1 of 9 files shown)


src/core/thermostat.cpp
```diff
@@ -234,3 +234,16 @@ void mpi_set_nptiso_gammas(double gamma0, double gammav) {
   mpi_call_all(mpi_set_nptiso_gammas_local, gamma0, gammav);
 }
 #endif
+
+Thermostat::GammaType
+handle_particle_gamma(Thermostat::GammaType const particle_gamma,
+                      Thermostat::GammaType const default_gamma) {
```
jngrad: here and elsewhere, make arguments take const references

src/core/grid_based_algorithms/lb_particle_coupling.cpp
```diff
@@ -132,8 +156,10 @@ Utils::Vector3d lb_drag_force(Particle const &p,
       LB::get_lattice_speed();
 
   Utils::Vector3d v_drift = interpolated_u + vel_offset;
+  auto const gamma = lb_handle_particle_anisotropy(p);
+
   /* calculate viscous force (eq. (9) @cite ahlrichs99a) */
-  return -lb_lbcoupling_get_gamma() * (p.v() - v_drift);
+  return -Utils::hadamard_product(gamma, (p.v() - v_drift));
```
jngrad: optimize out the minus sign

testsuite/python/lb_thermostat.py
```diff
@@ -24,87 +24,206 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+import unittest as ut
+import unittest_decorators as utx
+import thermostats_common
+import numpy as np
+
+import espressomd.lb
+
```
jngrad: copy paste error

jngrad (changes requested):

jngrad (approved):
Thanks!

PR #4747, opened 2023-06-28
Improve instrumentation support

### LINKED ISSUE #2509: Add documentation for Caliper
Depends on #2506.

### LINKED ISSUE #4740: Caliper bindings are broken
The Caliper bindings that mark the start and end of a portion of code to instrument are incorrect:
https://github.com/espressomd/espresso/blob/09c6958692e390bbd974528b48352e268ca9079f/src/profiler/include/profiler/profiler.hpp#L62-L64
https://github.com/espressomd/espresso/blob/09c6958692e390bbd974528b48352e268ca9079f/src/profiler/include/profiler/profiler.hpp#L71-L73
The second one should be `ESPRESSO_PROFILER_MARK_END`. The regression was introduced in 1df100905ca70bb134ecaca87af907fef744c327. When calling these functions inside the body of a Python loop, the Caliper summary creates a recursive data structure for the marked region, since the end marker is missing.

The Cython interface passes Python string objects to C++ functions that take `std::string` as arguments, triggering runtime exceptions. The interface should use `espressomd.utils.to_char_pointer()` to handle type conversion. The regression was introduced in df5d9d2956b38c34f365bf9e0f461902862d0a1e.

There are no test cases for this feature.

jngrad:
Fixes #4740, fixes #2509

Description of changes:
- third-party tools:
   - fix the broken Caliper integration and write an integration test
   - use standard Caliper macros instead of macro aliases
   - build Caliper as a CMake subproject to inherit options (e.g. CUDA and MPI support) and facilitate linking
   - add native support for CUDA-GDB (a few compiler flags were missing from `Debug` builds)
   - add native support for kernprof (new flag added to `pypresso`)
- documentation:
   - centralize scattered information about debuggers into a self-contained section of the user guide
   - document all supported debuggers and profilers: Caliper, Valgrind, GDB, CUDA-GDB, kernprof, perf, sanitizers
   - make the Doxygen docs more consistent with the Sphinx user guide (same fonts) and website (same logo)


RudolfWeeber (dismissed):

RudolfWeeber:
Thank you. In particular, the very extensive documentation will be helpful to developers sorting out performance issues.

reinaual (approved):
reviewed the merge commit only

PR #4748, opened 2023-07-06
Make observables run in parallel

### LINKED ISSUE #3154: Observables: Particle selection and filtering should be independent of observable
For particle observables the selection and filtering of the particles should be independent
of what is calculated. These are those currently derived from `Obsevables::PidObservable`,
and are already a separate class hierarchy. One plan of action could be to change the the
signature of `virtual std::vector<double> Obsevables::PidObservable::evaluate(PartCfg &partCfg) const` to `virtual std::vector<double> Obsevables::PidObservable::evaluate(Utils::Span<const Particle *>) const`, and evaluate the ids in the base class. Further, the actual implementation should potentially be a member and not be derived. This allows e.g. filtering out virtual particles in a central place and reuse the implementation of the observable function with other ways of selecting particles.

reinaual:
Fixes #3154

Description of changes:
- make observables run in parallel
   - only send the extracted data via MPI (the original implementation sent the entire particle data structure)
   - remove a major performance bottleneck in parallel simulations
- remove 8 callbacks from the deprecated `MpiCallbacks` framework


jngrad:
Here is an updated version of the plot https://github.com/espressomd/espresso/issues/3154#issuecomment-1548064703 which uses the new implementation of the observables, using a `TimeSeries` of the particle positions:

Benchmarks for a Lennard-Jones simulation with two types of writer: MPI-IO, observables accumulated with a time series. For simulations with 1 or 4 cores, observables are as fast as MPI-IO. For 2 cores, MPI-IO is faster. For 8 core or more, observables are faster.

The benchmark writes to an SSD, either raw unsorted particle positions for MPI-IO every 20 time steps, or sorted particle positions for observables written once at the very end (sampled every 20 time steps with an auto-accumulator).

Attachments: [benchmark.py](https://github.com/espressomd/espresso/files/12149302/benchmark.py.txt), [plot.py](https://github.com/espressomd/espresso/files/12149303/plot.py.txt)

jngrad (approved):

PR #4768, opened 2023-08-07
Separate short-range forces by signature

### LINKED ISSUE #4755: Separate short-range forces by signature
For the struct-of-arrays based short-range loop (#4754), short range forces need to be separated by their input/output signature:

This ticket is suitable for new developers, as the work is pretty localized and does not require advanced C++ concepts

* central, radial: 

    * returns scalar force (along the connecting vector)
    * takes distance and interaction parameters as input
    * examples: Lennard-Jones, soft-sphere, tabulated

* central radial with charge:

    * input: distance, product of charges, interaciton parameters
    * output: force along the connecting vector
    * Example: P3M short range
* non-central
    * dipoles short-range contributoin, elc short range contribution, etc.

Depending on which classes of forces are active, the struct-of-arrays short-range calculaoitn can decide, which quantities to pre-calculate

## Technical details

* Split the short-range force soccurring in 
forces_inline.hpp:add_nonbonded_pair_force and calc_nonbonded_pair_force into separate funciotns based on the categorizaiton listed above.
* From the add_nonbonded_pair_force function then call the functions for the individual categories and add up their result.
* Work on energy kernels should be deferred.

1234somesh:
Fixes #4755

Description of changes:
- split short-range force kernel into a central radial kernel, a central radial with charge kernel, and a non-central kernel

### DIFF
--- a/src/core/constraints/ShapeBasedConstraint.cpp
+++ b/src/core/constraints/ShapeBasedConstraint.cpp
@@ -91,8 +91,11 @@ ParticleForce ShapeBasedConstraint::force(Particle const &p,
 
     if (dist > 0) {
       outer_normal_vec = -dist_vec / dist;
-      pf = calc_non_bonded_pair_force(p, part_rep, ia_params, dist_vec, dist,
-                                      get_ptr(coulomb_kernel));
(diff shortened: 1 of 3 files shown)


jngrad (approved):
LGTM

PR #4769, opened 2023-08-08
Switch LB to pull scheme (un-optimized)

RudolfWeeber:
The integraiotn loop consist of the following sweeps:
* pull (stream)
* collide
* update velocity field from pdfs
* ghost communicatoin


RudolfWeeber:
By now, I combined stream/collide into a single kernel.
The remaingin issue is the boundary condition. Why exactly do we have a patched version? Was that only due to push scheme or some other reason? I do not really understand, what the patch changes.

Issue #4723, opened 2023-05-09
Move the walberla bridge to a self-contained git repo


RudolfWeeber:

RudolfWeeber:
Make the walberla_bridge (located in src/walberla_bridge) an independent library, i.e.,

* it should live in its own git repo
* and have its own CMake, such that it can be added to other cmae project, and that it knows how to run its own tests

The goal is that the library can be used by other codes than Espresso, e.g., our parterns in MultiXScale.

jngrad:
Requirements:
* [ ] Provide a copy of `cmake/unit_test.cmake`
* [ ] Replace `Utils::Vector3d` by a class with similar functionality
* [ ] Remove Boost dependency (see [walberla/walberla#190](https://i10git.cs.fau.de/walberla/walberla/-/issues/190))
   * [ ] Replace `boost::optional` by `std::optional`
   * [ ] Replace `boost::variant` by `std::variant`
   * [ ] Replace `boost::multi_array` by equivalent code
   * [ ] unit tests can probably still use Boost

RudolfWeeber:
The first step is to 
* copy the content of src/walberla_bridge into a new git repo outide espresso
`walbela_bridge`
* src/utils/include/utils/vector.hpp from Esprsso will also be needed for Vector3d and friends
* Modify the CMakeList.txt sot it can be compiled independently
* This involves copying in the combiler definitoins and walberla importing code form ESpresso's CMakeList.txt. Currently Espresso imports Walberla as a cake subproject. This will have to be done in the smae way for the walberla bridge repo.


Once the Walberla bridge can be built in its own repo independently of ESPRso, we'll sort out using that new repo from Espresso.

jngrad:
A proof-of-concept can be found in [jonscheunemann/walberla_bridge](https://github.com/jonscheunemann/walberla_bridge). Thanks @jonscheunemann and Marc Sauter!

Once we're done with the major LB/EK rewrite, we'll be able to use that work to create the bridge as part of the ESPResSo organization. I'll take care of resolving any remaining CMake subtleties with the `FetchContent` method.

PR #4773, opened 2023-08-11
Rewrite LB/EK interface

jngrad:
Description of changes:
- API changes:
   - the `espressomd.System` class now has a new member `lb`
   - the `espressomd.System` class member `ekreactions` is now the `reaction` attribute of `system.ekcontainer`
   - the `espressomd.System` class no longer has an `actors` member
   - further API changes will take place in an upcoming PR
- remove Boost dependency from LB/EK sources
   - follow-up to [**Remove remaining boost components** walberla/walberla#605](https://i10git.cs.fau.de/walberla/walberla/-/merge_requests/605)
   - partial fix for https://github.com/espressomd/espresso/issues/4723#issuecomment-1542383127
   - unit tests from the walberla bridge still depend on `Boost::unit_test_framework`
- fix regressions in EK
   - EK now throws an error when the box_l or node_grid changes, when the MD time step and EK tau disagree, when the box size is not an integer multiple of agrid
- rewrite the LB and EK code in the core from scratch to enforce SOLID principles
   - folder `src/grid_based_algorithms` was split into `src/core/lb` and `src/core/ek`
   - LB and EK global variables (`lattice_switch`, `ek_container`, `ek_reactions`, `lb_walberla_instance`,
   `lb_walberla_params_instance`) were removed; these objects are now members of the `System` class
    (partial fix for #2628)


src/python/espressomd/system.py
```diff
@@ -280,6 +273,14 @@ def __setstate__(self, params):
         self.call_method("set_system_handle", obj=params.pop("_system_handle"))
         for property_name in params.keys():
             System.__setattr__(self, property_name, params[property_name])
+        # note: several members can only be instantiated once
+        if has_features("WALBERLA"):
+            if self._lb is not None:
+                lb, self._lb = self._lb, None
```
itischler, line 279: is there a reason why this is set in 2 lines and not just `self.lb, self._lb = self._lb, None`?

src/core/integrate.cpp
```diff
@@ -407,24 +460,26 @@ int integrate(int n_steps, int reuse_forces) {
...
-          LB::propagate();
-          EK::propagate();
+          lb.propagate();
+          ek.propagate();
         }
         lb_lbcoupling_propagate();
       } else if (lb_active) {
-        auto const lb_steps_per_md_step = LB::get_steps_per_md_step(time_step);
+        auto &lb = system.lb;
+        auto const lb_steps_per_md_step = lb.get_steps_per_md_step(time_step);
```
itischler: I find this naming of the variables and functions misleading, because with tau >= dt (which is forced) there are multiple md steps in between 2 lb/ek steps. better would be `md_steps_per_lb_step`

src/core/ek/Solver.hpp
```diff
@@ -0,0 +1,101 @@
...
+   * All derived parameters and the fluid are reset to their default values.
+   */
+  void init() const {}
+
+  /**
+   * @brief Get the @c EK time step.
+   */
+  double get_tau() const;
+
+  auto get_steps_per_md_step(double time_step) const {
```
itischler: misleading naming see integrate.cpp

src/core/lb/Solver.hpp
```diff
@@ -0,0 +1,160 @@
...
+
+  /**
+   * @brief Add a force density to the fluid at the given position.
+   * @param pos            Position at which the force density is to be applied.
+   * @param force_density  Force density to apply.
+   */
+  void add_force_density(Utils::Vector3d const &pos,
+                         Utils::Vector3d const &force_density);
+
+  auto get_steps_per_md_step(double time_step) const {
```
itischler: misleading naming see integrate.cpp

itischler (commented):
Great work,
I only have a minor comment on the naming of steps_per_md_steps, which should probably be md_steps_per_lb/ek_steps

src/python/espressomd/system.py
jngrad: Unfortunately yes, it has to be written in two lines because of Python's evaluation order.

In more detail: `self.lb` is a callback method that internally assigns a value to `self._lb`. We want to back up the `self._lb` object from the checkpoint file in a temporary variable `lb`, then assign `None` to `self._lb`. Then the expression `self.lb = lb` can safely initialize `self._lb` under the hood.

I couldn't find an better syntax due to how pickling works in Python. Once we're done moving all global variables to the core `System` class, all of this unsustainable Python code will disappear.

src/core/integrate.cpp
jngrad: Good catch! Not only that, but the assertion was also incorrect, because the EK propagation counter was not synchronized with the LB propagation counter. Let me fix that.

jngrad: By the way, we should really think about writing a test for tau > dt with LB + EK active.

itischler (approved):

RudolfWeeber (approved):

PR #4777, opened 2023-08-22
Add option to skip ghost force reduction

### LINKED ISSUE #4556: Eliminate the need for a ghost force reduction
Currently, Es calculates forces/energies in the RegularDecomposition as follows:
\
* Iterate over all local (not ghost) cells
* For each of them consider all upper/right/back neighbor cells (i.e. some 14 of the 27), including ghost cells, and calculate forces. Forces are added to the particles on both sides. I.e. a force contribution is added to particles in local cells AND to particles in ghost cells
* At the end, a reduction is carried out across mpi node boundaries: the forces of physical particels on one node and the forces on the corresponding ghost particles on other nodes are accumulated, so that the total force from all interactiosn (including thos frome other mpi nodes) is available on the physical particle

The drawback of this scheme is the need for an additional communication step, i.e., the ghost force reduction. This is assumed to be one of the main obstacles to Espresso's scaleability, as all nodes have to wait for each other in the communicaiot step of the reduction.
The aim of this project is to gradually remove the need for this ghost force reducion.

## [ ] Phase 1: Short-range non-bonded forces

At the end of this step, an LJ fluid should run without ghost force reduciotn. Other featuers such as bonds and vitual sites will break.

1. [ ] Add a flag to the constructor of the RegularDecomposition `without_ghost_force_reduction`
2. [ ] Support it in the Python interface. See the additional parameters of the HybridDecomposition for inspiration.
3. [ ] If the flag is true, the local cells neighboring ghost cells ON ALL SIDES have to be added to the ghost cells `red_neighbors`. This is done in RegularDecomposition::init_cell_interactions()
4. [ ] If the flag is true, omit the ghost force reduction by returning an emtyp GhostCommunicator from RegularDecomopsition::collect_ghost_forces_comm()
5. [ ] Expose the flag via an interface function in the ParticleDecomposition class hierarchy (via a a virtual function returning false in the ParticleDecomposition base class). For the RegularDecomposition override it, returning the actual value of the flag.
6. [ ] In CellStructure::linked_cell(), query the without_ghost_force_reudciont flag of the decomposition stored in CellStructure::m_decomposition via the interface function added in the previous step. If true, run the algorithm::linked_cell() on the m_ghost_cells in addition to the current call on m_local_cells. You'll have to replicate the construciotn of fisrt/last the way it's done for m_local_cells.

capomav:
Fixes #4556

### DIFF
--- a/src/core/cell_system/CellStructure.cpp
+++ b/src/core/cell_system/CellStructure.cpp
@@ -264,9 +264,9 @@ void CellStructure::set_atom_decomposition(boost::mpi::communicator const &comm,
 
 void CellStructure::set_regular_decomposition(
     boost::mpi::communicator const &comm, double range, BoxGeometry const &box,
-    LocalBox<double> &local_geo) {
-  set_particle_decomposition(
-      std::make_unique<RegularDecomposition>(comm, range, box, local_geo));
+    LocalBox<double> &local_geo, bool without_ghost_force_reduction) {
+  set_particle_decomposition(std::make_unique<RegularDecomposition>(
+      comm, range, box, local_geo, without_ghost_force_reduction));
   m_type = CellStructureType::CELL_STRUCTURE_REGULAR;
   local_geo.set_cell_structure_type(m_type);
 }
--- a/src/core/cell_system/CellStructure.hpp
+++ b/src/core/cell_system/CellStructure.hpp
@@ -527,10 +527,12 @@ public:
    * @param range Interaction range.
    * @param box Box Geometry.
    * @param local_geo Geometry of the local box.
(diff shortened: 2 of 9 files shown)


RudolfWeeber:
Lgtm so far. Next step are

* skip setting up the collect_ghost_forces_comm in the constructor of RegularDecomposiotion (srccore/cell_systems/RegularDecomposition.cpp) if the flag is ture.
* Modify the testsuite/python/lj.py so that it runs one test without skipping the ghost force reduction and another with skipping it. That second test will fail. 
* check RegularDecomposition::init_cell_neighbors() to make sure, that the neighborship of ghost cells is set up correctly
* Modify the short range loop to also run on pairs of ghost cells and their red_neighbors.

PR #4779, opened 2023-08-25
Throw runtime error on incorrect usage of `ObjectList`

jngrad:
Adding the same object twice in an `ObjectList` can lead to unintentional side-effects. For example, an `EKSpecies` added twice would be integrated twice per time step, since the two shared pointers point to the same species, until the system is checkpointed, at which point the two species become deep copies of one another and no longer share the same memory address. In addition, when removing an object that was never added to an `ObjectList`, no error message is generated to notify users they may have used the wrong variable to remove an object. There are currently assertions in the core to prevent both duplicate objects and removal of nonexistent objects, but those are not triggered in release builds.

Description of changes:
- convert `ObjectList` assertions to recoverable runtime errors (i.e. the system reverts to the last valid state)
   - adding the same object twice in an `ObjectList` now raises a runtime error
   - removing an unknown object from an `ObjectList` now raises a runtime error

### DIFF
--- a/src/core/accumulators.cpp
+++ b/src/core/accumulators.cpp
@@ -60,24 +60,33 @@ int auto_update_next_update() {
                            });
 }
 
+namespace detail {
+struct MatchPredicate {
+  AccumulatorBase const *m_acc;
+  template <typename T> bool operator()(T const &a) const {
+    return a.acc == m_acc;
+  }
+};
+} // namespace detail
+
 void auto_update_add(AccumulatorBase *acc) {
-  assert(acc);
-  assert(std::find_if(auto_update_accumulators.begin(),
-                      auto_update_accumulators.end(), [acc](auto const &a) {
-                        return a.acc == acc;
-                      }) == auto_update_accumulators.end());
+  assert(not auto_update_contains(acc));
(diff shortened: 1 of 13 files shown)


christophlohrmann (commented):
Are there already python tests for the other containers or why are only the constraints tested?

jngrad:
> Are there already python tests for the other containers or why are only the constraints tested?

The python test currently verifies the `ObjectList` virtual base class is correctly throwing an error when an assertion fails. `Constraints` is used out of convenience. Whether the other containers correctly implemented the assertion is checked indirectly, since the assertions always run whenever an object is added or the container is cleared.

We could add more checks for every container, but is it even possible to implement an incorrect `bool has_in_core()`, without breaking all tests? Say that it always returns `true`, then one can never add an object to it; say that it always returns `false`, then one can never clear the list of objects at Python exit.

christophlohrmann (approved):

RudolfWeeber (approved):
Lgtm. As I understand, this amkes insertion of new constraints/accumulators etc. an O(N) operation. Since the containers are typically small, that should not be an issue. If it becoems one, we can implement an override for the check.

PR #4781, opened 2023-09-04
Sample: Using Dask with ESPREsSo

RudolfWeeber:
This was produced as a side project of learning Dask, but might be useful for others.


samples/high_throughput_with_dask/dask_espresso.py
```diff
@@ -0,0 +1,56 @@
...
+import base64
+import sys
+import subprocess
+import logging
+from dask import delayed
+
+
+def encode_transport_data(data):
+    """
+    Uses pickle and base64 to convert the provided data to a string which can be passed safely between the dask scheduler, worker and Esprsso
```
SamTov: This line seems a little bit long, was it not picked up by the linting?

samples/high_throughput_with_dask/dask_espresso.py
```diff
@@ -0,0 +1,56 @@
...
+def dask_espresso_task(pypresso, script, **kwargs):
+    """
+    Runs Esprsso asynchronesouly as a Dask task.
+
+    pypresso: string
+      Path to pypresso
+    script: string
+      Simulation script to run with pypresso
+    kwargs:
+      The keyword arguments are passed encoded and send to the
```
SamTov: spelling: `and sent`

samples/high_throughput_with_dask/dump_test_output.py
```diff
@@ -0,0 +1,9 @@
+import pickle
+import base64
+import sys
+
+base64_data = sys.stdin.read()
+pickle_data = base64.b64decode(base64_data)
+data = pickle.loads(pickle_data)
+
+print(data)
```
SamTov: Is the print statement intentional here? If not, as it is a test, should we try to assert something?

samples/high_throughput_with_dask/echo.py
```diff
@@ -0,0 +1,6 @@
+import dask_espresso as de
+
+data = de.get_data_from_stdin()
+data.update(processed=True)
+
+print(de.encode_transport_data(data))
```
SamTov, line 29: Same as above, is the print statement intentional here?

samples/high_throughput_with_dask/dask_espresso.py
```diff
@@ -0,0 +1,56 @@
+import pickle
```
SamTov, line 22: Should there be a header here? Or is it just in some of the modules?

samples/high_throughput_with_dask/echo.py
```diff
@@ -0,0 +1,6 @@
+import dask_espresso as de
```
SamTov: Also same as above, should there be a header here?

samples/high_throughput_with_dask/run_pv.py
```diff
@@ -0,0 +1,37 @@
...
+
+PYPRESSO = "/data/weeber/es/build/pypresso"
+SIM_SCRIPT = "lj_pressure.py"
+
+N_STEPS = int(2E4)
+N_PARTICLES = 100
+VOLUME_FRACTIONS = np.arange(0.1, 0.52, 0.01)
+
+
+client = dask.distributed.Client(sys.argv[1])
```
SamTov: is the argument theoretically supposed to be either a `Cluster` instance or `None` or is it something different altogether?

SamTov (commented):
I think it looks really nice and is a great addition to the samples as it will help with really high throughout simulation studies. I have some questions scattered throughout the review but I also wanted to ask one here:

How does it deal with having more jobs than can be run at one time. For example, if I open 5 workers on a slurm cluster, can I keep passing jobs to these 5 workers or do they close after a simulation is finished? I didn't see any closing in the script so I assume it is the former but how does that work exactly?

RudolfWeeber:
Answering the general questoin: the workers stay alive and can be re-used until they are explicitly shut down. Espresso globals are kept out of the worker by running Espresso in a sub-process, i.e., in an independent Python instance. This make sthe serialization of input and output via pickle and base64 necessary, so they can be safely passed via stdion and stdout.

samples/high_throughput_with_dask/dask_espresso.py
RudolfWeeber: No idea, why it wasn't formatted. broke it manually now

samples/high_throughput_with_dask/dump_test_output.py
RudolfWeeber: I removed this file entirely. It was there for testing but didn't use the dask_espresso module for decoding.

samples/high_throughput_with_dask/echo.py
RudolfWeeber: Yes. This script mocks a simulation for testing. I made the purpose clear in the doc string

samples/high_throughput_with_dask/dask_espresso.py
RudolfWeeber: there should. Is added now everywhere

samples/high_throughput_with_dask/run_pv.py
RudolfWeeber: I made it clear that this is a scheuler address that LocalCluster does not work and clusters with remote workers probably will.

RudolfWeeber:
I also added some docstrings and comments throughout the sample.

RudolfWeeber:
Anything still open here?

SamTov:
> Anything still open here?

I was asked by @jngrad to run this solution inside of our RL workflow in order to correctly assess whether it resolves the issues raised during our meetings. This will, however, take a little bit of time as we need to restructure the SwarmRL code such that it fits this structure. I think the code here works and is well written but whether it will solve the issues with our distributed deployment is still an open issue.

jngrad (approved):
LGTM

PR #4783, opened 2023-09-05
Remove more global variables

jngrad:
Partial fix for #2628

Description of changes:
- encapsulate cell structure, box geometry and local box in the `System` class
- gather MPI-related global variables in a global struct
- rewrite position folding/unfolding free functions as member functions of `BoxGeometry`

I took this opportunity to cleanup and modernize code, by e.g. replacing hard-coded loops by `Utils::Vector` operations, improving type safety (no mixed signed and unsigned int in arithmetic expression), using more expressive variable names, passing global variables by function argument, adding a collision detection test case, etc.


RudolfWeeber:
Offline discussion: It would be preferrable not to have the BoxGeometyr passed into angel bonds. This is currenlty the case, because they are called with acutal particle positoins rather than with distance vectors.
To separate concerns, minimum image handling should be outsie the bond kernel and only distance vectors should be passed in.

RudolfWeeber (approved):

PR #4785, opened 2023-09-05
add elc test on madelung energy

### LINKED ISSUE #4786: ELC energies (and forces?) might be of by factor 2
Working on the new ELC-IC tutorial (#4784) @Frieda95 and myself realized that there might be a factor 2 missing somewhere.
Looking through the test in [testuite/python/elc.py](https://github.com/espressomd/espresso/blob/python/testsuite/python/elc.py) I realized that there only the interaction of the charges with the walls and applied field is tested.

#4785 proposes to add a test on the Madelung energy: A cubic latttice unit cell is created between two grounded metal interfaces such that the image charge interactions are the same as in an infinite crystal lattice.

According to this, the energy obtained from ELC-IC is twice as large as expected...

schlaicha:
Fixes #4786 

For neutral systems ELC-IC still seems to work perfectly fine.

Description of changes:
- Improve testing of ELC using the Madelung energy

### DIFF
--- a/testsuite/python/elc.py
+++ b/testsuite/python/elc.py
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2010-2022 The ESPResSo project
+# Copyright (C) 2010-2023 The ESPResSo project
 #
 # This file is part of ESPResSo.
 #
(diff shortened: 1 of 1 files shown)


jngrad (approved):
LGTM

PR #4813, opened 2023-10-20
Fix OIF force summary function

jngrad:
Description of changes:
- bugfix: `OifCell.elastic_forces()` no longer throws a `TypeError`
- add a smoke test for that function

### DIFF
--- a/testsuite/python/oif_volume_conservation.py
+++ b/testsuite/python/oif_volume_conservation.py
@@ -27,40 +27,37 @@ import object_in_fluid as oif
 
 
 @utx.skipIfMissingFeatures("MASS")
-class OifVolumeConservation(ut.TestCase):
-
-    """Loads a soft elastic sphere via object_in_fluid, stretches it and checks
-       restoration of original volume due to elastic forces."""
+class Test(ut.TestCase):
 
     system = espressomd.System(box_l=(50., 50., 50.))
     system.time_step = 0.4
     system.cell_system.skin = 0.5
 
-    def check_relaxation(self, **kwargs):
-        self.system.part.clear()
+    def tearDown(self):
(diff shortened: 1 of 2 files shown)


testsuite/python/oif_volume_conservation.py
```diff
@@ -94,16 +91,42 @@ def check_relaxation(self, **kwargs):
...
+        cell = oif.OifCell(cell_type=self.make_cell_type(),
+                           particle_type=0, origin=half_box_l)
+        # stretch cell
+        partcls = self.system.part.all()
+        partcls.pos = (partcls.pos - half_box_l) * 1.5 + half_box_l
+        # reduce number of triangles to speed up calculation
+        cell.mesh.triangles = cell.mesh.triangles[:20]
+        # smoke test
+        results = cell.elastic_forces(el_forces=6 * [0], f_metric=6 * [1])
+        ref = [0., 1.36730815e-12, 22.4985704, 6838.5749, 7.3767594, 6816.6342]
```
christophlohrmann, line 121: where do these nice numbers come from?

jngrad: It's a verbatim copy of the output of the elastic forces calculation. I get the same values on ESPResSo 4.2.1, so the main branch is consistent with the last release. The second value in the list is sensitive to numerical errors and gives different results on ARM vs x86 architectures, which is why a non-zero absolute tolerance is used. As far as I understand it, these numbers should be fully deterministic on a given hardware.

christophlohrmann: not the best physics test then but the best we get I guess

christophlohrmann (approved):

RudolfWeeber (approved):

Issue #4613, opened 2022-11-26
Allow for several systems in a single Python instance


RudolfWeeber:
## Goal

Several instances of `espressomd.system` should be createable in a single Python script.

## Benefits

* The ML group needs this for optimizatnio tasks
* Stuff like Gibbs ensemble or replica exchange could be written down with standard Python theading and without a contorl script talking to two or more other scripts
* The test suite would run much faster, since  a lot of the runtime is consumed by the numpy/scipy modules being loaded over and over again
* Getting rid of the globals in the core, which prevent this at present,  is a goal in it self

## Steps

* [ ] Introduce a system class in the core mirroring espressomd.System

    Note: This can initially be made globally accessible to ease the transition

* [ ] Bundle the remaining "clusters of globals" into strucutes. These are mainly a bunch of globals relating to propagation and equaiton of motion (thermostat)
* [ ] Get rid of mpi callbacks, by moving the functionality to the script interface.

    The big ones are:

    * [ ] Setting/getting particle properties (this is in script interface, but uses the mpi callbacks)
    * [ ] The reaction methods, which make use of mpi callbacks. They have to be either fully parallelized or moved to Python

PR #4816, opened 2023-10-24
Refactor System class

jngrad:
Description of changes:
- encapsulate several global variables inside the `System` class
   - new members: cell_system, integrator, electrostatics, magnetostatics, analysis, comfixed, galilei, bond_breakage
   - required for #4613
   - partial fix for #2628
- make energy/pressure/force calculation functions members of the `System` class
- fully encapsulate event hooks in the `System` class
- simplify checkpointing of the `System` class
- remove `EspressoSystemStandAlone` class
- API changes:
   - `espressomd.galilei.GalileiTransform` is now a member of the `System` class, accessible via `system.galilei`


RudolfWeeber:
CellStructure::set_verlet_skina call of may lead to a situaoitn, where the ParticleDecompositoin needs to be rebuild, or at least the Verlet list becomes invalid. Can we safeguard, that this is not called without hte necessary steps being taken?

RudolfWeeber:
As I understand it, there is code in src/core/energy.cpp that implements member functions of the System class. Should we move all files that direclty contribute to the system class into a common directory src/core/system or the like?

RudolfWeeber:
void integrator_sanity_checks() should probably check for time_step <= 0

RudolfWeeber:
Could you please add some notes as to the purpose of code in leaf.hpp

RudolfWeeber:
Aside form previous commens, LGTM. This PR is a big step towards getting rid of globals and allowing more than one system in a script. Thanks alot.

jngrad:
> void integrator_sanity_checks() should probably check for time_step <= 0

Done.

> CellStructure::set_verlet_skina call of may lead to a situaoitn, where the ParticleDecompositoin needs to be rebuild, or at least the Verlet list becomes invalid. Can we safeguard, that this is not called without hte necessary steps being taken?

Done. I also took this opportunity to move the Verlet skin management and ghost update logic to `CellStructure`.

> Could you please add some notes as to the purpose of code in leaf.hpp

Done. Can be found in a self-contained .dox file that generates `doc/doxygen/html/SystemClassDesign.html`. Writing this down in the `System` class docstring would have forced a rebuild of most of the codebase after each update to the text.

> As I understand it, there is code in src/core/energy.cpp that implements member functions of the System class. Should we move all files that direclty contribute to the system class into a common directory src/core/system or the like?

I thought about it, in fact pressure, forces and tuning also implement `System` methods, but moving them would make the PR diff unreadable, and could introduce merge conflicts with the work of Julian or Thilo. I would prefer to leave these files here for now. It's also quite possible that we later end up merging the associated `*_inline.hpp` directly in the .cpp files, since most of the inline functions are actually only used there now, and can be made static.

RudolfWeeber (approved):

PR #4820, opened 2023-11-21
Propagators

### LINKED ISSUE #4603: Propagation: Add Particle property
Part of #4594

Add a particle property `propagation` which holds a set of flags, determining which propagation schemes should be applied to the particle

### 1. Mode enum in the core

* Add `src/core/PropagationModes.hpp`
* Therein, create an Enum class with the following keys, using values suitable for bitwise logic, i.e. 1,2,4,8

    * `NONE` (0)
    * `TRANS_SYSTEM_DEFAULT`
    * `TRANS_LANGEVIN`
    * `TRANS_VS_RELATIVE`
    * `TRANS_LB_MOMENTUM_EXCHANGE`
    * `TRANS_LB_TRACER`
    * `TRANS_BROWNIAN`
    * `TRANS_STOKESIAN`
    * `ROT_LANGEVIN`
    * `ROT_VS_RELATIVE`
    * `ROT_BROWNIAN`

* add a `is_valid_propagation_combination(int propagation)`. The following are allowed:

    * Any translational mode on its own
    * Any rotational mode on its own
    * The same mode for translation and rotation (where available)
    * translation vs_relative and rotation langevin or vice versa
    * translation lb momentum exchange + translation vs relative
    * These two + rotation langevin or rotation vs_relative
    * Did I miss any?

### Add particle property in core

* in `Particle.hpp` add `int propagation` to the `ParticleProperty` sub-struct and also to the serialization functions of the sub-struct
* in the `Particle` class, add a getter `propagation()` and a setter `set_propagation()`. The latter should use the `is_allowed_propagation_combination()` function from above to validate the given value and throw if it is not valid.
* In `particle_data.hpp` add the communication functions. Search for `gamma_rot` and use as template

### Add the property to the script_interface mapping between Python and C++

* In `src/script_interface/particle_data/Particle.cpp`, add the `propagation` property to the `add_parameters()` call. Again, use `gamma_rot` as template

### Add the Python particle property

* In `src/python/espressomd/` add `propagation.py` (may also be needed in `CMakeLists.txt`), therein define an `IntFlag` (from the enum package) reflecting the enum class in C++. (Note: I'm not aware of a lightweight way to synchronize this automatically).
* in `particle_data.py` add a `@property` for `propagation`.

    * The setter should take an instance of the `propagation` `IntFlag` and pass it to the script interface
    * The getter should take the int coming from the script interface and cast it to the `IntFlag`
    * Use `super().__setattr__` and `super.__getattr__()` to access the original script interface property setter/getter

### Test

* Add a test for the new property in `testsuite/python/particle.py`
* Add a checkpoint test in `testsuite/python/save_checkpoint.py` and  `testsuite/python/test_checkpoint.py`

ThPaul:
Fixes #4603

Description of changes:
- add a per-particle `propagation` flag to control which equations of motion are propagated during integration
- all virtual sites implementations can now be used simultaneously (`system.virtual_sites` was removed)
- virtual sites relative can now be coupled only to the Langevin and LB thermostats (possibly both: Langevin for rotation, LB for translation), using the new `particle.vs_relate_to()` keyword arguments `couple_to_lb=True` and `couple_to_langevin=True` (both are `False` by default)
- finer control over the propagation mode combinations can be obtained by setting the `particle.propagation` bitmask using combinations of `espressomd.propagation.Propagation` enum values; only a small number of combinations are currently allowed, but the list will grow as new propagation schemes are developed


RudolfWeeber:
## Work in progress: Test status

This tries to summarize what is currently tested and which (mixed cases) we are not catching. 

### Single propagaiton scheme
* [x] LB Tracers (virtual_sites_tracers*.py) - tests force distribution from tracer to fluid and advection of tracer
* [x] vs_relative 
    * virtual_sites_relative.py: - test placement, force distribution and runs a fluid of lj dumbells and verifies cofrces. 
    * engine.py, once the python branch is merged - relies on vs orientation. Since this pr let's users control translation and rotation of vs_relative separately, this will have to be revisited.
* [x] Unthermalized velocity verlet (integrate.py) tests Newton's laws
* [X] Velociyt verlet + Langevin: langevin.py (friction, mostly) and langevin_stats.py (Maxwell distribution, velocity auto-correlation function, noise properties. No MSD)
* [X] langevin for particles and box size (npt): thermostat_npt.py and thermostat_npt_stats.py) - however, the latter basically checks against a magic number, as far as I understand
* [X] Brownian dynamics: thermostat_anisotropic.py
* [X] stokesian dynamics: test for two sedimenting spheres against stored data from a paper
* [X] viscous (langevin-like coupling to LB) lb.py::test_viscous_coupling()



### Mixed cases

* [ ] LB tracers + vs_relative + lb viscous coupling
* [ ] LB viscous coupling for translation + langevin friction for rotation
* [ ] vs_relative for translation + langevin for rotation
* [X] Velocity Verlet + vs_relative (virtual_sites_relative.py:test_lj()) (it does not test Langeivn statistics, though), thermostat_anisotropic.py (friction coefficients per Cartesian axis in particle body frame)

### Further points

This is things I stumbled over while going through the tests. It's not all related to this PR.

* thermostat act_on_virtual filtering: we have to understand, whether this becomes obsolte. Probably yes, because one can just not set the respective propagation flag on those particles
* I don't think, we have an integration test for raspberries in a thermalized LB. 
* Are we testing an msd for a particle with Langevin eqn somewhere. I didn't find it in the testsuite, but there is a tutorial.

RudolfWeeber:
### Tentative merge strategy


* whitelist specific propagor settings on the particle level for which we have tessts
* incrementally add tests and enable more propagation modes

This way, we can merge the PR and still make sure that onl tested combinations can be used.


### Step 1

* allow all propagatoins on the system wide level (as it is now  in ES)
* allow setting of (TRANS_VS_RELATIVE+ROT_VS_RELATIVE) and (TRANS_LB_TRACERS) on the particle level


* [ ] Remove virtual flag from particle
* [ ] make sure, only above combinatoins can be set on the particle (aside from system default)
* [ ] adapt the act_on_virtual filter in the lb coupling to exclude vs_relative and lb_tracers propagations
* [ ] adapt other filters for the virtual property of particles, e.g., kinetic energy
* [ ] test non-virtual + vs_relative + inertialess tracers with lb

RudolfWeeber:
@jgrad the particels in testsuite/python/propagation.py are randomly placed. The forces between them are on the order of 1E6. 
Are you that the test failure at 3 mpi cores is actually due to a bug or isn't just really moveing a particle more than a local box length?

src/core/integrate.cpp
```diff
@@ -148,41 +152,109 @@ template <class Kernel> void run_kernel(BoxGeometry const &box_geo) {
...
 
+static void update_default_propagation() {
+  switch (integ_switch) {
+  case INTEG_METHOD_STEEPEST_DESCENT:
+    default_propagation = PropagationMode::NONE;
+    break;
+  case INTEG_METHOD_NVT: {
+    // NOLINTNEXTLINE(bugprone-branch-clone)
+    if (thermo_switch & THERMO_LB & THERMO_LANGEVIN) {
+      default_propagation = PropagationMode::TRANS_LB_MOMENTUM_EXCHANGE;
```
RudolfWeeber, line 139: I don't think, this is correct. Langeivn thermalioitn and LB thermalizaoitn exclude each other.

src/core/integrate.cpp
```diff
@@ -238,66 +310,145 @@ static void resort_particles_if_needed(System::System &system) {
   }
 }
 
+static bool should_propagate_with(Particle const &p, PropagationMode mode) {
+  return p.propagation() & mode or
```
RudolfWeeber: The funciotn should take default_propagation as  an argument and not rely on a global.

src/core/integrate.cpp
```diff
@@ -55,7 +55,9 @@
 #include "signalhandling.hpp"
```
RudolfWeeber, line 57: I'm not sure the if(p.propagaoitn() & bitmask) i sneeded. Isn't that hanlde dby the should_be_propagated function calls beow?

src/core/integrate.cpp
```diff
@@ -238,66 +310,145 @@ static void resort_particles_if_needed(System::System &system) {
...
-    early_exit = steepest_descent_step(particles);
-    break;
-  case INTEG_METHOD_NVT:
-    velocity_verlet_step_1(particles, time_step);
-    break;
+  // steepest decent
+  if (integ_switch == INTEG_METHOD_STEEPEST_DESCENT)
+    return steepest_descent_step(particles);
+
+  int bitmask = PropagationMode::TRANS_NEWTON | PropagationMode::ROT_EULER |
```
RudolfWeeber: Same idiom. I don't really understnad what htis does. Which propagatoisn are listed in the bitmask and what is the logic with the used_propagatois doing?

src/core/integrate.cpp
```diff
@@ -238,66 +310,145 @@ static void resort_particles_if_needed(System::System &system) {
...
+                PropagationMode::TRANS_BROWNIAN | PropagationMode::ROT_BROWNIAN;
+  if (::used_propagations & bitmask) {
+    if (default_propagation & bitmask) {
+      bitmask |= PropagationMode::SYSTEM_DEFAULT;
+    }
+    for (auto &p : particles) {
+      if (p.propagation() & bitmask) {
+        if (should_propagate_with(p, PropagationMode::TRANS_VS_RELATIVE))
+          continue;
+        if (should_propagate_with(p,
```
RudolfWeeber: Mention in a comment that VS Relative will be updated later, once all other particles are propagated.

src/core/integrate.cpp
```diff
@@ -84,6 +86,8 @@
 #endif
 
 int integ_switch = INTEG_METHOD_NVT;
+static int default_propagation = PropagationMode::NONE;
+static int used_propagations = PropagationMode::NONE;
```
RudolfWeeber: Can we avoid used_propagation beig global

src/script_interface/particle_data/ParticleHandle.cpp
```diff
@@ -114,6 +115,78 @@ static auto get_gamma_safe(Variant const &value) {
...
+  case TRANS_BROWNIAN | ROT_BROWNIAN:
+    return force_code_coverage(true);
+#ifdef STOKESIAN_DYNAMICS
+  case TRANS_STOKESIAN | ROT_STOKESIAN:
+    return force_code_coverage(true);
+#endif
+    // other mode combinations
+#ifdef NPT
+  case TRANS_LANGEVIN_NPT | ROT_EULER:
+    return force_code_coverage(true);
```
RudolfWeeber: TRANS_LANGEVIN_NPT | ROT_LANGEVIN. ROT_EULER woudl mean positoins are thermalied but orientaoitns not, that would not make sense.

RudolfWeeber (changes requested):

src/core/integrate.cpp
jngrad: This conditional branch activates translation LB + rotation Langevin. Maybe I misinterpreted your request for implementing the mixed case "LB viscous coupling for translation + langevin friction for rotation"?

src/core/integrate.cpp
jngrad: I started working on that for the other globals when I refactored the thermostats and integrator, but this project is on freeze until this PR converges.

RudolfWeeber:
So the main open issue is coupling of virtual sites.

As I understand it, the cleanest solution would be
* the flags TRANS_VS_RELATIVE and ROT_VS_RELATIVE do not imply any coupling to LB or Langevin noise/friction
* if that's wanted, TRANS_LANGEVIN or TRANS_LB_MOMENTUM_EXCHANGE have to be added to the virtual site
* There will not be a system default for this
* for convencience, p.vs_atuo_ralte_to() gets two additional keyword arguments couple_to_lb=False and apply_langevin_noise_and_frictoin=False, which the user can override when setting up the virtual site. This will then set the respective propagatoin flags on the virtual site.

Is this (in particular the absence of teh posibility to contorl this in one place for the entire system) too anoying for users?
A system default could in principle be implemente dbut it would make the decisoi logic on virtual sites really complicated.

src/script_interface/particle_data/ParticleHandle.cpp
jngrad: Well, ESPResSo's NpT implementation never thermalized angular velocities... But you know how I feel about our NpT implementation anyway :)

```python
import espressomd
system = espressomd.System(box_l=[1., 1., 1.], time_step=0.01)
system.cell_system.skin = 0.
p = system.part.add(pos=[0, 0, 0], v=[1, 2, 3], omega_lab=[1, 2, 3], rotation=3 * [True])
system.thermostat.set_npt(kT=0., gamma0=2., gammav=1e-6, seed=42)
system.integrator.set_isotropic_npt(ext_pressure=0.01, piston=1e6)
system.integrator.run(1)
print(f"{p.v=}")
print(f"{p.omega_lab=}")
```
Output on the python branch and in releases 4.2.1 and 4.1.4:
```
p.v=array([0.9801, 1.9602, 2.9403])
p.omega_lab=array([1., 2., 3.])
```

src/core/integrate.cpp
RudolfWeeber: Ah, OK. How do we prevent the following then:
system wide Langeivn an dLB thermostats with a non-zero gamma translaiton on the Langevin thermostat and
particle.propagaoi = systej default
This would then apply both translational Langevin noise and frictoin and couple the translational degree of feeom to LB, which would be wrong.

src/core/integrate.cpp
jngrad: It's a duplicate of the modes inside the loop. It doesn't seem needed: the testsuite still passes when removed.

src/core/integrate.cpp
jngrad: Once the integrator global variables are packed into a struct, this free function can be changed to a class method.

src/script_interface/particle_data/ParticleHandle.cpp
jngrad: addressed by #4843

src/core/integrate.cpp
jngrad: good point, I have added an extra conditional branch to cover that exact use case

src/core/integrate.cpp
jngrad: Done.

src/core/integrate.cpp
jngrad: It is not necessary. Removed.

src/core/integrate.cpp
jngrad: Removed.

src/core/integrate.cpp
jngrad: Done.

src/core/integrate.cpp
jngrad: Done.

RudolfWeeber (dismissed):

RudolfWeeber:
this implements most of the infrastructure to give people more flexibility w.r.t equations of motion, which is one of the main featues we promised for the next release. 
It turned out to be significantly more complex than initially expected.
Many thanks to Thilo and JN for making it happen.

RudolfWeeber (approved):

PR #4826, opened 2023-11-24
WIP add cutoff by type

### LINKED ISSUE #4819: Interface for querying non-bonded interactions
It should be possible to find out

* if only pairwise central forces are active (for struct-of-arrays based force calculation)
* What is the maximum required cutoff for a given list of particle types, or for all particle types except a given list of types (for automoatically determining the cutoff for the RegularDecompoistoin which is part of the HybridDecompositoin)

christophlohrmann:
Fixes #4819 

Description of changes:
- Added cutoffs by type
- no guarantees with regards to execution speed


christophlohrmann:
@RudolfWeeber does this look like what you had in mind?

RudolfWeeber:
LGTM. Just please rename the functions working on type lists to make it clear that they only concern non_bonded interactions
* Dipoles are also a non-central interation.

RudolfWeeber:
Once this is merged, the hzbrid decompositoin can be adapted to determine the cutoff automtically. However, then, electrostatics and dipolar and collision cutoffs need to be taken into account.

christophlohrmann:
> LGTM. Just please rename the functions working on type lists to make it clear that they only concern non_bonded interactions
> 
> * Dipoles are also a non-central interaction.

1. Is the rename really necessary if the functions are members of  ``InteractionsNonBonded``
2. which of the nonbonded interactions is "dipoles"?

PR #4827, opened 2023-11-24
LB+particles: guard against missed coupling due to round-off error

### LINKED ISSUE #4825: LB: Particle not coupled due to rounding error
Under veyr specific circumstances, particles are not fully coupled to LB:

Trigger: a particle coordinate is -epsilon but epsilon + box_l is = box_l due to a rounding error. In this case, neither the original position nor the folded positoin (-eps+box_l) is in the local domain and no force is added to the particle.

Probability of this occurring: probably very low in a system with randomly placed particles, but likely, if particles are initialized with a zero in one of the positoin coordinates. this can easily propagate to -epsilon due small numerical errors in the LB forces

MNWE (but this might be compiler dependent):
```
import espressomd

import espressomd.lb

import numpy as np

system = espressomd.System(box_l=[50,50,40])

system.time_step =0.01
system.cell_system.skin = 0.4

sigma = 1





if hasattr(espressomd.lb, "LBFluidWalberla"):
    lbf = espressomd.lb.LBFluidWalberla(
      agrid=1,
      density=1,
      tau=system.time_step,
      kinematic_viscosity=1)
else:
    lbf = espressomd.lb.LBFluid(
      agrid=1,
      dens=1,
      tau=system.time_step,
      visc=1)
if hasattr(system, "actors"): 
    system.actors.add(lbf)
else:
    system.lb = lbf
system.thermostat.set_lb(LB_fluid=lbf, gamma=1,seed=1, act_on_virtual=False)

p1 = system.part.add(pos=(0,-1E-25,0),v=(-1,0,0))
p2 = system.part.add(pos=(30,0,0),v=(1,0,0))

#system.thermostat.set_langevin(kT=0,gamma=1,seed=2)
for i in range(500):
  system.integrator.run(1)
  print(system.distance(p1,p2), p1.v[0],p2.v[0],p1.f[0],p2.f[0])
  np.testing.assert_allclose(np.copy(p1.f), -np.copy(p2.f), atol=1E-5)
```

Likely, both, the dev version and 4.2 are affected.

RudolfWeeber:
Fixes #4825

Description of changes:
* Bugfix: particles outside the simulation box are now properly coupled using PBC. The coordinates are folded before considering shifted positions in the LB particle coupling code. Also, a test that fails without the fix is added. To my understanding, if the particle position is folded, and then all combination of `folded_pos[i]` +/- `box_l[i]` for all Cartesian directions are considered, both the particle in the primary box and all potential halo regions are caught.

### DIFF
--- a/src/core/lb/particle_coupling.cpp
+++ b/src/core/lb/particle_coupling.cpp
@@ -251,7 +251,6 @@ void ParticleCoupling::kernel(Particle &p) {
   Utils::Vector3d force_on_particle = {};
   auto folded_pos = box_geo.folded_position(p.pos());
 
-
 #ifdef ENGINE
   if (not p.swimming().is_engine_force_on_fluid)
 #endif
--- a/src/core/lb/particle_coupling.cpp
+++ b/src/core/lb/particle_coupling.cpp
@@ -249,11 +249,13 @@ void ParticleCoupling::kernel(Particle &p) {
 
   // Calculate coupling force
   Utils::Vector3d force_on_particle = {};
+  auto folded_pos = box_geo.folded_position(p.pos());
+
 
 #ifdef ENGINE
   if (not p.swimming().is_engine_force_on_fluid)
 #endif
-    for (auto pos : positions_in_halo(p.pos(), box_geo, agrid)) {
+    for (auto pos : positions_in_halo(folded_pos, box_geo, agrid)) {
       if (in_local_halo(pos, agrid)) {
         auto const vel_offset = lb_drift_velocity_offset(p);
         auto const drag_force = lb_drag_force(m_lb, p, pos, vel_offset);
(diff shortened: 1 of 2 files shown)


RudolfWeeber:
in algorithsm/periodic_fold.hpp, first the lower, then the upper limit is checked, so here
-epsilon + box_l = box_l (due to round-off error)
but box_l >= box_l so
folded_pos = box_l - box_l =0

Let's consider the other case:
box_l +epislon is folded to
box_l +epsilon -box_l 
even if this evaluates to 0 (rather than epsilon) due to rounding, the resulting coordinate is still folded, ie., 0<=x<box_l

box_l-epsilon is < box_l so the folding code does nothing, the coordinate is already folded.

So as I understand it, Algorithm::periodic_fold() is correct in all cases. Can someone please double check this?

christophlohrmann (commented):
I agree with the code changes and all arguments made.
Still, a cpp unit test for algorithm/periodic_fold that explicitly tests the cases mentioned in @RudolfWeeber's comment would support the argument better than my agreement.
Also the python test could look not only at the fact that the particle *is* coupled, but also could check that the particle is coupled correctly, i.e. momentum conserving.

jngrad:
@RudolfWeeber's argument on the periodic folding w.r.t. roundoff errors is supported by the C++ unit tests:
https://github.com/espressomd/espresso/blob/2d656671063da890046dd6dd59d6d3ae88b403a5/src/core/unit_tests/periodic_fold_test.cpp#L101-L110
Here `res.first` (position) is 0.0 and `res.second` (image box count) is 0, so the folding was applied twice. This test also checks for the case box+epsilon, and for a non-zero image box counter.

jngrad (approved):

PR #4845, opened 2023-12-21
Thermostats refactoring

### LINKED ISSUE #3980: Split thermostats into separate classes
There is currently a single `Thermostat` class in the python interface that mixes the LB, NpT, DPD, Langevin, Brownian and Stokesian thermostats. This design is difficult to maintain, especially in the [checkpointing methods](https://github.com/espressomd/espresso/blob/dc0dfa199e9b2d5e59c3a5176ccceb8f8c99ce14/src/python/espressomd/thermostat.pyx#L110-L240) where all thermostat internals are visible in the same scope.

The `Integrate` class was in a similar situation in 4.1, and was split up into individual classes in 4.2 without breaking backward compatibility (#3390). We could apply the same refactoring to the `Thermostat` class. The difference with the `Integrator` class is that multiple thermostats can be active simultaneously. In the new design, `Thermostat` would store a list of currently active thermostats.

This new design would allow us to inactivate a single thermostat, as we do with `Actors`. At the moment, to inactivate a thermostat we have to inactivate all thermostats and re-enable those we didn't intend to inactivate, which means keeping their original parameters in a variable and setting them up again, *in the same order*, because each thermostat overwrites the global temperature variable.

### LINKED ISSUE #4266: Refactor thermostat switches
The low-level handling of how thermostats are enabled or disabled should not be exposed to the developer. Something like the following could be implemented:

``` c++
#include <bitset>
#include <cassert>
#include <iostream>

enum class Thermostat : unsigned int {
    LANGEVIN,
    DPD,
    NPT_ISP,
    NPT_ISO,
    LB,
    BROWNIAN,
    SD
};

template<std::size_t N>
struct ThermoSwitch {
  std::bitset<N> thermo_switch;
  void add(Thermostat thermostat) {
    assert(static_cast<unsigned int>(thermostat) < N);
    thermo_switch.set(static_cast<unsigned int>(thermostat), true);
  }

  void remove(Thermostat thermostat) {
    assert(static_cast<unsigned int>(thermostat) < N);
    thermo_switch.set(static_cast<unsigned int>(thermostat), false);
  }

  bool is_active(Thermostat thermostat) {
    assert(static_cast<unsigned int>(thermostat) < N);
    return thermo_switch[static_cast<unsigned int>(thermostat)];
  }

  void turn_off() {
    thermo_switch.reset();
  }
};

int main() {
   ThermoSwitch<7> thermo_switch;
   thermo_switch.add(Thermostat::LANGEVIN);
   std::cout << thermo_switch.thermo_switch << std::endl;
   thermo_switch.add(Thermostat::DPD);
   std::cout << thermo_switch.thermo_switch << std::endl;
   std::cout << thermo_switch.is_active(Thermostat::DPD) << std::endl;
   thermo_switch.turn_off();
   std::cout << thermo_switch.thermo_switch << std::endl;
   std::cout << thermo_switch.is_active(Thermostat::DPD) << std::endl;
}
```

### LINKED ISSUE #4848: LB particle coupling thermostat seed doesn't fully decorrelate independent simulations
**TL;DR**: the `seed` argument of `system.thermostat.set_lb()` doesn't actually set the random number generator (RNG) seed of the LB particle coupling code. Instead, it sets the lag time in the RNG sequence. Thus two independent simulations with a different seed will exhibit correlated dynamics. The LB fluid seed is not affected by this issue.

## Background

We use Philox counters to generate reproducible and cross-platform random number sequences. These are used to calculate the friction noise during thermalization. The RNG sequence is fully determined by 4 input arguments:
* the RNG salt, which is a unique id for each kind of thermostat (LB particle coupling, Langevin, Brownian, Stokesian, etc.)
* the RNG seed, which the user has control over
* two keys, which are typically particle ids (for LB particle coupling, the second key is always zero)

Running the same simulation twice with the same RNG seed produces the exact same trajectory, up to machine precision. Running the same simulation twice with a different RNG seed produces a different trajectory. For the latter, the magnitude of the difference between the two seeds is irrelevant; the RNG seed "decorrelates" RNG sequences, just like the RNG salt does. This is achieved by internally concatenating the salt and seed into a single 64bit integer.

The Philox RNG has an internal counter which is incremented at each time step. This counter essentially controls the current position in the RNG sequence. We do not provide a convenient way for the user to initialize this value. This counter value is saved during checkpointing.

## Problem statement

For historical reasons, thermostats in ESPResSo were all implemented differently, and didn't share a common base class. Yet they shared a unified Python interface, which could give users the false impression that they were related. This lead to bad user experience in the 4.1 line of ESPResSo where one thermostat was using the Mersenne Twister RNG instead of the Philox RNG.

The LB particle coupling thermostat in particular was implemented with a hardcoded `seed=0`, regardless of which `seed` the user gives as argument to `system.thermostat.set_lb()`. Yet the particle forces are different when giving a different seed, because the value of the `seed` argument is stored as the internal `counter` of the Philox RNG. Thus the LB particle coupling RNG sequence is always the same for all simulations, and the `seed` argument actually controls the initial time lag in that RNG sequence. This issue is specific to the LB particle coupling code, and does not affect the LB fluid code.

As a result, two independent simulations that differ only in the LB particle coupling seed with not truly be independent: the particle forces from simulation A will be correlated to the particle forces from simulation B with a time lag. This correlation is weak, since it's "diluted" by the LB fluid noise and all other interactions in the system.

It would be useful to know how weak this correlation truly is, and whether it can affect some observables more than other, for example Green-Kubo relations. To that end, I adapted the polymers tutorial to investigate how the correlation between independent simulations that differ only by the LB particle coupling seed affects simulation results. I measured the diffusion coefficient of a polymer with 4 beads connected by FENE bonds, repelled by a WCA potential, and coupled to a thermalized fluid in a 8x8x8 grid in units of the WCA sigma. The simulation was repeated 48 times for two scenarios: one where the Philox seed is 0 and the Philox counter initial value is controlled by the user ("current behavior"), and one where the Philox seed is controlled by the user and the Philox counter initial value is always 0 ("proposed behavior").

Figure 1 plots the mean and standard error of the mean of the diffusion coefficient as a function of the simulation time for both scenarios. We can see in the proposed behavior scenario that the error band tapers for longer simulation times, as expected. In the current behavior scenario, the error band is extremely narrow right at the beginning, and tapers more slowly. Figure 2 plots the standard error of the mean as a function of time to help better visualize how the correlation between independent measurements affects the magnitude of the standard error of the mean.

In my opinion, this is a design flaw. Data production is affected in two ways:
1. error bars in plots are artificially shrunk by a factor 2.5, because the underlying assumption of independent samples in the formula of the standard of the mean is not satisfied
2. users might decide to simulate for shorter times, because these independent simulations converge to the same value much faster than they should

Point 1 doesn't affect the samples mean, thus the calculated data point (or the shape of the curve for time-dependent observables) converges to the real value. Point 2 can affect the sample mean due to insufficient sampling time.

In the light of the data produced by this experiment, I'm not convinced we can simply simulate for longer to get the LB fluid RNG to introduce enough entropy to fully decorrelate independent simulations. The current behavior scenario was run again with a hardcoded seed of 1 (data not shown) to make sure my observations were not due to a statistical fluke with the seed of 0, and the same trend emerged.

<br>

Plot of the diffusion coefficient as a function of the simulation time, for both the current behavior of ESPResSo, and the new proposed behavior. In the current behavior, the standard error of the mean is extremely narrow and much smaller in magnitude than the magnitude of the fluctuations of the mean between two at short sampling times. In the proposed behavior, the standard error of the marge is more reasonable and contains the natural fluctuation of the mean at short simulation times, and tapers progressively towards longer simulation times.

<p align="center">
<b>Figure 1</b>: Convergence analysis of the polymer diffusion coefficient.
</p>

<br>

Plot of the standard error of the mean of the diffusion coefficient as a function of the simulation time. The proposed behavior is initially 3 times larger in magnitude compared to the current behavior, and 2.2 times larger towards large simulation times.

<p align="center">
<b>Figure 2</b>: Comparison of the standard error of the mean in both scenarios.
</p>

## Proposed solution

Make the `seed` parameter set the Philox seed, instead of the Philox counter. This was implemented by commit 9550aa1f24b215c090717a6b870b293293e9f535 in #4845. This fix could be backported to the 4.2 line of ESPResSo with an extra MPI callback and 3 lines of Cython code.

## Supporting information

<details>
<summary>Simulation script</summary>

```python
import sys
import numpy as np
import scipy.optimize

import espressomd
import espressomd.analyze
import espressomd.observables
import espressomd.polymer
import espressomd.lb

espressomd.assert_features(['WCA', 'WALBERLA'])

def build_polymer(system, n_monomers, polymer_params, fene):
    positions = espressomd.polymer.linear_polymer_positions(
        beads_per_chain=n_monomers, **polymer_params)
    p_previous = None
    for i, pos in enumerate(positions[0]):
        p = system.part.add(pos=pos)
        if p_previous is not None:
            p.add_bond((fene, p_previous))
        p_previous = p

def correlator_gk(pids_monomers, tau_max):
    com_vel = espressomd.observables.ComVelocity(ids=pids_monomers)
    com_vel_cor = espressomd.accumulators.Correlator(
        obs1=com_vel, tau_lin=16, tau_max=tau_max, delta_N=10,
    return com_vel_cor

def gk_integration(N, tau, acf):
    x = np.arange(2, 28)
    y = [1 / 3 * np.trapz(acf[:j], tau[:j]) for j in x]
    diffusion_coef = np.mean(y[10:])
    return diffusion_coef, tau[x], y

# Setup constants
BOX_L = 8.0
TIME_STEP = 0.01
LOOPS = 100
STEPS = 100
KT = 1.0
GAMMA = 5.0
POLYMER_PARAMS = {'n_polymers': 1, 'bond_length': 1, 'seed': 42, 'min_distance': 0.9}

# System setup
system = espressomd.System(box_l=3 * [BOX_L])
system.cell_system.skin = 0.4

# Lennard-Jones interaction
system.non_bonded_inter[0, 0].wca.set_params(epsilon=1.0, sigma=1.0)

# Fene interaction
fene = espressomd.interactions.FeneBond(k=7, r_0=1, d_r_max=2)
system.bonded_inter.add(fene)

com_vel_tau_results = []
com_vel_acf_results = []

N = 4
seed = int(sys.argv[-2])
counter = int(sys.argv[-1])
build_polymer(system, N, POLYMER_PARAMS, fene)
# energy minimization
system.time_step = 0.002
system.integrator.set_steepest_descent(
    f_max=1.0,
    gamma=10,
    max_displacement=0.01)
system.integrator.run(2000)
# thermalization with Langevin
system.integrator.set_vv()
system.time_step = TIME_STEP
system.thermostat.set_langevin(kT=1.0, gamma=50, seed=42)
system.integrator.run(2000)
system.thermostat.turn_off()
# thermalization with LB
lbf = espressomd.lb.LBFluidWalberla(kT=KT, seed=42, agrid=1, density=1,
                                    kinematic_viscosity=5, tau=system.time_step,
                                    single_precision=True)
system.lb = lbf
system.thermostat.set_lb(LB_fluid=lbf, gamma=GAMMA, seed=seed, philox_counter=counter)
system.integrator.run(1000)

# configure Green-Kubo correlator
com_vel_cors = []
n_loops = [1, 2, 4, 8, 12, 16, 24, 32, 40, 48, 56, 64, 80, 96, 112, 128, 160, 256, 384, 512]
n_steps = []
for i in n_loops:
    com_vel_cor = correlator_gk(np.arange(N), LOOPS * STEPS)
    system.auto_update_accumulators.add(com_vel_cor)
    com_vel_cors.append(com_vel_cor)

def save_data(n_steps, com_vel_tau_results, com_vel_acf_results):
    com_vel_tau_results = np.array(com_vel_tau_results)
    com_vel_acf_results = np.reshape(com_vel_acf_results, [len(n_steps), -1])

    diffusion_gk = []
    diffusion_time = []
    for length, tau, acf in zip(n_steps, com_vel_tau_results, com_vel_acf_results):
        diffusion_coef, xdata, ydata = gk_integration(N, tau, acf)
        diffusion_gk.append(diffusion_coef)
        diffusion_time.append(length * TIME_STEP)

    np.save(f"lb_seed_{seed}_counter_{counter}.npy", [diffusion_gk, diffusion_time])

last_sim_steps = 0
for i, n in enumerate(n_loops):
    com_vel_cor = com_vel_cors[i]
    sim_total = n * LOOPS * STEPS
    system.integrator.run(sim_total - last_sim_steps)
    system.auto_update_accumulators.remove(com_vel_cor)
    last_sim_steps = sim_total
    com_vel_cor.finalize()
    com_vel_tau_results.append(com_vel_cor.lag_times())
    com_vel_acf_results.append(com_vel_cor.result())
    n_steps.append(sim_total)
    save_data(n_steps, com_vel_tau_results, com_vel_acf_results)
```

</details>

<details>
<summary>Launch script</summary>

```sh
for counter in {1..16}; do ./pypresso mwe.py 0 $counter & done
for seed in {1..16}; do ./pypresso mwe.py $seed 0 & done
```

</details>

<details>
<summary>Data analysis</summary>

```python
import numpy as np
plt.rcParams.update({'font.size': 16})

def parse(n, pat, burn=0):
    root = "./"
    matrix = []
    for i in range(n):
        diffusion_gk, diffusion_time = np.load(f"{root}/{pat.format(i+1)}")
        matrix.append(diffusion_gk)
    n = min(len(x) for x in matrix)
    return diffusion_time[burn:n], np.array([x[burn:n] for x in matrix])

df_old_time, df_old_coef = parse(48, "lb_seed_0_counter_{}.npy", 2)
df_new_time, df_new_coef = parse(48, "lb_seed_{}_counter_0.npy", 2)

df_old_mean = np.mean(df_old_coef, axis=0)
df_old_stderr = np.std(df_old_coef, axis=0) / np.sqrt(df_old_coef.shape[1])
df_new_mean = np.mean(df_new_coef, axis=0)
df_new_stderr = np.std(df_new_coef, axis=0) / np.sqrt(df_new_coef.shape[1])

fig, (ax1, ax2) = plt.subplots(1, 2, sharey=True, figsize=(11, 4))
ax2.fill_between([], [], [])
ax1.fill_between(df_old_time, df_old_mean - df_old_stderr, df_old_mean + df_old_stderr, alpha=0.5)
ax2.fill_between(df_new_time, df_new_mean - df_new_stderr, df_new_mean + df_new_stderr, alpha=0.5)
ax1.plot(df_old_time, df_old_mean, label="Current behavior")
ax2.plot([], [])
ax2.plot(df_new_time, df_new_mean, label="Proposed behavior")
ax1.set_xlabel("Number of time steps")
ax2.set_xlabel("Number of time steps")
ax1.set_ylabel(r"Diffusion coefficient [$\sigma^2/t$]")
ax1.legend(loc="lower right")
ax2.legend(loc="lower right")
plt.tight_layout()
plt.show()

fig = plt.figure(figsize=(8, 4))
plt.plot(df_old_time, df_old_stderr, label="Current behavior")
plt.plot(df_new_time, df_new_stderr, label="Proposed behavior")
#plt.yscale('log')
#plt.xscale('log')
plt.legend()
plt.xlabel("Number of time steps")
plt.ylabel("Standard error of\n" r"the mean [$\sigma^2/t$]")
plt.tight_layout()
plt.show()
```

</details>

jngrad:
Description of changes:
- encapsulate thermostats and remove their Cython interface
   - use one C++ `ScriptInterface` class per core class (fixes #3980)
   - introduce the `ThermostatFlags` enum, allow deactivating a single thermostat, add an `is_active` flag accessible from the Python interface (fixes #4266)
- remove 20 MPI callbacks
   - partial fix for #4614
- remove thermostat and thermalized/rigid bond global variables
   - partial fix for #2628
- remove a significant amount of historical Cython code that relied on deprecated Cython features
   - partial fix for #4542
   - ESPResSo is now compatible with Cython 3.0
   - the `nogil` deprecation warning subsists
- improve testing of the thermostats interface
- API changes:
   - thermalized bond parameter `seed` was moved to `system.thermostat.set_thermalized_bond()`
      - this change better reflects the fact there is only one global seed for all thermalized bonds
      - until now this global seed was overwritten by any newly created thermalized bond, whether it was added to the system or not
   - LB thermostat `seed` parameter now gets written to the RNG seed (silent change)
      - until now, the RNG seed was hardcoded as `0` and the `seed` value was used to set the RNG counter, thus two independent simulations with a different seed would actually get the same particle coupling random noise sequence with a time lag equal to the difference between the two seeds
      -  this is a bugfix for ensemble runs
      - fixes #4848


jngrad:
Regarding the API change for thermalized bonds: the thermalized bond thermostat had always used a single, global RNG state that would be silently overwritten by each new thermalized bond. The new API no longer gives the false impression that each bond maintains its own RNG. The thermostat infrastructure is likely going to change in the near future anyway.

We should really consider rewriting the DPD thermostat to no longer get incremented by external components like constraints. This would allow us to have a single shared RNG counter from which all thermostats could extract their random numbers (the RNG salt decorrelates them). Then the DPD non-bonded potential and thermalized bonds would behave like any other interaction, since they would just fetch the RNG values from the system thermostat. This would also make the thermostat script interface a lot more maintainable. Right now it's quite convoluted, because the specificities of each thermostat prevent me from applying the SOLID principles. In particular, when turning a thermostat on, the arguments are dispatched to the thermostat container (`kT`) and to the actual thermostat (`gamma`, `seed`, etc.), which means that throwing a C++ exception requires rolling back the state of two script interface objects (both of which interact with the system class through distinct event code paths during that rollback) to gracefully restore the system into a valid state. I tried to refactor things further, but it proved to be a can of worm, and I could not find a solution that would both preserve the current API and play nice with the checkpointing mechanism.

In any case, I think this refactoring is a big improvement over the original Cython code, and from my side can be reviewed and merged as is. We can then consult with our users during the next software meeting how we want to proceed with further refactoring of the thermostats, and decide on which API changes would be acceptable. I would prefer not to delay the merge decision for this PR, because it simplifies the LB particle coupling code quite a bit and removes highly coupled global variables, and I would like to leverage that in my ongoing GPU LB work resp. ongoing globals removal work.

jngrad:
There might actually be a bug in the LB particle coupling RNG. I have produced a MWE in #4848.

src/python/espressomd/script_interface.pyx
```diff
@@ -192,6 +193,7 @@ cdef class PScriptInterface:
         for name, value in kwargs.items():
             self.sip.get().set_parameter(utils.to_char_pointer(name),
                                          python_object_to_variant(value))
+            utils.handle_errors(f"while setting parameter '{name}'")
```
reinaual, line 196: Do we care about the additional mpi-gathers introduced here? Do I get it correctly: This means that we have code left that adds our custom errors when calling a setter? When does this happen?

testsuite/python/particle.py
```diff
@@ -113,6 +113,8 @@ def test_propagation_enum(self):
         Propagation = espressomd.propagation.Propagation
         flags_core = self.system.call_method("get_propagation_modes_enum")
         flags_si = {e.name: e.value for e in Propagation}
+        # Python 3.11+ skips empty bitfields during enum iteration
+        flags_si["NONE"] = Propagation.NONE
```
reinaual, line 117: Out of interest, you have a link where this behaviour is described?

testsuite/scripts/tutorials/test_polymers.py
```diff
@@ -50,7 +50,7 @@ def test_diffusion_coefficients(self):
         # polymer diffusion
         ref_D = [0.0363, 0.0269, 0.0234]
         np.testing.assert_allclose(tutorial.diffusion_msd, ref_D, rtol=0.30)
-        np.testing.assert_allclose(tutorial.diffusion_gk, ref_D, rtol=0.15)
+        np.testing.assert_allclose(tutorial.diffusion_gk, ref_D, rtol=0.30)
```
reinaual, line 53: Why do we have to increase the tolerance?

doc/tutorials/polymers/polymers.ipynb
```diff
@@ -307,7 +307,7 @@
     "                                        kinematic_viscosity=5, tau=system.time_step,\n",
     "                                        single_precision=True)\n",
     "    system.lb = lbf\n",
-    "    system.thermostat.set_lb(LB_fluid=lbf, gamma=gamma, seed=42)"
+    "    system.thermostat.set_lb(LB_fluid=lbf, gamma=gamma, seed=0)"
```
reinaual, line 310: Is this a cosmetic change only?

src/script_interface/thermostat/thermostat.hpp
```diff
@@ -0,0 +1,731 @@
...
+            return;
+          }
+#ifdef PARTICLE_ANISOTROPY
+          static_assert(std::is_same_v<T, Utils::Vector3d>);
+          T gamma{};
+          try {
+            gamma = T::broadcast(get_value<double>(v));
+          } catch (...) {
+            gamma = get_value<T>(v);
+          }
```
reinaual: Why do we need this try-catch monstrosity? Do we allow to set a scalar `gamma` with `PARTICLE_ANISOTROPY` enabled?

src/core/lb/particle_coupling.hpp
```diff
@@ -83,79 +58,56 @@ void add_md_force(LB::Solver &lb, Utils::Vector3d const &pos,
...
   double m_noise_pref_wo_gamma;
+  bool m_thermalized;
 
 public:
-  ParticleCoupling(LB::Solver &lb, double time_step, double kT)
-      : m_lb{lb}, m_thermalized{kT != 0.}, m_time_step{time_step} {
-    assert(kT >= 0.);
+  ParticleCoupling(LBThermostat const &thermostat, LB::Solver &lb,
+                   BoxGeometry const &box_geo, LocalBox const &local_box,
+                   double time_step, double kT = -1.)
```
reinaual: Is this really a good design choice? I'm not a fan of "changing the coupling behaviour" based on an invalid default argument... However, if it's the best choice for the moment, leave it as it is.

src/python/espressomd/script_interface.pyx
jngrad: This can happen when setting a property of the system (e.g. `box_l`, `periodicity`) or calling a particle's convenience function `vs_auto_relate_to()`. There might more examples, but it's hard to tell. We have the same logic a bit earlier in the class inside the `call_method()` method. I think this change is helpful, because if we forget to add `utils.handle_errors()` in the consumer class (in this case `System` and `Particle`), the error message will be queued, the system will roll back to the previous valid state (therefore the current line of Python code in the simulation script had no effect), and the next call to `call_method()` will necessarily emit the queued warning, which is going to be very confusing for the user.

testsuite/python/particle.py
jngrad: Here is the commit that introduced the change in behavior: python/cpython@7aaeb2a3d682ecba125c33511e4b4796021d2f82

The official documentation updated the first paragraph of the `enum` page (compare [3.10](https://docs.python.org/3.10/library/enum.html) vs. [3.11](https://docs.python.org/3.11/library/enum.html)) to reflect that change. Iterating an `enum.IntFlag` should now only yield non-zero powers of 2, if I understand it correctly. They did not put a colored box about the change in behavior introduced by 3.11 in the `enum.IntFlag` section of the `enum` page. Took me a while to troubleshoot that failing test on Ubuntu noble :sweat_smile:

testsuite/scripts/tutorials/test_polymers.py
jngrad: I think the original tolerances in this test were just chosen based on the largest error plus a small safety margin for ARM architectures, which contrary to x86_64 architectures don't use extended precision floating-point numbers. When using a different seed, the magnitude of the error in the Green-Kubo relation becomes similar to the magnitude of the error of the MSD method.

doc/tutorials/polymers/polymers.ipynb
jngrad: This change is necessary to reproduce the exact same RNG sequence as on the python branch. The RNG sequence is then sampled 42 time steps ahead of the python branch, because it starts at 0, while on the python branch it started at 42. This change is actually what helped me outline the proof-of-concept in bug report #4848.

src/script_interface/thermostat/thermostat.hpp
jngrad: Yes, and the original Cython code to handle that case looked even uglier :laughing:

src/core/lb/particle_coupling.hpp
jngrad: I agree it's not great, but I wanted to avoid having two constructors. The core uses the default constructor with a negative temperature and gets the actual temperature from the LB fluid. The unit test passes a custom temperature to verify particle coupling is properly thermalized. I'll keep that in mind when rewriting the unit tests as part of #4734 and look at the feasibility of instantiating a LB fluid with a different temperature in every test to remove this default argument.

src/script_interface/thermostat/thermostat.hpp
reinaual: I understand that we don't want to divert from the original behaviour now, but I think this is really bad design and the interface should reflect what is actually happening...

reinaual (dismissed):

reinaual (approved):

src/core/lb/particle_coupling.hpp
jngrad: A much cleaner solution was now pushed to this PR.

Issue #4872, opened 2024-03-05
Document use of virtual environments


jngrad:
[PEP 0668](https://peps.python.org/pep-0668/) changes the behavior of `pip` to avoid accidentally breaking system-wide packages. This is achieved by disallowing the installation of packages via `pip`, including `pip install --user`, when outside of a virtual environment. This new policy has been adopted in Ubuntu 23.04 and will affect all users who migrate from Ubuntu 22.04 LTS to Ubuntu 24.04 LTS next month.

Virtual environments solve a lot of pain points that afflicted ESPResSo users over the years:
* getting the formatting and linting tools properly set up is difficult for new contributors, as we require a specific version that may conflict with the version requirements of other Python projects they contribute to, and the date at which we update the version requirements is unpredictable (we usually target Ubuntu LTS releases, but we often miss that target by several months)
* getting a compatible Jupyter implementation typically requires installing packages that break the dependency tree of other Jupyter implementations; this was the case in 2023 when we moved from Jupyter Notebook (Classic) to JupyterLab on ICP workstations and could no longer use Jupyter Notebook
* virtual environments can store the ESPResSo build folder in a `.pth` file, so that ESPResSo can be imported in Python without the need for the `pypresso` wrapper script, and without the need to touch the `$PYTHONPATH` (which would break `module load` and `module purge` in clusters); in fact many of our users still call `pypresso` on their simulation scripts even though it is not meant to be used in that way (but rather to help run debugging sessions: the script actually sets up paths to help find shared libraries of popular debuggers, and parses command line options; to be able to `import espressomd`, run `make install`)

Cluster users who `module load` ESPResSo instead of building it from sources should rely on the following syntax to re-use loaded Python packages instead of installing them again (and potentially breaking the dependency tree):

```sh
python3 -m venv --system-site-packages venv
source venv/bin/activate
python3 -m pip install -r requirements.txt
deactivate
```

For more details, see pm-blanco/pyMBE#12

jngrad:
Addressed by #4922.

PR #4884, opened 2024-03-17
WIP: Make espresso pip-installable

### LINKED ISSUE #4222: Make espresso "pip installable" by adding a setup.py file
... just like any other python package

jhossbach:
Closes #4222 

`scikit-build` is used to build espresso with CMake and install `espressond` in the `site-packages` directory.
This should also resolve the issue with paths in virtual environments without needing to add a `.pth` file (#4872) and should make espresso pretty easy to install in any environment.

Things I haven't touched:
 - Adding metadata to the `setup.py`, such as the license or description.
 - Building should happen in parralel with `ninja` but I am not sure if this was the case on my PC, so parallel building should be checked.
 - Documentation
 - of course updating the CI etc.


PythonFZ:
Are there any updates on this? I see that the package compiles given these changes but either installing it via `uv pip install .` or editable via `uv add` / `uv sync` gives me an import error `espressomd` not found.

PR #4898, opened 2024-04-02
Improve user guide chapters on observables and constraints

### LINKED ISSUE #4897: CI/CD: tutorial job times out
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/pipelines/21854

jngrad:
Fixes #4897

Description of changes:
- fix issues in the description of observable return values and array shapes
- fix vector notation in the expressions of single-particle force constraints
- make notation more consistent across particle-based observables

### DIFF
--- a/src/python/espressomd/constraints.py
+++ b/src/python/espressomd/constraints.py
@@ -194,7 +194,7 @@ class HomogeneousMagneticField(Constraint):
 
     :math:`U = -\\vec{\\mu} \\cdot \\vec{H}`
 
-    where :math:`\\vec{\\mu}` and :math:`\\vec{r}` are the particle dipole moment and position.
+    where :math:`\\vec{\\mu}` is the particle dipole moment.
 
     Attributes
     ----------
@@ -431,14 +431,15 @@ class LinearElectricPotential(Constraint):
     """
     Electric potential of the form
 
-    :math:`\\phi = -\\vec{E} \\cdot \\vec{r} + \\phi_0`,
+    :math:`\\phi = -\\vec{E} \\cdot \\vec{x} + \\phi_0`,
 
     resulting in the electric field :math:`\\vec{E}` everywhere.
     The resulting force on the particles are then
 
     :math:`\\vec{F} = q \\cdot \\vec{E}`
 
-    where :math:`q` and :math:`\\vec{r}` are the particle charge and position.
+    where :math:`q` and :math:`\\vec{x}` are the particle charge and position
+    in folded coordinates.
     This can be used to model a plate capacitor.
 
     Arguments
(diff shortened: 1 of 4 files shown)


pm-blanco (approved):
It is much more clear to me now. Thank you for taking care of this!

src/python/espressomd/constraints.py
```diff
@@ -420,21 +431,20 @@ class LinearElectricPotential(Constraint):
     """
     Electric potential of the form
 
-    :math:`\\phi = -E \\cdot x + \\phi_0`,
+    :math:`\\phi = -\\vec{E} \\cdot \\vec{r} + \\phi_0`,
```
reinaual: Actually I'm not a big fan of this change. For me $\vec{r}$ is a distance vector, whereas $\vec{x}$ is the absolute position. And since this is one of the few places the absolute position is relevant, I think the documentation should reflect that.

src/python/espressomd/constraints.py
```diff
@@ -463,15 +473,15 @@ class ElectricPlaneWave(Constraint):
     """
     Electric field of the form
 
-    :math:`E = E0 \\cdot \\sin(k \\cdot x + \\omega \\cdot t + \\phi)`
+    :math:`\\vec{E} = \\vec{E_0} \\cdot \\sin(\\vec{k} \\cdot \\vec{r} + \\omega \\cdot t + \\phi)`
```
reinaual: same as above

src/python/espressomd/constraints.py
```diff
@@ -520,9 +530,10 @@ class FlowField(_Interpolated):
     Viscous coupling to a flow field that is
     interpolated from tabulated data like
 
-    :math:`F = -\\gamma \\cdot \\left( u(r) - v \\right)`
+    :math:`\\vec{F} = -\\gamma \\cdot \\left( \\vec{u}(\\vec{r}) - \\vec{v} \\right)`
```
reinaual: If we decide to differentiate between $r$ and $x$, this should have been an $x$ in the first place.

src/python/espressomd/constraints.py
jngrad: I was not entirely sure about the notation. Quantum mechanics uses $|x\rangle$ and $\hat{x}$ ([position operator](https://en.wikipedia.org/wiki/Position_operator)), while classical mechanics seems to accept both $\vec{x}(t)$ and $\vec{r}(t)$ ([Newton equations](https://en.wikipedia.org/wiki/Classical_mechanics#Velocity_and_speed) and [geometry](https://en.wikipedia.org/wiki/Position_(geometry))). What confused me is how inconsistent the user guide is about notations. For example in chapter [Integrators and thermostats](https://espressomd.github.io/doc/integration.html), we use $\vec{x}(t)$ in VV, NpT and $\vec{r}(t)$ in steepest descent, Brownian. There is often no indication that $x$ is a vector (no arrow, no boldface), so I'm not even sure if the author meant the position vector or the x-component of the position vector.

reinaual (approved):

Issue #4762, opened 2023-07-25
More fine-grained event handling


RudolfWeeber:

jngrad:
We currently rely on pre-change resp. post-change hooks to let components of the system react to a proposed change, using functions named `veto_*_change()` resp. `on_*_change()`. The former typically throw an error to prevent a change that would leave the system in an undefined state, whereas the latter allow the component to react to a change, such as re-tuning long-range solvers parameters.

These methods sometimes take arguments, such as the proposed change to check if the value would lead to an undefined state, or a flag to help skip some checks when the corresponding pre-conditions were already validated upstream.

This mechanism is cumbersome and inflexible. In particular, each event needs a dedicated method, sometimes two if both pre- and post-change hooks are needed. These methods need to appear in all derived class, with the correct flags (which can be optional arguments).

To improve separation of concerns and the open/close principle, we could rewrite these hooks using a message-passing interface. Below is a prototype for this new design. This change is made possible by #4816, which encapsulated all event hooks into the core `System` class.

```c++
#include <any>
#include <functional>
#include <optional>
#include <stdexcept>

struct OnChange {
  enum class Property : int {
    none = 0,
    box_l,
    time_step,
  };
  enum class When : int {
    none = 0,
    before,
    after,
  };
  OnChange(Property property_, When when_, std::any &&value_) :
    property{property_}, when{when_}, value{value_}, callback(std::nullopt) {}
  Property property;
  When when;
  std::any value;
  std::optional<std::function<void()>> callback;
};

class Solver {
public:
  void on_change(OnChange const &message) {
    if (message.when == OnChange::When::before) {
      if (message.property == OnChange::Property::time_step) {
        auto const time_step = std::any_cast<double>(message.value);
        if (time_step /*...*/) {
          throw std::runtime_error("time step too small for feature Solver");
        }
      }
    }
  }
};

class System {
public:
  double box_l;
  double time_step;
  Solver *solver;
  void set_time_step(double value) {
    if (solver) {
      auto message = OnChange(OnChange::Property::time_step, OnChange::When::before, value);
      solver->on_change(message);
      this->time_step = value;
      message.when = OnChange::When::after;
      solver->on_change(message);
    }
  }
};

int main() {
  Solver solver{};
  System system{10., 0.01, &solver};
  system.set_time_step(0.05);
}
```

PR #4901, opened 2024-04-08
Disable box length reset when particles are present

### LINKED ISSUE #4834: Box length change invalidates particle positions
When manually modifying the box length, the new particle positions are recalculated with the wrong box size.

MWE:
```python
import espressomd
import numpy as np
np.set_printoptions(precision=2)
system = espressomd.System(box_l=[15., 15., 15.])
system.time_step = 0.01
system.cell_system.skin = 0.4
p = system.part.add(pos=[-0.1, 15.1, 30.])
expected = p.pos_folded * 10./15.
print(f"before:   {p.pos_folded}")
system.box_l = [10., 10., 10.]
print(f"after:    {p.pos_folded}")
print(f"expected: {expected}")
```
Output:
```
before:   [14.90 0.10 0. ]
after:    [ 4.90 0.10 0. ]
expected: [ 9.93 0.07 0. ]
```

This bug is reproducible on the python branch and with ESPResSo releases 4.2.1 and 4.1.4.

It is unclear to me, how to gracefully handle a box resize. One could fold the particle position onto the unit cell, apply the box resize via rescaling, and re-use the older image box counter. But then, a jump may become visible in trajectories in unfolded coordinates. One would also need to maintain the old and new value in the `BoxGeometry` object to apply the folding and then unfolding with the two box sizes. It is also unclear what should happen with Lees-Edwards boundary conditions, where particle positions gain an offset when going through one of the three periodic boundaries.

The easiest course of action is to prevent manually changing the box length if the box contains particles, and rely instead on the `system.change_volume_and_rescale_particles()`, which implements the correct behavior. Alternatively, one could make the box length setter call the rescale method under-the-hood, but this introduces a new problem: the `on_boxl_change` event would be called three times, causing long-range solvers to re-tune all solver parameters three times with a different box size. Since  the box might no longer be cubic for the first two re-tunes, the tuning code may become extremely slow, have to sample more mesh sizes, and throw an exception if the accuracy cannot be reached.

Note: this bug doesn't affect the NpT integrator, which uses the rescale algorithm.

jngrad:
Fixes #4834

Description of changes:
- Resizing the box via `system.box_l = new_box_l` now throws a runtime error if there are particles present, because particle position folding cannot be guaranteed to be correct; use `system.change_volume_and_rescale_particles()` instead, which properly rescales particle positions.

### DIFF
--- a/src/script_interface/system/System.cpp
+++ b/src/script_interface/system/System.cpp
@@ -111,6 +111,7 @@ System::System() : m_instance{}, m_leaves{std::make_shared<Leaves>()} {
            if (not(new_value > Utils::Vector3d::broadcast(0.))) {
              throw std::domain_error("Attribute 'box_l' must be > 0");
            }
+           m_instance->veto_boxl_change();
            m_instance->box_geo->set_length(new_value);
            m_instance->on_boxl_change();
          });
@@ -256,12 +257,15 @@ Variant System::do_call_method(std::string const &name,
     auto &box_geo = *m_instance->box_geo;
     auto const coord = get_value<int>(parameters, "coord");
     auto const length = get_value<double>(parameters, "length");
+    assert(coord != 3 or ((box_geo.length()[0] == box_geo.length()[1]) and
+                          (box_geo.length()[1] == box_geo.length()[2])));
     auto const scale = (coord == 3) ? length * box_geo.length_inv()[0]
                                     : length * box_geo.length_inv()[coord];
     context()->parallel_try_catch([&]() {
       if (length <= 0.) {
         throw std::domain_error("Parameter 'd_new' must be > 0");
       }
+      m_instance->veto_boxl_change(true);
     });
     auto new_value = Utils::Vector3d{};
     if (coord == 3) {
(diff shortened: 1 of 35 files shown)


src/script_interface/system/System.cpp
```diff
@@ -111,6 +111,7 @@ System::System() : m_instance{}, m_leaves{std::make_shared<Leaves>()} {
            if (not(new_value > Utils::Vector3d::broadcast(0.))) {
              throw std::domain_error("Attribute 'box_l' must be > 0");
            }
+           m_instance->veto_boxl_change();
            m_instance->box_geo->set_length(new_value);
            m_instance->on_boxl_change();
```
jngrad, line 116: Note to reviewers: there are long-term plans to replace this clumsy `veto/on_change` mechanism by a proper message-passing interface (#4762).

reinaual (approved):
I think we should add this change as a major one in the release-notes, to make sure it is as visible as possible to the users. This has a lot of potential to break old scripts, especially custom checkpointing solutions

PR #4924, opened 2024-05-21
Update Python requirements

jngrad:
Description of changes:
- replace `np.product()` (deprecated in NumPy 1.15 and removed in NumPy 2.0) by `np.prod()`
- bump Python version requirement of `requests` due to CVE-2024-35195
- adjust Python version requirements to resolve all dependency issues
   - `python3 -m pip install -r requirements.txt` now works again

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ cython>=0.29.21,<3.0.10
 setuptools>=68.1.2
 packaging>=24.0
 # required scientific packages
-numpy>=1.26.4
+numpy>=1.26.4,<2.0
 h5py>=3.10.0
 # optional scientific packages
 scipy>=1.11.4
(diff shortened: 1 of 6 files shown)


requirements.txt
```diff
@@ -18,16 +18,16 @@ pygame>=2.1.2
 # waLBerla dependencies
 pystencils==1.2
 lbmpy==1.2
-sympy==1.12
+sympy==1.9
```
reinaual, line 21: Why do we have to downgrade `sympy`?

jngrad:
```
ERROR: Cannot install -r requirements.txt (line 19) and sympy==1.12 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested sympy==1.12
    pystencils 1.2 depends on sympy<=1.11.1 and >=1.6

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
```
I've used sympy version 1.9 to generate the CUDA kernels, so I know it's compatible. At some point I'll have to look into updating the waLBerla commit revision to use lbmpy 1.3.4 and a more recent version of sympy, but this will take a while. I already tried once when preparing the CUDA kernels PR, but the `Context` API completely changed and I did not have the patience to fix everything.

reinaual (approved):

Issue #4847, opened 2024-01-09
WIP: P3M improvements notes


RudolfWeeber:
In preparation for performance portable electrostatics.

## ESPResSo interaciton

* Theere is still some immedaite dependent on ESPResSo data structures in the P3M code, e.g., 
    * the System class
    * BoxGeometry and LocalGeometry
This is in the initialization and for event triggers such as on_boxl_change()

* Some function signatures still rely on Espresso types, e.g. ParticleRange for assign_charges.

## Data structures general

* some quantities are scale dby the box length, but of type `double`. This is ambigious for non-cubic boxes. E.g. `alpha_L`

## Patterns wich micht be replaced by an algorithm 

* There are a lot of places, where something is done on all mesh siztes, so in essence this is a for_each with an enumerate, just that we have 3d indices rather than a linear one. This appears several times in the k-space kernels for electrostatics and dipolar p3m. 

## Abstraction of of meshes and fft

* mesh accesses for real and k space might be abstracted away, particular since the storage order is changed by the fft.
*  Fht fft details need to be abstracted away.
* Iteration boundaries depending on details of the fft such as `p3m.fft.plan[3].new_mesh[2]` need to be abstracted away

PR #4931, opened 2024-06-06
Modernize core

### LINKED ISSUE #4846: Increase required C++ version to C++20 and CUDA version to 12
**New features in C++20**

* [designated initializers](https://en.cppreference.com/w/cpp/language/aggregate_initialization#Designated_initializers) and [3-way comparison](https://en.cppreference.com/w/cpp/language/default_comparisons) (spaceship operator `<=>`)
  ```c++
  #include <compare>
  
  struct S {
    int i;
    int j;
    float f;
    constexpr auto operator <=>(const S&) const = default;
  };
  
  void foo() {
    constexpr S s1{.i = 2, .j = 3, .f = 3.14f};
    constexpr S s2{.i = 2, .j = 3, .f = 3.14f * 2.f};
    static_assert(!(s1 == s2));
    static_assert( (s1 != s2));
    static_assert( (s1 <  s2));
    static_assert( (s1 <= s2));
    static_assert(!(s1 >  s2));
    static_assert(!(s1 >= s2));
  }
  ```
* [constraints and concepts](https://en.cppreference.com/w/cpp/language/constraints)
  ```c++
  #include <concepts>
  
  auto add(std::integral auto lhs, std::integral auto rhs) {
    return lhs + rhs;
  }
  ```
* [mathematical constants](https://en.cppreference.com/w/cpp/numeric/constants)
* [string formatter](https://en.cppreference.com/w/cpp/utility/format/formatter) (Python syntax)
  ```c++
  #include <iostream>
  #include <format>
  #include <numbers>
  
  int main() {
    std::cout << "Python formatting syntax: ";
    std::cout << std::format("π≈{:.4f}", std::numbers::pi_v<double>) << "\n";
  }
  ```
  Output:
  ```
  Python formatting syntax:π≈3.1416
  ```
* [ranges](https://en.cppreference.com/w/cpp/ranges)
  ```c++
  #include <ranges>
  #include <vector>
  #include <algorithm>
  #include <iostream>
  
  int main() {
    std::cout << "Print the square of even numbers from a sorted list:\n";
    std::vector<int> values{1, 4, 5, 9, 2};
    std::ranges::sort(values);
    for (auto const val : values | std::views::filter([](int i) {return i%2 == 0;})
                                 | std::views::transform([](int i) {return i*i;})) {
      std::cout << val << "\n";
    }
  }
  ```
  Output:
  ```
  Print the square of even numbers from a sorted list:
  4
  16
  ```
* [`std::span`](https://en.cppreference.com/w/cpp/container/span) to abstract away `std::vector` and `std::array` in function arguments
* constexpr `std::vector` and `std::string` in GCC 12+ and Clang15+ (only inside a constexpr function)
* constexpr [algorithms](https://en.cppreference.com/w/cpp/algorithm) (sorting, finding, etc.)
* [`std::has_single_bit()`](https://en.cppreference.com/w/cpp/numeric/has_single_bit)
* [`std::unordered_map::contains()`](https://en.cppreference.com/w/cpp/container/unordered_map/contains)

Not fully supported by all compilers yet:

* [coroutines](https://en.cppreference.com/w/cpp/language/coroutines)
* [modules](https://en.cppreference.com/w/cpp/language/modules)
* optional typenames
* consteval functions

See [compiler support tables](https://en.cppreference.com/w/cpp/compiler_support#cpp20) for more details.

**Applicability**

We can replace all calls to `std::to_string` by corresponding calls to `std::format`. The standard function `std::to_string` is known to be ill-suited to represent floating-point numbers outside the range [1e6, 1e-6] due to the decimal format it uses, among other issues outlined in draft proposal [[D2587R1]](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2587r1.html). In fact, [`std::to_string`](https://en.cppreference.com/w/cpp/string/basic_string/to_string) will be redefined around`std::format` in C++26, but without addressing the underlying issue of precision loss in small numbers.  `std::to_string` is used 70 times in ESPResSo due to its convenience over alternatives based on `std::stringstream` or `snprintf`, approximately 20 of which involve a floating-point value. Likewise, several calls to `std::stringstream` are used to format error messages and could be replaced by a corresponding `std::format` call.

A lot of C++20 features were backported to ESPResSo in the `Utils` namespace about 6 years ago, which require extensive testing and in a few rare cases make it difficult to interface ESPResSo to third-party C++ libraries. We chose not to use equivalent features in Boost because some of these features were not just backported to C++14, but also ported to CUDA 9. These features are now available in the standard library and in CUDA 12. In particular, `std::span` can replace `Utils::Span`, mathematical constants can replace most of the values defined in `utils/constants.hpp`,  `std::unordered_map::contains()` can replace a lot of boilerplate code involving iterators or element counting, and `std::has_single_bit()` can replace hard-to-read bit operations to detect whether `thermo_switch` and similar bitfields have only one bit set (`std::bitset` is not used for performance reasons).

Concepts can replace trivial template declarations to help generate more helpful compiler error messages. Designated initializers can be used to help disambiguate constructor calls involving multiple arguments sharing the same type.

**Prior work**

ESPResSo is C++20-ready since b1f59e0da8f27d27ca00b5acd517e2ce961cdd2b and is tested in CI for builds without CUDA.

**Requirements**

We probably need to drop support for CUDA 11, which doesn't support C++20, or find a way to compile CUDA code in C++17 mode and C++ code in C++20 mode via CMake options. We already require a minimum of CUDA 11.3 when the compiler is Clang because recent GPUs with architectures sm_70+ require Thrust 1.11. For GCC the situation is a bit less clear, we currently require CUDA 11.0 but only test 11.5 in CI.

Bumping CUDA requirements to 12.0 would be the easiest solution. Ubuntu 24.04 ships CUDA 12.0 via [`nvidia-cuda-toolkit`](https://packages.ubuntu.com/noble/nvidia-cuda-toolkit). Many compute clusters have migrated to CUDA 12.0 in May of 2023 to mitigate several CVEs ([full list](https://nvidia.custhelp.com/app/answers/detail/a_id/5456/~/security-bulletin%3A-nvidia-cuda-toolkit---april-2023)), namely: bwUniCluster 2.0, bwForCluster JUSTUS 2, bwForCluster Helix, HLRS Vulcan, HPC Vega, Jülich JUWELS.

We would have to drop support for older compilers and require GCC 10+ and Clang 14+ (or Clang 17+ to build CUDA code without nvcc). I double-checked with our EasyBuild partners and increasing the compiler version requirements wouldn't be an issue for them. Several packages in `gompi/2023a` have already migrated to CUDA 12.1.

**Course of action**

* Update `CMAKE_CXX_STANDARD` to 20
* Drop support for CUDA 11, which doesn't support C++20.
* Figure out which GCC and Clang toolchains are compatible with CUDA 12.

Timeline: probably best to defer this change until we migrate to the Ubuntu 24.04 migration at the home institute. This way we don't need to sort out compiler toolchains twice.

jngrad:
Fixes #4846
Partial fix for #4847

Description of changes:
* migrate to C++20 in both C++ and CUDA code
   * CUDA 12+ is required
   * a recent C++ compiler is required, recommended: GCC 11+, Clang 18+, AppleClang 15+
* use STL containers
   * use `std::optional` instead of `boost::optional`
   * use `std::numbers` instead of `Utils::numbers`
   * use `std::span` instead of `Utils::Span`
   * use `std::ranges` and ranged-based algorithms instead of `boost::ranges` and `boost::algorithm`
   * cannot use `std::array` instead of `Utils::Array`, because its `constexpr` methods don't have the `__device__` qualifier ("calling a constexpr `__host__` function from a `__host__ __device__` function is not allowed")
* reduce dependency on Boost
   * remaining: `boost::mpi`, `boost::serialization`, `boost::unit_test`, `boost::algorithm` (for C++26 features)
* migrate to lbmpy/pystencils 1.3.3 and to the most recent version of the waLBerla development branch
* rewrite CMake by following best practices
   * the `object-in-fluid` module is now also installed when the `espressomd` module is installed
   * C++ and CUDA source files are now treated separately and no longer share the same compiler flags
   * CUDA shared objects are now properly linked against instrumentation tools (e.g. Caliper, UBSAN)
   * CUDA objects now generate code coverage information for host code
   * Boost unit tests can now be written in a CUDA source file
* rewrite P3M nested loops that operate on 3D grids
   * a utility function wraps the nested loops and takes a kernel as argument that operates on the grid
   * intermediate values are now systematically cached
   * the analytic cotangent sum kernel was inlined


jngrad:
@reinaual This PR should be considered a top priority. It is needed to complete the hand-over to Forschungszentrum Jülich ("MultiXscale deliverable" e-mail from May 10). It is also a pre-requisite for major refactoring campaigns, such as moving the waLBerla bridge to a separate repository (blocked by its dependency on `Utils::` functions and classes) and moving the Monte Carlo code to Python (blocked because it touches the same C++ code as this PR).

@rodrigoacb @phohenberger The P3M code now leverages C++20 features: `std::span`, `std::ranges`, `concepts`. In addition, to facilitate porting to Cabana and Kokkos, all nested loops of the form:

```c++
auto accumulate = 0.;
for (int nx = -mesh[0] / 2; nx < mesh[0] / 2; nx++) {
  auto const ctan_x = cotangent_sum(nx, mesh_i[0]);
  for (int ny = -mesh[1] / 2; ny < mesh[1] / 2; ny++) {
    auto const ctan_y = ctan_x * cotangent_sum(ny, mesh_i[1]);
    for (int nz = -mesh[2] / 2; nz < mesh[2] / 2; nz++) {
      if ((nx != 0) or (ny != 0) or (nz != 0)) {
        auto const n2 = nx * nx + ny * ny + nz * nz;
        auto const cs = ctan_y * cotangent_sum(nz, mesh_i[2]);
        auto const [alias1, alias2] =
            p3m_tune_aliasing_sums(nx, ny, nz, mesh, mesh_i, cao, alpha_L_i);
        auto const d = alias1 - Utils::sqr(alias2 / cs) / n2;
        if (d > 0.) {
          accumulate += d;
        }
      }
    }
  }
}
```

were rewritten as:

```c++
auto const m_stop = mesh / 2;
auto const m_start = -m_stop;
auto indices = Utils::Vector3i{};
auto values = Utils::Vector3d{};
auto accumulate = 0.;
for_each_3d(m_start, m_stop, indices,
  [&]() {
    if ((indices[0] != 0) or (indices[1] != 0) or (indices[2] != 0)) {
      auto const n2 = indices.norm2();
      auto const cs = Utils::product(values);
      auto const [alias1, alias2] =
          p3m_tune_aliasing_sums(indices, mesh, mesh_i, cao, alpha_L_i);
      auto const d = alias1 - Utils::sqr(alias2 / cs) / n2;
      if (d > 0.) {
        accumulate += d;
      }
  },
  [&](int dim, int n) {
    values[dim] = cotangent_sum(n, mesh_i[dim]);
  });
```

Some kernels rely on random access iterators (`std::size_t index` or `std::vector<double>::iterator it_energy`) that are incremented at every step of the innermost loop. This side-effect can be an obstacle to execution policies that rely on out-of-order execution or parallel execution, but the good news is, they can always be replaced by a `std::size_t pos` calculated on-the-fly using the loop counters stored in `indices`. The flattened arrays have row-major order. This can be a bit tricky when the loop counters don't start at 0; when in doubt, do `git show 7a91cfa` and search for the string `Utils::get_linear_index` until you find the hot loop of interest.

The execution policy for `for_each_3d()` may differ depending on the situation. Some of them (like in the example above) have a `Projector` kernel that populates an array of intermediate values that are expensive to compute and thus cannot be executed fully out-of-order without incurring a small performance penalty (calling the projector $3N^3$ times instead of $N + N^2 + N^3$ times). When no `Projector` kernel is provided, it defaults to a `constexpr` no-op function that generates no assembly code. See [for_each_3d.hpp](https://github.com/espressomd/espresso/blob/7a91cfa355e25094f62220141f0d6ef66281858a/src/core/p3m/for_each_3d.hpp) for more details. The `RX/RY/RZ` and `KX/KY/KZ` constants are used to transpose data; kernels that use them might not fully benefit from SIMD.

Not all nested loops contribute equally to the runtime. A Caliper report for the P3M benchmark is reproduced below. The nested loops in the top-level integrator are quite inexpensive compared to the reciprocal space (all-to-all FFT) and real space (N-square short-range loop) force calculations, although we have concrete plans to improve the performance of these two components. There is an expensive nested loop in the accuracy calculation kernel of the tuning algorithm, due to the $\mathcal{O}(N^6)$ complexity of evaluating a double sum over the 3D grid (it doesn't appear as a double sum in the Caliper report, because I disabled the nested instrumentation in that kernel to reduce overhead).

```
$ CALI_CONFIG=runtime-report mpiexec -n 4 ./pypresso ../maintainer/benchmarks/p3m.py \
    --volume_fraction=0.25 --particles_per_core=2000
Path                         Min time/rank Max time/rank Avg time/rank Time %
integrate                         3.783249      3.835922      3.808505  4.946649
  calculate_forces                2.312934      7.806445      5.088166  6.608728
    init_forces                   0.132315      0.138147      0.135768  0.176341
    calc_long_range_forces       32.141172     32.232919     32.174786 41.789985
      for_each_3d                 0.512280      0.530820      0.519396  0.674614
    short_range_loop             20.797703     26.168697     23.444353 30.450525
tune                              3.030472      3.126394      3.075959  3.995187
  calculate_accuracy              0.000367      0.000440      0.000406  0.000527
    for_each_3d                   2.264349      2.353698      2.311547  3.002335
  recalc P3M params               0.002077      0.002345      0.002197  0.002853
    for_each_3d                   0.056157      0.059715      0.057934  0.075247
      for_each_3d                 0.070111      0.073837      0.071683  0.093105
  integrate                       0.305274      0.306183      0.305689  0.397042
    calculate_forces              0.107690      1.059676      0.547243  0.710782
      init_forces                 0.005260      0.005526      0.005395  0.007007
      calc_long_range_forces      1.348340      1.419409      1.369864  1.779237
        for_each_3d               0.017730      0.018481      0.018187  0.023622
      short_range_loop            1.880777      2.821218      2.371574  3.080302
recalc P3M params                 0.001075      0.001241      0.001148  0.001491
  for_each_3d                     0.656495      0.701529      0.682490  0.886447
    for_each_3d                   0.817611      0.863255      0.842211  1.093899
```

src/core/electrostatics/p3m_gpu_cuda.cu
```diff
@@ -521,18 +525,16 @@ void assign_forces(P3MGpuData const &params,
...
-  auto const data_length =
-      3u *
-      static_cast<std::size_t>(parts_per_block *
-                               static_cast<unsigned int>(cao)) *
-      sizeof(float);
-  switch (cao) {
+  auto const data_length = std::size_t(3u) *
+                           static_cast<std::size_t>(parts_per_block) *
+                           static_cast<std::size_t>(cao) * sizeof(float);
+  switch (params.cao) {
```
reinaual, line 520: Out of interest, is there a reason not to use the local variable anymore?

jngrad:
General MultiXscale Meeting: Clang 18.1 has full support for [HIP, OpenCL, AMD GPUs, Intel GPUs](https://releases.llvm.org/18.1.0/tools/clang/docs/HIPSupport.html) and near-complete support for [OpenMP (including GPU kernel calls)](https://releases.llvm.org/18.1.0/tools/clang/docs/OpenMPSupport.html). This makes Clang a compiler with strategic importance, in particular Clang 19.0 because it implements an extra OpenMP feature that is needed to build the testsuite of another software in EasyBuild. This PR introduces support for Clang 18.1 in ESPResSo.

src/core/p3m/influence_function.hpp
```diff
@@ -52,44 +53,43 @@
...
+  auto km = Utils::Vector3d{};
+  auto fnm = Utils::Vector3d{};
+  auto numerator = 0.;
+  auto denominator = 0.;
+
+  for_each_3d(
+      m_start, m_stop, indices,
+      [&]() {
+        using namespace detail::FFT_indexing;
+        auto const U2 = std::pow(Utils::product(fnm), 2 * cao);
```
reinaual, line 78: In all the influence functions we could save one multiplication per iteration by caching the products in `fnm`, such that `fnm[0] = std::pow(math::sinc(...), 2 * cao);` and `fnm[1] = fnm[0] + std::pow(...)`, ... I'm not sure if it's worth the effort, since we probably don't do this often enough in a simulation

reinaual (approved):

src/core/electrostatics/p3m_gpu_cuda.cu
jngrad: To avoid implicit conversion compiler diagnostics. At line 503, the local `cao` is no longer a signed int: `auto const cao = static_cast<unsigned int>(params.cao);`

src/core/p3m/influence_function.hpp
jngrad: Could be. This specific diff already improves performance by caching the intermediate `sinc()` calls. But the P3M benchmarks reveal only 2% performance improvement for this entire PR. I gave up optimizing these 3D kernels, because saving a float multiplication is probably not worth it. It's also unclear, whether we will still be able to cache the `fnm` variable once we replace it with Kokkos parallel execution (OpenMP, potentially out-of-order).

PR #4944, opened 2024-06-27
Bugfixes

### LINKED ISSUE #4930: Velocity Verlet step in documentation has a typo
In the documentation (https://espressomd.github.io/doc/integration.html#velocity-verlet-algorithm) the force used for the first half step is $F(x(t), v(t-\Delta t/2), t)$ but in the code it uses the force at time $t$ instead:

https://github.com/espressomd/espresso/blob/abb087c7027b926687efb53a5ac9630d53488923/src/core/integrators/velocity_verlet_inline.hpp#L34-L45

Used in https://github.com/espressomd/espresso/blob/abb087c7027b926687efb53a5ac9630d53488923/src/core/integrate.cpp#L436

(Note the `calculate_forces` before it calls `integrator_step_1`

### LINKED ISSUE #4934: LB: velocity updated missing in population and force setters
Setting populatoins and/or last_applied force changes the velocity of an LB site.

Our implementation caches the velocities in a velocity field, which is calculated during the LB update.
An LB site's velocity field entry ahs to be updatd, when populaoitns or last_applied_force on that site are changed. This is currently missing. Among others, this results in an all-zero velocity field in the firt time step after restoring an LB from a checkpoint.

### LINKED ISSUE #4935: LB/Virtual sites: incomplete API change
With the propagatoin refactor, the coupling fo virtual sites to an LB fluid is now controlled on the particle level by setting the appropriate propagation flag or when constructing the virtual site with p.vs_relate_to().

So 
```
system.thermostat.set_lb(..., act_on_virtual=True)
```
has no effect. However, this keyword argument is still silently accepted.

To resolve:

* remove the `act_on_virtual` keywod argument from thermostat.set_lb()
* adapt the documentatoin w.r.t the propagaoitn flags

### LINKED ISSUE #4936: Lost virtual particle propagation flag after checkpointing
When creating a virtual particle using vs_relative_to() coupled to a lb fluid by setting couple_to_lb = True adding the TRANS_LB_MOMENTUM_EXCHANGE flag manually, this specific flag is lost after checkpointing. The other flags somehow are still there.
[minimum_example.txt](https://github.com/user-attachments/files/15790028/minimum_example.txt)

### LINKED ISSUE #4940: Particle image box is wrong with Verlet lists
When Verlet lists are used, and the Verlet skin is too large, particles are no longer resorted and their image box is miscounted.

MWE:
```python
import espressomd
import numpy as np

system = espressomd.System(box_l=[10.0, 10.0, 10.0])
system.time_step = 0.01
system.cell_system.skin = 0.4 # bug with 0.4, ok with 0.04 and 0.004
system.periodicity = [True, True, True]

x0 = np.array([9.95, 0.0, 0.0])
v0 = np.array([0.1, 0.0, 0.0])
p = system.part.add(pos=x0, v=v0)

pos_unfolded_ref = []
pos_unfolded = []
pos_folded = []
image_box = []

for i in range(10):
    pos_unfolded_ref.append(x0 + system.time * v0)
    pos_unfolded.append(p.pos)
    pos_folded.append(p.pos_folded)
    image_box.append(p.image_box)
    system.integrator.run(10)

pos_unfolded_ref = np.array(pos_unfolded_ref)
pos_unfolded = np.array(pos_unfolded)
pos_folded = np.array(pos_folded)
image_box = np.array(image_box)

plt.plot(pos_unfolded_ref[:,0], image_box[:,0], "o-")
plt.show()
```

Three skin values can be used in this MWE:
* `skin=0.4`: no particle resort happens besides the first time step (local resort)
* `skin=0.04`: particle resort (local resort) happens in one cell
* `skin=0.004`: particle resort (local resort) happens in multiple cells

Depending on the choice of initial conditions, e.g. with `x0=[9.50, 0.0, 0.0]` and `skin=0.4`, the image box is updated, but with a random time lag, i.e. it gets incremented when the position is anywhere between `box_l` and `box_l + skin`. The only particle folding test we have doesn't use Verlet lists.

This bug is reproducible on the Python branch, and the 4.2.2 and 4.1.4 releases. Other releases were not checked.

Many thanks to Xin Yuan for reporting this issue with a MWE on the mailing list thread [Position folding issue](https://lists.nongnu.org/archive/html/espressomd-users/2024-06/msg00000.html).

### LINKED ISSUE #4943: Ccache partial support for `-Xcompiler`
**TL;DR:** Ccache doesn't parse most flags passed to `-Xcompiler`. Code coverage files aren't regenerated without a workaround.

## Background

ESPResSo recently introduced support for code coverage of CUDA source files (e11f6fdb67e2d). When compiling a file with name `feature.cu`, an extra `feature.gcno` profile notes file is meant to be generated alongside the target `feature.cu.o` object file. When executing the code, a `feature.gcda` count data file is automatically written to disk. The execution counts are mapped to the corresponding source code line via the `feature.gcno` file. Thus, both the gcno and gcda files are required to generate the code coverage report.

Ccache is a compiler cache: it parses the compiler command line options and stores the output of the compiler in a database. When the same compiler invocation is detected, and the input source file(s) didn't change, the stashed output files are fetched from the database and the compilation is avoided. Ccache detects derived output files like gcno files in a best-effort manner. This is achieved by function [`process_option_arg()`](https://github.com/ccache/ccache/blob/v4.10/src/ccache/argprocessing.cpp#L338-L344), which relies on conditionals to handle flags like `"-fprofile-arcs"` and loops to handle tokens split from comma-separated lists via `util::Tokenizer()`.

## Problem statement

When compiling CUDA code with code coverage information using nvcc, we need to forward the coverage flags to the host compiler via the syntax `-Xcompiler=-Og -Xcompiler=--coverage -Xcompiler=-fprofile-abs-path`, or via the comma-separated syntax `-Xcompiler=-Og,--coverage,-fprofile-abs-path` (CUDA user guide: [4.2.4. Options for Passing Specific Phase Options](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/#options-for-passing-specific-phase-options)).

Ccache currently only support a small subset of nvcc compiler flags (https://github.com/ccache/ccache/issues/1325), and coverage flags aren't part of them. Adding support for host compiler coverage flags isn't straightforward. In the two syntaxes outlined above, the former is relatively easy to implement, but the latter requires parsing a list of tokens to extract the relevant information. The Ccache parser doesn't use recursion, and thus cannot properly handle a comma-separated list of flags. While it has a tokenizer, it is used to detect a single token from the list. Recursion is actually required here, because the parser is meant to return an error code and a borrowed copy of the compiler flag for each flag in the command line. Using a for loop would prevent us from generating actionable diagnostics when a token list contains a mix of supported and unsupported flags.

Introducing recursion would probably require altering the data structure used to collect flags, considering the data structure currently requires the entire flag to be added to the structure, and we cannot simply add a token extracted from the token list without breaking the parser, as I found out while attempting to implement a non-recursive solution. The diff is reproduced below for completeness. With it, the gcno file is cached with compiler flag `-Xcompiler=-Og,--coverage,-fprofile-abs-path`, but isn't cached with compiler flag `-Xcompiler=--coverage,-fprofile-abs-path`.

<details><summary>non-working patch (click to unroll)</summary>

```diff
diff --git a/src/ccache/argprocessing.cpp b/src/ccache/argprocessing.cpp
index ae74e522..275b8e64 100644
--- a/src/ccache/argprocessing.cpp
+++ b/src/ccache/argprocessing.cpp
@@ -851,6 +851,25 @@ process_option_arg(const Context& ctx,
     return Statistic::none;
   }
 
+  if (util::starts_with(arg, "-Xcompiler=")) {
+    for (const auto part : util::Tokenizer(&arg[11], ",")) {
+      if (part == "-fprofile-arcs") {
+        args_info.profile_arcs = true;
+        state.common_args.push_back(args[i]);
+      }
+      if (part == "-ftest-coverage") {
+        args_info.generating_coverage = true;
+        state.common_args.push_back(args[i]);
+      }
+      if (part == "--coverage" || part == "-coverage") {
+        args_info.profile_arcs = true;
+        args_info.generating_coverage = true;
+        state.common_args.push_back(args[i]);
+      }
+      return Statistic::none;
+    }
+  }
+
   if (arg == "-fstack-usage") {
     args_info.generating_stackusage = true;
     state.common_args.push_back(args[i]);
```

</details>

While one could adapt the parser to detect `-Xcompiler=--coverage` and its alternative forms (`-Xcompiler --coverage`, `--compiler-options --coverage`,  `--compiler-options=--coverage`) as a best effort and skip the comma-separated versions, that would go against the expectations of most CUDA developers and could lead to subtle build reproducibility issues.

## Reproducing the bug

Minimal working example:

```cuda
#include <cuda.h>
#include <stdio.h>

__global__ void cuda_hello() {
  printf("Hello World!\n");
}

int main() {
  cuda_hello<<<1,1>>>();
  cudaDeviceSynchronize();
  return 0;
}
```

**Current behavior:** using ccache either built from sources using commit https://github.com/ccache/ccache/commit/f03feea871099ee5d6276435b5cc56242d9b634a or built from release 4.9.1, nvcc release 12.0 (V12.0.140, build cuda_12.0.r12.0/compiler.32267302_0) and gcc 12.3.0-17ubuntu1 as host compiler:

```
$ ls
mwe.cu
$ nvcc -Xcompiler=-Og,--coverage,-fprofile-abs-path -c mwe.cu
$ ls
mwe.cu  mwe.gcno  mwe.o
$ rm -f mwe.gcno mwe.o
$ ccache nvcc -Xcompiler=-Og,--coverage,-fprofile-abs-path -c mwe.cu
$ ls
mwe.cu  mwe.gcno  mwe.o
$ rm -f mwe.gcno mwe.o
$ ccache nvcc -Xcompiler=-Og,--coverage,-fprofile-abs-path -c mwe.cu
$ ls
mwe.cu  mwe.o
```

**Expected behavior:** the `mwe.gcno` should have been restored from the cache.

## Outlook

Adding support for `-Xcompiler` flags in Ccache will probably require a significant code contribution.

**Workaround:** trick Ccache intro parsing host compiler flags by invoking nvcc like this:

```sh
ccache nvcc --forward-unknown-to-host-compiler --coverage -fprofile-abs-path -c mwe.cu
```

jngrad:
Description of changes:
- recalculate up-to-date particle box images in the python interface and hdf5 writer
   - fixes #4940
- address propagation flags regressions in the checkpointing and virtual sites codes
   - fixes #4935
   - fixes #4936
- update LB pdf/force/velocity fields together to avoid out-of-date cached velocities
   - fixes #4934
   - add unit tests for the LB field accessors (both  C++ and CUDA)
- extract the code coverage logic to a self-contained file and forward coverage compiler flags to Ccache
   - closes #4943
   - developers can now reproduce Codecov coverage reports locally
- add support for all CMake name boolean constants in the `pypresso` launcher
   - users are no longer required to use ON/OFF when configuring boolean variables of the CMake project
- improve documentation of thermostats and propagation flags
   - fixes #4930


RudolfWeeber:
I reviewed the commits 1-3 and 6-7. They can be merged.

RudolfWeeber:
For 3 and 4 it would be helpful to know in broad terms what the actual cahgnes are. That would make it clearer to me what to look for in the diff

jngrad:
> For 3 and 4 it would be helpful to know in broad terms what the actual cahgnes are. That would make it clearer to me what to look for in the diff

* commit 3: it is easier to look at the `*.tmpl.{h,cu,cuh}` template files. The pdf/vel/force files can no longer be edited independently from one another, since updating any one of them requires updating at least one of the other two. I also implemented a few missing slice accessors.
* commit 4: it is bad practice to declare CMake functions without a prefix like `espresso_`, because we can accidentally overshadow an existing CMake function and run into nasty undocumented behavior, e.g. infinite recursion. See Craig Scott's [Do Not Redefine CMake Commands](https://crascit.com/2018/09/14/do-not-redefine-cmake-commands/) blog post for more details (he's a CMake maintainer). A function called `unit_test()` doesn't exist in CMake yet, but it's not a far-fetched idea that such a function name could be introduced in a later release of CMake, or in one of its utility modules. All other ESPResSo-specific CMake functions had been renamed in older refactoring PRs. This one was renamed only now because it was quite complex and required CMake 3.25 functionality to be rewritten in a clean way. Also, it used to take multiple source code files as input, but it was totally unclear from the code how it was supposed to link them together, especially if they were mixing CUDA and C++ files. Now it only takes 1 source file a input, from which the name is deduced. So we can no longer have a fantasy test name that is unrelated to the source file name.
* commit 5: when running `cmake .. -D ESPRESSO_BUILD_WITH_UBSAN=YES`, the build is successful but `pypresso` fails to launch with a cryptic error message that took me 30 min to figure out. The `pypresso` launcher was too fragile and didn't consider that other keywords like true/yes/y/on and other less common named constants can evaluate to true.

reinaual:
I've checked commit [CMake: Rewrite unit_test() framework](https://github.com/espressomd/espresso/pull/4944/commits/d811c8682a4acc4b331bb1fe3191106005a3cf07) and [Shell scripts hardening](https://github.com/espressomd/espresso/pull/4944/commits/0b715cc8f6cd849f3e67527ecf2b6ddff9e804b8), they are both fine.

RudolfWeeber (approved):

PR #4947, opened 2024-07-03
Hide FFT implementation details

### LINKED ISSUE #4945: P3M: further refactoring
Refactoring so far has made P3M much more readable. Many thanks @jngrad 

Looking at the code, I think, the folowing things are sitll open:

* [ ] Disentangling general FFT info from specifics of the actual implementaiton for 3D ffts
* [ ] Related: abstracting away the changed data layout after FFT
* [ ] Maybe, we could use more speaking variable names in some places, e.g. replace "sm" and "fft.plan.new_mesh" by sth clearer

## Details w.r.t FFT

We currently use a "hand-written" implementation for 3d FFt with MPI. This will be replaced by  a library
We should split code/data strutures, that are independent of the specific implementtion.
This is, to my understanding:
* The global and mpi-rank local dimensions of the charge mesh (i.e. the input to the 3d fft)
* the global and mpi-3ank local indices for the charge grid in k-space (i.e., the output of the fft)
* access functions for the k-space mesh (because some FFt implementation, including the one we have now, have a different memory layout after transform. E.g., the ESPResSo implementaiton swaps between col-major and row-major during transofmr)
* functions for forward/backward FFT

Maybe this can be put in a base class, and specific implementations in sub-classes.

jngrad:
Fixes #4945

Description of changes:
- fully encapsulate the historic FFT implementation
   - electrostatics and magnetostatics now share the same FFT operations
   - FFT operations and mesh buffers are now better documented
   - all temporary mesh buffers are now private data members
- provide a type-erased API for the FFT backend
   - the P3M algorithm no longer has visibility of the FFT library
   - new FFT libraries can be used in place of the historic FFT implementation
- bugfixes:
   - the P3M algorithm no longer leaks memory when FFT plans are discarded
   - the FFT implementation now keeps the MPI environment alive until all FFT plans have been destroyed
      - avoids a race condition, since the FFT plan destructor need a MPI communicator

### DIFF
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -95,6 +95,10 @@ if(ESPRESSO_BUILD_WITH_WALBERLA)
             $<$<BOOL:${ESPRESSO_BUILD_WITH_CUDA}>:espresso::walberla_cuda>)
 endif()
 
+if(ESPRESSO_BUILD_WITH_FFTW)
+  add_subdirectory(fft)
+endif()
+
 add_subdirectory(accumulators)
 add_subdirectory(analysis)
 add_subdirectory(bond_breakage)
--- a/src/core/electrostatics/elc.cpp
+++ b/src/core/electrostatics/elc.cpp
@@ -1128,7 +1128,7 @@ void charge_assign(elc_data const &elc, CoulombP3M &solver,
   }
   /* prepare local FFT mesh */
   for (int i = 0; i < solver.p3m.local_mesh.size; i++)
-    solver.p3m.rs_mesh[i] = 0.;
+    solver.p3m.mesh.rs_scalar[i] = 0.;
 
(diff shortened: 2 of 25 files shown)


RudolfWeeber (dismissed):

jngrad:
@RudolfWeeber I realized just now we could improve separation of concerns in the P3M base struct further by moving the dipolar flag to the FFT class. Now it's ready to be merged.

@rodrigoacb We have disentangled the FFT logic from the P3M logic. There is still a small overlap due to how P3M communication buffers need to know about the FFT, but everything is encapsulated into the `FFTBackendLegacy` class. Alternative FFT libraries can be introduced via the CMake `FetchContent` mechanism and the `FFTBackend` pure virtual base class.

RudolfWeeber (approved):

PR #4948, opened 2024-07-04
Fix and test LB Lees Edwards particle coupling

RudolfWeeber:
* When finding ghosts of a position for applying forces to LB, take Lees-Edwards offset into account
* Correct interpolated fluid velocities for Lees-Edwards velocity shift when using an interpolation position across an LE boundary
* Simplify and correct the shifting and interpolation of LB PDF ghost layers, when a Lees-Edwards offset is applied
* Move application of forces from MD to the correct place for LB with pull scheme
* More extensive testing of LB Lees-Edwards particle coupling


jngrad (approved):
LGTM

PR #4950, opened 2024-07-05
Remove more global variables

### LINKED ISSUE #4821: Encapsulate auto-update accumulators
Accumulators registered via `system.auto_update_accumulators` are currently stored in a global variable in the C++ core. To help with multi-system simulations, the auto-update accumulators container needs to become a member of the core `System` class.

Procedure:
1. get familiar with the design patterns used in the core `System` class: opaque pointers, composite pattern (chapter [System class design](https://espressomd.github.io/dox/SystemClassDesign.html) in the Doxygen documentation)
1. create an `Accumulators::AutoUpdateAccumulators` class
   1. rewrite the functions in `src/core/accumulators.hpp` as methods of class `Accumulators::AutoUpdateAccumulators`
      - rename the `auto_update()` function to an `operator()()` method
      - rename all `auto_update_*()` functions to `*()` methods
      - rename all `auto_update_*()` function calls to `auto_update.*()` function calls
   1. add an `extern Accumulators auto_update;` in `src/core/accumulators.hpp`
   1. add an `Accumulators auto_update = {};` in `src/core/accumulators.cpp`
   1. make the `auto_update_accumulators` static global a member of the `Accumulators` class
   1. make all necessary adjustments until the C++ compiler is happy
   1. the testsuite should pass
1. rewrite `src/script_interface/accumulators/AutoUpdateAccumulators.hpp` as a leaf container class
   1. the class should publicly inherit from `ObjectList<AccumulatorBase, AutoParameters<ObjectList<AccumulatorBase, System::Leaf>, System::Leaf>>`
   1. the class should manage a `std::shared_ptr<::Accumulators::AutoUpdateAccumulators> m_handle;` that gets initialized inside `on_bind_system()` via a `m_params` handle that was initialized inside `do_construct()` (use `src/script_interface/electrostatics/Container.hpp` as a guide)
   1. the testsuite should pass
4. move the `auto_update` global to a member of the core `System` class
   1. add a `std::shared_ptr<Accumulators::AutoUpdateAccumulators> auto_update;` to the core `System` class, right after `bond_breakage`
   3. forward-declare `Accumulators::AutoUpdateAccumulators` at the top of `src/core/system/System.hpp`
   4. remove the `auto_update` global variable and adapt all function calls using the `->` dereference operator
   5. in the script interface, the current system's `auto_update` member is copied into `m_handle` inside `on_bind_system()`

jngrad:
Fixes #4821
Partial fix for #2628
Partial fix for #4613

Description of changes:
- encapsulate non-bonded interactions, bonded interactions, collision detection, particle list, cluster structure analysis, object-in-fluid, immersed boundaries, auto-update accumulators, constraints, MPI-IO, MMM1D
- API changes:
   - `Mpiio` and `ClusterStructure` now take a system object as argument


src/core/nonbonded_interactions/nonbonded_interaction_data.hpp
```diff
@@ -414,15 +415,15 @@ class InteractionsNonBonded {
    * @return Reference to interaction parameters for the type pair.
    */
   auto &get_ia_param(int i, int j) {
-    return *m_nonbonded_ia_params[get_ia_param_key(i, j)];
+    return *m_nonbonded_ia_params.at(get_ia_param_key(i, j));
```
reinaual: Isn't `get_ia_param_key` already checking the validity of the key? It asserts the size, so why this redundancy?

src/core/cluster_analysis/Cluster.hpp
```diff
@@ -32,6 +35,8 @@ namespace ClusterAnalysis {
 /** @brief Represents a single cluster of particles */
 class Cluster {
 public:
+  explicit Cluster(std::weak_ptr<BoxGeometry const> const &box_geo)
```
reinaual, line 38: Why can't we use an owning pointer here?

src/python/espressomd/particle_data.py
```diff
@@ -515,10 +515,9 @@ def bonds(self):
 
         """
         bonds = []
-        for bond_view in self.call_method("get_bonds_view"):
-            bond_id = bond_view[0]
-            partner_ids = bond_view[1:]
-            bonds.append((BondedInteractions()[bond_id], *partner_ids))
+        for bond_id, *partner_ids in self.call_method("get_bonds_view"):
+            bond = self.call_method("get_bond_by_id", bond_id=bond_id)
+            bonds.append((bond, *partner_ids))
```
reinaual, line 520: I understand that you didn't change this, but why do we have such a weird interface for this? Why don't we return `bond_id, List[partner_ids]` which is also more expressive?

reinaual (commented):
I tried to check everything within my capabilities. Most of it seems fine, but partially the commits are enormous and I can't say I understood the need for every change.

@RudolfWeeber do you want to take a look too? Otherwise I guess we trust in @jngrad and the testsuite and merge it.

src/core/cluster_analysis/Cluster.hpp
jngrad: The `BoxGeometry` should be owned by the `System` class, like most other pointer members of `System`. Objects of type `Cluster` may outlive the `System` class in Python. I don't think I left any `std::shared_ptr<BoxGeometry>` anywhere in the code, except for the `System` class. While it probably wouldn't hurt to share ownership of `BoxGeometry` (it's just a collection of arrays), for other classes it's not safe for them to outlive the `System` class (we have an exception mechanism to generate a meaningful error message in that case). I gave all system-owned pointers the same treatment with regards to lifetime: it can only be extended by client code like `Cluster` for the short duration of an operation.

Sometimes I have to store a `std::shared_ptr<X>` in a class. This is for example the case of Ekin and P3M, where an owning pointer to `boost::mpi::environment` is stored to guarantee we can deallocate the FFT plans even if the `System` class has expired. FFT plans deallocation requires the boost MPI communicator, whose lifetime is tied to the MPI environment. Similar argument for the CUDA handle, needed to `deallocate` thrust vectors stored in device memory.

src/core/nonbonded_interactions/nonbonded_interaction_data.hpp
jngrad: Only in `Debug` builds. In `Release` builds, if you make a mistake in the checkpointing mechanism and reload the wrong value for `max_seen_particle_type`, you silently enter undefined behavior territory. The `at()` version sends a signal even in `Release` builds with `-DNDEBUG`.

src/python/espressomd/particle_data.py
jngrad: We want to completely hide bond ids from the Python interface. The main driving force was recurring bugs in the `ScriptInterface`. Python bond ids could diverge from core bond ids during checkpointing, or refer to a different object when users erased and created a bond with the same type and same id as the erased one. Users could also run into subtle Python-specific oddities when making assumptions about data structures, e.g. tuples vs lists in https://github.com/espressomd/espresso/issues/4391#issuecomment-967064220.

Hiding bond ids also allows us to change the internal data structure from a `std::unordered_map` to a container with a more favorable time complexity for element search, which could be relevant in polymer melt simulations where there are a lot of bonds. Right now we are using a hash function, which scales as $\mathcal{O}(1)$ on average, but $\mathcal{O}(N)$ in worst case. We could imagine using a `std::vector`, which requires a multiplication instead of a hash, or a hybrid container that would use an array for small storage and `std::unordered_map` for large storage.

Lastly, we are very close to having multi-system capability in ESPResSo. A bond id is not useful if you don't know in which system this id is registered.

src/core/nonbonded_interactions/nonbonded_interaction_data.hpp
jngrad: On second thought, this is actually redundant, because the default-constructed `shared_ptr` will be a null pointer, and not a default-constructed non-bonded interaction. Thus, any mistake will be immediately visible in a `Release` build. Also, I just found out this check can't be disabled with compiler flag `-U_GLIBCXX_ASSERTIONS`.

src/python/espressomd/particle_data.py
reinaual: What I meant was only that it looked like the return type of `self.call_method("get_bonds_view")` is `bond_id, partner_id_0, partner_id_1,...`, so the size of this changes if there are several partners involved. To me this looked confusing, why not all the partner_ids are returned in a vector

reinaual (approved):

src/python/espressomd/particle_data.py
jngrad: Probably a relic from the Tcl days. This is part of the API, so we cannot really change it at the moment. At some point we will store bonds on all particles, rather than the first one; that project will give us an opportunity to explore more intuitive APIs.

PR #4958, opened 2024-07-17
LE: add fully connected option to RegularDecomposition

RudolfWeeber:
Description of changes:
- fully connected means that all cells of the decomposition at the (shear) normal boundary are neighbors with all cells across that boundary in the shear direction. 
- in this way, all pairs will be visible in the non-bonded loop, when a Lees-Edwards offset is applied across that boundary

### DIFF
--- a/src/core/cell_system/CellStructure.cpp
+++ b/src/core/cell_system/CellStructure.cpp
@@ -44,6 +44,7 @@
 #include <cstddef>
 #include <iterator>
 #include <memory>
+#include <optional>
 #include <set>
 #include <stdexcept>
 #include <string>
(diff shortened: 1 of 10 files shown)


RudolfWeeber:
@jngrad would you be weilling to hel pwith this? I ran out of ideas.
My understanding is that all failing tests fail because of a segfault which occurs when forces from the GPU are added back to the particles, because then, the cell_strure.localparticles() iterator seems to be invalid.

However, I have added trace code to the cell_strucutre.set_particle_decomposition() method whcih demonstraes that the iterator is valid after the cell system is updated.

While the PR touches the cell system, it should only update connections between cells, whih affect the short range loop, but not the cell_strucuter.local_particles() iterator.

jngrad:
@RudolfWeeber The entire testsuite runs without any segfault on my workstation using commit a4729bc269. I also don't see any obvious source of UB in the diff.

RudolfWeeber:
@jngrad What does the compiler want me to do w.r.t the modernize warning in the "empty" test on gitlab CI? Thanks

jngrad:
> @jngrad What does the compiler want me to do w.r.t the modernize warning in the "empty" test on gitlab CI? Thanks

Replace `m_fully_connected_boundary(fully_connected)` by `m_fully_connected_boundary(std::move(fully_connected))`.

jngrad (approved):

PR #4967, opened 2024-07-31
Add ZnDraw visualizer

phohenberger:
Description of changes:
- add support for the ZnDraw visualizer

### DIFF
--- a/src/python/espressomd/zn.py
+++ b/src/python/espressomd/zn.py
@@ -1,0 +1,624 @@
+#
+# Copyright (C) 2024 The ESPResSo project
+#
+# This file is part of ESPResSo.
+#
+# ESPResSo is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# ESPResSo is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+import subprocess
+import numpy as np
+import zndraw.zndraw
+import zndraw.utils
+import zndraw.draw
+import znsocket
+import znjson
+import espressomd
+import secrets
+import time
+import urllib.parse
+import typing as t
+
+
+# Standard colors
+color_dict = {"black": "#303030",
+              "red": "#e6194B",
+              "green": "#3cb44b",
+              "yellow": "#ffe119",
+              "blue": "#4363d8",
+              "orange": "#f58231",
+              "purple": "#911eb4",
+              "cyan": "#42d4f4",
+              "magenta": "#f032e6",
+              "lime": "#bfef45",
+              "brown": "#9A6324",
+              "grey": "#a9a9a9",
+              "white": "#f0f0f0"}
+
+
+class EspressoConverter(znjson.ConverterBase):
+    """
+    Converter for ESPResSo systems to ASEDict
+    """
+    level = 100
+    representation = "ase.Atoms"
+    instance = espressomd.system.System
+
+    def encode(self, system) -> zndraw.utils.ASEDict:
+        self.system = system
+        self.particles = self.system.part.all()
+        self.num_particles = len(self.particles)
+        self.params = system.visualizer_params
+
+        self.numbers = self.num_particles * [1]
+
+        if self.params["folded"] is True:
+            self.positions = self.particles.pos_folded
+        else:
+            self.positions = self.particles.pos
+
+        if self.params["colors"] is None:
+            self.colors = self.get_default_colors()
+        else:
+            self.colors = self.set_colors(self.params["colors"])
+
+        if self.params["radii"] is None:
+            self.radii = self.get_default_radii()
+        else:
+            self.radii = self.set_radii(self.params["radii"])
+
+        if self.params["bonds"] is True:
+            bonds = self.get_bonds()
+        else:
+            bonds = []
+
+        arrays = {
+            "colors": self.colors,
+            "radii": self.radii,
+        }
+        cell = [[system.box_l[0], 0, 0],
+                [0, system.box_l[1], 0],
+                [0, 0, system.box_l[2]]]
+        pbc = system.periodicity
+        calc = None
+        info = {}
+
+        if self.params["vector_field"] is not None:
+            vectors = self.params["vector_field"]()
+        else:
+            vectors = []
+
+        return zndraw.utils.ASEDict(
+            numbers=self.numbers,
+            positions=self.positions.tolist(),
+            connectivity=bonds,
+            arrays=arrays,
+            info=info,
+            calc=calc,
+            pbc=pbc.tolist(),
+            cell=cell,
+            vectors=vectors,
+        )
+
+    def decode(self, value):
+        value = None
+        return value
+
+    def get_default_colors(self):
+        return [color_dict["white"]] * self.num_particles
+
+    def get_default_radii(self):
+        return [0.5] * self.num_particles
+
+    def set_colors(self, colors):
+        color_list = list()
+        for p in self.particles:
+            color = colors[p.type]
+            # if color starts with #, assume it is a hex color
+            if color.startswith("#"):
+                color_list.append(color)
+            else:
+                if color not in color_dict:
+                    raise ValueError(
+                        f"Color {color} not found in color dictionary")
+                color_list.append(color_dict[color])
+        return color_list
+
+    def set_radii(self, radii):
+        radius_list = list()
+        for p in self.particles:
+            radius_list.append(radii[p.type])
+        return radius_list
+
+    def get_bonds(self):
+        bonds = []
+        for p in self.particles:
+            if not p.bonds:
+                continue
+            for bond in p.bonds:
+                if len(bond) == 4:
+                    bonds.append([p.id, bond[1], 1])
+                    bonds.append([p.id, bond[2], 1])
+                    bonds.append([bond[2], bond[3], 1])
+                else:
+                    for bond_partner in bond[1:]:
+                        bonds.append([p.id, bond_partner, 1])
+
+        self.process_bonds(bonds)
+
+        return bonds
+
+    def process_bonds(self, bonds):
+        half_box_l = 0.5 * self.system.box_l
+        num_part = len(self.positions)
+        bonds_to_remove = []
+        bonds_to_add = []
+
+        for b in bonds:
+            try:
+                if self.params["folded"] is True:
+                    x_a = self.system.part.by_id(b[0]).pos_folded
+                    x_b = self.system.part.by_id(b[1]).pos_folded
+                else:
+                    x_a = self.system.part.by_id(b[0]).pos
+                    x_b = self.system.part.by_id(b[1]).pos
+            except Exception:
+                bonds_to_remove.append(b)
+                continue
+
+            dx = x_b - x_a
+
+            if np.all(np.abs(dx) < half_box_l):
+                continue
+
+            if self.params["folded"] is False:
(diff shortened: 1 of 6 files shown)


phohenberger:
Use the `ZnDraw`-Visualizer to live visualize ESPResSo simulations in Jupyter Notebooks.

First you need to install `pip install zndraw==0.4.5`

Then you can import the relevant classes from the `/src/python/espressomd/zn.py` file.

Example usage:

```python
from espressomd.zn import Visualizer, LBField

color = {0: "red", 1: "#ff00ff"} # particle type mapping, there are some standard colors but you can also add hex
radii = {0: 0.2, 1: 0.5} # particle type mapping

LBField(system, step_x = 2, step_y=2,  step_z=2, scale=3.0)

vis = Visualizer(system=system, colors=color, radii=radii, vector_field = LBField, bonds=False, folded=False)

# Some integration loop
for i in range(1000):
    system.integrator.run(10)
    vis.update()
```
Of course stuff like `bonds`, `vector_field` are optional and can be left away. 

This currently only works in jupyter as the server closes down if the script is finished running to ensure no runaway processes.

phohenberger:
Some screenshots

image

image

src/python/espressomd/zn.py
```diff
@@ -0,0 +1,436 @@
+# Espresso visualizer with ZnDraw
```
jngrad: please remove this comment and add a copyright header with year 2024

src/python/espressomd/zn.py
```diff
@@ -0,0 +1,436 @@
...
+
+import subprocess
+import numpy as np
+from zndraw.zndraw import ZnDrawLocal, TimeoutConfig
+from zndraw.utils import ASEDict, get_port, direction_to_euler
+from zndraw.draw import Sphere, Plane, Cylinder, Rhomboid, Ellipsoid
```
jngrad: Here and elsewhere: consider writing `import zndraw.draw` instead to avoid namespace pollution. Keeping imported objects in a namespace is often preferred, since it avoids accidentally overshadowing an imported objects with a local object of the same name. There are no pylint rules to enforce it, because there are legitimate reasons for importing an object without a namespace, e.g. to provide API stability when a class is moved to a different file (by importing it in the original file: backward compatibility is preserved).

src/python/espressomd/zn.py
```diff
@@ -0,0 +1,436 @@
...
+    def _start_server(self):
+        """
+        Start the ZnDraw server through a subprocess
+        """
+        self.socket_port = get_port(default=6374)
+
+        self.server = subprocess.Popen(["zndraw", "--no-browser", f"--port={self.port}", f"--storage-port={self.socket_port}"],
+                                       stdout=subprocess.DEVNULL,
+                                       stderr=subprocess.DEVNULL
+                                       )
```
jngrad, line 486: The [subprocess.Popen()](https://docs.python.org/3/library/subprocess.html#popen-constructor) documentation has a long list of scary warnings and API changes. Basically every minor release of Python introduces 2-3 API changes, and they aren't sorted in chronological order. Isn't there a specialization available that has less API changes and less warnings, while still providing the same features, like [subprocess.check_call()](https://docs.python.org/3/library/subprocess.html#subprocess.check_call)?

src/python/espressomd/zn.py
```diff
@@ -0,0 +1,436 @@
...
+        objects = []
+
+        for shape in shapes:
+            if not isinstance(shape, t.Union[espressomd.shapes.Cylinder,
+                                             espressomd.shapes.Wall,
+                                             espressomd.shapes.Sphere,
+                                             espressomd.shapes.Rhomboid,
+                                             espressomd.shapes.Ellipsoid]):
+                raise ValueError(
+                    "Constraint must be of type Cylinder, Wall, Sphere, Rhomboid or Ellipsoid")
```
jngrad: The error message will get out of sync with the type union in the future. Consider inverting the error message, like so:
```python
raise NotImplementedError(f"Shape of type {shape.__class__.__name__} isn't available in ZnDraw")
```
This will actually tell the user which shape in the list of shape-based constraints is the culprit.

src/python/espressomd/zn.py
```diff
@@ -0,0 +1,436 @@
...
+
+            shape_type = shape.name()
+
+            if shape_type == "Shapes::Cylinder":
```
jngrad: I'm not sure we can guarantee the stability of these internal names. For example, the `Caliper` class has name `"ScriptInterface::Profiler::Caliper"` (extra `ScriptInterface::` namespace). These namespaces may change in the future, and since ZnDraw doesn't run in CI, any change to internal names will go unnoticed and get merged into the main branch. Because this if-else chain has no final `else` (the equivalent of `default` in a switch statement), any shape whose internal name has changed will be silently skipped.

Wouldn't it be more sustainable to adapt these if-else to check against the Python name, i.e. `shape.__class__.__name__`, rather than the internal name, and have a final `else` that throws the `NotImplementedError` mentioned above? Then you don't need to maintain a type union at lines 281-285 anymore.

src/python/espressomd/zn.py
```diff
@@ -0,0 +1,436 @@
...
+
+                objects.append(Ellipsoid(position=center,
+                               a=a, b=b, c=b, opacity=0.8))
+
+        self.zndraw.geometries = objects
+
+
+class LBField:
+    """
+    Convert the ESPResSo Lattice Boltzmann field to a vector field for visualization.
```
jngrad: here and elsewhere: it should be spelled "lattice-Boltzmann"

src/python/espressomd/zn.py
```diff
@@ -0,0 +1,436 @@
...
+        self.offset_y = offset_y
+        self.offset_z = offset_z
+        self.scale = scale
+        self.box = system.box_l
+
+        try:
+            self.lbf = system.lb
+            self.agrid = system.lb.agrid
+        except:
+            raise ValueError("System does not have a Lattice Boltzmann field")
```
jngrad: This try/catch is hiding the real reason the system LB doesn't have an agrid. Replace it with:
```python
if system.lb is None:
    raise ValueError("System does not have a lattice-Boltzmann solver")
self.lbf = system.lb
self.agrid = system.lb.agrid
```

src/python/espressomd/zn.py
```diff
@@ -0,0 +1,436 @@
...
+        velocities = origins + velocities
+        vector_field = np.stack([origins, velocities], axis=1)
+
+        return vector_field.tolist()
+
+
+class VectorField:
+    """
+    Give an array of origins and vectors to create a vector field for visualization.
+    Both must have the shape (VISUALIZATION_STEPS, N_VECTORS, 3)
```
jngrad: There is no mention of `VISUALIZATION_STEPS` and `N_VECTORS` anywhere else in this file. Please use the [`LBVelocityProfile`](https://github.com/espressomd/espresso/blob/2b7c8f20fca1d4abc5968267eeb86566701c996b/src/python/espressomd/observables.py#L294-L344) docstring as an example to improve this class documentation.

src/python/espressomd/zn.py
```diff
@@ -0,0 +1,436 @@
...
+              "white": "#f0f0f0"}
+
+
+class EspressoConverter(znjson.ConverterBase):
+    """
+    Converter for ESPResSo systems to ASEDict
+    """
+    level = 100
+    representation = "ase.Atoms"
+    instance = System
```
jngrad: Is this `instance` data member actually needed?

src/python/espressomd/zn.py
```diff
@@ -0,0 +1,436 @@
...
+                bonds.append([p.id, partner_id, 1])
+        return bonds
+
+
+znjson.config.register(EspressoConverter)
+
+
+class Visualizer():
+    def __init__(self,
+                 system: System = None,
```
jngrad: Documentation needed. Please add a section at the end of `doc/sphinx/visualization.rst` to give a minimal working example of how to instantiate and run this visualizer. Address the Sphinx warning about this file being not found. Look at variable `autodoc_mock_imports` in `doc/sphinx/conf.py.in` for how to disable imports of GUI Python modules by Sphinx.

src/python/espressomd/zn.py
```diff
@@ -0,0 +1,436 @@
...
+            if not p.bonds:
+                continue
+            for bond in p.bonds:
+                partner_id = bond[-1]
+                dist = np.linalg.norm(
+                    self.positions[p.id] - self.positions[partner_id])
+
+                # Remove bonds that are too long due to pbc
+                if dist > min(self.system.box_l) / 2: 
+                    continue
```
jngrad: The last particle in the bond list is not necessarily chemically bonded to the particle on which the list is stored.

In `src/python/espressomd/visualization.py`, look at:
* `_update_bonds()` for how to handle bonds with 2 or more partners
* `_draw_bonds()` for how to display bonds that go through PBC

src/python/espressomd/zn.py
```diff
@@ -0,0 +1,436 @@
...
+                    address=f"{self.url}:{self.socket_port}")
+                break
+            except:
+                time.sleep(0.5)
+
+        url = f"{self.url}:{self.port}"
+        self.zndraw = ZnDrawLocal(
+            r=self.r, url=url, token=self.token, timeout=config)
+        self.address = f"{self.zndraw.url}/token/{self.zndraw.token}"
+        self.address = self.address.replace("ws", "http")
```
jngrad: Is it safe to replace all occurrences of "ws" by "http"? Is it safe to replace the *first* occurrence of "ws" by "http"? There must be a function in the Python standard library to split the protocol from the address, and allow you to safely change the protocol without touching the address.

src/python/espressomd/zn.py
```diff
@@ -0,0 +1,436 @@
...
+        if self.frame_count == 0 and len(self.zndraw) > 0:
+            self.zndraw.__setitem__(0, znjson.dumps(
+                self.system, cls=znjson.ZnEncoder.from_converters(
+                    [EspressoConverter])
+            ))
+        else:
+            self.zndraw.append(znjson.dumps(
+                self.system, cls=znjson.ZnEncoder.from_converters(
+                    [EspressoConverter])
+            ))
```
jngrad: The `znjson.dumps()` statement seems repeated verbatim in both conditional branches. Can it be extracted in a variable?

src/python/espressomd/zn.py
```diff
@@ -0,0 +1,436 @@
...
+
+        if self.params["vector_field"] is not None and self.frame_count == 0:
+            self.zndraw.socket.sleep(1)
+            self.zndraw.config.arrows.colormap = self.arrow_config['colormap']
+            self.zndraw.config.arrows.normalize = self.arrow_config['normalize']
+            self.zndraw.config.arrows.colorrange = self.arrow_config['colorrange']
+            self.zndraw.config.arrows.scale_vector_thickness = self.arrow_config[
+                'scale_vector_thickness']
+            self.zndraw.config.arrows.opacity = self.arrow_config['opacity']
```
jngrad: These lines can probably be replaced by a for loop and [`setattr()`](https://docs.python.org/3/library/functions.html#setattr).

phohenberger: Yes, I will also make it possible to just give dictionaries containing a single key-value pair that you want to change from the standard config and keep the rest.

src/python/espressomd/zn.py
phohenberger: It is the correct syntax and used in the `ConverterBase`-class function `__eq__`. It is needed for correct execution, because the converter checks if the object to be converted matches this `instance`

src/python/espressomd/zn.py
phohenberger: I am not aware of a better alternative than `subprocess.Popen()`, because both `subprocess.run` and `subprocess.check_call` are blocking afaik. The server has to be permanently running in the background and the command used to starting the server does not complete until the server is shutdown, thus we need a non-blocking subprocess call. I'm aware `Popen` can be dangerous, but the only alternative Im aware of is starting the server manually in a commandline and then connecting to the server using jupyter in another commandline. If you have any suggestions on a safer option, I would be happy to use it.

jngrad: Good point. `Popen()` is our only option for ansynchronous processes.

jngrad (approved):

Issue #4963, opened 2024-07-23
P3M: further FFt refactoring


RudolfWeeber:
To my understanding, the FFT can be refactored furhter:

## move out unrelated halo communicatoin

The FFTBackend currently calls the halo communication before/after the FFT.
However, to my understanding, this halo communication is unrelated to the FFT. It should be done by the actual P3M classes.

Background: the halo communication is needed when interpolationg particle<->mesh. A particle sitting at the boundary needs to be able to read/write mesh sites across the boundary, so there is a halo layer. to my understanding, this is all in real space.

## Making explicit what is being transformed

To my understanding, we have the following FFTs

* forward

    * 3D scalar field (charge density for electrostatic P3M)
    * 3d vector field (dipole density for dipola P3M)

* back transform

    * 3d vector field (e.g., electric field)
    * 3d scalar field (e.g., electric potential)

IMO, it would be clearer, if the backend API provides these 4 functions, and the respective P3M implementations pass as arguments what they want transformed.

jngrad:
> The FFTBackend currently calls the halo communication before/after the FFT.

Can you show me an example? As far as I know, `fft.cpp` and `FFTBackendLegacy.cpp` have no knowledge of the particles and cell structure:

```
/work/jgrad/espresso/src/core/p3m/FFTBackendLegacy.cpp:112:30: error: ‘Particle’ was not declared in this scope
  112 | auto constexpr num1 = sizeof(Particle);
/work/jgrad/espresso/src/core/p3m/FFTBackendLegacy.cpp:113:30: error: ‘CellStructure’ was not declared in this scope; did you mean ‘CellStructureType’?
  113 | auto constexpr num2 = sizeof(CellStructure);
```

> IMO, it would be clearer, if the backend API provides these 4 functions, and the respective P3M implementations pass as arguments what they want transformed.

We can have 4 FFT methods. The new API would look like this:

```c++
// src/core/fft/fft.hpp
struct fft_data_struct {
  void forward_fft(boost::mpi::communicator const &comm, double *data);
  void backward_fft(boost::mpi::communicator const &comm, double *data, bool check_complex);
};

// src/core/p3m/data_struct.hpp
class FFTBackend {
  std::unique_ptr<fft_data_struct> fft;
  P3MFFTMesh &mesh;
  void update_mesh_data();
public:
  virtual void perform_scalar_fwd_fft() = 0;
  virtual void perform_vector_fwd_fft() = 0;
[...]

PR #4969, opened 2024-08-01
Simplify P3M C++ API

### LINKED ISSUE #4962: Template the floating point data type in P3M

jngrad:
Fixes #4962
Partial fix for #4963

Description of changes:
- disentangle `FFTBackend` methods: one per FFT transform operation, one per halo communication operation
- make the floating-point type and CPU/GPU architecture template parameters of P3M classes
- add support for P3M CPU and dipolar P3M CPU in single-precision mode


RudolfWeeber:
Offline discussion:
* we merge this PR onw.
* Next step is to change the data ownership. The actual meshes (rs_mesh, ks_mesh) should be owned by the solvers. ks_mesh is only needed by the dipolar p3m
* The forward and backward ffts in FFTBackend then take the meshs to transform as arguments
* The halo gather/scatter funciotns can then also be moved out of the FT Backend. They would be called by the actual solvers before the forward and after the backwards transforms, respectively
* The only link between the halo comm code and the fft is that they share the free functions to pack/unpack bllocks. These can be moved out of the fft and be used both, by the legacy fft implementaoitn and the halo communicaoitn (send_mesh)
* The main challange in all this is to hide implementatoin specific data types, e.g., fft_vector or some future Kokkos data type, fro mclients, especially the script interface.

RudolfWeeber (approved):

PR #4970, opened 2024-08-02
WIP: Add metatensor

### LINKED ISSUE #4933: Implement a Metatensor adaptor
## Data flow

See [Metatensor - Data flow](https://docs.metatensor.org/latest/atomistic/overview.html#data-flow-between-the-model-and-engine)

* ESPResSo needs to export a neighbor list in the form of a vector of distance vectors
* Metatensor sends back the central particle force, pressure tensor, energy, or observables like induced dipole

Metatensor expects distances, forces and energies in standard SI units. One has to set up the ESPResSo system such that e.g. the reduced unit is in Ångstroms, e.g. `box_l=[100, 100, 100]` means a box size of 100 Å. This can be achieved by designing a convenience function that leverages `pint` to create an empty ESPResSo system with the correct size, time step and energy units.

For domain decomposition: typically the ML model neighborhood radius is less than the interaction range. Thus we should be able to use metatensor in parallel without the need to tweak the ESPResSo domain decomposition and cell cutoff. One caveat: without LJ interaction the non-bonded cutoff will be zero, so we need to introduce a mechanism in the Metatensor plugin of the system call to adapt the `min_cutoff` cell system property.

Contact persons from the ML side: @SamTov, @Luthaf, @PicoCentauri

## Roadmap

* [ ] add pytorch dependency to the CI environment
* [ ] include Metatensor via CMake FetchContent
   * document in the user guide which CMake options are needed to help find pytorch (see [LAMMPS Metatensor docs](https://github.com/lammps/lammps/blob/ddc56d3738382444f16eac5c66d9abce40094dc0/doc/src/Build_extras.rst#ml-metatensor-package))
   * [Metatensor build instructions](https://docs.metatensor.org/latest/installation.html#install-c)
* [ ] write the C++ adaptor
   * adapt `std::vector<int> nbhood(System::System const &system, Utils::Vector3d const &pos, double dist)` to return a vector of particle distances (can be achieved with [`std::ranges::transform`](https://en.cppreference.com/w/cpp/algorithm/ranges/transform) and a `Particle` to `Vector3d` projector)
   * use pytorch's `autograd` feature to compute the force as the gradient of the energy
* [ ] introduce a Python API with the following input parameters:
   * ML model and any additional parameters, e.g. file paths
   * list of particle types for which the C++ core needs to export descriptors
      * which data structure would be suitable to represent the primary types and optional cross-terms?
   * figure out how to set `system.min_global_cut`
   * see [Metatensor - Exporting models](https://docs.metatensor.org/latest/atomistic/reference/models/export.html)
* [ ] write unit tests in the `testsuite/python` folder
* [ ] write a convenience function to instantiate an ESPResSo system with the correct reduced units

See https://github.com/lammps/lammps/pull/4163 for how Metatensor is being integrated into LAMMPS.

jhossbach:
Closes #4933


RudolfWeeber:
Progress: 


* Sketch of the main code for model loading and running with Lammps dependenceis removed. This will not compile. Just skteches. Also, sno Espresso specific code, yet.
* adding dependnceis in CMake. I failed to ge fetchcontent working. In the end, I built both metatnesor-core and metatensor-torch externally.

RudolfWeeber:
Notes on building:

## Torhc

```bash
wget "https://download.pytorch.org/libtorch/cu121/libtorch-cxx11-abi-shared-with-deps-2.4.0%2Bcu121.zip"

unzip libtorch-cxx11-abi-shared-with-deps-2.4.0+cu121.zip 
```

Note that this is libtorch, not torch propper.
Everything is pre-compiled. Dependencies included. the archive is huge. There is a cpu-only version of this as well, which, hopefully, is small enough to be included in CI.


## Intsall rust
I used the rustup script from the rust installatoin page, which installs to ~..cargo.

Source rust config
```bash
. ~/.cargo/env
```



## metatensor

git clone from lab-cosmo/metatensor
cd metatensor/metatensor-coe
cmake . -DCMAKE_INSTALL_PREFIX=~/es_mt_deps # or other user-owned target dir
make 
make install
```

## metatensor-torch

```
cd metatensor/metatensor-torch
cmake . -DTorch_DIR=whereever/libtorch/share/cmake/Torch -dmetatensor_DIR=~/es_mt_deps/lib/cmake/metatensor
make
make install
```

## Espresso


Some of the following is ICP specific. Main points are torhc and metatensor dirs.

```bash
GCC_VER=12; CUDA_VER=12.0; CC=gcc-${GCC_VER} CXX=g++-${GCC_VER} CUDACXX=/usr/local/cuda-${CUDA_VER}/bin/nvcc CUDAARCHS="61;75;86" cmake ~/es -D ESPRESSO_BUILD_WITH_CUDA=ON -D CUDAToolkit_ROOT=/usr/local/cuda-${CUDA_VER} -D ESPRESSO_BUILD_WITH_CCACHE=ON -D ESPRESSO_BUILD_WITH_WALBERLA=ON -D ESPRESSO_BUILD_WITH_WALBERLA_FFT=ON -D ESPRESSO_BUILD_WITH_WALBERLA_AVX=ON ESPRESSO_BUILD_WITH_HDF5=OFF -D ESPRESSO_BUILD_WITH_GSL=ON -D CMAKE_CUDA_FLAGS="--compiler-bindir=/usr/bin/g++-${GCC_VER}" \
   -DESPRESSO_BUILD_WITH_METATENSOR=on \
   -DTorch_DIR=/ssd/weeber/libtorch/share/cmake/Torch\
   -Dmetatensor_DIR=~/es_mt_deps/lib/cmake/metatensor\
   -Dmetatensor_torch_DIR=~/es_mt_deps/lib/cmake/metatensor_torch

```

RudolfWeeber:
Forgot two points:

* A small patch to metatensor is needed, seehttps://github.com/lab-cosmo/metatensor/pull/712
* Espresso uses CUDA (from feature configuraiton) as a macro. This causes a compiler error in torch. Before including anything toch, #undef CUDA.

jhossbach:
(Also do not forget to append `libtorch/lib` and `espresso_metatensor_deps/lib` to the `LD_LIBRARY_PATH`)

ceriottm:
hello there, we're putting the finishing touches on a [paper on the "metatensor ecosystem"](https://arxiv.org/abs/2508.15704) and we would very much love to mention also the integration in ESPRESSO. is there any way we can help to get this merged?

PicoCentauri:
I just had a very nice offline discussion with @jhossbach and @RudolfWeeber and we try to get this going again.

jhossbach:
Current setup that builds (on my machine):
```bash
export PATH=/usr/local/cuda-12.4/bin:$PATH
export LD_LIBRARY_PATH=/usr/local/cuda-12.4/lib64:$LD_LIBRARY_PATH
export CUDA_TOOLKIT_ROOT_DIR=/usr/local/cuda-12.4

rm -rf build
mkdir build
cd build

source ../.venv/bin/activate

echo $CMAKE_PREFIX_PATH
TORCH_PREFIX=$(python -c "import torch; print(torch.utils.cmake_prefix_path)")
MTS_PREFIX=$(python -c "import metatensor; print(metatensor.utils.cmake_prefix_path)")
MTS_TORCH_PREFIX=$(python -c "import metatensor.torch; print(metatensor.torch.utils.cmake_prefix_path)")
MTA_TORCH_PREFIX=$(python -c "import metatomic.torch; print(metatomic.torch.utils.cmake_prefix_path)")
CMAKE_PREFIX_PATH="$TORCH_PREFIX;$MTS_PREFIX;$MTS_TORCH_PREFIX;$MTA_TORCH_PREFIX"
echo $CMAKE_PREFIX_PATH

cmake ../ \
  -D ESPRESSO_BUILD_WITH_CUDA=OFF \
  -D ESPRESSO_BUILD_WITH_CCACHE=OFF \
  -D ESPRESSO_BUILD_WITH_WALBERLA=OFF \
  -D ESPRESSO_BUILD_WITH_WALBERLA_FFT=OFF \
  -D ESPRESSO_BUILD_WITH_WALBERLA_AVX=OFF \
  -D ESPRESSO_BUILD_WITH_GSL=OFF \
  -D ESPRESSO_BUILD_WITH_METATENSOR=ON \
  -D CMAKE_PREFIX_PATH="$CMAKE_PREFIX_PATH"
  ```

PR #4977, opened 2024-08-13
Introduce an event on_lees_edwards_change

### LINKED ISSUE #4874: Lees-Edwards update after LB construction
LB ignores Lees Edwares boundary condition, if it is set up after LB. To reproduce, switch the initializaiton order from 
1. LE
2. LB
to the opposite in the Lb Lees Edwards tests.

As an aside, the code that constructs the LeesEdwards Pack pased to LbWalberla should probbably be somewhere in teh core rather than in the script interface.

davidbbeyer:
Fixes #4874

Description of changes:
- introduced a new event function on_lees_edwards_change() to be triggered when the LE BC is changed
- the order of instantiation of LB and LEbc no longer matters: collision kernel is automatically updated

### DIFF
--- a/src/core/lb/LBWalberla.cpp
+++ b/src/core/lb/LBWalberla.cpp
@@ -38,6 +38,7 @@
 #include <utils/math/int_pow.hpp>
 
 #include <optional>
+#include <variant>
 
 namespace LB {
 
@@ -128,7 +129,9 @@ void LBWalberla::update_collision_model(LBWalberlaBase &lb,
                                         LBWalberlaParams &params, double kT,
                                         unsigned int seed) {
   auto const &system = ::System::get_system();
-  if (auto le_protocol = system.lees_edwards->get_protocol()) {
+  auto le_protocol = system.lees_edwards->get_protocol();
+  if (le_protocol and
+      not std::holds_alternative<LeesEdwards::Off>(*le_protocol)) {
     if (kT != 0.) {
       throw std::runtime_error(
           "Lees-Edwards LB doesn't support thermalization");
--- a/src/core/lb/LBWalberla.cpp
+++ b/src/core/lb/LBWalberla.cpp
@@ -30,6 +30,8 @@
 #include "system/System.hpp"
 #include "thermostat.hpp"
 
+#include "lees_edwards/lees_edwards.hpp"
+
 #include <walberla_bridge/lattice_boltzmann/LBWalberlaBase.hpp>
 
 #include <utils/Vector.hpp>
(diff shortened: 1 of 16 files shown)


RudolfWeeber:
LGTM. Could you please check the codecov tst for the patch (which shows that coverage deceased slightly) and make sure, we don't have important code untested?
Thanks!

davidbbeyer:
As far as I can tell, it is only complaining about the function `LBWalberla::on_lees_edwards_change()` which I added. However, this function is (indirectly) tested in the additional test in `/testsuite/python/lb_lees_edwards.py `, so I guess it should not be a problem.

src/core/lb/LBWalberla.cpp
```diff
@@ -112,6 +114,30 @@ void LBWalberla::sanity_checks(System::System const &system) const {
...
 
+void LBWalberla::on_lees_edwards_change() {
+  auto &system = System::get_system();
+  auto const energy_conversion =
+      Utils::int_pow<2>(lb_params->get_agrid() / lb_params->get_tau());
+  auto const lb_kT = lb_fluid->get_kT() * energy_conversion;
+  auto le_protocol = system.lees_edwards->get_protocol();
+  if (lb_kT != 0.) {
+    throw std::runtime_error("Lees-Edwards LB doesn't support thermalization");
+  }
```
jngrad: The code coverage report is highlighting this throw statement has having no coverage. The corresponding exception in the script interface does have code coverage, so we could find which line in the python test uses a thermalized LB kernel, duplicate it and flip the LB and LE order of instantiation.

src/core/lb/LBWalberla.cpp
```diff
@@ -112,6 +114,30 @@ void LBWalberla::sanity_checks(System::System const &system) const {
...
+  auto const &le_bc = system.box_geo->lees_edwards_bc();
+  auto lees_edwards_object = std::make_unique<LeesEdwardsPack>(
+      le_bc.shear_direction, le_bc.shear_plane_normal,
```
jngrad: I had a quick look at the Lees-Edwards code recently for an unrelated task, and vaguely remember that there are two ways for the user to remove Lees-Edwards: either set the `Off` protocol, or assign the Python value `None`, in which case variable `le_protocol` at line 122 in this file is a nullptr. It was quite a headache for @1234somesh to find all the places in the code where he had to introduce conditional checks for nullptr.

I'm mentioning this, because the script interface doesn't seem to call `system.on_lees_edwards_change()` when users set the protocol to `None`, in which case the LB keeps the LE collision operator. Could you please double-check that edge case? If you determine there is a need to call `system.on_lees_edwards_change()` when taking the `if (is_none(protocol))` branch in `src/script_interface/lees_edwards/LeesEdwards.hpp` (to reproduce the same behavior as setting the protocol to `Off`), then a nullptr check will be necessary in this function. An elegant way to introduce such a nullptr check will be described in the next code review comment, where an `update_collision_model()` method is proposed.

src/core/lb/LBWalberla.cpp
```diff
@@ -112,6 +114,30 @@ void LBWalberla::sanity_checks(System::System const &system) const {
                              system.get_time_step());
 }
 
+void LBWalberla::on_lees_edwards_change() {
+  auto &system = System::get_system();
+  auto const energy_conversion =
+      Utils::int_pow<2>(lb_params->get_agrid() / lb_params->get_tau());
```
jngrad: Out of curiosity, would there be a way to move the LE collision operator logic out of this event into a self-contained function? Because then we could call it from the event and also from the script interface, which would greatly reduce code duplication. We could call this function `set_collision_model()` and have two overloads, one for LE and one for thermalized. This would reflect the API of `LBWalberlaBase` in the core. This would also remove the need for the script interface to query the system global variable, and least in the particular case of setting the collision operator.

A possible implementation could look like this:
```c++
struct LBWalberla {
  void set_collision_model(std::unique_ptr<LeesEdwardsPack> &&lees_edwards_pack);
  void set_collision_model(double kT, unsigned int seed);
  void update_collision_model() {
    auto const &system = System::get_system();
    if (auto le_protocol = system.lees_edwards->get_protocol()) {
      // call Lees-Edwards overload
    } else {
      // call thermalized overload
    }
  }
  void LBWalberla::on_lees_edwards_change() {
    update_collision_model();
  }
};
```

jngrad (commented):
Thanks for taking care of this! I have selected a few lines were we potentially have edge cases that may need extra attention.

jngrad:
I noticed a lot of commits don't pass CI. Would you kindly update your virtual environment using our [pep8 instructions](https://github.com/espressomd/espresso/wiki/Tooling#user-content-Linter_and_code_formatter)?

jngrad (approved):
Thanks a lot!

Issue #4972, opened 2024-08-06
Add more ZnDraw-features


phohenberger:
- [x] Move `len(zndraw)` call in update loop to avoid socket-call
- [ ] ~~Allow non-Jupyter use~~ postponed due to unsolved runaway server problems
- [x] Allow multiple frames in the same file using different tokens
- [x] ~~Asynchronous Server startup~~
- [x] ~~Add color gradients for particle colors depending on the distance to reference~~
- [x] Make constraints proportionate to the box dimension, update ZnDraw shape
   - Add clipping mechanism for 2D shapes that cross the box boundaries
- [x] Make the box center the origin, instead of a box corner (confusing when rotating the box view)
- [x] Expose the ZnDraw script interface to Python
   - At least provide a way to set the camera angle from Python
- [ ] Fix inconsistencies in server communication
   - [X] on `lama`, ZnDraw creates a white rectangle in Jupyter instead of displaying the GUI (works fine in Chrome, on Firefox one can still click on the server URL to open the visualizer in a new tab and see everything)
      - solution: completely deactivate cookie blocking in the web browser settings (not just from third-party or unvisited sites, *allow all cookies*)
   - on CIP pool machines, using Firefox, once a ZnDraw server was used in a Notebook, when opening another Notebook in the same Jupyter session, ZnDraw cannot load the particle data
   - on other web browsers like GNOME Web, ZnDraw cannot load the particle data
- [x] ~~Find out if `datamodel-code-generator` can be made a soft dependency (it has a lot of dependencies)~~ not possible
- [ ] Visualize `ext_forces`

jngrad:
Here is a one-line change to make the complete CI/CD pipeline happy:
```diff
diff --git a/testsuite/scripts/importlib_wrapper.py b/testsuite/scripts/importlib_wrapper.py
index f3dd8e5b44..5bc879338c 100644
--- a/testsuite/scripts/importlib_wrapper.py
+++ b/testsuite/scripts/importlib_wrapper.py
@@ -396,3 +396,3 @@ class GetEspressomdVisualizerImports(ast.NodeVisitor):
     def __init__(self):
-        self.visualizers = {"visualization"}
+        self.visualizers = {"visualization", "zn"}
         self.namespace_visualizers = {
```

PythonFZ:
- [ ] Visualize `ext_forces`

phohenberger:
[...]

PR #4984, opened 2024-08-16
ZnDraw Visualizer patch

phohenberger:
Partial fix for #4972

Description of changes:
- Remove unnecessary socket-call in the update loop
- Allow multiple instances of the visualizer in the same notebook
- Have only one server run in the background when initializing the first `Visualizer`-instance
- Center the camera on initialization
- Clip wall shape at the box boundaries
- Expose the `self.zndraw`-object for direct manipulation of `ZnDraw`
- Bump to 'ZnDraw 0.4.6'

### DIFF
--- a/doc/sphinx/installation.rst
+++ b/doc/sphinx/installation.rst
@@ -132,7 +132,7 @@ To install the ZnDraw visualizer:
 
 .. code-block:: bash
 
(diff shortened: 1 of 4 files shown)


jngrad (approved):
Thank you very much!

PR #4991, opened 2024-09-09
Cuda 12.4 compatibility

reinaual:
Description of changes:

- add support for nvcc 12.4
   - sometimes `decltype` somehow fails to deduct the correct type of items in range-based for loops


jngrad (approved):
Thanks!

PR #4995, opened 2024-09-25
Implement manual trigger of bond breakage

### LINKED ISSUE #4665: Implement manual trigger of bond breakage (for use with Lees-Edwards)
* The breakable bonds feature removes bonds stretched beyond a certain limit.
* The breaking happens during the integration loop, after the force calculation
* Under normal conditions, no bonds exist with a length beyond the breakage limit after the integration has finished
* But when Lees-Edwards (LE) boundary conditions are active, the shift across the periodic boundary changes, as the time step is advanced. In this way, bound particles interacting across the periodic boundaries can be separated by more than the bond length in rare cases.

@bindgens1

Temporary "solution":  rewind the LE offset change applied in the last time step before stress measurement, measure stress, re-apply LE offset.

The proper solution is to introduce a script interface call to manually trigger the bond breakage. I'd hesitate to put this into the analysis routines, because analysis shouldn't actually modify the system.

### Core

* In `src/core/bond_breakage/bond_breakage.{cpp,hpp}`:
   * Write a bond handler which, if it gets a pair bond, evaluates the bond breakage criterion.
      * This can be a copy of the first few lines of `forces_inline.hpp::add_bonded_force()`
   * Write an `execute_bond_breakage()`  function
      * Call the method to clear the bond breakage queue
      * Use `CellStructure::bond_loop` with the bond kernel from the previous step (cf `short_range_loop.hpp`)
      * Run the method to process the breakage queue

### Script interface
* Create `src/script_interface/bond_breakage/BondBreakage.hpp` as an instance of `AutoParameters`. 
    * See `src/script_interface/particle_data/Poymer.{hpp,cpp}` as template
    * The constructor can be left empty (or omitted)
    * Only the `do_call_method()` has to be implemented
    * Add a single method called "execute", calling the `execute_bond_breakage()` method written in the core
    * Register the `BondBreakage` script interface class in `src/script_interface/bond_breakage/initialize.cpp`
    * Add the `BondBreakage.cpp` file to the `CMakeList.txt`

### Python

* Make the `BondBreakage` script interface class known to python.
   * Cf the Union class in `shapes.py`
   * Only the `_so_name` and `_bind_methods` are needed

1234somesh:
Fixes #4665

### DIFF
--- a/src/python/espressomd/bond_breakage.py
+++ b/src/python/espressomd/bond_breakage.py
@@ -42,6 +42,10 @@ class BreakageSpec(ScriptInterfaceHelper):
 class BreakageSpecs(ScriptObjectMap):
     _so_name = "BondBreakage::BreakageSpecs"
 
+    def execute(self):
+        """Execute the bond breakage on the current state of the system"""
+        self.call_method("execute")
+
     def _get_key(self, key):
         """Convert a bond object to a bond id."""
         if isinstance(key, interactions.BondedInteraction):
--- a/src/core/bond_breakage/bond_breakage.cpp
+++ b/src/core/bond_breakage/bond_breakage.cpp
@@ -204,4 +204,42 @@ void BondBreakage::process_queue_impl(System::System &system) {
   }
 }
 
+static bool bond_handler(BondBreakage &bond_breakage, Particle &p,
+                         std::span<Particle *> partners, int bond_id,
+                         BoxGeometry const &box_geo) {
+  auto retval = false;
+  if (partners.size() == 1u) { // pair bonds
+    auto d = box_geo.get_mi_vector(p.pos(), partners[0]->pos()).norm();
+    retval = bond_breakage.check_and_handle_breakage(
+        p.id(), {{partners[0]->id(), std::nullopt}}, bond_id, d);
+  } else if (partners.size() == 2u) { // angle bond
+    auto d =
+        box_geo.get_mi_vector(partners[0]->pos(), partners[1]->pos()).norm();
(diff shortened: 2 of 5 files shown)


RudolfWeeber (dismissed):

src/python/espressomd/bond_breakage.py
```diff
@@ -55,3 +55,19 @@ def __getitem__(self, key):
...
+    This class provides methods to manage and manipulate bond breakage.
+
+    Methods
+    -------
+    execute(parameters)
+        Executes the bond breakage with the given parameters.
+    """
+
+    _so_name = "BondBreakage::BondBreakage"
+    _so_bind_methods = ("execute",)
```
jngrad: @RudolfWeeber I'm not a big fan of having this feature as a stand-alone class. It currently only works on the main `System` instance, and ignores all other instances. How about making this feature a member method of the `BreakageSpecs` class? Then one could do `system1.bond_breakage.execute()` to execute all specs.

jngrad (changes requested):
The feature is missing a test. The feature throws an exception.

RudolfWeeber:
The probelm was that
* bond breakage handlers return true if they did something
* but kernels to cell_strucutre::bonded_loop (and short_range_loop) return true if a bond is broken


I.e. the return value of the former has to be discarded in the latter.

src/python/espressomd/bond_breakage.py
RudolfWeeber: moved it.

jngrad (approved):

PR #5003, opened 2024-10-10
avoid custom ubb

RudolfWeeber:
Use plain LbmPy bounce back boundary, but with streaming_pattern=pull, as we call it before the stream sweep. The patched boundary additional data handling is still needed. This accounts for the fact that the C++ boundary implementation queries the fluid node (not the boundary node) for the boundary velocity.


jngrad (approved):
Works fine, also with the new changes in [pycodegen/lbmpy!183](https://i10git.cs.fau.de/pycodegen/lbmpy/-/merge_requests/183) (as of this writing).

PR #5014, opened 2024-11-15
Improve LB GPU and thermostats API

jngrad:
Description of changes:
- LB GPU now supports VTK output
- LB GPU total pressure tensor and total momentum are now significantly faster
- thermostat parameter `gamma` is now required
   - fixes a regression introduced in the last thermostat API refactoring


hidekb:
Looks Good To Me. However, there is one question. The function set in some files generated_kernels/FieldAccessors* is simply moved to under the function std::vector<*> get. Does it mean anything?

jngrad:
> Looks Good To Me. However, there is one question. The function set in some files generated_kernels/FieldAccessors* is simply moved to under the function std::vector<*> get. Does it mean anything?

This move has no consequences. I re-ordered functions in the template files a while ago to make CPU and GPU header files more consistent and reduce the size of the diff between them, and forgot to push the regenerated kernels back then. Sometimes when regenerating the collide and stream kernels, the variables can also be re-ordered by pystencils, making the diff difficult to inspect.

hidekb (approved):
I understand. LGTM

RudolfWeeber (approved):

PR #5015, opened 2024-12-03
Update P3MGPU documentation

davidbbeyer:
Description of changes:
- Update documentation to reflect that P3MGPU is in fact able to execute the CPU energy and pressure kernels using parameters tuned for the GPU force kernel since #4506.


jngrad (approved):
Thanks!

PR #5016, opened 2024-12-04
Fix SimplePore distance function

jngrad:
The `SimplePore` shape used trigonometric functions incorrectly and generated NaN values in the core that would not propagate, but would still be a source of undefined behavior in if/else statements and would cause fatal errors when floating-point exceptions trapping was enabled.

The new distance calculation doesn't rely on trigonometric functions, is 5% faster, and is fully tested.


reinaual (approved):
LGTM

PR #5017, opened 2024-12-11
Add new P3M tuning options

jngrad:
Description of changes:
- new feature: mesh tuning can now be constrained to yield a value between two user-specified mesh sizes
- incompatible FFT domain decompositions are now discarded during CPU tuning to avoid undefined behavior

### DIFF
--- a/src/core/magnetostatics/dlc.cpp
+++ b/src/core/magnetostatics/dlc.cpp
@@ -50,6 +50,7 @@
 #include <cmath>
 #include <cstdio>
 #include <functional>
+#include <limits>
 #include <numbers>
 #include <numeric>
 #include <stdexcept>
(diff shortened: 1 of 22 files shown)


src/script_interface/electrostatics/CoulombP3M.hpp
```diff
@@ -103,6 +114,33 @@ class CoulombP3M : public Actor<CoulombP3M<Architecture>, ::CoulombP3M> {
...
+      std::vector<Variant> range;
+      try {
+        auto const val = get_value<std::vector<int>>(params, "tune_limits");
+        assert(val.size() == 2u);
+        range.emplace_back(val[0u]);
+        range.emplace_back(val[1u]);
+      } catch (...) {
+        range = get_value<std::vector<Variant>>(params, "tune_limits");
+        assert(range.size() == 2u);
+      }
```
reinaual: Why do we need this ugly try-catch? Don't we get the same result when always taking the `catch`-code path?

src/script_interface/magnetostatics/DipolarP3M.hpp
```diff
@@ -98,6 +109,33 @@ class DipolarP3M : public Actor<DipolarP3M<Architecture>, ::DipolarP3M> {
...
+      std::vector<Variant> range;
+      try {
+        auto const val = get_value<std::vector<int>>(params, "tune_limits");
+        assert(val.size() == 2u);
+        range.emplace_back(val[0u]);
+        range.emplace_back(val[1u]);
+      } catch (...) {
+        range = get_value<std::vector<Variant>>(params, "tune_limits");
+        assert(range.size() == 2u);
+      }
```
reinaual: Same thing

src/script_interface/electrostatics/CoulombP3M.hpp
jngrad: A Python list containing only integers is passed to the script interface as a `std::vector<int>`, while a list containing at least one None is considered mixed type, and is passed as `std::vector<Variant>`. The `get_value()` function cannot convert a `std::vector<int>` object to a `std::vector<Variant>` object:

`RuntimeError: Provided argument of type 'std::vector<int>{.size=2}' for parameter 'tune_limits' is not convertible to 'std::vector<ScriptInterface::Variant>' because it contains a value that is not convertible to 'ScriptInterface::Variant'`

I've rewritten this code using a type check instead of a try/catch to make it more readable.

reinaual (approved):

src/core/magnetostatics/dlc.cpp
```diff
@@ -430,15 +431,26 @@ double DipolarLayerCorrection::tune_far_cut() const {
...
+    auto const pref2 = g1_DLC_dip(gc, lz + h);
+    auto const pref0 = g1_DLC_dip(gc, lz);
+    auto const exp_term = 2. * gc * h;
+    auto const log_term1 = exp_term + std::log(9. * pref1);
+    if (exp_term >= exp_max or log_term1 >= log_max or -exp_term <= exp_min) {
+      // no solution can be found at larger k values due to overflow/underflow
+      break;
+    }
+    auto const fa0 = std::sqrt(9. * std::exp(+exp_term) * pref1 +
+                               9. * std::exp(-exp_term) * pref2 + 22. * pref0);
```
reinaual, line 453: It's probably a task for a coding day, but I think we should investigate where the need for such large values comes from. We either have the same problem in ELC or this situation can be prevented by a proper tuning. Either way I think it's worth taking a look at some point.

PR #5019, opened 2024-12-16
Sympy potential

ludogibbs:
Description of changes:
- new feature: generate non-bonded tabulated potentials from sympy expressions

### DIFF
--- a/src/python/espressomd/interactions.py
+++ b/src/python/espressomd/interactions.py
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2013-2022 The ESPResSo project
+# Copyright (C) 2013-2025 The ESPResSo project
 #
 # This file is part of ESPResSo.
 #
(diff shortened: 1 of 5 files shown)


RudolfWeeber:
Thanks, Lukas. Plesae address the following:
* Add your test case to the CMakeList.txt in testsuite/python, so it gets actually run in CI.
* translate the comments in your code to English
* pls use a more meaningful arument name than `f` for the energy expression
* I like the string-based expression input, but it does not cover all cases (e.g. piecewhise functions or other more unusual sympy expressions). So please keep that but also accept a sympy expression in additoin to a string for the argument.
* A furhter argument `distance_symbol` should be added, so people can use something else than `r`. Keeping `r` as default is fine, I guess.

src/python/espressomd/interactions.py
```diff
@@ -19,6 +19,8 @@
 
 import abc
 import enum
+from sympy import sympify, symbols,oo,nan
```
jngrad: Consider writing `import sympy` or `import sympy as sp` to limit module namespace pollution.

src/python/espressomd/interactions.py
```diff
@@ -350,6 +352,81 @@ def default_params(self):
...
+        params.update(kwargs)
+
+        if "energy" not in params or "force" not in params:
+            required_keys = {"min", "max", "steps", "f"}
+            missing_keys = required_keys - params.keys()
+            if missing_keys:
+                raise ValueError(
+                    f"Missing keys for table generation: {missing_keys}. "
+                    f"To generate energy and force tables, provide {required_keys}."
+                )
```
jngrad: I'm a bit uneasy about dynamically changing the set of expected arguments of a function depending on whether an argument is missing. From a user experience perspective, it's unclear what the function actually wants, or what happens when forces, energies, and a function are provided simultaneously. This also introduces a maintenance issue: the logic from the parent class `set_params()` method is duplicated here, thus if the parent class behavior changes, it won't be reflected here and `TabulatedNonBonded` will diverge from the rest of the code.

What would you think about creating a static class method like `make_tables(cls, f, min, max, steps)` that would return the energy and force tables, which can then be fed to `set_params(self, **kwargs)`?

src/python/espressomd/interactions.py
```diff
@@ -350,6 +352,81 @@ def default_params(self):
...
+        
+            # Berechnung von Energie- und Krafttabellen mit `get_table`
+            energy_tab, force_tab = self.get_table(
+                min=params["min"],
+                max=params["max"],
+                steps=params["steps"],
+                f=params["f"],
+                **{k: v for k, v in params.items() if k not in required_keys}  # Zusätzliche Parameter
```
jngrad: Here and elsewhere: please write comments in English language.

testsuite/python/tabulated_sympy.py
```diff
@@ -0,0 +1,70 @@
+#
+# Copyright (C) 2013-2022 The ESPResSo project
```
jngrad: The license header should reflect the year range you actually worked on the code, plus the years of the original code if you copied parts of it from somewhere else.

testsuite/python/tabulated_sympy.py
```diff
@@ -0,0 +1,70 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+import unittest as ut
+import unittest_decorators as utx
+import espressomd
+import espressomd.interactions
+import numpy as np
+from sympy import symbols, sympify
+
+class TabulatedTest(ut.TestCase):
```
jngrad: sympy should be available in some of the CI containers. You can disable `TabulatedTest` when sympy is missing with fixture `@utx.skipIfMissingModules("sympy")` above the class name. Don't forget to add a linter rule exception for unused imports.

src/python/espressomd/interactions.py
jngrad: Also: the present code makes `sympy` a hard requirement, even when `TABULATED` is not compiled in. Consider using a deferred import.

src/python/espressomd/interactions.py
```diff
@@ -350,6 +352,81 @@ def default_params(self):
...
+        filtered_params = {k: params[k] for k in relevant_keys if k in params}
+
+        err_msg = f"setting {self.__class__.__name__} raised an error"
+        self.call_method("set_params", handle_errors_message=err_msg, **filtered_params)
+
+        
+
+    def get_table(self, **kwargs):
+        """
+        Generate energy and force tables.
```
jngrad: "...from a function :math:\`f(r)\` that evaluates the force magnitude for a given inter-particle distance :math:\`r\`."

jngrad (commented):
Thanks! Don't forget to add the test to the CMake build system and to install and run the linters on your workstation. Here are the relevant sections in the developer guide:
[Linter and code formatter](https://github.com/espressomd/espresso/wiki/Tooling#user-content-Linter_and_code_formatter), [Python integration tests](https://github.com/espressomd/espresso/wiki/Testing#user-content-Python_integration_tests).

RudolfWeeber:
Looks good. Just 2 things:
* please rename the f argument to sth more meaningfull likle `energy_expr`
* the linter is complaining about the test importing something fro mSympy you aren't using.

RudolfWeeber:
@jngrad could you pls take a look? I am not sure, what the linter wants.
Once that's resoled, we can go ahead with this pr, form my side.

jngrad (approved):

PR #5020, opened 2024-12-17
Improve code portability

jngrad:
Description of changes:
- upgrade to latest waLBerla version and most recent pystencils/lbmpy/sympy releases
   - rewrite codegen pipeline to support multiple pystencils versions
- bugfixes:
   - remove sources of NaN, float overflow, and most float underflow
   - copy all waLBerla shared objects to the installation directory
   - add prefix to waLBerla and Caliper shared objects to avoid name clashes
- new feature: floating point exceptions instrumentation for x86 and Armv8
- run CI on Apple M1 (virtual machine, ARM 64bit architecture)
- give fine control over Clang-Tidy diagnostics for each CMake target


src/walberla_bridge/tests/LBWalberlaImpl_bspline_tests.cpp
```diff
@@ -119,9 +119,12 @@ BOOST_DATA_TEST_CASE(velocity_interpolation_bspline, bdata::make(all_lbs()),
...
 
-  for (double x = 0.0; x < params.box_dimensions[0]; x += 0.3) {
-    for (double y = 0.1; y < params.box_dimensions[1]; y += 0.3) {
-      for (double z = 0.2; z < params.box_dimensions[2]; z += 0.3) {
+  auto x = 0.0;
+  while (x < params.box_dimensions[0]) {
+    auto y = 0.1;
+    while (y < params.box_dimensions[1]) {
+      auto z = 0.2;
+      while (z < params.box_dimensions[2]) {
```
reinaual, line 127: Why does Clang-Tidy prefer this over the for loop? To me this seems less readable.

src/walberla_bridge/src/lattice_boltzmann/LBWalberlaImpl.hpp
```diff
@@ -673,7 +672,7 @@ class LBWalberlaImpl : public LBWalberlaBase {
     }
   }
 
-  void ghost_communication_pdf() override {
+  void ghost_communication_pdf() override { // TODO: is this never traversed???
```
reinaual: What about this todo?

src/core/electrostatics/elc.cpp
```diff
@@ -462,6 +458,11 @@ void ElectrostaticLayerCorrection::add_z_force(
           gblcblk[0] += elc.delta_mid_top * q;
       }
     } else {
+      // metallic boundaries
```
reinaual: This is not the metallic case, it's the dielectric case. The `elc.const_pot`-path is the metallic one.

src/core/reaction_methods/ReactionAlgorithm.cpp
```diff
@@ -274,7 +274,9 @@ double ReactionAlgorithm::make_reaction_mc_move_attempt(int reaction_id,
                                                         double bf,
                                                         double E_pot_old,
                                                         double E_pot_new) {
-  auto const exponential = std::exp(-(E_pot_new - E_pot_old) / kT);
+  auto constexpr exp_min = -708.4; // for IEEE-compatible double
+  auto const exponent = -(E_pot_new - E_pot_old) / kT;
+  auto const exponential = (exponent < exp_min) ? 0. : std::exp(exponent);
```
reinaual, line 279: Why do we care about underflow in this case? Isn't it just returning zero, which is totally fine for MC?

src/core/electrostatics/actor.hpp
```diff
@@ -34,6 +34,10 @@ void check_charge_neutrality(System::System const &system,
                              double relative_tolerance);
 
 template <typename Class> class Actor : public System::Leaf<Actor<Class>> {
+private:
+  Actor() = default;
+  friend Class;
```
reinaual, line 39: Why are all the `friend`s necessary?

maintainer/walberla_kernels/generate_lb_kernels.py
```diff
@@ -54,11 +39,27 @@
...
+import lbmpy.enums
+
+import lbmpy_walberla
+import lbmpy_espresso
+
+import lees_edwards
+import relaxation_rates
+import walberla_lbm_generation
+import code_generation_context
+import custom_additional_extensions
```
reinaual: Is there a benefit moving all the imports here?

maintainer/walberla_kernels/pystencils_espresso.py
```diff
@@ -88,11 +96,12 @@ def __init__(self, dim, time_step=ps.typing.TypedSymbol(
...
-            self,
-            "",
-            symbols_read=symbols_read,
-            symbols_defined=self.result_symbols)
+        if CustomCodeNode is not None:
+            CustomCodeNode.__init__(
+                self,
+                "",
+                symbols_read=symbols_read,
+                symbols_defined=self.result_symbols)
```
reinaual, line 104: Are we sure that the generated kernels for pystencils 2.0 are actually correct?

src/walberla_bridge/src/lattice_boltzmann/generated_kernels/PackInfoVecSinglePrecision.cpp
```diff
@@ -67,7 +67,6 @@ static FUNC_PREFIX void unpack_SW_BW_W_TW_NW_BS_S_TS_B_C_T_BN_N_TN_SE_BE_E_TE_NE
 } // namespace internal_1ccccad4ca561e07a0934cadb07d0fc1
 
 void PackInfoVecSinglePrecision::pack(Direction dir, unsigned char *byte_buffer, IBlock *block) const {
-  byte_buffer += sizeof(float) - (reinterpret_cast<std::size_t>(byte_buffer) - (reinterpret_cast<std::size_t>(byte_buffer) / sizeof(float)) * sizeof(float));
```
reinaual, line 70: Was this custom patched code? Why is it not necessary anymore?

src/walberla_bridge/src/electrokinetics/generated_kernels/DiffusiveFluxKernelWithElectrostaticThermalized_double_precision.cpp
```diff
@@ -56,99 +56,99 @@ static FUNC_PREFIX void diffusivefluxkernelwithelectrostaticthermalized_double_p
...
+          _data_j[_stride_j_0 + 12 * _stride_j_3] = D * (-f_ext_0 * z * _data_rho[_stride_rho_0] - f_ext_0 * z * _data_rho[_stride_rho_1 + _stride_rho_2] + f_ext_1 * z * _data_rho[_stride_rho_0] + f_ext_1 * z * _data_rho[_stride_rho_1 + _stride_rho_2] + f_ext_2 * z * _data_rho[_stride_rho_0] + f_ext_2 * z * _data_rho[_stride_rho_1 + _stride_rho_2] + kT * -2.0 * _data_rho[_stride_rho_1 + _stride_rho_2] + kT * 2.0 * _data_rho[_stride_rho_0] + z * _data_phi[_stride_phi_0] * _data_rho[_stride_rho_0] + z * _data_phi[_stride_phi_0] * _data_rho[_stride_rho_1 + _stride_rho_2] - z * _data_phi[_stride_phi_1 + _stride_phi_2] * _data_rho[_stride_rho_0] - z * _data_phi[_stride_phi_1 + _stride_phi_2] * _data_rho[_stride_rho_1 + _stride_rho_2]) * 0.04703213011469496 * ((1.0) / (kT)) + pow(D * (0.5 * _data_rho[_stride_rho_0] + 0.5 * _data_rho[_stride_rho_1 + _stride_rho_2]), 0.5) * (random_13_0 - 0.5) * 1.5025119784898082;
```
reinaual, line 85: I'm not sure about theses changes, if this new version or the previous version is actually correct. Apparently we lack test-coverage here, so this needs a proper investigation. I opened a ticket #5029 for it.

src/walberla_bridge/tests/LBWalberlaImpl_bspline_tests.cpp
jngrad: The Clang-Tidy check is overzealous and assumes that for-loops with a floating-point iterator are always a mistake. I also find the while-loop less aesthetically-pleasing, but it's better than explicitly disabling a security Clang-Tidy check. I looked online and it seems that for-loops are generally preferred when the number of iterations is known ahead of time (e.g. the number of elements in a STL container), whereas while-loop are generally preferred when the number of iterations is not well-defined (e.g. here, since the number of iterations of the first loop isn't the same depending on the CPU due to extended precision). Although nowadays for-loops are routinely used on iterator facades that have no well-defined number of elements...

src/walberla_bridge/src/lattice_boltzmann/LBWalberlaImpl.hpp
jngrad: Ah right, I still need to look into that one :sweat_smile:

src/core/electrostatics/elc.cpp
jngrad: Comment reproduced from the `z_energy()` method... I'll fix it in both methods.

src/core/reaction_methods/ReactionAlgorithm.cpp
jngrad: It's guaranteed to return zero, but it's also guaranteed to raise a CPU flag, which sends a hardware-defined signal when FPE instrumentation is enabled. There are still other sources of underflow in the codebase, but they are limited to a small set of python tests. I'm not sure it is worth addressing all of them, since they do not affect the numerical results. I addressed that one because it was a low-hanging fruit, and triggered FPE instrumentation in a dozen tests.

src/core/electrostatics/actor.hpp
jngrad: CRTP comes with many caveats: [`bugprone-crtp-constructor-accessibility`](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/crtp-constructor-accessibility.html). None of the proposed solutions are ideal. In this particular case, all methods are public, so using `friend` doesn't cause any harm.

maintainer/walberla_kernels/generate_lb_kernels.py
jngrad: For improved user experience, I would argue these imports should appear after
- `argparse.ArgumentParser()`, because we should be able to print the script command line usage via `--help` even if these modules are not available or do not work due to dependencies version mismatch,
- `assert packaging.specifiers.SpecifierSet(requirement).contains(module.__version__)`, so we can print a meaningful error message and exit, rather than have `import pystencils_espresso` throw an exception due to pystencils not being the correct version, and thus not having the same method names

maintainer/walberla_kernels/pystencils_espresso.py
jngrad: Unfortunately, this only offers partial support. If I remember correctly the LB boundary kernel didn't work yet, but the other LB kernels worked after a lot of tinkering. I only tested a few Python LB tests to check if the numerical precision issue in the LB collide kernel was fixed.

Below is the workflow I used back in November. It's broken again because the API is changing rapidly, but this PR solves most of the compatibility issues with 2.0-dev.
```sh
python3 -m venv codegen2
source codegen2/bin/activate
python3 -m pip install -c "$(git rev-parse --show-toplevel)/requirements.txt" \
    numpy cython jinja2 setuptools packaging
python3 -m pip install "git+file:///work/jgrad/walberla_deps/2.0/pystencils@v2.0-dev"
python3 -m pip install "git+file:///work/jgrad/walberla_deps/2.0/lbmpy@pystencils2.0-compat"
export PYTHONPATH="/work/jgrad/walberla_deps/2.0/walberla/python/"
cd $(git rev-parse --show-toplevel)/src/walberla_bridge/src/lattice_boltzmann/generated_kernels/
python3 $(git rev-parse --show-toplevel)/maintainer/walberla_kernels/generate_lb_kernels.py
```

src/walberla_bridge/src/lattice_boltzmann/generated_kernels/PackInfoVecSinglePrecision.cpp
jngrad: This patch was contributed upstream via [walberla/walberla!698](https://i10git.cs.fau.de/walberla/walberla/-/merge_requests/698).

src/walberla_bridge/src/lattice_boltzmann/LBWalberlaImpl.hpp
jngrad: Turns out there were two similarly named methods `ghost_communication_pdf()` and `ghost_communication_pdfs()`. I renamed the latter to `ghost_communication_full()`  to avoid confusion, and added a unit test for the former.

reinaual (approved):
Besides the EK-Stuff it LGTM

PR #5026, opened 2025-01-10
Allocating multiple blocks to one mpi rank in LB

hidekb:
Description of changes:

- LB CPU now supports allocating multiple blocks to one mpi rank
   - The default block number per mpi rank is 1
   - The block number per mpi rank is controlled by the argument  `blocks_per_mpi_rank` for `LBFluidWalberla`


maintainer/benchmarks/lb.py
```diff
@@ -50,6 +50,18 @@
 parser.add_argument("--output", metavar="FILEPATH", action="store",
                     type=str, required=False, default="benchmarks.csv",
                     help="Output file (default: benchmarks.csv)")
+parser.add_argument("--divided_block", action="store",
+                    type=int, default=1, required=False,
```
RudolfWeeber: Maybe sth like blocks_per_mpi_rank would be a more clear name for this?

src/script_interface/walberla/LBFluid.cpp
```diff
@@ -139,6 +139,12 @@ void LBFluidGPU::make_instance(VariantMap const &params) {
   auto const visc = get_value<double>(params, "kinematic_viscosity");
   auto const dens = get_value<double>(params, "density");
   auto const precision = get_value<bool>(params, "single_precision");
+  auto const blocks_per_mpi_rank = get_value_or<Utils::Vector3i>(
+      params, "blocks_per_mpi_rank", Utils::Vector3i{{1, 1, 1}});
+  if (blocks_per_mpi_rank != Utils::Vector3i{{1, 1, 1}}) {
+    throw std::runtime_error(
+        "GPU architecture PROHIBITED allocating many blocks to 1 CPU.");
+  }
```
RudolfWeeber, line 147: "Using more than one block per MPI rank is not supported for GPU LB" (but why, actually?)

src/walberla_bridge/src/BoundaryPackInfo.hpp
```diff
@@ -96,7 +96,7 @@ class BoundaryPackInfo : public PackInfo<GhostLayerField_T> {
     WALBERLA_ASSERT_EQUAL(bSize, buf_size);
 #endif
 
-    auto const offset = std::get<0>(m_lattice->get_local_grid_range());
+    auto const offset = to_vector3i(receiver->getAABB().min());
```
RudolfWeeber: Wouldn't it be better to have functions for this in the Lattice class, so they can be used by EK as well?
After all, LB and EK probably need to agree about both, the mpi and the block decompositoin?

src/walberla_bridge/src/LatticeWalberla.cpp
```diff
@@ -50,21 +51,28 @@ LatticeWalberla::LatticeWalberla(Utils::Vector3i const &grid_dimensions,
...
           "Lattice grid dimensions and MPI node grid are not compatible.");
     }
+    if (m_grid_dimensions[i] % block_grid[i] != 0) {
+      throw std::runtime_error(
+          "Lattice grid dimensions and block grid are not compatible.");
+    }
   }
 
   auto constexpr lattice_constant = real_t{1};
-  auto const cells_block = Utils::hadamard_division(grid_dimensions, node_grid);
```
RudolfWeeber, line 56: cells_per_block?

src/walberla_bridge/src/LatticeWalberla.cpp
```diff
@@ -50,21 +51,28 @@ LatticeWalberla::LatticeWalberla(Utils::Vector3i const &grid_dimensions,
...
+      uint_c(block_grid[0]), uint_c(block_grid[1]), uint_c(block_grid[2]),
       // number of cells per block in each direction
       uint_c(cells_block[0]), uint_c(cells_block[1]), uint_c(cells_block[2]),
       lattice_constant,
       // number of cpus per direction
       uint_c(node_grid[0]), uint_c(node_grid[1]), uint_c(node_grid[2]),
       // periodicity
-      true, true, true);
+      true, true, true,
+      // keep global block information
```
RudolfWeeber, line 75: What does this do/mean?

src/walberla_bridge/src/LatticeWalberla.cpp
```diff
@@ -73,11 +81,44 @@ LatticeWalberla::LatticeWalberla(Utils::Vector3i const &grid_dimensions,
 [[nodiscard]] std::pair<Utils::Vector3d, Utils::Vector3d>
```
RudolfWeeber, line 82: Are we surr that there is no funciot in block forest or uniform blcok forest which supplies the answer to this?
If not, and we really have to calculate htis ourselves, it could probably be a bit more compact. Chatgpt came up with:
```
    // Initialize bounds to extreme values
    std::array<double, 3> globalMin = {std::numeric_limits<double>::max(),
                                       std::numeric_limits<double>::max(),
                                       std::numeric_limits<double>::max()};
    std::array<double, 3> globalMax = {std::numeric_limits<double>::lowest(),
                                       std::numeric_limits<double>::lowest(),
                                       std::numeric_limits<double>::lowest()};

    for (const auto& block : blocks) {
        const auto& blockMin = block.min();
        const auto& blockMax = block.max();
        for (size_t i = 0; i < 3; ++i) {
            globalMin[i] = std::min(globalMin[i], blockMin[i]);
            globalMax[i] = std::max(globalMax[i], blockMax[i]);
        }
    }
```

src/walberla_bridge/src/lattice_boltzmann/LBWalberlaImpl.hpp
```diff
@@ -389,8 +390,56 @@ class LBWalberlaImpl : public LBWalberlaBase {
...
+      }
+    }
+    auto block_upper_corner = to_vector3i(block.getAABB().max());
+    if (lower_corner[0] > block_upper_corner[0] or
+        lower_corner[1] > block_upper_corner[1] or
+        lower_corner[2] > block_upper_corner[2]) {
+      return std::nullopt;
+    }
+    for (uint_t f = 0u; f < 3u; ++f) {
+      if (block_upper_corner[f] > upper_corner[f]) {
```
RudolfWeeber: std::max() ?

src/walberla_bridge/src/lattice_boltzmann/LBWalberlaImpl.hpp
```diff
@@ -868,40 +918,62 @@ class LBWalberlaImpl : public LBWalberlaBase {
...
+          auto const upper_cell = bci->max();
+          // The field data "values" knows about block-local indices
+          // In the loop, x,y,z are in block-local coordinates
+          // It is converted to BlockForest (lattice) coordinates assigned to a
+          // mpi rank The same applies to other get_slice methods
+          for (auto x = lower_cell.x(); x <= upper_cell.x(); ++x) {
+            for (auto y = lower_cell.y(); y <= upper_cell.y(); ++y) {
+              for (auto z = lower_cell.z(); z <= upper_cell.z(); ++z) {
+                auto const node = local_offset + Utils::Vector3i{{x, y, z}};
+                auto const index = stride_x * (node[0] - lower_corner[0]) +
```
RudolfWeeber: Can these index calculations be moved out to separate functions? It looks relateed to Utils::get_linear_index()?

src/walberla_bridge/src/lattice_boltzmann/LBWalberlaImpl.hpp
```diff
@@ -912,17 +984,55 @@ class LBWalberlaImpl : public LBWalberlaBase {
...
-      auto force_field =
-          block.template getData<VectorField>(m_last_applied_force_field_id);
-      auto vel_field = block.template getData<VectorField>(m_velocity_field_id);
-      assert(++(lattice.get_blocks()->begin()) == lattice.get_blocks()->end());
       assert(velocity.size() == 3u * ci->numCells());
-      std::vector<FloatType> const values(velocity.begin(), velocity.end());
-      lbm::accessor::Velocity::set(pdf_field, vel_field, force_field, values,
-                                   *ci);
+      int64_t const stride_y = (ci->max().z() - ci->min().z() + 1u);
+      int64_t const stride_x = (ci->max().y() - ci->min().y() + 1u) * stride_y;
```
RudolfWeeber: Pls document wht these strides are for

RudolfWeeber: Also, these stride calculation seem to appear several times in the cell interval code. Can that be ousourced to a function which is then re-used?

testsuite/python/lb_couette_xy.py
```diff
@@ -0,0 +1,147 @@
...
+
+    """
+    u = x / h - 0.5
+    for k in np.arange(1, k_max + 1):
+        wave = 2 * np.pi * k / h
+        u += np.exp(-nu * wave ** 2 * t) * np.sin(wave * x) / (np.pi * k)
+    return v * u
+
+
+LB_PARAMS = {'agrid': 1.,
```
RudolfWeeber: Pls avoid 1s in unit tests, as wrong exponents don't get cauth.

RudolfWeeber:
Thank you. Looks good in general.
One thing worth looking into is the cell interval business. It looks like there are a lot of very similar loops in the getters and setters for slices. Would it it be possible to pull that out into a function, which then is called with different lambdas for the individual cases?
@jngrad could you maybe take a look?

src/walberla_bridge/src/LatticeWalberla.cpp
hidekb: If "keep global block information" is true, each process keeps information about remote blocks that reside on other processes.

maintainer/benchmarks/lb.py
hidekb: I changed the argument name to blocks_per_mpi_rank.

src/walberla_bridge/src/LatticeWalberla.cpp
hidekb: I changed cells_block to cells_per_block.

src/walberla_bridge/src/LatticeWalberla.cpp
hidekb: In my understanding, a BlockForest knows the domain size for the whole space, and blocks know their domain range for the entire space. Hence, there is no class that knows the domain range in 1 CPU. Then, I rewrote the code to make it compact.

src/walberla_bridge/src/lattice_boltzmann/LBWalberlaImpl.hpp
hidekb: I rewrote the code using std::max() and std::min().

src/walberla_bridge/src/lattice_boltzmann/LBWalberlaImpl.hpp
hidekb: I rewrote the code, obtaining a linear index using Utils::get_linear_index() and refactoring the code related to set_slice and get_slice.

src/walberla_bridge/src/lattice_boltzmann/LBWalberlaImpl.hpp
hidekb: By refactoring the code, stride is not used now.

hidekb: I rewrote the code following your suggestions. Thus, the code related to set_slice and get_slice became more compact.

testsuite/python/lb_couette_xy.py
hidekb: I moved this function.

maintainer/benchmarks/lb.py
```diff
@@ -105,6 +108,8 @@
 print(f"LB shape: {lb_grid}")
 print(f"LB agrid: {agrid:.3f}")
 
+blocks_per_mpi_rank = args.blocks_per_mpi_rank
+
```
jngrad: This alias is not really needed, you can use `args.blocks_per_mpi_rank` directly at line 153.

maintainer/benchmarks/lb_weakscaling.py
```diff
@@ -0,0 +1,166 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+"""
+Benchmark Lattice-Boltzmann fluid + Lennard-Jones particles.
+"""
```
jngrad: Wouldn't it be more sustainable if we only had one LB benchmark file to maintain? argparse is quite flexible, surely we can come up with a way to select strong vs. weak scaling with command line options?

src/python/espressomd/lb.py
```diff
@@ -141,6 +141,8 @@ class LBFluidWalberla(HydrodynamicInteraction,
         Required for a thermalized fluid. Must be positive.
     single_precision : :obj:`bool`, optional
         Use single-precision floating-point arithmetic.
+    blocks_per_mpi_rank : (3,) array_like of :obj:`int`, optional
+        Ditribute more than one block to each CPU.
```
jngrad: typo

src/script_interface/walberla/LBFluid.cpp
```diff
@@ -139,6 +139,12 @@ void LBFluidGPU::make_instance(VariantMap const &params) {
   auto const visc = get_value<double>(params, "kinematic_viscosity");
   auto const dens = get_value<double>(params, "density");
   auto const precision = get_value<bool>(params, "single_precision");
+  auto const blocks_per_mpi_rank = get_value_or<Utils::Vector3i>(
+      params, "blocks_per_mpi_rank", Utils::Vector3i{{1, 1, 1}});
```
jngrad: Here and elsewhere, do we really need a default value, considering we already provide a default value in the python class?

src/script_interface/walberla/LBFluid.cpp
jngrad: how about "GPU LB only uses 1 block per MPI rank"?

src/script_interface/walberla/LatticeWalberla.hpp
```diff
@@ -53,14 +54,25 @@ class LatticeWalberla : public AutoParameters<LatticeWalberla> {
...
+    m_blocks_per_mpi_rank = get_value_or<Utils::Vector3i>(
+        args, "blocks_per_mpi_rank", Utils::Vector3i{{1, 1, 1}});
     auto const n_ghost_layers = get_value<int>(args, "n_ghost_layers");
+    auto const block_grid =
+        Utils::Vector3i{{static_cast<int>(::communicator.node_grid[0] *
+                                          m_blocks_per_mpi_rank[0]),
+                         static_cast<int>(::communicator.node_grid[1] *
+                                          m_blocks_per_mpi_rank[1]),
+                         static_cast<int>(::communicator.node_grid[2] *
+                                          m_blocks_per_mpi_rank[2])}};
```
jngrad: `Utils::hadamard_product()`?

src/walberla_bridge/src/utils/types_conversion.hpp
```diff
@@ -68,6 +68,10 @@ inline Utils::VectorXd<9> to_vector9d(Matrix3<float> const &m) {
           double_c(m[3]), double_c(m[4]), double_c(m[5]),
           double_c(m[6]), double_c(m[7]), double_c(m[8])};
 }
+inline Utils::Vector3i to_vector3i(Vector3<float> const &v) {
+  return Utils::Vector3i{
+      {static_cast<int>(v[0]), static_cast<int>(v[1]), static_cast<int>(v[2])}};
```
jngrad: Are these static casts safe? For example, what is the output when the input vector is `v = {1.f, 2.f, 6.99999f}`? What happens if a vector of doubles is passed? Should we warn the user through assertions if the input floats are not "close enough" to round numbers?

testsuite/python/lb.py
```diff
@@ -517,7 +517,11 @@ def test_agrid_rounding(self):
         phi = 0.05
         lj_sig = 1.0
         l = (n_part * 4. / 3. * np.pi * (lj_sig / 2.)**3 / phi)**(1. / 3.)
-        system.box_l = [l] * 3 * np.array(system.cell_system.node_grid)
+        if hasattr(self, 'blocks_per_mpi_rank'):
+            system.box_l = [
+                l] * 3 * np.array(system.cell_system.node_grid) * np.array(self.blocks_per_mpi_rank)
+        else:
+            system.box_l = [l] * 3 * np.array(system.cell_system.node_grid)
```
jngrad: Can one replace `[l] * 3` by `l`?

testsuite/python/lb_couette_xy.py
```diff
@@ -0,0 +1,145 @@
+#
+# Copyright (C) 2021-2023 The ESPResSo project
```
jngrad: Year range should probably include 2025.

testsuite/python/lb_mass_conservation.py
```diff
@@ -41,7 +41,7 @@ class LBMassCommon:
 
     """Check the lattice-Boltzmann mass conservation."""
 
-    system = espressomd.System(box_l=[3.0, 3.0, 3.0])
+    system = espressomd.System(box_l=[6.0, 6.0, 6.0])
```
jngrad: Several mass conservation tests are know to take a lot of time, especially in code coverage builds. The runtime can significantly increase when multiple CI jobs run on the same runner, which is then starved of resources. This change may potentially increase the test runtime by a factor 8. Can you please confirm the runtime did not significantly change in the clang and coverage CI jobs, compared to the python branch?

hidekb:
> Thank you. Looks good in general. One thing worth looking into is the cell interval business. It looks like there are a lot of very similar loops in the getters and setters for slices. Would it it be possible to pull that out into a function, which then is called with different lambdas for the individual cases? 

I rewrote the code related to the getters and setters for slices. Similar loops are pulled into a function which calls different lambdas for the individual cases.

testsuite/python/lb_shear.py
```diff
@@ -31,7 +31,7 @@
 # Box size will be H +2 AGRID to make room for walls.
 # The number of grid cells should be divisible by four and 3 in all directions
 # for testing on multiple mpi nodes.
-H = 12 * AGRID
+H = 10 * AGRID
```
jngrad: Why change the box height?

maintainer/benchmarks/lb.py
hidekb: I have rewritten it as per your comment.

maintainer/benchmarks/lb_weakscaling.py
hidekb: I unified lb.py and lb_weakscaling.py by adding option --weak_scaling to argparse for lb.py

src/python/espressomd/lb.py
hidekb: I have rewritten it.

src/script_interface/walberla/LBFluid.cpp
hidekb: I added default value for the python class `LatticeWalberla`. And `blocks_per_mpi_rank` in `LBFluid` is settle by `get_value<Utils::Vector3i>(m_lattice->get_parameter("blocks_per_mpi_rank"))`.

src/script_interface/walberla/LatticeWalberla.hpp
hidekb: I have rewritten it, as you suggest.

testsuite/python/lb.py
hidekb: I have rewritten it, as you suggest.

testsuite/python/lb_couette_xy.py
hidekb: I changed it.

testsuite/python/lb_shear.py
hidekb: In the present test script, we don't need this modification anymore, so I am going to put it back.

src/walberla_bridge/src/lattice_boltzmann/LBWalberlaImpl.hpp
```diff
@@ -389,8 +391,52 @@ class LBWalberlaImpl : public LBWalberlaBase {
...
+  // Interval within local block
+  [[nodiscard]] std::optional<CellInterval> get_block_interval(
+      Utils::Vector3i const &lower_corner, Utils::Vector3i const &upper_corner,
+      Utils::Vector3i const &block_offset, IBlock const &block) const {
+    auto block_lower_corner = to_vector3i(block.getAABB().min());
+    if (upper_corner[0] < block_lower_corner[0] or
+        upper_corner[1] < block_lower_corner[1] or
+        upper_corner[2] < block_lower_corner[2]) {
+      return std::nullopt;
+    }
```
jngrad: In the past, ESPResSo was affected by subtle bugs where indices in the compared vectors wouldn't be the same on the left-hand-side and right-and-side of the comparison operator. To avoid this type of mistake, one can write here and elsewhere `if (not(block_lower_corner > upper_corner))`. Comparison operators on `Utils::Vector` objects will internally call `Utils::detail::all_of(lhs, rhs, cmp)`, which is why sometimes the lhs and rhs have to be exchanged.

src/walberla_bridge/src/lattice_boltzmann/LBWalberlaImpl.hpp
```diff
@@ -389,8 +391,52 @@ class LBWalberlaImpl : public LBWalberlaBase {
...
+    block_upper_corner -= Utils::Vector3i::broadcast(1);
+    Cell const block_lower_cell =
+        Cell(static_cast<int>(block_lower_corner[0] - block_offset[0]),
+             static_cast<int>(block_lower_corner[1] - block_offset[1]),
+             static_cast<int>(block_lower_corner[2] - block_offset[2]));
+    Cell const block_upper_cell =
+        Cell(static_cast<int>(block_upper_corner[0] - block_offset[0]),
+             static_cast<int>(block_upper_corner[1] - block_offset[1]),
+             static_cast<int>(block_upper_corner[2] - block_offset[2]));
+    return {CellInterval(block_lower_cell, block_upper_cell)};
```
jngrad: * `Cell const` -> `auto const` since the rhs already indicates the type.
* are the static casts necessary? the result of the operations seem to already be `int`
* would `return {{block_lower_cell, block_upper_cell}};` work here? I never remember the exact rule, but I think the outermost braces should invoke the optional constructor, and the innermost braces should invoke the cell interval constructor. For initializer lists and pointer-valued element different rules would apply.

testsuite/python/lb_mass_conservation.py
hidekb: I changed the box_l from 6 to 4 to reduce the runtime. For testing with blocks_per_mpir_rank(i.e. [1,1,2]), box_l = 4 is needed.

src/walberla_bridge/src/utils/types_conversion.hpp
hidekb: I added the function for a vector of doubles and assertions if the difference between input values and round numbers is larger than  10^(-5), which is also debatable.

src/walberla_bridge/src/lattice_boltzmann/LBWalberlaImpl.hpp
hidekb: I have rewritten it, as you suggest.

src/walberla_bridge/src/utils/types_conversion.hpp
hidekb: I removed the free function `to_vector3i()` and added the member function of LatticeWalberla `LatticeWalberla::get_block_corner(IBlock &block)`. This change has reduced the scope for using integerisation functions compared to before.

src/walberla_bridge/src/lattice_boltzmann/LBWalberlaImpl.hpp
hidekb: I have rewritten `Cell const` to `auto const` and removed static_cast. And I have removed Braces and wrote `return CellInterval(block_lower_cell, block_upper_cell);`

src/walberla_bridge/src/BoundaryPackInfo.hpp
hidekb: I removed the free function `to_vector3i` and added the member function of LatticeWalberla `LatticeWalberla::get_block_corner`.

testsuite/python/lb_mass_conservation.py
hidekb: The total runtime for the clang CI jobs in this branch is 291.31 sec, compared to 271.33 sec in the python branch, which increases by around 7%.
The total runtime for the coverage CI jobs in this branch is 338.22 sec, compared to 320.22 sec in the python branch, which increases by around 6%.
These amounts cannot be ignored but are not so significant.

src/script_interface/walberla/LBFluid.cpp
hidekb: Following Rudolf's suggestion, I rewrote it to "Using more than one block per MPI rank is not supported for GPU LB."

src/script_interface/walberla/LBFluid.cpp
hidekb: Therefore, we don't need to provide a default value in C++ code, so I changed `get_value_or` to `get_value`.

testsuite/python/lb_couette_xy.py
hidekb: I changed the value of agrid from 1. to 0.6.

jngrad (approved):
Thanks a lot!

PR #5031, opened 2025-01-23
Encapsulate HDF5 module

jngrad:
Description of changes:
- bump version requirements of h5xx
- source h5xx via FetchContent instead of submodules
   - the ESPResSo repository can now be cloned without `--recursive`
   - the creation of GitHub releases no longer depends on third-party tools and can be fully automated
- hide hdf5 and h5xx header files and compiler flags from consumer targets
- fix portability issues revealed by newer compiler toolchains

### DIFF
--- a/src/core/io/writer/h5md_core.cpp
+++ b/src/core/io/writer/h5md_core.cpp
@@ -30,7 +30,10 @@
 
 #include <utils/Vector.hpp>
 
+#include <boost/array.hpp>
+#include <boost/filesystem.hpp>
 #include <boost/mpi/collectives.hpp>
+#include <boost/multi_array.hpp>
 
 #include <mpi.h>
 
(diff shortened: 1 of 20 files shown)


src/core/io/writer/h5md_core.cpp
```diff
@@ -215,34 +230,36 @@ static void write_attributes(h5xx::file &h5md_file) {
 }
 
 void File::write_units() {
+  auto const &datasets = *m_datasets;
   if (!mass_unit().empty() and (m_fields & H5MD_OUT_MASS)) {
-    h5xx::write_attribute(datasets["particles/atoms/mass/value"], "unit",
+    h5xx::write_attribute(datasets.at("particles/atoms/mass/value"), "unit",
```
reinaual, line 235: Is there a benefit to making it a bound-checked access?

jngrad: Muscle memory. I didn't think too much about it.

Marginal benefit: if the key doesn't exist, a default-constructed hdf5 dataset is created and returned by reference by `operator[]`. Your file then contains a dataset with no elements. Since this method is called only once per simulation, at file creation, overhead of exceptions isn't relevant, and is probably smaller than default-constructing an empty dataset. Plus, we get notified that the map isn't properly initialized.

Whether we need bound-checked accesses in the hot loop (`File::write()`) is debatable. The h5xx module will be removed from ESPResSo eventually, in favor of a yet-to-be-determined C++ wrapper library for hdf5. Whether that new library will offer bound-checked access is likely, but not guaranteed.

reinaual (approved):

hidekb:
LGTM!

hidekb (approved):
LGTM!

PR #5032, opened 2025-01-24
add an observable for pairwise distance and an accumulator for contact times

pm-blanco:
Description of changes:
- `observables.PairwiseDistances`: an observable that tracks pairwise distances between two sets of particles
- `accumulators.ContactTimes`: an accumulator that tracks contact times when particles are within some distance
- `samples/contact_time.py`: a sample script showcasing how to use these new features

The contact time is defined as:

$$\tau = t_f - t_0$$

where  $t_f$ is the last measured time at which those particles were in contact and  $t_0$ the first measured time at which those particles were in contact. For example, if two particles have been only in contact during two consecutive configurations then their  contact time is $\tau = \Delta t$, where $\Delta t$ is the time step. If they are only in contact for a single configuration then  their contact time is $\tau = 0$.


RudolfWeeber:
Thank you Pablo. This is a very useful addition.
IMO, we should try to split this into 
* an observable `PairwiseDistance(pids1,pids2)` 
* and an accumulator `ParticlecontactTimes(pairwise_distance_obs, contact_distance)`

In my impression, this would allow us to leave the observabl genuinely stateless (and avoid cirumenting teh them being const). All the state would be in the accumulator.

Should we do a short video call, to discuss?

pm-blanco:
Hi Rudolf! The solution you propose makes sense to me, the "hack" I had to do felt a bit against the current phylosophy of the observable/accumulator framework. I am simply not too familiar with the c++ core of the accumulators/observables so I did not want to tinker with it too much. I think I got myself somewhat familiar now with the observable framework, so I can try to refactor  the feature to match what you propose.

Regarding the videocall, sounds good. I will send you an e-mail so we can schedule a time :smile:

pm-blanco:
@jngrad I have the PR mostly ready for review, but my test is skipped by the CI for some reason.  Do you know why?
```
The following tests did not run:
	 30 - contact_time (Skipped)
```

jngrad:
> @jngrad I have the PR mostly ready for review, but my test is skipped by the CI for some reason. Do you know why?
> 
> ```
> The following tests did not run:
> 	 30 - contact_time (Skipped)
> ```

The class is missing a `test_something(self)` method to check the result of the simulation.

testsuite/python/contact_time.py
```diff
@@ -0,0 +1,221 @@
...
+        obs = espressomd.observables.PairwiseDistances(ids=[1], 
+                                                       target_ids=[1])
+
+        with self.assertRaisesRegex(RuntimeError, "The contact threshold must be a positive number."):
+            espressomd.accumulators.ContactTimes(obs=obs, 
+                                                 delta_N=1,
+                                                 contact_threshold=-1)
+
+    if __name__ == "__main__":
+        ut.main()
```
jngrad: Remove indentation.

jngrad:
@pm-blanco I think something went wrong during the rebase. Could you please try this?
```sh
git reset --soft upstream/python # rewind changes accidentally taken from the python branch
git commit -m 'Add contact time observable'
git show # here carefully inspect the diff to make sure it only contains your feature
git push -f
```
If this doesn't help, use `git reflog` to look through the history of orphaned commits and resurrect e.g. `a38417ee4`.

pm-blanco:
@jngrad I have fixed the broken rebase, sorry for the mess! The PR is now ready for review.

jngrad:
Could you please address the Clang compiler errors too? Thanks!

pm-blanco:
@jngrad the clang compilation errors should be solved now, thanks for the help!

samples/contact_time.py
```diff
@@ -76,4 +66,6 @@
 # Get the contact time calculated on the fly
 contact_times = contact_time_accumulator.contact_times()
 
-print(f"The contact times are {contact_times}")
+print("The contact times are:")
+for pid, tau in zip(ids, contact_times):
+    print(f"  particle {pid}: {tau:.3f}")
```
pm-blanco: I find this  misleading. `contact_times`  tracks all the contact times measured in the simulation so it does not make sense to link it to any specific particle id.

jngrad: My bad, I'll fix it.

testsuite/python/contact_time.py
```diff
@@ -0,0 +1,233 @@
...
+            dist2 += dist**2
+        return np.sqrt(dist2)
+
+    @classmethod
+    def calculate_contact_times(
+            cls, system, accumulator_pos, contact_threshold):
+        """
+        Calculate the contact times from the time series of the coordinates
+        stored in ``accumulator_ids1`` and ``accumulator_ids2``.
+
```
pm-blanco: Note to self: I need to fix this broken docs.

jngrad: My plan is to merge this ASAP. Do you need more time to review my patch?

pm-blanco: I quickly went through your patch and it look fine to me. The only two issues  I spotted  these two.  If you can fix the docs here then I think you can proceed to merge it on my side!

jngrad (approved):

PR #5034, opened 2025-01-31
Support modern compilers, Python, Jupyter and NumPy releases

jngrad:
Description of changes:
- bump all compiler version requirements
   - C++20 core features are now almost fully supported
- remove now obsolete re-implementations of C++20 algorithms
- massively reduce the complexity of the CMake build system for tutorials
- fix deprecation warnings from recent versions of NumPy, SciPy, JupyterLab, and the Python standard library
- improve code coverage and remove unreachable code


testsuite/python/tests_common.py
```diff
@@ -322,11 +322,12 @@ def lj_generic_force(espressomd, r, epsilon, sigma, cutoff, offset=0., e1=12,
...
     else:
+        has_ljgen_softcore = espressomd.has_features("LJGEN_SOFTCORE")
         h = (r - offset)**2 + delta * (1. - lam) * sigma**2
         f = (r - offset) * epsilon * lam * (
-            b1 * e1 * np.power(sigma / np.sqrt(h), e1) - b2 * e2 * np.power(sigma / np.sqrt(h), e2)) / h
-        if (not espressomd.has_features("LJGEN_SOFTCORE")) and generic:
-            f *= np.sign(r - offset)
+            b1 * e1 * np.power(sigma / np.sqrt(h), e1) -
+            b2 * e2 * np.power(sigma / np.sqrt(h), e2)) / h
+        f *= np.sign(r - offset) if generic and not has_ljgen_softcore else 1
```
reinaual: What is the benefit in using a ternary operator here? If this line wasn't reached in coverage before, I'm not in favor of "green washing" this line with this.

testsuite/python/collision_detection.py
```diff
@@ -186,19 +186,8 @@ def verify_state_after_bind_at_poc_triplet(self, expected_np):
...
-                continue
-            # Skip vs that doesn't have a bond
-            if p.bonds == ():
-                continue
-            if p.id == 1:
-                # Parse the bond
-                self.assertEqual(len(p.bonds), 2)
-                # Bond type
-                self.assertEqual(p.bonds[0][0], self.bond_pair)
-                self.assertEqual(p.bonds[1][0], self.bond_vs)
```
reinaual, line 200: For my understanding, it means that this branch was never reached in CI?

doc/sphinx/installation.rst
```diff
@@ -250,15 +250,14 @@ to use, we recommend installing JupyterLab:
 
 .. code-block:: bash
 
-    python3 -m pip install --user -c requirements.txt \
-        nbformat nbconvert jupyterlab
+    python3 -m pip install -c requirements.txt \
+        jupyterlab>=4.3 nbformat nbconvert lxml[html_clean]
```
reinaual, line 254: What does this `lxml`-package do?

src/utils/include/utils/mask.hpp
```diff
@@ -44,24 +45,18 @@ auto mask_impl(Integral mask, T t, std::index_sequence<I...>) {
  * instance of the type).
  *
  * Example:
- *   mask(0b1011, {1, 2, 3, 4}) => {1, 0, 3, 4}
+ *   <tt>mask(0b1011u, {1, 2, 3, 4}) => {1, 2, 0, 4}</tt>
```
reinaual, line 48: Why this change? If this is true, I don't understand the masking.

doc/sphinx/Readme.rst
```diff
@@ -5,7 +5,8 @@ How to get started with Sphinx
 
    .. code-block:: bash
 
-       pip3 install --user --upgrade 'sphinx>=2.3.0,!=3.0.0' 'sphinxcontrib-bibtex>=0.3.5'
+       python3 -m pip install -c requirements.txt \
+         sphinx sphinx-toggleprompt sphinxcontrib-bibtex numpydoc pybtex
```
reinaual, line 9: What do we need `pybtex` for?

src/core/particle_node.cpp
```diff
@@ -231,29 +231,21 @@ static std::vector<Particle> mpi_get_particles(std::span<const int> ids) {
...
+    assert(ignore.empty());
   }
 
-  /* Copy local particles */
-  std::transform(node_ids[this_node].cbegin(), node_ids[this_node].cend(),
-                 parts.begin(), [](int p_id) {
-                   auto const p = get_cell_structure().get_local_particle(p_id);
-                   assert(p != nullptr);
-                   return *p;
-                 });
```
reinaual, line 251: Does this mean that we don't have particles of the head node in the cache after the call to `prefetch_particle_data`?

testsuite/python/collision_detection.py
jngrad: indeed, this is a leftover from the bond angle collision protocol that was removed last year

doc/sphinx/installation.rst
jngrad: Parser for XML and HTML. It's a hard dependency of JupyterLab >= 4, but the `html_clean` component is a soft dependency that doesn't get installed by default. We need it to convert Jupyter notebooks into HTML files in CI. We can use the syntax `lxml[html_clean]` inside `requirements.txt`, but not when using the `-c` flag, since square brackets are not allowed in a constraint file (it was in earlier versions of pip, but not anymore).

src/utils/include/utils/mask.hpp
jngrad: Proof by computation:

```c++
#include <cstdio>
#include <concepts>
#include <cstddef>
#include <limits>
#include <type_traits>
#include <tuple>
#include <utility>

namespace Utils {
namespace detail {
template <class T, std::size_t... I>
auto mask_impl(std::unsigned_integral auto mask, T const &t,
               std::index_sequence<I...>) {
  return T{((mask & (1u << I)) ? std::get<I>(t) : std::tuple_element_t<I, T>{})...};
}
} // namespace detail

template <class T> T mask(std::unsigned_integral auto mask, T const &t) {
  auto constexpr size_in_bits = std::numeric_limits<decltype(mask)>::digits;
  static_assert(size_in_bits >= std::tuple_size_v<T>);
  return detail::mask_impl(mask, t,
                           std::make_index_sequence<std::tuple_size_v<T>>{});
}
} // namespace Utils

int main() {
    auto values = std::tuple<int, int, int, int>{1, 2, 3, 4};
    auto mask = 0b1011u;
    auto out = Utils::mask(mask, values);
    printf("{%i, %i, %i, %i}\n", std::get<0>(out), std::get<1>(out), std::get<2>(out), std::get<3>(out));
}
```
Output:
```
{1, 2, 0, 4}
```

The least significant bit appears on the right, the most significant bit appears on the left. The `1` next to `u` refers to the 0th index in the tuple-like sequence, the `1` next to `b` refers to the n-th index.

reinaual: Oh, I also fell for it. Of course the least significant bit is on the right...

doc/sphinx/Readme.rst
jngrad: https://github.com/espressomd/espresso/blob/adf636d9916a19485604c376d925d7849fba3076/doc/sphinx/conf.py.in#L279-L282

src/core/particle_node.cpp
jngrad: To my understanding, the cache only holds copies of particles from worker nodes, because particles on the head node are always immediately accessible. In fact, there is a pre-filtering of the user input to remove particles who have a valid pointer on rank 0, so this transform was always a no-op. In fact, when a simulation only has 1 MPI rank, the function is never called:
https://github.com/espressomd/espresso/blob/adf636d9916a19485604c376d925d7849fba3076/src/core/particle_node.cpp#L264-L268

testsuite/python/tests_common.py
jngrad: Looking at this file again, something about this function feels a bit weird. It's supposed to be the derivative of the corresponding energy function immediately above, but the latter seems to hardcode `generic=True`. That's quite dangerous, because the testsuite doesn't always check if `LJGEN_SOFTCORE` is present when `LENNARD_JONES_GENERIC` is required.

jngrad: After some digging, the function seems to be correct. I've reverted this change and added an extra test to get real coverage.

This investigation helped find new was to improve the script interface. For some reason the bonded interactions map was maintaining two `std::unordered_map` and was only using one of them as a stop-gap solution to avoid a nasty MPI deadlock due to `std::shared_ptr` lifetime being different on the head node and worker nodes.

testsuite/python/lb_pressure_tensor.py
```diff
@@ -19,10 +19,11 @@
 import unittest as ut
 import unittest_decorators as utx
 import numpy as np
+# import scipy.optimize
+# import scipy.integrate
```
reinaual, line 23: What are with the imports?

jngrad: They are required for a test case at the bottom of the file which is currently commented out.

reinaual (approved):

PR #5043, opened 2025-02-18
Splitting kinetic energy into contributions

davidbbeyer:
Fixes #1400 

Description of changes:
- Splits the kinetic energy returned by `espressomd.analyze.Analysis.energy()` into contributions due to translation and rotation

PR #5045, opened 2025-02-18
Add DPD contribution to the pressure

### LINKED ISSUE #5030: DPD pressure not part of total pressure
which is pretty confusing. DPD stress has to be calculated separately using the DPDStress observable.

mebrito:
Fixes #5030

### DIFF
--- a/src/python/espressomd/analyze.py
+++ b/src/python/espressomd/analyze.py
@@ -652,6 +652,8 @@ class Analysis(ScriptInterfaceHelper):
             reduction(out, "dipolar")
         if has_features("VIRTUAL_SITES"):
             reduction(out, "virtual_sites")
+        if has_features("DPD"):
+            reduction(out, "dpd")
 
         if dim == 1 or calc_sp:
             return out
--- a/src/core/Observable_stat.cpp
+++ b/src/core/Observable_stat.cpp
@@ -46,6 +46,11 @@ Observable_stat::Observable_stat(std::size_t chunk_size, std::size_t n_bonded,
 #else
   constexpr std::size_t n_vs = 0;
 #endif
+#ifdef DPD
+  constexpr std::size_t n_dpd = 1;
+#else
+  constexpr std::size_t n_dpd = 0;
+#endif
   auto const n_non_bonded = get_non_bonded_offset(max_type, max_type) + 1ul;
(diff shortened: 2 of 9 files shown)


src/python/espressomd/analyze.py
```diff
@@ -492,7 +492,11 @@ def energy(self):
 
         """
         observable = self.call_method("calculate_energy")
-        return self._generate_summary(observable, 1, False)
+        observable = self._generate_summary(observable, 1, False)
+        if has_features("DPD"):
+            # Remove meaningless (key, value)
+            del observable["dpd"]
+        return observable
```
jngrad: Wouldn't it be easier to make DPD stress part of the pressure calculation in the core, like in #5043? With the currently proposed changes, ESPResSo scripts written in C++ and Python disagree about the total pressure.

testsuite/python/dpd_press_contrib.py
```diff
@@ -0,0 +1,103 @@
...
+        self.assertNotIn("dpd", pressure_dict,
+                         "'dpd' key should not be in the pressure dictionary")
+
+        pressure_tensor_dict = self.system.analysis.pressure_tensor()
+        self.assertNotIn("dpd", pressure_tensor_dict,
+                         "'dpd' key should not be in the pressure-tensor dictionary")
+
+
+@utx.skipIfMissingFeatures("DPD")
+class DPDPressContrib(ut.TestCase):
```
jngrad: Wouldn't it be easier to instead extend the existing DPD pressure test?

https://github.com/espressomd/espresso/blob/5c7ae80bbe4943c157e4e3940c61a60700f00295/testsuite/python/dpd.py#L395-L399

One line to get the DPD pressure via `system.analysis.pressure_tensor()["DPD"]` and one line to check it's equal to `-stress` (pressure should be the negative of the stress).

src/python/espressomd/analyze.py
mebrito: I also thought on doing it in this alternative way, but since energy and pressure use the same class in the core for calculation, that would lead to an extra meaningless "dpd" key in the energy dictionary.

jngrad: That would have been fine. Other features also have zero contribution in several fields, for example the cold Drude thermostat doesn't contribute to the energy.

jngrad (approved):

PR #5047, opened 2025-02-18
Add storage-agnostic function to add bonds between particles

RudolfWeeber:
* adds a function `add_bond(bond_id, particle_ids)`
* use it (rather than storage-aware code) in method "add_bond" for the Particle Handle script interface
* use it in the BindCenters collision detection scheme. Here, we check via a flag, that one-sided bond storage is used, because the current implementation doesn't guarantee that cross-mpi-rank pairs are handled on both involved ranks

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -373,36 +373,6 @@ tutorials-samples-no-gpu:
   rules:
     - if: $CI_PIPELINE_SOURCE == "schedule"
 
-rtx4090:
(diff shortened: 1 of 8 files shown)


RudolfWeeber:
@jngrad once you are fine with the changes, we can merge this and the remaining occurences (in the other collision detection modes) can be done as a relatively easy task next coding day.

jngrad (dismissed):

jngrad (approved):

Issue #5038, opened 2025-02-03
Python access to instantaneous pressure


RudolfWeeber:
Teh instantaneous pressure (on the fly calculated during force calculaiton for NPT) is currently not accessible from Python.
To allow testings, add
`system.analysis.get_instantaneous_pressure()` fand `...virials()`
This can then return the value from the core (currently i the `nptiso` global, but to be moved into the `System` class)

PR #5053, opened 2025-03-13
New implementation of NPT ensemble based on Andersen equation

### LINKED ISSUE #5037: Separate data structs for NPt Parameters and instantaneous parameters
In Preparation for NPt rework.

Both, the parameters (.e.g. Piston mass, external pressrure) and cached pressure values from last force calculation (instantaneous pressure, p_inst) are mixed up in a singel struct `NPTIsoParameters` in `src/core/npt.hpp`

* [ ] Step 1: Split NPTIsoParameters into NPTIsoParameters and InstantaneousPressure, so that the firest one really only contains parametes of the algorithm and the second one contains the pressure that is calculated on the fly during force calculaiton. 
    Note: The easiest way to refactor is to temporarily usie two globals, the current `nptiso` and a new `npt_inst_pressure`.
* [ ] step 2: get rid of the two globals by making them leafs of the `System` class

hidekb:
Fixes #5037 and #5038.

Description of changes:

- Instantaneous pressure during NPT integration can be obtained from Python.
  - NptIsoParameters nptiso becomes a variable in the System class.
- NPTIsoParameters is divided into NPTIsoParameters and InstantaneousPressure, where the former contains only the algorithm's parameters, and the latter includes the pressure calculated during force calculation.
- The previous implementation of the NPT ensemble cannot calculate isothermal compressibility correctly. To solve this problem, we split thermal fluctuation from momentum time evolution and replaced the algorithm for thermal fluctuation based on [Benedict's paper](https://doi.org/10.1063/1.4802990).
- Mass dependence of the prefactor for thermal fluctuation was also included in the friction and thermal noise prefactor.


RudolfWeeber:
Thank you Hideki.

A few things to check:
* in `void velocity_verlet_npt_finalize_p_inst(double time_step)`, would it be possible to aggreagate the mpi reductions into a single one so as to avoid three syncrhonization barriers?
* your implementation relies on storing prefactors for particle masses in a std::unordered_map. Are you sure this is actually faster than calculating the factor on the fly? Element lookups from an std::unorderd_map tend to be surprisingly slow.
* If you keep the cache-based implementation, please make sure that the rebuild of the cache is triggerd, when `System::on_particle_change()` is run.
* Could you please also annote the equations in the code (or equations numbers from the paper)?
* In the Python test case, could you please annote meanings of variables at first use, e.G., `avpV_inst`
* Please also annote a reference for the expected .5 compressibility 
* Lastly, please check if an update of the documentation is neededd.

hidekb:
Dear Rudolf,

I  checked and modified the code and the document following your suggestions.

> in `void velocity_verlet_npt_finalize_p_inst(double time_step)`, would it be possible to aggreagate the mpi reductions into a single one so as to avoid three syncrhonization barriers?

I aggregate two MPI reductions into a single one.


> your implementation relies on storing prefactors for particle masses in a std::unordered_map. Are you sure this is actually faster than calculating the factor on the fly? Element lookups from an std::unorderd_map tend to be surprisingly slow.

I compared computational time per 5000 steps in the system with WCA potential, p_ext=1.0, and particle number 1000 on giraffe with condor_drain.  And I concluded that both ways are almost the same;
* storing prefactors for particle masses 1.732 $\pm$ 0.002 ms
* calculating the factor on the fly 1.741 $\pm$ 0.004 ms
 
> If you keep the cache-based implementation, please make sure that the rebuild of the cache is triggerd, when `System::on_particle_change()` is run.

In my understanding, the cache is rebuilt when `m_resort_particles` is not 0. In the present code, after running `System::on_particle_change()`, `m_resort_particles` is always greater than 0. Then, the cache is rebuilt after running `CellStructure::update_ghosts_and_resort_particle`. 

>     Could you please also annote the equations in the code (or equations numbers from the paper)?

I annotated the equations in the code.

>     In the Python test case, could you please annote meanings of variables at first use, e.G., `avpV_inst`

I annotated the meaning of variables.

>     Please also annote a reference for the expected .5 compressibility

No paper directly writes the compressibility value, $\kappa_{T}$ of the WCA fluid at pressure $p=1.0$ and $k_{B}T=1.0$ in LJ unit. But, with using the Carnahan-Starling equation of state, we can estimate $\kappa_{T}$ at $k_{B}T=1.0$ and arbitrary pressure, because the Carnahan-Starling equation of state can well describe the pressure of the WCA fluid at $k_{B}T=1.0$. For $p=1.0$ and $k_{B}T=1.0$,  $\kappa_{T}=0.5$. It is annotated in the test Python script.
 
>     Lastly, please check if an update of the documentation is neededd.

I updated the documentation about NpT integration.

RudolfWeeber (dismissed):

jngrad (approved):
Thank you very much!

PR #5064, opened 2025-03-24
Ek flux boundary continuity

### LINKED ISSUE #5049: EK density continuity broken for for boundaries in the ghostlayer
I found a bug, where the global ek density increases over time. This happens when the sample script is run with multiple mpi processes, put it works fine when only using 1 process.
```python
import numpy as np
import espressomd
import espressomd.electrokinetics
import espressomd.shapes

box_l = 20
time_step = 0.2

# Parameters
kT = 1.0
density = 0.2
valency = 1.0
diffusion = 0.2
ext_efield = [0.01, 0., 0.]
sphere_radius = 4.0


# Set-up the system
system = espressomd.System(box_l=[box_l] * 3)
system.time_step = time_step
system.cell_system.skin = 1.0


# Set-up the EK
lattice = espressomd.electrokinetics.LatticeWalberla(agrid=1.0, n_ghost_layers=1)
ek_solver = espressomd.electrokinetics.EKNone(lattice=lattice)
system.ekcontainer = espressomd.electrokinetics.EKContainer(
    solver=ek_solver, tau=system.time_step)


# Set-up the EK species
ek_species_pos = espressomd.electrokinetics.EKSpecies(
    lattice=lattice,
    single_precision=False,
    kT=kT,
    tau=system.time_step,
    density=density,
    valency=valency,
    diffusion=diffusion,
    advection=False,
    friction_coupling=False,
    ext_efield=ext_efield,
    thermalized=False
)
system.ekcontainer.add(ek_species_pos)

# Set-up the boundary
sphere = espressomd.shapes.Sphere(center=[box_l/2]*3, radius=sphere_radius)

ek_species_pos.add_boundary_from_shape(
    shape=sphere, value=[0.0,0.0,0.0], boundary_type=espressomd.electrokinetics.FluxBoundary)
    

for i in range(20):
    system.integrator.run(100)
    print(np.mean(ek_species_pos[:,:,:].density))
```

itischler:
Fixes #5049

Adds the ghost communication for the ek flux boundaries.
However this only fixes the issue when the size of the ghost layer is two (n_ghost_layers=2). This is due to the boundary needs all its neighbors for the kernel to work properly and we need to run the kernel on the first ghost layer cells,  so that the fluxes into and out of the neighboring real cells has the correct value.


src/walberla_bridge/src/electrokinetics/EKinWalberlaImpl.hpp
```diff
@@ -327,6 +348,18 @@ class EKinWalberlaImpl : public EKinWalberlaBase {
     m_full_communication = std::make_shared<FullCommunicator>(blocks);
     m_full_communication->addPackInfo(
         std::make_shared<PackInfo<DensityField>>(m_density_field_id));
+    m_boundary_communicator =
+        std::make_shared<BoundaryFullCommunicator>(blocks);
+    m_boundary_communicator->addPackInfo(
+        std::make_shared<field::communication::BoundaryFlagPackInfo<FlagField>>(
+            m_flag_field_flux_id));
+    auto flux_boundary_packinfo = std::make_shared<
+        field::communication::BoundaryPackInfo<FlagField, BoundaryModelFlux>>(
```
reinaual, line 357: Did you check that the Density boundaries are not suffering from the same issue?

src/walberla_bridge/src/electrokinetics/EKinWalberlaImpl.hpp
```diff
@@ -606,6 +652,7 @@ class EKinWalberlaImpl : public EKinWalberlaBase {
 
     // is this the expected behavior when reactions are included?
     kernel_boundary_density();
+    m_pending_ghost_comm.set(GhostComm::DENS);
```
reinaual, line 655: I strongly vote for this to be properly tested. Your change of the `no_flux.py`-testcase to at lest two ranks will hopefully have the side effect, that this will catch the missing sync for multiple MPI-ranks, however, it will not catch a missing PBC. I suggest doing it either, as you mentioned in our offline discussion, by changing the diffusion testcase to reach over the PBC, which would then also break on one MPI-rank, but increase it's complexity... Alternatively, we could use a new testcase. My low-complexity suggestion would be to have a diffusion-only system with a very high diffusion coefficient and wait for a concentration peak to homogeneously diffuse in the domain and check for density conservation.

itischler: i tested that the ek_diffusion.py, which is a concentration peak that diffuses over 70 timesteps and reaches into the PBC, works with this flag and does not without, because the without this flag the density is not conserved.

src/walberla_bridge/src/electrokinetics/EKinWalberlaImpl.hpp
itischler: I did not and it probably does, but as the density is synced after the integration step regardless of the presence of a density boundary it should not matter.

itischler: i looked it up, as the density boundary kernel was created with a D3Q1 stencil it is purely local and as for the adding of boundary nodes the ghostlayers are being considered, they should not have this issue in the first place.

reinaual: You're right, I remember. We don't remove the domain flag from boundary fields, which is why I went for the D3Q1 hack, at the cost that only simple Dirichlet boundary conditions can be implemented for the density field. Thanks for checking again 👍

src/walberla_bridge/src/electrokinetics/EKinWalberlaImpl.hpp
reinaual: Okay, you checked that this test fails already without this line. I thought we wouldn't leak enough for the threshold to trigger - that's even better!

reinaual (approved):

PR #5069, opened 2025-04-03
Reduce overhead of particle creation

### LINKED ISSUE #5011: Particle creation is slow in parallel simulations
The time for creating a particle scales linearly in the number of mpi ranks for high core numbers (>~64)
E.g 30 ms per particle creatoin (explicitly specifying the particle id) for 512 MPi ranks on the ant cluster at ICP.

In my opionion, particle insertoin should comprise
* a broadcast from Python on rank 0 to all ranks (scales n log n in number of ranks)
* maybe some point to point communicaoitn at constatn time
## Steps to investigate

Use a Python profiler on the particle creation code in src/python/espressomd.particle_data.py with different number of mpi ranks to figure out, where in the creation process the poor scaling appears.

The script was

```python
import espressomd
import numpy as np
from time import time

s=espressomd.System(box_l=[1,1,1])
grid =s.cell_system.node_grid
n=10*np.prod(grid)
print(grid)
for n in 1,10,100:
    tick=time()
    s.part.add(id=range(n),pos=np.random.random((n,3))*s.box_l)
    tock=time()
    print(n, (tock-tick)/n)
    s.part.clear()
```

And the commands to run were

```bash
module load spack/default gcc/12.3.0 cuda/12.3.0 openmpi/4.1.6 \
            fftw/3.3.10 boost/1.83.0 cmake/3.27.9 python/3.12.1

`
srun --cpu-bind=cores --ntasks=512 --time=60 -u ./pypresso ./test.py 
```

RiccardoFrenner:
Description of changes:
- Return the ParticleHandle object directly from add_particle in the C++ Script Interface instead of just its ID.
- This eliminates an unnecessary call to retrieve the particle by ID after creation, improving efficiency, as investigated in #5011.

This does not fix #5011 but is still an improvement, especially for lower core counts.

### DIFF
--- a/src/python/espressomd/particle_data.py
+++ b/src/python/espressomd/particle_data.py
@@ -1129,8 +1129,7 @@ class ParticleList(ScriptInterfaceHelper):
             bonds = p_dict.pop("bonds")
             if nesting_level(bonds) == 1:
                 bonds = [bonds]
-        p_id = self.call_method("add_particle", **p_dict)
-        p = self.by_id(p_id)
+        p = self.call_method("add_particle", **p_dict)
         for bond in bonds:
             if len(bond):
                 bond = p.normalize_and_check_bond_or_throw_exception(bond)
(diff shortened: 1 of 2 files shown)


jngrad (approved):
Thanks!

PR #5072, opened 2025-04-04
for each particle

RudolfWeeber:
Description of changes:
- introduce `for_each_{local,ghost}_particle()`
- make `RuntimeErrorCollector` thread-safe

This introduces `cell_structure.for_each_local_particle()` and `cell_structure.for_each_ghost_particle()` templates and uses them in integrators / virtual sites / force setup etc. For now, I rolled it out only where the per-particle operation has no side effects and is safe to execute in parallel. The one exception is the `RuntimeErrorCollector`, where a mutex was added.

Once we have Kokkos-support in, we can implement parallel versions of the templates without modifying the call sites. Probably some sort of heuristics will be needed to only use the parallel version for a high enough number of particles.

The most relevant places still missing (because they have side effects) is bonded force calculation and LB coupling. I'm not sure about side effects of constraints, in particular how DPD-based constraints handle the counter.

### DIFF
--- a/src/core/cell_system/CellStructure.hpp
+++ b/src/core/cell_system/CellStructure.hpp
@@ -44,6 +44,7 @@
 
 #include <algorithm>
 #include <cassert>
+#include <concepts>
 #include <iterator>
 #include <memory>
 #include <optional>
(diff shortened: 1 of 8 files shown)


jngrad (approved):
LGTM

PR #5073, opened 2025-04-07
Fix DPD thermostat

jngrad:
Description of changes:
- DPD thermostat prefactors are now always up-to-date, regardless of the order in which the DPD thermostat and DPD non-bonded interaction are set up

### DIFF
--- a/src/core/system/System.cpp
+++ b/src/core/system/System.cpp
@@ -295,8 +295,7 @@ void System::on_short_range_ia_change() {
 void System::on_non_bonded_ia_change() {
   nonbonded_ias->recalc_maximal_cutoffs();
   rebuild_cell_structure();
-  thermostat->recalc_prefactors(time_step);
-  reinit_thermo = false;
+  on_thermostat_param_change();
   propagation->recalc_forces = true;
 }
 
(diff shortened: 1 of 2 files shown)


Yashas17:
Looks good !

Yashas17 (approved):

RudolfWeeber:
Since the prefactors depend on timestep, are we sure, this is also called, when someone changes the timestep after setting up the thermostat or the interaction?

Yashas17:
I just checked, `reinit_thermo` is set to `true` when the time step is changed, even in the current build. So this should work flawlessly. However, maybe we can add a test case for this as well.

jngrad:
I think so: `on_timestep_change()` calls `on_thermostat_param_change()` unconditionally, which sets the `reinit_thermo` flag to true, such that thermostat prefactors are recalculated on integration start. I'm actually wondering if we should not set the flag to true in `on_non_bonded_ia_change()` to queue the prefactors update, instead of updating them right away. I don't think we'll be gaining anything in terms of performance given how cheap the operation is, but I could look into it in terms of readability or maintainability in the next refactoring PR.

Yashas17 (approved):
Looks okay.

RudolfWeeber (approved):

PR #5075, opened 2025-04-16
LB pull scheme and zero-centered storage format

### LINKED ISSUE #4514: Walberla: Switch to pull scheme
The pull-scheme (stream-collide) is already used for Lees-Edwards. It should be used everywhere, and the combined stream-collide kernel should be used rathan individual ones.
If this should not be possible for some reason, at least the streaming step can be vectorized.

### LINKED ISSUE #5071: lb momentum broken for single precision
Playing around with the lb_momentum_conservation testcase i found some strange behaviors. The testcase sets a particle in a lb fluid and applies a external force opposite to the fluids external force. The particle also has initial velocity. For simplification I set the force direction to one cardinal axis and turned the thermostat off. In the end the momentum is measured by `system.analysis.linear_momentum()`. I would expect to get similar results regardless of the axis i chose in the beginning. 
```
import espressomd
import espressomd.lb
import numpy as np

TIME_STEP = 0.008
AGRID = 1.0
GRID_SIZE = 10
KVISC = 4
DENS = 2.3
F = 5.5 / GRID_SIZE**3
GAMMA = 1
lb_params = {"single_precision": True}
direction = np.array([0,0,1])
LB_PARAMS = {'agrid': AGRID,
             'density': DENS,
             'kinematic_viscosity': KVISC,
             'tau': TIME_STEP,
             'ext_force_density': direction * 0.7 * F}


system = espressomd.System(box_l=[GRID_SIZE * AGRID] * 3)
system.time_step = TIME_STEP
system.cell_system.skin = 0.01

system.cell_system.set_regular_decomposition()
lbf = espressomd.lb.LBFluidWalberla(**LB_PARAMS, **lb_params)


system.lb = lbf

ext_fluid_force = system.volume() * LB_PARAMS["ext_force_density"]

p = system.part.add(
    pos=system.box_l / 2, ext_force=-ext_fluid_force, v=direction * 0.2)

initial_momentum = np.array(system.analysis.linear_momentum())

for i in range(10):
    system.integrator.run(500)
    measured_momentum = system.analysis.linear_momentum()
    print(measured_momentum)
```
Running this system with double precision works fine.
`[2.15400000e-01 7.63278329e-11 2.22044605e-10]`

however in single precision the momentum in the non chosen direction are too large and also vary depending on the axis:
`direction = np.array([1,0,0])`: `[0.29751236 0.12665987 0.09685755]`

`direction = np.array([0,1,0])`: `[-0.06705523  0.23790772  0.01117587]`

`direction = np.array([0,0,1])`: `[-0.00745058  0.14156103  0.12906964]`

This is with the current kernels in espresso, and with newly generated ones, where we know that the momentum_conservation test fails it does not get better. So the question is, is this a purely lbmpy problem, or is there also some things wrong on our side.

itischler:
Fixes #4514
Fixes #5071

Description of changes:
* use pull scheme and merge the stream and collide kernels together
* use zero-centered storage format (only storing the differences from the equilibrium) to increase accuracy
* API change: LB populations are not longer publicly accessible
   * LB populations are sensitive to implementation details:
      * with push/pull scheme, the populations are obtained before/after collision
      * with zero-centered storage format, populations are relative to a reference distribution


itischler:
i changed the momentum conservation from absolute to relative tolerance and increased the velocity and forces to circumvent the floating point precision background velocity mentioned in #5071 .

RudolfWeeber:
There seems to be a lot of commented out code in the LB Lees Edwards particle coupling test. Is that intentional?

maintainer/walberla_kernels/generate_lb_kernels.py
```diff
@@ -167,23 +176,37 @@ def generate_collide_kernels(ctx, method, data_type):
...
+def generate_macroscopic_value_getter_kernels(ctx, method, fields):
+    precision_prefix = pystencils_espresso.precision_prefix[ctx.double_accuracy]
+    for params, target_suffix in paramlist(parameters, ("GPU", "CPU", "AVX")):
+        pystencils_walberla.generate_sweep(
+            ctx,
+            f"UpdateVelFromPDF{precision_prefix}{target_suffix}",
+            lbmpy.macroscopic_value_kernels.macroscopic_values_getter(
+                method, None, fields["velocity"], fields["pdfs"],
+                use_pre_collision_pdfs=True),
+            **params)
```
RudolfWeeber, line 225: Why do we need pre_collisoin symbols. What difference does this make in the results?

maintainer/walberla_kernels/generate_lb_kernels.py
```diff
@@ -112,53 +113,61 @@ def generate_init_kernels(ctx, method):
...
-        shear_dir_normal)
-
+        shear_dir_normal, True)
+    optimization = {"cse_global": True,
+                    "double_precision": ctx.double_accuracy}
+    print(le_collision_rule_unthermalized)
     for params, target_suffix in paramlist(parameters, ("GPU", "CPU", "AVX")):
-        pystencils_espresso.generate_collision_sweep(
+        stem = f"StreamCollideSweep{
+            precision_prefix}LeesEdwards{target_suffix}"
```
RudolfWeeber: what does "stem"stnad for? Also, the kernel name construiont logic appears repeatedly. Coudl we outsource it to a `kernel_name("StreamColide", target, precision) or similar?

maintainer/walberla_kernels/lees_edwards.py
```diff
@@ -59,16 +59,22 @@ def velocity_offset_eqs(config, method, pdfs, shear_dir_normal, stencil):
 
     grid_size = TypedSymbol("grid_size", dtype=default_dtype)
 
+    # When the kernels are combined, the integration loop also reaches
+    # one grid cell further to pull the populations. This changes the
+    # position where to apply the LE boundary by one cell.
+    boundary_offset = [0, grid_size - 1]
+    if (combined_kernel):
+        boundary_offset = [1, grid_size - 0]
     # +,-1 for upper/lower boundary layers, 0 otherwise.
```
RudolfWeeber, line 69: Please clarify: in the generated stream collide kernels based on the pull scheme, the loops in the generate kernels run on (1<=i<=n). For pure collisoin kernels they run on (0<=i<=n-1).

src/walberla_bridge/src/lattice_boltzmann/LBWalberlaImpl.hpp
```diff
@@ -633,45 +623,37 @@ class LBWalberlaImpl : public LBWalberlaBase {
...
-    // LB stream
-    integrate_stream(blocks);
+    // LB stream collide
+    integrate_stream_collide(blocks);
     // Mark pending ghost layer updates
-    m_pending_ghost_comm.set(GhostComm::PDF);
+    // As pdf and laf are communicated direcly afterwards they are not set.
     m_pending_ghost_comm.set(GhostComm::VEL);
     m_pending_ghost_comm.set(GhostComm::LAF);
-    // Refresh ghost layers
```
RudolfWeeber, line 645: As I understand it, the last applied force is communicated by the pdf_streaming_communicator. If this is the case, LAF should probably not be marked as pending, here.

src/walberla_bridge/src/lattice_boltzmann/LBWalberlaImpl.hpp
```diff
@@ -633,45 +623,37 @@ class LBWalberlaImpl : public LBWalberlaBase {
...
     auto const &blocks = get_lattice().get_blocks();
     // Reset force fields
     integrate_reset_force(blocks);
-    // LB collide
-    integrate_collide(blocks);
-    m_pdf_streaming_communicator->communicate();
-    // Handle boundaries
-    if (m_has_boundaries) {
-      integrate_boundaries(blocks);
-    }
```
RudolfWeeber, line 638: Should integrate_boundaires mark pdf and velocity ghost layer communicatoin as pending?

testsuite/python/lb.py
```diff
@@ -656,7 +656,7 @@ def test_viscous_coupling_rounding(self):
         self.system.lb = lbf
         self.system.thermostat.set_lb(LB_fluid=lbf, seed=3, gamma=self.gamma)
         p = self.system.part.add(pos=[-1E-30] * 3, v=[-1, 0, 0])
-        self.system.integrator.run(1)
+        self.system.integrator.run(2)
         for _ in range(20):
```
RudolfWeeber, line 660: Here (and in some other places), why do we need one extra time step? In the integrate_pull() function in LBWalberlaImpl.hpp ResetForces is the first call, so shouldn't the force be applied immediately?

maintainer/walberla_kernels/templates/FieldAccessors.tmpl.h
```diff
@@ -600,8 +604,14 @@ namespace PressureTensor
          Cell const & cell )
    {
         const {{dtype}} & xyz0 = pdf_field->get(cell, uint_t{ 0u });
+        const int x = cell[0];
+        const int y = cell[1];
+        const int z = cell[2];
         {% for i in range(Q) -%}
-            const {{dtype}} f_{{i}} = pdf_field->getF( &xyz0, uint_t{ {{i}}u });
+            const {{dtype}} f_{{i}} = pdf_field->get({{inv_neighbor_dir[i][0]}} + x,
+                                                     {{inv_neighbor_dir[i][1]}} + y,
```
RudolfWeeber, line 616: Why do we have to acces the neighbors rather than the actual cell here and in the other accessors? Hasn't the streaming alredy taken place in the StreamCollide sweep?

RudolfWeeber (dismissed):

maintainer/walberla_kernels/templates/FieldAccessors.tmpl.h
itischler: Yes, however, for the pressure tensor it makes a huge difference if it is calculated after streaming or after collision, as the collision step drives the pressure tensor towards its equilibrium state. As for our cases we want to look at the pressure tensor mostly in a steady state away from equilibrium and for those cases it is preferred to calculate the pressure tensor after the streaming step. Therefore by looking at the pdfs from the neighbors, we obtain the pressure tensor as it would be after the next streaming step.

maintainer/walberla_kernels/generate_lb_kernels.py
itischler: As i see it it uses the neighboring cells pdf to calculate the local velocity as is the next streaming step has already happened. this is necessary because the boundary conditions are handled after the stream collide sweep, therefore the local velocity in the cells next to the boundary would be wrong. So using this pre_collision_pdfs flag fixes this, however it also makes the LB velocity be 1 timestep ahead of the md, ( or half a time step depending on how we look at it). This is also the reason, why in many testcases we add 1 when comparing against the system time.
Quite frankly this is something i found out by try and error until i found a way in which it worked.

src/walberla_bridge/src/lattice_boltzmann/LBWalberlaImpl.hpp
itischler: Yes, however, there are some unit tests (LBWalberlaImpl_unit_tests, LBWalberlaImpl_ bspline_tests) which need the flag to be set to true after the integration step, so that the forces can be communicated. it would probably be better to have function, which can communicate regardless of the flag.

maintainer/walberla_kernels/generate_lb_kernels.py
itischler: this name was chosen before, but i'll rename it

maintainer/walberla_kernels/lees_edwards.py
itischler: a pure collision kernel runs over grid_size number of cells in each direction (0<=i<=n-1). However if the pull stream_collide kernel it also iterates over one ghostlayer in each direction to pull the pdfs from (0<=i<=n+1). This however makes the cells to apply the LEBC to shift by one cell.

src/walberla_bridge/src/lattice_boltzmann/LBWalberlaImpl.hpp
itischler: I'll mark the pdf to be communicated

testsuite/python/lb.py
itischler: Here it is not needed anymore (it was at first) so i remove it. After testing the other cases i found out, that they are still needed.

src/walberla_bridge/src/lattice_boltzmann/LBWalberlaImpl.hpp
itischler: I removed it again, as it lead to many tests failing

jngrad:
@itischler for zero-centered pdfs, one has to pass the options to the `LBMConfig` object directly, e.g.

```diff
diff --git a/maintainer/walberla_kernels/generate_lb_kernels.py b/maintainer/walberla_kernels/generate_lb_kernels.py
index 5720a0704..ad17429dc 100644
--- a/maintainer/walberla_kernels/generate_lb_kernels.py
+++ b/maintainer/walberla_kernels/generate_lb_kernels.py
@@ -181,5 +181,4 @@ def generate_collide_kernels(ctx, method, data_type):
     lb_collision_rule_thermalized = lbmpy.creationfunctions.create_lb_collision_rule(
         method,
-        zero_centered=False,
         fluctuating={
             "temperature": kT,
@@ -328,4 +327,6 @@ with code_generation_context.CodeGeneration() as ctx:
     method = lbmpy.creationfunctions.create_mrt_orthogonal(
         stencil=stencil,
+        delta_equilibrium=True,
+        zero_centered=True,
         compressible=True,
         weighted=True,
```
and similar for the Lees-Edwards kernel, your new kernel, and `generate_pack_info_pdfs_field_assignments()`.

maintainer/walberla_kernels/generate_lb_kernels.py
reinaual: @RudolfWeeber Just for clarification, it's a universal term in filepaths for the name of a file without it's extension. See [C++](https://en.cppreference.com/w/cpp/filesystem/path/stem) or [Python](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.stem)

jngrad:
Parallel simulations are now 15% faster.

Strong scaling plot of CPU LB with a box size of 128x64x128. The parallel performance increased by 15% from 4 to 16 cores.

Parallel efficiency has improved marginally.

Weak scaling plot of CPU LB with a local box size of 48x48x48 per core. The parallel efficiency improves by about 5 percentage points up to 4 cores, and 1 percentage point up to 16 cores..

jngrad (approved):
Thanks a lot!

PR #5076, opened 2025-04-16
Fix Cabana integration

phohenberger:
Resolve compatibility issues between external libraries.


RudolfWeeber:
While kokkos was found by cmake with this patch, I cannot actually include kokkos headers, e.g., "Kokkos_Core.hpp". Probably KOKKOS has to be linked with espresso_core or similar?

jngrad:
@RudolfWeeber weird, we do have `#include <Kokkos_Core.hpp>` in `communication.cpp` and it compiles in CI. Did you update the build folder with `cmake . -D ESPRESSO_BUILD_WITH_CABANA=ON`?

RudolfWeeber:
The espresso-p3m library needs to be target_link_library'd to Cabana, too.

jngrad (approved):

PR #5078, opened 2025-04-18
Kokkos version of for_each_particle

RudolfWeeber:
There are 2 implementations. When there are many cells (regular decomposition) the parallel loop is over cells, when there is only one (N-square) the parallel loop is over particles.

Notes: 
* `Kokkos::Experimental::for_each()` only works on Kokkos views, therefore `parallel_for()` is used instead
* So far, the `for_each_particle()` template is only used in cases, where no shared data is accessed. With 4 OpenMP threads, this reduces the time for the LJ benchmark by a few percents.
* next step is to make access to the most important shared data thread safe, particularly, the particle forces and the P3M real space charge density field

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -503,6 +503,8 @@ if(ESPRESSO_BUILD_WITH_FFTW)
 endif()
 
 if(ESPRESSO_BUILD_WITH_SHARED_MEMORY_PARALLELISM)
+  find_package(OpenMP REQUIRED)
+
   if(NOT EXISTS ${FETCHCONTENT_BASE_DIR}/kokkos-src)
     find_package(Kokkos 4.3 QUIET)
   endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -829,6 +829,14 @@ if(ESPRESSO_BUILD_TESTS)
             "Extra arguments to give to ctest calls (separated by semicolons)")
   set(ESPRESSO_TEST_NP "4" CACHE STRING
                                  "Maximal number of MPI ranks to use per test")
+  add_library(espresso_tests_cpp_flags INTERFACE)
+  add_library(espresso::tests::cpp_flags ALIAS espresso_tests_cpp_flags)
+  if(ESPRESSO_BUILD_WITH_COVERAGE)
+    target_compile_options(
+      espresso_tests_cpp_flags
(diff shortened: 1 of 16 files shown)


jngrad:
With 2 OpenMP threads per MPI rank, we end up oversubscribing the resources in the docker container. While the second thread does almost no work, it still leads to a massive performance penalty, between 45% and 90% slowdown depending on the load on the runner. Not sure what causes this delay, maybe frequent context switches? I've just conducted quick benchmarks with a runner that exposes 6 CPU cores, and found out that when CTest is asked to use at most 4 cores ("check cores") and ignore the 2 remaining cores ("buffer cores"), the total runtime is close to the runtime we have on the python branch.

```
python branch // CI job on saturated runner // 4 check cores
lb_momentum_conservation_1_core ........................................   Passed   67.94 sec
lb_momentum_conservation ...............................................   Passed   33.44 sec
lb .....................................................................   Passed   26.22 sec
dpd_stats ..............................................................   Passed    8.05 sec
integrator_npt_stats ...................................................   Passed   25.65 sec
Total Test time (real) = 409.04 sec

PR branch // CI job on saturated runner // 4 check cores
lb_momentum_conservation_1_core ........................................   Passed  141.65 sec
lb_momentum_conservation ...............................................   Passed  132.59 sec
lb .....................................................................   Passed   55.15 sec
dpd_stats ..............................................................   Passed    9.31 sec
integrator_npt_stats ...................................................   Passed   49.19 sec
Total Test time (real) = 669.74 sec

PR branch // isolated CI job // 6 check cores
lb_momentum_conservation_1_core ........................................   Passed   99.01
lb_momentum_conservation ...............................................   Passed   57.67 sec
lb .....................................................................   Passed   27.66 sec
dpd_stats ..............................................................   Passed    8.86 sec
integrator_npt_stats ...................................................   Passed   41.68 sec
Total Test time (real) = unknown, canceled job

PR branch // isolated CI job // 4 check cores + 2 buffer cores
lb_momentum_conservation_1_core ........................................   Passed   60.31 sec
lb_momentum_conservation ...............................................   Passed   36.94 sec
lb .....................................................................   Passed   20.95 sec
dpd_stats ..............................................................   Passed    8.90 sec
integrator_npt_stats ...................................................   Passed   40.90 sec
Total Test time (real) = 428.06 sec
```

jngrad (approved):

PR #5081, opened 2025-04-29
Lees-Edwards bugfixes

Yashas17:
Fixes #

Description of changes:
- Changed the sign in lees-edwards distance function
- Updated test cases to use the correct sign
- Corrected velocity difference calculation


RudolfWeeber (dismissed):
From my side, this can be merged. I added two additional test cases to Yashas'  chagnes, which explicitly test the consistency of the signs between the push step applied to a particle crossing the boundary and the velocity and positipn difference functions used in (short range) interaction calculation.

The more extensive steepest descent in the VS relative test might help with the flakyness of this test case which existed independently of these changes.

jngrad (approved):
As per https://github.com/espressomd/espresso/pull/5081#pullrequestreview-2841125153

PR #5082, opened 2025-04-29
Fix inclusion of external libraries

jngrad:


CMakeLists.txt
```diff
@@ -484,46 +451,119 @@ endif()
...
+      endif()
+    endforeach()
   endif()
-  install(TARGETS Heffte LIBRARY DESTINATION "${ESPRESSO_INSTALL_LIBDIR}")
   if(ESPRESSO_BUILD_WITH_CLANG_TIDY)
-    set(HEFFTE_CXX_CLANG_TIDY "${ESPRESSO_CXX_CLANG_TIDY}")
+    set(KOKKOS_CXX_CLANG_TIDY "${ESPRESSO_CXX_CLANG_TIDY}")
     set(SKIP_CLANG_TIDY_CHECKS "")
     set(SKIP_CLANG_TIDY_CHECKS_CXX "")
     # silence heFFTe diagnostics
```
reinaual:
```suggestion
    # silence kokkos diagnostics
```

reinaual (changes requested):

cmake/cabana.patch
```diff
@@ -1,16 +1,6 @@
-https://github.com/ECP-copa/Cabana/pull/815
 --- CMakeLists.txt
 +++ CMakeLists.txt
-@@ -216,12 +216,6 @@
-   enable_testing()
+@@ -56,3 +56,2 @@ endforeach()
+ if(Kokkos_ENABLE_CUDA)
+-  kokkos_check(OPTIONS CUDA_LAMBDA)
```
reinaual, line 5: For my understanding, why do we have to disable this check in kokkos?

jngrad: The `kokkos_check()` function is only defined in the CMake-generated `KokkosConfigCommon.cmake` file. Including this file completely breaks the `FetchContent` build system.

reinaual (approved):

PR #5083, opened 2025-04-29
LB OpenMP kernels

### LINKED ISSUE #5080: Turn on OpenMP for waLBerla kernels
probably by adding cpuo[penmp=True to the parameters["CPU"] and parameters["AVX"] dictionaires in src/maintainer/walberla_kernels/generate_lb_kernels.py

itischler:
Fixes #5080

changes the lb kernel generation to use openMP


RudolfWeeber:
Doing a "grep -l pragma.*omp *.?pp"on the generated kernels, that the AVX kernels didn't get the mememo...
Probably the omp stuff needs to go to the parameters["AVX"] dict as well.
I didn't yet try with a non-avx build, though.

RudolfWeeber (dismissed):

jngrad (changes requested):
In the generated kernels, `SHARED_MEMORY_PARALLELISM` didn't exist, because `espresso::walberla` doesn't depend on `espresso::config`. We need to check for `_OPENMP` instead, which is only defined when the compiler is passed `-fopenmp`. I've adapted the code accordingly, and now the kernels are built with OpenMP.

The OpenMP kernels seem to be broken though. Even the C++ unit tests are failing when using 2 threads.

RudolfWeeber:
OK, we had a few successful CI runs now. Seems like we got the remaining race condition by setting the thread save receive to false on the boundary pack info.

@jngrad should w merge now?

jngrad:
@RudolfWeeber I'll run it locally too and then merge it.

jngrad (approved):

PR #5086, opened 2025-05-06
P3M OpenMP kernels

jngrad:


src/core/p3m/P3MFFT.hpp
```diff
@@ -69,8 +72,16 @@ template <typename FloatType> class P3MFFT {
     auto const global_box = heffte::box3d<>(
         {0, 0, 0}, to_array(m_global_mesh - Utils::Vector3i::broadcast(1)),
         to_array(m_memory_layout));
+    auto const n_procs = Utils::product(node_grid);
+    auto best_grid = node_grid;
+    for (auto i : {0u, 1u, 2u}) {
+      if (m_global_mesh[i] % 2 * n_procs == 0) {
```
hidekb: I assume that `n_procs` is always an integer greater than one. In that case, `if (m_global_mesh[i] % 2 * n_procs == 0)` would behave the same as `if (m_global_mesh[i] % 2 == 0)`. Could you clarify the reasoning behind multiplying by `n_procs`?

RudolfWeeber: Are there braces missing? % (2* nproc) ==0? No Idea what the operator precedence is.

jngrad: Good catch, we actually meant to write `m_global_mesh[i] % (2 * n_procs) == 0`. Modulo has higher precedence than multiply. Let me fix that. I think we multiply by 2 because the `m_global_mesh` is guaranteed to be a multiple of 2.

hidekb (approved):
LGTM

PR #5088, opened 2025-05-07
Kokkos parallel for_each_3d_lin

RudolfWeeber:
- **Kokkos implementation of for_each_local_particle()**
- **p3m: Use OpenMP kernels**
- **Kokkos implementation of for_each_3d_lin**


RudolfWeeber:
@jgrad this is now rebased onto your p3m branch.
Note that the kernel signature for for_each_3d_lin() changed, as the previous one (sung a shard refence for the index) is not suitable for parallel executoin

RudolfWeeber:
@jngrad could you please take a look at the concept for the kernel for for_each_3d_lin. 
to my understanding, the kernel takes two args, namely somthing which is a 3d index vector, and an integer for the linear loop index. However, enabling the concpet results in a flood of compiler errors.

jngrad:
> @jngrad could you please take a look at the concept for the kernel for for_each_3d_lin. to my understanding, the kernel takes two args, namely somthing which is a 3d index vector, and an integer for the linear loop index. However, enabling the concpet results in a flood of compiler errors.

The `std::invocable` instantiation involves `detail::IndexVectorConcept auto`, which is not a valid type since `auto` cannot be deduced. Rewriting as
```c++
template <Utils::MemoryOrder memory_order, class Kernel, class IndexVector>
  requires detail::IndexVectorConcept<IndexVector> and std::invocable<Kernel, IndexVector, int>
```
make the `requires` syntax valid, however that won't work in this case because the `IndexVector` cannot be automatically deduced from the lambda `operator()` method(s). A clean solution isn't immediately obvious to me.

RudolfWeeber:
OK, thanks. I'll remove the concept, then.

hidekb (dismissed):
LGTM

jngrad (approved):
As per https://github.com/espressomd/espresso/pull/5088#pullrequestreview-2844348172

PR #5089, opened 2025-05-14
DPD: consistent noise w.r.t particle traversal order

RudolfWeeber:
The DPD noise function already ensured a consistent key w.r.t the order, in which a pair of particles was passed, by always using `(smaller_id, larger_id)`. However, the fact that `f(id1,id2) = -f(id2,id1)` was not accounted for, so the noise switched sign depending on the order particles were passed. This PR fixes this by applying a +/-1 prefactor to the noise depending on whether the first or second particle id is smaller.

I do not think the existing code lead to wrong physics, as the statistics for the noise was still correct. However, we prefer to have the same actual numbers result independent of cell system or MPI decomposition.

There is a second issue w.r.t DPD constraints, not addressed here. I'll open a separate ticket.


hidekb (approved):
LGTM

PR #5092, opened 2025-05-20
Fix ZnDraw visualizer

ludogibbs:


RudolfWeeber:
Thank you, Lukas. 
As I understand the code, particle types are queried from ESPResSo twice, in some cases?
If this function is called on every visualizer update, you might pre-determine, if you need `system.part.all().type` and obtain it only once rather than twice from the core: `all_types = np.array(system.part.all().type)`

RudolfWeeber (approved):

PR #5093, opened 2025-05-20
Bump dependencies

jngrad:
Description of changes:
* bump all libraries version requirements
* add full support for Cython 3.0.x and drop support for Cython 0.29.x
* bugfix: OpenMPI 5.0 now longer triggers random PRRTE errors at system exit
* migrate to the Ninja build system in CI for significantly faster build times


jngrad:
@hidekb this bumps the Cabana and Kokkos dependencies to the most recent commits that are compatible with one another, and fixes the `no member named 'make_kokkos_reducer' in namespace 'Reduction'` regression in the unit test `particle_reduction_test.cpp`. The way C++ unit tests are implemented has completely changed too: whenever global variables from the ESPResSo core are involved, one has to use `BOOST_TEST_GLOBAL_CONFIGURATION(GlobalConfig)` instead of the custom `int main()` function, in order to properly control the global variables lifetime.

More commits are planned for this PR to handle the waLBerla version bump, but the existing commits won't change.

jngrad:
This PR is needed for EESSI deployment. I can confirm on HPC Vega that ESPResSo can be built with heFFTe, Kokkos, Cabana, HighFive and waLBerla, using EasyBuild 4.9.4 and the foss-2023b toolchain.

.github/workflows/push_pull.yml
```diff
@@ -22,7 +22,8 @@ jobs:
       - name: Get runner specifications
         run: system_profiler SPHardwareDataType
       - run: |
-         brew install boost boost-mpi fftw ccache
+         brew install boost boost-mpi fftw
+         brew list ccache || brew install ccache
```
reinaual, line 26: Why do we have to check for `ccache` and `ninja` first? Shouldn't it be safe to install something that might already be installed?

maintainer/CI/build_cmake.sh
```diff
@@ -258,16 +258,16 @@ else
...
-  cmake "${srcdir}" ${cmake_params} "${cmake_param_protected}" || exit 1
+  cmake -G Ninja "${srcdir}" ${cmake_params} "${cmake_param_protected}" || exit 1
 fi
 end "CONFIGURE"
 
 # BUILD
 start "BUILD"
 
-make -k -j${build_procs} || make -k -j1 || exit ${?}
+time ninja -k 8 -j${build_procs} ${ninja_params} || exit ${?}
```
reinaual, line 270: Did we intentionally add `time` here to measure build times in CI?

maintainer/walberla_kernels/templates/FieldAccessors.tmpl.h
```diff
@@ -581,11 +581,11 @@ namespace MomentumDensity
             for(uint_t y = 0; y < pdf_field->ySize(); ++y) {
                 for(uint_t x = 0; x < pdf_field->xSize(); ++x) {
                     const {{dtype}} & xyz0 = pdf_field->get(x, y, z, uint_t{ 0u });
-                    {% for i in range(Q) -%}
+                    {% for i in range(1, Q) -%}
```
reinaual, line 584: Is it guaranteed that the first element in the range always corresponds to the 0-vector? Probably thats true for the stencil we use in lbmpy, but I guess this would mean that the code generated would not compile for custom orderings without that restriction, right? Might be something one should be aware of for the future.

.github/workflows/push_pull.yml
jngrad: It's safe but generates a warning that gets promoted to a GitHub Action Annotation.

maintainer/CI/build_cmake.sh
jngrad: Yes, it will help with debugging. When a CI job hangs and the testsuite doesn't complete, the testsuite runtime doesn't get printed out, and it's unclear whether the issue comes from the test stage or from the build stage.

maintainer/walberla_kernels/templates/FieldAccessors.tmpl.h
jngrad: Up till now we had inconsistent range numbers between field accessors. Some started at zero but didn't use the 0th value, sometimes they started at 1. Now we only extract values that are actually needed, which allows us to enable compiler flags to detect unused intermediate variables and remove them. The extra diagnostics I've enabled aren't set in stone; if we see they cause any trouble, we can selectively remove them. Regarding stencils with custom orderings, the new code would indeed not work, and a compiler error would alert the developer. If there is a need for custom fields, we could decorate all 19 or 27 variables with `[[maybe_unused]]` to avoid triggering compiler diagnostics.

reinaual: Okay, I'm not against these changes, I only wanted to note about the possible restrictions that you already seem to be aware of

reinaual (approved):

PR #5097, opened 2025-05-30
Shared memory parallelization for short range forces

hidekb:
Follow up on #5051.

Convert arrays of structs (AoS) to structs of arrays (SoA) to improve performance of the short-range loop.


jngrad (dismissed):
Thanks a lot! OpenMP performance is now in the same ballpark as MPI.

LJ with 8000 particles in total:
* 480 ps/step with 4 MPI ranks
* 530 ps/step with 4 OMP threads (before this PR: 1340 ps/step)

P3M with 8000 particles in total:
* 3.3 ms/step with 4 MPI ranks
* 4.1 ms/step with 4 OMP threads (before this PR: 7.0 ms/step)

jngrad (approved):

PR #5100, opened 2025-06-16
Improve ESPResSo versioning

jngrad:
Description of changes:
- bump project version from 4.3 to 5.0
- remove static global variables containing the feature set
- add proper CMake guards for git commands when determining the project commit version at build time
- make our versioning scheme compliant with PEP 440 and SemVer 2.0.0
   - remove the unsupported `-dev` label from the CMake project version number (deprecated anyway)
   - remove the convoluted boost string splitting mechanism that was needed to remove the `-dev` label


jngrad:
This PR will help reduce the patch size in the EESSI development stack (https://github.com/multixscale/dev.eessi.io-espresso/pull/1).

For reviewers: historically we had the `-dev` suffix in development commits and would remove it in a release commit. However, we would often forget to re-introduce it when backporting bugfix commits, for example all intermediate commits between 4.2.0 and 4.2.1 were missing the `-dev` suffix, making the project version inconsistent. CMake doesn't actually support version labels ([cmake/cmake#16716](https://gitlab.kitware.com/cmake/cmake/-/issues/16716)), and we used to introduce this label by overriding an internal variable, which meant any project that included ESPResSo via FetchContent was unable to check the ESPResSo version with mathematical expressions, and variables like `PROJECT_VERSION_MAJOR` were undefined. There is also the risk of forgetting to update the hardcoded version number in `version.hpp.in` when updating the same version number in `CMakeLists.txt`. By having `version.hpp.in` read the version number from CMake, we reduce the risk for human error in our release workflow. Finally, a quick look at the git history reveals the `-dev` label was introduced in 2016 as a side-effect of calculating the `SOVERSION` for packaging .so files, but that packaging feature was removed in 2019 and we have, to my knowledge, never had a use for the `-dev` label.

Regarding version numbers of development commits being meaningless, the way SemVer and PEP 440 are worded make it clear that nothing can be said about the version number of a commit that is not a release commit. Commits on the python branch use the next available version number to help distinguish the python branch from the last stable release, even if that version number will never be released (e.g. 4.3.0 will never exist). Not providing a version number would break scripts that check for the version number for API compatibility (e.g. pyMBE and SwarmRL, and many user scripts too), so I would prefer to keep providing a version number on the python branch and between bugfix releases. We cannot encode in ESPResSo whether the current commit corresponds to a release, because we cannot predict the commit hash before it's generated, and we don't want to establish a POST request to GitHub since HPC clusters might not allow internet access.

RudolfWeeber (approved):

reinaual (approved):

PR #5101, opened 2025-06-17
GPU EK

### LINKED ISSUE #5058: GPU support for EK

### LINKED ISSUE #5079: EKIN GPU roadmap
* [ ] generate GPU versions of all kernels
* [ ] Add a GPU version of the EK kernel traits template (additoinal template parameter), mark the exisitng one as CPU.
* [ ] add the field type to the traits class (or use template logic in EK impl template). Currently, GhostLayerField is hard-coded, which is not suitable for GPU.
* [ ] Add architecture as template parameter to all EK classes templtes and initialization code
* [ ] The field setters/getters probbaly need to be adapted. Accessors to the last applied force field in LB (which is also a plain vector field) can be used as templtes
* [ ] Code checking that LB and EK agree about CPU/GPU and float type for the couping
* [ ] Python interface and testing

itischler:
Fixes #5079
Fixes #5058

Description of changes:
- Implements a GPU EK solver
- Poisson Solver using gpu heffte FFT
- Added a accessor for the electrostatic potential of the EK
- Added a flux accessor to obtain the flux vector of a species
- EK reaction ported to GPU
- Added testcases for electric potential and ion conductivity


src/walberla_bridge/tests/EKinWalberlaImpl_unit_tests.cpp
```diff
@@ -74,7 +74,11 @@ BOOST_DATA_TEST_CASE(dimensions, bdata::make(all_eks()), ek_generator) {
 }
 
 BOOST_AUTO_TEST_CASE(stencil_size) {
-  auto constexpr stencil_size = std::size_t{9u};
+  // Does it makes sense to check  for a stencil of 9 (D3Q19) as advection
+  // needs all 27 direction to operate properly anyway. I we want it we need
+  // to add the flux accesses for these different FluxCounts
+  // auto constexpr stencil_size = std::size_t{9u};
+  auto constexpr stencil_size = std::size_t{13u};
```
reinaual: The number you put here is not physically inspired, it's a unit test for the stencil size. The discretization (and the codegen) should work for any number of fluxes, only the error order changes."
```suggestion
  auto constexpr stencil_size = std::size_t{13u};
```

src/script_interface/walberla/EKPoissonSolverSlice.hpp
```diff
@@ -0,0 +1,78 @@
...
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+namespace ScriptInterface::walberla {
+
+class EKPoissonSolverSlice : public LatticeSlice<EKFieldSerializer> {
+  using LatticeModel = ::walberla::PoissonSolver;
+  std::shared_ptr<LatticeModel> m_ek_poisson_solver;
+  std::shared_ptr<EKPoissonSolver> m_ek_sip;
```
reinaual: Same as above. I think we should rename the member.

src/script_interface/walberla/EKPoissonSolverSlice.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+Variant EKPoissonSolverSlice::do_call_method(std::string const &name,
+                                             VariantMap const &params) {
+  if (name == "get_slice_size") {
+    return {m_slice_upper_corner - m_slice_lower_corner};
+  }
+  if (name == "get_slice_ranges") {
+    return {std::vector<Variant>{m_slice_lower_corner, m_slice_upper_corner}};
+  }
+  if (name == "get_ek_sip") {
```
reinaual: I find the name `get_ek_sip` a bit confusing. I had to check the type to see that it's not the EK, but the poisson-solver. Maybe it's more clear to rename it to something more clear, like `get_ek_solver_sip` or `get_solver_sip`?

src/script_interface/walberla/EKReaction.hpp
```diff
@@ -62,6 +62,13 @@ class EKReaction : public AutoParameters<EKReaction, LatticeIndices> {
     return get_value<double>(lattice->get_parameter("agrid"));
   }
 
+  auto get_is_gpu(VariantMap const &args) const {
+    auto reactants = get_value<std::vector<Variant>>(args, "reactants");
+    auto reactant =
+        get_value<std::shared_ptr<EKReactant>>(reactants[0])->get_instance();
```
reinaual:
```suggestion
    auto const reactants = get_value<std::vector<Variant>>(args, "reactants");
    auto const reactant =
        get_value<std::shared_ptr<EKReactant>>(reactants[0])->get_instance();
```

src/walberla_bridge/include/walberla_bridge/electrokinetics/PoissonSolver/FFT_GPU.hpp
```diff
@@ -0,0 +1,224 @@
...
+
+#include "../../../../src/electrokinetics/generated_kernels/EK_FieldAccessors_double_precision.h"
+#include "../../../../src/electrokinetics/generated_kernels/EK_FieldAccessors_double_precision_CUDA.cuh"
+#include "../../../../src/electrokinetics/generated_kernels/EK_FieldAccessors_single_precision.h"
+#include "../../../../src/electrokinetics/generated_kernels/EK_FieldAccessors_single_precision_CUDA.cuh"
+#include "../../BlockAndCell.hpp"
+
+#include <blockforest/communication/UniformBufferedScheme.h>
+#include <domain_decomposition/BlockDataID.h>
+#include <fft/Fft.h>
```
reinaual: Do we really need all the includes? At least the `#include <fft/Fft.h>` seems unneeded to me, since we don't use walberla's FFT for this one
```suggestion
```

src/walberla_bridge/src/LatticeModel.cpp
```diff
@@ -45,7 +45,7 @@ std::shared_ptr<VTKHandle> LatticeModel::create_vtk(
   auto const &blocks = get_lattice().get_blocks();
   auto const write_freq = (delta_N) ? static_cast<unsigned int>(delta_N) : 1u;
   auto vtk_obj = walberla::vtk::createVTKOutput_BlockData(
-      blocks, identifier, uint_c(write_freq), uint_c(0), false, base_folder,
+      blocks, identifier, uint_c(write_freq), uint_c(0), true, base_folder,
```
reinaual, line 48: Why do we force an unstructured output now?

maintainer/walberla_kernels/walberla_ek_generation.py
```diff
@@ -0,0 +1,86 @@
...
+    from pystencils_walberla.jinja_filters import add_pystencils_filters_to_jinja_env
+
+    default_dtype = config.data_type.default_factory()
+
+    # TODO Find a better way to create these stencils lists/dirs
+    # Directions according to walberla directions.h
+    stencils = ["C", "N", "S", "W", "E", "T", "B",
+                "NW", "NE", "SW", "SE", "TN", "TS",
+                "TW", "TE", "BN", "BS", "BW", "BE",
+                "TNE", "TNW", "TSE", "TSW", "BNE", "BNW", "BSE", "BSW"]
```
reinaual: I guess you can get these easily with lbmpy.
`stencils = list(map(ps.stencil.offset_to_direction_string, lbmpy.stencils.LBStencil("D3Q27")))` gives `['C', 'N', 'S', 'W', 'E', 'T', 'B', 'NW', 'NE', 'SW', 'SE', 'TN', 'TS', 'TW', 'TE', 'BN', 'BS', 'BW', 'BE', 'TNE', 'TNW', 'TSE', 'TSW', 'BNE', 'BNW', 'BSE', 'BSW']`

maintainer/walberla_kernels/walberla_ek_generation.py
```diff
@@ -0,0 +1,86 @@
...
+    # Staggered directions according to pystencils field.py
+    staggeredStencils = {"W": 0, "S": 1, "B": 2,
+                         "SW": 3, "NW": 4, "BW": 5,
+                         "TW": 6, "BS": 7, "TS": 8,
+                         "BSW": 9, "TSW": 10, "BNW": 11, "TNW": 12}
```
reinaual: `ps.field.Field.create_generic("test", spatial_dimensions=3, index_shape=(13,), field_type=ps.field.FieldType.STAGGERED_FLUX).staggered_stencil`?

maintainer/walberla_kernels/walberla_ek_generation.py
```diff
@@ -0,0 +1,86 @@
...
+    # Inverse directions to the staggered access
+    invStencils = {"C": -1, "N": 1, "S": 1, "W": 0, "E": 0, "T": 2, "B": 2,
+                   "NW": 4, "NE": 3, "SW": 3, "SE": 4, "TN": 7, "TS": 8,
+                   "TW": 6, "TE": 5, "BN": 8, "BS": 7, "BW": 5, "BE": 6,
+                   "TNE": 9, "TNW": 12, "TSE": 11, "TSW": 10, "BNE": 10,
+                   "BNW": 11, "BSE": 12, "BSW": 9}
```
reinaual: Can't help with that, since I don't understand where the numbers are coming from

maintainer/walberla_kernels/templates/EK_FieldAccessors.tmpl.cu
```diff
@@ -0,0 +1,573 @@
...
+                    j_out[1u] += add_flux;
+                {% elif "S" in Stencils[i] -%}
+                    j_out[1u] -= add_flux;
+                {% endif -%}
+                {% if "T" in Stencils[i] -%}
+                    j_out[2u] += add_flux;
+                {% elif "B" in Stencils[i] -%}
+                    j_out[2u] -= add_flux;
+                {% endif -%}
+            {% endfor %}
```
reinaual, line 462: Is there any way we can use sympy and the codegen to get an expression? Or maybe we can use the offsets from the stencil directly instead of inverting the strings to offsets in the template? I guess that would get rid of all the complexity with the letters in the stencil string

maintainer/walberla_kernels/walberla_ek_generation.py
itischler: thanks that helped

maintainer/walberla_kernels/walberla_ek_generation.py
itischler: thanks that helped

maintainer/walberla_kernels/walberla_ek_generation.py
itischler: I was able to figure something out

src/walberla_bridge/src/LatticeModel.cpp
itischler: i did this because the espresso vtk reader used in the lattice_vtk.py test was nor able to operate on the generated vti files for the structured potentials. The boundaries make the flux and the density vtk unstructured, which lets the vtk reader to work normally. So i thought making them unstructured should be fine, as it is consistent.

reinaual: Offline discussion: the presence of boundaries changes the file format from structured to unstructured. Since the FFT solver does not allow for any boundaries, it always chooses the structured format, in contrast to the LB/EK formats, that are unstructured in this case. This leads to incompatibilities and probably surprises the user. With this change, all fields are always written in the unstructured format, which effectively only changes the case when there is no boundary present. FYI @jngrad

jngrad: Thanks for the heads-up. The benchmarks were done with unstructured grids, and the testsuite only checked unstructured grids. Here is the relevant code in waLBerla:
```c++
   if( !configured_ ) {
      if( !forcePVTU_ && cellInclusionFunctions_.empty() && cellExclusionFunctions_.empty() &&
         blockStorage_->getNumberOfLevels() == 1 && ghostLayers_ == 0 ) // uniform data -> vti
      {
         uniformGrid_ = true;
      }
      configured_ = true;
   }
```
Somehow we always ended up taking the false branch, even though we have no ghost layer, 1 level of recursion, and never forced PVTU. That's because we always add a cell exclusion callback for boundaries, even when no boundaries are present. I need to implement a VTI reader in a follow-up PR, because the VTI format is three times smaller in size than the VTU format, and this is sensible in a HPC context (such as in the MultiXscale demonstrator).

maintainer/walberla_kernels/templates/EK_FieldAccessors.tmpl.cu
itischler: I now use the direction instead of the strings, which makes in cleaner

jngrad (approved):
Thanks a lot!

Issue #5068, opened 2025-03-31
LB cells and slices do not check for valid property-setters


reinaual:
Our LB nodes and slices allow the user to set any value without throwing an error. The getters do throw.

```python
import espressomd
import espressomd.lb


system = espressomd.System(box_l=[3, 3, 3])
system.cell_system.skin = 0.1

lbf = espressomd.lb.LBFluidWalberla(kinematic_viscosity=1.0, density=1.0, agrid=1.0, tau=1.0, single_precision=False)
system.lb = lbf

# None of these should work...
lbf[1,1,1].whatever_you_want = [1.0] * 19
lbf[1,:,1].whatever_you_want = [1.0] * 19

# this does not work, as expected
print(lbf[1,:,1].whatever_you_want)
```

jngrad:
This is actually true for all script interface classes. In addition, the attribute is *volatile*, in the sense that it only exists on the script interface instance where it was set:
```python
import espressomd
system = espressomd.System(box_l=[3, 3, 3])
wca=system.non_bonded_inter[0,0].wca
wca.my_value = 5
print(wca.my_value) # shows 5
print(system.non_bonded_inter[0,0].wca.my_value) # AttributeError
```
There are rare exceptions in custom classes, for example `lbf.my_value` and `system.lb.my_value` both refer to the same memory location.

Solving this issue requires a clever use of `hasattr` in the setter, such as:
```diff
--- a/src/python/espressomd/script_interface.pyx
+++ b/src/python/espressomd/script_interface.pyx
@@ -463,4 +463,9 @@ class ScriptInterfaceHelper(PScriptInterface):
 
     def __setattr__(self, attr, value):
+        if not hasattr(self, attr):
+            raise AttributeError(
+                f"Object '{self.__class__.__name__}' has no attribute '{attr}'; "
+                "ESPResSo script interface objects are thin wrappers for C++ "
+                "objects and cannot store extra data in user-defined attributes")
         if attr in self._valid_parameters():
             self.set_params(**{attr: value})
```

Then we would have to adjust a few edge cases in the script interface where we intentionally use volatile attributes. But the biggest change here would be requiring Cython 3.1. Previous versions of Cython implement the `hasattr` behavior from Python 2, which is known to be broken: instead of capturing `AttributeError` to detect the attribute absence, it captures all errors, even `exit(1)`. MWE:

```python
import espressomd.constraints
dubious_constraint = espressomd.constraints.ShapeBasedConstraint()
[...]

PR #5102, opened 2025-06-24
Fix testsuite

jngrad:
Description of changes:
-  use Ninja everywhere in CI/CD
- fix regressions in Python tests

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -398,12 +398,12 @@ installation:
   script:
     - bash maintainer/CI/build_cmake.sh
     - cd build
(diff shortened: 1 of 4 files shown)


jngrad:
This PR fixes some of the issues revealed by a prototype I wrote to explore the technical feasibility of #5068.

It also fixes a regression detected by the EESSI build pipeline (https://github.com/multixscale/dev.eessi.io-espresso/pull/1): if we're not in a git clone (e.g. we downloaded a release tarball), the git state is an empty string.

reinaual (approved):

PR #5104, opened 2025-06-26
Improve MPI and CUDA portability

jngrad:
Description of changes:
- address compiler diagnostics from GCC 14.1 and nvcc 12.9
- modernize and simplify functions that collect GPU metadata
- remove hardcoded hostname and device name string lengths
- handle `OMP_NUM_THREADS` default value at the C++ level


jngrad:
Note to reviewers: multi-GPU support is still a work-in-progress. The fact that `list_devices` only returns devices on the head node is justified by its historic use case for single-GPU simulations: it's meant to help find out on which GPU device to pin when more than one GPU is available, if one of them is too old to be supported (e.g. it's only used for the LCD display). Figuring out a better Python interface to query GPU metadata is a task for another PR.

reinaual (dismissed):

reinaual (approved):

PR #5105, opened 2025-07-01
dpd_pair_force without particle params

NiklasMueller2001:
The `dpd_pair_force function()` currently takes two references to `Particle` objects as input parameters. Due to a future change in the data model of `Particle` structs, this function should rather take references to data members.


RudolfWeeber (approved):

RudolfWeeber:
@jngrad I think this pr is ready.

PR #5109, opened 2025-07-08
Improve resilience to precision loss

### LINKED ISSUE #5108: LB tests tolerances are too strict
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/pipelines/24423

jngrad:
Fixes #5108

Description of changes:
- properly fold coordinates in the histogram and RDF classes to avoid off-by-one errors
- adjust test tolerances to better reflect recent changes to waLBerla and heFFTe algorithms

### DIFF
--- a/src/utils/include/utils/Histogram.hpp
+++ b/src/utils/include/utils/Histogram.hpp
@@ -165,7 +165,7 @@ private:
     return std::ranges::all_of(pos, [&it_limits](U const value) {
       auto const [lower, upper] = *it_limits;
       ++it_limits;
-      return value >= lower and value <= upper;
+      return value >= lower and value < upper;
     });
   }
 
--- a/src/utils/include/utils/Histogram.hpp
+++ b/src/utils/include/utils/Histogram.hpp
@@ -105,13 +105,10 @@ public:
       throw std::invalid_argument("Wrong dimensions for the value");
     }
     if (check_limits(pos)) {
-      boost::array<array_index, M + 1> index;
-      for (std::size_t i = 0; i < M; ++i) {
-        index[i] = calc_bin_index(pos[i], m_limits[i].first, m_bin_sizes[i]);
-      }
-      for (array_index i = 0; i < static_cast<array_index>(N); ++i) {
-        index.back() = i;
-        m_array(index) += value[static_cast<std::size_t>(i)];
+      auto index = calc_bin_index(pos);
+      for (std::size_t i = 0; i < N; ++i) {
+        index.back() = static_cast<array_index>(i);
+        m_array(index) += value[i];
         m_count(index)++;
       }
     }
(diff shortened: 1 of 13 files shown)


jngrad:
The waLBerla bridge is now tested on ARM. That wasn't possible before due to a long-standing round-off bug in the histogram class (logfile [45562052553](https://github.com/jngrad/espresso/actions/runs/16145194460/job/45562052553)) that use to trigger a write past-the-end in LB profile observables.

src/utils/include/utils/Histogram.hpp
```diff
@@ -153,7 +172,7 @@ class Histogram {
    * \brief Check if the position lies within the histogram limits.
    * \param pos     Position to check.
    */
-  bool check_limits(std::span<const U> pos) const {
+  bool check_limits(std::span<const U> const &pos) const {
```
reinaual, line 162: Don't we want to use an early breakout here? Or probably even better `std::ranges::all_of`?

src/utils/include/utils/Histogram.hpp
```diff
@@ -121,20 +118,42 @@ class Histogram {
...
 
+protected:
+  // internal function exposed for unit testing purposes
+  long correct_off_by_one(long bin, long n_bins) const {
+    if (bin == -1l) {
+      ++bin;
+    } else if (bin == n_bins) {
+      --bin;
+    }
+    return bin;
```
reinaual: Why not use `std::clamp` and check in the assert [`calc_bin_index`](https://github.com/espressomd/espresso/pull/5109/commits/a7d55cb18558950b3bc87ac33af11a3fa745b332#diff-d0b0322fc635ab11d4dd280b913d2a94fb766a672bf4c01d5a5544a405736499R153) that we're in range? I think that way we can also drop the testcase. What do you think?

src/utils/include/utils/Histogram.hpp
```diff
@@ -153,19 +159,19 @@ class Histogram {
...
-    for (std::size_t i = 0; i < M; ++i) {
-      if (pos[i] < m_limits[i].first or pos[i] >= m_limits[i].second)
-        within_range = false;
-    }
-    return within_range;
+    auto it_limits = m_limits.begin();
+    return std::ranges::all_of(pos, [&it_limits](U const value) {
+      auto const [lower, upper] = *it_limits;
+      ++it_limits;
+      return value >= lower and value <= upper;
```
reinaual: Looking at the code, previously, we used $[\mathrm{lower}, \mathrm{upper})$ as the valid range. No idea if there are any side effects if we leave it in...

```suggestion
      return value >= lower and value < upper;
```

reinaual (changes requested):

src/utils/include/utils/Histogram.hpp
jngrad: indeed, and we can do it in the RDF class too, which duplicated the same code

src/utils/include/utils/Histogram.hpp
jngrad: good point, this boundary check was fixed and is now properly documented in the user guide

src/utils/include/utils/Histogram.hpp
```diff
@@ -121,20 +118,29 @@ class Histogram {
...
    */
-  array_index calc_bin_index(double value, double offset, double size) const {
-    return static_cast<array_index>(std::floor((value - offset) / size));
+  auto calc_bin_index(std::span<const U> const &pos) const {
+    boost::array<array_index, M + 1> index;
+    for (std::size_t i = 0; i < M; ++i) {
+      auto const offset = m_limits[i].first;
+      auto const size = m_bin_sizes[i];
+      auto const n_bins = static_cast<long>(m_n_bins[i]);
+      auto const bin = static_cast<long>(std::floor((pos[i] - offset) / size));
```
reinaual, line 137: Is there a reason why we use `long` for the index and not `std::size_t`?

jngrad: to be able to detect -1, otherwise it would be a large positive value and would be folded on the right side of range instead of the left side

reinaual (approved):

PR #5123, opened 2025-07-22
LB interpolation

jngrad:
Description of changes:
- fix bug in particle coupling near LB boundaries
- improve performance of LB profile observables
   - make a single GPU kernel call when calculating a LB histogram
   - pre-calculate sampling positions for LB histograms that don't depend on particle positions


jngrad:
Benchmarks: `LBVelocityProfile` is 52% faster on CPU, `CylindricalLBVelocityProfile` is 20% faster on CPU. Both are now several orders of magnitude faster on GPU. Cylindrical LB GPU observables at particle positions now scale better with the number of particles.

For the benchmarks, add this to `maintainer/benchmarks/lb.py`:
```py
import espressomd.observables
params = {
    'n_x_bins': int(system.box_l[0] / agrid),
    'n_y_bins': int(system.box_l[1] / agrid),
    'n_z_bins': int(system.box_l[2] / agrid),
    'min_x': 0.0, 'min_y': 0.0, 'min_z': 0.0,
    'max_x': system.box_l[0],
    'max_y': system.box_l[1],
    'max_z': system.box_l[2],
    'sampling_delta_x': agrid,
    'sampling_delta_y': agrid,
    'sampling_delta_z': agrid,
    'sampling_offset_x': 0.5 * agrid,
    'sampling_offset_y': 0.5 * agrid,
    'sampling_offset_z': 0.5 * agrid,
    'allow_empty_bins': False}
obs = espressomd.observables.LBVelocityProfile(**params)
acc = espressomd.accumulators.TimeSeries(obs=obs, delta_N=1)
system.auto_update_accumulators.add(acc)
```

or:

```py
import espressomd.observables
cyl_transform_params = espressomd.math.CylindricalTransformationParameters(
        center=system.box_l / 2, axis=[1, 0, 0], orientation=[0, 0, 1])
params = params = {
    'transform_params': cyl_transform_params,
    'n_r_bins': 12,
    'n_phi_bins': 12,
    'n_z_bins': 12,
    'min_r': 0.0, 'max_r': 20.0,
    'min_z': -10.0, 'max_z': 10.0,
    'min_phi': -np.pi, 'max_phi': np.pi,
    'sampling_density': 2,
}
obs = espressomd.observables.CylindricalLBVelocityProfile(**params)
acc = espressomd.accumulators.TimeSeries(obs=obs, delta_N=1)
system.auto_update_accumulators.add(acc)
```

Run with:

```sh
OMP_NUM_THREADS=1 OMP_PROC_BIND=true OMP_PLACES=0 mpiexec -n 1 --bind-to core \
  ./pypresso ../maintainer/benchmarks/lb.py --particles_per_core 0 --box_l=48
```

---

The boundary slip velocity is written directly to the velocity field. This is safe, because the velocity and pdf fields contain undefined values inside the boundary regions anyway. The cost of overwriting the velocity field on GPU is negligible. We could re-implement the `std::unordered_map` as a `cuco::static_map` from cuCollections, however this library is still under active development. Its maintainers are currently introducing Clang in CI and migrating the hash maps to the new API. Once this refactoring is complete, they will publish the first release of cuCollections.

RudolfWeeber:
In general, lgtm.
Few points:
* Are we sure we don't need to update the velocity field for other cases?
* Please add some comments explaining what's happening in the velocities at pos getter in LBWalberlaImpl. Maybe also more clear variabe names in place of d_idx and d_ubb (boundary_indices, boundary_velocities or similar).
* Is Interpolation the right name space for the set_from_list kernel? as far as I understand, it doesn't interpolate?

jngrad:
> * Are we sure we don't need to update the velocity field for other cases?

I don't think so. Velocity-related kernels call the node getter, which takes care of the slip velocity. The slice getter already takes care of the slip velocity. VTK writers use the boundary field as a mask to filter out indeterminate values.
 
> * Please add some comments explaining what's happening in the velocities at pos getter in LBWalberlaImpl. Maybe also more clear variabe names in place of d_idx and d_ubb (boundary_indices, boundary_velocities or similar).

Done. I'm using the `dev_` prefix on device memory containers to mirror the `host_` prefix on host memory containers.

> * Is Interpolation the right name space for the set_from_list kernel? as far as I understand, it doesn't interpolate?

I've moved it to a more appropriate namespace.

src/core/observables/CylindricalLBFluxDensityProfileAtParticlePositions.cpp
```diff
@@ -48,21 +48,31 @@ CylindricalLBFluxDensityProfileAtParticlePositions::evaluate(
   auto &system = System::get_system();
   auto const &box_geo = *system.box_geo;
   auto &lb = system.lb;
-  auto const vel_conv = lb.get_lattice_speed();
   lb.ghost_communication_pdf();
   lb.ghost_communication_vel();
 
+  std::vector<Utils::Vector3d> unfolded_pos{};
+  std::vector<Utils::Vector3d> folded_pos{};
+  folded_pos.reserve(local_particles.size());
```
reinaual, line 56:
```suggestion
  folded_pos.reserve(local_particles.size());
  unfolded_pos.reserve(local_particles.size());
```

src/core/observables/CylindricalLBFluxDensityProfileAtParticlePositions.cpp
```diff
@@ -48,21 +48,31 @@ CylindricalLBFluxDensityProfileAtParticlePositions::evaluate(
...
   for (auto const &p : local_particles) {
-    auto const pos = box_geo.folded_position(traits.position(p));
+    unfolded_pos.emplace_back(traits.position(p));
+    folded_pos.emplace_back(box_geo.folded_position(traits.position(p)));
+  }
+  auto const interpolated_vel =
+      lb.get_coupling_interpolated_velocities(folded_pos);
+  auto const interpolated_rho = lb.get_interpolated_densities(unfolded_pos);
```
reinaual, line 63: What is the reason we use folded coordinates for the velocity and unfolded for the density? We used both folded before, was this a bug?

src/walberla_bridge/src/lattice_boltzmann/LBWalberlaImpl.hpp
```diff
@@ -1146,7 +1186,14 @@ class LBWalberlaImpl : public LBWalberlaBase {
       auto const gl = lattice.get_ghost_layers();
       auto field =
           block.template uncheckedFastGetData<VectorField>(m_velocity_field_id);
-      auto const res = lbm::accessor::Interpolation::get(field, host_pos, gl);
+      // the velocity field has indeterminate values inside boundary regions;
+      // we overwrite them with boundary slip velocities before interpolation
+      auto const [dev_idx, dev_vel] = m_boundary->get_flattened_map_device();
```
reinaual, line 1191: Why is this not an issue on CPU? And is there a reason we decided against using an additional boundary handling for the velocity field from walberla?

I have the impression that we're introducing more and more code and complexity for the fact that we're not maintaining meaningful values in the fluid fields for cells that have no meaning for LB, which are not maintained by walberla for performance reasons (e.g. proxying the access and checking if it is a boundary or here, correcting the velocity field). To me, it seems like we would remove a lot of code and complexity by introducing additional "cosmetic" boundary handlings from walberla with Dirichlet boundary conditions for the velocity field and possibly also the density field, which could be called in the integration loop. This would then remove the need for proxying the access to the fields to check if it is a boundary, and would also fix the issue being targeted with this PR with probably significantly less code. What do you think @jngrad?

RudolfWeeber (approved):

src/core/observables/CylindricalLBFluxDensityProfileAtParticlePositions.cpp
jngrad: Yeah it's really confusing. The `LBSolver` class sometimes expects unfolded coordinates and will fold them for us, sometimes expects already folded positions. I think in the future we should give every method in `LBSolver` that deals with positions an additional boolean flag to document whether positions need folding. Likewise, some methods return quantities in LB units, other times in MD units. I updated the docstrings accordingly, but we probably want a more fool-proof solution.

src/walberla_bridge/src/lattice_boltzmann/LBWalberlaImpl.hpp
jngrad: I initially created an extra FlagField for the GPU, but I got weird compiler errors. I could not find a waLBerla function to fill `uint8_t` on a GPU field. I then made some attributes in the `DynamicUBB` class public, because there we already have the flat map in thrust device vectors. However, if I understood the code correctly, these vectors don't contain the halo region, because we wrote an optimized communicator that only transfers the bounce-back field itself, and not the vectors used to pre-calculate the bounce-back field (the `OUTER` vector is empty). I'm not entirely satisfied by the extra complexity introduced here, but I also don't have much time left in my budget for the waLBerla integration.

src/core/observables/CylindricalLBFluxDensityProfileAtParticlePositions.cpp
jngrad: Good point, I'll add this to my next refactoring PR.

PR #5124, opened 2025-07-24
Batch particle properties update

NiklasMueller2001:
Until now the `update()` method of the `ParticleHandle` class just sequentially set its parameters to new values, making one expensive update call for each new parameter. Now a batch update is performed.

### DIFF
--- a/src/python/espressomd/particle_data.py
+++ b/src/python/espressomd/particle_data.py
@@ -605,6 +605,14 @@ class ParticleHandle(ScriptInterfaceHelper):
                 else:
                     self.propagation |= Propagation.ROT_LANGEVIN
 
+    def _bond_sanity_checks(self, bond):
+        if self.id in bond[1:]:
+            raise Exception(
+                f"Bond partners {bond[1:]} include the particle {self.id} itself")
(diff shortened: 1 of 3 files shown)


jngrad:
Overall looks good to me. I'm currently working on this PR to deduplicate some of the common code.

Is there a particular reason to prefer the current syntax
```python
p.update({"bonds": new_bonds, "pos": [1, 2, 3]})
```
over the following syntax?
```python
p.update(bonds=new_bonds, pos=[1, 2, 3])
```

The latter is closer to the syntax of `system.part.add()`, and wouldn't require `additional_properties` to store overridden values. The function would take `**new_properties` as argument, so it would be safe to override a value without affecting the parent scope. I had to look up PEP 584 to double-check the pipe operator uses the last occurrence of a key/value pair in the return value (nothing against it, the old union syntax with expansions is also cryptic in my opinion, but removing unions makes the code easier for newcomers).

jngrad:
I see: the `system.part.update()` takes a dict as argument, so we mimic that in the particle class.

jngrad (approved):
Thanks!

Issue #5119, opened 2025-07-17
Reduce Boost footprint


jngrad:
This tracks our efforts to reduce Boost's footprint in ESPResSo.

* [ ] `boost::mpi`: see tracking ticket #4833
* [ ] `boost::serialization`: can probably be replaced by cereal
* [x] `boost::filesystem`: can be replaced by C++17 `std::filesystem`
* [x] `boost::string_ref`: replaced by C++17 `std::string_view`
* [x] `boost::transform`: replaced by C++20 range-based `std::transform`
* [x] `boost::optional`: replaced by C++17 `std::optional`
* [x] `boost::variant`: can be replaced by C++17 `std::variant`, recursive variant requires GCC >= 12
* [ ] `boost::accumulate`: can be replaced by C++23 range-based `std::fold_left`
* [ ] `boost::unit_test_framework` : needed for unit tests
* [ ] `boost::multi_array`: needed for histograms
* [ ] `boost::qvm`: needed for quaternions

RudolfWeeber:
I think we might limit the scope a bit, here to
* stuff that becomes available in the stl
* boost compoents that cause significant challenges (such as boost-mpi, which is often not pre-uilt on clusters)


When it comes to the unit test framework, the quaternions and the histograms, I'm not sure there is much to gain by using somthign else.



*

PR #5130, opened 2025-08-01
New variant implementation

jngrad:
Partial fix for #5119

Description of changes:
- roll out `std::variant` everywhere

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -805,11 +805,6 @@ endif()
 
 find_package(Boost 1.83.0 REQUIRED ${ESPRESSO_BOOST_COMPONENTS})
 
-# enable boost::variant with more than 20 types
-target_compile_options(
-  espresso_cpp_flags INTERFACE -DBOOST_MPL_CFG_NO_PREPROCESSED_HEADERS
-                               -DBOOST_MPL_LIMIT_LIST_SIZE=30)
-
 #
 # Paths
 #
--- a/src/core/bond_breakage/actions.hpp
+++ b/src/core/bond_breakage/actions.hpp
@@ -20,7 +20,6 @@
 #pragma once
 
 #include <boost/functional/hash.hpp>
-#include <boost/variant.hpp>
 
 #include <array>
 #include <cstddef>
@@ -74,7 +73,7 @@ struct DeleteAllBonds {
 } // namespace BondBreakage
 
 // Hash support for std::unordered_set
-namespace boost {
+namespace std {
 template <> struct hash<BondBreakage::DeleteBond> {
   std::size_t operator()(BondBreakage::DeleteBond const &t) const noexcept {
     return t.hash_value();
@@ -91,4 +90,4 @@ template <> struct hash<BondBreakage::DeleteAllBonds> {
     return t.hash_value();
   }
 };
-} // namespace boost
+} // namespace std
--- a/src/core/bond_breakage/bond_breakage.cpp
+++ b/src/core/bond_breakage/bond_breakage.cpp
@@ -30,18 +30,18 @@
 
 #include <boost/mpi.hpp>
 #include <boost/serialization/access.hpp>
-#include <boost/variant.hpp>
 
 #include <cassert>
 #include <memory>
 #include <unordered_set>
 #include <utility>
+#include <variant>
 #include <vector>
 
 namespace BondBreakage {
 
 // Variant holding any of the actions
-using Action = boost::variant<DeleteBond, DeleteAngleBond, DeleteAllBonds>;
+using Action = std::variant<DeleteBond, DeleteAngleBond, DeleteAllBonds>;
 
 // Set of actions
 using ActionSet = std::unordered_set<Action>;
@@ -160,7 +160,7 @@ static void remove_pair_bonds_to(Particle &p, int other_pid) {
 }
 
 // Handler for the different delete events
-class execute : public boost::static_visitor<> {
+class execute {
   CellStructure &cell_structure;
 
 public:
@@ -199,7 +199,7 @@ void BondBreakage::process_queue_impl(System::System &system) {
 
   // Execute actions
   for (auto const &a : actions) {
-    boost::apply_visitor(execute(cell_structure), a);
+    std::visit(execute(cell_structure), a);
     system.on_particle_change();
   }
 }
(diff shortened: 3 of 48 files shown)


jngrad:
I think this is how one implements recursive variants with the C++ standard library. It's not a widely used design pattern, and the Boost library relies on a radically different design, so I'll give an overview of the PR implementation here.

The traditional variant uses a [tagged union](https://en.wikipedia.org/wiki/Tagged_union) data structure, where the union stores any object whose type belongs to the variant type list, and the tag indicates the type of the currently held object. Here is how the containers are implemented in the STL and Boost, in pseudo-code:

```c++
// factory to dynamically construct a C++ union type that can hold
// any value whose type appears in the union template parameters
template<class First, class... Rest>
union variadic_union {
  variadic_union(First &&value, Rest&&... rest)
    : m_first(std::forward<First>(value)),
      m_rest(std::forward<Rest>(rest)...) {}

  First m_first;
  variadic_union<Rest...> m_rest;
};

template <class... Ts>
struct std::variant {
  union variadic_union<Ts...> data;
  uint8_t type_index;
};

template <class... Ts>
struct boost::variant {
  uint32_t type_index;
  union variadic_union<Ts...> data;
};
```

The variant type list `Ts...` cannot contain the variant itself for two reasons. First, the variant type is not known before the type list ends, so one must use a CRTP. Second, since the union is as large as the largest element it can store, and the integer tag has a non-zero size, the recursive variant has an infinite size at compile time. That's because nothing prevents us from writing e.g. `Variant{Variant{Variant{Variant{value}}}}` with an arbitrary depth level, and the compiler must provide a valid constructor for any recursion depth, thus the size is not bounded.

The Boost library implements recursive variants by replacing the variant type by a pointer to the variant type, i.e. `boost::make_recursive_variant<int, double, boost::recursive_variant_>::type` yields a type that is almost like `boost::variant<int, double, this*>` internally, but uses clever metaprogramming techniques to hide the pointer from the type list (visitors that match against the variant implicitly dereference the underlying pointer). This trick comes with a performance penalty, which becomes noticeable when passing large 3D arrays of doubles to the core via LB/EK setters and getters, and that's one of the reasons why we added `std::vector<int>` and `std::vector<double>` specializations to the list of type alternatives, despite `std::vector<Variant>` being already in the type list.

The STL doesn't provide this design pattern, which means the variant can only appear in its own type list in the form of a pointer. Which is fine with ESPResSo, since we only need the variant inside STL containers like vectors and maps, both of which hold a pointer to a data structure in heap memory. The PR implementation looks like this in pseudo-code:

```c++
template <class... Ts>
struct recursive_variant
  : public std::variant<Ts...,
                        std::vector<recursive_variant<Ts...>>,
                        std::unordered_map<int, recursive_variant<Ts...>>,
                        std::unordered_map<std::string, recursive_variant<Ts...>>> {
  using BaseClass = std::variant<Ts..., ___> // here repeat the full signature of the base class
  using BaseClass::BaseClass;
private:
  friend class boost::serialization::access;
  template <typename Archive>
  void serialize(Archive &ar, unsigned const) {
    BaseClass &self = *this;
    ar & self;
  }
};
```

One pain point with this solution, is that like many CRTPs, one has to spell out the base class twice: once in the inheritance list, and once in a typedef so it can be referred to in the constructor and in the self. To avoid code duplication, one can declare the base class as a templated typedef outside the class, and forward-declare the CRTP class to be able to use it in the templated typedef [^1]. This works for simple cases, but when the forward-declared type is part of a STL template instance, type information can be lost. This is unfortunately the case with the STL of GCC 11, where the `std::pair` iterator of `std::unordered_map` complains about the `value_type` being forward declared; this was resolved in the STL shipped with GCC 12. I couldn't find the proposal that explains how this was fixed, but for what it's worth, a similar-looking issue in `std::stack` iterators was addressed by [P1425R4](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p1425r4.pdf). Same problem with Intel Classic: not supported in icx 2022 but fixed in icx 2023. The PR implementation works against both libstdc++ and libc++. Tested in the Compiler Explorer with GCC 12–15, Clang 18–22, Intel Classic icx 2023–2025, MSVC 19, nvc++ 22–25, and nvcc 12.0–12.9 using gcc 12–13 as host compiler.

[^1]: While the C++ type trait `std::tr2::direct_bases` was proposed in N2965 to provide syntactic sugar to solve this exact issue with CRTPs, the proposal was rejected.

RudolfWeeber:
The replacement of the non-recursive variants in the core is probably uncontroversial. However, when it comes to the recursive_variant in the script interface, we need to discuss whether replacing the relatively clear boost::make_recursive_variant with very advanced custom C++ that few people will be able to maintain really improves our position in terms of sutainability.

jngrad:
I'm open to this discussion. The old implementation, while more readable, comes with its one caveats: one has to remember that every time a new type is introduced in `Variant`, it has to be mirrored in the exact same way in `PackedVariant` to avoid cryptic compiler errors, and whenever the number of types exceeds 20/30/40, one has to increment a C++ define from the Boost::mpl library (and that constant has an upper limit of 50).

jngrad:
For completeness, here is a more concise solution that uses reflections:
```c++

template <class Base> struct ReflectionGetParent : public Base {
  using Base::Base;
  using first_parent = Base;
};

template <class... Ts>
struct recursive_variant
    : public ReflectionGetParent<std::variant<
                 Ts..., std::vector<recursive_variant<Ts...>>,
                 std::unordered_map<int, recursive_variant<Ts...>>,
                 std::unordered_map<std::string, recursive_variant<Ts...>>>> {
  using ParentClass = typename recursive_variant<Ts...>::first_parent;
  using BaseClass = ReflectionGetParent<ParentClass>;
  using BaseClass::BaseClass;

private:
  friend class boost::serialization::access;

  template <typename Archive>
  void serialize(Archive &ar, unsigned const /*version*/) {
    ParentClass &self = *this;
    ar & self;
  }
};
```
Works with libstdc++ but not with libc++, which complains about `recursive_variant` template specializations being incomplete until after the struct closing brace. Taking the `serialize` declaration outside the class body doesn't help.

jngrad:
Doxygen in CI doesn't behave like on our workstations, even though the same version is used. This is getting annoying.

RudolfWeeber (approved):

PR #5136, opened 2025-08-08
Fix P3M edge case in charge assignment

### LINKED ISSUE #5135: P3M fails when a particle is close to the local domain boundary
Continuation of #4103 and related to #4825.

Reproducible on a CPU with [extended precision](https://en.wikipedia.org/wiki/Extended_precision) and with `-D CMAKE_BUILD_TYPE=RelWithAssert` (i.e. `-O3` without `-DNDEBUG`):

```python
import espressomd
import espressomd.version

ref_box_l = 5.81825
system = espressomd.System(box_l=3 * [ref_box_l])
system.time_step = 0.01
system.cell_system.skin = 0.
system.part.add(pos=[[0, 0, 0], [1, 1, 1]], q=[-1, +1])
solver = espressomd.electrostatics.P3M(
    prefactor=2., accuracy=1e-2, mesh=[10, 10, 10], cao=7,
    r_cut=1.5, alpha=29.4842, tune=False)
if espressomd.version.major() == 4:
    system.actors.add(solver)
else:
    system.electrostatics.solver = solver
```
Output:
```
python3.12: /home/user/espresso/src/utils/include/utils/math/bspline.hpp:34:
T Utils::bspline(int, T) [order = 7, T = double]: Assertion `(x >= T(-0.5))' failed.
```

Here is the source of the issue:

https://github.com/espressomd/espresso/blob/08abad428c7aaf9ebb31e151bd4bdfedf12cf295/src/core/p3m/interpolation.hpp#L152-L160

The calculation of `pos` at line 154 is prone to [catastrophic cancellation](https://en.wikipedia.org/wiki/Catastrophic_cancellation), specifically [radix conversion](https://en.wikipedia.org/wiki/Catastrophic_cancellation#Example:_Radix_conversion). The `-O3` is crucial to get 80-bit floating-point arithmetic on hardware that supports it. The particle is at the origin of the local mesh, and due to accumulated precision loss, the conversion from Cartesian units to lattice units causes the particle to end up outside the local mesh by a ridiculously small amount. To emulate the issue in Python:

```py
>>> import numpy as np
>>> (np.float128('0.') - np.float128('-0.80808979848680806413')) * np.float128('3.0937155804730926967') - np.float128('2.5')
np.longdouble('-8.5651971626354850287e-17')
>>> (np.float64( '0.') - np.float64( '-0.80808979848680806413')) * np.float64( '3.0937155804730926967') - np.float64( '2.5')
np.float64(0.0)
```

Affected ESPResSo versions:
- 5.0-dev (python branch): crash when built with assertions
- 4.2: crash when built with assertions
- 4.1: warnings about underflow sent to stderr

Proposed bugfix: tweak the assertions to allow a deviation of a few machine epsilons from the lower and upper limits. The deviation is tiny (16 orders of magnitude) and unlikely to have any meaningful impact on simulation results.

jngrad:
Fixes #5135

Description of changes:
- P3M, P3MGPU and DipolarP3M no longer trigger an assertion when particles are close to a local mesh boundary

### DIFF
--- a/src/utils/include/utils/math/bspline.hpp
+++ b/src/utils/include/utils/math/bspline.hpp
@@ -22,6 +22,11 @@
 #include "sqr.hpp"
 #include "utils/device_qualifier.hpp"
 
+#if defined(__CUDACC__)
+#include <cfloat>
+#else
+#include <limits>
+#endif
 #include <stdexcept>
 
 namespace Utils {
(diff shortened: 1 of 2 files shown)


hidekb (approved):
I think this looks good to me.

PR #5137, opened 2025-08-14
Use new particle data layout in the P3M code

jngrad:


jngrad:
With the new P3M shared memory implementation, Coulomb P3M is 17% faster and Dipolar P3M is 7% faster with 2000 particles per thread. The MPI implementation is also faster now. Weak scaling benchmarks will run over the week-end.

jngrad:
The benchmarks are inconclusive due to the high variance. Running the P3M benchmark with 4 threads multiple times yields two different runtimes reliably: 3.7ms/step and 4.6ms/step. Since the difference is significant, one cannot take the mean of multiple runs. Taking the minimum (or first quartile) of multiple runs requires a larger sample size than what I currently have.

src/core/electrostatics/icc.cpp
```diff
@@ -126,6 +138,22 @@ void ICCStar::iteration(CellStructure &cell_structure,
                    elc_kernel);
     system.coulomb.calc_long_range_force(particles);
     cell_structure.ghosts_reduce_forces();
+#ifdef SHARED_MEMORY_PARALLELISM
+    // force reduction
+    int num_threads = execution_space().concurrency();
+    kokkos_parallel_range_for<Kokkos::RangePolicy<execution_space>>(
+        "reduction", std::size_t{0}, unique_particles.size(),
+        [&local_force, &unique_particles, num_threads](std::size_t const i) {
+          Utils::Vector3d force{};
```
hidekb, line 147: I think it is possible to apply `local_force` directly to `unique_particles.at(i)->force()`, and the calculation became a bit faster due to not generating Utils::Vector3d.

src/core/electrostatics/elc.hpp
```diff
@@ -115,13 +115,41 @@ struct elc_data {
...
+  void dielectric_layers_contribution(BoxGeometry const &box_geo,
+                                      std::size_t p1, std::size_t p2,
+                                      auto &aosoa, double q1q2,
+                                      auto &&kernel) const {
+    if (aosoa.position(p1, 2) < space_layer) {
+      auto const q_eff = delta_mid_bot * q1q2;
+      auto pos2 = copy_aosoa_vector_elc(p2, aosoa.position);
+      auto pos1 = copy_aosoa_vector_elc(p1, aosoa.position);
+      pos1[2] *= -1.;
+      auto const d = box_geo.get_mi_vector(pos2, pos1);
```
hidekb, line 133: When we use aosoa_pack, it would be faster to assign each value of aosoa.position directly to `get_mi_vector(T const &a0, T const &a1, T const &a2, T const &b0, T const &b1, T const &b2)` than to use `get_mi_vector(const Utils::Vector<T, 3> &a, const Utils::Vector<T, 3> &b)`.

src/core/electrostatics/elc.hpp
```diff
@@ -115,13 +115,41 @@ struct elc_data {
...
+      kernel(q_eff, d);
+    }
+    if (aosoa.position(p1, 2) > (box_h - space_layer)) {
+      auto const q_eff = delta_mid_top * q1q2;
+      auto const z = 2. * box_h - aosoa.position(p1, 2);
+      auto pos2 = copy_aosoa_vector_elc(p2, aosoa.position);
+      auto pos1 = copy_aosoa_vector_elc(p1, aosoa.position);
+      pos1[2] = 2. * box_h - pos1[2];
+      auto const d = box_geo.get_mi_vector(pos2, pos1);
```
hidekb, line 142: Same as line 133.

hidekb (approved):
I think this looks good to me. Thanks a lot for the great work.

src/core/electrostatics/elc.hpp
jngrad: Here I'm leveraging compiler optimizations that elide out the temporary Vector3d objects from the stack. Verified on the Compiler Explorer using GCC 15.2 and Clang 19.1.

src/core/electrostatics/icc.cpp
jngrad: A quick benchmark reveals no major speed-up, but I'll include that in my next PR because it improves readability.

PR #5140, opened 2025-08-19
Zn draw tester

ludogibbs:
- Switching back from ZnDrawLocal to ZnDraw as the Visualizer object to prevent occasional errors.

- Test script designed to verify that ZnDraw is properly loaded as the Visualizer object, ensuring that all default methods are correctly initialized and accessible.


RudolfWeeber:
For it to execute, in CI, you need to include the test in testsuite/python/CMakeList.txt.
It will probably also be necessary to update the docker container to include the dependencies.

src/python/espressomd/zn.py
```diff
@@ -519,8 +519,8 @@ def _start_zndraw(self):
                 time.sleep(0.5)
 
         url = f"{self.url}:{self.SERVER_PORT}"
-        self.zndraw = zndraw.zndraw.ZnDrawLocal(
-            r=self.r, url=url, token=self.token, timeout=config)
+        self.zndraw = zndraw.zndraw.ZnDraw(
+            url=url, token=self.token, timeout=config)
```
PythonFZ, line 523: I'd recommend not using internal zndraw modules and use `from zndraw import ZnDraw`.

testsuite/python/zn_interface.py
```diff
@@ -0,0 +1,58 @@
...
+            options = [x for x in elements[j].find_elements(
+                By.TAG_NAME, "option")]
+            option_names = [x.get_attribute("value") for x in options]
+            if option_names == next_option_names:
+                continue
+            else:
+                assert option_names == method_list[idx], f"Difference: {
+                    set(option_names) ^ set(method_list[idx])}"
+                idx += 1
+                next_option_names = option_names
```
PythonFZ, line 54: There is a good chance this will break for future versions of ZnDraw for several reasons
- I won't be using bootstrap anymore, see WIP https://github.com/zincware/ZnDraw/pull/799
- the options on the different drop-down menus might change over time.

Are there any ZnDraw extensions currently used by Espresso or within any tutorial?

jngrad:
Closing as we won't be able to work on this PR further due to the significant deviation from the development branch and ZnDraw 0.6.

Issue #5144, opened 2025-08-20
Collision detection fails randomly with Cabana Verlet list


jngrad:
In the 4.2 release, Verlet lists were a vector of vectors, which allowed arbitrary neighbor list lengths. With the Cabana implementation, the neighbor list is a 2D matrix, which limits the number of neighbors to a small value (close to 16).

A heuristic is used to estimate the density of particles based on the interaction range and select the size of the Verlet list. There are several scenarios where this strategy is inadequate:

* when the system is out-of-equilibrium, with particles in very close proximity (much closer than sigma), the interaction range is not a good way to estimate local density; this is often the case when using the steepest descent algorithm (see also #5141), or when setting up the system and forgetting to run a minimization
* when collision detection is active in `BindAtPointOfCollision` mode, for each particle pair in close proximity, two virtual particles with zero radius are created; so for a volume containing $N$ particles initially, we can expect $N + N^2$ particles to be present in the same volume after one time step
* dumbbells (`VirtualSites.test_lj` test case in `testsuite/python/virtual_sites_relative.py`)
* maybe other simulation setups are affected (Stokesian Dynamics? solid raspberries?)

To properly handle out-of-equilibrium and high density systems, we would need to explore other types of data structures that allow for dynamic sizing, at the expense of cache locality.

MWE  for a `-DCMAKE_BUILD_TYPE=RelWithAssert -DESPRESSO_BUILD_WITH_SHARED_MEMORY_PARALLELISM=ON` build:
```python
import espressomd
import espressomd.interactions
import numpy as np

system = espressomd.System(box_l=[2.0, 2.0, 2.0])
np.random.seed(seed=42)

bond_center = espressomd.interactions.HarmonicBond(k=5000, r_0=0.1)
bond_vs = espressomd.interactions.HarmonicBond(k=25000, r_0=0.02)
system.bonded_inter.add(bond_center)
system.bonded_inter.add(bond_vs)
time_step = 0.001
system.time_step = time_step
system.cell_system.skin = 0.05
system.min_global_cut = 0.112
system.non_bonded_inter[0, 0].wca.set_params(epsilon=1, sigma=0.05)
system.part.add(pos=np.random.random((2000, 3)))

# scenario 1: steepest descent
system.thermostat.turn_off()
system.integrator.set_steepest_descent(
    f_max=0, gamma=1, max_displacement=0.001)
system.integrator.run(10000)

# scenario 2: collision detection
[...]

PR #5145, opened 2025-08-20
Fix regressions from recently merged PRs

jngrad:
Partial fix for #5144

Description of changes:
- provide a temporary solution for issues related to the Cabana Verlet list
- gracefully handle file system latency issues that surfaced recently in nightly builds

### DIFF
--- a/src/core/integrate.cpp
+++ b/src/core/integrate.cpp
@@ -506,12 +506,12 @@ int System::System::integrate(int n_steps, int reuse_forces) {
     lb_active = lb.is_solver_set();
     ek_active = ek.is_ready_for_propagation();
 #ifdef SHARED_MEMORY_PARALLELISM
-    cell_structure->set_max_prefactor(5);
+    cell_structure->set_max_prefactor(8);
 #endif
   }
 #ifdef SHARED_MEMORY_PARALLELISM
   else {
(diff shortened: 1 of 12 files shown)


src/core/integrate.cpp
```diff
@@ -506,12 +506,12 @@ int System::System::integrate(int n_steps, int reuse_forces) {
...
     ek_active = ek.is_ready_for_propagation();
 #ifdef SHARED_MEMORY_PARALLELISM
-    cell_structure->set_max_prefactor(5);
+    cell_structure->set_max_prefactor(8);
 #endif
   }
 #ifdef SHARED_MEMORY_PARALLELISM
   else {
-    cell_structure->set_max_prefactor(8);
+    cell_structure->set_max_prefactor(12);
```
hidekb: Since `m_max_prefactor` has a default value, and `max_counts` is set to the number of particles in CellStructure.hpp when `integ_switch` is `INTEG_METHOD_STEEPEST_DESCENT`, this part of the code might be redundant. Could you clarify if it serves a specific purpose?

jngrad: Indeed, it's now redundant. I've removed this code.

hidekb (approved):
looks good to me!!

PR #5147, opened 2025-08-21
Improve performance of ParticleSlice property access

RudolfWeeber:
Description of changes:
* address some of the performance regressions introduced in `ParticleSlice` since 4.2.2

### DIFF
--- a/src/python/espressomd/particle_data.py
+++ b/src/python/espressomd/particle_data.py
@@ -19,6 +19,7 @@
 
 import numpy as np
 import collections
+import itertools
 import functools
 from .interactions import BondedInteraction
 from .utils import nesting_level, array_locked, is_valid_type
@@ -26,7 +27,6 @@ from .utils import check_type_or_throw_except
 from .code_features import assert_features, has_features
 from .script_interface import script_interface_register, ScriptInterfaceHelper
 from .propagation import Propagation
-import itertools
 
 
 @script_interface_register
@@ -886,9 +886,13 @@ class ParticleSlice(ScriptInterfaceHelper):
     _so_name = "Particles::ParticleSlice"
     _so_checkpointable = False
     _so_creation_policy = "GLOBAL"
+    _particle_cache_size = 10000  # size of the particle cache for slices
 
     def __iter__(self):
         return self._id_gen()
+
+    def _get_particle_impl(self, p_id):
+        return self.call_method("get_particle", p_id=p_id)
 
     def _id_gen(self):
         """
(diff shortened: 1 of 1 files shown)


src/python/espressomd/particle_data.py
```diff
@@ -1432,8 +1441,7 @@ def get_attribute(particle_slice, attribute):
 
         # get first slice member to determine its type
         p_id = particle_slice.id_selection[0]
-        target = getattr(
-            particle_slice.call_method("get_particle", p_id=p_id), attribute)
+        target = getattr(particle_slice._get_particle(p_id), attribute)
```
jngrad, line 1444: Here and elsewhere, I re-introduced the original `getattr()` and `setattr()` implementation, because not all particle attributes have the same name in the script interface (e.g. `exclusions`), and some of them require pre- or post-processing, such as `propagation` which is serialized to an int and deserialized to a `Propagation` class.

jngrad:
Benchmark script:
```python
import time
import argparse
import espressomd
import numpy as np
parser = argparse.ArgumentParser(description="Benchmark particle list")
parser.add_argument("--particles_per_core", metavar="N", action="store",
                    type=int, default=1000, required=False)
args = parser.parse_args()
np.random.seed(42)
system = espressomd.System(box_l=[10., 10., 10.])
n_proc = system.cell_system.get_state()["n_nodes"]
n_part = n_proc * args.particles_per_core
system.cell_system.set_regular_decomposition(use_verlet_lists=True)
system.time_step = 0.01
system.cell_system.skin = 0.5
system.part.add(pos=np.random.random((n_part, 3)) * system.box_l)
system.integrator.set_vv()
n_loops = 60
timings = []
plist = system.part.all()
for _ in range(30):
    t1 = time.time()
    for i in range(n_loops):
        np.array(plist.pos)
        np.array(plist.q)
        np.array(plist.f)
        np.array(plist.v)
    t = time.time() - t1
    timings.append(t)
    print(f"{t*1000 / n_loops:.1f} ms/loop")
print(f"avg: {np.mean(t)*1000 / n_loops:.1f} ms/loop")
```
Results with 1k particles per core:
* python branch: 35.4 ms/loop
* this PR: 8.6 ms/loop
* v4.2.2: 2.7 ms/loop

Results with 20k particles per core (twice the cache size):
* python branch: 700 ms/loop
* this PR: 860 ms/loop (slower due to cache invalidation)
* v4.2.2: 2.7 ms/loop (unaffected!)

While we could in principle halve the runtime by calling `get_parameter` on attributes that are not part of a shortlist (`["exclusions", "propagation"]`, maybe more), this is quite risky, because it's easy to forget to update that list when adding a new non-trivial particle attribute as part of a feature request, and it might take weeks before someone notices.

@NiklasMueller2001 is working on a generalization of #5124 to particle slices, where we can leverage the struct of arrays data structure to avoid resolving particles, if the user only queries the most popular attributes (id, type, force, velocity, position, charge).

jngrad (approved):

PR #5150, opened 2025-08-25
Implement LB boundary forces

itischler:
Description of changes:
- Added a getter for the lb forces acting on the boundaries


maintainer/walberla_kernels/generate_lb_kernels.py
```diff
@@ -318,6 +318,21 @@ def patch_boundary_kernel(content, target_suffix):
...
         return content
 
+    def patch_boundary_force_accessor(content):
+        # make ids accessable from outside
+        content = content.replace("BlockDataID forceVectorID;", "")
+        content = content.replace(
+            "BlockDataID pdfsID;", "BlockDataID pdfsID;\nBlockDataID forceVectorID;\n")
+        content = content.replace("BlockDataID indexVectorID;", "")
+        content = content.replace(
+            "BlockDataID pdfsID;", "BlockDataID pdfsID;\nBlockDataID indexVectorID;")
```
jngrad: Can this be replaced by overloading static method `additional_member_variable()` in lbmpy_espresso with a function that injects a multi-line string containing `public:` followed by several getters? See e.g. the output of `print(ubb_data_handler.additional_member_variable)`, to be added after line 302.

maintainer/walberla_kernels/generate_lb_kernels.py
```diff
@@ -318,6 +318,21 @@ def patch_boundary_kernel(content, target_suffix):
...
+        # use range-based for loop
+        content = content.replace(
+            "for(std::vector<ForceStruct>::iterator it = cpuVector_.begin(); it != cpuVector_.end(); ++it)",
+            "for(auto & it : cpuVector_)")
+        content = content.replace("] += it->F_", "] += it.F_")
```
jngrad: patch submitted upstream: https://i10git.cs.fau.de/walberla/walberla/-/merge_requests/781

jngrad: See `jngrad/lb_boundary_force` for the merge conflicts resolution. The extra commit simplifies UBB code generation. It is not complete, though. For it to work, one has to use the `maintainer/walberla_kernels/templates/Boundary.tmpl.h` template instead of the one currently in waLBerla. The Ekin codegen script does exactly that, but I don't remember how.

I replaced the `constexpr std::vector` getter function by a `constexpr std:array`, because constexpr suport in std::vector is partial as of C++20. Please bear in mind the `constexpr` specifier doesn't make the function constexpr, but marks it as available in a constexpr context. Since you didn't mark variable `neighbor_offset` as constexpr, the compiler used the non-constexpr version of the function and didn't warn you about vector resize operations not being constexpr :-)

Live demo: https://godbolt.org/z/xqfM7K1oz

itischler:
The only thing missing is that the lbmpy version needs to be updated when the [lbmpy-pr](https://i10git.cs.fau.de/pycodegen/lbmpy/-/merge_requests/193) for the UBB boundary force is merged

src/walberla_bridge/src/lattice_boltzmann/LBWalberlaImpl.hpp
```diff
@@ -1753,14 +1753,14 @@ class LBWalberlaImpl : public LBWalberlaBase {
...
-    for (auto const &block : *get_lattice().get_blocks()) {
+    for (auto &block : *get_lattice().get_blocks()) {
       auto const offset = get_lattice().get_block_corner(block, true);
       auto const &force_field = m_boundary->get_force_vector(&block);
       auto const &index_field = m_boundary->get_index_vector(&block);
       for (int i = 0; i < raster_flat.size(); i++) {
         if (raster_flat[i] != 0) {
           auto node = flat_index_to_node(i);
-          if (get_lattice().node_in_local_domain(node)) {
+          if (get_lattice().node_in_local_halo(node)) {
```
jngrad, line 1763: I had to change from local domain to local halo, otherwise contributions would be lost when a boundary stretches across a periodic boundary. To reproduce the error with the original code, run `lb_buoyancy_force.py` with sphere parameters `RADIUS = 0.8 * AGRID` and `center=self.system.box_l / 2 + 0.5 * AGRID`. This will create a single boundary cell on the periodic boundary (`print(self.lbf[8:11,8:11,8:11].is_boundary.astype(int))`). Integrate for a few time steps and print the force on the boundary, using 1 or 2 MPI ranks to see the difference in the per-shape force.

src/walberla_bridge/src/BoundaryHandling.hpp
```diff
@@ -231,16 +231,18 @@ class BoundaryHandling {
...
   }
 
-  BlockDataID get_force_vector_id() const { return m_boundary->forceVectorID; }
-  BlockDataID get_index_vector_id() const { return m_boundary->indexVectorID; }
-  auto const &get_force_vector(const IBlock *block) {
+  auto const &get_force_vector(IBlock *block) const {
+    using ForceVector = BoundaryClass::ForceVector;
+    auto const force_vector_id = m_boundary->getForceVectorID();
+    auto *forceVector = block->getData<ForceVector>(force_vector_id);
+    forceVector->syncCPU();
```
jngrad, line 242: The call to `syncCPU()` was missing. The total LB force was correct, but the force per shape was always zero with the GPU kernels.

src/walberla_bridge/src/lattice_boltzmann/generated_kernels/DynamicUBBSinglePrecisionCUDA.h
```diff
@@ -105,8 +117,12 @@ class DynamicUBBSinglePrecisionCUDA {
...
     IndexInfo *pointerGpu(Type t) { return gpuVectors_[t]; }
     void syncGPU() {
-      for (auto &gpuVec : gpuVectors_)
-        WALBERLA_GPU_CHECK(gpuFree(gpuVec));
+      for (auto &gpuVec : gpuVectors_) {
+        if (gpuVec) {
+          WALBERLA_GPU_CHECK(gpuFree(gpuVec));
+          gpuVec = nullptr;
+        }
+      }
```
jngrad, line 125: Nasty double-free bug. Patch submitted upstream: https://i10git.cs.fau.de/walberla/walberla/-/merge_requests/785/diffs?commit_id=b5a78afcc8d3affc6312c1a9272f44de1cf0c0a2

jngrad (approved):
Thanks a lot!

PR #5151, opened 2025-08-25
Py: Script interface performance

RudolfWeeber:


jngrad (approved):
This leads to a 5% speed-up for particle data access.

PR #5152, opened 2025-08-25
Improve management of compile-time constants

jngrad:
Description of changes:
- make compile-time constants strongly typed and move them to more relevant header files
- add prefix to all ESPResSo macros to avoid name clashes with third-party libraries


jngrad:
Required for https://github.com/espressomd/espresso/pull/4970#issuecomment-2278377638

hidekb (approved):
I checked all macro by "$grep -rFh ifdef | sort -u". Then, looks good to me!

PR #5157, opened 2025-08-26
On a single MPI rank, don't use ghost cells in RegularDecomposition

RudolfWeeber:
When running on a single MPI rank, ghost cells aren't necessary, as the 
corresponding physical cell is available on the same MPI rank.

In this case, the physical cells can be directly connected as neighbors and the ghost communication deactivated.

the PR somewhat improves the performance of shared-memory parallelizaiton simulations on a single MPI rank by avoiding the non-parallel ghost communication.

@jngrad do we have a CI job that runs the test on just one mpi rank? Aside from this PR, there are other optimizations in ESPREsSo which are only active on sinlge mpi rank simulations.


jngrad:
One could add:
```cmake
python_test(FILE regular_decomposition.py MAX_NUM_PROC 1 SUFFIX 1_core)
```
after line 287 in `testsuite/python/CMakeLists.txt` to run the test with 1 core.

RudolfWeeber:
It's not obvious, which tests exacly one should run this way, as the cell system details interact with a lot of things from Lees Edwards to LB particle coupling.
I added a gitlab job to run the entire test suite with one mpi rank in the additional_checks stage.

jngrad:
That's going to put a bit more strain on the CI fleet. Additional jobs can only run on two runners, and one of them was down yesterday, stalling all of the CI pipelines. If we go forward with this, we should consider investing in a new CI runner if there is still budget left for infrastructure. We are already pushing the existing runners to the limit by oversubscribing cores by a factor 2.5 and taking advantage that hyperthreads are idle when only 1 CI pipeline is running, and that some python tests don't consume CPU resources.

RudolfWeeber:
We'll have to discuss about the runners at the ICP. In any case, I think, not testing single core is the worse option, as this is the most common way in which Espersso is used, and wil become more so, whit shared memory parallelism.

jngrad (approved):
With 2500 particles per thread and 1 MPI rank, the new code is 11% faster with 1 thread and 20% faster with 4 threads according to the LJ benchmark.

PR #5159, opened 2025-08-28
CI/CD maintenance

jngrad:
Description of changes:
- use the new CI runners, which allocate 8 CPU cores per job (25% improvement in runtimes)
- move Fedora job to GitHub Actions and combine the style and doxygen jobs to reduce the number of CPU-only jobs
- factor in number of threads in test properties to reduce oversubcription when CTest schedules tests in parallel
- introduce extra interface checks to improve code coverage
- refactor C++ core and Ekin Poisson solvers


testsuite/python/ek_interface.py
```diff
@@ -472,13 +539,18 @@ class EKTestWalberlaSinglePrecision(EKTest, ut.TestCase):
...
+@ut.skipIf(EKTest.system.cell_system.get_state()["n_nodes"] != 1,
+           "GPU EK runs for 1 MPI rank")
 class EKTestWalberlaGPU(EKTest, ut.TestCase):
 
     """Test for the Walberla implementation of the EK in double-precision."""
 
     lb_fluid_class = espressomd.lb.LBFluidWalberlaGPU
     ek_lattice_class = espressomd.electrokinetics.LatticeWalberla
     ek_species_class = espressomd.electrokinetics.EKSpeciesGPU
+    ek_solver_class = espressomd.electrokinetics.EKFFTGPU
```
jngrad, line 553: @itischler A few issues surfaced when attempting to write extra interface checks for the EKFFTGPU code in `ek_interface.py` and `lattice_vtk.py`. Once this PR gets merged, could you please have a look at it? You will need a Debug build to investigate the error message. I had to run GDB in two terminals with:
```
$ mpiexec -np 2 xterm -fa 'Monospace' -fs 12 -e ./pypresso --gdb \
    ../testsuite/python/ek_interface.py EKTestWalberlaGPU
(gdb) break MPI_Abort
(gdb) run
```

itischler: sure I can do that. it is probably because the i did not check for the gpu fft to work with multiple mpi ranks, however i assume this is the same error as i had for the cpu heffte yesterday, which i fixed in #5161 . In the current GPU version the heffte box has the same size as the whole system, which causes the boxes to overlap when running this on multiple mpi ranks.

src/core/communication.cpp
```diff
@@ -123,12 +123,13 @@ void init(std::shared_ptr<boost::mpi::environment> mpi_env) {
 }
 
 void deinit() {
-  ErrorHandling::deinit_error_handling();
-  Communication::m_callbacks.reset();
-
```
hidekb, line 128: For just curiosity, are there any problems with the previous order: Communication::m_callbacks.reset(); kokkos_handle.reset();?

jngrad: Not really, but it looked odd to me that Kokkos was the only managed resource on the stack that didn't follow the last in, first out (LIFO) approach to resource management.

hidekb (approved):
It looks good to me.

PR #5162, opened 2025-08-29
Add interface for ASE calculators

RudolfWeeber:
This extends the ASE plugin for ESPREsSo. Among others, it provideas an integrate() method, which handles the details of transferring data between Espresso and ASE.

At this point, there is a significant problem remaining:
While I was not able to write down a (simple) failing test, a Lennard-Jones liquid simulatinos based on this (sampels/lj_liquid_ase.py) is nto stable (heats up).

One possible cause is that, due to the two-step integrator, cahging particle external forces in every time step, destroys the statistics of the Langevin noise. (The update of particle properties done by ase triggers an invalidation of existing particle forces, which might change the first lj half step)


RudolfWeeber:
Update: A draft for the single step symplectic euler integration is  in #5163 
This improves on the problem but does not fully fix it.
On my Github (RudolfWeeber) in the branch
ase_sym_euler
is the emrge of this PR and the symplectic Euler work
With sympplectic euler integrator:
2 particlse connected by a harmonic spring from ASE (samples/ase_2part.py) 
* in NVE ensemble conserve the energy (up to 1E-4 rel eoror)
* in NVT ensemble, kinetic and potential energy are maintained at 1/2kT per quadratic term: 3kT for Ekin of 2 partiles, 1.5kT for harmonic potential in 3D.

The ASE-based Lennard-Jones (samples/lj_liquid_ase.py) still heats up until it explodes, though, even with symplectic euler integrator.

RudolfWeeber:
Correction: With symplectic Euler integration, the ASE-based Lj simulation is also stable (with velocity verlet, it is not).

In summary, when changing partile external forces from Python in every time step, at this point, we need a single step integrator.

jngrad (approved):

PR #5163, opened 2025-09-01
Support for single step symplectic euler integrator

RudolfWeeber:
This adds preliminary support for a sinlge step integraotr using
system.integrator.set_symplectic_euler()

Main missing thing is more extensive testing in cooperation with Langevin thermostat and blocks for combinations that we don't test, e.g., LB coupling.


jngrad:
Offline discussion: the symplectic Euler integrator needs to be revisited after the summer school, and possibly removed if we can find a better way to integrate particles with time-dependent external forces. It is an experimental feature.

jngrad (approved):

PR #5167, opened 2025-09-08
remove dangerous ABI-breaking warning suppression

eli-schwartz:
The Strict Aliasing rule is actually important, and not a mere style recommendation. Breaking it leads to Undefined Behavior and subsequently to mysterious crashes.

Building with -Wno-strict-aliasing does not get rid of the crashes but does tell the compiler to refrain from telling you about them. This flag does not, in the end, do anything when there are no warnings to ignore! But it does prevent catching future mistakes. The flag appears to have been added via cargo cult.

If not possible to fix a genuine aliasing issue correctly, a better option in 100% of cases is anyways to use -fno-strict-aliasing. This isn't a warning suppression flag -- it is a "disable the compiler optimizations" flag that reduces the risk of aliasing bugs resulting in runtime errors (because the compiler tries to avoid generating certain types of efficient code that knowingly rely on aliasing being correct).

Fixes #

Description of changes:
-

### DIFF
--- a/src/python/espressomd/CMakeLists.txt
+++ b/src/python/espressomd/CMakeLists.txt
@@ -59,7 +59,6 @@ target_compile_options(
     $<$<NOT:$<CXX_COMPILER_ID:IntelLLVM>>:-Wno-deprecated-declarations>
     $<$<CXX_COMPILER_ID:IntelLLVM>:-diag-disable=1224>
     $<$<CXX_COMPILER_ID:GNU>:-Wno-cpp>
-    $<$<CXX_COMPILER_ID:GNU>:-Wno-strict-aliasing>
     $<$<CXX_COMPILER_ID:GNU>:-Wno-maybe-uninitialized>
     $<$<CXX_COMPILER_ID:GNU>:-Wno-volatile>
     $<$<CXX_COMPILER_ID:GNU>:-Wno-shadow=compatible-local>


jngrad:
Thank you for raising our attention on this matter and contributing a pull request!

I don't think the intention of this flag was to avoid crashes, but rather to silence false positives from `-Wall` and let us use `-Werror` to force developers to address all warnings. We only use `-Wno-strict-aliasing` in automatically-generated source code, such as Cython code and in waLBerla kernels via `#pragma GCC diagnostic ignored "-Wstrict-aliasing"`. Cython in particular forces us to periodically re-evaluate which diagnostics flags we have to disable (the most recent one being e18cc41e). The offending flag was introduced in 2017 and was most likely necessary back then. A lot has changed since: we now use Cython as a thin wrapper and no longer expose ESPResSo core features, we no longer support older GCC and Clang versions, and we now require recent Cython versions. I'm not sure why this flag survived our last cleanup. CI confirms that there are no false positives for GCC 12, 13, and 14.

jngrad (approved):

eli-schwartz:
Ha, my theory was that it was a remnant of python 2, see https://wiki.gentoo.org/wiki/Project:Python/Strict_aliasing (and note that that legacy page suggests fixing it by passing -fno-* specifically because it did lead to crashes).

But I admit I didn't spelunk through history to see when it was introduced.

I did check with my default GCC 14 compiler to see that no warnings were produced with the flag removed (Gentoo quality assurance tooling greps the compilation logs for any -Wstrict-aliasing issues and raises an alert, which is why I noticed this -- I have another alert for packages that pass -Wno-strict-aliasing rather than the `-f` version and then I try to figure out why it's passing it).

jngrad:
Thanks for sharing that page! I suspected it had something to do with Python 2, which matches the timeline perfectly, but I didn't want to put the blame on it without material evidence to support my claim. We did support both Python 2 and 3 simultaneously for a long time after Python 2 was officially decommissioned, and therefore had to keep it for a while.

PR #5169, opened 2025-09-11
Draft:Thermalize Lees Edwards LBM kernel

Yashas17:
Fixes #

Description of changes:
- Add thermal fluctations to Lees Edwards LB kernel


src/walberla_bridge/src/lattice_boltzmann/LBWalberlaImpl.hpp
Yashas17, line 1: Since I am trying to thermalize Lees Edwards, I need to pass time-step to the kernel. The time-step getter and setter functions at the moment are called only if the collision model is the basic thermalized model. Since now even Lees Edwards will be thermalized and no unthermalized models remain, should I remove this check completely or add Lees Edwards collision model as an option as well. You can refer to line 535, 1777 and 1785 to check the current implementation?

Do you have any opinion on this @jngrad @itischler @RudolfWeeber?

jngrad: If all collision models are thermalized, you no longer need to check the collision model type (and we can probably drop the `Thermalized` suffix from the regular kernels), but you still need to check for m_kT being zero.

PR #5170, opened 2025-09-12
Batch particle slice properties setter

NiklasMueller2001:
Until now setting properties on the `ParticleSlice `class made multiple expensive calls **per particle in the slice** into the script interface to set each particle's value iteratively. Now a batch update is performed, similar to [#5124](https://github.com/espressomd/espresso/pull/5124).


jngrad (approved):
Particle properties setter benchmark: major speed-up compared to the python branch

* 1 MPI rank: from 7.91 ms/loop down to 1.67 ms/loop
* 2 MPI ranks: from 31.5 ms/loop down to 5.2 ms/loop
* 4 MPI ranks: from 81.1 ms/loop down to 11.6 ms/loop

Using 1000 particles per MPI rank on an AMD Ryzen 9 9950X.

Notifying @hidekb.

Particle properties getter benchmark: no difference. Here @NiklasMueller2001 could apply the same recipe to the parallel getter in a follow-up PR. At the moment, we manually iterate over each particle to extract its attribute value, which requires the constant creation of a new `ParticleHandle`. This PR introduced a `ParticleModifier` thin wrapper that is created once by the particle slice and gets its internal particle id updated between each attribute read to save a massive amount of time.

PR #5173, opened 2025-09-16
Simplify build system

### LINKED ISSUE #5166: Impossible to run ctest directly because test dependencies cannot be requested to be built.
cmake tests are traditionally built by `all` and can't be excluded, since unlike build systems such as automake or Meson... ctest doesn't know how to build tests before running them. There is no target to manually build tests either. And since https://github.com/espressomd/espresso/issues/3861 it is not possible to control the arguments passed to ctest after already configuring, so even assuming one very much wishes to wrap ctest in layers of cmake configure argument indirection the experience will be suboptimal.

A good model here would be something akin to https://mesonbuild.com/Release-notes-for-1-7-0.html#test-targets-no-longer-built-by-default

- create an alias (.PHONY) target "test-prereq" that builds all test programs
- have the manually defined test target depend on that in turn
- ideally also provide a way to exclude nondefault tests entirely. cmake provides "labels", which I believe would function the same as Meson's [`test(suite: 'foobar')`](https://mesonbuild.com/Reference-manual_functions.html#test_suite), although I'm not aware of a good way to duplicate the effect of [`add_test_setup(exclude_suites: ['foobar'], is_default: true)`](https://mesonbuild.com/Reference-manual_functions.html#add_test_setup_exclude_suites). Still it should at least be possible to manually restrict tests to a specific group.

jngrad:
Description of changes:
- Combine C++ and CUDA source files in the same target
   - leverage modern CMake native support for CUDA and HIP source code
   - avoid the creation of separate targets `espressso::core` and `espresso::core_cuda`
   - reduce the complexity of the linking stage:
      - tightly-coupled object files (cyclic dependency) are now part of a single shared object
      - dynamic loading no longer depends on dubious runpaths transitivity rules
- Overhaul CMake logic for the testsuite
   - decouple building and testing stage in C++ and python tests to help with packaging on EESSI and Gentoo
   - packagers can now run `make espresso_packaging_dependencies` to build ESPResSo source code, unit test executables, and Python test dependencies, while `ctest -j$(nproc) --output-on-failure -L "unit_test|python_test" -LE long` can be used to run all tests relevant for packaging (fixes #5166)


jngrad:
This work integrates major CMake improvements made in waLBerla ([walberla/walberla!774](https://i10git.cs.fau.de/walberla/walberla/-/merge_requests/774), [walberla/walberla!781](https://i10git.cs.fau.de/walberla/walberla/-/merge_requests/781)) and completes the CUDA target refactoring initiated in f3efdfae4542815fa8fe57740f66f1acab38732d.

I can confirm this PR works on the following clusters: Ant, BinAC2, Helix, bwUniCluster3.0.

jngrad:
Full disclosure: on bwUniCluster3.0 there is a harmless UCX warning about a stray `MPI_Send` or `MPI_Recv`. Since the GDB command `rbreak ^MPI_` doesn't catch anything after `MPI_Finalize` in a debug build of ESPResSo, there is a good chance the issue comes from how OpenMPI is packaged on that cluster, rather than from ESPResSo. It's a CUDA-aware UCX-ready OpenMPI build, and the warning is only emitted when a CUDA GPU is available. I won't investigate this further, since GDB cannot help diagnose the warning.

cmake/FindCUDACompilerNVCC.cmake
```diff
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2009-2022 The ESPResSo project
+# Copyright (C) 2009-2025 The ESPResSo project
```
hidekb, line 2: I agree — the years here were updated from 2022 to 2025. Just curious, though: why does the year in `cmake/espresso_unit_test.cmake` remain 2024?

jngrad: I like to update the date for major changes to the file. This way, when people re-use ESPResSo files in other projects, they have the most up-to-date license header. I also must change the date range when I copy-paste code from a file to another one, when the date range from the source file is more recent than the destination file (copyright year ranges cannot be decreased). For small changes, I typically don't bother, because there is little creative value, and I don't want to make the git diff more difficult to read; at every software release, we run a script to update all year ranges anyway, in a single commit that doesn't annoy anyone.

hidekb (approved):
LGTM. Thanks for the great work.

PR #5174, opened 2025-09-23
MLIP and Water Tutorial

PythonFZ:
This PR adds the MLIP/Water Tutorial.
It is spilt into three parts
1. TIP4P Water by @hidekb
2. MLIP Setup and evaluation by @PythonFZ 
3. MLIP MD by @hidekb

For testing until merged, you can use

```bash
git clone https://github.com/PythonFZ/espresso.git
cd espresso
git checkout mlip-water
cd doc/tutorials/mlip-water/
# use your favorite way of installation, e.g. uv sync / pip install . or pre-installed
dvc pull
```


RudolfWeeber:
Thank you! It looks very good.

A few points from an initial quick read:

## Part 1

* considr using GPU P3M, if a GPu is available. 
* maybe extend description of how the rigid body with Virtual Sites works

## Part 2

* For newcomers, some of the terminology might not be fully clear, e.g., hyperparameter, descriptor...
* Maybe it can be clarified what "true energy" erfers to, i.e., make more explicit that this refers to the result from qm calcuations on which the model was trained
* Eventhough the models are pre-trained, can one show some samples from the training data visually?
* What is the role of zntrack? 


## Part 3

* Maybem we can make it clear that this script in essence works for other MLIPs, as long as they provide an ASE calculator
* Explain thaat ase.integrate() assignes to the particles' ext_force. This means that
    * this attribute is overwritten
    * but otherwise, all ESPResSo interactions can be used in combination with the ASe calculator

doc/tutorials/mlip-water/README.md
```diff
@@ -0,0 +1,24 @@
...
+## Setup
+
+1. Install [UV](https://astral.sh/uv/) if you haven't already `curl -LsSf https://astral.sh/uv/install.sh | sh`
+2. Create the virtual environment and download the data and model files
+
+```bash
+uv sync --frozen --no-cache # --no-cache is optional to avoid caching into your home directory
+source .venv/bin/activate # activate venv
+dvc pull # download the model and data files
+```
```
PythonFZ: add `pip install .` instructions

doc/tutorials/mlip-water/01_TIP4P_water.ipynb
```diff
@@ -0,0 +1,583 @@
+{
```
schlaicha, line 1: <p>Would it not be more readable/less error prone to get the constants from <code>scipy.constants</code>?</p><p>And if you use pint and units everywhere, also the lengths should get there units IMHO (and for the charges and masses below). Alternatively (maybe preferred), add the explanation that these constants are defined in order to convert to the unit-less quantities below - as it is now it is not clear intuitively which quantities have or don't have units.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/espressomd/espresso/pull/5174/discussion/#comment-2380034236'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='3'/>

schlaicha, line 1: <pre class="ql-syntax" spellcheck="false">Line #4.    def rigid_constrain(p, C):
</pre><p>Why C? This relates two particle positions, right? Maybe center would be more instructive?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/espressomd/espresso/pull/5174/discussion/#comment-2380034244'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='16'/>

schlaicha, line 1: <p><em>Typiically</em> the hydrogens have no LJ interaction, but for example in the CHARMM compatible versions they do. To be fair I would simply add the <em>typical </em>in the sentence.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/espressomd/espresso/pull/5174/discussion/#comment-2380034249'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='19'/>

schlaicha, line 1: <pre class="ql-syntax" spellcheck="false">Line #2.        epsilon=OO_EPSILON, sigma=OO_SIGMA, cutoff=0.85, shift="auto"
</pre><p> </p><p>This is the original cutoff, and water properties at this small cutoff are very sensitive</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/espressomd/espresso/pull/5174/discussion/#comment-2380034251'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='20'/>

schlaicha, line 1: <pre class="ql-syntax" spellcheck="false">Line #3.    system.thermostat.set_langevin(kT=1, gamma=1, seed=1)
</pre><p> </p><p>This will not give the correct dynamics. I see that we have no other choices for the thermostat, but at least we should tune the friction coefficient to match the dynamics of TIP4P/2005, $k_\mathrm{B}T/\gamma = D = 2.08 \cdot 10^9 \mathrm{m^2/s)$?</p><p> </p><p>Would this yield the desired behavior?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/espressomd/espresso/pull/5174/discussion/#comment-2380034257'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='25'/>

schlaicha, line 1: <pre class="ql-syntax" spellcheck="false">Line #1.    # replace singel O particle by molecules
</pre><p>comment makes no sense to me</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/espressomd/espresso/pull/5174/discussion/#comment-2380034259'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='26'/>

schlaicha, line 1: <pre class="ql-syntax" spellcheck="false">Line #5.    r_max = 0.7  # nm
</pre><p>Why not L/2? Then this looks much more familar...</p><p>Correspondingly</p>

<pre class="ql-syntax" spellcheck="false">rdf_bins = int(L/2 * 100)
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/espressomd/espresso/pull/5174/discussion/#comment-2380034262'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='32'/>

schlaicha, line 1: <pre class="ql-syntax" spellcheck="false">Line #5.    data = np.load("rdf_experiment.npz")
</pre><p> </p><p>File is missing, so I cannot judge on the quality of the rdf...</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/espressomd/espresso/pull/5174/discussion/#comment-2380034269'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='34'/>

schlaicha:
I now had a look at the first part, this is looking very impressive to me (see my comments above)!

I just did a slightly longer (100x) run to test the pressure in the simulation, at least with the current friction coefficient (see my comment above) it seems reasonable: $$313 \pm 45$$ bar, which seems very plausible to me...

I will look into parts 2 and 3 tomorrow.

doc/tutorials/mlip-water/02_MLIPs.ipynb
```diff
@@ -0,0 +1,742 @@
+{
```
schlaicha, line 1: <p>Maybe one should add a hint on the unit, rmax is in Angstrom?</p><p> </p><p>Also, concerning the <strong>n_radial</strong>, I would add the explanation from below: "(...) in message passing, which affects angular resolution"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/espressomd/espresso/pull/5174/discussion/#comment-2381881832'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='4'/>

schlaicha, line 1: <p>typo, "we used on to set" - "on" needs to be removed</p><p> </p><p>The note on units could be moved up to where the cutoff is introduced</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/espressomd/espresso/pull/5174/discussion/#comment-2381881843'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='7'/>

schlaicha, line 1: <pre class="ql-syntax" spellcheck="false">Line #12.    ax[1].set_ylabel("Density")
</pre><p>Should also be labeled 'counts' in order not to confuse.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/espressomd/espresso/pull/5174/discussion/#comment-2381881857'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='12'/>

schlaicha, line 1: <pre class="ql-syntax" spellcheck="false">Line #7.        packmol="/data/fzills/tools/packmol-20.15.1/packmol",  # Path to PACKMOL executable
</pre><p>This needs to be adressed before merge, especially wrt. CI</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/espressomd/espresso/pull/5174/discussion/#comment-2381881869'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='22'/>

doc/tutorials/mlip-water/01_TIP4P_water.ipynb
hidekb: <p>I modified this cell for using Scipy.constants and added "these constants are defined in order to convert to the unit-less quantities below" below the cell.</p>

hidekb: <p>I modified the argument <code>C</code> to <code>center_mol</code>  becuase <code>center</code>  is already defined. </p>

hidekb: <p>I added typical in front of TIP4P.</p>

hidekb: <p>I modified <code>cutoff</code> from <code>2.5 * OO_SIGMA</code> to 0.85.</p>

hidekb: <p>In this tutorial, we only discuss the radial distribution function. Then, for fast equilibrium, we set <code>gamma=1</code>.</p>

hidekb: <p>I removed it.</p>

hidekb: <p>After using $dvc pull, the necessary data is provided.</p>

hidekb: <p>In this tutorial, all RDFs are compared with each other. For the third part, the maximum length for RDF is 7 angstroms since the system size is 14 angstroms. To match that length, we set it to 7 angstroms.</p>

PythonFZ:
In the ASE interface, we are setting the calculator in a loop.

```python
for _ in tbar:
    atoms = system.ase.get()
    atoms.calc = box.calc
```
The `.calc` is a property (https://gitlab.com/ase/ase/-/blob/master/ase/atoms.py#L396) and there might be a performance consideration here.

doc/tutorials/mlip-water/02_MLIPs.ipynb
PythonFZ: moved to the very top

PythonFZ: I am using density for the forces, because there are so many force component but keep the non-normalized histogram for the energies, because the numbers still make sense to the reader.

PythonFZ: I moved the note on units to the top and replaced the description ( it was wrong before, thanks for highlighting!)

PythonFZ:
@PicoCentauri we are using the GMNN[^1] as implemented in APAX[^2] 

[^1]: https://pubs.acs.org/doi/10.1021/acs.jctc.0c00347 
[^2]: https://pubs.acs.org/doi/full/10.1021/acs.jcim.5c01221

schlaicha:
Hi,

I had another look and all my points seem well adressed. I just pushed quite a few of typo corrections and didactic rephrasings. From my side this is finished and can be used as tutorial!

A few open points:

- Concerning the CUDA support for P3M in the first parg suggested by @RudolfWeeber , the way it is implemented now always demands CUDA, so the code will fail without CUDA installed - I think this should be a switch checking *if* CUDA is available and then using it.

- I don't understand the last paragraph in the third tutorial part,

   > Because our MLIP learns all interactions, it can happen that minimum bond distances are not obeyed, (...)

   I guess you are refering to the data $$r_\text{max} = 2\text{\AA}?$$
   Then I would simply conclude that such a small cutoff misses the fundamental interactions and therefore produces nonphysical results, leaving out the rest of the discussion.

   > this can be mitigated by including short-range two-body

   why would you want to do that?

PythonFZ:
I can shorten this, but I think it includes some valuable information

> Because our MLIP learns all interactions, it can happen that minimum bond distances are not obeyed, leading to unphysical behavior in the simulation that are reflected in the RDF plots.

I think this highlights again, that there are e.g. no springs between particles and if not enough information is given, the system can blow up. Maybe including another sentence to explicitly refer to the 2 A data might help?

> Although this can be mitigated by including short-range two-body potential terms, fundamentally, if such behavior occurs, the model parameters or training data should be improved.

Including the `Ziegler-Biersack-Littmark (ZBL)` short-range correction potential is common practice in many MLIPs but for our model, we have seen that including these correction terms might not necessarily improve the model accuracy.

If this is information is to coarse and doesn't really benefit the model, I'll shorten the section as you've suggested?

schlaicha:
> If this is information is to coarse and doesn't really benefit the model, I'll shorten the section as you've suggested?

Yes, I think either you need to provide more explanations (like the one you provided above) or simplify the section...

jngrad:
@hidekb Some of the ASE changes in this PR have been merged in the python branch. Could you please resolve the outstanding merge conflicts?

@PythonFZ I guess the time has come for me to finally look into CI workflows with large data artifacts :smiley:  It will probably be easier if we meet in person to look into this. Here are already a few questions to refresh my memory. There are a lot of files in the tutorial folder, e.g. `data.dvc`, `dvc.lock`, `config/*.yaml`, `.python-version`, `uv.lock`, `zntrack.json`; can any of those be regenerated automatically? There is also an S3 bucket path in `.dvc/config`; do we need to archive it somewhere safe?

doc/tutorials/mlip-water/01_TIP4P_water.ipynb
jngrad: The box length was eventually resized to 25 Angstroms. The RDF is now calculated up to 8 Angstroms. While larger values are possible, we save the RDF to a file in this tutorial to plot it against the MLIP RDF in Part 3 for comparison, which is only accurate up to 7 Angstroms when r_max is 5.5 Angstroms.

jngrad (approved):
Thanks a lot everyone!

PR #5179, opened 2025-10-13
Remove particle pointer-based access in forces_cabana and force_inline

### LINKED ISSUE #5155: Remove remaining acces to Particle in forces_cabana
* Add the following properties to aosoa pack:
    * velocity
    * scalar dipole moment
    * quaternion
    * number of exclusions

* make Particle::calc_dip() a free funciotn calc_dip(quat, scalar_dipm) and use in dipolar short range kernel
* analogously for Gay-Bern (Particle::calc_director())
* probably, dipolar layer correction has to be adapted the same way as ELC

With this, to my understanding, the only instance in which the full particle is needed is, if num_exclusions for that particle >0

Further points:
* NPT code should only be used if npt is active (see function in central force PR)

With the additional properties in tha AosoaPack, it should also be possible to rewrite the VerletCriterion to use the particle properties directly and avoid access to the particle pointers.

hidekb:
Fixes #5155

Description of changes:
- Added `VelocityViewType = Kokkos::View<double *[3], Kokkos::LayoutRight, Kokkos::HostSpace>`, `DirectorViewType = Kokkos::View<double *[3], Kokkos::LayoutRight, Kokkos::HostSpace>` and `DipmViewType = Kokkos::View<double *, Kokkos::HostSpace>` in aosoa_pack.hpp to avoid accessing the Particle pointer in DPD and dipole interactions
- Added new interface in `gay_berne.hpp`


jngrad (approved):
LGTM

PR #5183, opened 2025-10-16
parallel particle slice getter

NiklasMueller2001:
Until now getting particle properties from `ParticleSlice` objects was very expensive since multiple expensive MPI calls **for every particle in the slice** were made.

- Moved parts of the implementation of ParticleHandle's get_attribute method to the script interface.

- Implemented optimized getter for positions, types and charges and general getter for all other attributes except special attributes that the python class still handles.


jngrad (approved):
Weak scaling benchmark for a total of 4000 particles evenly split across MPI ranks, using the following hot loop:

```python
n_loops = 30
timings = []
plist = system.part.all()
for _ in range(30):
    t1 = time.time()
    for i in range(n_loops):
        _ = plist.pos
    t = time.time() - t1
    timings.append(t / n_loops)
print(f"avg: {np.mean(timings)*1e9 / n_loops / n_part:.2f} ns/loop/part")
```

`pos` is part of the 4 properties that only trigger one MPI gather_v operation, `force` is part of the less optimized properties that have extra MPI reductions.

| MPI ranks | property | time per access | speed-up |
|-----------|----------|-----------------|----------|
| 1 | force (old) | 38.8 ns/loop/part | -   |
| 1 | force (new) | 33.8 ns/loop/part | 13% |
| 1 | pos (old)   | 39.8 ns/loop/part | -   |
| 1 | pos (new)   | 33.3 ns/loop/part | 16% |
| 2 | force (old) | 42.2 ns/loop/part | -   |
| 2 | force (new) | 34.2 ns/loop/part | 19% |
| 2 | pos (old)   | 44.4 ns/loop/part | -   |
| 2 | pos (new)   | 34.6 ns/loop/part | 22% |
| 4 | force (old) | 43.2 ns/loop/part | -   |
| 4 | force (new) | 35.1 ns/loop/part | 19% |
| 4 | pos (old)   | 44.7 ns/loop/part | -   |
| 4 | pos (new)   | 35.3 ns/loop/part | 21% |

PR #5184, opened 2025-10-16
Guard against verlet list overflow

### LINKED ISSUE #5165: Verlet List Overflow Handling
## Guard against overflow

To avoid silent failurs.

* set_overflowed(bool) and get_overflowed() members in CustomVerletList
* any thread buildin the verlet lit can set_overflowed(true) without atomics
* after Verlet List construction throw if get_overflowed()

Then, users can at least continue with linked cell by setting the use_verlet_lists=False on the cell system

## Pre-calculating maximum number of neigbors

Assumig half shells and no load balancing:
The maximum neighbors a particle in Cell c can have is
c.n_particles -1 +sum[neighbor_cell.n_particles for neighbor_cells in c.red_neighbors()])

This can be evaluated by a KOKKOs parallel_reduce()
```C++
    int max=0;
    Kokkos::parallel_reduce("max_neighbors", local_cells.size(), 
   KOKKOS_LAMBDA (const int& cell_i, int& lmax) {
        
        lmax = std::max(lmax, max_neighbors_for_cell(local_cells[cell_i]));
    }, max, Kokkos::Max<int>(max));
```

Using a view for `max` and a parallel_for rather than a reduce, this can also be collected per Cell to calculate the offsets in a CSR format.

## Alternative: Linear format + reduction

Along the lines of
```C++
struct VerletList {
  st::vecotr<std::pair<std::size_t, std::size_t>> pairs;
  // Alterntaively: 2D KOKKOS view with layout such that pairs(i,0) and pairs(i,1) are adjacent in memory
  Kkkos::view<int*> counts; // initialize to number of unique particles on the MPI rank
}
```

Phase 1: Kokkos::parallel_for over cells. Each thread builds one VerletList
Phase 2: reductoin: Glue pair vectors together and sum counts views from different threads.

This list can then used directly with a KOKKOS::parallel_for or alternatively, one can implement the Cabana::Neighborlist concept. The latter requries that neighbors for each particle are stored contiguously (no balancing of neighbors per particle)

## Alternative 2 Using Cabanas CSR Verlet List

I think this implies writing a custom Builder.
Using the maximum neihgbor count per cell from above, we know the (worst case) counts for all particles in that cell. From a cumulative sum over the counts, we calcuate the offset of the CSR data for that particle in the CSR Verlet Data Structure.
The Verlet List can then be populated using a parallel_for .
C.f. cabana/core/src/Cabana_Verletlist.hpp struct VerletListData<memorySpace, VerletLayoutCSR>

@hidekb @jngrad

hidekb:
Fixes #5165

Description of changes:
- Added new function `setOverflow()` and `hasOverflow()` in `custom_verlet_list` and automatically switch to using link cell when Verlet list overflow is detected
- Added new test function in `cell_system.py`


RudolfWeeber:
Are you sure, you have to collect this per thread? Since, during parallel execution, you only ever set it to true, I don't think you need to guard against parallel access. I think, a plain `bool` would do?

hidekb:
I changed `overflows` from `Kokkos::View` to `bool`. And I wrote the comments in the file.

RudolfWeeber (dismissed):

RudolfWeeber (dismissed):

jngrad (approved):

PR #5188, opened 2025-10-23
Introduce Magnetodynamics with thermal Stoner-Wohlfarth

stekajack:
### Magnetodynamics Overview

**Magnetodynamics** refers to the ability to evolve the degrees of freedom of a dipole moment using a defined set of rules (a *model*), such that magnetic relaxation is not strictly tied to the Brownian relaxation of a particle.

There are three relevant internal relaxation mechanisms for magnetic nanoparticles (on timescales typically interesting for MD simulations):

- **Brownian relaxation** ($\tau_B$)  
- **Néel relaxation** ($\tau_N$)  
- **Rotational diffusion of the dipole moment** ($\tau_D$)

---

### Thermal Stoner–Wohlfarth Model

The **thermal Stoner–Wohlfarth (SW) model** provides a state-of-the-art, scalable method for including Néel relaxation in simulations of single-domain magnetic nanoparticles.

Broadly speaking, the model:

1. Finds the **critical points** of the magnetic energy of a Stoner–Wohlfarth particle.  
2. Based on the **relative orientation** of the local magnetic field and the particle’s director, determines the correct orientation for the dipole moment.  
3. Computes the **probability** of the dipole moment flipping its orientation with respect to the easy axis (i.e. Néel relaxation), derived from the energy barrier between critical points.  
4. Executes a **kinetic Monte Carlo move** according to that probability.

---

### Implementation Details

The feature is implemented under the compile-time flag:  
`THERMAL_STONER_WOHLFARTH`

If enabled, the following particle properties become available:

```cpp
/** Flag indicating if the particle is real and its director should be used */
bool sw_real = false;

/** Flag indicating if the particle is virtual and its dipole moment should be evolved */
bool sw_virt = false;

/** Angle between the director and dipole moment of a SW particle */
double phi0 = 0.0;

/** Saturation magnetization of the particle */
double sat_mag = 0.0;

/** 
 * Hk = 2 * K1 / (mu0 * Ms)
 *  -> anisotropy field [A/m], where K1 is the magnetic anisotropy constant [kg/(m·s²)].
 * On the particle, we store the inverse of Hk (Hkinv) in reduced units.
 */
double Hkinv = 0.0;

/** Energy units parameter for the kinetic MC step */
double kT_KVm_inv = 0.0;

/** Time units parameters for the kinetic MC step */
double tau0_inv = 0.0;
double tau_trans_inv = 0.0;

/** Time between kinetic MC steps */
double dt_incr = 0.0;
```

References
	•	Publication: [Phys. Rev. B 111, 014438 (2025)](https://doi.org/10.1103/PhysRevB.111.014438)
	•	Optimization Library: [NLopt](https://github.com/stevengj/nlopt) (used to find critical points in magnetic energy)

⸻

Notes

This method relies on the DIPOLE_FIELD_TRACKING feature of ESPResSo to correctly incorporate dipole interactions.
For interacting systems, it should only be used with magnetostatic actors that support this feature.

⸻

Tasks:

- [x] replace marsenne-twister with philox
    - NOTE: now using rng state from thermostat
- [x] replace sw_real/sw_virtual flags with single enable_magnetodynamics=true on virtual
- [x] group magnetodynamics particle params in a typed enum magnetodynamics (similar to swimmers)
- [x] enable access to RNG seed via script interface
    - NOTE: seed and step are taken directly from thermostat, so no need for extra interface  
- [x] sw_main refactor:
  - [x] for_each loop over virtual particles
  - [x] separate external field summation
  - [x] separate main into smaler clearer steps
- [x] improve docstrings:
  - [x] write out SW abbreviations
  - [x] annotate equation numbers from reference paper in code
  - [x] docstrings for minimization helpers


stekajack:
Note: the integration test is not run as a part of CI. I choose not to meddle with the current CI setup files because there are other features that are not tested (scafacos for example), and it’s not my place to decide which features are strictly tested and which are not. Given that tSW is fairly self-contained, I thought better to leave it as is.

RudolfWeeber:
Thank you for this extensive contribution, Deniz! Magnetodynamics is a great addition for ESPREsSo.

Based on an initial read of the code, I think we can aim for inclusion in 5.0.

Some observations/suggestions:

### particle properties

* please write out the abbreviation SW in the docstrings
* I think `sw_real` and `sw_virtual` can be bundled into a single `magnetodynamics_flags` which holds a typed enum (also in view of other planned magnetodynamics methods). Rather than `real` and `virtual`, maybe `particle` and `interal_mag_moment` or something similar?
* Is the MC time-step really something that should be chosable per particle, or is this rather an algorithm parameter
* more speaking names for the other prefactors. And should the thermal energy really be part of a ratio on the particle, or should kT be a parameter of the algorithm?

### Random number generation

* We use counter-based (Philox) RNG in ESPREsSo, mainly for reproducibility reasons. I suggest using the same infrastructure for SW. 
* Access to the counter (or RNG) seed) needs to be provided via the algorithms script interface in any case for deterministic tests

### SW main function

* As I understnad the collection of sw_real and sw_virtual particles, the assumption is that they will always be iterated in the same order. I'm pretty sure, we do not guarantee this. I.e., the relationship between real and virtual particles is not stable. I think the safe way to do this is:
    * iterate over the virtual particles
    * and loop up the corresponding real particle based on the vs relative attribute (c.f. virtual_sites/relative.cpp:vs_relative_update_particles())
    * this will also let you write the main loop as an std::for_each()
* please outsource the summing of external magnetic fields from constraints to a separate function
* In the two versions of the main loop (iterating over pairs of SW real and SW virt particles), pls consider movint the respective loop bodies into separate functions.
* Also, the two versions (with and without external magnetic field, if I understand the code correctly) look somewhat similar. Can this be unified?
* Please also annotate the code with the equation numbers from the paper
* Some docstrings for the minimization helpers would help with code clarity

### Propagation

* What exaclty is the meaning of the ROT_VS_INDEPENDNET flag? Is that different from just having no rotatinal propagation scheme engaged?


### Interface and testing

* I think there is already a function that scans particle propagation flags before integration to figure out, whether, e.g. virtual sites relatvie loops need to run. SW could probably hook in there so we only run the loop, if any particle actually has the flag
* We need to add SW to the maxset configuration and add the CMake flag to download the dependency in a suitable CI runner, so the algorithm gets tested in our CI.

### Parallelism safeguards

* As far as Ï understnad the code, it requires that corresponding sw real and sw_virtual particles are on the same rank. Since they have the same position, this is guaranteed for the regular decomposition. It is also true for the AtomDecomposition (nsquare) for a single MPI rank, but not for several MPI ranks. I.e., we should probably throw an error in the unsupported cases?

stekajack:
Thanks Rudolf for the detailed review and suggestions. I will update the initial comment with a task list as I see it. Here, I will just quickly answer some of the points/questions you raised:

- The MC time parameter can be chosen by the user to correspond to the time resolution they would like, similar to how we can choose not to propagate flow fields in every integration. I don’t think of it as an algorithm parameter.

- The ratio between the anisotropy and thermal energy is per-particle, so that users can simultaneously simulate various sizes/materials. I suppose that KV could be saved on particle and the temperature could be queried from the core if necessary.

- The two branches of sw_main are there as an early opt-out in case there is no applied field, as the possible dipole moment orientation calculation simplifies significantly. The no-field branch is in general not necessary.

- ROT_VS_INDEPENDENT propagation mode is different from just disabling rotational integration. Without an integration scheme, the torques due to the dipole moment being carried by the virtual particle would not be applied to the real particle. This is necessary; otherwise, for example, ESPResSo would give magnetization for a ferrofluid that is below Langevin. The standard VS_RELATIVE scheme propagates the torque but also resets the quaternion, which we don’t want if the magnetodynamics scheme is to govern the dipole moment. Therefore, the ROT_VS_INDEPENDENT propagation mode.

RudolfWeeber (dismissed):

RudolfWeeber:
Thank you Deniz! @jngrad from my side, this PR is ready. Can you please take over for the merge?

testsuite/python/CMakeLists.txt
```diff
@@ -319,6 +319,7 @@ python_test(FILE collision_detection.py MAX_NUM_PROC 4)
 python_test(FILE collision_detection_interface.py MAX_NUM_PROC 2)
 python_test(FILE lj.py MAX_NUM_PROC 4)
 python_test(FILE dipole_field_tracking.py MAX_NUM_PROC 2 GPU_SLOTS 1)
+python_test(FILE thermal_stoner_wohlfarth.py MAX_NUM_PROC 4)
```
jngrad: Are 4 MPI ranks (and thus 3 MPI ranks in the `additional_checks` stage) providing more insights than 2 MPI ranks? As far as I understand the feature, the virtual site is always placed on top of the real particle. I'm asking this, because the non-linear optimizer takes 92% of the runtime, making the test case really slow on a code coverage build (about 2min30s), and with 4 MPI ranks and 2 OpenMP threads it consumes all 8 available cores, introducing a bottleneck in the test scheduler.

I tried to split the test into an interface test (flipping test with minima search) that always runs and a statistical tests (fluid and solid cases) that only runs in non-coverage build, but then the C++ code coverage drops.

src/core/magnetostatics/stoner_wolfarth_thermal.cpp
```diff
@@ -149,128 +153,88 @@ double get_phi_at_energy_min(double theta, double h, double phi0,
...
-        p.stoner_wolfarth_phi_0() = M_PI;
-        p.dipm() = dipm;
-        p.quat() = quat;
-      }
-    }
+    auto const dist_0 = std::abs(p.stoner_wolfarth_phi_0() - 0.);
+    auto const dist_pi = std::abs(p.stoner_wolfarth_phi_0() - pi);
+    // Compare the differences and determine the nearest angle (simplifies
+    // down to an XNOR operation, aka equality operator for boolean types)
+    kernel(flip == (dist_0 < dist_pi));
```
jngrad: Wrote a more compact form for this if/else chain. Please double-check my reasoning is correct. The provided test case didn't cover most of the branches ([code coverage report](https://app.codecov.io/gh/espressomd/espresso/commit/d479def1fb06c611c3260bd2768bd3e9f262d911/blob/src/core/magnetostatics/stoner_wolfarth_thermal.cpp#L175)), so if I accidentally introduced a regression, it might go detected.

testsuite/python/test_checkpoint.py
```diff
@@ -792,6 +792,24 @@ def test_virtual_sites(self):
...
+        Propagation = espressomd.propagation.Propagation
+        prop_flag = Propagation.TRANS_VS_RELATIVE | Propagation.ROT_VS_INDEPENDENT
+        self.assertEqual(p_real.propagation, Propagation.SYSTEM_DEFAULT)
+        self.assertEqual(p_virt.propagation, prop_flag)
+        self.assertEqual(p_real.vs_relative[0], -1)
+        self.assertEqual(p_virt.vs_relative[0], p_real.id)
+        self.assertEqual(p_real.vs_relative[1], 0.)
+        self.assertEqual(p_virt.vs_relative[1], 0.)
+        np.testing.assert_allclose(
+            np.copy(p_virt.vs_relative[2]), [1., 0., 0., 0.], atol=1e-10)
```
jngrad, line 813: Added a checkpointing test, seems to work as expected. But just for my understanding: in `Particle.hpp`, parameter `dt_incr` is not part of the `serialize()` function, and can be expected to be lost during ghost communication. Is this a quantity that gets regenerated from other parameters?

CMakeLists.txt
```diff
@@ -701,21 +701,24 @@ if(ESPRESSO_BUILD_WITH_STOKESIAN_DYNAMICS)
...
-    set(NLOPT_GUILE off CACHE BOOL "")
-    set(NLOPT_SWIG off CACHE BOOL "")
-    add_subdirectory("${nlopt_SOURCE_DIR}" "${nlopt_BINARY_DIR}")
-  endif()
+    nlopt
+    GIT_REPOSITORY https://github.com/stevengj/nlopt.git
+    GIT_TAG        v2.10.0
+  )
+  # cmake-format: on
+  set(BUILD_SHARED_LIBS off)
```
jngrad: I had to turn this off, because the .so file is not installed in the correct path. In fact, the previous code used `set(BUILD_SHARED_LIBS on CACHE BOOL "")`, which was ignored by CMake since `BUILD_SHARED_LIBS` is not a cache variable, and the variable was OFF by default, which is why the CI pipeline didn't fail.

I need to take a closer look at NLopt integration in the build system and consult with my EasyBuild contacts. Building it as a shared library is straightforward, but we then would have to manually override the install path and possibly the R-path, like we did with waLBerla, which is a bit tedious.

jngrad (commented):
Overall, looks good to me and can go in the 5.0 release. I'll try to sort out the NLopt installation step as soon as possible.

testsuite/python/test_checkpoint.py
stekajack: No, this quantity doesn't get regenerated from other parameters! It should be a part of the serialize() function with the rest of the parameters. Thanks a lot for the catch!

src/core/magnetostatics/stoner_wolfarth_thermal.cpp
stekajack: seems legit, thanks

testsuite/python/test_checkpoint.py
jngrad: A C++ unit test was added to check serialization of the struct.

jngrad (approved):

PR #5191, opened 2025-10-25
Short range force performance

RudolfWeeber:


RudolfWeeber:
# Summary of Changes: Cabana/Kokkos Performance Optimizations

This changeset implements several performance optimizations for the Cabana-based shared-memory parallelization in ESPResSo, focusing on minimizing redundant data transfers and unnecessary operations in the short-range force calculation loops.


## 1. Conditional Data Loading in Force Kernel (`src/core/forces_cabana.hpp`)
- Purpose: Minimize memory access overhead by loading particle data only when needed
- Implementation:
  - Added helper functions to check if specific algorithms are active: `gay_berne_active()`, `npt_active()`, `thole_active()`, `dipoles_active()`
  - Particle directors are now loaded conditionally (only when Gay-Berne or dipole interactions are active)
  - Particle pointers are loaded conditionally (only when exclusions exist or Thole interactions are active)
  - Added `has_exclusion()` flag check to avoid loading full particle data for exclusion checking

## 2. Optimized AoSoA Storage (`src/core/aosoa_pack.hpp`)
- Added `flags` view to store particle metadata (currently exclusion flags)
- Methods: 
  - `set_has_exclusion(int i, bool value)`: Set exclusion flag
  - `has_exclusion(int i)`: Check if particle has exclusions
- Enables fast exclusion checking without dereferencing particle pointers

## 3. Partial vs Full Commit Strategy (`src/core/short_range_cabana.hpp`)
- Split particle data commit into two categories:
  - Always commit: Position, velocity, charge, director, dipole moment (changes every step)
  - Commit on rebuild only: ID, type, exclusion flags (static during simulation)
- Implementation:
  - Added `rebuild` parameter to `commit_particle()`
  - Partial commits skip static data, reducing memory writes by ~20-30%

## 4. Force Calculation Signature Changes
- `calc_central_radial_force()` now returns `Utils::Vector3d` instead of `ParticleForce` (i.e., no torque)
- Electrostatic Layer Correction (ELC) kernels now operate on `Utils::Vector3d&` instead of `ParticleForce&` (no torque)
- Reason: Eliminates unnecessary struct construction/destruction in hot loop
- Files affected: 
  - `src/core/forces_inline.hpp`
  - `src/core/pressure_inline.hpp`
  - `src/core/constraints/ShapeBasedConstraint.cpp`
  - `src/core/electrostatics/elc.hpp`
  - `src/core/electrostatics/solver.hpp`
  - `src/core/electrostatics/coulomb_inline.hpp`

## 6. Caliper Profiling
- Added: Caliper markers for more fine-grained profiling
  - `rebuild_local_properties()` - Memory allocation and setup
  - `reset_local_force()` - Force array zeroing
  - `set_index_map()` - ID-to-index mapping
  - AoSoA commit operations (full vs partial)
  - Cabana Verlet list creation

RudolfWeeber:
I have re-enabled torques from non-central forces in constraints and pressure calculation, although, this is probably not used at presence. Still, in case someone wants a Gay-Berne-like constraint on walls in the future, not silintely dropping the torque will save that person some headaches...

jngrad (approved):

src/core/pressure_inline.hpp
```diff
@@ -66,13 +66,15 @@ inline void add_non_bonded_pair_virials(
...
+    pf.f = calc_central_radial_force(ia_params, d, dist) +
 #ifdef ESPRESSO_THOLE
-                       thole_pair_force(p1, p2, ia_params, d, dist, bonded_ias,
-                                        kernel_forces) +
+           thole_pair_force(p1, p2, ia_params, d, dist, bonded_ias,
+                            kernel_forces) +
 #endif
-                       calc_non_central_force(p1, p2, ia_params, d, dist).f;
-    auto const stress = Utils::tensor_product(d, force);
+           Utils::Vector3d{};
```
reinaual, line 75: @jngrad I guess it's something for your next cleanup commit

jngrad: Indeed! Nice catch. I'm rewriting these files anyway for my upcoming P3M Kokkos PR.

PR #5204, opened 2025-11-20
Improve heFFTe integration

jngrad:
* implement real-to-complex FFT to improve performance
* sort out data layouts for all buffers in real-space and k-space
* unit test P3M algorithm against all data layout combinations
* provide a partial heFFTe re-implementation of DipolarP3M


jngrad:
Benchmark results for 1000 particles per OpenMP thread on an AMD Ryzen 9 9950X 16-core processor:
P3M CPU weak scaling benchmark. The real-to-complex version of P3M has lost 15 percentage points of parallel efficiency at 2 threads, but this gap decreases at higher thread counts.
P3M CPU weak scaling benchmark using the time-to-solution as the performance metric. The real-to-complex version of P3M outperforms the python branch version of P3M for all thread counts.


The real-to-complex version of P3M outperforms the python branch version of P3M for all thread counts in terms of time-to-solution. In terms of parallel efficiency, we are worse off, most likely because the complex-to-complex FFT was artificially increasing the fraction of time spent executing a highly parallel FFT code. The real-to-complex FFT decreases the size of the k-space grid by a factor of almost two. The k-space differentiation is also currently purely serial code, since the iterator used in the innermost for loop "skips" the missing redundant FFT data and doesn't get incremented as quickly as the for loop indices.

Further performance improvements are planned within the development access on LUMI in the context of the 5.1 release.

src/core/p3m/P3MFFT.hpp
```diff
@@ -31,70 +32,73 @@
...
+    auto constexpr row_major_order = std::array<int, 3>{2, 1, 0};
+    auto constexpr col_major_order = std::array<int, 3>{0, 1, 2};
+    auto constexpr in_box_order =
+        (FFTConfig::r_space_order == Utils::MemoryOrder::ROW_MAJOR)
+            ? row_major_order
+            : col_major_order;
+    auto constexpr out_box_order =
+        (FFTConfig::k_space_order == Utils::MemoryOrder::ROW_MAJOR)
+            ? row_major_order
+            : col_major_order;
```
reinaual, line 76: Is it worth to extract the memory order, since it's also used in `p3m_heffte.impl.hpp`?

reinaual (approved):

src/core/p3m/P3MFFT.hpp
jngrad: These indices are actually reversed compared to those in `p3m_heffte.impl.hpp` in the `CoulombTuningAlgorithm` class, because heFFTe uses the Fortran indexing convention while ESPResSo uses the C indexing convention.

PR #5214, opened 2025-11-28
Refactor force calculation code

RudolfWeeber:


RudolfWeeber:
@jngrad could you please help with the timeouts? I wasn't able to trigger this locally.

src/core/forces.cpp
```diff
@@ -162,6 +162,98 @@ static void reinit_dip_fld(CellStructure const &cell_structure) {
...
+  auto const &local_virial = system.cell_structure->get_local_virial();
+#endif
+  auto const &aosoa = system.cell_structure->get_aosoa();
+
+  return /* ForcesKernel */ {*system.bonded_ias,
+                             *system.nonbonded_ias,
+                             get_ptr(coulomb_kernel),
+                             get_ptr(dipoles_kernel),
+                             get_ptr(elc_kernel),
+                             get_ptr(coulomb_u_kernel),
```
jngrad, line 175: Function `get_ptr()` returns a non-owning pointer to a temporary closure that expires when returning from this function. I'm currently working on rewriting these lambdas and should have a working prototype tomorrow. We can then build upon my work to completely remove these temporary lambdas.

jngrad:
To follow-up on the idea of extracting the P3M pointer if it is active (branch `jngrad/p3m_shortcut`), the speed-up is about 1.5%. The inlining is somehow degraded by the else branch, probably because `coulomb_kernel` contains P3M as a possible jump and the compiler is trying to reduce code duplication in the first branch by recycling some of the jumps.

I tried generating the Cartesian product of `ForcesKernel` symbols with compile-time optimizations for each combination of electrostatics and magnetostatics solvers, like so:
```c++
template <typename CoulombSolver, typename DipolesSolver,
          bool HasCoulombSolver, bool HasDipolesSolver, bool HasNpT>
struct ForcesKernel;
```
The idea was to avoid the expensive runtime checks on the type index of the variants for each particle pair, by resolving them once at the beginning of integration. But then we have to store the template specialization in an opaque pointer, and I was not sure how this would interfere with Kokkos inlining pragmas.

jngrad (approved):

PR #5216, opened 2025-12-03
Tie thermal Stoner-Wohlfarth to Langevin

jngrad:
Description of changes:
- thermal Stoner-Wohlfarth now requires the Langevin thermostat
- improve code coverage of thermal Stoner-Wohlfarth

### DIFF
--- a/src/core/thermostat.cpp
+++ b/src/core/thermostat.cpp
@@ -87,62 +87,6 @@ void Thermostat::Thermostat::philox_counter_increment() {
   }
 }
 
-uint64_t Thermostat::Thermostat::get_philox_counter() const {
-  if (thermo_switch & THERMO_LANGEVIN) {
-    return langevin->rng_counter();
-  }
-  if (thermo_switch & THERMO_BROWNIAN) {
-    return brownian->rng_counter();
-  }
-#ifdef ESPRESSO_NPT
-  if (thermo_switch & THERMO_NPT_ISO) {
-    return npt_iso->rng_counter();
-  }
-#endif
-#ifdef ESPRESSO_DPD
-  if (thermo_switch & THERMO_DPD) {
-    return dpd->rng_counter();
-  }
-#endif
-#ifdef ESPRESSO_STOKESIAN_DYNAMICS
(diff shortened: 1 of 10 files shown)


src/core/thermostat.cpp
```diff
@@ -87,62 +87,6 @@ void Thermostat::Thermostat::philox_counter_increment() {
...
-#ifdef ESPRESSO_STOKESIAN_DYNAMICS
-  if (thermo_switch & THERMO_SD) {
-    return stokesian->rng_counter();
-  }
-#endif
-  if (thermo_switch & THERMO_BOND) {
-    return thermalized_bond->rng_counter();
-  }
-  return 0;
-}
```
jngrad, line 116: @stekajack This code raised concerns w.r.t sustainability. There is one philox counter per thermostat, and this list takes the first one available. When switching thermostat from e.g. Langevin to Brownian, and using the same seed, the random noise generator would cycle through the same random numbers again. This function was only tested in the first line of code. While writing new test cases to improve code coverage, I found out this function was missing the LB thermostat. Since the LB philox counter doesn't necessarily get incremented at every MD time step, adding LB in this list would lead to correlated random numbers.

We think the long-term solution would be to introduce a Stoner-Wohlfarth thermostat. Since the 5.0 release is only a few weeks away, we settled on Langevin for now, which is the most relevant use case for this feature.

stekajack: The way I see it, tSW is not a thermostat but essentially a propagation scheme for the rotational degrees of freedom of particles. It should therefore stay in step with the state of a thermostat. The LB thermostat state should also increment at every integration, because the particles are being thermostatted at every integration. It’s the flow field that does not necessarily have to be updated at every timestep, and thus should have its own counter. I don’t think it’s a good solution to block functionality for a range of thermostats just because one thermostat is inconsistent with the rest. It’s acceptable for the release, but there should be a discussion on how to proceed, especially in light of the additional methods that will be introduced in the magnetodynamics solver family.

reinaual (approved):

Issue #4954, opened 2024-07-16
Allow passing ParticleHandle and ParticleSlice to observables


RudolfWeeber:
rather than numerical ids, as it is required right now.

This would include

* renaming the `ids` property to `particles` or the like
* writing a python wrapper for the getter of that attribute that maps between ParticleHanldes/ParticleSlice and numeric ids.
* adapt samples/tests/tutorials

Nikhil-4595:
Hey @RudolfWeeber and @jngrad, as my mentor already mentioned the eligibility to work on two issues. I have worked on #5039, and the second one is this. I will start working on it, but I have one doubt regarding Observables. Since we have different observables accepting different parameters, shall I make changes to only observables having ids based on #3992 last discussion?

jngrad:
This task requires a significant API change that will impact many users. Since we are in the process of releasing ESPResSo 5.0, I feel rather uneasy about it. If you insist on working on it, be aware that we won't be able to review it until after 5.0 is released, which might take a few weeks.

Nikhil-4595:
Hey @jngrad, no problem about that. I will work on it and you can focus on the current release. However, can you please confirm this :

Currently, particle-based observables require explicit integer IDs, for example:

obs.ParticlePositions(ids=[1, 2, 3])

To confirm the intended direction of this issue, I want to clarify whether the following renamed and simplified API is acceptable, with ids fully replaced by particles.

Proposed new usage examples:

Passing a list of particle handles

particles = [p1, p2, p3]
obs.ParticlePositions(particles=particles)


Passing a ParticleSlice directly

sel = system.part[1:4]
obs.ParticlePositions(particles=sel)


Passing a single ParticleHandle

p = system.part[2]
obs.ParticlePositions(particles=p)

Internally, these inputs would be converted to numeric IDs particle before passing further, while preserving order. The existing ids argument would be removed rather than kept as an alias.

Is this correct? And shall the renaming be done to all Observables accepting ids as parameters?

jngrad:
Your proposed interface sounds reasonable to me.

Nikhil-4595:
Great, then I will work on this as proposed. After the relevant changes, I will create a PR, and then you can look into it whenever you will have time. Thanks.

PR #5226, opened 2026-01-08
Rename and document `Analysis.particle_energy()`

### LINKED ISSUE #5039: system.analysis.particle_energy() rename/clarify
This should probably be particle_non_bonded_energy(), and the docstring should clarify whether or not short range electrostatic contribtuions are included.

Nikhil-4595:
Fixes #5039

Description of changes:
- API change: rename `Analysis.particle_energy()` to `Analysis.particle_non_bonded_energy()` to better reflect the calculated quantity (kinetic, bonded, electrostatic and magnetostatic contributions are not part of this energy)
- bugfix: Thole corrections are no longer part of the energy calculated by `Analysis.particle_non_bonded_energy()`
- improve documentation of the per-particle non-bonded energy analysis function and of the Thole correction

### DIFF
--- a/src/python/espressomd/analyze.py
+++ b/src/python/espressomd/analyze.py
@@ -335,7 +335,7 @@ class Analysis(ScriptInterfaceHelper):
         -------
         :obj:`float`
             Short-range non-bonded energy contribution of that particle.
-            
+
     particle_energy()
         Deprecated alias for :meth:`particle_non_bonded_energy`.
 
--- a/src/python/espressomd/analyze.py
+++ b/src/python/espressomd/analyze.py
@@ -529,29 +529,6 @@ class Analysis(ScriptInterfaceHelper):
         """
         return self.call_method("particle_energy", pid=particle.id)
 
-    def particle_energy(self, particle):
-        """
-        Deprecated alias for :meth:`particle_non_bonded_energy`.
-
-        Parameters
-        ----------
-        particle : :class:`~espressomd.particle_data.ParticleHandle`
-
-        Returns
-        -------
-        :obj:`float`
-            Short-range non-bonded energy contribution of that particle.
-
-        """
-        import warnings
-
-        warnings.warn(
-            "particle_energy() is deprecated, use particle_non_bonded_energy() instead",
-            DeprecationWarning,
-            stacklevel=2,
-        )
-        return self.particle_non_bonded_energy(particle)
-
     def particle_bond_energy(self, particle, bond):
         """
         Calculate the bonded energy for the given particle and bond.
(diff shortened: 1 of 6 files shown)


src/python/espressomd/analyze.py
```diff
@@ -506,22 +506,50 @@ def energy(self):
...
-    def particle_energy(self, particle):
+    def particle_non_bonded_energy(self, particle):
         """
-        Calculate the non-bonded energy of a single given particle.
+        Calculate the short-range non-bonded energy contribution associated with
+        a single particle.
+
+        This includes only short-range non-bonded interaction terms (e.g. Lennard-Jones,
+        WCA, etc., depending on enabled features). Electrostatic energy contributions
+        are not included.
```
jngrad: This statement about electrostatics energies is in contradiction with the summary you wrote in the PR first comment. In addition, it would probably be better if your summary was more concise. It epilogues on `add_non_bonded_pair_energy()`, which as far as I know doesn't participate in `particle_energy()`.

The summary also doesn't explain why `particle_energy()` needs the Coulomb kernel if it doesn't contribute to the non-bonded forces. I think it has to do with Drude oscillators, and it is not obvious to me we can state Coulomb is excluded from the particle energy since the Thole correction for Drude oscillators is included in the particle energy (seems like a bug to me). What's your take on this @RudolfWeeber ?

Nikhil-4595: Thanks for pointing this out - you’re right, that statement can be misleading.

My intention with the docstring was only to clarify that long-range (k-space) electrostatic contributions are not part of the per-particle energy, but I agree that stating “electrostatic energy contributions are not included” is incorrect given the presence of Coulomb kernels and possible short-range terms (e.g. Thole corrections for Drude oscillators).

I’m happy to edit the comment summary to use more conservative wording (e.g. explicitly excluding only long-range electrostatics, without making claims about all Coulomb contributions). Before doing so, could you please confirm what level of detail you’d prefer here?

Thanks again for the careful review.

Nikhil-4595: Hey @jngrad, based on your input, I have changed the docstring and summary based on the findings.  Thanks :)

jngrad: After looking into the Thole paper, I'm pretty confident that its inclusion in the per-particle non-bonded energy is a bug in ESPResSo. The Thole correction is not meaningful without the electrostatic contribution, and both the short-range and long-range parts of electrostatics solvers are excluded from the non-bonded energy calculation.

I'll take over the PR from here, since it requires tricky changes to the C++ core and a new test to prevent this bug from being re-introduced again in the future.

Nikhil-4595: Hey @jngrad, Sure. Thanks for letting me know. Moreover, I have been looking into another issue #4954 which i want to work on. Can you please adddress the doubt I mentioned in the comments there.

jngrad (approved):
Thanks a lot!

PR #5228, opened 2026-01-13
Improve LB+LEbc domain decomposition

### LINKED ISSUE #5023: LB+Lees-Edwards doesn't work for certain node grids
The shear profile from lb plane couette flow is wrong, if the mpi node grid is not 1 in the shear_normal directoin. This case is currently not checked for.
Most likely cause is that the Lees Edwards interpolation kernels are always run on the upper and lower ghost layers in LBWalberlaImpl. However they should be only run on ghost layers which actualy represent a domain boundary.

jngrad:
Fixes #5023

Description of changes:
- allow LB domain decomposition along the shear normal direction


maintainer/walberla_kernels/lees_edwards.py
```diff
@@ -58,25 +67,41 @@ def velocity_offset_eqs(config, method, pdfs,
...
-    grid_size = TypedSymbol("grid_size", dtype=default_dtype)
+    lebc_top_index = TypedSymbol("lebc_top_index", counters[1].dtype)
+    lebc_bot_index = TypedSymbol("lebc_bot_index", counters[1].dtype)
 
     # When the kernels are combined, the integration loop also reaches
     # one grid cell further to pull the populations. This changes the
     # position where to apply the LE boundary by one cell.
-    boundary_offset = [0, grid_size - 1]
+    boundary_offset = [0, -1]
     if combined_kernel:
```
Yashas17, line 77: Does it make sense to have this `if` statement here, since we are anyways using combined kernel? We can perhaps remove this extra parameter altogether since we don't have separate stream and collide kernels.

maintainer/walberla_kernels/lees_edwards.py
Yashas17, line 1: Overall, I understand how the logic works, but don't understand how diff, and _diff_wrt work and what do they affect.

Yashas17 (commented):
Overall the changes look good to me, except a small part I don't understand and possibility of removing the combined_kernels option, which might not be needed.

maintainer/walberla_kernels/lees_edwards.py
jngrad: I couldn't remember if we still have long-term plans with the push scheme, and chose to leave it in.

maintainer/walberla_kernels/lees_edwards.py
jngrad: Try removing them and see what happens :-)

The `CastFunc` is required to transform variables of any type (e.g. `int64_t`) into a type that supports packing (e.g. `__m256d` for a double 4-pack). The implementation details of `CastFunc` involve the definition of a multivariate sympy function `f(x,t)` where `t` is the type to convert `x` to, and `t` is defined as a sympy Symbol, which has no definition for a partial derivative. Thus one cannot derive `f` with respect to `t`, nor `t` with respect to another variable, and symbolic differentiation fails. We cannot derive `t` from a new class `CppType(BasicType)` that implements symbolic differentiation, because pystencils throws an exception during the simplification stage if there is any object left in the expression that is not of type `BasicType` (must be an exact match, they don't use `isinstance`). So we bind the symbolic differentiation methods to the object directly. These methods implement the behavior of a constant: zero is returned when derived w.r.t any variable.

Yashas17: Got it, then everything looks fine and ready to merge.

Yashas17 (approved):

hidekb (approved):
LGTM!

PR #5229, opened 2026-01-14
Grand canonical Monte Carlo tutorial

### LINKED ISSUE #5213: Grand canonical Monte Carlo tutorial has convergence issues
The deviation from ideality in the last plot is sensitive to initial conditions and has low reproducibility. When Coulomb interactions are deactivated, the first data point doesn't fit on the ideal curve. One possible explanation could be the number of particles is too small.

hidekb:
Fixes #5213 

In original tutorial, the ion ration R=c_{M-}^{sys}/c_{X+,X-}^{res} spans the range 0.5 < R < 50.
At R = 50, the ion number in the system is extremely low, and the insertion acceptance is vanishingly small.
As expected, the results then depend strongly on the initial configuration and the random seed.

Description of changes:
-Added Ideal-ion (no interaction) calculation
-Implemented log-based acceptance for numerical stability, e.g. a vanishingly small acceptance rate.
-R range limited to the physically relevant region, 0.17 < R < 1.7, where the Donnan theory and the distinction between ideal and interacting ions become evident, as shown in Fig. 2a,b of the paper: "Grand-Reaction Method for Simulations of Ionisation Equilibria Coupled to Ion Partitioning" (https://pubs.acs.org/doi/10.1021/acs.macromol.0c00260).


jngrad:
@pm-blanco would you be willing to review this PR this week, or nominate someone familiar with the Monte Carlo C++ code? @hidekb had to rewrite the C++ code to work in logspace to avoid loss of precision that lead to numerical instabilities in the Grand canonical ensemble. A side-effect of this refactoring is a silent change in the Python interface to `ReactionAlgorithm.calculate_acceptance_probability()` which now returns the logarithm of the acceptance probability. It looks like a private function (it's not even tested), but it is accessible to users and could lead to surprises. ESPResSo 5.0 is a major release, and as such allows silent API changes. We are also going to replace all of this code by your Python implementation after 5.0 is released. Nevertheless, if you think this is a function that shouldn't change behavior silently, we can rename it to make this a non-silent API change.

pm-blanco:
@jngrad I will take a look during the week. On a quick look, my first impression is that the change `ReactionAlgorithm.calculate_acceptance_probability` should not be silent if it is accessible to users, specially if returns something (slightly) different than it used to, but let me think about it.

src/core/reaction_methods/ReactionAlgorithm.cpp
```diff
@@ -270,17 +271,15 @@ ReactionAlgorithm::create_new_trial_state(int reaction_id) {
...
+double ReactionAlgorithm::make_reaction_mc_move_attempt_logarithmic(
+    int reaction_id, double ln_bf, double E_pot_old, double E_pot_new) {
   auto constexpr exp_min = -708.4; // for IEEE-compatible double
   auto const exponent = -(E_pot_new - E_pot_old) / kT;
   auto const exponential = (exponent < exp_min) ? 0. : std::exp(exponent);
   auto &reaction = *reactions[reaction_id];
   reaction.accumulator_potential_energy_difference_exponential(
       std::vector<double>{exponential});
-  if (get_random_uniform_number() >= bf) {
+  if (-get_random_logarithmic_number() >= ln_bf) {
```
pm-blanco, line 283: The logarithmic acceptance MC criterion is correct, but it is not immediately obvious why.
`get_random_logarithmic_number()` draws from `std::exponential_distribution`, i.e. it returns an Exp(1) variate corresponding to `−log(u)` with `u ∼ U(0,1)`. 
With this convention, the condition
```c++
-get_random_logarithmic_number() >= ln_bf
```

is algebraically equivalent to the previous uniform MC criterion  `u >= bf`. However, this double change of sign is not obvious and confused me at first sight.

I suggest adding a short comment or renaming the helper (e.g. get_random_minus_log_uniform()) to make this equivalence explicit and avoid confusion about sign conventions.

src/core/reaction_methods/tests/reaction_methods_utils_test.cpp
```diff
@@ -26,18 +26,22 @@
...
+                              std::log(std::tgamma(N + nu + 1));
   };
 
   for (int N0 = 0; N0 < 6; ++N0) {
     for (int nu = -4; nu <= 4; ++nu) {
-      auto const val = factorial_Ni0_divided_by_factorial_Ni0_plus_nu_i(N0, nu);
+      auto const val =
+          ln_factorial_Ni0_divided_by_factorial_Ni0_plus_nu_i(N0, nu);
       auto const ref = reaction_ensemble_combinations(N0, nu);
+      std::cout << N0 << " " << nu << " " << val << " " << ref << std::endl;
```
pm-blanco: Consider removing the `cout`  statement.

src/python/espressomd/reaction_methods.py
```diff
@@ -425,7 +430,7 @@ def reaction(self, steps):
 
     def calculate_acceptance_probability(self, reaction_id, E_pot_diff):
```
pm-blanco: I would be in favor of renaming this function to something like `log_calculate_acceptance_probability` just to avoid surprises to any user that might be using the function and make the API non-silent. Even considering the plans to re-factor the MC code once the 5.0 release is done, this version of espresso might be the stable one for a while and it makes no harm to make explicit the change.

pm-blanco (changes requested):
In general, I think that implementing the MC criterion in the logarithmic scale is a good idea, since it should be more robust against numerical inaccuracies. I checked the code and the implementation and it looks correct to me. For my peace of mind, I also compiled the version of espresso in this PR and ran the pyMBE functional testsuite which also ran without problems. Looking back at the MC C++ code, there are a few bits that are a bit of an spaguetti, but that goes beyond this PR and should be addressed during the refactoring of the MC code in Python in any case. I also checked the GCMC tutorial, which I quite like. I only have a few minor comments that can be quickly addressed and I will approve the PR.

src/core/reaction_methods/ReactionAlgorithm.cpp
jngrad: I'll add a comment to this effect. I also had a hard time convincing myself of the validity of this transformation, but here is a visual proof:
```c++
// g++ -std=c++20 main.cpp && ./a.out

#include <cmath>
#include <iomanip>
#include <iostream>
#include <map>
#include <random>

void make_histogram(auto const &kernel) {
  std::map<int, int> hist;
  for (int n = 0; n != 1'000'000; ++n)
    ++hist[static_cast<int>(8. * kernel())];

  for (auto const& [x, y] : hist) {
    auto len = y / 1600;
    std::cout << std::fixed << std::setprecision(1)
              << x / 8. << '-' << (x + 1) / 8.
              << ' ' << std::string(len, '*') << '\n';
    if (len == 0) break;
  }
}

int main() {
  std::mt19937 m_generator(42);
  m_generator.discard(1'000'000);
  std::exponential_distribution<double> m_exponential_distribution(1.);
  std::uniform_real_distribution<double> m_uniform_real_distribution(0., 1.);
  auto get_random_logarithmic_number = [&]() {
    return m_exponential_distribution(m_generator);
  };
  auto get_random_logarithmic_number_from_uniform = [&]() {
    return -std::log(m_uniform_real_distribution(m_generator));
  };

  make_histogram(get_random_logarithmic_number);
  make_histogram(get_random_logarithmic_number_from_uniform);
}
```

src/core/reaction_methods/tests/reaction_methods_utils_test.cpp
jngrad: done

src/python/espressomd/reaction_methods.py
jngrad: I agree.

pm-blanco (approved):
LGTM

jngrad (approved):
@hidekb @pm-blanco Thanks a lot!

PR #5230, opened 2026-01-15
Change Python API of GPU algorithms

jngrad:
API changes:
- when an algorithm has two implementations, one on CPU and one on GPU, fuse them into a single Python class and add a constructor argument `gpu=True` to select the GPU backend
- strip suffix "Walberla" from LB-related classes, for consistency with EK-related classes
- prevent the creation of more than 1 virtual site center of mass (VS COM) per molecule id
- remove `vs_com` member of the particle class, since this information can be stored in `mol_id` directly


jngrad:
Note to reviewers: these changes don't really break backward compatibility with ESPResSo 4.2, since they only affect features whose API either already broke in a previous change (long-range solvers and waLBerla-based algorithms), or did not exist back then (VS COM). Still, they need to be labeled as API changes and appear in the changelog for ESPResSo 5.0.

testsuite/python/ek_eof.py
```diff
@@ -63,7 +63,7 @@ def test_eof(self):
 
         density = 0.0006
 
-        lattice = self.ek_lattice_class(n_ghost_layers=1, agrid=self.AGRID)
+        lattice = self.ek_lattice_class(n_ghost_layers=2, agrid=self.AGRID)
```
hidekb, line 67: I understand that this PR modifies the Python interface only; therefore, the computational results should remain consistent with the previous implementation. Could you clarify the rationale for changing the value of `n_ghost_layers`?

hidekb (approved):

testsuite/python/ek_eof.py
jngrad: The EK code prints a warning when the ghost layer thickness is 1 and throws an error when using two or more MPI ranks, because the flux boundary requires information past the first ghost layer. I fixed it here but a few other tests might be affected. I intend to address this more thoroughly in a future PR, and also revisit the warning message because it complains about multiple MPI ranks even when only 1 rank is used (most likely a copy-paste error).

hidekb (approved):
LGTM!

PR #5231, opened 2026-01-19
Unified Thermalized and Lees-Edwards Kernel Generation

Yashas17:
Description of changes:
- Thermalized and Lees-Edwards kernels now use the same LB collision rule.
- Unified thermalized and Lees-Edwards kernel generation function.


jngrad (approved):
Thanks a lot!

PR #5237, opened 2026-01-26
Cabana: base max neighbor estimation on local density

RudolfWeeber:
For the Cabana based Verlet list, we need to know the maximum number of neighbors any particle has before the list creation is begun.
The existing code made an assumption about the local density (it considered the number of particles but not the local box size). Hence, the code failed for high densities, e.g., for a dpd fluid, and fell back to linked cell.

This PR bases the heuristics on the local partilce denisty (local here means: for the mpi rank)


hidekb (dismissed):
I understand that this PR reduces the likelihood of falling back to the linked cell by refining the estimation of max_counts using local density (e.g., local particle number normalised by local volume). This looks good to me.

jngrad (approved):

PR #5240, opened 2026-01-29
Early exit from short range kernels (cabana and classic)

RudolfWeeber:
This exits early from the short range kernels, when particles are further apart than the maximal cutoff (non-bonded AND others)
wq
As a result, slightly larger skins are possible. The Lj benchmark is faster by some 10%, for P3M it's less.


hidekb (commented):
I understand that this PR introduces an early exit from the short-range kernel before generating `ParticleForce` and `ia_params`, which improves computational efficiency. However, I am concerned about the coexistence of two similarly named variables, `max_cutoff` and `ia_params.max_cut`, which may cause ambiguity. It might be clearer to rename `max_cutoff` to `system_maximal_cutoff` to better reflect its role and avoid potential confusion.

jngrad (approved):

PR #5244, opened 2026-02-02
Shared memory parallelization for the bond loop

hidekb:
Description of changes:
- Convert arrays of structs (AoS) to structs of arrays (SoA) to improve the performance of the bond loop.
- The bond loop is decomposed into three distinct components, namely pair bond, angular relationships, and dihedral configurations.


RudolfWeeber:
I wonder whether we can reduce the complexity of the bond storage creation a bit and avoid the global comm.
In this, I'd trade simpler code there for slightly lower performance for the bond-brekage and collision deteciotn features.

Would the following work.

1. Bond storage format: (particle index, bond type, parter idx1, partner_idx2, partner_idx3), i.e 5 ints per bond independent of bond types. For a pair bond, we just ignore partner_idx2 and partner_idx3. I.e. do the id_to_index translation when constructing the bond storage
2. Alongside the unique_particles create a bond_counts view that has the number of bonds par particle. This can be populated either when constructing unique_particles or in a separate parallel_for
3. Use an exclusive prefix sum on the bond_conts view (Se parallel_scan in "Kokkos "parallel dipatch" chapter) to construct a particle_bond_offsets view. This (multiplied by 5 for the size of the bond info) will provide the storage position in the bond storae for the bond info of a specific particle
4. Use a parallel_for over the particles to populate the bond storage. Since the storage indices for the bonds of all particles are knwon, this avoids the atomic.
5. Bond force calculation can then be done in a parallel_for over the number of bonds in the bond storage.
6. the livetime of the bond storage is linked to the one of unique_particles

Collision detection and bond_breakage;

* After adding/removing bonds in collision detection and bond breakage: Cacl the same bond storage construciton code as above. If the performance turns out to be unaccepptable, we can implement on-the-fly updates of the bond storage. I'd prefer this until it becomes clear from profiling that its worth it.

RudolfWeeber:
Thank you.
The overall design is good.


Below are some points raised by Claude Code.


Remarks:
* I'd defer point 2. This is a task in its own that we take on later.
* Point 3 is correct, imo. We need to use mass=1 if ESPRESSO_MASS is not compiled in
* Regarding 4: it is not clear to me that registering the bond error will actually prevent the bond loop from running. If it does run, there may well be a segfault. We can either increment the counters, even if the partners can't be resolved, or have a bond_list.valid flag, which can be set to false (irrespective o fparallelism) if any thread cannot resolve partners.
* Point 5 is probably a false positive. If reduction of ghost forces wouldnt work, we'd have seen it with short range forces.
* 6 and 9: can we unify the serial and shared mem CellStructure::add_new_bond()   and merge it with src/core/bonds.cpp:add_bond()
* 7: confirmed
* 8: not idea, if the fence is needed.
* 10: I think, we should add dihedrals. There is currently no client, but whenever someone should add one, it will be hard to track the issue.
the others I didn't check.

Here's part of the Claude Code output.



## Critical Issues

1. omp_get_thread_num() is not portable across Kokkos backends
bonds_kokkos.hpp:87,178,246 — All three kernels use omp_get_thread_num() to
index thread-local force buffers. With Kokkos::Threads (pthreads) backend,
this returns 0 for all threads, causing silent data races and wrong forces.
The TODO comments acknowledge this but it remains unfixed. Use
Kokkos::OpenMP::impl_hardware_thread_id() or a team policy with
team.team_rank().

2. KOKKOS_INLINE_FUNCTION on host-only code
bonds_kokkos.hpp:67,161,231 — The operator() bodies use std::optional,
std::variant, std::get_if, std::mutex (via bond breakage), and
std::unordered_map::at() — none of which are device-callable.
KOKKOS_INLINE_FUNCTION expands to __host__ __device__ with CUDA, so this will
fail to compile with a GPU-enabled Kokkos. Since all views use HostSpace,
consider using plain inline or KOKKOS_FUNCTION with explicit HostSpace
execution policies.

3. ThermalizedBond divergence when ESPRESSO_MASS is not defined
bonds_kokkos.hpp:100-119 — The Kokkos path wraps ThermalizedBond handling in
#ifdef ESPRESSO_MASS. Without that define, the bond falls through to
calc_bond_pair_force, which throws BondUnknownTypeError. The serial path
handles this case (using default mass 1.0). This is a behavioral regression in
 the Kokkos path.

4. Bond count mismatch between counting and filling
CellStructure.cpp:276-299 vs 239-257 — set_index_map counts all bonds (only
checks !partner_ids().empty()), while update_bond_storage only fills bonds
whose partners can be resolved via resolve_bond_partners. Unresolvable bonds
leave m_local_*_bond_numbers overcounted, causing the Kokkos kernel to iterate
 over uninitialized view slots.

5. Ghost particle bond forces may be stranded
bonds_kokkos.hpp / forces.cpp:213-264 — Bond kernels write forces for partner
particles into local_force(j, thread_id, ...). If partner j is a ghost
particle, reduce_cabana_forces_and_torques must include that index in its
reduction loop, or those forces are silently dropped. Verify that
unique_particles includes ghost entries that appear as bond partners.


## Important Issues

6. add_new_bond called outside get_local_particle guard
BindAtPointOfCollision.cpp:169-172 — The add_new_bond call is unconditional,
but the actual bond insertion into p->bonds() is inside an if (auto p = ...)
guard. This creates phantom bond entries in the Kokkos bond list when the
particle doesn't exist locally.

7. "inter" / "intra" labels swapped
short_range_cabana.hpp:154-158 — intra_kernel is launched with label "inter"
and vice versa. Misleading for profiling tools.

8. No Kokkos::fence() between bond loops and pair loop
short_range_cabana.hpp:322-341 — The three bond parallel_for calls have no
fence before the pair force loop. On OpenMP this is synchronous in practice,
but not guaranteed by the Kokkos programming model.

9. BindCenters collision protocol never calls add_new_bond
New bonds from BindCenters are added via ::add_bond to particle bond lists but
 never registered in the Kokkos bond storage. They only become visible after
the next full Verlet list rebuild. Should be documented or fixed.

10. Dihedral bonds excluded from rebuild_bond_list
CellStructure.cpp:425-433 — rebuild_bond_list only handles pair and angle
bonds. No assertion or comment documents this limitation.

11. Dead code in rebuild_bond_list_impl
CellStructure.cpp:409-421 — The else branch (in-place overwrite) is
unreachable given the counting logic. Should be removed or documented.

## Minor / Code Quality

12. AoSoA_pack::id_to_index is dead code — Declared but never allocated in
resize(). Remove it.

13. bond_broken_error(std::span<const int>) overload is unused —
Declared/defined but never called. Remove or add a call site.

14. [[maybe_unused]] on NCols — It's actually used; the attribute is
misleading.

15. bonds_cabana_test.cpp uses wrong macro names — Guards on
SHARED_MEMORY_PARALLELISM instead of ESPRESSO_SHARED_MEMORY_PARALLELISM, so
the test body is never compiled. The test always passes vacuously.

16. const & on ForceType in BondsKernelData — Kokkos View const semantics
allow mutation of data through a const view. Technically correct but
misleading to readers.

RudolfWeeber:
What ws the reason for the additoin of kvisc=[nearly ero] in the oif_volume_conservation test?

RudolfWeeber:
Given the rather large number of member variables this adds to the CellStructure, would it be possible to move all the bond list managment business inclduing the clearing/allocating functions into its own class, say LocalBondState and only hold the unique pointer to that in the CellStructure?

hidekb:
> What ws the reason for the additoin of kvisc=[nearly ero] in the oif_volume_conservation test?

I modified the implementation within the conditional branch `if (kvisc > tiny_oif_elasticity_coefficient)`, and the accompanying test script must execute this path to satisfy coverage requirements. The comment within this branch indicates that it is not fully implemented. Hence, we may disable it entirely by commenting it out. At the time, I used a very small value of `kvisc` for preliminary testing.

hidekb:
> Given the rather large number of member variables this adds to the CellStructure, would it be possible to move all the bond list managment business inclduing the clearing/allocating functions into its own class, say LocalBondState and only hold the unique pointer to that in the CellStructure?

I refactored the code. Now, variable- and function-related bond storage has been moved from CellStructure to LocalBondState.

RudolfWeeber (dismissed):

jngrad (approved):
Node-level performance improvement: a LJ simulation with harmonic bonds is now faster with 4 OpenMP threads and 1 MPI rank vs. 1 OpenMP thread and 4 MPI ranks.

Issue #5036, opened 2025-02-03
Prevent implicit bool/int conversion in script interface


RudolfWeeber:

adr1ja:
Hi, I would like to work on this.

To prevent these implicit conversions and improve overall type safety, I plan to introduce static type checking infrastructure to the project.

My plan is to add a `mypy.ini` configuration file to the root and apply strict type hints to the `espressomd.visualization` module as a first step to demonstrate the setup.

jngrad:
Hi @adr1ja,

"script interface" has a special meaning in ESPResSo: it refers to a custom C++ framework we designed to interface the C++ core with Cython and Tcl. More details can be found [on the wiki](https://github.com/espressomd/espresso/wiki/Software-design#user-content-Script_interface).

Python type hints are not used consistently in the project yet, mainly due to a concern that types declared in the docstrings might get out-of-date with the type hints (at some point I'll have to look into [`sphinx-autodoc-napoleon-typehints`](https://pypi.org/project/sphinx-autodoc-napoleon-typehints/)).

adr1ja:
Hi @jngrad,

Thanks for the clarification regarding the C++ script interface framework; I'll keep the distinction in mind.

Regarding the type hints synchronization concern: I can address this directly in my PR. I can set up `sphinx-autodoc-typehints` (or the napoleon extension) in the build configuration. This will automatically populate the documentation from the Python type signatures, effectively removing the maintenance burden of keeping docstrings in sync manually.

I will start by setting up this infrastructure and applying it to a single module (e.g., `espressomd.visualization`) as a proof of concept.

jngrad:
Please be aware this Sphinx extension is currently not available in the docker images used to test ESPResSo. We currently cannot easily update our infrastructure due to the upcoming ESPResSo 5.0 release.

adr1ja:
Hi @jngrad, I will drop the Sphinx/mypy infrastructure plan due to the infrastructure freeze and focus on modifying the C++ get_value logic to enforce strict type checking between bool and int.

PR #5246, opened 2026-02-03
core: prevent implicit bool conversion in ScriptInterface

adr1ja:
Fixes ##5036

Description of changes:
- 
This PR addresses issue #5036 by enforcing strict type checking for boolean values within the ScriptInterface framework. I modified the allow_conversion template trait in src/script_interface/get_value.hpp to explicitly exclude bool from implicit conversions to arithmetic types (integers and floating-point numbers). This ensures that boolean arguments passed from the Python interface are no longer silently cast to 0 or 1 in the C++ core, preventing potential silent physics bugs.

Also, verified the fix with a runtime check:
`./pypresso -c "import espressomd; s=espressomd.System(box_l=[1,1,1]); s.part.by_id(True)"`

image

The system now correctly raises a `RuntimeError`: `Provided argument of type 'bool' for parameter 'p_id' is not convertible to 'int'`.


adr1ja:
Hi @jngrad, the C++ fix is verified and the style/formatting issues are now resolved (the CI style check is green).

However, the build stage is failing because this change is enforcing strict type safety on existing tests and tutorials that were relying on implicit bool-to-int conversions. For example, I see failures in the default and tutorials-samples jobs.

Since this is a breaking change for the current test suite, how would you like me to proceed? Should I attempt to fix the offending tests?

jngrad:
I think you should identify and report here which parts of the code rely on implicit bool-to-int conversions. I suspect you will need to use GDB to find out some of the culprits. Bear in mind that implicit bool-to-int conversion is the Pythonic behavior, so there might be legitimate reasons for features to leverage that behavior. This ticket is about making the ESPResSo behavior diverge from the Python behavior, so we need to be able to justify why we are making this change. Here are instructions on how to use [GDB with ESPResSo](https://espressomd.github.io/doc/running.html#gdb) (you won't need MPI) and how to [run checkpoint tests](https://github.com/espressomd/espresso/wiki/Testing#user-content-Running_tests-2).

For more context: [cannot subclass from bool type](https://mail.python.org/pipermail/python-dev/2002-March/020822.html), which also explains why `isinstance(np.True_, bool)` evaluates to `False` and one has to use `isinstance(x, (bool, np.bool_))` to check if a value is a boolean type ([NumPy types docs](https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.bool)), and why Cython created the special `bint` type ([Cython types docs](https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#types)).

PR #5172, opened 2025-09-16
Modernize zndraw interface

reinaual:
Fixes #5171, fixes #5035

- Removed unused/unnecessary code
- Fixed zndraw interface not respecting PBC
- (Re-?) Introduced bonds to zndraw
- Ready for current zndraw version (0.5.11)

PR #5251, opened 2026-02-04
ASE Cleanup and MLIP tutorial update

reinaual:
Description of changes:
- added missing update of particle types in ASE interface
- removed type_mapping: replaced with direct access to numbers in ASE (work originally done in #5172)
- re-added the force export to the ASE Interface, necessary for the visualizer
- updated the visualizer to be able to export forces


jngrad (approved):
Thanks a lot!

PR #5257, opened 2026-02-10
Improve code portability

### LINKED ISSUE #5256: [Solved] undefined reference to `__cxa_call_terminate`
**Solutions**

* if you can, use the exact same GCC version to build and link files in your project as the one used to package third-party libraries
   * if third-party libraries were built by different GCC toolchains, use the highest version
* otherwise, if your project links to a third-party library built with GCC 14 but you cannot use GCC 14 in your project, call CMake with `-D CMAKE_EXE_LINKER_FLAGS=-L/usr/lib/gcc/x86_64-linux-gnu/14 -D CMAKE_SHARED_LINKER_FLAGS=-L/usr/lib/gcc/x86_64-linux-gnu/14` to override GCC 13's `libstdc++.so` with the one from GCC 14 that contains the missing symbols
* otherwise, if you must use two GCC versions in a CMake project, e.g. GCC 14 for C++ source files and GCC 13 for CUDA host code, call CMake with `CXX=g++-14 CUDAHOSTCXX=g++-13 CMAKE_CXX_IMPLICIT_LINK_DIRECTORIES_EXCLUDE=/usr/lib/gcc/x86_64-linux-gnu/14 cmake ..` and add `target_link_directories(target BEFORE PUBLIC /usr/lib/gcc/x86_64-linux-gnu/14)` to all targets
* otherwise, if none of these options work for your project, use the Clang toolchain with `CXX=clang++-20 CUDACXX=clang++-20 cmake .. -D CMAKE_CXX_FLAGS="-I/usr/include/x86_64-linux-gnu/c++/15 -I/usr/include/c++/15" -D CMAKE_CUDA_FLAGS="-I/usr/include/x86_64-linux-gnu/c++/15 -I/usr/include/c++/15 --cuda-path=/usr/lib/cuda"`

**Problem statement**

libstdc++ version 14 introduced a new symbol `__cxa_call_terminate` which can be found in several shared libraries built with GCC 14 and up. Linking one of the affected shared libraries to a C++ project built with GCC 13 or lower will result in a linking error. Examples of shared libraries include Boost.mpi and Boost.serialization in Ubuntu 26.04.

CUDA packages are directly impacted by this ABI change, because CUDA &lt; 12.8 only supports GCC versions up to 13 [^1], and therefore GCC 12 or 13 has to be the host compiler, which triggers a linker error: GCC 14 or up has to be used during linking to get the correct `libstdc++.so` file and satisfy the undefined symbol `__cxa_call_terminate`. Likewise, some shared objects may require the additional symbol `CXXABI_1.3.15`, but it is not available in GCC 13 and lower.

While manually linking with GCC 14 or GCC 15 is a valid workaround, we need solution that is natively supported by the build system. CMake projects are in a bind, because when either [`ENV{CUDAHOSTCXX}`](https://cmake.org/cmake/help/v3.31/envvar/CUDAHOSTCXX.html) or [`-D CMAKE_CUDA_HOST_COMPILER`](https://cmake.org/cmake/help/v4.0/variable/CMAKE_CUDA_HOST_COMPILER.html) is used to select GCC 13 as the host compiler (CMake will forward it as a `--compiler-bindir`/`--ccbin` nvcc flag), CMake sets the internal variable `CMAKE_CUDA_HOST_LINK_LAUNCHER` to `/usr/bin/g++-13`, but we need `/usr/bin/g++-14` as the linker. Manually editing that internal variable in `./CMakeFiles/3.31.6/CMakeCUDACompiler.cmake` goes against best practices and may break in future CMake versions. Gentoo Linux modified the CMake source files to provide a new environment variable `ENV{CUDAHOSTLD}` that lets CMake projects set the value of internal variable `CMAKE_CUDA_HOST_LINK_LAUNCHER` ([patch](https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=5a76747ea313441e2eefffcd1badc7ac59e75ae6)), but that won't work in other package managers.

Target property [`LINKER_LANGUAGE`](https://cmake.org/cmake/help/v4.0/prop_tgt/LINKER_LANGUAGE.html) could probably be overridden to `"CXX"` on CUDA targets to select the main compiler, but that wouldn't solve another related issue: CUDA-enabled CMake subprojects may accidentally pick up GCC 13 and transitively forward the path to the GCC 13 `libstdc++.so` file to targets built with GCC 14 in the parent CMake project. In this case, the GCC 13 path takes precedence over the GCC 14 path, and one cannot prepend the GCC 14 path to targets in the parent project since CMake will detect it as redundant and remove it (see [`CMAKE_CXX_IMPLICIT_LINK_DIRECTORIES`](https://cmake.org/cmake/help/v4.0/variable/CMAKE_LANG_IMPLICIT_LINK_DIRECTORIES.html)). The issue here is that we are using two GCC versions, 13 and 14, but CMake can only keep track of one of them in its internal variables.

**Solution for CMake projects**

The proper way to do this is through the combination of:

* [`ENV{CUDAHOSTCXX}`](https://cmake.org/cmake/help/v4.0/envvar/CUDAHOSTCXX.html) (or [`-D CMAKE_CUDA_HOST_COMPILER`](https://cmake.org/cmake/help/v4.0/variable/CMAKE_CUDA_HOST_COMPILER.html)) to select GCC 13 as the host compiler (CMake will forward it as a `--compiler-bindir`/`--ccbin` nvcc flag)
* [`ENV{CXX}`](https://cmake.org/cmake/help/v4.0/envvar/CXX.html) to select GCC 14 as the linker and main C++ compiler
* [`ENV{CMAKE_CXX_IMPLICIT_LINK_DIRECTORIES_EXCLUDE}`](https://cmake.org/cmake/help/v4.0/envvar/CMAKE_LANG_IMPLICIT_LINK_DIRECTORIES_EXCLUDE.html) to remove `/usr/lib/gcc/x86_64-linux-gnu/14` from the list of implicit library directories
* [`target_link_directories()`](https://cmake.org/cmake/help/v4.0/command/target_link_directories.html) to prepend `/usr/lib/gcc/x86_64-linux-gnu/14` to the list of explicit library directories (and thus take precedence over `/usr/lib/gcc/x86_64-linux-gnu/13` from CUDA targets and targets in subprojects)

Minimal working example for Ubuntu 26.04 (it won't work on Ubuntu 24.04!):
```cmake
==> CMakeLists.txt <==
cmake_minimum_required(VERSION 3.27.6)
cmake_policy(VERSION 3.27.6)
project(MyProject VERSION "5.0.0" LANGUAGES CXX CUDA)
find_package(CUDAToolkit 12.0 REQUIRED)
add_library(core SHARED a.cpp a.cu)
# override libstdc++.so search path to reproduce the linker error
target_link_directories(core BEFORE PUBLIC /usr/lib/gcc/x86_64-linux-gnu/13)
# override libstdc++.so search path to fix the linker error (no effect: automatically removed by CMake)
target_link_directories(core BEFORE PUBLIC /usr/lib/gcc/x86_64-linux-gnu/14)
add_executable(main main.cpp)
target_link_libraries(main PRIVATE core)

==> a.cpp <==
int foo();
void err() {
  __cxa_call_terminate(nullptr);
}
int bar() {
  return foo();
}

==> a.cu <==
int foo() {return 1;}

==> main.cpp <==
int bar();
int main(){return bar();}
```
Build commands to reproduce the linker error:
```sh
mkdir build
cd build
CXX=g++-14 CUDAHOSTCXX=g++-13 cmake ..
make main VERBOSE=1
```
Output:
```
-- The CXX compiler identification is GNU 14.3.0
-- The CUDA compiler identification is NVIDIA 12.4.131 with host compiler GNU 13.4.0
-- Check for working CXX compiler: /usr/bin/g++-14 - skipped
-- Check for working CUDA compiler: /usr/bin/nvcc - skipped
-- Found CUDAToolkit: /usr/include (found suitable version "12.4.131", minimum required is "12.0")
-- Configuring done (0.9s)
-- Generating done (0.0s)
[ 20%] Building CXX object CMakeFiles/core.dir/a.cpp.o
/usr/bin/g++-14 -Dcore_EXPORTS  -fPIC -o CMakeFiles/core.dir/a.cpp.o -c /home/user/project/a.cpp
[ 40%] Building CUDA object CMakeFiles/core.dir/a.cu.o
/usr/bin/nvcc -forward-unknown-to-host-compiler -ccbin=/usr/bin/g++-13
  "--generate-code=arch=compute_52,code=[compute_52,sm_52]" -Xcompiler=-fPIC -x cu
  -c /home/user/project/a.cu -o CMakeFiles/core.dir/a.cu.o
[ 60%] Linking CXX shared library libcore.so
/usr/bin/g++-14 -fPIC -Wl,--dependency-file=CMakeFiles/core.dir/link.d -shared
  -Wl,-soname,libcore.so -o libcore.so CMakeFiles/core.dir/a.cpp.o CMakeFiles/core.dir/a.cu.o
  -L/usr/lib/gcc/x86_64-linux-gnu/13
  -L/usr/lib/x86_64-linux-gnu/stubs
  -lcudadevrt -lcudart_static -lrt -lpthread -ldl
[ 80%] Building CXX object CMakeFiles/main.dir/main.cpp.o
/usr/bin/g++-14 -o CMakeFiles/main.dir/main.cpp.o -c /home/user/project/main.cpp
[100%] Linking CXX executable main
/usr/bin/g++-14 -Wl,--dependency-file=CMakeFiles/main.dir/link.d
  CMakeFiles/main.dir/main.cpp.o -o main
  -L/usr/lib/gcc/x86_64-linux-gnu/13
  -Wl,-rpath,/usr/lib/gcc/x86_64-linux-gnu/13:/home/user/project/build
  libcore.so
/usr/bin/x86_64-linux-gnu-ld.bfd: libcore.so: undefined reference to `__cxa_call_terminate'
```
To prevent CMake from removing `/usr/lib/gcc/x86_64-linux-gnu/14` in the last linker command:
```sh
mkdir build
cd build
CXX=g++-14 CUDAHOSTCXX=g++-13 CMAKE_CXX_IMPLICIT_LINK_DIRECTORIES_EXCLUDE=/usr/lib/gcc/x86_64-linux-gnu/14 cmake ..
make main VERBOSE=1
```
Output:
```
...
[100%] Linking CXX executable main
/usr/bin/cmake -E cmake_link_script CMakeFiles/main.dir/link.txt --verbose=1
/usr/bin/g++-14 -Wl,--dependency-file=CMakeFiles/main.dir/link.d
  CMakeFiles/main.dir/main.cpp.o -o main
  -L/usr/lib/gcc/x86_64-linux-gnu/14
  -L/usr/lib/gcc/x86_64-linux-gnu/13
  -Wl,-rpath,/usr/lib/gcc/x86_64-linux-gnu/14:/usr/lib/gcc/x86_64-linux-gnu/13:/home/user/project/build
  libcore.so
[100%] Built target main
```

For reference, here are the problematic symbols in shared libraries packaged on Ubuntu 26.04:
```
$ strings /usr/lib/x86_64-linux-gnu/libboost_mpi.so.1.88.0 | grep -P "CXXABI_|__cxa_call_terminate"
__cxa_call_terminate
CXXABI_1.3
CXXABI_1.3.8
CXXABI_1.3.9
CXXABI_1.3.15
$ strings /lib/gcc/x86_64-linux-gnu/13/libstdc++.so | grep -P "CXXABI_|__cxa_call_terminate"
CXXABI_1.3
CXXABI_1.3.1
...
CXXABI_1.3.14
$ strings /lib/gcc/x86_64-linux-gnu/14/libstdc++.so | grep -P "CXXABI_|__cxa_call_terminate"
__cxa_call_terminate
CXXABI_1.3
CXXABI_1.3.1
...
CXXABI_1.3.14
CXXABI_1.3.15
```

Note to EasyBuild users: for the `foss/2025a` and `foss/2025b` toolchains (GCC 14), use `CUDA/12.8.0` which is compatible [^2].

[^1]: https://docs.nvidia.com/cuda/archive/12.6.3/cuda-installation-guide-linux/index.html#host-compiler-support-policy
[^2]: https://docs.nvidia.com/cuda/archive/12.8.0/cuda-installation-guide-linux/index.html#host-compiler-support-policy

### LINKED ISSUE #5263: Utils::Vector is unsafe
The `Utils::Storage`, `Utils::Array` and `Utils::Vector` rely on implementation-defined behavior in their implicit constructors.

The underlying storage is a C-style array *without a default initializer*. Aggregate types like C-style arrays `T array[N]` have a different behavior depending on the type `T`:
* for class types, the class constructor is called for each element
* for arithmetic types, the values are left undetermined

Since we don't explicitly zero-initialize the array, it is default-initialized by the implicit constructor. We are relying on compilers zero-initializing arithmetic types during default-initialization of C-style arrays. This is often true with `-O0` and `-O1`, but not always true at `-O2`, for example with the NVHPC compiler, values are random.

Quoting from the C++ standard, section 9.4 Initializers **[dcl.init]**:

> 7. To *default-initialize* an object of type `T` means:
>    * (7.1.) If `T` is a (possibly cv-qualified) class type (Clause 11), constructors are considered. The applicable
>      constructors are enumerated (12.2.2.4), and the best one for the *initializer* () is chosen through overload
>      resolution (12.2). The constructor thus selected is called, with an empty argument list, to initialize the object.
>    * (7.2) If `T` is an array type, each element is default-initialized.
>    * (7.3) Otherwise, no initialization is performed.

In practice, this means all statements of the form `return Utils::Vector3d{}` aren't guaranteed to return a zero-initialized vector. With the NVHPC toolchain, they are all in an undeterminate state, which breaks several features, in particular the P3M and DPD features. Even when these features are not used in a simulation, simply enabling them in the `myconfig.hpp` file is sufficient to break other unrelated features. For example, the pressure tensor contains random values when `ELECTROSTATICS` is enabled, because the Coulomb tensor is uninitialized.

We cannot make the `Vector` constructor zero-initialize the underlying storage, because it's a constexpr `__device__` `__host__` function, and therefore providing a non-default constructor would prevent it from being called by another constexpr `__device__` `__host__` function like `Utils::Vector::broadcast()`.

jngrad:
Description of changes:
- address compiler diagnostics for GCC 13 to 16, Clang 19 to 21, nvcc 12.0 to 12.8, Xcode, CrayClang, IntelLLVM
- add support for NVHPC compiler toolchain
- skip compilation of deprecated Kokkos and waLBerla features
- bump all dependencies to get full C++20 support
- modernize user guide with Sphinx tabs to wrap build instructions for different toolchains
   - closes #5256
- remove multiple sources of undefined behavior
   - fixes #5263
- redesign runtime error system to use Python's `stderr` instead of the lower-level `std::cerr` (which cannot be redirected to a buffer/filehandle/stream at the Python level)


src/core/BoxGeometry.hpp
```diff
@@ -293,8 +292,8 @@ class BoxGeometry {
       auto const shear_plane_normal = le.shear_plane_normal;
       auto const shear_direction = le.shear_direction;
       auto const dy = x[shear_plane_normal] - y[shear_plane_normal];
-      if (fabs(dy) > length_half()[shear_plane_normal]) {
-        ret[shear_direction] -= Utils::sgn(dy) * le.shear_velocity;
+      if (std::fabs(dy) > length_half()[shear_plane_normal]) {
+        ret[shear_direction] -= std::copysign(le.shear_velocity, dy);
```
jngrad, line 296: This [signum function](https://en.wikipedia.org/wiki/Sign_function) call was removed in favor of [`std::copysign`](https://en.cppreference.com/w/cpp/numeric/math/copysign), which doesn't have an edge case at 0. As I understand it, the case x=0 cannot occur since it would require the half box length to be negative, something that cannot be achieved from the Python interface (and would lead to fatal errors in the cell structure anyway). LEbc was the last user of signum; it was progressively phased out over the years, due to it being frequently misused, and this PR completely removes it. Notifying @Yashas17 just in case I'm missing something.

src/core/cluster_analysis/ClusterStructure.cpp
```diff
@@ -134,68 +127,57 @@ void ClusterStructure::add_pair(const Particle &p1, const Particle &p2) {
...
 
   // Now fill the cluster objects with particle ids
   // Iterate over particles, fill in the cluster map
   // to each cluster particle the corresponding cluster id
-  for (auto it : cluster_id) {
-    // If this is the first particle in this cluster, instance a new cluster
-    // object
-    if (clusters.find(it.second) == clusters.end()) {
-      clusters[it.second] = std::make_shared<Cluster>(m_box_geo);
-    }
```
jngrad, line 171: As I understand it, the body of this conditional was never executed because the merged clusters are already instantiated at line 144. The refactored version of this function now uses C++20's `std::map<T,U>::contains()` to make the intention clear and the code duplication more visible. Notifying @RudolfWeeber just in case I'm missing something.

jngrad:
Confirmed to work on LUMI, MareNostrum, macOS/arm64 (also in debug builds: [job 63522833231](https://github.com/jngrad/espresso/actions/runs/21986756878/job/63522833231)), pre-release Ubuntu 26.04/x86, Fedora Rawhide ([task 142230523](https://koji.fedoraproject.org/koji/taskinfo?taskID=142230523)), and EESSI-dev (https://github.com/multixscale/dev.eessi.io-espresso/pull/6).

CMakeLists.txt
```diff
@@ -633,14 +640,14 @@ if(ESPRESSO_BUILD_WITH_SHARED_MEMORY_PARALLELISM)
   find_package(OpenMP REQUIRED COMPONENTS CXX)
 
   if(NOT EXISTS ${FETCHCONTENT_BASE_DIR}/kokkos-src)
-    find_package(Kokkos 4.3 QUIET)
+    find_package(Kokkos 4.6 QUIET)
```
itischler, line 652: should this not be 5.0 or do we support both versions?

src/core/electrostatics/elc.cpp
```diff
@@ -1129,7 +1130,7 @@ void charge_assign(elc_data const &elc, CoulombP3M &solver,
 
 #ifdef ESPRESSO_SHARED_MEMORY_PARALLELISM
   // multi-threading -> cache sizes must be equal to the number of particles
-  auto const include_neutral_particles = Kokkos::num_threads() > 1;
+  auto constexpr include_neutral_particles = true;
```
itischler, line 1133: why was the thread number being greater than 1 a requirement for this and does it still work, when SMP is enabled but only run one one core?

reinaual (commented):
Besides the `Kokkos`-version in CMake, it looks good to me.

CMakeLists.txt
jngrad: I would like to support both, so that we can use package [`kokkos-devel`](https://packages.fedoraproject.org/pkgs/kokkos/kokkos-devel/) when packaging ESPResSo on Fedora. We are only using stable features from Kokkos 4.6 that are still present in Kokkos 5.0.

src/core/electrostatics/elc.cpp
jngrad: I'm not sure why this bug only surfaced on macOS. All architectures should be affected. I've now introduced a neutral particle in the ELC test and can confirm this bug now gets detected right away on x86 CPUs.

reinaual (approved):

itischler (approved):

PR #5258, opened 2026-02-12
fix ek friction forces

itischler:
The ek lb friction force is now calculated by all ek species and not just from the last one, which previously had overwritten the forces of the other species


reinaual:
Good catch, this is a serious oversight on my end! Could you please add a testcase for this? Maybe the EOF testcase for multiple species or a new one, where species are pushing momentum into the fluid via an external field?
And did you check that LB sees the correct forces now?

testsuite/python/ek_fluid_coupling.py
```diff
@@ -0,0 +1,232 @@
...
+            expected_force = np.full_like(forces, expected_force)
+            np.testing.assert_allclose(
+                forces,
+                expected_force,
+                rtol=1e-5)
+        else:
+            np.testing.assert_allclose(
+                forces,
+                np.zeros_like(forces),
+                atol=1e-10)
```
jngrad:
```suggestion
            np.testing.assert_allclose(
                forces,  expected_force, rtol=1e-5, atol=1e-9)
        else:
            np.testing.assert_allclose(
                forces, np.zeros_like(forces), atol=1e-9)
```

jngrad (approved):

PR #5259, opened 2026-02-13
LBWalberlaImpl readability

RudolfWeeber:
The LBWalberlaImpl tempalte has gotten too long to be understandable by new people. This is WIP to improve on that by moving out some details to separate files and re-order the main class template.


RudolfWeeber:
I think, we should keep the section headings in the LBWalberlaImpl template. The last commit seems to remove them.

jngrad (approved):

PR #5279, opened 2026-03-27
Fix checkpointing of dihedral angles

### LINKED ISSUE #5277: Checkpoint loading fails for simulations that include Dihedral interactions.
Checkpoint loading fails for simulations that include Dihedral interactions in ESPResSo 5.0.

The issue appears to be that the dihedral parameter mult is restored as a float during checkpoint loading, while the interaction expects a positive integer.

Error message
mult must be a positive integer -- got a float
Observed behavior
Simulations run normally when started fresh
Checkpoints can be written
Loading a checkpoint fails if dihedral interactions are present
The checkpoint is rejected and the simulation restarts from scratch
Expected behavior
Checkpoint loading should restore dihedral interactions correctly, preserving mult as an integer.

I suppose it would be enough to patch ~/espresso/src/core/bonded_interactions/dihedral.hpp :
struct DihedralBond {
  int mult;
  double bend;
  double phase;
.
.
.
};

jngrad:
Fixes #5277

Description of changes:
- fix dihedral bond checkpointing bug (regression introduced in 5.0.0)
- re-enable virtual sites relative checkpointing test (regression introduced in 5.0.0)
- improve testing of bonds, particle properties, virtual sites

### DIFF
--- a/src/core/bonded_interactions/dihedral.hpp
+++ b/src/core/bonded_interactions/dihedral.hpp
@@ -45,7 +45,7 @@ inline constexpr auto dihe_tiny_sin_value{1e-10};
 
 /** Parameters for four-body angular potential (dihedral-angle potentials). */
 struct DihedralBond {
-  double mult;
+  int mult;
   double bend;
   double phase;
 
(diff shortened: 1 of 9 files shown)


reinaual (approved):

jngrad:
Credit goes to @NiklasMueller2001 for identifying the code coverage regression in the checkpointing test.

PR #5283, opened 2026-04-02
Remove performance bottleneck in GPU VTK writers

itischler:
Description of changes:
- The gpu lb vtk now uses the correct field to write the vtk 
- The pressure is written to an array before the vtk write process to utilize parallelization during the calculation


src/walberla_bridge/src/electrokinetics/EKinWalberlaImpl.hpp
```diff
@@ -1114,149 +1094,128 @@ class EKinWalberlaImpl : public EKinWalberlaBase {
...
+    Vector3<uint_t> m_dims;
 
-#if defined(__CUDACC__) and defined(WALBERLA_BUILD_WITH_CUDA)
-  template <typename OutputType = float>
-  class DensityVTKWriter : public VTKWriter<DensityFieldCpu, 1u, OutputType> {
   public:
-    using Base = VTKWriter<DensityFieldCpu, 1u, OutputType>;
-    using Base::Base;
-    using Base::evaluate;
+    void set_content(VecType content) { m_content = content; }
```
jngrad: Here and in LB, I think a `std::move()` is needed to enable copy elision of pass-by-value arguments.

MWE:
```c++
#include <cstdio>
#include <utility>
struct S {
    S() { std::printf("S() at %p\n", (void*)this); }
    ~S() { std::printf("~S() at %p\n", (void*)this); }
    S(const S&) { std::puts("S(const S&)"); }
    S(S&&other) noexcept { std::printf("S(S&&) at %p from %p\n", (void*)this, (void*)&other); }
    S& operator=(const S&other) { std::printf("operator=(const S&) at %p from %p\n", (void*)this, (void*)&other); return *this; }
    S& operator=(S&& other) noexcept { std::printf("operator=(S&&) at %p from %p\n", (void*)this, (void*)&other); return *this; }
};
struct P {
    S member;
    P(S value) : member{std::move(value)} {}
};
auto foo() {
    return P(S{});
}
int main() {
    auto p = foo();
}
```

jngrad (approved):

PR #5285, opened 2026-04-08
short-range force loop optimisations

RudolfWeeber:
This has a bunch of optimizations for force calculation and others.

* Reduce impact of compiled-in but nto used features
* reduce Vector3d{} instanciation, sqrt and data loads in hot loops
* allow inlining for_each_particle() calls
* optimizations to langevin per particle
* On my machine, brings some 10% on the lj benchmark for the default myconfig

## changes

nonbonded_interaction_data.hpp/.cpp
- Add PairPotential enum and pair_potential_bit() helper
- Add active_pair_mask bitmask field to IA_parameters, populated by
  recalc_maximal_cutoffs() via new consider() helper; tracks which
  potentials are actually configured for each type pair

forces_inline.hpp, energy_inline.hpp:
- Guard every pair-potential branch with the active_pair_mask bit so
  inactive potentials cost nothing even when compiled in

VerletCriterion.hpp, short_range_cabana.hpp, CellStructure.hpp,
VerletCriterion_test.cpp
- Pass squared distance (double) directly to VerletCriterion instead
  of a Distance struct; avoids sqrt on the Verlet-list rebuild path
- VerletCriterion::operator() now takes double dist2

BoxGeometry.hpp
- Add get_mi_dist2() that accumulates the three squared components
  directly for cuboid boxes, avoiding an intermediate Vector3d

forces_cabana.hpp, energy_cabana.hpp, short_range_cabana_helpers.hpp
- Store system_max_cutoff squared; compare dist_sq for early exit
  before taking the sqrt
- Load particle positions/directors lazily, only where algorithms
  that need them are actually active (Gay-Berne, dipoles, DPD, ELC,
  NPT Coulomb virial)
- Add dpd_active() predicate checking both the thermostat switch and
  the active_pair_mask DPD bit

CellStructure.hpp/.cpp, for_each_particle.hpp (new)
- Replace std::function / ParticleUnaryOp with a template Callable
  in for_each_local_particle(), for_each_ghost_particle() and
  parallel_for_each_particle_impl() to eliminate virtual dispatch
- Move parallel_for_each_particle_impl template body into new header
  for_each_particle.hpp; add include in forces.cpp, integrate.cpp,
  icc.cpp, virtual_sites/com.cpp, virtual_sites/relative.cpp

langevin_inline.hpp
- For isotropic particles (all friction components equal) bypass the
  rotation-matrix operations entirely; use hadamard_product() directly
  for both the friction and noise terms


src/core/short_range_cabana_helpers.hpp
```diff
@@ -43,23 +61,15 @@ KOKKOS_INLINE_FUNCTION bool thole_active(IA_parameters const &ia_params,
 #endif
 
 struct PairDataFlags {
-  bool need_directors = false;
   bool need_particle_pointers = false;
```
hidekb: If only `need_particle_pointers` is required, the `PairDataFlags` struct may be unnecessary. Instead, `compute_pair_data_flags` simply returns a bool indicating `need_particle_pointers`.

hidekb (dismissed):
Looks good to me.

jngrad (approved):
as per @hidekb's review

PR #5297, opened 2026-04-20
Rewrite features parser and config file generator

jngrad:
Process feature implication rules in topological order to guarantee that transitive chains are always evaluated. Add test case to check the features parser. Remove unused features checker.


itischler (approved):
looks good to me

PR #5300, opened 2026-04-22
DPD: avoid std::pow() for standard ramps

RudolfWeeber:


Yashas17 (commented):
Does this really have any performance impact? We are replacing `std::pow()` with `if` statements, which themselves will cost some time.  In the case of a non-standard ramp, it will even degrade the performance.

RudolfWeeber:
I think so. The std::pwoer() was visible in the profile with a few percent, and the branch prediciton should get the ifs right, as there is typically only one dpd ramp type in a simulation.

Yashas17:
The code logic looks correct to me. So as long as it has performance gains for the majority of cases, it should be fine, I suppose.

Yashas17 (approved):

jngrad (dismissed):
8% speed-up for linear and quadratic ramps, using 500 particles and 1 core.

jngrad (approved):

PR #5301, opened 2026-04-22
LE: avoid one std::round() call in distance calculation

RudolfWeeber:


src/core/lees_edwards/LeesEdwardsBC.hpp
```diff
@@ -41,11 +41,10 @@ struct LeesEdwardsBC {
...
     auto const le_plane_normal = static_cast<unsigned int>(shear_plane_normal);
     auto const le_direction = static_cast<unsigned int>(shear_direction);
-    auto const n_le_crossings =
-        std::round(res[le_plane_normal] * l_inv[le_plane_normal]);
-    if (n_le_crossings >= 1.)
+    auto const le_norm_frac = res[le_plane_normal] * l_inv[le_plane_normal];
+    if (le_norm_frac >= 0.5)
       res[le_direction] -= pos_offset;
-    if (n_le_crossings <= -1.)
+    if (le_norm_frac <= -0.5)
```
Yashas17: Why not use an `else if` here? It will ensure only 1 `if` statement gets executed for `le_norm_frac>=0.5`.

Yashas17 (commented):
The new logic replacing `std::round()` looks correct to me.

RudolfWeeber:
It doesn't make a difference, I think. as there are 3 options, we'd need
```
if (a) {}
else if (b) {}
else {}
```
which would still evaluate two ifs.

A more general solution would be replacing std::round() to std::rint(), which however changes the tie-breaking rule for 0.5, which would need further analysis. So I think, this pr is what we can do at the moment - if you agree it doesn't change the current behavior.

Yashas17:
We will just need

```
if (le_norm_frac >= 0.5){
      res[le_direction] -= pos_offset;}
else if (le_norm_frac <= -0.5) {
      res[le_direction] += pos_offset;}
```

Why is there a third `else` statement here? In case the first if is evaluated, we will not evaluate the second `if`, thus providing some performance improvement. Right now, we are evaluating both `if` statements, even if the first one is true.

RudolfWeeber:
Ah, got it. If the one if triggers, we can skip the 2nd.

Yashas17 (approved):

jngrad (approved):

PR #5302, opened 2026-04-22
Improve NVHPC toolchain support

jngrad:
Fix double-free in prvalue temporary objects passed as const lvalue reference to functions arguments (affects nvc++ <= 26.3). Select host code compiler flags based on the host compiler id. Address NVHPC diagnostics.


itischler (approved):
looks good. if the failed CI jobs get fixed it is fine

jngrad:
The CI runner interrupted both docker containers for some reason. I re-ran the interrupted jobs 4 times each just to be safe, and they all passed.

PR #5304, opened 2026-04-23
Fix EK testcase and add unit conversions

reinaual:
Description of changes:
- API change: EK electrostatics solver now needs a time step `tau` for proper unit conversion
- API change: `EKNone` is now available on GPU
- bugfix: EK unit conversion fixed for `tau != 1`


testsuite/python/ek_potential_periodic.py
```diff
@@ -0,0 +1,144 @@
...
+        I, J, K = np.meshgrid(i, j, k, indexing="ij")
+        phase = 2.0 * np.pi * (nx * I / Nx + ny * J / Ny + nz * K / Nz)
+
+        phi_exact = rho0 * np.cos(phase)
+
+        lap = (
+            (np.roll(phi_exact, -1, axis=0) - 2.0 * phi_exact + np.roll(phi_exact, 1, axis=0)) / hx**2 +
+            (np.roll(phi_exact, -1, axis=1) - 2.0 * phi_exact + np.roll(phi_exact, 1, axis=1)) / hy**2 +
+            (np.roll(phi_exact, -1, axis=2) - 2.0 *
+             phi_exact + np.roll(phi_exact, 1, axis=2)) / hz**2
```
itischler: the formatting is a bit off here

src/script_interface/walberla/EKFFT.hpp
```diff
@@ -59,7 +60,10 @@ class EKFFT : public EKPoissonSolver {
   void make_instance(VariantMap const &args) override {
     // unit conversions
     auto const agrid = get_value<double>(m_lattice->get_parameter("agrid"));
-    m_conv_permittivity = Utils::int_pow<2>(agrid);
+    auto const tau = get_value<double>(args, "tau");
```
itischler, line 62: this change of the interface should also be mentioned in the electrokinetics.py so that the docu is correct. also i still find it kind of odd that a electrostatics solver needs a timestep, but that is how it is now.

src/walberla_bridge/src/electrokinetics/PoissonSolverNone.hpp
```diff
@@ -19,59 +19,173 @@
...
-                      Utils::Vector3i const &) const override {
-    throw std::runtime_error("PoissonSolverNone has no potential field");
+  std::size_t get_potential_field_id() const noexcept override {
+    return static_cast<std::size_t>(m_potential_field_id);
   }
 
-  void solve() override {}
-  void setup_fft(bool) override {}
+  [[nodiscard]] std::optional<double>
+  get_node_potential(Utils::Vector3i const &node,
```
itischler, line 137: shouldn't the potential getter for the SolverNone simply throw an error or maybe return 0.

src/walberla_bridge/src/electrokinetics/PoissonSolverNone.hpp
```diff
@@ -19,59 +19,173 @@
...
+      return std::nullopt;
+
+    auto const potential_field =
+        bc->block->template getData<PotentialField>(m_potential_field_id);
+    return {double_c(
+        walberla::ek::accessor::Scalar::get(potential_field, bc->cell))};
   }
+
+  [[nodiscard]] std::vector<double>
+  get_slice_potential(Utils::Vector3i const &lower_corner,
```
itischler, line 164: same as above

itischler (changes requested):

testsuite/python/ek_potential_periodic.py
reinaual: Well, I guess for that you have to complain to our CI. I didn't manage to convince it...

src/walberla_bridge/src/electrokinetics/PoissonSolverNone.hpp
reinaual: The reasons I added these was that we allocate zero-initialized fields, and that way one is actually showing the user what is going on, not just showing what we expect is going on. I've added the possibility to manipulate the fields in the EKNone solver now, since we allocate the memory anyway. That way the user can set the potential landscape from the interface, if there's ever a need for it.

itischler (dismissed):
seems good now, thanks

jngrad (approved):
as per @itischler's review

PR #5306, opened 2026-04-27
Boundaries includable in lb and ek vtk

itischler:
Description of changes:
- The vtk data inside of boundaries cells can be written by toggeling the flag include_boundaries in the vtk_wrtier class.
- Boundaries can also be tracked as an observable by the vtk writers


testsuite/python/lattice_vtk.py
```diff
@@ -352,6 +415,20 @@ def test_vtk(self):
             self.assertEqual(sorted(vtk_obj.observables), sorted(vtk_obs))
             self.assertEqual(vtk_obj.valid_observables(), set(self.valid_obs))
 
+            if self.include_boundaries:
+                # also write a snapshot that includes boundary cells and the
+                # ``boundary`` mask observable
+                label_vtk_with_boundaries = f"test_vtk_{
+                    self.vtk_id}_with_boundaries"
```
jngrad: Here and elsewhere: there is a Python version we support, but don't have in CI, where this syntax is invalid. Our autopep8 version has a bug and introduces line breaks in f-strings. The bug report came after the autopep8 version was frozen in Ubuntu. We need `  # nopep8` (with two leading whitespaces) to prevent carriage returns.

jngrad (approved):
Thanks!

PR #5310, opened 2026-04-30
Exchanging from hand-made duplication View to ScatterView

### LINKED ISSUE #5284: Use Kokkos ScatterView for forces/torques
## Background

* when calculating short rnage forces in parallel, several threads can add forces to the same particle simultaneously, hence, we have to avoid rae conditions
* There are two options:
    1. duplicaiton: each thread writes to forces to a private view/array. After force calculation, the sum of the contributions from all threads is summed for each particle
    2. atomics: all threads write to a single view/array using atomics.
* Which one is better depeds on the target. For GPU, atomics are often fine, on the CPU, they are quite slow.
* Right now, we use duplication in the force calculation. This is implemented "by hand" in ESPREsSo.
* Kokkos ScatterView abstracts these details away. One can switch between duplication and atomics via a template parameter.

## Goal

* There should be central views and corresponding scatter views for particle forces and torques valid throughout the integraiton loop
* short range force calculation (and everything else that contributes foresces/torques) in paralllel via Kokkos should use the access() method of the scatter view
* The contribute() call can probably be placed just once, before the fores are applied to the particles

## Notes

* There are other places, where scatter view should probably be used, e.g., P3M charge assignment.
* A simple benchmark for a histogram is here: https://github.com/RudolfWeeber/kokkos_scatterview_bench
* On my machine, CPU is significantly faster with duplication and GPU with atomics.

hidekb:
Fixes #5284

Description of changes:
- The variables local_force, local_torque and local_virial implemented by Kokkos::View is exchanged to ones implemented by Kokkos::ScatterView

### DIFF
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -76,8 +76,8 @@ install(TARGETS espresso_core
 
 target_link_libraries(
   espresso_core
-  PUBLIC  Cabana::Core Kokkos::kokkos OpenMP::OpenMP_CXX
-          $<$<BOOL:${ESPRESSO_BUILD_WITH_CUDA}>:OpenMP::OpenMP_CUDA>
+  PUBLIC Cabana::Core Kokkos::kokkos OpenMP::OpenMP_CXX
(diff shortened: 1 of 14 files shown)


jngrad (approved):
Great work!

PR #5320, opened 2026-05-18
Fix regressions

jngrad:
Description of changes:
- make small CMake adjustments for the NVHPC compiler toolchain
- reduce anti-patterns in the C++ core and handle an edge case in the immersed boundaries method
- add missing feature guards and raise an exception when setting a thermostat that is not compiled in
- relax tolerances of the Grand canonical tutorial test case to reduce the frequency of CI failures

AI disclosure: some of the C++ corrections were suggested by Claude Sonnet based on a local copy of the ISO C++ guidelines. These suggestions were manually verified for correctness.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -333,7 +333,6 @@ enable=dangerous-default-value, # W0102
        unused-argument, # W0613
        unused-wildcard-import, # W0614
        deprecated-method, # W1505
-       cyclic-import, # R0401
        trailing-comma-tuple, # R1707
        bad-classmethod-argument, # C0202
        undefined-variable, # E0602
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -644,7 +644,9 @@ endif()
 #
 
 set(ESPRESSO_OPENMP_COMPONENTS CXX)
-if(ESPRESSO_BUILD_WITH_CUDA)
+if(ESPRESSO_BUILD_WITH_CUDA AND NOT CMAKE_CUDA_COMPILER_ID STREQUAL NVHPC)
+  # FindOpenMP doesn't fully support NVHPC+CUDA as of CMake 4.3, see ticket:
+  # https://gitlab.kitware.com/cmake/cmake/-/work_items/23003
   list(APPEND ESPRESSO_OPENMP_COMPONENTS CUDA)
 endif()
 # ESPResSo and Kokkos require some of the new features from OpenMP 5.0, but a
--- a/src/core/immersed_boundary/ibm_tribend.cpp
+++ b/src/core/immersed_boundary/ibm_tribend.cpp
@@ -54,7 +54,7 @@ IBMTribend::calc_forces(BoxGeometry const &box_geo, Utils::Vector3d const &pos1,
   n2 /= Aj;
 
   // Get the prefactor for the force term
-  auto const sc = std::min(1.0, n1 * n2);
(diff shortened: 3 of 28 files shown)


.pylintrc
```diff
@@ -333,7 +333,6 @@ enable=dangerous-default-value, # W0102
        unused-argument, # W0613
        unused-wildcard-import, # W0614
        deprecated-method, # W1505
-       cyclic-import, # R0401
```
jngrad, line 336: Removing this check because a bug in pylint causes false positive to be detected inside comment lines of the license headers. Happens at random in CI ([example](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/418023)). Reported by @NiklasMueller2001.

CMakeLists.txt
```diff
@@ -644,7 +644,9 @@ endif()
 #
 
 set(ESPRESSO_OPENMP_COMPONENTS CXX)
-if(ESPRESSO_BUILD_WITH_CUDA)
+if(ESPRESSO_BUILD_WITH_CUDA AND NOT CMAKE_CUDA_COMPILER_ID STREQUAL NVHPC)
+  # FindOpenMP doesn't fully support NVHPC+CUDA as of CMake 4.3, see ticket:
+  # https://gitlab.kitware.com/cmake/cmake/-/work_items/23003
```
jngrad, line 649: This change doesn't prevent building ESPResSo on MareNostrum, when the NVHPC toolchain is packaged to use nvcc as the backend to build CUDA device code. For systems where the NVHPC toolchain uses `nvc++ -cuda` as the backend to build CUDA device code, the situation is more complicated, and we need to exclude the CUDA component since FindOpenMP doesn't fully support this toolchain, at the time of writing.

src/core/immersed_boundary/ibm_tribend.cpp
```diff
@@ -54,7 +54,7 @@ IBMTribend::calc_forces(BoxGeometry const &box_geo, Utils::Vector3d const &pos1,
   n2 /= Aj;
 
   // Get the prefactor for the force term
-  auto const sc = std::min(1.0, n1 * n2);
+  auto const sc = std::clamp(n1 * n2, -1., 1.);
```
reinaual, line 57: Are these numerical accidents really something one sees? A scalar product of two normalized vectors to exceed the range [-1, 1]?

jngrad:
```c++
#include <cmath>
#include <cstdio>
#include <valarray>

int main() {
    std::valarray<double> v1{-1., -2., -1.};
    std::valarray<double> v2{+1., +2., +1.};
    auto const v1_sqr = v1 * v1;
    auto const v2_sqr = v2 * v2;
    auto const v1_norm = std::sqrt(v1_sqr[0] + v1_sqr[1] + v1_sqr[2]);
    auto const v2_norm = std::sqrt(v2_sqr[0] + v2_sqr[1] + v2_sqr[2]);
    v1 /= v1_norm;
    v2 /= v2_norm;
    auto const product = v1 * v2;
    auto const scalar = product[0] + product[1] + product[2];
    printf("scalar=%.16f\n", scalar);
    return 0;
}
```
Output on x86:
```
$ g++-14 mwe.cpp
$ ./a.out 
scalar=-1.0000000000000002
```

reinaual (approved):

PR #5322, opened 2026-05-26
Dihedral pressure tensor

### LINKED ISSUE #5311: Presure contrib for dihedrals is missing
As far as I can tell, we do not calculate pressure tensor contributions for dihedrals..

hidekb:
Fixes #5311

Description of changes:
- Added calculation of dihedral pressure tensor
- Added the test for calculating the dihedral pressure tensor


jngrad (approved):
The implementation matches perfectly with GROMACS:

Plots showing 4 components of the pressure tensor for a single dihedral bond over the angle range 0 to 2*pi, calculated in ESPResSo and GROMACS, with excellent agreement

PR #5324, opened 2026-05-26
Fix fits_in_box check for Interpolation-Based Constraints

### LINKED ISSUE #5323: Interpolation-Based Constraints can cause Segfaults
Passing the constraint a wrong grid_spacing can cause it's `fits_in_box` check to pass, but the interpolation to access out of bounds.

```python
import numpy as np
import espressomd
import espressomd.constraints

BOX_L = 20.0
system = espressomd.System(box_l=[BOX_L, BOX_L, BOX_L])
system.time_step = 0.01
system.cell_system.skin = 0.4

system.part.add(pos=[BOX_L * 0.75, BOX_L / 2, BOX_L / 2])

xy_points = 3
# wrong xy_spacing
xy_spacing = BOX_L / (xy_points - 1)

z_spacing = 2.0
z_points = int(np.ceil(BOX_L / z_spacing)) + 2

field = np.zeros((xy_points, xy_points, z_points, 3))
constraint = espressomd.constraints.ForceField(
    field=field,
    grid_spacing=[xy_spacing, xy_spacing, z_spacing],
    default_scale=1.0,
)
system.constraints.add(constraint)

system.integrator.run(1)
```
The reason for this is the implicit shifting of the array positions by half a grid spacing, effectively reducing the usable size for the interpolation by one grid spacing. We warn about this in the docstring, but never check for it.
https://github.com/espressomd/espresso/blob/6c413d81536837c53046eb348f8406bfb03745fa/src/python/espressomd/constraints.py#L224-L230)

reinaual:
Fixes #5323

@jngrad I think the error for the user is not very helpful. Would make a good beginners Coding-Day Ticket to improve the error message in the ScriptInterface.

### DIFF
--- a/src/core/field_coupling/fields/Interpolated.hpp
+++ b/src/core/field_coupling/fields/Interpolated.hpp
@@ -136,7 +136,10 @@ public:
 
   bool fits_in_box(const Utils::Vector3d &box) const {
     auto const box_shape = shape();
-    auto const grid_size = Utils::hadamard_product(m_grid_spacing, box_shape);
+    // Due to origin-shifting, we require an additional grid point
+    auto const usable_shape = box_shape - Utils::Vector3i::broadcast(1);
+    auto const grid_size =
+        Utils::hadamard_product(m_grid_spacing, usable_shape);
     return (m_origin < Utils::Vector3d::broadcast(0.)) &&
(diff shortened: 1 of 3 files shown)


jngrad (approved):
Thanks!

PR #5325, opened 2026-05-27
added slice setter and getter for indexed reactions

itischler:
Description of changes:
- added slice setter and getter for indexed reactions


jngrad (approved):
Thanks!

PR #5327, opened 2026-05-27
Fix EKin slice/node getters and VTK output

jngrad:
Description of changes:
- EK node and slice getters now return the imposed value when a cell has a boundary condition
- EK GPU and EK CPU objects can no longer be mixed in the same EK container
- EKNone now writes the potential in VTK files
- EKNone GPU potential node/slice getters no longer return uninitialized data before the first time step

### DIFF
--- a/src/script_interface/walberla/EKFFT.hpp
+++ b/src/script_interface/walberla/EKFFT.hpp
@@ -50,7 +50,6 @@ namespace ScriptInterface::walberla {
 class EKFFT : public EKPoissonSolver {
 protected:
   std::unique_ptr<ResourceManager> m_resources_lock;
-  std::shared_ptr<LatticeWalberla> m_lattice;
   double m_conv_permittivity;
   bool m_gpu;
   bool m_single_precision;
@@ -97,7 +96,6 @@ public:
     m_lattice = get_value<decltype(m_lattice)>(args, "lattice");
     m_vtk_writers =
         get_value_or<decltype(m_vtk_writers)>(args, "vtk_writers", {});
-
     make_instance(args), m_resources_lock = std::make_unique<ResourceManager>();
     // MPI communicator is needed to destroy the FFT plans
     m_resources_lock->acquire_lock(::communication_environment->get_mpi_env());
@@ -123,6 +121,8 @@ public:
         {"lattice", AutoParameter::read_only, [this]() { return m_lattice; }},
         {"shape", AutoParameter::read_only,
          [this]() { return m_instance->get_lattice().get_grid_dimensions(); }},
+        {"vtk_writers", AutoParameter::read_only,
+         [this]() { return serialize_vtk_writers(); }},
     });
   }
 
--- a/src/script_interface/walberla/EKIndexedReactionSlice.cpp
+++ b/src/script_interface/walberla/EKIndexedReactionSlice.cpp
@@ -47,8 +47,8 @@ Variant EKIndexedReactionSlice::do_call_method(std::string const &name,
     auto const attr = get_value<std::string>(params, "name");
     if (not m_shape_val.contains(attr)) {
       context()->parallel_try_catch([&]() {
-        throw std::runtime_error("Unknown EK indexed reaction property '" +
-                                 attr + "'");
+        std::string const class_prop{"EK indexed reaction property"};
+        throw std::runtime_error("Unknown " + class_prop + " '" + attr + "'");
       });
     }
     return m_shape_val.at(attr);
--- a/src/script_interface/walberla/EKNone.hpp
+++ b/src/script_interface/walberla/EKNone.hpp
@@ -48,8 +48,6 @@
 namespace ScriptInterface::walberla {
 
 class EKNone : public EKPoissonSolver {
-  std::shared_ptr<::walberla::PoissonSolver> m_instance;
-  std::shared_ptr<LatticeWalberla> m_lattice;
   bool m_gpu;
   bool m_single_precision;
 
@@ -90,11 +88,18 @@ public:
     m_gpu = get_value_or<bool>(args, "gpu", false);
     m_single_precision = get_value_or<bool>(args, "single_precision", m_gpu);
     m_lattice = get_value<decltype(m_lattice)>(args, "lattice");
+    m_vtk_writers =
+        get_value_or<decltype(m_vtk_writers)>(args, "vtk_writers", {});
     auto const agrid = get_value<double>(m_lattice->get_parameter("agrid"));
     m_tau = get_value<double>(args, "tau");
     set_potential_conversion(agrid, m_tau);
-
     make_instance(args);
+    for (auto &vtk : m_vtk_writers) {
+      vtk->attach_to_lattice(m_instance, get_lattice_to_md_units_conversion());
+    }
+  }
+
+  EKNone() {
     add_parameters({
         {"tau", AutoParameter::read_only, [this]() { return m_tau; }},
         {"single_precision", AutoParameter::read_only,
@@ -103,6 +108,8 @@ public:
         {"lattice", AutoParameter::read_only, [this]() { return m_lattice; }},
         {"shape", AutoParameter::read_only,
          [this]() { return m_instance->get_lattice().get_grid_dimensions(); }},
+        {"vtk_writers", AutoParameter::read_only,
+         [this]() { return serialize_vtk_writers(); }},
     });
   }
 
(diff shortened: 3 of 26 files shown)


src/script_interface/walberla/EKFFT.hpp
```diff
@@ -50,7 +50,6 @@ namespace ScriptInterface::walberla {
 class EKFFT : public EKPoissonSolver {
 protected:
   std::unique_ptr<ResourceManager> m_resources_lock;
-  std::shared_ptr<LatticeWalberla> m_lattice;
```
jngrad, line 53: The same shared pointer exists in a parent class with the same name. Thus the parent class' pointer was shadowed and never initialized, leading to confusing error messages.

src/script_interface/walberla/EKFFT.hpp
```diff
@@ -123,6 +121,8 @@ class EKFFT : public EKPoissonSolver {
         {"lattice", AutoParameter::read_only, [this]() { return m_lattice; }},
         {"shape", AutoParameter::read_only,
          [this]() { return m_instance->get_lattice().get_grid_dimensions(); }},
+        {"vtk_writers", AutoParameter::read_only,
+         [this]() { return serialize_vtk_writers(); }},
```
jngrad, line 125: needed for checkpointing

src/script_interface/walberla/EKIndexedReactionSlice.cpp
```diff
@@ -47,8 +47,8 @@ Variant EKIndexedReactionSlice::do_call_method(std::string const &name,
     auto const attr = get_value<std::string>(params, "name");
     if (not m_shape_val.contains(attr)) {
       context()->parallel_try_catch([&]() {
-        throw std::runtime_error("Unknown EK indexed reaction property '" +
-                                 attr + "'");
+        std::string const class_prop{"EK indexed reaction property"};
+        throw std::runtime_error("Unknown " + class_prop + " '" + attr + "'");
```
jngrad, line 51: split over 2 lines to get proper code coverage

src/script_interface/walberla/EKNone.hpp
```diff
@@ -48,8 +48,6 @@
 namespace ScriptInterface::walberla {
 
 class EKNone : public EKPoissonSolver {
-  std::shared_ptr<::walberla::PoissonSolver> m_instance;
-  std::shared_ptr<LatticeWalberla> m_lattice;
```
jngrad, line 52: these pointer are already present in a parent class with the same name

src/walberla_bridge/src/electrokinetics/PoissonSolverNone.hpp
```diff
@@ -224,10 +224,94 @@ class PoissonSolverNone : public PoissonSolver {
   }
   void ghost_communication() override {}
 
-  void solve() override {}
+  void solve() override { integrate_vtk_writers(); }
```
jngrad, line 232: Here and below: EKNone can now write its potential field to a file, in case it's non-zero. Until now, it would write a VTK file with the connectivity information but without the `potential` data field.

testsuite/python/ek_interface.py
```diff
@@ -488,6 +489,15 @@ def test_ek_indexed_reactions(self):
...
             ek_reaction[3:5, :, :].is_boundary, values)
 
+        # getters
+        ek_slice = ek_reaction[0:2, -4:-1, 1:2]
+        np.testing.assert_array_equal(
+            np.copy(ek_slice.call_method("get_slice_size")), [2, 3, 1])
+        np.testing.assert_array_equal(
+            np.copy(ek_slice.call_method("get_slice_ranges")),
+            [[0, 8, 1], [2, 11, 2]])
+        self.assertEqual(ek_slice.call_method("get_reaction_sip"), ek_reaction)
```
jngrad, line 541: needed to get code coverage (see CI report of a5d5f0756eddda8db58ad2c8e7956b12cd7d605e)

testsuite/python/save_checkpoint.py
```diff
@@ -100,8 +100,15 @@
...
-        ek_solver = espressomd.electrokinetics.EKNone(
-            lattice=lb_lattice, tau=system.time_step)
+        ek_solver_class = espressomd.electrokinetics.EKNone
+        ek_solver_params = {
+            "tau": system.time_step, "single_precision": False, "gpu": False}
+        if "LB.GPU" in modes:
+            ek_solver_params["gpu"] = True
+        if espressomd.has_features("WALBERLA_FFT") and "LB.ASCII" in modes:
+            ek_solver_class = espressomd.electrokinetics.EKFFT
+            ek_solver_params["permittivity"] = 0.1
```
jngrad: Completely arbitrary criterion: if writing checkpoint data in ASCII mode, use EKFFT, if writing in binary mode, use EKNone. This way both solvers get their VTK Poisson code tested.

src/script_interface/walberla/EKFFT.hpp
jngrad: @itischler could you please have a look at [this job log](https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/jobs/419596#L3390)? I don't see how my changes could cause the EKNone potential field to no longer be zero-initialized. This is not reproducible locally, and difficult to reproduce on a CI runner.

itischler: it seem as the problem is because walberla's `gpu:addGPUFieldToStorrage` (PoissonSolverNone.hpp:112) is not zero-initialized, as written in their docu, while in the CPU we set it to be 0. For the FFT case this also would not mater as the potential is calculated anew each step. So we have to zero it on initialization
```
#if defined(__CUDACC__)
    m_potential_field_id = gpu::addGPUFieldToStorage<PotentialField>(
        blocks, "potential field", 1u, field::fzyx,
        get_lattice().get_ghost_layers());
    for (auto &block : *blocks) {
      auto *potential_field =
          block.template getData<PotentialField>(m_potential_field_id);
      ek::accessor::Scalar::initialize(potential_field, FloatType{0});
    }
#else  // __CUDACC__
```

jngrad: Should be fixed by 6d52ec2bcc. To reproduce the bug, use this patch:
```diff
diff --git a/src/walberla_bridge/src/electrokinetics/PoissonSolverNone.hpp b/src/walberla_bridge/src/electrokinetics/PoissonSolverNone.hpp
index ad1eae245..e2cb961f9 100644
--- a/src/walberla_bridge/src/electrokinetics/PoissonSolverNone.hpp
+++ b/src/walberla_bridge/src/electrokinetics/PoissonSolverNone.hpp
@@ -116,3 +116,3 @@ class PoissonSolverNone : public PoissonSolver {
       auto field = block.template getData<PotentialField>(m_potential_field_id);
-      ek::accessor::Scalar::initialize(field, FloatType{0});
+      ek::accessor::Scalar::initialize(field, FloatType{5});
     }
```

testsuite/python/test_checkpoint.py
```diff
@@ -410,6 +426,62 @@ def test_ek_vtk(self):
...
+            # check file numbering when resuming VTK write operations
+            vtk_root = pathlib.Path("vtk_out") / \
+                f"manual_ek_poisson_{vtk_suffix}"
+            filename = "simulation_step_{}.vtu"
+            self.assertTrue((vtk_root / filename.format(0)).exists())
+            self.assertFalse((vtk_root / filename.format(1)).exists())
+            self.assertFalse((vtk_root / filename.format(2)).exists())
+            # check VTK objects are still synchronized with their EK objects
+            if isinstance(ek_solver, espressomd.electrokinetics.EKNone):
+                old_potential = ek_solver[0, 0, 0].potential
```
itischler: it should be fine, but this test only checks for pos = [0,0,0]

itischler (dismissed):

itischler (dismissed):

itischler (dismissed):

jngrad:
The testsuite should now pass in all environments.

@itischler @reinaual I made the `gpu` and `single_precision` arguments of `EKNone` match the EKSpecies, because it's the only way to make all kernels agree on data types and memory spaces. Making `EKNone` act like `None` introduces workarounds like `extract_solver(params.contains("solver") ? params.at("solver") : Variant{none})` that did not actually work (`none` crashes the code). Also, from the core, one cannot determine if the EK solver is EKNone since it's type-erased, and thus mixing GPU species with CPU EKNone would make it impossible to determine whether the EK container was GPU or CPU; the species `is_gpu` flag was used, but one could always change the solver after adding a species, and in doing so, create the right conditions to trigger a segfault. The code already checks if `EKNone` is single-precision and `EKSpecies` double-precision and prevents mixing them, and from my side it makes sense to have a similar check for CPU/GPU mixups.

itischler (approved):
All good now. Most of these issues should have caught my eye much sooner, so sorry for that. Thanks

PR #5330, opened 2026-06-01
Update CI/CD workflows

### LINKED ISSUE #5329: VTK test times out
https://gitlab.icp.uni-stuttgart.de/espressomd/espresso/-/pipelines/26366

jngrad:
Fixes #5329

Description of changes:
- update GitHub Actions to Node.js v24
- adjust tolerances and timeout of the VTK test

### DIFF
--- a/.github/workflows/push_pull.yml
+++ b/.github/workflows/push_pull.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@main
       - name: Setup ccache
-        uses: hendrikmuhs/ccache-action@v1.2
+        uses: hendrikmuhs/ccache-action@v1.2.23
         with:
           key: macos
           save: ${{ github.ref == 'refs/heads/python' }}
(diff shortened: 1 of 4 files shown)


.github/workflows/push_pull.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@main
       - name: Setup ccache
-        uses: hendrikmuhs/ccache-action@v1.2
+        uses: hendrikmuhs/ccache-action@v1.2.23
```
jngrad, line 18: The 1.2 rolling release tag is no longer updated since November 2025 (https://github.com/hendrikmuhs/ccache-action/issues/437). Node.js v20 will be removed from GitHub Actions in June 2026:
https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

testsuite/python/lattice_vtk.py
```diff
@@ -581,25 +582,6 @@ class LBWalberlaVTKDoublePrecisionCPU(TestLBVTK, ut.TestCase):
     vtk_id = "lb_double_precision_cpu"
 
 
-@utx.skipIfMissingGPU()
-@utx.skipIfMissingFeatures(["WALBERLA", "CUDA"])
-class LBWalberlaVTKDoublePrecisionGPU(TestLBVTK, ut.TestCase):
```
itischler, line 586: We do all CPU tests in double precision and GPU in single precision. Would it be better if for some of those test we switch it around, to test if the overall pypeline works e.g. test EKNoneDoubleGPU and EKNoneSingleCPU. it is probably not needed for all cases as with each added flag the number of tests scale exponentially.

jngrad: Good point. We can use the boundaries flag as a confounding factor for the precision and hardware.

itischler (approved):

PR #5334, opened 2026-06-11
Fix exclusion check for ghost particles

RudolfWeeber:
Ghost copies carry empty exclusion lists because the ghost communicator does
not serialize the `el` field.  `do_nonbonded()` combined the two one-sided
`none_of` results with `or`, which evaluates to `true` (not excluded)
whenever the ghost partner has an empty list, so the excluded LJ interaction
was computed anyway.

**Fix:** change `or` to `and` in `src/core/exclusions.hpp:43`.

**Regression test:** `ExclusionsGhostPairTest` added to
`testsuite/python/exclusions.py` — covers the cross-rank ghost case
(requires ≥ 2 MPI ranks) and a local control case.

Fixes bug #1 of the adversarial bug sweep.


RudolfWeeber:
This is genuine IMO (and test produces different outcomes on 1 and 2 mpi ranks). Test needs fixing.

itischler:
the test seems to work fine on 1 or two MPI ranks, and only failed earlier due to trying to double instantiate the system

jngrad (approved):

PR #5335, opened 2026-06-11
core/accumulators: fix Correlator::finalize() discarding compression results

RudolfWeeber:
In `finalize()` the compress calls did not assign their return values, so
the higher-hierarchy buffer slots (`A[i+1][newest[i+1]]`,
`B[i+1][newest[i+1]]`) were never updated.  The subsequent correlation loop
read stale ring-buffer data into the long-lag-time tail, silently corrupting
results for any multi-block correlator (`hierarchy_depth >= 2`, the standard
`tau_max/dt >= tau_lin` case).

**Fix:** assign the compress results in `finalize()` (lines 477-480 of
`src/core/accumulators/Correlator.cpp`), mirroring the identical assignment
already present in `update()`.

**Regression test:** `test_finalize_preserves_long_lag_tail` added to the
`CorrelatorTest` class in `testsuite/python/accumulator_correlator.py`.

Fixes bug #2 of the adversarial bug sweep.

### DIFF
--- a/src/core/accumulators/Correlator.cpp
+++ b/src/core/accumulators/Correlator.cpp
@@ -474,10 +474,12 @@ int Correlator::finalize(boost::mpi::communicator const &comm) {
         newest[i + 1] = (newest[i + 1] + 1) % (m_tau_lin + 1);
         n_vals[i + 1] += 1;
 
-        (*compressA)(A[i][(newest[i] + 1) % (m_tau_lin + 1)],
-                     A[i][(newest[i] + 2) % (m_tau_lin + 1)]);
-        (*compressB)(B[i][(newest[i] + 1) % (m_tau_lin + 1)],
-                     B[i][(newest[i] + 2) % (m_tau_lin + 1)]);
+        A[i + 1][newest[i + 1]] =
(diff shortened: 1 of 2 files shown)


RudolfWeeber:
@jngrad Is this the corrleator issue that has been discussed?
My impression is that the issue and fix are genuine.

jngrad:
This indeed looks like the MSD bug reported by a student of @pm-blanco to @1234somesh on February 2025.

pm-blanco:
That would indeed explain the behaviour we observed two years ago. Great to see that finally you got to the root of the issue!

jngrad:
I don't have @pm-blanco's original script, but I worked something out based on the new test case.

Before the bugfix:

MSD plot before the bugfix, showing correlation data being lost at large tau values when finalized() is called

After the bugfix:

MSD plot after the bugfix, showing correlation data being preserved at large tau values when finalized() is called

jngrad (approved):

pm-blanco:
@jngrad that is exactly what we observed. That confirms that the bug is fixed, which is great since we have been having issues with having good accuracy at long lag times for the MSD for quite a long time.

PR #5336, opened 2026-06-11
Make ParticleSlice getter return pid-sorted values

RudolfWeeber:
`get_particles_properties()` sorted gathered `(pid, value)` pairs by
ascending pid before returning them, but setters write values positionally
against the user-supplied `id_selection` order.  For any non-ascending
`by_ids([3,1,2])` selection the set/get round-trip silently returned
reordered data.

**Fix:** replace the sort-then-iterate with a `pid→value` unordered map
that returns values in the original `id_selection` (`pids`) order
(`src/script_interface/particle_data/ParticleSlice.hpp`).

**Regression tests:** `test_charge_roundtrip_non_ascending` and
`test_pos_roundtrip_non_ascending` added to `ParticleSliceTest` in
`testsuite/python/particle_slice.py`.

Fixes bug #4 of the adversarial bug sweep.


RudolfWeeber:
I think, the performance decline due to the std::unorderd_map is probably not significant compared to the python stuff (but not tactually testsed). We can come back to it, if it turns out to be too slow in practical scenarios.

jngrad:
The performance penalty was reduced by eliding one of the two `std::move` operations. If need be, we could look into introducing a new branch for node-level simulations (1 MPI rank) where data is stored in the correct order to avoid move operations.

jngrad (approved):

PR #5337, opened 2026-06-11
Fix rescaling of non-cubic boxes

RudolfWeeber:
`change_volume_and_rescale_particles("xyz")` computed a single scalar scale
from the x-axis (`length * length_inv()[0]`) and applied it to all three
particle coordinates, silently corrupting y/z positions on any non-cubic
box.  The cubic-box assert was compiled out under NDEBUG.

**Fix:** for `coord==3`, capture old per-axis lengths, apply shrinking axes
before the box resize and growing axes after, each with its own
`length/old_lengths[axis]` scale factor
(`src/script_interface/system/System.cpp`).  Also removes the now-vacuous
cubic-box assert.

**Regression test:** `test_isotropic_rescale_noncubic_box` added to `Test`
in `testsuite/python/system.py`.

Fixes bug #16 of the adversarial bug sweep.


RudolfWeeber:
Bug is genuine IMO. Test case needs fixing.

jngrad (approved):

PR #5339, opened 2026-06-11
Fix change_reaction_constant rejecting odd reaction_ids

RudolfWeeber:
`change_reaction_constant` had a contradictory guard:
`if (reaction_id % 2 == 1) throw "Only forward reactions can be selected"`.
This interpreted `reaction_id` as a raw index into the doubled
`m_reactions` container, but `get_reaction_index()` (and the public
docstring) treat it as a logical index and multiply by 2 internally.
Every odd logical `reaction_id` (the 2nd, 4th, … added reaction) was
wrongly rejected.

**Fix:** remove the `% 2 == 1` guard from
`src/script_interface/reaction_methods/ReactionAlgorithm.cpp`.
Range validation is performed by `get_reaction_index()`.

Also updated the existing `test_exceptions` in
`testsuite/python/reaction_methods_interface.py` to expect the correct
`IndexError` from `get_reaction_index()` for an out-of-range id.

**Regression test:** `test_change_constant_of_second_reaction` added to
`ReactionMethods` in `testsuite/python/reaction_methods_interface.py`.

Fixes bug #36 of the adversarial bug sweep.


RudolfWeeber:
@ltarrach, could you please take a look?

testsuite/python/reaction_methods_interface.py
17ena: The note `reaction id = 2 * reaction index` is not consistent with the function `get_reaction_index()` because there `reaction index = 2 * reaction id`. The latter version must be consistent with the rest of the code and, thus, be correct. The `reaction_id` is assigned according to the order in which `add_reaction()` was called. To each `reaction_id` belongs a forward reaction with `reaction index = 2 * reaction id` and a backward reaction with  `reaction index = 2 * reaction id +1`.

testsuite/python/reaction_methods_interface.py
```diff
@@ -264,7 +264,7 @@ def test_exceptions(self):
                 method.delete_reaction(reaction_id=i)
             with self.assertRaisesRegex(IndexError, f"No reaction with id {2 * i}"):
                 method.get_acceptance_rate_reaction(reaction_id=2 * i)
```
17ena, line 266: As far as I understand, `get_acceptance_rate_reaction()` uses the reaction index, but not the reaction id. Therefore, the nomenclature of the parameter `reaction_id` in the function `get_acceptance_rate_reaction()` appears misleading to me as it must be an reaction index.

17ena (commented):
I approve the changes made to the code, but I have two comments on other lines involved in the code related to the pull request.

17ena (approved):

jngrad:
@pm-blanco neither `get_acceptance_rate_reaction()` nor `change_reaction_constant()` are called by pyMBE, so I do not see any further action required beyond this PR.

That said, I would be in favor of dropping this reaction id vs index API entirely, and instead let users pass around `SingleReaction` objects directly, perhaps in ESPResSo 5.1 already. That might be a bit tricky w.r.t. checkpointing, since we would have to guarantee two objects are equal via some unique id (uuid) when reloading from a checkpoint file. It's also unclear to me how much work that would cause in pyMBE. Are reaction ids used for bookkeeping in data frames?

jngrad (approved):

pm-blanco:
@jngrad just to say that I am also in favour of dropping the reaction id logic entirely (I find it very confusing) and leverage on the SingleReaction class. Regarding the bug, good catch! I don't know any user that was using those methods though, I guess that is why they were undetected for long

pm-blanco:
@jgrad regarding the bookkeeping of reactions on pyMBE - we do not use the espresso reaction id at all

PR #5343, opened 2026-06-12
core/pressure: fix DPD virial dropping excluded pairs

RudolfWeeber:
Move the DPD force accumulation inside the particle exclusion loop to mirror the DPD pressure exclusion.

Fixes bug #7 of the adversarial core bug sweep.


RudolfWeeber:
@Yashas17 @hidekb The bug is genuine, stress and force should be consistent. But which way? Is it actually intended, that the that the dpd force between excluded particles is clauclated? I think it needs to be fixed the other way around: exclude for both, force and pressure?

jngrad (approved):

PR #5345, opened 2026-06-12
core/p3m: fix influence function zeroing valid k-modes for odd mesh sizes

RudolfWeeber:
The influence-function skip condition `indices[d] % (mesh/2) == 0` correctly
identifies only DC and Nyquist for even meshes, but for odd meshes it
spuriously zeroes legitimate low/mid-frequency modes, corrupting dipolar
P3M forces and energies.

**Fix:** replace the modulo shortcut with an `is_dc_or_nyquist` lambda that
returns true only for index == 0 (DC) or (mesh even AND index == mesh/2,
Nyquist), in both `influence_function.hpp` and `influence_function_dipolar.hpp`.

**Regression test:** `test_influence_function_odd_mesh` added to
`testsuite/python/p3m_madelung.py`.

Fixes bug #9 of the adversarial core bug sweep.

### DIFF
--- a/src/core/p3m/influence_function.hpp
+++ b/src/core/p3m/influence_function.hpp
@@ -166,13 +166,21 @@ std::vector<FloatType> grid_influence_function(
   }
 
   auto const wavevector = (2. * std::numbers::pi) * inv_box_l;
-  auto const half_mesh = params.mesh / 2;
   auto indices = Utils::Vector3i{};
 
+  /* Skip G_opt only for DC (index == 0) and, for even mesh, the Nyquist mode
+   * (index == mesh/2). The old half_mesh % test was incorrect for odd mesh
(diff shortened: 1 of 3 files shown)


RudolfWeeber:
@jngrad I think the issue might be genuine. The test thac Claude wrote is meaingless, as we prevent odd grid sizes. How shuold we proceed?

jngrad:
> @jngrad I think the issue might be genuine. The test thac Claude wrote is meaingless, as we prevent odd grid sizes. How shuold we proceed?

There are other parts of the codebase that assume the mesh is even, for example `calc_p3m_mesh_shift()` takes an optional boolean parameter to zero out the Nyquist frequency, and that would require fixing too. I'll try to come up with a more thorough fix for it. It won't really change anything for Coulomb P3M since we only tune even meshes for performance reasons, but for dipolar P3M this indeed looks like a bug.

PR #5346, opened 2026-06-12
shapes: fix Torus::calculate_dist producing NaN for particles on the torus axis

RudolfWeeber:
When a particle lies on the torus symmetry axis (`r == 0`), `r_vec` is the zero
vector and `r_vec * m_rad / r` computes `0 * inf = NaN`, propagating NaN into
the constraint force direction.  Sibling axial shapes (`Cylinder`, `SimplePore`,
`SpheroCylinder`) all branch on `r == 0` to avoid the division.

**Fix:** add an `e_r_axis` member computed via Gram-Schmidt; replace the bare
`r_vec * m_rad / r` with `e_r * m_rad` where
`e_r = (r == 0) ? e_r_axis : r_vec / r`
(`src/shapes/include/shapes/Torus.hpp`, `src/shapes/src/Torus.cpp`).

**Regression test:** `src/shapes/unit_tests/Torus_test.cpp` (Boost.Test).

Fixes bug #14 of the adversarial core bug sweep.

### DIFF
--- a/samples/visualization_constraints.py
+++ b/samples/visualization_constraints.py
@@ -67,31 +67,31 @@ if args.shape == "Wall":
 
 elif args.shape == "Sphere":
     system.constraints.add(shape=espressomd.shapes.Sphere(
-        center=[25, 25, 25], radius=15, direction=1),
-        particle_type=0, penetrable=True)
+        center=[25, 25, 25], radius=15, direction=-1),
+        particle_type=0, penetrable=False)
 
 elif args.shape == "Ellipsoid":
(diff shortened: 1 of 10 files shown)


RudolfWeeber:
OK, I understand, the r=0 case muost be caught. I'm not sure about the fallback.

jngrad:
There is also a singularity when the position lies on the torus surface, a singularity when the position lies on the core circle, and a numerical instability when the position is within 10e-15 of the core circle.

jngrad (approved):

PR #5347, opened 2026-06-12
core/bonded_interactions: fix AngleCosine force not sanitizing cosine at collinear configs

RudolfWeeber:
`AngleCosineBond::forces()` called `angle_generic_force(..., sanitize_cosine=false)`,
so at near-collinear arrangements (including the default `phi0=pi` equilibrium)
`sin_phi -> 0` and the `/sin_phi` force factor yielded `NaN`/`inf`, irreversibly
corrupting the trajectory.  The sibling `AngleHarmonic` passes `true`.

**Fix:** change the `false` to `true` on line 82 of
`src/core/bonded_interactions/angle_cosine.hpp`.

**Regression test:** `src/core/unit_tests/angle_cosine_test.cpp` (Boost.Test).

Fixes bug #15 of the adversarial core bug sweep.

### DIFF
--- a/src/core/bonded_interactions/angle_cosine.hpp
+++ b/src/core/bonded_interactions/angle_cosine.hpp
@@ -79,7 +79,7 @@ AngleCosineBond::forces(Utils::Vector3d const &vec1,
     return -bend * (sin_phi * cos_phi0 - cos_phi * sin_phi0) / sin_phi;
   };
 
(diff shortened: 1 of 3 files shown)


RudolfWeeber:
@jgrad is this about the argument to arccos remaining valid? Or a false positive

PR #5348, opened 2026-06-12
utils: fix linear_interpolation reading one past end at top of range

RudolfWeeber:
`linear_interpolation` returned `table[uind] * (1-dx) + table[uind+1] * dx`.
When `x == maxval` (the clamped upper bound), `uind == table.size()-1` and
`table[uind+1]` is an out-of-bounds heap read.  Although `dx == 0` makes the
OOB read coefficient zero, it is ASan-flagged UB.  `TabulatedPotential` always
clamps to the closed interval `[minval, maxval]`, so `x == maxval` is a normal
path for any tabulated potential evaluated at or beyond the cutoff.

**Fix:** clamp `ind` to at most `table.size()-2` in `linear_interpolation`
(`src/utils/include/utils/linear_interpolation.hpp`).

**Regression test:** `interpolation_at_endpoint` case added to
`src/utils/tests/linear_interpolation_test.cpp`.

Fixes bug #17 of the adversarial core bug sweep.

### DIFF
--- a/src/utils/include/utils/linear_interpolation.hpp
+++ b/src/utils/include/utils/linear_interpolation.hpp
@@ -19,6 +19,7 @@
 
 #pragma once
 
+#include <algorithm>
 #include <cassert>
 #include <cstddef>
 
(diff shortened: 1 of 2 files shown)


RudolfWeeber:
Not sure this is genuine. The previous code looked plausibel, but so does the new test case.

PR #5360, opened 2026-06-16
core: guard null central particle in bond breakage angle branch (bug-sweep #26)

RudolfWeeber:
The REVERT_BIND_AT_POINT_OF_COLLISION angle-bond branch of
actions_for_breakage guarded only the two partner particles
(if (p1 and p2)) and then dereferenced vs->is_virtual() without
checking vs != nullptr. The global breakage queue is broadcast to
every MPI rank, so each rank runs actions_for_breakage for every
entry, including entries whose central virtual-site particle is not
local/ghost on this rank. On a rank where both partners resolve but
the central particle does not, get_local_particle(e.particle_id)
returns nullptr and vs->is_virtual() is a null dereference (SIGSEGV).

Fix: guard with if (vs and p1 and p2), mirroring the pair branch which
already guards its central particle. Absence-on-rank is a legitimate
locality condition, so the branch returns an empty action set rather
than throwing.

To make the function unit-testable, de-static actions_for_breakage and
declare it in bond_breakage.hpp (the Action/ActionSet aliases move to
actions.hpp). Adds a regression test that reproduces the locality
condition on a single rank (partners present, central absent).

### DIFF
--- a/src/core/bond_breakage/actions.hpp
+++ b/src/core/bond_breakage/actions.hpp
@@ -23,6 +23,8 @@
 
 #include <array>
 #include <cstddef>
+#include <unordered_set>
+#include <variant>
 
 namespace BondBreakage {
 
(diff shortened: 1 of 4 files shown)


RudolfWeeber:
I think this is legit, although it is not obvious to me, how to trigger it from Python.

PR #5379, opened 2026-06-16
utils: fully reset RunningAverage in clear()

RudolfWeeber:
RunningAverage::clear() only reset the sample count m_n, leaving the
running average, variance, and min/max accumulators stale. After a
clear() followed by re-sampling, var() folded the pre-clear variance
into the new data and min()/max() returned the pre-clear extrema,
silently corrupting a reused accumulator.

Reset all members in clear() so it mirrors the constructor's clean
state. Add a clear_then_resample unit test that re-samples after
clear() and checks avg/var/min/max.


jngrad (approved):
LGTM

PR #5384, opened 2026-06-16
script_interface: fix LB node _population OOB read

RudolfWeeber:
The LBFluidNode "set_population" branch forwarded the user-supplied
vector to LBWalberlaBase::set_node_population verbatim, with no
length check at any layer (Python, SI, or core). The core
LBNodeAccess loop reads population[f] for f in [0, Stencil::Size)
(19 for D3Q19) via unchecked operator[], so a too-short vector (e.g.
lbf[0,0,0]._population = [1., 2.]) reads past the std::vector buffer,
injecting garbage populations or crashing. The slice path already
validates length via get_value_shape; the node path did not.

Fix: PREVENT the bad input. Before calling set_node_population, check
pop.size() == m_lb_fluid->stencil_size() inside parallel_try_catch so
the throw on the owning rank propagates uniformly to a Python
exception instead of an MPI desync. A malformed population vector is a
user error with no meaningful completion; padding/truncating would
silently inject incorrect physics. This mirrors the existing
checkpoint "population size mismatch" validation (LBFluid.cpp).

### DIFF
--- a/src/script_interface/walberla/LBFluidNode.cpp
+++ b/src/script_interface/walberla/LBFluidNode.cpp
@@ -33,6 +33,7 @@
 #include <boost/mpi/communicator.hpp>
 
 #include <optional>
+#include <stdexcept>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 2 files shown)


jngrad (approved):
LGTM

PR #5385, opened 2026-06-16
core/immersed_boundary: fix Tribend concave reference angle (bug-sweep #5)

RudolfWeeber:
IBMTribend::calc_forces() builds the dynamic dihedral angle as
theta = acos(n1*n2) * copysign(1, desc), confined to [-pi, pi], while
initialize() maps a concave reference dihedral (desc = dx1 . (n1 x n2) < 0)
onto theta0 = 2*pi - acos(...) in [pi, 2*pi]. The two angles thus live on
different branches and the raw difference DTh = theta - theta0 was never
reduced modulo 2*pi. At a concave reference equilibrium DTh = -2*pi instead
of 0, applying a spurious ~2*pi*kb bending force at the relaxed shape --
exactly the canonical biconcave-RBC / vesicle membrane use case.

Fix: reduce DTh to (-pi, pi] via std::remainder(theta - theta0, 2*pi),
keeping the signed theta and the copysign(1, theta) orientation factor
unchanged. This zeroes the force at both convex and concave reference
shapes and yields correctly restoring forces on both sides of equilibrium.
(Note: the alternative of remapping the dynamic theta to [0, 2*pi] would
collapse copysign(1, theta) to +1 and make concave references unstable.)

Adds two regression tests to testsuite/python/ibm.py: a convex control
(passes even on buggy code) and the concave equilibrium case (force ~6 on
buggy HEAD, < 1e-6 after the fix).

### DIFF
--- a/src/core/immersed_boundary/ibm_tribend.cpp
+++ b/src/core/immersed_boundary/ibm_tribend.cpp
@@ -61,7 +61,14 @@ IBMTribend::calc_forces(BoxGeometry const &box_geo, Utils::Vector3d const &pos1,
   auto const desc = (dx1 * direc);
   auto const theta = std::acos(sc) * std::copysign(1., desc);
 
-  auto const DTh = theta - theta0;
+  // The dynamic angle theta is confined to [-pi, pi], whereas the reference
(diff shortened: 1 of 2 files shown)


RudolfWeeber:
not obvious to me. Needs more detailed analysis.

Issue #5067, opened 2025-03-31
Missing contributions to stress tensor


RudolfWeeber:
some contributions to stress tensor are not implemented, either in `system.analysis.pressure_tensor()` or in NPt instantaneous pressure

* [ ] contribution for rattle (called "Rigid bond" in ESPResSo)
* [ ] pressure from dipolar interactions

Electrostatic and dipolar layer correction might be conceptually difficult w.r.t presusre due to the empty gap region they require.

PR #5388, opened 2026-06-29
Add virial contribution for rigid bonds

hidekb:
Fixes Partially #5067 

Description of changes:
- Added virial calculation for rigid constraint in pressure.cpp and rattle.cpp
- Added new test script `rigid_bond_virial.py` that verifies the constraint virial


jngrad (approved):
LGTM

PR #5392, opened 2026-07-06
Integrate Exscalicos

rhalver:
Adds a first draft for a potential interface to the ExScaLiCoS electrostatic solvers library:

Description of changes:
- reuses the ScaFaCoS library interface add performs a RMS error comparison of the results
- provides / uses tests (34-37) based on systems from ScaFaCoS

Issues:
- needs to be streamlined for production use (e.g. removal of testing RMS computation)
- near-field contribution delegation needs to be adjusted
- further integration into Python layer missing

[ ... ]


rhalver:
The Kokkos error is presumably from the change of find_package(Kokkos QUIET) to find_package(Kokkos REQUIRED), which was necessary as we compiled our tests with a preinstalled version of ExScaLiCoS and its dependencies, I will have a look at it, if I can provide an auto-installer similar to the ones used for Kokkos, Cabana and HeFFTe, if no installed version is present.

jngrad:
The CMake integration now supports both the `FindPackage` and `FetchContent` mechanisms for ExScaLiCoS. The latter seamlessly replaces `find_package(Cabana REQUIRED)` in ExScaLiCoS by the Cabana version built from sources in ESPResSo. There are still linker issues with CUDA and runtime issues due to the Python ScaFaCoS object somehow outliving the MPI communicator, which I need to figure out.

PR #5399, opened 2026-07-21
Node-level performance improvements for P3M and LJ

RudolfWeeber:
# Short-range and P3M performance backport

This branch backports the single-rank CPU performance work to the `python`
branch. It has two independent thrusts — the P3M reciprocal-space path and the
short-range Verlet pair loop — plus the refactors and tests that support them.

Unscientific benchmarks give some 15% on LJ and some 30% on P3M, as well as improved shared-memory parallel efficiency for the latter.

## P3M reciprocal space

* **Thread heFFTe's single-rank reshapes** (`cmake/heffte.patch`, applied via
  FetchContent, mirroring the existing cabana patch). heFFTe 2.4.1 runs its
  pack/unpack/transpose reshapes and post-FFT scaling serially, which caps
  OpenMP scaling on one rank. The pragmas are `#ifdef _OPENMP`-guarded so
  OpenMP-less builds still compile. 
* **Parallelize the mesh crop/pad copies** and **skip the zero-init of the
  crop buffer** (`field_layout_helpers.hpp`). The lane-by-lane `memmove` was a
  large serial slice of the profile; per-lane offsets make the outer loop an
  `omp parallel for` over disjoint lanes. `extract_block` writes every output
  element, so the buffer is now default-initialized instead of value-
  initialized. 
* **Optional kokkos-fft reciprocal-space backend, on by default**
  (`ESPRESSO_BUILD_WITH_KOKKOS_FFT`, default ON; `OFF` avoids the dependency).
  A new abstract `P3MFFTBackend` interface fronts the FFT: heFFTe stays behind
  a thin adapter (and is still selected at runtime for multi-rank runs), while
  single-rank CPU runs get `P3MFFTKokkos`, an r2c transform on Kokkos host
  views reproducing heFFTe's layout and scaling convention. The backend runs
  zero-copy on the P3M buffers: plans are built on and cached by the exact
  buffers they execute on, and the charge density is extracted straight into
  the FFT input buffer (`forward_input_buffer()` + `extract_block_into`),
  eliminating the per-step staging copies and allocation.
* **Solver renamed to backend-neutral names**: `CoulombP3MHeffte` →
  `CoulombP3MImpl`, `p3m_heffte.*` → `p3m.cpp` / `p3m.impl.hpp` /
  `p3m.impl.definitions.hpp` (matching upstream naming). Rename only. The
  dipolar solver keeps its `dp3m_heffte` name — it still uses heFFTe
  unconditionally.
* **Benchmark support**: `p3m.py` gains `--mesh` to prescribe the (cubic) P3M
  mesh while still tuning cao/r_cut/alpha.

## Short-range Verlet pair loop

* **Branchless cuboid minimum-image fold**: periodicity is encoded in a masked
  inverse box length so the fold is `dx - rint(dx * inv_masked) * L` with no
  branch; box parameters are hoisted by value into the kernels via
  `CuboidMinimumImage`.
* **Flat per-type-pair squared-cutoff table** in `VerletCriterion` replaces
  the per-candidate walk of the interaction pointer table; inactive pairs hold
  a negative sentinel. A new `ShortRangeOnly` template parameter additionally
  compiles the electrostatics/dipolar/collision branches out of the build loop
  when those cutoffs are inactive.
* **Specialized non-bonded pair kernel** (`SpecializedForcesKernel`): owns the
  Verlet-list loop (one invocation per particle instead of per pair), so the
  ScatterView accessor — a per-pair `omp_get_thread_num` in the profile — is
  obtained once per particle, with position/type/charge hoisted out of the
  neighbor loop. Neighbors are processed in fixed-size tiles: a scalar gather,
  a vectorized (packed-AVX) minimum-image + squared-distance pass, then a
  scalar force pass only for pairs that survive the cutoff gate. The i-side
  force accumulates in a register and hits the ScatterView once per particle.
  It is installed through an optional `ShortRangeVerletPairLoop` hook and
  selected only for cuboid boxes with no NPT virial, dipolar/ELC kernel, DPD
  thermostat, Thole/Gay-Berne pair, or exclusions; everything else falls back
  to the generic `ForcesKernel`. Exclusion presence is now an O(1) aggregate
  maintained during the AoSoA commit sweep instead of a per-call sweep.
* **Translation-unit split**: force initialization and the Verlet-list build
  move to `forces_init.cpp` / `short_range_verlet.cpp` (pure code motion,
  backport of 3a54ec10ee).

## Build/compat notes

* New CMake option `ESPRESSO_BUILD_WITH_KOKKOS_FFT` (default ON); multi-rank
  runs select heFFTe at runtime regardless.
* heFFTe patch pragmas are no-ops without OpenMP.

## Commits (initial)

- **build: thread heFFTe's reshape kernels via a FetchContent patch**
- **p3m: parallelize the mesh crop/pad copies**
- **p3m: skip the zero-init of the crop buffer**
- **benchmarks: let the p3m benchmark prescribe the P3M mesh**
- **p3m: optional single-rank kokkos-fft reciprocal-space backend**
- **p3m: run the kokkos-fft backend zero-copy on the P3M buffers**
- **p3m: extract the charge density straight into the FFT input buffer**
- **core: split force initialization and verlet-list build into own translation units**
- **core: branchless minimum-image fold and flat Verlet-list cutoff table**
- **core: own-the-loop specialized non-bonded pair kernel**
- **docs: use a code span for the get_mi_coord_masked reference**
- **core: batch the specialized kernel's minimum-image gate over neighbors**
- **core: compile inactive-feature branches out of the Verlet criterion**
- **docs: describe the criterion call operator without a doxygen ref**
- **style: apply maintainer formatters to backported files**
- **p3m: keep the FFT backend's real-space scalar real for c2c configs**
- **core: share the central-radial + coulomb pair force between the cabana kernels**
- **core: track exclusion presence during the aosoa commit**
- **Revert "core: share the central-radial + coulomb pair force between the cabana kernels"**
- **p3m: enable the kokkos-fft backend by default**
- **build: guard the heffte patch pragmas for compilers without OpenMP**
- **core: accumulate the specialized kernel's i-side force in a register**
- **core: rename the P3M solver implementation to backend-neutral names**
- **core: unit-test the cuboid minimum-image fold against the BoxGeometry path**
- **core: define the always-inline attribute macro in one header only**


jngrad (approved):
LGTM.

Here are benchmark results for 1000 particles per thread:

| benchmark | threads | python branch (ms/step) | PR (ms/step) | speed-up |
| --- | --- | ---| ---| ---|
| lj.py | 1 | 0.172 | 0.162 | 6.2% |
| lj.py | 2 | 0.202 | 0.189 | 6.5% |
| lj.py | 4 | 0.237 | 0.220 | 7.1% |
| p3m.py | 1 | 0.845 | 1.008 | 16% |
| p3m.py | 2 | 1.207 | 1.365 | 12% |
| p3m.py | 4 | 1.633 | 1.305 | 20% |

The MPI version of the P3M remains unaffected. The OpenMP patches to heFFTe only bring a measurable improvement for large grids, e.g. P3M with a 80x80x80 mesh is unaffected, but ekin with a 128x128x64 mesh shows a 2.5% speed-up.

PR #5400, opened 2026-07-22
Add support for Kokkos on GPU

jngrad:
Description of changes:
- make CPU the default Kokkos execution space and prepare infrastructure for the CUDA execution space
- sort out toolchain compiler flags for combined C++/CUDA compilation
- add a function to query the compiler toolchain used to build ESPResSo


jngrad:
There are temporary workarounds in several places that will disappear once we migrate to Ubuntu 26.04 and CUDA >= 12.6. More adjustments will be made in future PRs, when we integrate ExScaLiCoS and the waLBerla commit with the new build system.

This Kokkos CUDA integration doesn't rely on the Kokkos nvcc wrapper, which would have required even more complex CMake workarounds to restore the ccache functionality and fix compiler diagnostics. The most difficult aspect of this integration is that every C++ translation unit that sees Kokkos must be compiled in CUDA mode, whereby nvcc builds the device code and delegates the host code compilation to the CUDA host compiler (gcc or nvc++). Relocatable device code is disabled since it introduces a lot of complexity to the build system.

Calls to CMake function `espresso_set_kokkos_properties_on_source_files()` must be done towards the end of CMake files, after the last `include()` statement, since it iterates the list of C++ source files and updates their property to be CUDA source files. That must happen once the target is fully populated with all source files from all subfolders.

Many C++ functions must be made `constexpr` even though they are almost never called in a `constexpr` context. This is a side-effect of CUDA's flag `-expt-relaxed-constexpr`, which allows device kernels to call `constexpr` host functions. Although such a code path is never encountered for  a `Particle` pointer (and cannot happen, since host pointers do not live in the same address space as device memory), the compiler has to consider such a possibility during template instantiation.

src/core/electrostatics/p3m_heffte.impl.hpp
```diff
@@ -315,9 +315,12 @@ void CoulombP3MHeffte<FloatType, Architecture, FFTConfig>::init_cpu_kernels() {
...
   p3m.local_mesh.calc_local_ca_mesh(p3m.params, local_geo, skin, elc_layer);
-  p3m.fft = std::make_shared<P3MFFT<FloatType, FFTConfig>>(
-      ::comm_cart, p3m.params.mesh, p3m.local_mesh.ld_no_halo,
-      p3m.local_mesh.ur_no_halo, ::communicator.node_grid);
+  for (int d = 0; d < 3; ++d) {
+    p3m.ffts[d] = std::make_shared<P3MFFT<FloatType, Arch::CPU, FFTConfig>>(
+        nullptr, ::comm_cart, p3m.params.mesh, p3m.local_mesh.ld_no_halo,
+        p3m.local_mesh.ur_no_halo, ::communicator.node_grid);
+  }
+  p3m.fft = p3m.ffts[0];
```
reinaual: Maybe I missed it, but what's the purpose of these 3 FFTs?

jngrad: A leftover from the failed P3M GPU rewrite. Meant to stagger the FFTs, such that one transform runs while the GPU sends the result from the previous FFT to the CPU. I'll check if it can be easily removed, since it's of little use on the host.

reinaual (approved):
I'll spare myself a comment on this “beauty”

PR #5403, opened 2026-07-27
Benchmark improvements

RudolfWeeber:
Description of changes:
* allow for tuning and measuring benchmakrs in separate runs, to get tuning out of perf/caliper profiles
* allow to set skin via commandline for lj and p3m
* allow to specify mesh or mesh tuning limits via cmd line for p3m
* Make setting the particle count more intuitive: either --num_particles for an absolute value or --particles_per_core which counts omp_threads*mpi_ranks
* Number of omp threads is available in the python interface via the cell system state


RudolfWeeber:
@jngrad The tests that test the benchmark scripts fiddle with module level globals to run short versions of the test. 
This pr still allows this, but I'm wondering whether we shouldn't rather have a --quick option or similar. That way, we only test the public interface and not internal details.

jngrad:
> I'm wondering whether we shouldn't rather have a --quick option or similar. That way, we only test the public interface and not internal details.

I agree, but let's merge this for now. I've been working on benchmark improvements too, and some of your changes (e.g. passing a P3M mesh size) are identical to the ones I already implemented :sweat_smile:

I'll look into adding a `--quick` option in a future PR.

jngrad (approved):

PR #5405, opened 2026-07-31
Split the ek tutorial

itischler:
Ek tutorial was split in 3 parts so that each discussed ek system has its own notebook


jngrad (approved):
LGTM

PR #5407, opened 2026-08-03
Dipolar P3M dipole field tracking

stekajack:
This PR adds dipole-field tracking support to the Kokkos implementation of `DipolarP3M`.

When `DIPOLE_FIELD_TRACKING` is enabled, DP3M now calculates and stores the total magnetic field acting on each particle as part of the existing force-and-torque calculation. This avoids a separate field-calculation pass and keeps the field consistent with the torque used by the integrator.

The implementation includes all DP3M contributions:

- Real-space pair contribution
- Reciprocal-space mesh contribution
- Dipolar surface term
- Local and ghost-particle contributions across MPI ranks

## Implementation

A Kokkos `ScatterView` is added for the local dipole field, following the existing force and torque accumulation model.

During force calculation:

1. The dipole-field buffers are reset together with the force and torque buffers.
2. The DP3M real-space kernel returns the field acting on both particles in each pair.
3. The reciprocal-space torque interpolation also accumulates the corresponding field.
4. The surface-term field is added directly to each particle.
5. The Kokkos scatter buffer is reduced into `Particle::dip_fld()`.
6. Dipole fields accumulated on ghost particles are reduced back to their owning particles.

Consequently, field tracking uses the same real-space and reciprocal-space evaluations as the force and torque calculation and does not require an additional DP3M evaluation. The implementation supports both distributed-memory and shared-memory parallelism:
- MPI ghost communication now supports dipole-field reduction.
- Kokkos accumulation uses a dedicated `ScatterView`, consistent with the current force and torque implementation.
- The field is reset and reduced according to the existing force-calculation lifecycle.


testsuite/python/dipole_field_tracking.py
```diff
@@ -127,6 +127,17 @@ def test_dds_gpu(self):
                     np.copy(p.torque_lab), np.cross(p.dip, p.dip_fld),
                     rtol=1e-9, atol=1e-5)
 
+    @utx.skipIfMissingFeatures(["DP3M"])
+    def test_dp3m(self):
+        self.system.periodicity = [True, True, True]
+        solver = espressomd.magnetostatics.DipolarP3M(
+            prefactor=1., accuracy=1e-8, epsilon=1.)
```
jngrad: I don't think `epsilon` is needed here. It only modifies the surface correction term in the energy calculation, which doesn't seem to play a role in the dipolar field calculation, but I could be wrong. Let me know if this parameter was added here intentionally. Please note that `epsilon` is unrelated to the magnetostatic prefactor, instead it controls the boundary conditions at infinity.

jngrad: I stand corrected: the energy is actually added to the dipole fields. Let's see if I can come up with a test that fails when epsilon has the wrong value.

jngrad: nevermind, that was actually intentional and correctly tested :sweat_smile:

jngrad (approved):

PR #5413, opened 2026-08-07
Remove script interface overhead and MPI+OpenMP overhead

jngrad:
Description of changes:
- skip superfluous MPI serialization and communication in the MpiCallbacks framework when the world size is 1
- avoid OpenMP fork/join in `Kokkos::deep_copy` and `Kokkos::parallel_for` when using only 1 thread per rank
- improve cache re-use in the energy and pressure observables
- reduce overhead of the Python script interface
   - make conversion from C++ `Variant` to Python `ScriptInterfaceHelper` much cheaper
   - make conversion from Python `object` to C++ `Variant` much cheaper
   - make string conversion operations much cheaper by leveraging interned strings
   - bind C++ methods only once per class definition, rather than once per class instantiation
- bound C++ methods can now display their docstring in Jupyter

### DIFF
--- a/src/core/kokkos_helpers.hpp
+++ b/src/core/kokkos_helpers.hpp
@@ -1,0 +1,62 @@
+/*
+ * Copyright (C) 2025-2026 The ESPResSo project
+ *
+ * This file is part of ESPResSo.
+ *
+ * ESPResSo is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * ESPResSo is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include <algorithm>
+#include <type_traits>
+
+#include <Kokkos_Core.hpp>
+
+/**
+ * @brief Wrapper for ``Kokkos::deep_copy`` that skips fork/join when
+ * the number of threads is 1.
+ */
+ESPRESSO_ATTR_ALWAYS_INLINE inline void
+kokkos_deep_copy(auto const &exec_space, auto const &view, auto const &value) {
+  using View = std::remove_cvref_t<decltype(view)>;
+  if constexpr (Kokkos::SpaceAccessibility<
+                    Kokkos::DefaultHostExecutionSpace,
+                    typename View::memory_space>::accessible) {
+    if (Kokkos::num_threads() == 1 and view.span_is_contiguous()) {
+      std::fill_n(view.data(), view.span(),
+                  static_cast<typename View::value_type>(value));
+      return;
+    }
+  }
+  Kokkos::deep_copy(exec_space, view, value);
+}
+
+/**
+ * @brief Wrapper for ``Kokkos::parallel_for`` that skips fork/join when
+ * the number of threads is 1.
+ */
+template <class ExecutionSpace = Kokkos::DefaultHostExecutionSpace>
+ESPRESSO_ATTR_ALWAYS_INLINE inline void
+kokkos_parallel_range_for(auto const &name, auto start, auto end,
+                          auto const &kernel) {
+  if (Kokkos::num_threads() > 1) {
+    Kokkos::RangePolicy<ExecutionSpace> policy(start, end);
+    Kokkos::parallel_for(name, policy, kernel);
+  } else {
+    for (auto p_index = start; p_index < end; ++p_index) {
+      kernel(p_index);
+    }
+  }
+}
--- a/src/python/espressomd/script_interface.pyx
+++ b/src/python/espressomd/script_interface.pyx
@@ -643,15 +643,15 @@ class ScriptObjectList(ScriptInterfaceHelper):
     """
 
     def __getitem__(self, key):
-        return self.call_method("get_elements")[key]
+        return self.call_method(b"get_elements")[key]
 
     def __iter__(self):
-        elements = self.call_method("get_elements")
+        elements = self.call_method(b"get_elements")
         for e in elements:
             yield e
 
     def __len__(self):
-        return self.call_method("size")
+        return self.call_method(b"size")
 
 
 class ScriptObjectMap(ScriptInterfaceHelper):
@@ -677,22 +677,22 @@ class ScriptObjectMap(ScriptInterfaceHelper):
         Remove all elements.
 
         """
-        self.call_method("clear")
+        self.call_method(b"clear")
 
     def __len__(self):
-        return self.call_method("size")
+        return self.call_method(b"size")
 
     def __getitem__(self, key):
-        return self.call_method("get", key=key)
+        return self.call_method(b"get", key=key)
 
     def __setitem__(self, key, value):
-        self.call_method("insert", key=key, object=value)
+        self.call_method(b"insert", key=key, object=value)
 
     def __delitem__(self, key):
-        self.call_method("erase", key=key)
+        self.call_method(b"erase", key=key)
 
     def keys(self):
-        return self.call_method("keys")
+        return self.call_method(b"keys")
 
     def __iter__(self):
         for k in self.keys():
--- a/src/python/espressomd/script_interface.pyx
+++ b/src/python/espressomd/script_interface.pyx
@@ -567,8 +567,9 @@ class ScriptInterfaceHelper(PScriptInterface):
 
     def __init_subclass__(cls, **kwargs):
         super().__init_subclass__(**kwargs)
+        cls.__doc__, docstrings = cls._extract_method_docstrings()
         for method_name in cls._so_bind_methods:
-            cls._install_bound_method(method_name)
+            cls._install_bound_method(method_name, docstrings.get(method_name))
 
         cdef vector[string] features_vec
         if cls._so_features:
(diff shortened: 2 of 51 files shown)


src/python/espressomd/script_interface.pyx
```diff
@@ -394,24 +413,13 @@ cdef Variant python_object_to_variant(value) except *:
...
-    cdef Vector3d vec3d
-    cdef Vector4d vec4d
-    cdef cnp.ndarray[cnp.float64_t, ndim = 2] arrayNvec3d
-    cdef size_t index
-    cdef size_t nrows
+    # handle inexpensive types first
     if is_none(value):
         return None
     if is_type[cbool](value):
         return get_value[cbool](value)
```
jngrad, line 420: These object<->variant conversion functions were expensive, because Cython pulls every `cdef` variable to the top of the function. Therefore, for trivial types like `int` and `str`, we would always pay the price of several allocations of `std::vector`, `std::unordered_map`, and large C-style arrays on the stack. These expensive allocations are now only performed when really needed, by calling dedicated functions.

src/python/espressomd/script_interface.pyx
```diff
@@ -532,18 +555,51 @@ class ScriptInterfaceHelper(PScriptInterface):
...
+        if self._so_features_error:
+            raise RuntimeError(self._so_features_error)
         super().__init__(self._so_name, policy=self._so_creation_policy,
                          **kwargs)
-        self.define_bound_methods()
+
+    def __init_subclass__(cls, **kwargs):
+        super().__init_subclass__(**kwargs)
+        for method_name in cls._so_bind_methods:
+            cls._install_bound_method(method_name)
```
jngrad: This leverages [PEP 487](https://peps.python.org/pep-0487/) introduced in Python 3.6.

C++ methods are now bound once at the class level, rather than every time a new instance is created (by the now removed `def define_bound_methods(self)`). This massively reduces the overhead of Monte Carlo simulations in the proposed Python rewrite in #4765, where `ParticleHandle` objects are now created much more frequently. `ParticleHandle` is a thin wrapper than only contains an int and two pointers, but the `define_bound_methods` would walk through the list of particle properties and synthesize each one, wasting a few percentage points of the Monte Carlo runtime.

src/python/espressomd/script_interface.pyx
```diff
@@ -532,18 +555,51 @@ class ScriptInterfaceHelper(PScriptInterface):
...
+        assert method_name.isidentifier(), f"{method_name!r} isn't a suitable name for a function"  # nopep8
+        filename = f"<espressomd bound method: {cls.__qualname__}.{method_name}>"  # nopep8
+        src = (f"def {method_name}(self, **kwargs):\n"
+               f"    return self.call_method({method_name.encode()!r}, **kwargs)\n")
+        namespace = {}
+        exec(compile(src, filename, "exec"), namespace)
+        fn = namespace[method_name]
+        fn.__qualname__ = f"{cls.__qualname__}.{method_name}"
+        fn.__module__ = cls.__module__
+        setattr(cls, method_name, fn)
```
jngrad, line 666: This is the modern approach to bind methods to a class. Here is the authoritative source from `dataclasses`: [python/cpython@3.10:Lib/dataclasses.py#L426-L433](https://github.com/python/cpython/blob/3.10/Lib/dataclasses.py#L426-L433)

I have added an extra `compile(src, filename, "exec")` step, whose job is to compile a *code object*, which `exec()` can then convert to a function and bind to a name. While `eval()` processes a single statement, `exec()` can process multiple statements. Both can take a string as argument, which is the most common use case, but for a function it would lead to `co_filename = "<string>"`. In that scenario, a profiler would show all bound C++ methods of a class (e.g. system.part.add, system.part.clear, etc.) merged together inside a block called `<string>`. By adding an extra `compile()` with a custom `filename`, profilers are able to create one properly labeled block per bound C++ method, making the time graph more fine-grained and easier to analyze.

src/python/espressomd/utils.pyx
```diff
@@ -94,13 +107,18 @@ def to_bytes(s):
     s : :obj:`str` or :obj:`bytes`
 
     """
-    if isinstance(s, str):
+    if type(s) is str:
+        # dict lookup is much cheaper than conversion
+        if s in _fast_map_str2bytes:
+            return _fast_map_str2bytes[s]
+        return s.encode()
```
jngrad, line 114: The fast map is treated as a C++ object rather than a Python object, massively lowering the overhead of key lookup. The `str.encode()` method is surprisingly expensive, about 0.8% of the total runtime of a Monte Carlo simulation. This is due to the memory allocation (strings are stored on the heap, not on the stack) and memcopy.

In Python, string literals like `a = "hello"` that are ASCII are [*interned*](https://en.wikipedia.org/wiki/String_interning), i.e. stored in a central location, such that other string literals that contain the same text can re-use the same pointer and hash, thus saving a memory allocation. The hash is lazily evaluated once on the first hashing and then re-used by all strings that share the same pointer. A non-literal string would look like this: `b = "".join("hello")`, and while `a == b` is True, `a is b` is False, since they don't share the same pointer.

The hash map lookup is actually really cheap for interned strings because the hash doesn't need to be computed, and disambiguating multiple strings in the same bucket is achieved by pointer comparison. Thus, we have no string traversal during lookup. Inspection of the C++ code generated by Cython reveals the dict lookup and value retrieval decay to two C API calls, `PyDict_Contains()` and `PyDict_GetItem()`, which are the fast paths for dict objects.

The same logic doesn't apply to `to_str()`, because it works with a char pointer as argument from the C++ interface, and lookup would require creating a temporary bytestring, which would incurr a memory allocation, memcopy, and hash.

src/core/kokkos_helpers.hpp
```diff
@@ -0,0 +1,62 @@
...
+                    Kokkos::DefaultHostExecutionSpace,
+                    typename View::memory_space>::accessible) {
+    if (Kokkos::num_threads() == 1 and view.span_is_contiguous()) {
+      std::fill_n(view.data(), view.span(),
+                  static_cast<typename View::value_type>(value));
+      return;
+    }
+  }
+  Kokkos::deep_copy(exec_space, view, value);
+}
```
jngrad, line 44: `Kokkos::deep_copy` and `Kokkos::parallel_for` create a task pool even when the number of threads is 1, in which case the code is slowed down by a fork+join operation that is superfluous, since there is a single task to join. Use a serial `std::fill_n` when there is only 1 thread, and the data layout has no gap, and the data resides in host memory. For non-contiguous data (pitched memory) and device memory, fall back to the regular `Kokkos::deep_copy`, which uses specialized initialization kernels.

jngrad:
The performance gains are too small for LJ and P3M. The main winner is Monte Carlo, where the script interface overhead was substantial and the frequent cache invalidation was detrimental. Here are benchmark results for `mc_acid_base_reservoir.py` with 118 particles per thread and Debye-Hückel electrostatics on an AMD Ryzen 9 9950X3D, only showing the reaction step times (MD step times are identical in the old and new code):

| threads |  old code |new code | speed-up |
| --- | --- | --- | --- |
| 1 | 45.5 us/step | 37.2 us/step | 18% |
| 2 | 58.5 us/step | 51.3 us/step | 12% |
| 4 | 79.2 us/step | 72.2 us/step | 9%  |

This refactoring will actually be instrumental to #4765, where the Python rewrite of the reaction methods introduced massive overhead due to the frequent creation of `ParticleHandle` objects, which have to re-bind C++ methods for every new particle instance, and to the frequent calls to `call_method()` that need to marshal Python strings.

Currently, #4765 has a performance of 155.1 us/step. With these changes, the performance becomes 82.3 us/step, i.e. this PR brings a 47% speed-up. While 82.3 us/step represents a 1.8x slowdown compared to the old code, that's quite reasonable considering all the Monte Carlo logic is now carried out in Python.

This PR can be backported to ESPResSo 5.0, albeit with less spectacular speed-ups:

| threads |  old code |new code | speed-up |
| --- | --- | --- | --- |
| 1 | 63.3 us/step | 55.4 us/step | 13% |
| 2 | 96.8 us/step | 90.2 us/step | 7% |
| 4 | 151.4 us/step | 143.9 us/step | 5%  |

notifying benchmark results to @pm-blanco

AI disclosure: Opus 4.8 was used to analyze py-spy profiles and suggest refactoring operations. All refactoring was done manually and profiled to confirm performance gains. Suggestions that would degrade the maintainability of ESPResSo were rejected, e.g. providing a new CMake option to build in Serial mode instead of OpenMP mode, or implementing custom Kokkos containers that already exist in Cabana.

jngrad:
In the original code, synthetic methods had their docstrings written inside the class docstring. This works fine with Sphinx, but not with Jupyter notebooks and the Python `help()` function. The new code now extracts the method docstrings from the class docstring and binds them to the object functions.

MWE:
```python
import espressomd.observables
obs = espressomd.observables.Energy()
acc = espressomd.accumulators.Correlator(
    obs1=obs, corr_operation="scalar_product", tau_max=10, tau_lin=2, delta_N=1)
help(acc.finalize)
```

**Old behavior:**

`help(acc.finalize)`:

> Help on method finalize in module espressomd.accumulators:
> 
> finalize(**kwargs) method of espressomd.accumulators.Correlator instance

Typing `acc.finalize(` followed by Shift+Tab in JupyterLab:

> Signature: acc.finalize(**kwargs)
> Docstring: <no docstring>
> File:      /work/jgrad/espresso/build/\<espressomd bound method: Correlator.finalize\>
> Type:      method

**New behavior:**

`help(acc.finalize)`:

> Help on method finalize in module espressomd.accumulators:
> 
> finalize(**kwargs) method of espressomd.accumulators.Correlator instance
>     Trigger a compression for the last remaining values in the buffer.
>     After that, the correlator can no longer be updated.

Typing `acc.finalize(` followed by Shift+Tab in JupyterLab:

> Signature: acc.finalize(**kwargs)
> Docstring:
> Trigger a compression for the last remaining values in the buffer.
> After that, the correlator can no longer be updated.
> File:      /work/jgrad/espresso/build/\<espressomd bound method: Correlator.finalize\>
> Type:      method

hidekb (approved):
I understand this PR reduce communication overhead and object <-> variant conversions overhead, and can show the docstring of bound C++ method in jupyter and Python `help()` function.
LGTM!

PR #5415, opened 2026-08-11
Fix stale hdf5 superblock

jngrad:
hdf5 files are now properly closed by the `H5md.close()` method to avoid stale EOA in the superblock.


hidekb (approved):
I understand this PR avoid stale EOA in the superblock. LGTM!

PR #5417, opened 2026-08-14
Make strategic Utils::Vector operations noexcept

jngrad:
Kokkos kernels are automatically labeled with the `noexcept` specifier, since exception propagation is not possible on device code and inside OpenMP parallel regions. Add `noexcept` specifiers and `std::floating_point` constraints in key `BoxGeometry` kernels and `Utils::Vector` kernels and constructors. When a throw statement is accidentally introduced in one of those `noexcept` functions, compiler diagnostics will flag the regression. For debug and coverage builds, using a `noexcept` constructor over a regular one also allows elision of the exception handling block in the assembly code, saving one comparison operation per instantiation. The speed-up is minimal, about 1% for a pure LJ simulation with an `-Og` build. Finally, rely on synthetic generation of `Utils::Vector` comparison functions using C++20.

### DIFF
--- a/doc/sphinx/running.rst
+++ b/doc/sphinx/running.rst
@@ -1289,7 +1289,7 @@ Pass option ``--native`` to capture C++ function names; to get accurate traces
 with minimal overhead, build |es| with
 ``-D CMAKE_CXX_FLAGS="-g -fno-omit-frame-pointer" -D CMAKE_BUILD_TYPE=Release``.
 
-To record samples in a format that can be analyzed by AI agents, run::
+To record samples in a format that can be analyzed by AI agents, run:
 
 .. code-block:: bash
 
--- a/src/core/BoxGeometry.hpp
+++ b/src/core/BoxGeometry.hpp
@@ -28,6 +28,7 @@
 #include <bitset>
 #include <cassert>
 #include <cmath>
+#include <concepts>
 #include <limits>
 #include <stdexcept>
 #include <utility>
@@ -50,8 +51,9 @@ namespace detail {
  * @return Shortest distance from @p b to @p a across periodic images,
  *         i.e. <tt>a - b</tt>. Can be negative.
  */
-template <typename T>
-T get_mi_coord_masked(T a, T b, T box_length, T box_length_inv_masked) {
+template <std::floating_point T>
+T get_mi_coord_masked(T a, T b, T box_length,
+                      T box_length_inv_masked) noexcept {
   auto const dx = a - b;
   return dx - std::rint(dx * box_length_inv_masked) * box_length;
 }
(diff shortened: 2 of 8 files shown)


jngrad:
I deliberately did not mark `operator+(Vector<T, N> const &, Vector<U, N> const &)` and other mathematical operations on vectors as `noexcept` because the syntax is really complex. Here is a possible implementation:

```c++
namespace detail {
template <typename R, std::size_t N>
inline constexpr bool nothrow_elementwise_v =
    std::is_nothrow_move_assignable_v<R> and
    std::is_nothrow_move_constructible_v<Vector<R, N>>;
}

#define ESPRESSO_VECTOR_BINARY_OP_NOEXCEPT(op)                                 \
  noexcept(                                                                    \
      noexcept(std::declval<T const &>() op std::declval<U const &>()) and     \
      detail::nothrow_elementwise_v<                                           \
          decltype(std::declval<T const &>() op std::declval<U const &>()), N>)

template <std::size_t N, typename T, typename U>
auto operator+(Vector<T, N> const &a, Vector<U, N> const &b)
    ESPRESSO_VECTOR_BINARY_OP_NOEXCEPT(+) {
  return detail::binary_op(a, b, std::plus<>());
}

#undef ESPRESSO_VECTOR_BINARY_OP_NOEXCEPT
```

It is a lot of boilerplate code that is difficult to maintain. For example, the `std::declval<T const &>() op std::declval<U const &>()` is meant to express `using R = T + U` outside the function body. It is also not transferable to in-place operations like `+=`, where a dedicated syntax is needed. It is also conservative, rather than exact, and we cannot use this reliably in an expression of the type `noexcept(noexcept(vec_a + vec_b))` for non-arithmetic types `T` and `U`. Note: in this kind of expression, the first `noexcept` is the specifier and the second `noexcept` is an operator that returns `true` if the expression is guaranteed to be fully `noexcept`. The STL `<functional>` header did not bother to make `std::plus` and the other operators conditionally `noexcept`, probably for the same reason.

For all mathematical operations, I let the compiler decide recursively if the functions are `noexcept`. That's trivial here because everything is inlined. Compiler diagnostics will detect when an uncaught exception is reachable from within a `noexcept` function (like a Kokkos kernel) but if the throw statement is hiding within several layers of function calls, diagnostics won't be able to show the offending line. The handful of `noexcept` that are introduced by this PR will help us detect accidental regressions with a finer granularity in Vector-related code. Note: a `noexcept` function is allowed to call a function that throws an exception, in which case the exception calls `std::terminate` instead of the stack unwinding mechanism. Kokkos kernels cannot really support exceptions anyway.

hidekb (approved):
LGTM!

PR #5421, opened 2026-09-01
Code cleanup

jngrad:
Description of changes:
- replace `&*buffer.end()` with `std::to_address(buffer.end())` to avoid dereferencing past-the-end iterators
- modernize code based on new features in C++20, remove unused code, add missing headers
- reduce `particle_node.hpp` footprint

### DIFF
--- a/src/core/MpiCallbacks.hpp
+++ b/src/core/MpiCallbacks.hpp
@@ -84,7 +84,7 @@ auto invoke(F f, boost::mpi::packed_iarchive &ia) {
 
   /* This is the local receive buffer for the parameters. We have to strip
      away const so we can actually deserialize into it. */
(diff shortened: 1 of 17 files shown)


jngrad:
The ongoing project of removing `particle_node.cpp` to allow multiple ESPResSo systems has revealed tight coupling between `particle_node.cpp`  and `CellStructure.cpp`. In the end, everything should be part of `CellStructure`. This PR disentangles some of the obvious coupling and reduces the noise in the `particle_node.cpp` removal branch.

The sources of UB are non-critical and eliminated by compilers already at `-O0`. Even our UBSAN builds ignore them. Nevertheless, they are technically not allowed by the C++ standard and can potentially trip other UB detection tools. These `std::to_address()` improvements will be backported to 5.0.2.

itischler (approved):
looks good, thanks

PR #5424, opened 2026-09-05
Assign ELCIC mirror forces to the correct particles

reinaual:
I've stumbled upon another subtle bug in ELCIC force calculation.

When we add the `dielectric_layers_contribution(...)`, we mirror the particle1 position and calculate the distance to particle2. However, we then assign the force to particle 1, which is wrong - it's the force on particle 2.

elc_image_force_bug


jngrad (approved):
LGTM

PR #5428, opened 2026-09-16
Fix Propagation enum printing as raw int

RudolfWeeber:
Since Python 3.11, enum.IntFlag mixes in ReprEnum, which makes str()/print()/f-strings fall back to int.__str__() instead of the descriptive flag name. As a result, printing a particle's propagation property showed a bare integer (e.g. "4") even though isinstance(particle.propagation, Propagation) was already True.

Override __str__ on Propagation to restore flag-name formatting, and add a regression test asserting the string representation for a single flag, a combination, and NONE.


RudolfWeeber:
@RebeccaStephan

jngrad:
Note: the CI pipeline failed due to a maintenance operation on the electrical circuits of the CI machines. We will restart the CI pipeline tomorrow, once the maintenance operation is over.

PR #5429, opened 2026-09-16
Allow free rotation of virtual sites, expose default propagation

RudolfWeeber:
* Allow relative virtual sites to rotate freely by Euler or Langevin equation. This is done by assiging `Propagatio.ROT_EULER` or `Propagation.ROT_LANGEVIN` in addition to `Propagation.TRANS_VS_RELATIVE`
* count freely rotating virtual sites' rotational kinetic energy
* Add `System.integrator.default_propagation` so users can actually tell what that propagation flag resolves to depending on active integrators and thermostats.

Note that this PR contains teh propagation interface fix PR. I separated them, because the interface fix should go into 5.0.2 but this one into 5.1.


RudolfWeeber:
@RebeccaStephan

PR #5432, opened 2026-09-18
Improve support for CUDA 13.4 and Clang 22

jngrad:
Description of changes:
- silence C29-specific warnings from the Boost unit tests framework
- add missing header includes, in particular nvcc 13.4 no longer indirectly includes `<cstdint>` via `<memory>`


hidekb (approved):
LGTM! Thank you for the good work.

